PR #9, opened 2016-12-05
Drop Rails 4 support

jcoyne:


jcoyne:
@jeremyf I'm reducing the amount of variation we need to support.  Rails 5 has been out for 6 months now, so there's been plenty of time to shake out the bugs.  As Hyrax is a new product, I'd like to take the opportunity to start fresh.

jeremyf:
@jcoyne That's good by me!

mjgiarlo:
I'll add another reason: to strip out our janky polling solution for notifications and replace it with ActionCable.

PR #13, opened 2016-12-06
Make transfers and proxies optional features

jcoyne:


mjgiarlo:
What about the UI bits? Should those be behind a Flipflop check too, or do you think that's a follow-up PR?

jcoyne:
@mjgiarlo I'd say let's follow up with that.  I would hope that a good portion of the UI is already affected by `can? :create, TransferRequest`

mjgiarlo:
@jcoyne Ah, yes, worth QAing.

config/features.rb
```diff
@@ -4,6 +4,14 @@
   strategy :active_record, class: Hyrax::Feature
   strategy :default
 
+  feature :proxy_deposit,
+          default: true,
+          description: "Depositors may designate proxies to deposit works on their behalf"
+
+  feature :transfer,
```
jeremyf: I believe an additional modifier for the feature name could be helpful: `transfer_works`

spec/models/flipflop_spec.rb
```diff
@@ -5,8 +5,19 @@
     subject { described_class.assign_admin_set? }
     it { is_expected.to be true }
   end
+
   describe "enable_mediated_deposit?" do
     subject { described_class.enable_mediated_deposit? }
     it { is_expected.to be false }
   end
+
+  describe "proxy_deposit?" do
```
jeremyf: The spec you are describing appears to be `default value of proxy_deposit?` It may help legibility to explain that.

spec/models/flipflop_spec.rb
```diff
@@ -5,8 +5,19 @@
...
+    subject { described_class.proxy_deposit? }
+    it { is_expected.to be true }
+  end
+
+  describe "transfer?" do
```
jeremyf: The spec you are describing appears to be `default value of transfer?` It may help legibility to explain that.

PR #35, opened 2016-12-14
Restyle the banner so the gray doesn't cover the tusks

jcoyne:
screen shot 2016-12-14 at 10 59 13 am


assets/css/style.css
```diff
@@ -708,10 +708,10 @@
 
 	#banner {
 		background-image: url("../images/hyrax_logo_horizontal_white_background.png");
-		background-position: center center;
-		background-size: contain;
-                background-repeat: no-repeat;
-		height: 28em;
+		background-position: 50% 10px;
+		background-size: auto 140px;
+    background-repeat: no-repeat;
```
jcoyne: Works for me:
screen shot 2016-12-14 at 11 00 21 am

I think this is because the templates you have used tabs rather than spaces.

PR #39, opened 2016-12-15
Removes duplicate chevron from facets.

andjsmit:
Fixes https://github.com/projecthydra-labs/hyku/issues/532 (Partially)

Current admin dashboard uses li for menus but with the same classes as the div menus in facets. Styles for the chevrons are clashing. Adding li to CSS separates the two.

Same change will be needed in hyku as well.

@projecthydra/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_dashboard.scss
+++ b/app/assets/stylesheets/hyrax/_dashboard.scss
@@ -123,7 +123,7 @@ ul.collapsing > li > a {
   padding-left: 40px;
 }
 
-.collapse-toggle {
+.admin-sidebar .collapse-toggle {
   margin-bottom: 0;
 
   &::after {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/hyrax/_dashboard.scss
```diff
@@ -123,7 +123,7 @@ ul.collapsing > li > a {
   padding-left: 40px;
 }
 
-.collapse-toggle {
+li > .collapse-toggle {
```
jcoyne: Can we use `.nav-stacked` (or add a new class to the admin sidebar) instead of `li >`. I think that makes it clearer that it's for the admin menu.

andjsmit: Sounds like a better approach. I was trying too hard to walk lightly on the code. Code updated. I'll push the class to hyku as well if this PR gets merged.

mjgiarlo:
@jcoyne re-review?

PR #40, opened 2016-12-15
Breaking out visibility notes into a different I18n string

### LINKED ISSUE #32: Translation field is missing parameter
I believe any calls to the `t` method should pass the parameter `type: document.human_readable_type` (See [Code]( https://github.com/projecthydra-labs/hyrax/blob/062c3e3592ec6b97c578c6838b4be56657a03285/app/helpers/hyrax/hyrax_helper_behavior.rb#L221-L229))

new_generic_work____hyrax

Discovered in #31

danhorst:
Fixes #32 ; refs #32

Fixing interpolation of internationalization terms for visibility labels. Bonus: removing descriptive text from the `label_html` term. Now you can call `label_html` and just get the label. 🎉 

@projecthydra/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_form_visibility_component.html.erb
+++ b/app/views/hyrax/base/_form_visibility_component.html.erb
@@ -10,7 +10,7 @@
           <label>
             <%= f.radio_button :visibility, Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC, data: { 'target': '#collapsePublic' }  %>
             <%= t('hyrax.visibility.open.label_html') %>
-            <%= t('hyrax.visibility.open.note', type: f.object.human_readable_type).html_safe %>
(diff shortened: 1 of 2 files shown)


app/views/hyrax/base/_form_visibility_component.html.erb
```diff
@@ -10,6 +10,7 @@
           <label>
             <%= f.radio_button :visibility, Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC, data: { 'target': '#collapsePublic' }  %>
             <%= t('hyrax.visibility.open.label_html') %>
+            <%= t('hyrax.visibility.open.note', type: f.object.human_readable_type).html_safe %>
```
jcoyne: if we make the i18n key `note_html`, then the `html_safe` will not be required.

PR #41, opened 2016-12-15
Generate Spanish locale file into applications when creating a new work type

mjgiarlo:


lib/generators/hyrax/work/templates/locale.es.yml.erb
```diff
@@ -0,0 +1,37 @@
...
+      <%= file_name %>:     'fa fa-file-text-o'
+    select_type:
+      <%= file_name %>:
+        # TODO: translate `human_name` into Spanish
+        name:               "<%= human_name.titleize %>"
+        # TODO: translate `human_name` into Spanish
+        description:        "<%= human_name %> trabajos"
+  simple_form:
+    labels:
+      <%= file_name %>:
```
jcoyne: Shouldn't we just be relying on the default translations?

mjgiarlo: @jcoyne I'm not sure. We did this for a reason, I assume: https://github.com/projecthydra/sufia/commit/f6decc18996b78c3a59c9371b89761cccb6614b8

mjgiarlo: The drawback is that we potentially have multiple places to fix/add translations. The benefit is that downstream users already have the translations in their apps, which makes it easier to tweak.

mjgiarlo:
@jcoyne Build should go :green_heart: momentarily. Re-review?

PR #45, opened 2016-12-15
Add default descriptions for metadata fields, so that downstream apps can fallback to good defaults

mjgiarlo:


mjgiarlo:
@jcoyne Those lines do differ from the other, admittedly very simple hints, in that they say `collection` instead of `work`.

PR #53, opened 2016-12-16
Fix routing for featured works on the home page. Fixes #548

jcoyne:


jcoyne:
@mjgiarlo those partials are covered by view tests. However view tests are not aware of engine routing 😞

mjgiarlo:
@jcoyne new feature spec? *winces*

PR #64, opened 2016-12-19
Merge latest from Sufia

mjgiarlo:


mjgiarlo:
@jcoyne last travis build running. This is ready for review (in case I messed up the merge).

spec/services/hyrax/actor_factory_spec.rb
```diff
@@ -13,9 +13,10 @@
                          Hyrax::Actors::AttachFilesActor,
                          Hyrax::Actors::ApplyOrderActor,
                          Hyrax::Actors::InterpretVisibilityActor,
+                         Hyrax::DefaultAdminSetActor,
+                         Hyrax::Actors::InitializeWorkflowActor,
                          Hyrax::ApplyPermissionTemplateActor,
-                         Hyrax::Actors::GenericWorkActor,
-                         Hyrax::Actors::InitializeWorkflowActor]
+                         Hyrax::Actors::GenericWorkActor]
```
jcoyne: This seems like a problem. Why did you move Initialize up 3 steps?

spec/services/hyrax/actor_factory_spec.rb
```diff
@@ -30,9 +31,10 @@
         Hyrax::Actors::AttachFilesActor,
         Hyrax::Actors::ApplyOrderActor,
         Hyrax::Actors::InterpretVisibilityActor,
+        Hyrax::DefaultAdminSetActor,
+        Hyrax::Actors::InitializeWorkflowActor,
```
jcoyne: Why is Initialize moving up?

app/services/hyrax/actor_factory.rb
```diff
@@ -9,9 +9,10 @@ def self.stack_actors(curation_concern)
        Hyrax::Actors::AttachFilesActor,
        Hyrax::Actors::ApplyOrderActor,
        Hyrax::Actors::InterpretVisibilityActor,
+       DefaultAdminSetActor,
+       Hyrax::Actors::InitializeWorkflowActor,
```
jcoyne: Why is Initialize moving up?  Don't we need to save the model before we activate the workflow?

mjgiarlo: The order of actors in Hyrax is different from the order of actors in Sufia. See https://github.com/projecthydra/sufia/blob/master/app/services/sufia/actor_factory.rb. I went with the order from Sufia where I've seen workflow working before rather than Hyrax. Do you think Sufia is currently wrong too?

mjgiarlo:
@jcoyne Fixed. Re-review?

PR #194, opened 2016-12-31
Fix application template link

mark-dce:
Link needs to point to 'projecthydra-labs' instead of 'projecthydra'
https://raw.githubusercontent.com/projecthydra-labs/hyrax/master/template.rb

PR #196, opened 2017-01-03
Fix rails template link in README.md

no-reply:
The rails template link had fallen out of date, pointing to `projecthydra` instead of `projecthydra-labs`. The update fixes the app generation instructions.


jcoyne:
Duplicate of #194

PR #209, opened 2017-01-05
Adding a link to the work into the notification message

### LINKED ISSUE #197: Include a link in notification when changes are required
### Descriptive summary

In Hyrax master, when an item is not approved and changes are required the notification should include a link to the item needing changes. https://github.com/projecthydra-labs/hyrax/blob/master/app/services/hyrax/workflow/changes_required_notification.rb

This is needed becuase the user does not have any access to the work via their My Work Listing or in the Catalog listings.  Since I could not access the work via the notification or the listing I was left hacking at the URLs.

### Expected behavior

The notification should include a link to the item needing changes.

`Your deposit requires changes	Test Admin Set (http://server.url/concern/generic_works/3n583xt00x) requires additional changes before approval. `

### Actual behavior

`Your deposit requires changes	Test Admin Set (3n583xt00x) requires additional changes before approval. `

1. Upload a work to an Admin Set that requires approval
1. Then Request Changes before Approval (on the work screen)
1. Then check out your notifications

### Related work

cam156:
Fixes #197 

Includes a link to the work in the workflow change request notification

screen shot 2017-01-05 at 1 57 15 pm

### DIFF
--- a/app/services/hyrax/workflow/abstract_notification.rb
+++ b/app/services/hyrax/workflow/abstract_notification.rb
@@ -13,6 +13,7 @@ module Hyrax
         @comment = comment.respond_to?(:comment) ? comment.comment.to_s : ''
         @recipients = recipients
         @user = user
+        @entity = entity
       end
 
       def call
@@ -27,6 +28,16 @@ module Hyrax
 
         def message
           "#{title} (#{work_id}) was advanced in the workflow by #{user.user_key} and is awaiting approval #{comment}"
+        end
+
+        # @return [ActiveFedora::Base] the document (work) the the Abstract WorkFlow is creating a notification for
+        def document
+          @entity.proxy_for
+        end
+
+        def document_path
+          key = document.model_name.singular_route_key
+          Rails.application.routes.url_helpers.send(key + "_path", document.id)
         end
 
       private
--- a/app/services/hyrax/workflow/changes_required_notification.rb
+++ b/app/services/hyrax/workflow/changes_required_notification.rb
@@ -1,6 +1,8 @@
 module Hyrax
   module Workflow
     class ChangesRequiredNotification < AbstractNotification
+      include ActionView::Helpers::UrlHelper
+
       protected
 
         def subject
@@ -8,13 +10,13 @@ module Hyrax
         end
 
         def message
-          "#{title} (#{work_id}) requires additional changes before approval.\n\n '#{comment}'"
+          "#{title} (#{link_to work_id, document_path}) requires additional changes before approval.\n\n '#{comment}'"
         end
 
       private
 
         def users_to_notify
-          user_key = ActiveFedora::Base.find(work_id).depositor
+          user_key = document.depositor
           super << ::User.find_by(email: user_key)
         end
     end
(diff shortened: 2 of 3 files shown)


app/services/hyrax/workflow/changes_required_notification.rb
```diff
@@ -8,15 +12,24 @@ def subject
         end
 
         def message
-          "#{title} (#{work_id}) requires additional changes before approval.\n\n '#{comment}'"
+          "#{title} (#{link_to work_id, document_path}) requires additional changes before approval.\n\n '#{comment}'"
         end
 
       private
 
+        def solr_document
```
jcoyne: Instead of loading the solr document, could we just add an accessor for the `entity` (from https://github.com/projecthydra-labs/hyrax/blob/master/app/services/hyrax/workflow/abstract_notification.rb#L11)?

cam156: @jcoyne I gave that a try originally and the entity seems to be some kind of sipity item, not a solr document.  The original code was loading the document from fedora.

cam156:
@mjgiarlo I'm happy to move the code into the abstract notification.  It seems like a good fit.

app/services/hyrax/workflow/changes_required_notification.rb
jcoyne: Ok. Thanks for checking. Would you mind documenting that?

cam156:
@jcoyne Moved to use the entity proxy_for.  @mjgiarlo moved the code to the abstract class.

app/services/hyrax/workflow/abstract_notification.rb
```diff
@@ -29,6 +30,15 @@ def message
           "#{title} (#{work_id}) was advanced in the workflow by #{user.user_key} and is awaiting approval #{comment}"
         end
 
+        def document
```
jcoyne: Please add YARD doc about what this returns.

cam156: @jcoyne technically it is protected so it does not need yard doc, but I can certainly add some.

app/services/hyrax/workflow/abstract_notification.rb
```diff
@@ -29,6 +30,16 @@ def message
           "#{title} (#{work_id}) was advanced in the workflow by #{user.user_key} and is awaiting approval #{comment}"
         end
 
+        # @return [document] the document (work) the the Abstract WorkFlow is creating a notification for
```
jcoyne: The thing in the square brackets should be a class or list of classes.  So is it a `ActiveFedora::Base`?

cam156: @jcoyne Unfortunately I do not know. The public method for Entity does not have any yard comments, so I can not be certain what it is returning in all cases.  For the current case it returns a GenericWork, but I can not see how that would always be the case.  https://github.com/projecthydra-labs/hyrax/blob/master/app/models/sipity/entity.rb#L28

Maybe you have a better idea...

PR #222, opened 2017-01-09
Adding link to referenced work in all workflow notifications

### LINKED ISSUE #216: Workflow notifications should contain links to the item
### Descriptive summary

When I receive a notification of a workflow event it would be nice to include a link directly to the item that needs action.

See https://github.com/projecthydra-labs/hyrax/blob/master/app/services/hyrax/workflow/changes_required_notification.rb#L13 for example of how to include a link in the notification.

cam156:
Fixes #216 

Adds a link to the workflow notifications instead of just the work id.

screen shot 2017-01-09 at 1 25 37 pm

### DIFF
--- a/app/services/hyrax/workflow/abstract_notification.rb
+++ b/app/services/hyrax/workflow/abstract_notification.rb
@@ -1,6 +1,8 @@
 module Hyrax
   module Workflow
     class AbstractNotification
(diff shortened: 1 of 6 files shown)


jcoyne:
On of the builds failed. I'm restarting it.

PR #234, opened 2017-01-10
Unify deposited/complete states from two baked-in workflows

mjgiarlo:


mjgiarlo:
@jcoyne @jeremtf I wonder if this is going to cause issues with tenants/instances that have data in the old "complete" state. Do we need a rake task to deal with this, or something?

Issue #224, opened 2017-01-10
add a "no state change" radio


mjgiarlo:
<a href="https://github.com/jcoyne"><img src="https://avatars.githubusercontent.com/u/92044?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jcoyne](https://github.com/jcoyne)**
_Friday Sep 23, 2016 at 15:12 GMT_
_Originally opened as https://github.com/projecthydra/curation_concerns/issues/1027_

----

This allows a user to leave a comment without changing the state.
- [x] Update [CurationConcerns::Workflow::WorkflowSchema](https://github.com/projecthydra/curation_concerns/blob/4760ee0d7019059702a1c9413bdf28a599ff41b7/app/services/curation_concerns/workflow/workflow_schema.rb#L13) (`required(:transition_to)` to be `optional(:transition_to)`)
- [x] Update [CurationConcerns::Forms:: WorkflowActionForm](https://github.com/projecthydra/curation_concerns/blob/4760ee0d7019059702a1c9413bdf28a599ff41b7/app/forms/curation_concerns/forms/workflow_action_form.rb#L55 to conditionally update the workflow state of an object.
- [ ] Update initial workflow to include "No State Change" radio button; This would be done by defining the "no_state_change" action that is available from each state for the given roles that should have access in [workflow.json.erb](https://github.com/projecthydra/curation_concerns/blob/workflow/lib/generators/curation_concerns/work/templates/workflow.json.erb) but not including the `transition_to` link.
  - See [CurationConcerns::Workflow::WorkflowSchema](https://github.com/projecthydra/curation_concerns/blob/workflow/app/services/curation_concerns/workflow/workflow_schema.rb#L19-L24) for guidance concerning schema definition.

mjgiarlo:
<a href="https://github.com/mjgiarlo"><img src="https://avatars.githubusercontent.com/u/131982?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mjgiarlo](https://github.com/mjgiarlo)**
_Tuesday Jan 10, 2017 at 17:33 GMT_

----

Inclined to take this up again, @jeremyf ?

mjgiarlo:
@vantuyls Do you want to slot this into MD milestone 2 or 3, or leave in its current milestone?

mjgiarlo:
This was closed by #238, but that PR did not reference this issue directly.

PR #238, opened 2017-01-11
Adding no_state_change action to default workflow

### LINKED ISSUE #244: Collect new and returning users per week from Google Analytics
Display these numbers as a bar chart in the "User activity" section of the "activity summary"

See projecthydra-labs/hyku#431 for mockup.

jeremyf:
For states in which the role can make a state change, add a new action
that allows is non-state changing.

Closes #244

### DIFF
--- a/lib/generators/hyrax/templates/mediated_deposit_workflow.json.erb
+++ b/lib/generators/hyrax/templates/mediated_deposit_workflow.json.erb
@@ -58,6 +58,12 @@
                             "to": ["approving"]
                         }
                     ]
+                }, {
+                    "name": "no_state_change",
+                    "from_states": [
+                        { "names": ["pending_review", "deposited"], "roles": ["approving"] },
+                        { "names": ["changes_required"], "roles": ["depositing"] }
+                    ]
                 }
             ]
         }
--- a/spec/features/workflow_state_changes_spec.rb
+++ b/spec/features/workflow_state_changes_spec.rb
@@ -1,0 +1,59 @@
+require 'spec_helper'
+
+RSpec.describe "Workflow state changes", type: :feature do
+  let(:workflow_name) { 'with_comment' }
+  let(:approving_user) { create(:user) }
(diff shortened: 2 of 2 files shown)


mjgiarlo:
Since #224 is about letting users leave comments without changing states, is this sufficient to close the ticket? I'm wondering specifically how this will look in the UI, whether this change communicates to users that it's a mechanism for leaving a comment independent of a state change.

jeremyf:
@mjgiarlo The state could be "leave_a_comment" instead

mjgiarlo:
@jeremyf That might be clearer for users, so :+1:.

lib/generators/hyrax/templates/mediated_deposit_workflow.json.erb
```diff
@@ -58,6 +58,12 @@
                             "to": ["approving"]
                         }
                     ]
+                }, {
+                    "name": "leave_a_comment",
```
jcoyne: Are there any states that come from this state?  I'm not sure why this should be a terminal state?

mjgiarlo: This is a new action, right, not a state? The way I read this change is that `leave_a_comment` is a new action that can be triggered from three current states, and the new action doesn't trigger a state change. Have you tested this in the UI to ensure it does what you expect, @jeremyf?

jeremyf: @mjgiarlo You read the data correctly. I can write a feature test for this particular behavior if that would be more appropriate (I shudder at UI tests for data driven behavior as its verifying a certain moment, but not verifiable over the course of the project).

jcoyne:
@jeremyf I wonder if these new failures are due to ActiveFedora 11.1.0.  There was some work done around the logging implementation in that release.

jeremyf:
@jcoyne Perhaps :(

PR #239, opened 2017-01-11
Add resource sync changelist

### LINKED ISSUE #61: ResourceSync implementation should support change lists
Decomposed from projecthydra/sufia#2915.

Add support for change lists: http://www.openarchives.org/rs/1.0.9/resourcesync#DesChanges

jcoyne:
Fixes #61

### DIFF
--- a/app/search_builders/hyrax/abstract_type_relation.rb
+++ b/app/search_builders/hyrax/abstract_type_relation.rb
@@ -1,0 +1,44 @@
+module Hyrax
+  class AbstractTypeRelation < ActiveFedora::Relation
+    def initialize(opts = {})
+      super(DummyModel, opts)
+    end
+
+    def allowable_types
+      raise NotImplementedException, "Implement allowable_types in a subclass"
+    end
+
+    def equivalent_class?(klass)
+      allowable_types.include?(klass)
+    end
+
+    def search_model_clause
+      clauses = allowable_types.map do |k|
+        ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: k.to_s)
+      end
+      clauses.size == 1 ? clauses.first : "(#{clauses.join(' OR ')})"
+    end
+
+    class DummyModel
+      def self.primary_concern
+        Hyrax.config.curation_concerns.first
+      end
+
+      def self.delegated_attributes
+        primary_concern.delegated_attributes
+      end
+
+      def self.solr_query_handler
+        primary_concern.solr_query_handler
+      end
+
+      def self.default_sort_params
+        primary_concern.default_sort_params
+      end
+
+      def self.id_to_uri(*args)
+        primary_concern.id_to_uri(*args)
+      end
+    end
+  end
+end
--- a/app/search_builders/hyrax/exposed_models_relation.rb
+++ b/app/search_builders/hyrax/exposed_models_relation.rb
@@ -1,0 +1,8 @@
+module Hyrax
+  # A relation that scopes to all user visible models (e.g. works + collections + file sets)
+  class ExposedModelsRelation < AbstractTypeRelation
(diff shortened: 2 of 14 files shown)


app/search_builders/hyrax/abstract_type_relation.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def equivalent_class?(klass)
+      allowable_types.include?(klass)
+    end
+
+    def search_model_clause
+      clauses = allowable_types.map do |k|
+        ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: k.to_s)
+      end
+      clauses.size == 1 ? clauses.first : "(#{clauses.join(' OR ')})"
```
dbrower: Doesn't `clauses.join(' OR ')` do the right thing when there is only one item in the list?

app/search_builders/hyrax/exposed_models_relation.rb
```diff
@@ -0,0 +1,8 @@
+module Hyrax
+  # A relation that scopes to all user visible models (e.g. works + collections + file sets)
+  class ExposedModelsRelation < AbstractTypeRelation
+    def allowable_types
+      Hyrax.config.curation_concerns + [Collection, ::FileSet]
```
mjgiarlo: Why collection and not admin set?

app/search_builders/hyrax/abstract_type_relation.rb
mjgiarlo: Yes, but the false expression in the ternop also builds a string with parentheses.

dbrower: And it is an error to include parentheses if there is not an `OR`...? this is only a nitpick.

jcoyne: @dbrower this isn't logic I'm changing. I've just moved it from a concrete class to an abstract class.

app/search_builders/hyrax/exposed_models_relation.rb
jcoyne: We've never exposed AdminSets via resource synch before, but we have Collections. I'm trying to be consistent with what is already there.

app/search_builders/hyrax/abstract_type_relation.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    class DummyModel
```
jeremyf: Is this class something that we won't be exposing to the public? If so, what are your thoughts on adding a `private_constant :DummyModel` after the close of the class?

lib/hyrax/resource_sync/change_list_writer.rb
```diff
@@ -0,0 +1,90 @@
+module Hyrax
+  module ResourceSync
+    # TODO: the big assumption I'm making here is that the repository has fewer
+    # than 50,000 resources to list. The Sitemap protocol is limited at 50,000
+    # items, so if we require more than that, we must have multiple Change
+    # lists and add a Change List Index to point to all of them.
+    class ChangeListWriter
```
jeremyf: Can we use pagination for generating sitemap generation?

I was crawling this site http://grognardia.blogspot.com/sitemap.xml when I discovered the sitemap leveraged pagination.

jcoyne: We absolutely need too, but each page can have 50k objects. See https://github.com/projecthydra-labs/hyrax/issues/59 if you'd like to work on that.

app/search_builders/hyrax/abstract_type_relation.rb
jcoyne: This PR was already merged, but feel free to add that.

PR #264, opened 2017-01-12
Updating workflow load to delete removed states and actions.

### LINKED ISSUE #171: Modifications to workflows cannot happen if there are items already in the workflow
<a href="https://github.com/vantuyls"></img></a> **Issue by [vantuyls](https://github.com/vantuyls)**
_Friday Sep 02, 2016 at 16:34 GMT_
_Originally opened as https://github.com/projecthydra/sufia/issues/2623_

----

If there are items currently in a Mediated Deposit Workflow, no changes can be made to the workflow until in-process items are cleared out. 

This is a first step towards a smarter way of being able to modify workflows with items in-process. In the future, we're hoping to be able to modify workflows with items in-process without orphaning or otherwise "screwing up" those in-process items. 

Decomposes #2442

cam156:
Also to error if a state has active entities and is being deleted.

Fixes #171

This PR does two things: remove unused actions and remove unused states.  It stemmed from not wanting to update a workflow and remove a state if it had any items currently in that state.  While looking into that problem I realized load was only additive and old actions and old states just remained after a load.

The deletion code for the states is what actually solves #171.  Before a workflow is updated the states that the workflow transition to are checked against the new configuration transition to.  If any transition to state is removed from the new configuration the state.entities.count is checked to make sure there are no active entities in the state. 

If there are any spanish speaking code reviewers that can check my es translation.  I just depended on google translate.  I was guessing it was better to have bad spanish than no spanish, but correct me if we do not want to take that stance.

### DIFF
--- a/app/services/hyrax/workflow/sipity_actions_generator.rb
+++ b/app/services/hyrax/workflow/sipity_actions_generator.rb
@@ -9,6 +9,7 @@ module Hyrax
       # @param workflow [Sipity::Workflow]
       # @param actions_configuration [Hash] as defined in Hyrax::Workflow::WorkflowSchema
       # @return [Sipity::Workflow]
+      # @raise [Hyrax::Workflow::InvalidStateRemovalException] Trying to remove a state that is in use
       def self.call(**keywords, &block)
         new(**keywords, &block).call
       end
@@ -30,7 +31,8 @@ module Hyrax
 
       public
 
-      # @raises [Hyrax::Workflow::InvalidStateRemovalException] Trying to remove a state that is in use
+      # @return [Sipity::Workflow]
+      # @raise [Hyrax::Workflow::InvalidStateRemovalException] Trying to remove a state that is in use
       def call
         generate_state_diagram!
         workflow
@@ -40,10 +42,10 @@ module Hyrax
 
         def generate_state_diagram!
           # look for old states that need to be removed and validate the states can be removed.
-          removed_states = validate_states_to_be_removed
+          unused_states_to_remove = validate_states_to_be_removed
 
           # remove any actions that will no longer be needed by the workflow
-          remove_unused_actions
+          remove_unused_actions!
 
           actions_configuration.each do |configuration|
             Array.wrap(configuration.fetch(:name)).each do |name|
@@ -54,24 +56,31 @@ module Hyrax
           # remove any states that are no longer needed by the workflow
           #  Note I am doing this after the update so nothing is still linking to the
           #       state to keep it from being deleted.
-          removed_states.each(&:delete)
+          unused_states_to_remove.each(&:destroy)
         end
 
         def validate_states_to_be_removed
-          new_states = actions_configuration.map { |a| a[:transition_to] }
-          removed_states = workflow.workflow_states.reject { |state| new_states.include?(state.name) }
-          removed_states.each do |state|
-            # a state  must have been removed, we should remove it only if there is not entities currently in that state
-            raise InvalidStateRemovalException.new("can not delete a state with entites", state) if state.entities.count > 0
+          new_state_names = actions_configuration.map { |a| a[:transition_to] }.flatten
+          states_to_remove = []
+          states_that_cannot_be_destroyed = []
+          workflow.workflow_states.each do |state|
+            next if new_state_names.include?(state.name)
+            states_to_remove << state
+            states_that_cannot_be_destroyed << state if state.entities.count > 0 # Choosing count so we pre-warm the query
           end
-          removed_states
+          if states_that_cannot_be_destroyed.any?
+            exception_message = "Cannot delete one or more states because they have one or more entities associated with them."
+            raise InvalidStateRemovalException.new(exception_message, states_that_cannot_be_destroyed)
+          end
+          states_to_remove
         end
 
-        def remove_unused_actions
+        def remove_unused_actions!
           # clear any actions from the workflow that do not exists in the new configuration
-          new_actions = actions_configuration.map { |a| a[:name] }
-          remove_actions = workflow.workflow_actions.reject { |action| new_actions.include?(action.name) }
-          remove_actions.each(&:delete)
+          new_action_names = actions_configuration.map { |a| a[:name] }.flatten
+          workflow.workflow_actions.each do |workflow_action|
+            workflow_action.destroy unless new_action_names.include?(workflow_action.name)
+          end
         end
     end
   end
--- a/app/services/hyrax/workflow/sipity_actions_generator.rb
+++ b/app/services/hyrax/workflow/sipity_actions_generator.rb
@@ -30,6 +30,7 @@ module Hyrax
 
       public
 
+      # @raises [Hyrax::Workflow::InvalidStateRemovalException] Trying to remove a state that is in use
       def call
         generate_state_diagram!
         workflow
@@ -38,11 +39,39 @@ module Hyrax
       private
 
         def generate_state_diagram!
+          # look for old states that need to be removed and validate the states can be removed.
+          removed_states = validate_states_to_be_removed
+
+          # remove any actions that will no longer be needed by the workflow
+          remove_unused_actions
+
           actions_configuration.each do |configuration|
             Array.wrap(configuration.fetch(:name)).each do |name|
               StateMachineGenerator.generate_from_schema(workflow: workflow, name: name, **configuration.except(:name))
             end
           end
+
+          # remove any states that are no longer needed by the workflow
+          #  Note I am doing this after the update so nothing is still linking to the
+          #       state to keep it from being deleted.
+          removed_states.each(&:delete)
+        end
+
+        def validate_states_to_be_removed
+          new_states = actions_configuration.map { |a| a[:transition_to] }
+          removed_states = workflow.workflow_states.reject { |state| new_states.include?(state.name) }
+          removed_states.each do |state|
+            # a state  must have been removed, we should remove it only if there is not entities currently in that state
+            raise InvalidStateRemovalException.new("can not delete a state with entites", state) if state.entities.count > 0
+          end
+          removed_states
+        end
+
+        def remove_unused_actions
+          # clear any actions from the workflow that do not exists in the new configuration
+          new_actions = actions_configuration.map { |a| a[:name] }
+          remove_actions = workflow.workflow_actions.reject { |action| new_actions.include?(action.name) }
+          remove_actions.each(&:delete)
         end
     end
   end
--- a/app/services/hyrax/workflow/workflow_importer.rb
+++ b/app/services/hyrax/workflow/workflow_importer.rb
@@ -38,12 +38,20 @@ module Hyrax
       def self.generate_from_json_file(path:, **keywords)
         contents = path.respond_to?(:read) ? path.read : File.read(path)
         data = JSON.parse(contents)
+        generate_from_hash(data: data, **keywords)
+      end
+
+      # @api public
+      #
+      # Responsible for generating the work type and corresponding processing entries based on given pathname or JSON document.
+      #
+      # @return [Array<Sipity::Workflow>]
+      def self.generate_from_hash(data:, **keywords)
         importer = new(data: data, **keywords)
-        workflow = importer.call
+        workflows = importer.call
         self.load_errors ||= []
         load_errors.concat(importer.errors)
-
-        workflow
+        workflows
       end
 
       # @param data [#deep_symbolize_keys] the configuration information from which we will generate all the data entries
--- a/app/services/hyrax/workflow/workflow_importer.rb
+++ b/app/services/hyrax/workflow/workflow_importer.rb
@@ -1,12 +1,20 @@
 module Hyrax
   module Workflow
+    # Responsible for loading workflows from a data source.
+    #
+    # @see .load_workflows
+    # @see .generate_from_json_file
     class WorkflowImporter
       class << self
-        attr_accessor :load_errors
-
-        def clear_load_errors
+        def clear_load_errors!
           self.load_errors = []
         end
+
+        attr_reader :load_errors
+
+        private
+
+          attr_writer :load_errors
       end
 
       # @api public
@@ -14,9 +22,8 @@ module Hyrax
       # Load all the workflows in config/workflows/*.json
       # @return [TrueClass]
       def self.load_workflows
-        clear_load_errors
+        clear_load_errors!
         Dir.glob(Rails.root.join('config', 'workflows', '*.json')) do |config|
-        Dir.glob(Rails.root + "config/workflows/*.json") do |config|
           Rails.logger.info "Loading workflow: #{config}"
           generate_from_json_file(path: config)
         end
@@ -83,9 +90,11 @@ module Hyrax
           begin
             find_or_create_from(configuration: configuration)
           rescue InvalidStateRemovalException => e
-            error = I18n.t('hyrax.workflow.load.state_error', workflow_name: e.state.workflow.name, state_name: e.state.name, entity_count: e.state.entities.count)
-            Rails.logger.error(error)
-            errors << error
+            e.states.each do |state|
+              error = I18n.t('hyrax.workflow.load.state_error', workflow_name: state.workflow.name, state_name: state.name, entity_count: state.entities.count)
+              Rails.logger.error(error)
+              errors << error
+            end
             Sipity::Workflow.find_by(name: configuration[:name])
           end
         end
@@ -95,7 +104,7 @@ module Hyrax
 
         def find_or_create_from(configuration:)
           workflow = Sipity::Workflow.find_or_initialize_by(name: configuration.fetch(:name))
-          generate_state_diagram(workflow: workflow, actions_configuration: configuration.fetch(:actions))
+          generate_state_diagram!(workflow: workflow, actions_configuration: configuration.fetch(:actions))
 
           find_or_create_workflow_permissions!(
             workflow: workflow, workflow_permissions_configuration: configuration.fetch(:workflow_permissions, [])
@@ -108,7 +117,7 @@ module Hyrax
 
         extend Forwardable
         def_delegator WorkflowPermissionsGenerator, :call, :find_or_create_workflow_permissions!
-        def_delegator SipityActionsGenerator, :call, :generate_state_diagram
+        def_delegator SipityActionsGenerator, :call, :generate_state_diagram!
 
         module SchemaValidator
           # @param data [Hash]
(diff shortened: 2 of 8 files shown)


app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -4,24 +4,28 @@ class WorkflowImporter
       # @api public
       #
       # Load all the workflows in config/workflows/*.json
-      # @return [TrueClass]
+      # @return [Array] List of errors during processing
```
jcoyne: Noting that this is a public API change (and will affect Hyku)

cam156: @jcoyne I could make it a getter and still return true always.

cam156:
@jcoyne I am not sure I am in love with the way the errors are reported now, but it is similar to other ruby constructs and keeps the public api the same.

app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -1,27 +1,41 @@
...
-        Dir.glob(Rails.root + "config/workflows/*.json") do |config|
-          Rails.logger.info "Loading workflow: #{config}"
-          generate_from_json_file(path: config)
+      class << self
+        attr_accessor :load_errors
+
+        # @api public
+        #
+        # Load all the workflows in config/workflows/*.json
+        # @return [Array] List of errors during processing
```
jeremyf: An intention of the `@api public` is to say that the method signature and return values do not change (without ample deprecation warning).  The new documentation says it returns an Array of load_errors (though the method still returns true).

I believe the return value should remain true.

app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -1,27 +1,41 @@
...
+        # @return Array<Sipity::Workflow> Workflow
+        def generate_from_json_file(path:, **keywords)
+          contents = path.respond_to?(:read) ? path.read : File.read(path)
+          data = JSON.parse(contents)
+          importer = new(data: data, **keywords)
+          workflow = importer.call
+          self.load_errors ||= []
+          load_errors.concat(importer.errors)
+
+          workflow
```
jeremyf: If this is an Array of workflows, set the variable name as plural (workflows).

app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -1,27 +1,41 @@
...
     class WorkflowImporter
-      # @api public
-      #
-      # Load all the workflows in config/workflows/*.json
-      # @return [TrueClass]
-      def self.load_workflows
-        Dir.glob(Rails.root + "config/workflows/*.json") do |config|
-          Rails.logger.info "Loading workflow: #{config}"
-          generate_from_json_file(path: config)
+      class << self
```
jeremyf: Moving all of the methods into a `class << self` and removing the `self.` part of method definition has made reviewing the changes more difficult. Please preserve the original methods and add the `:load_errors` accessor via the following:

```ruby
class << self
  attr_accessor :load_errors
end
```

One side-effect of the :load_errors is that they will remain for the duration of the script. If these are loaded via the Rails UI (e.g. by allowing JSON files to be uploaded and load the WorkflowImporter), the load_errors will not clear out.

app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -60,25 +74,34 @@ def validate!
...
+            errors << error
+            Sipity::Workflow.find_by(name: configuration[:name])
+          end
         end
       end
 
       private
 
         def find_or_create_from(configuration:)
-          workflow = Sipity::Workflow.find_or_initialize_by(name: configuration.fetch(:name)) do |wf|
```
jeremyf: The reorganization of this method does not look as though it changes anything. Why was it reorganized?

I'm concerned that we would be interacting with a non-persisted workflow. The previous version found the workflow, updated attributes, then saved (and threw an exception if invalid). Then it proceeded to generate the state_diagram and find the workflow permission.

The new version, finds or initialized, makes transformative changes on the state_diagram and permissions, then attempts to save the workflow (with validation). I think this could result in inconsistent data.

cam156: @jeremyf It does not change anything unless the workflow fails becuase it removes a state it is not allowed.  At that point if you save changes to the workflow and then error later the workflow was half updated.  All the tests still passed when I moved the code around, but my tests for failing did not.  Do you have an example of your failure scenario?

app/services/hyrax/workflow/workflow_importer.rb
cam156: @jeremyf I'll move it back.  My thought was that was the prefered syntax when you are creating many class methods.

app/services/hyrax/workflow/workflow_importer.rb
jeremyf: @cam156 I'll write up a spec (but for now consider this one tabled and okay to keep as you have).

mjgiarlo:
@cam156 :+1: to "better to have a bad Spanish translation than a missing one." Thank you!

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -30,6 +30,7 @@ def actions_configuration=(input)
 
       public
 
+      # @raises [Hyrax::Workflow::InvalidStateRemovalException] Trying to remove a state that is in use
```
mjgiarlo: I believe this should be `@raise` rather than `@raises`

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
       private
 
         def generate_state_diagram!
+          # look for old states that need to be removed and validate the states can be removed.
+          removed_states = validate_states_to_be_removed
+
+          # remove any actions that will no longer be needed by the workflow
+          remove_unused_actions
```
mjgiarlo: We aren't :100: consistent about this convention, but I wonder if you would consider making this a `!` method since we don't care about the return value.

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
       private
 
         def generate_state_diagram!
+          # look for old states that need to be removed and validate the states can be removed.
+          removed_states = validate_states_to_be_removed
```
mjgiarlo: Is there some value in creating this variable at this position in the method? Or could we avoid the variable and change line 57 to `validate_states_to_be_removed.each(&:delete)`?

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
...
           actions_configuration.each do |configuration|
             Array.wrap(configuration.fetch(:name)).each do |name|
               StateMachineGenerator.generate_from_schema(workflow: workflow, name: name, **configuration.except(:name))
             end
           end
+
+          # remove any states that are no longer needed by the workflow
+          #  Note I am doing this after the update so nothing is still linking to the
+          #       state to keep it from being deleted.
+          removed_states.each(&:delete)
```
mjgiarlo: Can you use `#destroy_all` here instead of `#each`?

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
...
+        end
+
+        def validate_states_to_be_removed
+          new_states = actions_configuration.map { |a| a[:transition_to] }
```
mjgiarlo: @jeremyf @jcoyne Is this a reliable way to get a full list of all the states? 

Also, is `:transition_to` always a scalar, or do we need some logic here to perhaps flatten the result?

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
...
+          removed_states = workflow.workflow_states.reject { |state| new_states.include?(state.name) }
```
mjgiarlo: If both `new_states` and `workflow.workflow_states` are arrays, you may be able to simplify this to `removed_states = workflow.workflow_states - new_states`

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
...
+          removed_states.each do |state|
```
mjgiarlo: Won't this only raise on the first removed state? Perhaps try something like this:

```ruby
removed_states_with_entities = removed_states.select { |state| state.entities.any? }
raise InvalidStateRemovalException.new("Workflow states cannot be removed when they contain entities: #{removed_states_with_entities.to_sentence}") if removed_states_with_entities.any?
```

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
...
+            # a state  must have been removed, we should remove it only if there is not entities currently in that state
+            raise InvalidStateRemovalException.new("can not delete a state with entites", state) if state.entities.count > 0
+          end
+          removed_states
+        end
+
+        def remove_unused_actions
+          # clear any actions from the workflow that do not exists in the new configuration
+          new_actions = actions_configuration.map { |a| a[:name] }
+          remove_actions = workflow.workflow_actions.reject { |action| new_actions.include?(action.name) }
```
mjgiarlo: Same comment as above re: array substraction.

app/services/hyrax/workflow/sipity_actions_generator.rb
```diff
@@ -38,11 +39,39 @@ def call
...
+          remove_actions.each(&:delete)
```
mjgiarlo: Same comment as above re: `#destroy_all`.

app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -1,11 +1,20 @@
 module Hyrax
   module Workflow
     class WorkflowImporter
+      class << self
+        attr_accessor :load_errors
+
+        def clear_load_errors
```
mjgiarlo: Would you consider making this a `!` method?

app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -60,25 +74,34 @@ def validate!
...
       def call
+        self.errors = []
         Array.wrap(data.fetch(:workflows)).map do |configuration|
-          find_or_create_from(configuration: configuration)
+          begin
+            find_or_create_from(configuration: configuration)
+          rescue InvalidStateRemovalException => e
+            error = I18n.t('hyrax.workflow.load.state_error', workflow_name: e.state.workflow.name, state_name: e.state.name, entity_count: e.state.entities.count)
+            Rails.logger.error(error)
+            errors << error
```
mjgiarlo: Should this be `self.errors`?

app/services/hyrax/workflow/workflow_importer.rb
```diff
@@ -60,25 +74,34 @@ def validate!
...
       private
 
         def find_or_create_from(configuration:)
-          workflow = Sipity::Workflow.find_or_initialize_by(name: configuration.fetch(:name)) do |wf|
-            wf.label = configuration.fetch(:label, nil)
-            wf.description = configuration.fetch(:description, nil)
-            wf.save!
-          end
+          workflow = Sipity::Workflow.find_or_initialize_by(name: configuration.fetch(:name))
+          generate_state_diagram(workflow: workflow, actions_configuration: configuration.fetch(:actions))
```
mjgiarlo: Isn't that a `!` method?

config/locales/hyrax.en.yml
```diff
@@ -588,6 +588,8 @@ en:
       default:
         deposit: "Deposit"
       unauthorized: "The work is not currently available because it has not yet completed the approval process"
+      load:
+        state_error: "The workflow: %{workflow_name} has not been updated.  You are removing a state: %{state_name} with %{entity_count} entitie(s).  A state may not be removed while it has active enities!"
```
mjgiarlo: Change `entitie(s)` to `entity/ies`. And change `enities` to `entities`

config/locales/hyrax.es.yml
```diff
@@ -607,6 +607,8 @@ es:
     workflow:
       default:
         deposit: "Depositar"
+      load:
+        state_error: "El flujo de trabajo por %{workflow_name} no se ha actualizado.  Estás eliminando un estado: %{state_name} con %{entity_count} entitie(s).  ¡Un estado no puede ser removido mientras tiene enidades activas!"
```
mjgiarlo: Change `entitie(s)` to `enidades`

spec/services/hyrax/workflow/workflow_importer_spec.rb
```diff
@@ -42,8 +46,67 @@
         result = described_class.generate_from_json_file(path: path)
       end.to change { Sipity::Workflow.count }.by(1)
       expect(result).to match_array(kind_of(Sipity::Workflow))
+      expect(described_class.load_errors).to be_empty
       expect(result.first.label).to eq "This is the label"
       expect(result.first.description).to eq "This description could get really long"
     end
   end
+  context "when I load twice" do
+    let!(:workflow1) { described_class.generate_from_json_file(path: path)[0] }
```
mjgiarlo: Would you consider using `#first` here instead of `[0]`? I find `first` harder to miss when I scan this with my eyes.

spec/services/hyrax/workflow/workflow_importer_spec.rb
```diff
@@ -42,8 +46,67 @@
...
+    let(:workflow2) { described_class.generate_from_json_file(path: path)[0] }
```
mjgiarlo: Would you consider using `#first` here instead of `[0]`? I find `first` harder to miss when I scan this with my eyes.

spec/services/hyrax/workflow/workflow_importer_spec.rb
```diff
@@ -42,8 +46,67 @@
...
+    it "creates the same results" do
+      expect(workflow2).to eq(workflow1)
+      expect(workflow2_errors).to be_empty
+    end
+    context "When the json changes" do
+      let(:workflow_name) { "awsome workflow" }
+      let(:action_name)  { "awesome action" }
+      let(:state_name)   { "awesome state" }
+      let(:second_path) { double(read: json2) }
+      let(:workflow2) { described_class.generate_from_json_file(path: second_path)[0] }
```
mjgiarlo: Would you consider using `#first` here instead of `[0]`? I find `first` harder to miss when I scan this with my eyes.

spec/services/hyrax/workflow/workflow_importer_spec.rb
```diff
@@ -42,8 +46,67 @@
...
+      context "when the workflow name stays the same" do
+        let(:workflow_name) { "ulra_submission" }
+        it "modifies the same workflow" do
+          expect(workflow2.label).not_to eq(workflow1.label)
+          expect(Sipity::Workflow.count).to eq(1)
+          expect(workflow2_errors).to be_empty
+          expect(Sipity::WorkflowAction.count).to eq(1)
+          expect(Sipity::WorkflowState.count).to eq(1)
+        end
+        context "when entites are in the state" do
```
mjgiarlo: `s/entites/entities/`

spec/services/hyrax/workflow/workflow_importer_spec.rb
```diff
@@ -42,8 +46,67 @@
...
+          let!(:entity) { Sipity::Entity.create(workflow_state: workflow1.reload.workflow_states[0], proxy_for_global_id: "abc123", workflow_id: workflow1.id) }
```
mjgiarlo: Would you consider using `#first` here instead of `[0]`? I find `first` harder to miss when I scan this with my eyes.

app/services/hyrax/workflow/sipity_actions_generator.rb
cam156: @mjgiarlo If we compare the new states to the old states after we update the workflow then we will never see the changes.  The old state will  be the new state. at line 57.  The other issue is that we can not delete the states until they are no longer referenced by actions.

app/services/hyrax/workflow/sipity_actions_generator.rb
cam156: Not sure I can give it a try

app/services/hyrax/workflow/sipity_actions_generator.rb
cam156: @mjgiarlo Yeah that might be better, but since we have no validation some is better than none.  At this point I'm going to unassign myself from this ticket as there seems to be a lot of work left.  I'm off the sprint this week and I will let the next person pick back up with this work.

cam156:
@mjgiarlo @jeremyf @jcoyne At this point I am no longer working on the mediated deposit sprint. I was hoping that this would get through, but since you all seems to want many changes I'm going to leave it for the next person.

jeremyf:
@cam156 Thanks for the heads up, I'll pull down the branch and work through suggestions.

app/services/hyrax/workflow/sipity_actions_generator.rb
jeremyf: @mjgiarlo Given that the methods being called are private, the naming is a bit less vital. However, I will make the change.

app/services/hyrax/workflow/sipity_actions_generator.rb
jeremyf: @mjgiarlo As this is not an ActiveRelation object, I believe we need to iterate and `destroy`

app/services/hyrax/workflow/sipity_actions_generator.rb
jeremyf: Yes, the configuration is a Hash (we don't have a more explicit data structure). The `transition_to` is always a String. 

See https://github.com/projecthydra-labs/hyrax/blob/5907e1d4f265139b554909dc442dbc91a585ae8c/app/services/hyrax/workflow/workflow_schema.rb#L43

app/services/hyrax/workflow/sipity_actions_generator.rb
jeremyf: `new_states` is an array of strings. `workflow.workflow_states` is an array of WorkflowState objects. I believe we could reduce the number of iterations through the states, but the current code is reasonable.

jeremyf:
This will require https://github.com/projecthydra-labs/hyrax/pull/270 to be merged

jeremyf:
This depends on https://github.com/projecthydra-labs/hyrax/pull/270

jcoyne:
@jeremyf rebase?

mjgiarlo:
@cam156 @jeremyf Thanks for your work on this!

PR #275, opened 2017-01-17
Adding support for i18n-debug

jeremyf:
See https://github.com/fphilipe/i18n-debug. We may need to observe if
this adds any significant time to builds.

Below is example output that is written to
`.internal_test_app/log/test.log`:

```console
[i18n-debug] en.simple_form.labels.file_set.files => nil
[i18n-debug] en.simple_form.labels.file_set.files => nil
[i18n-debug] en.simple_form.labels.defaults.files => "Upload a file"
[i18n-debug] en.simple_form.required.text => "required"
[i18n-debug] en.simple_form.required.mark => "*"
[i18n-debug] en.simple_form.required.html => "<span class=\"label label-info required-tag\">required</span>"
[i18n-debug] en.simple_form.hints.file_set.files => nil
[i18n-debug] en.simple_form.hints.file_set.files => nil
[i18n-debug] en.simple_form.hints.defaults.files => nil
[i18n-debug] en.simple_form.hints.defaults.files => nil
[i18n-debug] en.simple_form.hints.file_set.files => nil
[i18n-debug] en.simple_form.hints.file_set.files => nil
[i18n-debug] en.simple_form.hints.defaults.files => nil
[i18n-debug] en.simple_form.hints.defaults.files => nil
[i18n-debug] en.activemodel.models.file_set => nil
[i18n-debug] en.activemodel.models.active_fedora/base => nil
[i18n-debug] en.helpers.submit.file_set.update => nil
[i18n-debug] en.helpers.submit.update => "Save"
```

While the specs are running, I was running the following command to see
what translation keys was hit or missed:

```console
tail -f .internal_test_app/log/test.log
```

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -79,5 +79,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'shoulda-matchers', '~> 3.1'
   spec.add_development_dependency 'rails-controller-testing', '~> 0'
   spec.add_development_dependency 'webmock'
-  spec.add_development_dependency 'i18n-debug'
+  spec.add_development_dependency 'i18n-debug' unless ENV['TRAVIS']
 end
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -79,4 +79,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'shoulda-matchers', '~> 3.1'
   spec.add_development_dependency 'rails-controller-testing', '~> 0'
   spec.add_development_dependency 'webmock'
+  spec.add_development_dependency 'i18n-debug'
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -51,6 +51,7 @@ require 'support/speedup'
 require 'webmock/rspec'
 WebMock.disable_net_connect!(allow_localhost: true)
 
+require 'i18n/debug' unless ENV['TRAVIS']
 require 'byebug' unless ENV['TRAVIS']
 
 Capybara.default_driver = :rack_test      # This is a faster driver
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -51,7 +51,7 @@ require 'support/speedup'
 require 'webmock/rspec'
 WebMock.disable_net_connect!(allow_localhost: true)
 
-require 'i18n/debug' unless ENV['TRAVIS']
+require 'i18n/debug' if ENV['I18N_DEBUG']
 require 'byebug' unless ENV['TRAVIS']
 
 Capybara.default_driver = :rack_test      # This is a faster driver
@@ -80,7 +80,7 @@ ActiveJob::Base.queue_adapter = :inline
 # HttpLogger.ignore = [/localhost:8983\/solr/]
 # HttpLogger.colorize = false
 
-if ENV['TRAVIS'] == 'true'
+if ENV['TRAVIS']
   # Monkey-patches the FITS runner to return the PDF FITS fixture
   module Hydra::Works
     class CharacterizationService


jeremyf:
Related to https://github.com/projecthydra-labs/hyku/pull/629

hackmastera:
@jeremyf how frequent is that output? is that the full output right now? It looks like it may have the potential to make the (already noisy) logs very noisy. Did you consider a report (rake task?). Or are you hoping the noise will cause us to address issues with i18n?

jeremyf:
@hackmastera At present, anytime there is a request to I18n it would be logged.

It would be possible to require an ENV be set for the logging to occur (e.g. `require 'i18n-debug' if ENV['i18n_debug']`)

hackmastera:
+1 on the ENV idea; I would even be okay with the ENV turning it off if more people want to see this in the logs on a regular basis (i.e. expect to incorporate whatever will be required to deal with the messages into their workflow)

mjgiarlo:
@hackmastera @jeremyf Made the change y'all discussed. I know @jcoyne also discussed this with @jeremyf but I couldn't find that thread, so I'd be glad to hear your thoughts on this too.

Note that I also created this wiki page as a companion to this PR: https://github.com/projecthydra-labs/hyrax/wiki/Testing-internationalization-(i18n)-support

mjgiarlo:
Rebased.

hyrax.gemspec
```diff
@@ -80,4 +80,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'shoulda-matchers', '~> 3.1'
   spec.add_development_dependency 'rails-controller-testing', '~> 0'
   spec.add_development_dependency 'webmock'
+  spec.add_development_dependency 'i18n-debug'
```
jcoyne: How about we don't even install it on travis? https://docs.travis-ci.com/user/languages/ruby/#Speeding-up-your-build-by-excluding-non-essential-dependencies

mjgiarlo: @jcoyne Would adding `unless ENV['TRAVIS'] == 'true'` at the end of this line suffice?

jcoyne: I think that would work.

mjgiarlo: Done

mjgiarlo:
@jcoyne Re-review?

hyrax.gemspec
```diff
@@ -79,4 +79,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'shoulda-matchers', '~> 3.1'
   spec.add_development_dependency 'rails-controller-testing', '~> 0'
   spec.add_development_dependency 'webmock'
+  spec.add_development_dependency 'i18n-debug' unless ENV['TRAVIS'] == 'true'
```
jcoyne: why are we checking for a specific value here, and just the existence of the variable in `spec_helper.rb`?

mjgiarlo: Because `spec_helper.rb` was itself internally inconsistent. 😆  Fixed that in my follow-up. Thx.

mjgiarlo:
@jcoyne OK, think I've got it now.

spec/spec_helper.rb
```diff
@@ -51,6 +51,7 @@ def coverage_needed?
 require 'webmock/rspec'
 WebMock.disable_net_connect!(allow_localhost: true)
 
+require 'i18n/debug' if ENV['I18N_DEBUG']
```
mjgiarlo: Does this introduce a weirdness if we add `ENV['I18N_DEBUG']` in Travis? I.e., should the `hyrax.gemspec` change above also check the value of `ENV['I18N_DEBUG']`, or rectify this when needed?

jcoyne: NBD ATM

PR #283, opened 2017-01-18
Test on the most recent Ruby release, 2.4.0

jcoyne:


jcoyne:
@jeremyf mostly I'm interested in having our tests run fast.  I don't want to consume all our available workers on different versions. Feel free to add a second build if you want to see it on other versions.

jeremyf:
I like the reasoning…just wanting to understand them.

PR #284, opened 2017-01-19
fix for #221

### LINKED ISSUE #221: Lexvo languages rake task fails to add anything to the db because of a difference in namespace
### Descriptive summary

Observed in 0.0.1.alpha. When running rake hyrax:controlled_vocabularies:language no languages are added to the database. The cause is a difference in the namespace used in the lexvo data (http://www.w3.org/2008/05/skos#) and RDF::Vocab::SKOS (http://www.w3.org/2004/02/skos/core). The easiest fix would be to change the predicate in /lib/hyrax/controlled_vocabulary/importer/language.rb to match the one in lexvo (http://www.w3.org/2008/05/skos#prefLabel). I'd be happy to submit a PR with this change made.

### Expected behavior

Almost 8000 langages should be added to the qa_local_authority_entries table

### Actual behavior

No language terms are added to the database table.

1. Run rake hyrax:controlled_vocabularies:language
2. Once it completes check http://localhost:3000/authorities/terms/local/languages - if it's empty, you have the problem

geekscruff:
Fixes #221 

Fixes incorrect namespace in lexvo languages importer

rake hyrax:controlled_vocabularies:languages fails to import any languages because the namespace in RDF::Vocab::SKOS is different to the one in the Lexvo data dump. Switching the importer (lib/hyrax/controlled_vocabulary/importer/language.rb) to use the Lexvo skos namespace fixes this.

### DIFF
--- a/lib/hyrax/controlled_vocabulary/importer/language.rb
+++ b/lib/hyrax/controlled_vocabulary/importer/language.rb
@@ -23,7 +23,7 @@ module Hyrax
             'languages',
             [rdf_path],
             format: 'rdfxml',
-            predicate: RDF::Vocab::SKOS.prefLabel
+            predicate: RDF::URI('http://www.w3.org/2008/05/skos#prefLabel')
           )
           logger.info "Import complete"
         end
--- a/spec/lib/hyrax/controlled_vocabulary/importer/language_spec.rb
+++ b/spec/lib/hyrax/controlled_vocabulary/importer/language_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe Hyrax::ControlledVocabulary::Importer::Language do
       'languages',
       [rdf_path],
       format: 'rdfxml',
-      predicate: RDF::Vocab::SKOS.prefLabel
+      predicate: RDF::URI('http://www.w3.org/2008/05/skos#prefLabel')
     )
     instance.import
   end


jcoyne:
This is because Lexvo is not using the current standard predicate. Compare:
https://www.w3.org/2009/08/skos-reference/skos.html#prefLabel which specifies http://www.w3.org/2004/02/skos/core#prefLabel
https://www.w3.org/TR/2008/WD-skos-reference-20080829/skos.html#prefLabel which specifies http://www.w3.org/2008/05/skos#prefLabel

lib/hyrax/controlled_vocabulary/importer/language.rb
```diff
@@ -23,7 +23,7 @@ def import
             'languages',
             [rdf_path],
             format: 'rdfxml',
-            predicate: RDF::Vocab::SKOS.prefLabel
+            predicate: 'http://www.w3.org/2008/05/skos#prefLabel'
```
jcoyne: Should this be an `RDF::URI`?

geekscruff: It works either way so if you'd rather I use RDF::URI('http://www.w3.org/2008/05/skos#prefLabel') I'll happily change it.

jcoyne:
@geekscruff would you mind changing the commit message to say what you fixed and reference the issue number. This makes the message understandable even if you don't have access to github issues.

lib/hyrax/controlled_vocabulary/importer/language.rb
jcoyne: I think using the object is better.

geekscruff:
@jcoyne I've updated the commit message but my attempt to squash the commits went awry. If you want to reject the PR I can start again and try and get this right.

jcoyne:
@geekscruff you can do it. What you're going to do is the following:
`git rebase -i HEAD~6`  <- start an interactive rebase with the last 6 commits
It will show you as screen in your editor like this:
```
pick c3ad157 fix for #221
pick 54b72 update spec test for languages rake task
pick 32bc fix for #221 - hardcoding...
pick aaaaa message here
pick bbbb message here
pick ccccc message here

# Rebase 710f0f8..a5f4a0d onto 710f0f8
#
# Commands:
#  p, pick = use commit
#  r, reword = use commit, but edit the commit message
#  e, edit = use commit, but stop for amending
#  s, squash = use commit, but meld into previous commit
#  f, fixup = like "squash", but discard this commit's log message
#  x, exec = run command (the rest of the line) using shell
#
```

You will then change the word pick to "squash" for all of the commits after the first one (leave that as pick). Then save.  Then it will open a new editor and you can edit the commit message. Finally, you'll force push to this branch.
`git push origin +fix_221`

geekscruff:
Thanks!

jeremyf:
@geekscruff You can also do `git rebase master -i` and it will rebase all commits that are not on master.

PR #288, opened 2017-01-19
Tidying up rubocop violations and extracting a service class

jeremyf:
## Removing MethodLength exclusion

@4849cb548dd61664c2ae7547192548808ec408cb

In a perfect world, a Hyrax::FileSet::CreateDerivativesService would be
created, then we would delegate the public facing methods (e.g.
:cleanup_derivatives and :create_derivatives) to the underlying class.

However, this is not a perfect world.

## Extracting a class for FileSetDerivativesService

@7eb62f0e7d36ddc438d1328ce3a0274f9c864cfe

The purpose of this refactor is to build towards a more perfect world
by:

1) Reducing the number of methods added to a model, thus narrowing its
   interface (both public and private)
2) Exposing a clear collaborating class with a narrow purpose
3) To show a methodology for not abusing ActiveSupport::Concern
4) To be explicit in how a module is dependent on the class it mixes
   into, as well as what methods are important for that class to get
   from the module - see the delegate methods for how the service and
   FileSet are to interact.

Related to projecthydra/hydra-works#315

### DIFF
--- a/app/models/concerns/hyrax/file_set/derivatives.rb
+++ b/app/models/concerns/hyrax/file_set/derivatives.rb
@@ -15,34 +15,49 @@ module Hyrax
       # before derivatives so that we have a credible mime_type field to work with.
       def create_derivatives(filename)
         case mime_type
-        when *self.class.pdf_mime_types
+        when *self.class.pdf_mime_types             then create_pdf_derivatives(filename)
+        when *self.class.office_document_mime_types then create_office_document_derivatives(filename)
+        when *self.class.audio_mime_types           then create_audio_derivatives(filename)
+        when *self.class.video_mime_types           then create_video_derivatives(filename)
+        when *self.class.image_mime_types           then create_image_derivatives(filename)
+        end
+      end
+
+      private
+
+        def create_pdf_derivatives(filename)
           Hydra::Derivatives::PdfDerivatives.create(filename,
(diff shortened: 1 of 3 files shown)


app/models/concerns/hyrax/file_set/derivatives.rb
```diff
@@ -4,61 +4,22 @@ module Derivatives
       extend ActiveSupport::Concern
 
       included do
+        # TODO: Should these be moved to an initializer? If we are setting a class variable at include time,
+        # it seams that we could set them in the configuration.
```
jcoyne: Typo: 'seems'. And yes, they could go to an initializer

app/models/concerns/hyrax/file_set/derivatives.rb
```diff
@@ -4,61 +4,22 @@ module Derivatives
...
-
-        def cleanup_derivatives
-          derivative_path_factory.derivatives_for_reference(self).each do |path|
-            FileUtils.rm_f(path)
-          end
-        end
-
-        def derivative_path_factory
-          Hyrax::DerivativePath
+        def file_set_derivatives_service
```
jcoyne: 👍

app/models/concerns/hyrax/file_set/derivatives.rb
jeremyf: @jcoyne Moved to an initializer.

PR #293, opened 2017-01-19
Don't show guest users on the workflow roles page. Fixes #252

### LINKED ISSUE #252: Filter out guest users from the "workflow roles" page
/admin/workflow_roles

jcoyne:


app/presenters/hyrax/admin/workflow_role_presenter.rb
```diff
@@ -2,7 +2,7 @@ module Hyrax
   module Admin
     class WorkflowRolePresenter
       def users
-        ::User.all
+        ::User.where(guest: false)
```
mjgiarlo: Now that we've got this in a couple places, it might be good to extract this into the User model as a new class method called `.guests`, but I don't think you need to do that in this PR.

jcoyne: The method would be `.registered_users`, but sure. 👍

jcoyne: Added in https://github.com/projecthydra-labs/hyrax/pull/294

mjgiarlo: @jcoyne You think `.guests` shouldn't return non-guests or something?

PR #294, opened 2017-01-19
Add a User.registered scope

jcoyne:


mjgiarlo:
@jcoyne You can also use this in the WorkflowRolePresenter now that I've merged #293

jcoyne:
@mjgiarlo I've changed the WorkflowRolePresenter to use the new user scope

Issue #295, opened 2017-01-20
Running the test suite is not idempotent


danhorst:
Running the test suite using `bundle exec rspec` or `bundle exec rake spec` will only complete successfully the _first_ time it is run. Running the same command a second time will fail with:
```ruby
NameError: uninitialized constant Catapult
```

This is because while the test suite is running `config/initializers/hyrax.rb` gets modified to include:
```ruby
# Injected via `rails g hyrax:work Catapult`
config.register_curation_concern :catapult
```
This change is not reverted afterwards.

This problem isn’t present using `bundle exec rake ci` or in the Travis builds because the test app is rebuilt every time the tests are run.

---

I would like to be able to run one spec at a time using `bundle exec rspec TEST_FILE` while I am working on a particular feature. My current workaround is to keep the contents of `.internal_test_app` tracked in git so it is easy to revert changes to `config/initializers/hyrax.rb` when needed. I only use the `ci` task to check for regression failures because it takes about 12 minutes to complete on my development machine.

Places to look: `spec/features/work_generator_spec.rb` specifically line 18:
```ruby
Rails::Generators.invoke('hyrax:work', ['Catapult'], behavior: :revoke, destination_root: Rails.root)
```

atz:
If we separate out a new ticket "hyrax:work generator is not reversible", we can solve the test case with a basic hack-around.  Just copy the `hyrax.rb` initializer before the test and copy back after.

jcoyne:
Fixed by #302

PR #302, opened 2017-01-23
Make work generator clean up after itself. Fixes #301, #295

### LINKED ISSUE #301: hyrax:work generator is not reversible
Specifically the `register_work` method as described in #295.  

We ask the method to do a mildly complicated task (for a generator), since it has to preserve the order of inserted lines.  Note that logical variation in the content to be inserted (the additional comment line) is problematic (it would never get deleted, since `destroy` would only be run when lines are already added).  Probably can be removed.

mjgiarlo:


lib/generators/hyrax/work/work_generator.rb
```diff
@@ -1,70 +1,41 @@
 require 'rails/generators'
 
-class Rails::Generators::NamedBase
```
mjgiarlo: As far as I can tell, this is cruft from Curate that's not used.

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -1,70 +1,41 @@
...
-
-  def create_actor_spec
-    return unless rspec_installed?
-    template('actor_spec.rb.erb', "spec/actors/hyrax/actors/#{file_name}_actor_spec.rb")
-  end
-
-  def create_form_spec
-    return unless rspec_installed?
-    template('form_spec.rb.erb', "spec/forms/hyrax/#{file_name}_form_spec.rb")
+  def banner
```
mjgiarlo: Added this method so that the first step of the generator is displaying a status message.

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -1,70 +1,41 @@
...
-end
-
 class Hyrax::WorkGenerator < Rails::Generators::NamedBase
   source_root File.expand_path('../templates', __FILE__)
 
   argument :attributes, type: :array, default: [], banner: 'field:type field:type'
 
   # Why all of these antics with defining individual methods?
   # Because I want the output of Hyrax::WorkGenerator to include all the processed files.
-  def create_model_spec
```
mjgiarlo: I reordered the generator methods to generate its files in alphabetical order (app, then config, then spec).

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -73,7 +44,7 @@ def register_work
     lastmatch = nil
     in_root do
       File.open(config).each_line do |line|
-        lastmatch = line if line =~ /^\s*config.register_curation_concern :.*\n$/
+        lastmatch = line if line =~ /config.register_curation_concern :(?!#{file_name})/
```
mjgiarlo: Using negative lookahead in the anchor allows for this method to work for revoke/destroy as well as invoke/generate.

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -85,20 +56,47 @@ def register_work
...
+  end
+
+  def create_model_spec
+    return unless rspec_installed?
+    template('model_spec.rb.erb', "spec/models/#{file_name}_spec.rb")
   end
 
-  def create_readme
-    readme 'README'
+  def display_readme
```
mjgiarlo: Made this method's intention clearer. A README is not created here.

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -85,20 +56,47 @@ def register_work
...
+    readme 'README' unless revoking?
```
mjgiarlo: Don't display the README if destroying the work type.

lib/generators/hyrax/work/work_generator.rb
atz: I think we need to do away with prepending the `Note: order of registration...` comment unless lastmatch.  During revocation, the comment may or may not be there, depending on what versions of the generator have been invoked previously.

mjgiarlo: Done.

PR #305, opened 2017-01-24
Switching permissions templates to use workflow_id instead of workflo…

### LINKED ISSUE #191: Change PermissionTemplate workflow_name column to workflow_id
<a href="https://github.com/jcoyne"></img></a> **Issue by [jcoyne](https://github.com/jcoyne)**
_Thursday Dec 01, 2016 at 17:19 GMT_
_Originally opened as https://github.com/projecthydra/sufia/issues/2954_

----

Becuase the workflows are also in the database. https://github.com/projecthydra/sufia/blob/master/db/migrate/20161021175854_create_permission_template.rb#L6

scholarworks:
Updates to use workflow_id instead of workflow_name.

Fixes #191; refs #issuenumber

*  Switched workflow_id in place of workflow_name throughout
*  Added has_many relationship to Sipity::Workflow
*  Added belongs_to relationship to PremissionTemplate for Workflow

### DIFF
--- a/app/actors/hyrax/default_admin_set_actor.rb
+++ b/app/actors/hyrax/default_admin_set_actor.rb
@@ -36,7 +36,7 @@ module Hyrax
       # rubocop:disable Lint/HandleExceptions
       def create_default_admin_set
         AdminSet.create!(id: DEFAULT_ID, title: ['Default Admin Set']).tap do |_as|
-          PermissionTemplate.create!(admin_set_id: DEFAULT_ID, workflow_name: 'default')
+          PermissionTemplate.create!(admin_set_id: DEFAULT_ID)
         end
       rescue ActiveFedora::IllegalOperation
         # It is possible that another thread created the AdminSet just before this method
--- a/app/models/sipity/workflow.rb
+++ b/app/models/sipity/workflow.rb
@@ -10,10 +10,16 @@ module Sipity
     has_many :workflow_states, dependent: :destroy, class_name: 'Sipity::WorkflowState'
     has_many :workflow_actions, dependent: :destroy, class_name: 'Sipity::WorkflowAction'
     has_many :workflow_roles, dependent: :destroy, class_name: 'Sipity::WorkflowRole'
+    has_many :permission_templates, dependent: :restrict_with_exception, class_name: 'Hyrax::PermissionTemplate'
 
     DEFAULT_INITIAL_WORKFLOW_STATE = 'new'.freeze
     def initial_workflow_state
       workflow_states.find_or_create_by!(name: DEFAULT_INITIAL_WORKFLOW_STATE)
+    end
+
+    def self.default_workflow
+      log_and_raise('No workflows found') if Sipity::Workflow.none?
+      Sipity::Workflow.order('id ASC').first!
     end
   end
 end
--- a/app/services/hyrax/workflow/default_workflow_strategy.rb
+++ b/app/services/hyrax/workflow/default_workflow_strategy.rb
@@ -3,9 +3,9 @@ module Hyrax
     class DefaultWorkflowStrategy
       def initialize(_work, _attributes); end
 
-      # @return [String] The name of the workflow to use
-      def workflow_name
-        'default'
+      # @return [String] The id of the workflow to use
+      def workflow_id
+        Sipity::Workflow.default_workflow.id
       end
     end
   end
--- a/app/services/hyrax/workflow/workflow_by_admin_set_strategy.rb
+++ b/app/services/hyrax/workflow/workflow_by_admin_set_strategy.rb
@@ -6,9 +6,9 @@ module Hyrax
       end
 
       # @return [String] The name of the workflow by admin_set to use
-      def workflow_name
-        return 'default' unless @admin_set_id
-        Hyrax::PermissionTemplate.find_by!(admin_set_id: @admin_set_id).workflow_name
+      def workflow_id
+        return unless @admin_set_id
+        Hyrax::PermissionTemplate.find_by!(admin_set_id: @admin_set_id).workflow_id
       end
     end
   end
--- a/spec/services/hyrax/workflow/workflow_by_admin_set_strategy_spec.rb
+++ b/spec/services/hyrax/workflow/workflow_by_admin_set_strategy_spec.rb
@@ -2,21 +2,21 @@ describe Hyrax::Workflow::WorkflowByAdminSetStrategy, :no_clean do
   context "when using default workflow strategy" do
     let(:workflow_strategy) { described_class.new(nil, {}) }
 
-    describe '#workflow_name' do
-      subject { workflow_strategy.workflow_name }
-      it { is_expected.to eq 'default' }
+    describe '#workflow_id' do
+      subject { workflow_strategy.workflow_id }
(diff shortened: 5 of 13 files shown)


app/actors/hyrax/default_admin_set_actor.rb
```diff
@@ -35,7 +35,7 @@ def default_exists?
       # Creates the default AdminSet and an associated PermissionTemplate with workflow
       def create_default_admin_set
         AdminSet.create!(id: DEFAULT_ID, title: ['Default Admin Set']).tap do |_as|
-          PermissionTemplate.create!(admin_set_id: DEFAULT_ID, workflow_name: 'default')
+          PermissionTemplate.create!(admin_set_id: DEFAULT_ID, workflow_id: 1)
```
jeremyf: Not liking this; that is a more magical number than the magic string 'default'

How about:

```ruby
  PermissionTemplate.create!(admin_set_id: DEFAULT_ID, workflow_id: default_workflow.id)

# @raises ActiveRecord::RecordNotFound when we don't have a workflow
def default_workflow
  Sipity::Workflow.order('id ASC').first!
end
```

This will throw an exception if it does

app/models/sipity/workflow.rb
```diff
@@ -10,6 +10,7 @@ class Workflow < ActiveRecord::Base
     has_many :workflow_states, dependent: :destroy, class_name: 'Sipity::WorkflowState'
     has_many :workflow_actions, dependent: :destroy, class_name: 'Sipity::WorkflowAction'
     has_many :workflow_roles, dependent: :destroy, class_name: 'Sipity::WorkflowRole'
+    has_many :permission_templates, dependent: :restrict_with_exception
```
jeremyf: I believe this will need a `class_name: 'Hyrax::PermissionTemplate'` option

app/services/hyrax/workflow/default_workflow_strategy.rb
```diff
@@ -3,9 +3,9 @@ module Workflow
     class DefaultWorkflowStrategy
       def initialize(_work, _attributes); end
 
-      # @return [String] The name of the workflow to use
-      def workflow_name
-        'default'
+      # @return [String] The id of the workflow to use
+      def workflow_id
+        1
```
jeremyf: Again, I am very hesitant about this magic number.

app/services/hyrax/workflow/workflow_by_admin_set_strategy.rb
```diff
@@ -6,9 +6,9 @@ def initialize(_work, attributes)
       end
 
       # @return [String] The name of the workflow by admin_set to use
-      def workflow_name
-        return 'default' unless @admin_set_id
-        Hyrax::PermissionTemplate.find_by!(admin_set_id: @admin_set_id).workflow_name
+      def workflow_id
+        return 1 unless @admin_set_id
```
jeremyf: Again, I'm concerned about the magic number.

spec/services/hyrax/workflow/workflow_by_admin_set_strategy_spec.rb
```diff
@@ -2,21 +2,21 @@
   context "when using default workflow strategy" do
     let(:workflow_strategy) { described_class.new(nil, {}) }
 
-    describe '#workflow_name' do
+    describe '#workflow_id' do
       subject { workflow_strategy.workflow_name }
-      it { is_expected.to eq 'default' }
+      it { is_expected.to eq 1 }
```
jeremyf: Magic number :(

spec/views/hyrax/admin/admin_sets/_form_workflow.html.erb_spec.rb
```diff
@@ -4,11 +4,11 @@
   let(:admin_set) { create(:admin_set) }
   let(:permission_template) { Hyrax::PermissionTemplate.find_or_create_by(admin_set_id: admin_set.id) }
   before do
-    create(:workflow, name: "my_name", description: "random workflow", label: "my label")
+    create(:workflow, id: 1, description: "random workflow", label: "my label")
```
jeremyf: I don't know if you can create with a specified :id.

app/models/sipity/workflow.rb
```diff
@@ -16,5 +16,15 @@ class Workflow < ActiveRecord::Base
     def initial_workflow_state
       workflow_states.find_or_create_by!(name: DEFAULT_INITIAL_WORKFLOW_STATE)
     end
+
+    def self.default_workflow
+      Sipity::Workflow.order('id ASC').first! unless Sipity::Workflow.all.empty?
+      nil
```
jeremyf: This method will always return `nil`. I assume that is not your intention.

app/models/sipity/workflow.rb
```diff
@@ -16,5 +16,15 @@ class Workflow < ActiveRecord::Base
...
+    end
+
+    def self.default_workflow_id
+      default_workflow.id unless default_workflow.nil?
```
jeremyf: This method will always return `nil`.

app/actors/hyrax/default_admin_set_actor.rb
```diff
@@ -35,8 +35,15 @@ def default_exists?
       # Creates the default AdminSet and an associated PermissionTemplate with workflow
       def create_default_admin_set
         AdminSet.create!(id: DEFAULT_ID, title: ['Default Admin Set']).tap do |_as|
-          PermissionTemplate.create!(admin_set_id: DEFAULT_ID, workflow_name: 'default')
+          # PermissionTemplate.create!(admin_set_id: DEFAULT_ID, workflow_id: Sipity::Workflow.default_workflow.id)
```
jeremyf: Remove commented code

app/models/sipity/workflow.rb
```diff
@@ -10,10 +10,16 @@ class Workflow < ActiveRecord::Base
...
     has_many :workflow_roles, dependent: :destroy, class_name: 'Sipity::WorkflowRole'
+    has_many :permission_templates, dependent: :restrict_with_exception, class_name: 'Hyrax::PermissionTemplate'
 
     DEFAULT_INITIAL_WORKFLOW_STATE = 'new'.freeze
     def initial_workflow_state
       workflow_states.find_or_create_by!(name: DEFAULT_INITIAL_WORKFLOW_STATE)
     end
+
+    def self.default_workflow
+      return if Sipity::Workflow.none?
```
jeremyf: I'm not excited about returning `nil` if there aren't any workflows. It is likely to create upstream errors.

@jcoyne Is it a reasonable assumption that there exists a default workflow within the application? If so, I'd like this to remove this line and allow the exception from `first!` to bubble up.

jcoyne: I think raising an exception is the best option.

jeremyf: @scholarworks I believe you can get rid of the `Sipity::Workflow.none?` and let the next line's `Sipity::Workflow.order('id ASC').first!` raise the

jeremyf: @scholarworks Go ahead and remove the above `return if Sipity::Workflow.none?`

jeremyf:
@scholarworks Thanks for stewarding this PR to completion!

PR #314, opened 2017-01-26
Remove the route to versions, which is no longer used.

jcoyne:


cam156:
@jcoyne & @mjgiarlo as of Sufia 7.2 the version route was still coming from Sufia.  Removing this would be non backward compatible.  


```
Routes for Sufia::Engine:
                             homepage_index GET      /homepage(.:format)                                  homepage#index
...
                                       help GET      /help(.:format)                                      static#help
                                      terms GET      /terms(.:format)                                     static#terms
                                     zotero GET      /zotero(.:format)                                    static#zotero
                                   mendeley GET      /mendeley(.:format)                                  static#mendeley
                                  agreement GET      /agreement(.:format)                                 static#agreement
                                   versions GET      /versions(.:format)                                  static#versions
```

jcoyne:
@cam156 I removed it because it causes an action not found error, sufia provides the route, but it doesn't implement any action for it.

cam156:
@jcoyne interestingly it works without the controller action if the view is present.  There is no action here: https://github.com/psu-stewardship/scholarsphere/blob/develop/app/controllers/static_controller.rb but the view exists here in SS, https://github.com/psu-stewardship/scholarsphere/blob/develop/app/views/static/versions.html.erb

I'm guessing it is Rails magic...

mjgiarlo:
@jcoyne @cam156 Could it be this was never fully extracted to Sufia? It could be this has only ever worked in ScholarSphere. We have at least two options:

1. Remove from Sufia altogether (this PR) and let it all live in ScholarSphere. (Possibly a breaking change, at least for Penn State.)
2. Pull ScholarSphere versions stuff back into Sufia. (Though I can't find evidence that anyone beside Penn State is using this.)

jcoyne:
Yes, it works if you have the template present. Since hyrax doesn't have the versions.html.erb template and it doesn't have a link to `versions_path`, maybe it doesn't belong in hyrax.

cam156:
@mjgiarlo @jcoyne I'm fine with either option for  hyrax, since it is an entirely new product with no semantic version history.

jcoyne:
We can wait for this change until after 0.1.0 is cut.

mjgiarlo:
@jcoyne @cam156 OK. Putting this in the 1.0.0 milestone.

jeremyf:
I'm away from my login. Go ahead and dismiss my change request.

Jeremy Friesen

On Mar 23, 2017 4:22 PM, "Justin Coyne" <notifications@github.com> wrote:

@jeremyf <https://github.com/jeremyf> this is unblocked. Care to re-review?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/projecthydra-labs/hyrax/pull/314#issuecomment-288848459>,
or mute the thread

PR #315, opened 2017-01-26
Flash messages must be marked as html_safe, else HTML is unescaped

mjgiarlo:
Fixes projecthydra-labs/hyku#694

Bug introduced in https://github.com/projecthydra-labs/hyrax/commit/be4a648f9e0f9191f514a241780dabab29c73777 because `<br/>` was not marked as html_safe.

### DIFF
--- a/app/views/_flash_msg.html.erb
+++ b/app/views/_flash_msg.html.erb
@@ -2,7 +2,7 @@
   <% if flash[type].present? %>
     <div class="alert <%= flash_dom_class %> alert-dismissable" role="alert">
       <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-      <%= safe_join(Array.wrap(flash[type]), '<br/>') %>
+      <%= safe_join(Array.wrap(flash[type]), '<br/>'.html_safe) %>
     </div>
     <% flash.delete(type) %>
   <% end %>
--- a/spec/views/_flash_msg.html.erb_spec.rb
+++ b/spec/views/_flash_msg.html.erb_spec.rb
@@ -1,0 +1,31 @@
+# coding: utf-8
+describe '/_flash_msg.html.erb', type: :view do
+  before do
+    allow(view).to receive(:flash).and_return(flash)
+  end
+
+  let(:flash) { { notice: notice } }
+  let(:i18n_html) { t('hyrax.works.new.after_create_html', application_name: 'Whatever') }
+
+  context 'with a single flash notice' do
+    let(:notice) { i18n_html }
+    it 'renders with HTML unescaped' do
+      render
+      expect(rendered).not_to have_content '</span>'
+    end
+  end
+
+  context 'with multiple flash notices' do
+    let(:notice) do
+      [
+        i18n_html,
+        'Lorem ipsum!'
+      ]
+    end
+    it 'renders the notice joined with unescaped line break' do
+      render
+      expect(rendered).not_to have_content '<br/>'
+      expect(rendered).not_to have_content '</span>'
+    end
+  end
+end


app/views/_flash_msg.html.erb
```diff
@@ -2,7 +2,7 @@
   <% if flash[type].present? %>
     <div class="alert <%= flash_dom_class %> alert-dismissable" role="alert">
       <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-      <%= safe_join(Array.wrap(flash[type]), '<br/>') %>
+      <%= safe_join(Array.wrap(flash[type]).map(&:html_safe), '<br/>'.html_safe) %>
```
jcoyne: These flashes should already be html safe. I think it's only the second argument that needs to be marked `html_safe`

mjgiarlo: @jcoyne They're not, afaict. When I remove the `#map` call, the tests fail.

mjgiarlo: @jcoyne You can see the commit that pulled in `safe_join` removed explicit calls to `#html_safe`: https://github.com/projecthydra-labs/hyrax/commit/be4a648f9e0f9191f514a241780dabab29c73777 So I believe we still need it.

jcoyne: @mjgiarlo but you aren't marking them html_safe in your test.

mjgiarlo: Good call.  My test was not setup correctly -- now it uses i18n properly.

mjgiarlo:
@jcoyne round two?

mjgiarlo:
@jcoyne @jeremyf @ggeisler I'm doing some in-browser QA and I still see these as escaped HTML. Can one of you verify that you are seeing the same w/ the latest from Hyrax? I'm back to wondering if we need to coerce `html_safe`ty on each entry in the flash. Is there a better solution?

atz:
@mjgiarlo et al.: There is already a convention for rails i18n including HTML-markup or not.  The convention is that the key end in `_html`.  See [docs](http://guides.rubyonrails.org/i18n.html#using-safe-html-translations).

So the first added test with `hyrax.works.new.after_create_html` should have been passing before this change.  

We don't know what strings are being added into flash overall.  It is up to the code appending values to flag them.  While tagging the separator `html_safe` is fine, the broader problems this doesn't solve include: 
- controller code, for example, adding messages with tags that are not flagged `html_safe`, 
- i18n values that contain tags but are not `_html`-named

In the latter category I find:
```
config/locales/hyrax.en.yml:        based_near: "A place name related to the collection, such as its site of publication, or the city, state, or country the collection contents are about. Calls upon the <a href='http://www.geonames.org'>GeoNames web service</a>."
config/locales/hyrax.en.yml:        based_near: "A place name related to the work, such as its site of publication, or the city, state, or country the work contents are about. Calls upon the <a href='http://www.geonames.org'>GeoNames web service</a>."
config/locales/hyrax.es.yml:        based_near: "Nombre de un lugar relacionado con la colección, tal como el sitio de publicación, o ciudad, estado, o país, que esté relacionado con los contenidos de la colección. Llamados a través del servicio de <a href='http://www.geonames.org'>GeoNames</a>."
config/locales/hyrax.es.yml:        based_near: "Nombre de un lugar relacionado con el trabajo, tal como el sitio de publicación, o ciudad, estado, o país, que esté relacionado con los contenidos de el trabajo. Llamados a través del servicio de <a href='http://www.geonames.org'>GeoNames</a>."
```

That might be the source of the escaping you are seeing.  (Solution is to change the key name.)

mjgiarlo:
@atz This is why I'm confused: because in this case, AFAICT, there is only one flash entry and it uses the `_html` convention for automatic `html_safe`ty. And yet, there are those HTML tags in the UI.

jcoyne:
This is odd:
```
I18n.t('hyrax.works.new.after_create_html').html_safe?
=> false
```

jcoyne:
The docs say:
> When you use them in views the HTML will not be escaped.

So, maybe there's something special about calling the `t` method in the view context that does it. Probably need to mark that translation explicitly html_safe in the controller.

jcoyne:
Yep: https://github.com/rails/rails/blob/a4dbbde4c91a52e28b6c4772b0032b9c16d636ca/actionview/lib/action_view/helpers/translation_helper.rb#L79-L89

jcoyne:
Looks like calling `view_context.t` will do it:
```
irb(main):011:0> CatalogController.new.t('hyrax.works.new.after_create_html').html_safe?
=> false
irb(main):012:0> CatalogController.new.view_context.t('hyrax.works.new.after_create_html').html_safe?
=> true
```

PR #350, opened 2017-02-02
Adds file count to admin sets on admin dash.

### LINKED ISSUE #241: Count files per collection, display on activity summary
See projecthydra-labs/hyku#431

For a first cut only count number of child (not all decedent) files

andjsmit:
Fixes #241 

Adds files count to administration dashboard under collections.

Accomplished by counting the file sets from the Solr query that identities the number of works in a admin set. Includes English and Spanish locales for column label.

### DIFF
--- a/app/services/hyrax/admin_set_service.rb
+++ b/app/services/hyrax/admin_set_service.rb
@@ -18,9 +18,10 @@ module Hyrax
       response.documents
     end
 
-    # This performs a two pass query, first getting the AdminSets and then getting the work counts
+    # This performs a two pass query, first getting the AdminSets
+    # and then getting the work and file counts
     # @param [Symbol] access :read or :edit
-    # @return [Array<Array>] a list with document, then work count
+    # @return [Array<Array>] a list with document, then work and file count
     def search_results_with_work_count(access)
       documents = search_results(access)
       ids = documents.map(&:id).join(',')
@@ -32,8 +33,9 @@ module Hyrax
                   'facet.field' => join_field }
       )
       counts = results['facet_counts']['facet_fields'][join_field].each_slice(2).to_h
+      file_counts = count_files(results)
       documents.map do |doc|
-        [doc, counts[doc.id].to_i]
+        [doc, counts[doc.id].to_i, file_counts[doc.id]]
       end
     end
 
@@ -63,6 +65,19 @@ module Hyrax
         attrs['data-release-before-date'] = true if permission_template.release_before_date?
         attrs['data-visibility'] = permission_template.visibility unless permission_template.visibility.blank?
         attrs
+      end
+
+      # Count number of files from admin set works
+      # @param [Array] results Solr search result
+      # @return [Hash] admin set id keys and file count values
+      def count_files(results)
+        file_counts = Hash.new(0)
+        results['response']['docs'].each do |doc|
+          doc['isPartOf_ssim'].each do |id|
+            file_counts[id] += doc['file_set_ids_ssim'].length
+          end
+        end
+        file_counts
       end
   end
 end
--- a/app/views/hyrax/admin/_collections.html.erb
+++ b/app/views/hyrax/admin/_collections.html.erb
@@ -13,6 +13,7 @@
             <tr>
               <th><%= t('.collection') %></th>
               <th><%= t('.works') %></th>
+              <th><%= t('.files') %></th>
             </tr>
           </thead>
           <tbody>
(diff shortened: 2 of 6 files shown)


app/services/hyrax/admin_set_service.rb
```diff
@@ -64,5 +66,17 @@ def data_attributes(admin_set)
         attrs['data-visibility'] = permission_template.visibility unless permission_template.visibility.blank?
         attrs
       end
+
+      # Count number of files from admin set works
+      # @param [Array] Solr search result
+      def count_files(results)
```
jcoyne: This approach is only going to count files for the number of works returned.  This defaults to 10, right?

app/services/hyrax/admin_set_service.rb
```diff
@@ -64,5 +66,17 @@ def data_attributes(admin_set)
         attrs['data-visibility'] = permission_template.visibility unless permission_template.visibility.blank?
         attrs
       end
+
+      # Count number of files from admin set works
+      # @param [Array] Solr search result
```
jcoyne: YARD requires the variable name

jcoyne: Add a `@return` YARD comment that explains it returns a hash with id as keys and counts as values.

app/services/hyrax/admin_set_service.rb
jcoyne: Ah, this will be a problem because we don't have works here, we have admin sets due to https://github.com/projecthydra-labs/hyrax/blob/master/app/search_builders/hyrax/admin_set_search_builder.rb#L16.  I'm not sure you ever see file_set_ids_ssim in your results, will you?

andjsmit:
As far as I can tell, I see file sets in the returned results. Found this when checking the rails log. The numbers match what I expect testing on a demo app.

andjsmit:
Example results:

```
{
  "responseHeader"=>{
    ...
    }
  },
  "response"=>{
    "numFound"=>2,
    ...
    "docs"=>[
      {
        ...
        "isPartOf_ssim"=>["gq67jr16q"],
        ...
        "member_ids_ssim"=>["44558d285", "6m311p28w", "qj72p712h"],
        "file_set_ids_ssim"=>["44558d285", "6m311p28w", "qj72p712h"],
        "admin_set_tesim"=>["First Admin Set"],
        ....
      },
  "facet_counts"=>{
    "facet_queries"=>{},
    "facet_fields"=>{
      "isPartOf_ssim"=>["gq67jr16q", 1, "q811kj606", 1]
    },
   ...
  }
}
```

jcoyne:
Okay, this must be doing a join query then.  Second problem: what's the maximum number of rows this query is returning?  By default, it'll probably be 10 (or maybe 1000), but this does not seem sufficient.  We may have more than `500,000` files in an admin set.

jcoyne:
Hmm, I'm not seeing that join anywhere and I'm seeing this in fq: `{!terms f=has_model_ssim}AdminSet`

jcoyne:
@andjsmit it looks to me like you might have misread the logs. Here's what I get:
https://gist.github.com/jcoyne/08c98a4c073910b9d3719de0dbbc3b25

andjsmit:
@jcoyne I'm not sure I get your gist :-) I'm pulling out the file count using the Solr query inside the search_results_with_work_count method, not from it's result. So I'm thinking this might be closer to what I'm doing: https://gist.github.com/andjsmit/9b74479dde18b3c4bce6405a05a09dcf

jcoyne:
>I'm pulling out the file count using the Solr query inside the search_results_with_work_count method

@andjsmit can you show me where you are doing that? As far as I can tell `results` can only contain AdminSet objects.  Perhaps you could create an integration test that shows that the correct counts are displaying on the admin dashboard?

andjsmit:
@jcoyne Doesn't `results` contain all the solr documents that are assigned to any admin set because of the query asking for all docs that hold an admin set id it's isPartOf_ssim field? Sorry if my understanding is weak on this. It just seems like a the info I need is readily available without much fuss.

jcoyne:
@andjsmit As far as I can tell, results only contains AdminSets.  I could be wrong. That's why I'm asking you to create an integration test to show me.

andjsmit:
@jcoyne Ok, I'll work on creating an integration test. Would the proper place for such a test be in `spec/features/admin_admin_set_spec.rb`? Thanks.

jcoyne:
@andjsmit if that scenario already renders this page, then that's an idea place.

andjsmit:
@jcoyne @mjgiarlo PR now includes an integration test. Ready for a re-review.

jcoyne:
@andjsmit Thanks for adding that feature spec. That really aided in my understanding. Good job.

PR #355, opened 2017-02-03
Merge the latest commits from sufia

jcoyne:


config/locales/hyrax.en.yml
```diff
@@ -12,6 +12,7 @@ en:
       admin_set:
         new: "Create new administrative set"
       cancel: "Cancel"
+      delete: "Delete"
```
mjgiarlo: Can you toss another commit in this branch that includes Spanish translations to make sure our locales stay in sync?

jcoyne: Done

PR #367, opened 2017-02-07
Ensuring User groups can be built as expected

jeremyf:
## Ensuring User groups can be built as expected

@4a52156a849ac8be5208df8672cedf6de86cc4f5

Prior to this commit the following was happening:

```ruby
admin_set = FactoryGirl.create(:admin)
admin_set.groups == ['admin']
=> true
found_admin_set = User.find(admin_set.id)
found_admin_set.groups == ['admin']
=> false
```

This is because the method redefinition was applying only to the in
memory version of the user that was being created. If we were to run
`User.find` and get back a new in-memory version, the user's `groups`
override would not be applied.

This change leverages:

* [ActiveRecord's identity][1]
* [RSpec::Mocks.allow_message][2]
* [FactoryGirl's transient attributes][3]

And with this change, we can now specify what each user's groups should
be as part of the `FactoryGirl.create` or `.build` method.

In addition this commit replaces direct reference to `::RoleMapper`
instead preferring [`::User.group_service`][4]. The change also removes
some of the now needless spec stubbing of ::RoleMapper by leveraging
the FactoryGirl stubbing that now automatically occurs.

[1]:http://api.rubyonrails.org/classes/ActiveRecord/Core.html#method-i-3D-3D
[2]:https://github.com/rspec/rspec-mocks/blob/2d67315c09e8d21a470f25d511bbdea287d89c4c/lib/rspec/mocks.rb#L57-L71
[3]:https://github.com/thoughtbot/factory_girl/blob/d42595f4696a986d626a6b84e2592e85c5bb4c4f/GETTING_STARTED.md#transient-attributes
[4]:https://github.com/projecthydra/hydra-head/blob/v10.4.0/hydra-access-controls/lib/hydra/user.rb#L8-L11

### DIFF
--- a/lib/hyrax/collections/search_service.rb
+++ b/lib/hyrax/collections/search_service.rb
@@ -33,7 +33,7 @@ module Hyrax
         # Grant access based on user id & role
         unless @user_key.blank?
           # for roles
-          ::RoleMapper.roles(@user_key).each do |role|
+          ::User.group_service.roles(@user_key).each do |role|
             user_access_filters << "#{solr_access_control_suffix(:group)}:#{escape_slashes(role)}"
           end
           # for individual person access
--- a/spec/abilities/ability_spec.rb
+++ b/spec/abilities/ability_spec.rb
@@ -63,7 +63,7 @@ describe 'Hyrax::Ability', type: :model do
     end
 
     describe 'as admin' do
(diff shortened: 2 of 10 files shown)


mjgiarlo:
LGTM, but I'd like @jcoyne to take a look.

Issue #249, opened 2017-01-11
Manage users page should have a filter for only admins


jcoyne:
Making it easy to find the admin users amongst all the non-admins. see projecthydra-labs/hyku#444 for mockup

mjgiarlo:
Related to #248, #247. Consider using datatables for the implementation (already being used for admin/admin_sets index view).

laritakr:
The use of datatables in the admin users page takes care of this requirement.

Issue #248, opened 2017-01-11
Manage users page should have a search for user


jcoyne:
This filters the rows that display to those that match the search.

see projecthydra-labs/hyku#444 for mockup

mjgiarlo:
Related to #249, #247. Consider using datatables for the implementation (already being used for admin/admin_sets index view).

laritakr:
The use of datatables in the users page takes care of this requirement.

Issue #247, opened 2017-01-11
Admin page should have a paginated list of users.


jcoyne:
On the sidebar it's called "Manage users"

Show user name, avatar if they are an admin or not and their date of last access. This should link to the user-role edit page.

see projecthydra-labs/hyku#444 for mockup

- [ ] Add a section for user count and administrator count
- [ ] Add pagination for user page
- [ ] Add columns of Name, Username, Roles, Department, Last Access
- [ ] Add select option for updating pagination page size
- [ ] Ensure the templates for this page in Hyrax exist in Hyku as overrides
- [ ] Answer what roles we are rendering from the mockup (e.g. the Hyku `user_roles` table values, or the Hyrax `sipity_roles + sipity_workflow_roles` join)

At present, #248 is for the Filter by User Name, #249 is for Filter by Role

mjgiarlo:
Related to #248, #249. Consider using datatables for the implementation (already being used for admin/admin_sets index view).

mjgiarlo:
Feel free to take what's currently available at `/admin/users` (currently a submenu of Roles and Permissions), and move it into the menu structure as specified in the mockup.

jeremyf:
Note, self assigned, but @laritakr and I are going to spend this afternoon on task breakdown.

jeremyf:
@mjgiarlo For the Roles column in the UI, what is the source of those Roles?

laritakr:
@ggeisler the [manage users page as designed](https://cloud.githubusercontent.com/assets/101482/20907182/edf68dd6-bb09-11e6-872b-4ec63c7601eb.png) shows content that is not exposed for entry (user name, department). Should they still be exposed on this page with blank data, or is other information desired instead?

ggeisler:
@laritakr I don't think we should show fields that we know will be blank for all users, so I'd leave those out for now. I'll make another ticket that proposes to add other fields I think we'd want to enable users to add. Sorry for the confusion.

laritakr:
closed via https://github.com/projecthydra-labs/hyrax/pull/547

PR #385, opened 2017-02-09
Create a paginated list of users

laritakr:
**Resolves:** 
- https://github.com/projecthydra-labs/hyrax/issues/247
- https://github.com/projecthydra-labs/hyrax/issues/248
- https://github.com/projecthydra-labs/hyrax/issues/249

**Creates a paginated list of users**
- make a UserPresenter class and make it functional
- add breadcrumbs for page
- add User scope :not_batch_or_audit_user for user count
- modify specs accordingly
- incorporate Spanish translation (other than for datatables)

**Note:** By using datatables, this PR also incorporates the sorting and searching capabilities as specified in issues 248 and 249. However datatables currently does not support Spanish translation. This needs to be handled in a separate issue.

Not all of the columns requested in the issues are currently in use. Page was created with only currently valid content.
screen shot 2017-02-10 at 2 11 52 pm


app/models/concerns/hyrax/user.rb
```diff
@@ -24,6 +24,7 @@ module Hyrax::User
 
     scope :guests, ->() { where(guest: true) }
     scope :registered, ->() { where(guest: false) }
+    scope :not_batch_or_audit_user, -> { where("#{Devise.authentication_keys.first} not in (?)", [::User.batch_user_key, ::User.audit_user_key]) }
```
mjgiarlo: What would you think renaming this something like `:without_system_accounts` so that we can add other accounts to this in the future without touching the places in the code that use this scope?

app/presenters/hyrax/users_presenter.rb
```diff
@@ -0,0 +1,55 @@
...
+      @users = search(query)
+    end
+
+    def user_count
+      ::User.not_batch_or_audit_user.count
+    end
+
+    def repository_admin_count
+      count_admin = 0
+      ::User.not_batch_or_audit_user.each do |user|
```
mjgiarlo: Would a more targeted iterator work here? Like this one-liner:

```ruby
::User.without_system_accounts.select { |user| admin?(user) }.count
```

app/presenters/hyrax/users_presenter.rb
```diff
@@ -0,0 +1,55 @@
...
+        count_admin += 1 if admin?(user)
+      end
+      count_admin
+    end
+
+    def user_roles(user)
+      roles = user.groups
+      return roles if roles.any?
+      ['']
```
mjgiarlo: Can you say a bit about why this returns `['']` if `roles` is empty?

app/presenters/hyrax/users_presenter.rb
```diff
@@ -0,0 +1,55 @@
...
+      @authentication_key = authentication_key
+    end
+
+    # @return [Array] an array of Users
+    def users
+      @users = search(query)
+    end
+
+    def user_count
+      ::User.not_batch_or_audit_user.count
```
mjgiarlo: Should this also include the `registered` scope so it doesn't include guest accounts?

app/views/hyrax/users/index.html.erb
```diff
@@ -1,21 +1,50 @@
...
-  <%= paginate @users, theme: 'blacklight', route_set: hyrax %>
+<% content_for :page_header do %>
+  <h1><span class="fa fa-user"></span> <%= t('hyrax.users.index.title') %></h1>
+<% end %>
+
+<div class="panel panel-default users-listing">
+  <div class="panel-heading">
+    <% count_users = @presenter.user_count %>
+    <% admin_count = @presenter.repository_admin_count %>
+    <% if count_users == 1 %>
```
mjgiarlo: Hm, in a repository with a single user, I would expect that user to also be an admin. I'm not sure that's necessarily the case, though. I wonder if we could move this logic into the presenter, as something like `def heading`?

app/views/hyrax/users/index.html.erb
```diff
@@ -1,21 +1,50 @@
...
+            <th></th>
+            <th><%= t('.id_label') %></th>
+            <th><%= t('.role_label') %></th>
+            <th><%= t('.access_label') %></th>
+          </tr>
+        </thead>
+        <tbody>
+          <% @presenter.users.each do |user| %>
+            <tr>
+              <td><%= link_to hyrax.profile_path(user) do %>
```
mjgiarlo: Can you add a line break after the `<td>`?

app/views/hyrax/users/index.html.erb
```diff
@@ -1,21 +1,50 @@
...
+                    <%= image_tag(user.avatar.url(:thumb), width: 30) if user.avatar.file %>
+                  <% end %>
+              </td>
+              <td><%= link_to user.email, hyrax.profile_path(user) %></td>
+              <% @presenter.user_roles(user).each do |role| %>
+                <td><%= role %></td>
```
mjgiarlo: Does this mean that rows may have different numbers of cells depending on the number of roles for the user in that particular row?

app/views/hyrax/users/index.html.erb
```diff
@@ -1,21 +1,50 @@
...
+<% content_for :page_header do %>
+  <h1><span class="fa fa-user"></span> <%= t('hyrax.users.index.title') %></h1>
+<% end %>
+
+<div class="panel panel-default users-listing">
+  <div class="panel-heading">
+    <% count_users = @presenter.user_count %>
+    <% admin_count = @presenter.repository_admin_count %>
+    <% if count_users == 1 %>
+      <%= t('hyrax.users.index.describe_users_singular_html', nbr_admin: admin_count) %>
```
mjgiarlo: Can you spell out the word 'number' in these calls to `#t()`?

config/locales/hyrax.es.yml
```diff
@@ -11,6 +11,47 @@ es:
           based_near: "Ubicación"
           contributor: "Colaboradores"
           keyword: "Palabra Clave"
+
+  datetime:
```
mjgiarlo: Can you add these to the English locale file too?

mjgiarlo:
Fixes #247 
Fixes #248 
Fixes #249

mjgiarlo:
Can you also include a screenshot or animated gif showing the view changes? Thx!

app/models/concerns/hyrax/user.rb
laritakr: That's a good suggestion. Will do.

app/presenters/hyrax/users_presenter.rb
laritakr: I'm not sure on this one. In this, I replicated the prior functionality, which included guest. I was actually going to ask which was preferred.

app/presenters/hyrax/users_presenter.rb
laritakr: As this is a presenter, if roles come back as nil, I had to qualify, it misaligns the columns on the view. The logic needed to either be in here or in the view file so I put it here.

app/views/hyrax/users/index.html.erb
laritakr: Yes, unlikely to happen in reality, though it could happen depending on how tests are set up. I believe I mimicked the logic from the way it was showing elsewhere, but I'm find with putting it into the presenter. That definitely would simplify the view.

app/views/hyrax/users/index.html.erb
laritakr: Possibly... I intended it to be separate lines in the row! Though currently the ONLY possible role is 'admin' so I probably didn't test this.

app/views/hyrax/users/index.html.erb
laritakr: I'm not sure what the purpose would be of this. It just makes the spacing look strange. Amy I missing something?

mjgiarlo: Nevermind this one. Thanks!

app/presenters/hyrax/users_presenter.rb
```diff
@@ -0,0 +1,55 @@
...
+    # @return [Array] an array of Users
+    def users
+      @users = search(query)
+    end
+
+    def user_count
+      ::User.registered.without_system_accounts.count
+    end
+
+    def repository_admin_count
```
mjgiarlo: Would a more targeted iterator work here? Like this one-liner:

```ruby
::User.without_system_accounts.select { |user| admin?(user) }.count
```

app/views/hyrax/users/index.html.erb
```diff
@@ -1,21 +1,47 @@
...
+              <td><%= link_to hyrax.profile_path(user) do %>
+                    <%= image_tag(user.avatar.url(:thumb), width: 30) if user.avatar.file %>
+                  <% end %>
+              </td>
+              <br>
+              <td><%= link_to user.email, hyrax.profile_path(user) %></td>
+              <td><% roles = @presenter.user_roles(user) %>
+                  <ul><% roles.each do |role| %>
+                    <li><%= role %></li>
+                    <% end if roles.any? %>
```
mjgiarlo: I'm not sure you need the `#any?` check here. If `roles` is empty, none of the code in the block will be executed.

laritakr: Right... missed removing that one.

app/presenters/hyrax/users_presenter.rb
jcoyne: This is a very expensive operation. It will iterate over all the users in the system.  This could be in the thousands. Can we count admin roles instead? I'm comfortable with the assumption that the system accounts will not be granted an admin role.

mjgiarlo:
@jcoyne this doesn't use a SQL COUNT? or even then are you concerned about performance?

jcoyne:
This doesn't use a count query. The select reifies each record.

app/views/hyrax/users/index.html.erb
```diff
@@ -1,21 +1,46 @@
...
+        <tbody>
+          <% @presenter.users.each do |user| %>
+            <tr>
+              <td><%= link_to hyrax.profile_path(user) do %>
+                    <%= image_tag(user.avatar.url(:thumb), width: 30) if user.avatar.file %>
+                  <% end %>
+              </td>
+              <br>
+              <td><%= link_to user.email, hyrax.profile_path(user) %></td>
+              <td><% roles = @presenter.user_roles(user) %>
```
jcoyne: This page is used to show the list of users to the public.  I don't think we want to put the user roles on this page. Rather it should be a new page that is only accessible to admins (See issue #247)

laritakr: I implemented the page to spec, and this is accessible off of the admin menu so it is the appropriate place to show user roles. I don't see a need for me to second-guess the design that @ggeisler came up with. If someone wants another page for general users to see publicly, then that should be a separate implementation.

jcoyne: The problem is, we already are using this template for another purpose. You've repurposed it for the admin page.  And changed it's original purpose.

jcoyne: You can see here, this template is for the existing public view of the users: https://hierophant.demo.hydrainabox.org/users?locale=en.

laritakr: Yes, that's what the specs said to do. 

I believe that wee need to discuss this at today's meeting, because I don't think simply removing the column is the answer. We may want to redo my code into a new page and leave the existing one as is (or maybe @ggeisler would want changes to the existing one as well), but if we want a users page for the admin menu, we don't want to simply undo what I've already done.

jcoyne: Yes, I think copying this template to a new location (probably `admin/users/`) and then making your additions there is the right approach.

laritakr: Yes that location makes more sense to me too. But we probably still want to discuss at the meeting whether we want this issue to include the similar changes to be on the existing user list or if there will be another issue to update that menu.

mjgiarlo: IMO, let's punt on changes to the existing (publicly visible) user list. Thanks for making this change, @laritakr, and sorry this issue was unclear!

laritakr: The existing list will require some changes anyway, given that I pulled out a bunch of the code into a presenter class. So if I'm making changes anyway, it makes some sense to consider what we want it to look like.

mjgiarlo: OK, @laritakr. Let's do discuss this briefly on today's stand-up, then. Thank you.

laritakr:
Replaced with https://github.com/projecthydra-labs/hyrax/pull/547

PR #393, opened 2017-02-13
Check current visibility value on form

laritakr:
Add checked: option to radio button for visibility form to show the currently selected visibility value.

Fixes https://github.com/projecthydra-labs/hyku/issues/744


app/views/hyrax/admin/admin_sets/_form_visibility.html.erb
```diff
@@ -44,7 +44,7 @@
                   <h3><%= t('.visibility.title') %></h3>
                   <p><%= t('.visibility.description') %></p>
                   <%# List each option in a <div class='radio'> tag %>
-                  <%= f.collection_radio_buttons :visibility, f.object.visibility_options, :first, :last, item_wrapper_tag: :div, item_wrapper_class: 'radio' do |b| %>
+                  <%= f.collection_radio_buttons :visibility, f.object.visibility_options, :first, :last, item_wrapper_tag: :div, item_wrapper_class: 'radio', checked: f.object[:visibility] do |b| %>
```
jcoyne: This is the default behavior for simple_form, isn't it? Is there a reason we need to make this explicit?

laritakr: This was not behaving as default in this situation, and making it explicit solves the issue. I'm not sure why it wasn't finding it by default... perhaps because the permission template has so many levels and it wasn't finding the right items to check? The collection of radio buttons is based on f.object.visibility_options, but the checked item comes from f.object[:visibility]

jcoyne: @laritakr could it be that simple_form is by default doing `f.object['visibility']` (with a String, not a Symbol)? Can you check that the form object is returning the correct value if either a string or symbol is passed to the `[]` method?

jcoyne: The reason this isn't working is because Tag::Base gets the value by sending `:visibility` to the form object (https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/actionview/lib/action_view/helpers/tags/base.rb#L28)  this visibility method has been used by this method: https://github.com/projecthydra-labs/hyrax/blob/91729ff3bf48e1d7be1084bbe418d0c393ce46a2/app/forms/hyrax/forms/permission_template_form.rb#L15. So you probably want to rename that method.  Then make a "visibility" method that just returns the correct value.

jcoyne: @laritakr this could fix it: https://github.com/projecthydra-labs/hyrax/pull/476

laritakr:
Closing: replaced with PR 490 due to inability to squash commits on this one.

PR #410, opened 2017-02-14
Wrap the actor stack in a transaction

jcoyne:
This should roll back any workflow or other database changes if
there is an exception somewhere in the actor stack (e.g. can't write to
solr).  This will allow us to repeat the action without encountering
RecordNotUnique exceptions when getting to the InitializeWorkflowActor
when attempting an action a second time.

Fixes https://github.com/projecthydra-labs/hyku/issues/746

### DIFF
--- a/spec/actors/hyrax/actors/transactional_request_spec.rb
+++ b/spec/actors/hyrax/actors/transactional_request_spec.rb
@@ -1,0 +1,30 @@
+require 'spec_helper'
+
+RSpec.describe Hyrax::Actors::TransactionalRequest do
+  let(:bad_actor) do
+    Class.new(Hyrax::Actors::AbstractActor) do
+      def create(attributes)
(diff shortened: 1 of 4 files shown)


spec/actors/hyrax/actors/transactional_request_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+                                                    Hyrax::Actors::InitializeWorkflowActor])
+  end
+
+  let(:depositor) { create(:user) }
+  let(:work) { create(:work) }
+
+  describe "create" do
+    subject { actor_stack.create({}) }
+
+    it "returns true and calls the next actor without the version attribute" do
```
jeremyf: I think this spec description is not reflective of what is happening.

jcoyne: yeah, I just changed it.

PR #412, opened 2017-02-14
Move all actors into the same namespace for consistency

jcoyne:


jeremyf:
@jcoyne Looks like a morning of rebases

PR #414, opened 2017-02-14
Refactor actors to pass Ability

### LINKED ISSUE #408: CanCan should be used to check if users can append to a collection.
https://github.com/projecthydra-labs/hyrax/blob/master/app/actors/hyrax/actors/add_as_member_of_collections_actor.rb#L20

That should be something like:

```ruby
          other_collections = curation_concern.member_of_collections.select { |coll| !ability.can?(:edit, coll) }
```

Related to #407

jcoyne:
Fixes #408

### DIFF
--- a/app/actors/hyrax/actors/abstract_actor.rb
+++ b/app/actors/hyrax/actors/abstract_actor.rb
@@ -16,11 +16,11 @@ module Hyrax
     class AbstractActor
       attr_reader :next_actor
 
-      def initialize(_curation_concern, _user, next_actor)
+      def initialize(_curation_concern, _ability, next_actor)
         @next_actor = next_actor
       end
 
-      delegate :curation_concern, :user, to: :next_actor
+      delegate :curation_concern, :ability, to: :next_actor
 
       delegate :create, to: :next_actor
 
--- a/app/actors/hyrax/actors/actor_stack.rb
+++ b/app/actors/hyrax/actors/actor_stack.rb
@@ -1,20 +1,20 @@
 module Hyrax
   module Actors
     class ActorStack
-      attr_reader :curation_concern, :user, :first_actor_class, :more_actors
-      def initialize(curation_concern, user, more_actors)
+      attr_reader :curation_concern, :ability, :first_actor_class, :more_actors
+      def initialize(curation_concern, ability, more_actors)
         @curation_concern = curation_concern
-        @user = user
+        @ability = ability
         @more_actors = more_actors
         @first_actor_class = @more_actors.shift || RootActor
       end
 
       def inner_stack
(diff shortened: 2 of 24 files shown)


tpendragon:
This is a pretty major contract change, yeah?

jcoyne:
@tpendragon that's why it's 2.0.

tpendragon:
I understand why we'd want to change it to this eventually, but in the interest of making Hyrax 2.0 an easy CC migration target, can we accomplish this by just doing

```ruby
def ability
  ::Ability.new(user)
end
```
?

jcoyne:
@tpendragon I'm trying to be considerate of the case where Ability is influenced by inputs other than just the user, such as IP range.

tpendragon:
@jcoyne Absolutely, and injecting the ability rather than relying on the concretion of the `Ability` class is a good idea! But I think in the past we've been too laissez-faire with our contracts. Maybe a happy medium is a way to deprecate the use of a user, but still accept it?

jcoyne:
@tpendragon I'm hesitant to bother without some evidence that anyone is actually customizing their ActorStack.  Do you know of any?

tpendragon:
@jcoyne Off the top of my head, no. Are we at the point where we're changing promises because we personally don't know of anyone that took us up on it, though? This is the primary processing pipeline and the mechanism we provide for changing that - it seems short-sighted to assume nobody's using it. And even if they aren't, do we have no responsibility towards providing those paths as a matter of practice?

This doesn't effect me personally, but we said Hyrax 2.0 was UI changes that CC people could migrate to. This points towards "whatever breaking changes we can fit in."

tpendragon:
I'm working on an alternative to this now, to talk about.

app/actors/hyrax/actors/abstract_actor.rb
```diff
@@ -16,11 +16,11 @@ module Actors
     class AbstractActor
       attr_reader :next_actor
 
-      def initialize(_curation_concern, _user, next_actor)
+      def initialize(_curation_concern, _ability, next_actor)
```
jeremyf: Wondering if we can have the following signature:

```ruby
def initialize(_curation_concern, user, next_actor, ability: default_ability)
  # insert setters here
  @next_actor = next_actor
  self.user = user
end

private

def default_ability
  Ability.new(@user)
end

def user=(input)
  # Deprecation notice saying to favor an ability
  @user = user
end
```

It preserves the current interface, but allows for an ability to be passed. I believe there would also be

tpendragon:
https://github.com/projecthydra-labs/hyrax/compare/refactor_actors_and_deprecate I think this is what it would look like if we allowed backwards compatibility.

PR #418, opened 2017-02-15
Fix contract for FileSetsControllerBehavior

### LINKED ISSUE #400: FileSetsController is broken
There's no concept in Hyrax of uploading a file straight to a FileSet, so I assume this code just got orphaned in the move. However, plum's specs brought it to light:

https://github.com/projecthydra-labs/hyrax/blob/master/app/controllers/concerns/hyrax/file_sets_controller_behavior.rb#L107

That code is the wrong signature for `create_metadata`. It should be something like

```
actor.create_metadata(params[:file_set])
actor.attach_file_to_work(find_parent_by_id)
```

tpendragon:
Closes #400

We'll probably end up removing this bit of the FileSetsController at
some point, as it no longer appears to be used.

### DIFF
--- a/app/controllers/concerns/hyrax/file_sets_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/file_sets_controller_behavior.rb
@@ -104,8 +104,7 @@ module Hyrax
           empty_file_response(file)
         else
           update_metadata_from_upload_screen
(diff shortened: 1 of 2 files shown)


tpendragon:
@jcoyne Adjusted the existing specs.

PR #419, opened 2017-02-15
Fix default workflow strategy.

### LINKED ISSUE #364: Can't override WorkflowStrategy
### Descriptive summary

You should be able to set the WorkflowStrategy on the WorkflowFactory, but it's being overridden by the initializer defined in Hyrax's engine. Should just be able to update https://github.com/projecthydra-labs/hyrax/blob/master/lib/hyrax/engine.rb#L60 to be `||=`

However, it was definitely confusing that there's a default strategy in the factory (https://github.com/projecthydra-labs/hyrax/blob/master/lib/hyrax/engine.rb#L60), but it'll never get called - probably an artifact of Sufia behavior.

tpendragon:
We were originally setting this to a default and immediately overriding
it, which was a side effect of this behavior being in
CurationConcerns.

Closes #364.

### DIFF
--- a/app/services/hyrax/workflow/workflow_factory.rb
+++ b/app/services/hyrax/workflow/workflow_factory.rb
@@ -2,11 +2,11 @@ module Hyrax
   module Workflow
     class WorkflowFactory
       class_attribute :workflow_strategy
-      self.workflow_strategy = DefaultWorkflowStrategy
+      self.workflow_strategy = Hyrax::Workflow::WorkflowByAdminSetStrategy
 
       # @param work [#to_global_id]
       # @param attributes [Hash]
-      # @param strategy [#name] strategy for finding which workflow to use. Defaults to an instance of DefaultWorkflowStrategy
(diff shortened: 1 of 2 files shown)


app/services/hyrax/workflow/workflow_factory.rb
```diff
@@ -2,7 +2,7 @@ module Hyrax
   module Workflow
     class WorkflowFactory
       class_attribute :workflow_strategy
-      self.workflow_strategy = DefaultWorkflowStrategy
+      self.workflow_strategy = Hyrax::Workflow::WorkflowByAdminSetStrategy
```
mjgiarlo: Two questions:

1. I wonder if this means we should also dump the `DefaultWorkflowStrategy` since it's no longer being used?
2. Update line 9 below?

tpendragon: I think it's probably useful to provide, especially for CC migrators who may have gotten to CC 2.0 and not customized their workflow strategy. We had already customized it to support geo concerns.

tpendragon: Fixed comment.

PR #425, opened 2017-02-15
Fix FileManagerForm

tpendragon:
It was referencing non-existent methods before.


jcoyne:
This has already been backported. Let's just merge instead, ok?

tpendragon:
The fix was ported? Because without this the file manager doesn't work

jcoyne:
@tpendragon can you confirm that this is no longer required?

tpendragon:
Confirmed this is no longer needed.

PR #433, opened 2017-02-16
Add rake task to create default_admin_set

### LINKED ISSUE #237: Default admin set should be created earlier
Default admin set should be created earlier than when the first work w/o an admin set is submitted.

scholarworks:
Adds the default_admin_set generator.

Fixes #237 

Adds a generate called by:

rails generate hyrax:default_admin_set

This is NOT included in the insall generator stack as teh admin set requires a running fedora instance on create.


scholarworks:
@mjgiarlo That was my original goal, yes. So any guidance in how we may tackle that based on the need for the fedora server would be great.

jeremyf:
I don't know the impact of this being part of the boot sequence of the application. However I had suggested allowing the default admin set to be part of the `Hyrax.config`.

I believe we could add an after_prepare hook to the Rail tie that would ensure the AdminSet is created. See Hydra::Groupy for examples of `to_prepare` (https://github.com/projecthydra-labs/hydra-groupy/blob/master/lib/hydra/groupy/railtie.rb)

mjgiarlo:
That would work for me, and I don't know if we have any better hooks for firing when e.g. Fedora spins up. IMO it's reasonable to assume that backend services are already up when Rails initializes.

scholarworks:
@jeremyf Cool. So the to_prepare in railtie you have here runs at the end of the rails boot process? Then checking if there is a Default admin set as part of the config makes a lot more sense. I'll take a stab at that this afternoon sometime. Thanks!

scholarworks:
FYI - This isn't working yet. Don't merge just because it passed ;-)

lib/hyrax/default_admin_set/railtie.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails/railtie'
+
+module Hyrax
+  module DefaultAdminSet
+    # Connect into the boot sequence of a Rails application
+    class Railtie < Rails::Railtie
+      config.to_prepare do
+        begin
+          AdminSet.find(Hyrax::DefaultAdminSetActor::DEFAULT_ID)
```
mjgiarlo: I have thoughts about how we might streamline this Railtie and DRY out some other code (the DefaultAdminSetActor). Mind if I toss a commit at this branch?

scholarworks: Please do.

jcoyne:
@scholarworks please update the title of this PR to reflect what is actually happening.

app/models/concerns/hyrax/admin_set_behavior.rb
```diff
@@ -32,6 +32,18 @@ def self.default_set?(id)
         id == DEFAULT_ID
       end
 
+      # Creates the default AdminSet and an associated PermissionTemplate with workflow
+      def self.create_default!
```
jcoyne: Please use `# rubocop:disable Lint/HandleExceptions` as we had before (and remove the exception from the rubocop config`  this ensure that only this method is allowed this exception and not the whole of the file.

scholarworks:
Closing in lieu of clean branch PR #477.

PR #445, opened 2017-02-17
Derivative service shared examples & Configurable Derivatives

tpendragon:
This does a few things:

1. Creates a shared example for DerivativeServices that can be called in other apps, for cases where users want their own derivative logic.

2. Adds a `#valid?` method to DerivativeServices which checks to see if it can handle a given fileset.

3. Adds a factory method `Hyrax::DerivativeService.for` which will return the first appropriate DerivativeService given a configured set.

Some questions:

A. Do we like this pattern?
B. Is the shared spec correct? For example, is `#uri` and `#mime_type` a part of the interface, or just for that specific derivative service? Is `#file_set` part of the interface? SHOULD we be checking if `#valid?` works?
C. Should we blow up FileSetDerivativeService into one per mime type, now that it's easy? (Maybe another PR, but that's a benefit of this refactor.)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,7 @@ AllCops:
     - 'script/**/*'
     - 'spec/test_app_templates/**/*'
     - 'vendor/**/*'
+    - 'lib/hyrax/specs/**/*'
 
 Lint/ImplicitStringConcatenation:
   Exclude:
@@ -114,6 +115,10 @@ RSpec/MessageSpies:
     - 'spec/controllers/hyrax/api/zotero_controller_spec.rb'
     - 'spec/controllers/hyrax/generic_works_controller_spec.rb'
     - 'spec/services/hyrax/workflow/workflow_importer_spec.rb'
+
+RSpec/InstanceVariable:
+  Exclude:
+    - 'spec/services/hyrax/derivative_service_spec.rb'
 
 RSpec/ExpectActual:
   Enabled: false
--- a/app/models/concerns/hyrax/file_set/characterization.rb
+++ b/app/models/concerns/hyrax/file_set/characterization.rb
@@ -19,7 +19,7 @@ module Hyrax
         class_attribute :characterization_terms, :characterization_proxy
         self.characterization_terms = [
           :format_label, :file_size, :height, :width, :filename, :well_formed,
-          :page_count, :file_title, :last_modified, :original_checksum, :mime_type,
+          :page_count, :file_title, :last_modified, :original_checksum,
           :duration, :sample_rate
         ]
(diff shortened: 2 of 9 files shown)


tpendragon:
For what it's worth, I made sure this works in Plum too.

no-reply:
> A. Do we like this pattern?:+1:

no-reply:
> A. Do we like this pattern?
:+1:

> B. Is the shared spec correct? For example, is #uri and #mime_type a part of the interface, or just for that specific derivative service?

It looks good to me. `#mime_type`, at least seems like something the derivative service should know. Can we think of examples where this won't be appropriate?

> Is #file_set part of the interface?

Yes. You're passing it in as an argument, so exposing it as a reader doesn't seem to add any dependencies for clients; they already have to know about it, and possibly want to get it back. The only question I have is about whether you have the right name. Is there something more abstract than `file_set` that the argument could be called?

> SHOULD we be checking if #valid? works?

Looks right to me. Are there other reasons the object could be invalid other than the fileset being so? If so, the shared example might be overly specific. But it also seems fine to let it be that way until/unless other facets of validity emerge.

> C. Should we blow up FileSetDerivativeService into one per mime type, now that it's easy? (Maybe another PR, but that's a benefit of this refactor.)

Sounds neat. :+1:.

tpendragon:
> It looks good to me. #mime_type, at least seems like something the derivative service should know. Can we think of examples where this won't be appropriate?

There are cases where it just isn't used - for instance, we store files external to Fedora, so `mime_type` never actually gets set, it's stored in a different property. If `mime_type` is part of the interface to a DerivativeService it points towards overriding that method on our service, but then you end up with different services with the same fileset reporting different mime types.

> Looks right to me. Are there other reasons the object could be invalid other than the fileset being so? If so, the shared example might be overly specific. But it also seems fine to let it be that way until/unless other facets of validity emerge.

I asked this because that spec used to test for `#invalid` cases too, but then the default service turned out to always return `true`. This would only be an issue if there were a service which was hard to craft a valid relevant fileset for.

tpendragon:
Probably didn't need a WIP for this branch. Removed that, and this is now open for proper review.

jeremyf:
I like the pattern.

jcoyne:
I'd love to see one of the commit messages expanded to include something about why this change is needed with a concrete example. The code looks :+1:

tpendragon:
@jcoyne Expanded the commit message here: https://github.com/projecthydra-labs/hyrax/pull/445/commits/b3d139200850fd1a4aed91b05f10410bcffd1472

jcoyne:
Thank you

PR #450, opened 2017-02-21
Tidying up slow specs by consolidation

jeremyf:
From the travis build prior to this update

```console
Hyrax::My::CollectionsController logged in user #index shows only collections that I own
    56.95 seconds ./spec/controllers/hyrax/my/collections_controller_spec.rb:23
```

### DIFF
--- a/spec/controllers/hyrax/my/collections_controller_spec.rb
+++ b/spec/controllers/hyrax/my/collections_controller_spec.rb
@@ -1,29 +1,23 @@
 describe Hyrax::My::CollectionsController, type: :controller do
   describe "logged in user" do
-    let(:user)  { create(:user) }
-    let(:other) { create(:user) }
-
-    let!(:my_file)              { create(:work, user: user) }
-    let!(:first_collection)     { create(:public_collection, user: user) }
-    let!(:unrelated_collection) { create(:public_collection, user: other) }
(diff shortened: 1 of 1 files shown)


spec/controllers/hyrax/my/collections_controller_spec.rb
```diff
@@ -1,29 +1,22 @@
...
-          expect(assigns[:document_list].length).to eq 2
-          get :index, params: { per_page: 2, page: 2 }
-          expect(assigns[:document_list].length).to be >= 1
-        end
-      end
+      let(:user)  { create(:user) }
+      let(:other) { create(:user) }
+      before { sign_in user }
+
+      let!(:my_file)              { create(:work, user: user) }
```
jcoyne: One of the recent recommendations that rubocop has (that we haven't upgraded to yet) is that we should not name things that we are not referencing and that they should just go into a before block.  I'd prefer we follow that guideline so that we eventually have a simpler upgrade.

jcoyne: If you *really* want to make this fast you could just test that the correct search_builder is used. Alternately, just add fixtures to solr (not to fedora) as this controller action should not be hitting fedora.

PR #459, opened 2017-02-21
Workflows should be admin set specific

### LINKED ISSUE #256: Workflows should be Admin Set-specific
The use case here is that installations (such as ours at Stanford but also @scholarworks's @vantuyls's and @cam156's?) want to use the one-step-mediated-deposit workflow in multiple Admin Sets, but we do not want workflow **roles** shared between Admin Sets.

Discussed the implementation with @jcoyne and @jeremyf on 1/25 and we think this means:

- [x] Workflows are not shared between Admin Sets -- no workflows exist outside the context of Admin Sets. This can be enforced by `PermissionTemplate has_many Workflows`; The `PermissionTemplate` should also have one active Workflow.
- [x] Each Admin Set `has_many` workflows
- [x] Each Admin Set can only have one active workflow at a given time.
- [x] Whenever an Admin Set is created, *n* workflows are imported (where *n = the number of workflow templates*) from the available JSON templates and associated with the Admin Set
~~- [ ] We may need to be careful not to allow users to change an Admin Set's active workflow if there are items moving through the current workflow. (Is this a problem already? We should verify this before entangling that with this issue.)~~

jeremyf:
## Adding migration for workflow permission template

@2c24899e9c46eb2a6413c7d30fb561c9b57c53e8

Unblocks projecthydra-labs/hyku#347
Unblocks projecthydra-labs/hyrax#228
Closes #256

## Adding PermissionTemplate.has_many :workflows

@170fed12ab64b7d3981c1f58e856ece05f76dc15

This is a concession that each AdminSet has many workflows, but given
that AdminSet is an ActiveFedora::Base and Sipity::Workflows are an
ActiveRecord::Base, we can't cross the streams. So instead, I'm
leveraging the fact that each AdminSet has one and only one Permission
Template and each Permission Template belongs to one and only one
AdminSet.

## Ensuring Workflow requires permission template

@48d6166a62d1bbea520963cc30b0f8a5a0c2b6bc

## Requiring permission_template for WorkflowImport

@7c58cfecb28e8ca2ee114cd52f59375079d08db1

## Working towards works must be in AdminSet

@555fdf97acc5dc6096ec3721465f19e21656620c

## Fixing rubocop violation

@5e7c25da0315f19270c059fcb566715c299ccbd6

Style/EmptyLines: Extra blank line detected.

## Ensuring permission template admin_set_id unique

@a81df61c6f4b411733d646ced2736c7f6211d823

* Each AdminSet has one and only one PermissionTemplate
* Each PermissionTemplate belongs to one and only one AdminSet

## Adding AdminSet.after_destroy hook

@0837016a6fc11c1494e9fd083641170a7861fdd1

Ensuring that when we destroy an admin set that we clean up the
corresponding permission template.

## Allow admin_set to be created with perm template

@bd56123468d713368ba3092e586b501a44660011

## Improving factories for AdminSet and PermissionTemplate

@68d871079a0ae3d0342248044a0c766bffa91bbc

## Removing log_and_raise as method is undefined

@b10bba1236aa404d1c94faad2f1a2ca1d18c8864

## Updating permission_template factory for workflow

@ce559c2bc1dc41088ccfa766a05795d98f729333

As part of the permission template factories, allow for workflows to
be easily loaded:

```ruby
create(:permission_template, with_workflows: true)
```

## Allow for works to create permission templates

@0d0cd5734d5163935628b23e2cf9f696d9581991

Leveraging the other work, this process allows for chaining together
the creation of several different concepts

## Adding import workflow for one PermissionTemplate

@5fde4746822d41272a9c233ec1be5ad0cc1e49c6

Given that each PermissionTemplate has many workflows that are
available for assignment, I wanted a method to contain that logic.

## Importing workflows when AdminSet is created

@5df0c542468a9e1b652393689bca940b7cb07d93

When we create an admin set, we need to load the workflows that are to
be associated with the permission_template (and by extension AdminSet).

## Skipping auto-create AdminSet for works factory

@78f341dd367425170d75f14d476127b5bef3916f

It appears to shave ~0.1 seconds off of each test that creates a work.

## Narrowing workflows for permission template

@1dcfd94e3376f47f3fe1bed886823861c9a68433

Given that we are loading all workflows for each permission template,
we don't want to query for all workflows; Instead we scope to the given
permission template.

## Fixing specs related to active workflows

@e7ae2a80a7976739ba49ab204a5e496ae31b7d0b

* Adding Sipity::Workflow.activate!
* Working on permission template to allow workflow
* Adding active_workflow for InAdminSet
* Improving workflow strategy lookup
* Testing GenericWork actor with action behavior
* Resolving default workflow behavior

Close #256

## Amendments for code review

@a266feb5ba1d69ede626f4441f15a7b538f9c346

## Renaming PermissionTemplate.available_workflows

@61f0ac1437ca8830cc1670e02c40059437fff0b6

Reflecting that these workflows are not active, but instead available
for activation.


.rubocop_todo.yml
```diff
@@ -23,6 +23,7 @@ Metrics/ClassLength:
     - 'lib/generators/hyrax/templates/catalog_controller.rb'
     - 'lib/generators/hyrax/install_generator.rb'
     - 'lib/hyrax/configuration.rb'
+    - 'app/forms/hyrax/forms/permission_template_form.rb'
```
jcoyne: Please order this alphabetically.

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -5,13 +5,23 @@ class PermissionTemplateForm
 
       self.model_class = PermissionTemplate
       self.terms = []
-      delegate :access_grants, :access_grants_attributes=, :release_date, :release_period, :visibility, :workflow_id, to: :model
+      delegate :access_grants, :access_grants_attributes=, :release_date, :release_period, :visibility, to: :model
+      delegate :workflows, :active_workflow, admin_set, to: :model
+      # def admin_set_id
```
jcoyne: Can you be a little more verbose here.

app/models/concerns/hyrax/in_admin_set.rb
```diff
@@ -5,5 +5,9 @@ module InAdminSet
     included do
       belongs_to :admin_set, predicate: ::RDF::Vocab::DC.isPartOf
     end
+
+    def active_workflow
+      Sipity::Workflow.find_active_workflow_for_admin_set_id(admin_set_id)
```
jcoyne: How about using kwargs here and doing: `find_active_workflow_for(admin_set_id: admin_set_id)`

app/models/hyrax/permission_template.rb
```diff
@@ -10,9 +10,13 @@ module Hyrax
   class PermissionTemplate < ActiveRecord::Base
     self.table_name = 'permission_templates'
 
-    has_many :access_grants, class_name: 'Hyrax::PermissionTemplateAccess'
+    has_many :access_grants, class_name: 'Hyrax::PermissionTemplateAccess', dependent: :destroy
     accepts_nested_attributes_for :access_grants, reject_if: :all_blank
 
+    has_many :workflows, class_name: 'Sipity::Workflow', dependent: :destroy
```
jcoyne: Can we give this a better name? `:workflow_options, :potential_workflows, :selectable_workflows, or :workflow_choices`?

app/models/sipity/workflow.rb
```diff
@@ -10,16 +10,40 @@ class Workflow < ActiveRecord::Base
     has_many :workflow_states, dependent: :destroy, class_name: 'Sipity::WorkflowState'
     has_many :workflow_actions, dependent: :destroy, class_name: 'Sipity::WorkflowAction'
     has_many :workflow_roles, dependent: :destroy, class_name: 'Sipity::WorkflowRole'
-    has_many :permission_templates, dependent: :restrict_with_exception, class_name: 'Hyrax::PermissionTemplate'
+    belongs_to :permission_template, class_name: 'Hyrax::PermissionTemplate', required: true
```
jcoyne: Add a comment here about this just being one of the potential workflows that could be selected for that permission_template.

app/services/hyrax/workflow/workflow_factory.rb
```diff
@@ -1,37 +1,35 @@
 module Hyrax
   module Workflow
+    # Responsible for stitching a work into it's workflow and granting permissions accordingly
```
jcoyne: Typo: `its`

spec/actors/hyrax/actors/generic_work_actor_spec.rb
```diff
@@ -19,42 +19,41 @@
...
+        subject.create(title: ['Foo Bar'], admin_set_id: admin_set.id)
       end
     end
 
     context 'valid attributes' do
       let(:visibility) { Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED }
       before do
         redlock_client_stub
-        create(:workflow_action)
+        # create(:workflow_action)
```
jcoyne: remove?

spec/forms/hyrax/forms/permission_template_form_spec.rb
```diff
@@ -168,17 +181,14 @@
...
-
       it "gives the managers workflow roles" do
         expect { subject }.to change { Sipity::WorkflowResponsibility.count }.by(4)
       end
     end
 
-    context "when a new workflow is not changed" do
+    context "when a workflow is not changed" do
       it "does nothing" do
+        subject # Need to get things setup
```
jcoyne: Change this to "Sets up the things" otherwise someone may read this as a TODO

app/models/hyrax/permission_template.rb
jeremyf: Yeah…I'll renamed them. I agree just looking at where all to find and replace.

PR #460, opened 2017-02-21
Have a single way to draw permission badges

### LINKED ISSUE #137: SufiaHelperBehavior#render_visibility_label should use PermissionBadge
<a href="https://github.com/awead"></img></a> **Issue by [awead](https://github.com/awead)**
_Thursday Jun 30, 2016 at 02:28 GMT_
_Originally opened as https://github.com/projecthydra/sufia/issues/2312_

----

CurationConcerns::PermisisonBadge contains all the logic and presenter code necessary to do what SufiaHelperBehavior#render_visibility_label is doing.

jcoyne:
Prior to this there was still the Sufia way and the CurationConcerns
way.

Fixes #137

### DIFF
--- a/app/presenters/hyrax/permission_badge.rb
+++ b/app/presenters/hyrax/permission_badge.rb
@@ -1,62 +1,43 @@
 module Hyrax
   class PermissionBadge
     include ActionView::Helpers::TagHelper
-    include HyraxHelper
 
-    def initialize(solr_document)
-      @solr_document = solr_document
+    # @param visibility_or_document [String,#visibility] the current visibility or an object
+    #                                                    that has a method returning visibility
+    def initialize(visibility_or_document)
+      self.visibility = visibility_or_document
     end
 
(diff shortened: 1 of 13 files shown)


app/presenters/hyrax/permission_badge.rb
```diff
@@ -1,62 +1,38 @@
...
   class PermissionBadge
     include ActionView::Helpers::TagHelper
-    include HyraxHelper
 
-    def initialize(solr_document)
-      @solr_document = solr_document
+    # @param visibility_or_document [String,#visibility] the current visibility or an object
+    #                                                    that has a method returning visibility
+    def initialize(visibility_or_document)
+      @visibility = if visibility_or_document.respond_to?(:visibility)
```
jeremyf: Extract if statement into a method; If someone were extend PermissionBadge it would provide a place to change.

(As a matter of practice, I prefer minimal business logic in initialize, instead preferring methods for any complex setter decisions)

PR #466, opened 2017-02-22
Keep the menu section open when a subitem is selected

### LINKED ISSUE #462: Admin sidebar item doesn't stay open when selecting a sub-item
It should look like this:

screen shot 2017-02-22 at 7 54 15 am

jcoyne:
Fixes #462

### DIFF
--- a/app/presenters/hyrax/collapsable_section_presenter.rb
+++ b/app/presenters/hyrax/collapsable_section_presenter.rb
@@ -1,0 +1,50 @@
+module Hyrax
+  # Draws a collapsable list widget using the Bootstrap 3 / Collapse.js plugin
+  class CollapsableSectionPresenter
+    def initialize(view_context:, text:, id:, icon_class:, open:)
+      @view_context = view_context
+      @text = text
+      @id = id
+      @icon_class = icon_class
+      @open = open
+    end
+
+    attr_reader :view_context, :text, :id, :icon_class, :open
+    delegate :content_tag, :safe_join, to: :view_context
+
+    def render(&block)
+      button_tag + list_tag(&block)
+    end
+
+    private
+
+      def button_tag
+        content_tag(:a,
+                    role: 'button',
+                    class: "#{button_class}collapse-toggle",
+                    data: { toggle: 'collapse' },
+                    href: "##{id}",
+                    'aria-expanded' => open,
+                    'aria-controls' => id) do
+                      safe_join([content_tag(:span, '', class: icon_class),
+                                 content_tag(:span, text)], ' ')
+                    end
+      end
+
+      def list_tag
+        content_tag(:ul,
+                    class: "collapse #{workflows_class}nav nav-pills nav-stacked",
+                    id: id) do
+                      yield
+                    end
+      end
+
+      def button_class
+        'collapsed ' unless open
+      end
+
+      def workflows_class
+        'in ' if open
+      end
+  end
+end
--- a/app/presenters/hyrax/menu_presenter.rb
+++ b/app/presenters/hyrax/menu_presenter.rb
@@ -1,0 +1,37 @@
+module Hyrax
+  # view-model for the admin menu
+  class MenuPresenter
+    def initialize(view_context)
+      @view_context = view_context
+    end
+
+    attr_reader :view_context
+
+    delegate :controller, :controller_name, :action_name, :content_tag,
+             :current_page?, :link_to, to: :view_context
+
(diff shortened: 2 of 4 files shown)


app/presenters/hyrax/menu_presenter.rb
```diff
@@ -0,0 +1,32 @@
...
+             :current_page?, :link_to, to: :view_context
+
+    # Returns true if the current controller happens to be one of the controllers that deals
+    # with workflow.  This is used to keep the parent section on the sidebar open.
+    def workflows_section?
+      controller.instance_of? Hyrax::Admin::WorkflowRolesController
+    end
+
+    def nav_link(href, **link_options)
+      content_tag(:li, class: ('active' if current_page?(href))) do
```
jeremyf: Inline terinary operators for hash values strikes me as a bit stinky. It's hard to insert a debugger to see the results.

```ruby
content_tag_options = {}
content_tag_options[:class] = 'active' if current_page?(href)
content_tag(:li, content_tag_options) do
```

Optionally, I can get behind:

```ruby
content_tag_options = current_page?(href) ? { class: 'active' } : {}
content_tag(:li, content_tag_options) do
```

app/presenters/hyrax/menu_presenter.rb
```diff
@@ -0,0 +1,32 @@
...
+        link_to(href, link_options) do
+          yield
+        end
+      end
+    end
+
+    # Draw a collaspable menu section. The passed block should contain <li> items.
+    def collapsable_section(text, **kwargs, &block)
+      CollapsableSectionPresenter.new(view_context, text, kwargs).render(&block)
```
jeremyf: Shouldn't the `kwargs` be `**kwargs`?

I believe we should fully prefer keyword args (instead of mixed).
```ruby
CollapsableSectionPresenter.new(view_context: view_context, text: text, id: id, icon_class: icon_class, open: open)
```

app/presenters/hyrax/menu_presenter.rb
```diff
@@ -0,0 +1,32 @@
...
+    def nav_link(href, **link_options)
+      content_tag(:li, class: ('active' if current_page?(href))) do
+        link_to(href, link_options) do
+          yield
+        end
+      end
+    end
+
+    # Draw a collaspable menu section. The passed block should contain <li> items.
+    def collapsable_section(text, **kwargs, &block)
```
jeremyf: Given that the underlying class relies on the kwargs of `id:, icon_class:, open:` these should be indicated on the method signature.

Also, move text to a kwarg.

app/presenters/hyrax/collapsable_section_presenter.rb
```diff
@@ -0,0 +1,50 @@
+module Hyrax
+  # Draws a collapsable list widget using the Bootstrap 3 / Collapse.js plugin
+  class CollapsableSectionPresenter
+    def initialize(view_context, text, id:, icon_class:, open:)
```
jeremyf: Make both `view_context` and `text` keywords.

Is there a reasonable default assumption for `open`

app/presenters/hyrax/menu_presenter.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    attr_reader :view_context
+
+    delegate :controller, :controller_name, :action_name, :content_tag,
+             :current_page?, :link_to, to: :view_context
+
+    # Returns true if the current controller happens to be one of the controllers that deals
+    # with workflow.  This is used to keep the parent section on the sidebar open.
+    def workflows_section?
+      controller.instance_of? Hyrax::Admin::WorkflowRolesController
```
jeremyf: Hmm…should the `Hyrax::Admin::WorkflowRolesController` be a class configuration?

jcoyne: No, I don't think so. One would override this whole method, as this condition could get much more complex (checking action_name, multiple controllers, etc.)

app/presenters/hyrax/menu_presenter.rb
jcoyne: Moving text to a kwarg would make this unlike any other view helper which passes the text as a positional argument.  For example: http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

app/presenters/hyrax/collapsable_section_presenter.rb
jcoyne: One could provide a default, but it is not needed in our case. We are always passing something.

PR #468, opened 2017-02-22
Update to the master branch of flipflop

jcoyne:
Stop using justin's fork

Thanks so much to @rolftimmermans for the updates to flipflop.

### DIFF
--- a/app/controllers/hyrax/admin/features_controller.rb
+++ b/app/controllers/hyrax/admin/features_controller.rb
@@ -2,11 +2,18 @@ module Hyrax
   module Admin
     class FeaturesController < Flipflop::FeaturesController
       layout 'admin'
+
       before_action do
         authorize! :manage, Hyrax::Feature
+      end
+
+      # overriding so we can have a layout https://github.com/voormedia/flipflop/issues/18
+      def index
         add_breadcrumb t(:'hyrax.controls.home'), root_path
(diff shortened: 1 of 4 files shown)


app/controllers/hyrax/admin/features_controller.rb
```diff
@@ -2,11 +2,18 @@ module Hyrax
...
       before_action do
         authorize! :manage, Hyrax::Feature
+      end
+
+      # overriding so we can have a layout https://github.com/voormedia/flipflop/issues/18
+      def index
         add_breadcrumb t(:'hyrax.controls.home'), root_path
         add_breadcrumb t(:'hyrax.toolbar.admin.menu'), hyrax.admin_path
         add_breadcrumb t(:'hyrax.admin.sidebar.settings'), hyrax.admin_features_path
+        @feature_set = Flipflop::FeaturesController::FeaturesPresenter.new(Flipflop::FeatureSet.current)
```
jeremyf: This looks like an addition to the code; Should this be a separate method?

jcoyne: @jeremyf I'm trying not to deviate from https://github.com/voormedia/flipflop/blob/master/app/controllers/flipflop/features_controller.rb#L8, so that eventually if https://github.com/voormedia/flipflop/issues/18 is fixed, we can call `super`.

PR #480, opened 2017-02-23
Add YAMLStrategy for Flipflop

### LINKED ISSUE #372: Support YAML as configuration for FlipFlop
For CC users who want configuration as part of the codebase (Plum users included.)

tpendragon:
This allows users to configure their application without using the admin
panel, if they so wish.

Closes #372.


tpendragon:
@jcoyne Added some text to the README.

Issue #444, opened 2017-02-17
Add left sidebar to Profile page where logged-in user manages their profile


hannahfrost:
Per https://github.com/projecthydra-labs/hyrax/issues/441

![profile](https://cloud.githubusercontent.com/assets/1801076/23085469/d6aa1f6a-f51c-11e6-81eb-c468d49cf2ce.png)

Issue #443, opened 2017-02-17
Add left sidebar to Works page where logged-in user manages their work(s)


hannahfrost:
Per https://github.com/projecthydra-labs/hyrax/issues/441

![works](https://cloud.githubusercontent.com/assets/1801076/23085378/49751820-f51c-11e6-88d5-751a999a2d7c.png)

jcoyne:
@ggeisler as this user is not an admin, should the breadcrumbs say "Administration" here?

hannahfrost:
@jcoyne I discussed a similar question with @ggeisler recently. The answer is yes: "Administration" in this context is about function, not meant to be used in the UI only for explicitly admin (or manager) roles. Logged-in users (eg Depositors) who are not Repository Administrators have tasks that are administrative and content management-y in nature, and this is the best overarching term we could come up with for the navigation. Note that #441 describes how "Administration" is the term used in the upper right menu for a user to navigate to this page. Do you have suggestions for a different term to use for this section of the site?

Issue #442, opened 2017-02-17
Add left sidebar to Collections page where logged-in user manages their collection(s)


hannahfrost:
Per https://github.com/projecthydra-labs/hyrax/issues/441

![collections](https://cloud.githubusercontent.com/assets/1801076/23085317/e5ec80a4-f51b-11e6-8607-6425dfb64a60.png)

vantuyls:
I believe we can close this?

mjgiarlo:
@vantuyls You mean because the CE work cycle is covering it, or what? Pulling in @elrayle to confirm this or something like it is already in-scope for that work. Also pinging @hannahfrost and @ggeisler to loop 'em in.

ggeisler:
Yes, I believe this is now covered by #1417.

mjgiarlo:
Thanks, @ggeisler @vantuyls

Issue #441, opened 2017-02-17
Add left sidebar to logged-in user pages


hannahfrost:
To build on the UX introduced with the Hyrax admin dashboard, the left sidebar should be added to all of the pages that a logged-in user uses for administrative and content management tasks.

The attached mockups show how this sidebar looks for the roles that do not have administrative privileges for the repository or for admin sets: registered user, admin set depositor, admin set reviewer. (Other parts of the mockups are obviously skeletal, and not intended as a focus for this ticket.) 

With these functions moved to the left sidebar, the top menu bar is impacted. The “Dashboard”, “Works” and “Collections” menus are no longer presented, as the pages they encompass are now accessed from the left sidebar when the user is in the administration context. "Administration" is accessed on a menu revealed by clicking on the user name in the upper right. Clicking on “Administration” takes the user to the "Your Activity" Screen. "Notifications" link remains in the top bar to the left of the user name.

This is an overarching explanatory ticket. Individual tickets are linked for each of the user pages to get the left sidebar. 

![activitysummary](https://cloud.githubusercontent.com/assets/1801076/23084971/db9bf294-f519-11e6-9036-d50025a77e00.png)

hannahfrost:
Still need to add a ticket about making the left sidebar collapsible, as shown in https://github.com/projecthydra-labs/hyku/issues/447

vantuyls:
can we close this?

mjgiarlo:
I think so. @hannahfrost?

PR #482, opened 2017-02-24
Consolidated user and admin dashboard

jcoyne:
This is an epic ticket.
Ref #184 
Ref #362 
Ref projecthydra-labs/hyku#502

Mockups at projecthydra-labs/hyku#447

TODO: 
- [x] #442 
- [x] #443 
- [x] #444 
- [x] #441 
- [x] Add create buttons to work / collection pages
- [x] Add the highlighted column on "my works" page
- [x] Move "Filtering by" under the filters.
- [x] Create works should use the new layout
- [x] notification as a bell icon.
- [x] notifications page should be on the dashboard
- [x] remove dropdown on the search.

Screencaps:
screen shot 2017-03-02 at 12 15 31 pm


hannahfrost:
@jcoyne I'm so happy to see this ticket, and would be even happier to pull all the subtasks tickets to the Ready column if you agree that they are in fact ready.

mjgiarlo:
Heads-up to the @projecthydra-labs/hyrax-ui-ux-advisors... this is a UI change that we've been moving towards for some time now, after discovering:

1. The "topbar" is becoming too crowded. Knowing what we know about the current crop of supported features and what the community wants in the coming year, we need more screen real estate and more flexibility (e.g., submenus and subsubmenus) to contain all the functionality.
1. Drift between the two dashboards: the classic Sufia dashboard and the new administrative dashboard. Given past discussions surfaced doubt about the elements of the classic dashboard being the right ones going forward, and the existence of fuller designs for the admin dashboard, this PR consolidates the two dashboards. Now all logged-in users get the leftbar-style dashboard.

See the attached screenshot above, and to see more elements of the design and the rationale, see
the https://github.com/projecthydra-labs/hyku/issues/447

What do y'all think of the consolidated design (to drop in Hyrax *after* the 1.0.0 release so as not to complicate the upgrade path).

config/locales/hyrax.en.yml
```diff
@@ -11,6 +11,13 @@ en:
           based_near: Location
           contributor: Contributors
           keyword: Keyword
+        facet:
```
mjgiarlo: Can you run the yaml sorter on both these files?

mjgiarlo: See https://github.com/projecthydra-labs/hyrax/pull/523

mjgiarlo:
Should the "Administration" breadcrumb now be "Dashboard"?

mjgiarlo:
Should the user profile show page live outside the dashboard? I expect that page to be visible by all including non-authenticated users.

hackmastera:
TL;DR, We vote to remove the image from the sidebar menu, and move configuration to the bottom. 


My initial reaction is that it really privileges the user avatar image, which we don't use. I'd expect / prefer to see that on a profile page rather than a menu.

I had my heaviest admin user look at this and she thinks it seems fine. Here were her comments / questions looking at the screenshot:
* i'd get most use out of the repository objects section, but i don't really mind that it's under activity that much (although the profile/activity stuff also shows up under right top corner profile name too)
* generally, i'm also used to settings/configuration being at the very end. but maybe i wouldn't even be seeing that, who knows
* What is the default page when you log in?

Other comments on the page displayed, which I think are out of scope for this ticket but which someone may find interesting:
* i'd prefer add new work to be to the left of create batch, oddly enough, probably because of the way i think through our workflow
* the filter is interesting, it seems like a mashup right now of workflow status, admin stuff, and facets?

hannahfrost:
@hackmastera This is good feedback. Thank you. @ggeisler, what do you think?

I agree that we can do without the avatar image on the dashboard.

Configuration could move to the bottom of the menu. It will indeed only be available to the repo admin, but I wouldn't expect it to be an area that is accessed frequently.

Default page on login is Activity Summary for Repository Admin, My Summary for less-privileged users.

jcoyne:
@mjgiarlo the profile is on the admin page so the current user can edit it. There's also a public view of the profile (the user show page)

mjgiarlo:
When I create a collection, I am redirected to the collection show page (outside the dashboard), but my breadcrumbs are: `My Dashboard > Your Collections > TitleOfCollection`

I hadn't seen `My Dashboard` previously (see above re: Administration vs. Dashboard), and we should probably standardize on a pronoun (my vs. your) or drop them altogether.

mjgiarlo:
Create button on my/collections is 'New Collection' and create button on my/works is 'Add new work'. Suggest being consistent re: whether we use a verb or not.

mjgiarlo:
Also :+1: to moving "Add new work" to the left of "Create batch of works" on my/works view.

mjgiarlo:
Batch select dropdown/checkbox is available in my/works but not in dashboard/works (all works).

mjgiarlo:
@jcoyne :speech_balloon: 
> @mjgiarlo the profile is on the admin page so the current user can edit it. There's also a public view of the profile (the user show page)

Ah, thanks. I was surprised how hard it was to find links in the UI to this page. I had to click into the work show view, then the fileset show view, then on the depositor value. Where else do we expose this in the UI? (This is where I wonder if our work show view should be showing the depositor, even though that's kinda outside the scope of this PR.)

mjgiarlo:
In https://github.com/projecthydra-labs/hyku/issues/447, the leftbar is all one color. In the implementation in this PR, the block behind the avatar is a darker grey. Also, the section headings (the h5 elements like Activity, Configuration, and Tasks) are low contrast and somewhat hard to read. (We may already have a ticket for this?)

mjgiarlo:
When I click the link for Activity > Statistics (/admin/stats), I see:

```
NoMethodError at /admin/stats
undefined method `profile_path' for #<ActionDispatch::Routing::RoutesProxy:0x0000000166e378>
```

jcoyne:
@mjgiarlo I pushed a change to the view to address that routing error.

Regarding the styles, this is not complete.  I was hoping that we could follow this up with some polishing of the styles.

You are correct that batch select was not present in the all works. I wasn't sure of the utility of that for works I didn't own.  There's no reason we couldn't add that.

Re: Administration vs. Dashboard, I'd alread asked @hannahfrost and @ggeisler about this and they said that "Administration" was the activity.

As there is so much feedback here, would you mind creating a ticket for everything? I'm sure many of them need discussion before they are actionable.

hackmastera:
Thanks for attending to my/our comments. would love to see a new screenshot.

mjgiarlo:
@hackmastera Sorry, which changes that you suggested did you want to see reflected in a new screenshot?

mjgiarlo:
Hey @hackmastera, with LDCX coming up so soon and bearing in mind the cost of needing to keep rebasing this branch, I'm inclined to :white_check_mark: and `:merge:` this pull request. I would like to work with you to make sure any changes get slotted in before any 2.x releases are cut. Hit me up and I will put in issues.

PR #488, opened 2017-02-24
Don't try to remove that which is not there

jeremyf:


mbklein:
Most zen commit message evar.

jeremyf:
In the dark days of a prior job, all of my maintenance messages (pre-source control commit messages) were written in Haiku form. It was the little things.

PR #490, opened 2017-02-24
Fix admin set visibility radio buttons

laritakr:
- Resolves  https://github.com/projecthydra-labs/hyku/issues/744
- Resolves https://github.com/projecthydra-labs/hyrax/issues/463
- Specs now test for valid release option combinations, and return appropriate notifications if the selections aren't correct. 
- Radio buttons are now checked based on currently saved options.

Related issue https://github.com/projecthydra-labs/hyrax/pull/476 allowed the Visibility options to be correctly checked, but the Release Varies option needed an explicit statement to correctly know when to show the radio button as checked. 
`<%= f.radio_button :release_period, '', checked: !f.object.release_varies.blank? %>`

This is required because the button is based on simply a value of blank (because release period can contain either "before" or "embargo"), so the determining factor as to whether to check it relies on the existence of one of the underlying release_varies sub-options being selected. This can be removed if https://github.com/projecthydra-labs/hyrax/issues/489 is completed to persist all of the permission template data. 

I was unable to create a failing test for this radio button issue. Future clean-up of the admin set & permission templates should hopefully allow a simpler structure that can be validated. 

I have documented the current logic regarding the content of each of the visibility release attributes,  as well as which combinations throw which errors in the following table. (One correction to the table is that for "no delay", the date gets set to today's date after update).

screen shot 2017-02-24 at 5 10 26 pm

@jeremyf 

### DIFF
--- a/app/forms/hyrax/forms/permission_template_form.rb
+++ b/app/forms/hyrax/forms/permission_template_form.rb
@@ -38,13 +38,50 @@ module Hyrax
         select_release_varies_option(model)
       end
 
+      # @return [Hash{Symbol => String, Boolean}] { :content_tab (for confirmation message), :updated (true/false), :error_code (for flash error lookup) }
       def update(attributes)
-        update_admin_set(attributes)
-        update_permission_template(attributes)
-        grant_workflow_roles(attributes)
+        return_info = { content_tab: tab_to_update(attributes) }
+        error_code = nil
+        case return_info[:content_tab]
+        when "participants"
+          update_participants_options(attributes)
+        when "visibility"
+          error_code = update_visibility_options(attributes)
+        when "workflow"
+          update_workflow_options(attributes)
+        end
+        return_info[:error_code] = error_code if error_code
+        return_info[:updated] = error_code ? false : true
+        return_info
       end
 
       private
+
+        # @return [String]
+        def tab_to_update(attributes)
+          return "participants" if attributes[:access_grants_attributes].present?
+          return "workflow" if attributes[:workflow_id].present?
+          return "visibility" if attributes.key?(:visibility)
+        end
+
+        # @return [Void]
+        def update_participants_options(attributes)
+          update_admin_set(attributes)
+          update_permission_template(attributes)
+        end
+
+        # @return [String, Nil] error_code if validation fails, nil otherwise
+        def update_visibility_options(attributes)
+          error_code = validate_visibility_combinations(attributes)
+          return error_code if error_code
+          update_permission_template(attributes)
+        end
+
+        # @return [Void]
+        def update_workflow_options(attributes)
+          update_permission_template(attributes)
+          grant_workflow_roles(attributes)
+        end
 
         def activate_workflow_from(attributes)
           new_active_workflow_id = attributes[:workflow_id] || attributes['workflow_id']
@@ -86,8 +123,10 @@ module Hyrax
           end
         end
 
+        # @return [Nil]
         def update_permission_template(attributes)
           model.update(permission_template_update_params(attributes))
+          nil
         end
 
         # Maps the raw form attributes into a hash useful for updating the admin set.
@@ -103,7 +142,6 @@ module Hyrax
         def grants_as_collection(attributes)
           return [] unless attributes[:access_grants_attributes]
           attributes_collection = attributes[:access_grants_attributes]
-
           if attributes_collection.respond_to?(:permitted?)
             attributes_collection = attributes_collection.to_h
           end
@@ -131,10 +169,11 @@ module Hyrax
           end
         end
 
+        # Handles complex defaults
+        # Removes release_varies and release_embargo from the returned attributes
+        # These form fields are only used to update release_period
         # @return [Hash] attributes used to update the model
         def permission_template_update_params(raw_attributes)
-          # Remove release_varies and release_embargo from attributes
-          # These form fields are only used to update release_period
           attributes = raw_attributes.except(:release_varies, :release_embargo)
           # If 'varies' before date option selected, then set release_period='before' and save release_date as-is
           if raw_attributes[:release_varies] == Hyrax::PermissionTemplate::RELEASE_TEXT_VALUE_BEFORE_DATE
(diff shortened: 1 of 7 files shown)


app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -38,14 +38,61 @@ def initialize(model)
...
+          if validated_attributes[:valid] == true
+            update_permission_template(attributes)
+          else
+            return_info[:error_code] = validated_attributes[:error_code]
+          end
+          return_info[:updated] = validated_attributes[:valid]
+          return_info
+        end
+
+        # @return [Hash] { :updated } = true
```
atz: Seems like an odd return value, since it is never informative to the caller.  Might as well return null or whatever `grant_workflow_roles` returns, right?  The `{}` are not conventional in a YARD description, but that aside, I think you want `true` inside them (matching to the other `@return` statements you have).

spec/forms/hyrax/forms/permission_template_form_spec.rb
```diff
@@ -195,6 +220,95 @@
...
+
+    context "validate all release option attribute combinations" do
+      def self.expect_attributes_to_be_valid(attributes = {})
+        it "will be valid? for #{attributes.inspect}" do
+          attributes = ActionController::Parameters.new(**attributes).permit!
+          expect(form.update(attributes)).to eq(content_tab: "visibility", updated: true)
+        end
+      end
+
+      def self.expect_attributes_not_to_be_valid(attributes = {})
```
atz: This is an undesirable spot to define a method.  It looks like it is scoped to this context block, but in execution it is not.  Can we rework these using rspec `shared_examples`?  I can propose a revision or assist if that is unfamiliar.

jeremyf:
I disagree with extracting a shared example. These methods are to create a
more legible set of tested scenarios below.

The return value of the form#update method is atypical, but this form and
the underlying PermissionTemplate model has some complicating data behavior.

LaRita has a spreadsheet of the complicated data validations. There are
ways to refactor this to be a better form and data model, but that is for
another implementor.

Jeremy Friesen

atz:
What do you think `shared_examples` does, if not that?  The structure and position in the file could be the same without the scope and contamination issues.  Alternatively, this would be more conventional as a helper method.  

But `self` in that context is a `RSpec::ExampleGroups::...` object: defining methods directly on it is anti-pattern.  RSpec provides tools for this case: why not use them?

jeremyf:
@atz Yes. A shared example defined inside that block would be reasonable. I'm opposed to moving the implementation of the it blocks outside of this file.

app/forms/hyrax/forms/permission_template_form.rb
laritakr: Not sure if I understand your question here or not. The return value here is based on whether or not the validation passes. If it doesn't pass, we do no updates. If it passes, we do.  So the results of the validation are what we return (both true/false, as well as an error code to be used for the flash notice in the re-routing).  The other return statements ONLY return true; Visibility tab changes are the only ones that may return without updating.

spec/forms/hyrax/forms/permission_template_form_spec.rb
laritakr: Yes I will need assistance with this. I sent you a message on slack.

config/locales/hyrax.en.yml
```diff
@@ -52,8 +52,14 @@ en:
           permission_update_notices:
             new_admin_set:      "The administrative set '%{name}' has been created. Use the additional tabs to define other aspects of the administrative set."
             participants:       "The administrative set's participant rights have been updated"
-            visibility:         "The administrative set's visibility settings have been updated."
+            visibility:         "The administrative set's release & visibilty settings have been updated."
```
atz: Typo here: `visibilty` should be `visibility`

spec/forms/hyrax/forms/permission_template_form_spec.rb
```diff
@@ -195,6 +220,95 @@
...
+      # fixed, with date selected
+      expect_attributes_to_be_valid(
+        visibility: '',
+        release_date: Time.zone.today + 2.months,
+        release_period: Hyrax::PermissionTemplate::RELEASE_TEXT_VALUE_FIXED,
+        release_varies: nil,
+        release_embargo: nil
+      )
+      # varies, but no subsequent options
+      expect_attributes_not_to_be_valid(
```
atz: This case is the same as `varies, with embargo option but no period` at line 294, i.e. all empty string/nil values!

app/forms/hyrax/forms/permission_template_form.rb
atz: My comment is about the Hash returned by `update_workflow_options`, described in this line.  I think you are discussing `update_visibility_options`.  

The only thing `update_workflow_options` can return is `{:updated => true}`, which makes it semantically the same as returning nil.  If it didn't throw exception, it updated.

spec/forms/hyrax/forms/permission_template_form_spec.rb
laritakr: You're right... I apparently typed over something as i was reformatting for Rubocop. Fixed this.

config/locales/hyrax.en.yml
laritakr: Fixed it.

app/forms/hyrax/forms/permission_template_form.rb
laritakr: Yes you're right. I wasn't looking at the same line you referred to (update_participant_options does the same thing by the way).

But unless I also make changes in the controller, I do need the value of true (or false in case of failure) returned to the controller. So for consistency and clarity purposes, I mimicked the same return structure in each of the three update methods, and then merged the resulting hash to return to the controller. I'm not sure what you'd recommend that would be any more clear, but I'm open to suggestions.

atz: Since you are defining the method here, you have the luxury of knowing that you are the only one calling it!  So you could handle that at line 51 above.  I'll comment there.

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -38,14 +38,61 @@ def initialize(model)
...
-        grant_workflow_roles(attributes)
+        return_info = { content_tab: tab_to_update(attributes) }
+        update_status = {}
+        case return_info[:content_tab]
+        when "participants"
+          update_status = update_participants_options(attributes)
+        when "visibility"
+          update_status = update_visibility_options(attributes)
+        when "workflow"
+          update_status = update_workflow_options(attributes)
```
atz: If you change the return value from `update_workflow_options`, this block could become 

```ruby
update_workflow_options(attributes)
update_status = { updated: true }
```

Similarly for `update_participants_options`.  

Of the 3 cases, only `update_visibility_options` returns a dynamic response in its Hash, and that is for only two return values.  Even there, the presence of `:error_code` implies `:updated = false`.  So it could just return error_code value or nil.

jcoyne:
@laritakr oh conflicts! Please rebase.

jeremyf:
@laritakr I went ahead and rebased; You're tackling other things.

laritakr:
This is still under work and not ready to merge in yet.

laritakr:
Ready for review @projecthydra-labs/hyrax-code-reviewers

atz:
I'm seeing duplication of commits now. e.g.:
- "Convert ad hoc methods to shared_examples" `1bde732`
- "Convert ad hoc methods to shared_examples" `f7c331b`

@jeremyf, maybe you rebased off an outdated version of the branch?  Even the main "	Fix admin set visibility radio buttons" commit is duplicated.  The end result looks good though.

jeremyf:
@atx we'll work on a squash; Hopefully this works out.

PR #504, opened 2017-02-27
Add config option to show/suppress deposit agreement

### LINKED ISSUE #435: Add ability to disable deposit agreement functionality entirely.
Our tool is internal, and much of the content that gets ingested is public domain, making this a confusing thing to have on the form.

escowles:
Fixes #435 

Adds a configuration option to completely disable the deposit agreement.

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -319,6 +319,12 @@ module Hyrax
       @active_deposit_agreement_acceptance
     end
 
+    attr_writer :show_deposit_agreement
+    def show_deposit_agreement?
+      return true if @show_deposit_agreement.nil?
+      @show_deposit_agreement
+    end
+
     attr_writer :work_requires_files
     def work_requires_files
       return true if @work_requires_files.nil?
--- a/app/views/hyrax/base/_form_progress.html.erb
+++ b/app/views/hyrax/base/_form_progress.html.erb
@@ -25,7 +25,8 @@
     <% end %>
   </div>
   <div class="panel-footer text-center">
-    <% if Hyrax.config.active_deposit_agreement_acceptance %>
+    <% if Hyrax.config.show_deposit_agreement? %>
+      <% if Hyrax.config.active_deposit_agreement_acceptance %>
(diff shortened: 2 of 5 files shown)


lib/hyrax/configuration.rb
```diff
@@ -319,6 +319,11 @@ def active_deposit_agreement_acceptance
       @active_deposit_agreement_acceptance
     end
 
+    attr_writer :show_deposit_agreement
+    def show_deposit_agreement
+      @show_deposit_agreement || true
```
mjgiarlo: If `@show_deposit_agreement` is `false`, won't this method return `true`?

escowles: Dang, you're right — update coming shortly.

lib/hyrax/configuration.rb
```diff
@@ -319,6 +319,12 @@ def active_deposit_agreement_acceptance
       @active_deposit_agreement_acceptance
     end
 
+    attr_writer :show_deposit_agreement
+    def show_deposit_agreement
```
jcoyne: Please change this to `show_deposit_agreement?`

escowles: There are a bunch of existing boolean options `enable_ffmpeg`, `display_microdata`, etc., and none of them have question marks now.  Should all of them have question marks?

jcoyne: Yes, imho. But you don't have to handle that if you don't want to.

jcoyne:
Would this be better for flipflop?

lib/hyrax/configuration.rb
escowles: OK, I've updated to `show_deposit_agreement?` and created #508 for changing the rest of the boolean getter methods to also end with question marks.

jcoyne:
@escowles don't you have the ability to set the flipflop defaults via a yaml file?  I can see that this would be something that someone who uses a Hyku product (no access to change configuration code) would want.

escowles:
Closing in favor of #510

PR #507, opened 2017-02-28
744 cleanup

atz:
See if this helps simplify some of those internal methods.  Tried to reduce the complexity of `validate_visibility_combinations` and had some success, but not enough to please rubocop.

I also did some consolidation in the spec.


atz:
@laritakr: how does this look?

PR #511, opened 2017-02-28
Admin Set hint text

laritakr:
Add hint text in both English and Spanish to display on the
add and edit pages for Admin Sets. Also add Title to the default labels
list in order to allow for Spanish translation.

Resolves issue https://github.com/projecthydra-labs/hyrax/issues/201

### DIFF
--- a/config/locales/hyrax.en.yml
+++ b/config/locales/hyrax.en.yml
@@ -723,6 +723,7 @@ en:
         admin_set_id:              "Add as member of administrative set"
         based_near:                "Location"
         collection_ids:            "Add as member of collection"
+        creator:                   "Creator"
         date_created:              "Date Created"
         description:               "Abstract or Summary"
         embargo_release_date:      'until'
@@ -730,6 +731,8 @@ en:
         keyword:                   "Keyword"
         lease_expiration_date:     'until'
         related_url:               "Related URL"
+        rights:                    "Rights"
+        title:                     "Title"
         visibility_after_embargo:  'then open it up to'
         visibility_after_lease:    'then restrict it to'
         visibility_during_embargo: 'Restricted to'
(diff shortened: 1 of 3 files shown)


mjgiarlo:
@laritakr What would you think of whipping up a quick view spec to ensure that these labels are rendered in the right place?

config/locales/hyrax.en.yml
```diff
@@ -727,6 +730,7 @@ en:
         keyword:                   "Keyword"
         lease_expiration_date:     'until'
         related_url:               "Related URL"
+        title:                     "Title"
```
jcoyne: Why are you adding this to defaults rather than admin_set?  I would think it would only make sense to be in defaults if there is more than one type that is using it.

laritakr: "Title" is the default value for title. I added it because I would assume that anywhere using the default would want it to be translated, and this allows for that if it IS used elsewhere. I'm not changing the text that would be displayed for English, which is why I put it into default. It only made sense to me to put it into admin_set if we wanted it to not use the default value of "Title".

jcoyne: Let's say admin set stops using the title field altogether.  At this point I know that I can remove the translation for `admin_set.title`, however, if it's in `default.title`, I don't know that it can be safely removed.  So putting it somewhere generic, like default, prevents us from doing clean up down the road.

laritakr: That's fine, I can move it if you prefer.  I just thought it would be better to catch all of the untranslated titles at once. And additionally, since title hints were under default, I thought it made sense to put title under default as well for consistency purposes.

laritakr:
Ok I added a test for the title hint for @mjgiarlo, and separated the commit containing the missing simple-form entries to identify those changes as a separate issue as requested by @jcoyne.

PR #518, opened 2017-03-01
Merge latest commits from CurationConcerns

mjgiarlo:


.rubocop_todo.yml
```diff
@@ -32,6 +32,7 @@ Metrics/ParameterLists:
 
 Metrics/AbcSize:
   Exclude:
+    - 'app/actors/hyrax/actors/add_to_work_actor.rb'
```
jcoyne: 😞

app/actors/hyrax/actors/add_to_work_actor.rb
```diff
@@ -22,13 +22,17 @@ def add_to_works(new_work_ids)
             work.save
           end
 
-          # add to new
+          # add to new so long as the depositor for the parent and child matches, otherwise inject an error
           (new_work_ids - curation_concern.in_works_ids).each do |work_id|
             work = ::ActiveFedora::Base.find(work_id)
-            work.ordered_members << curation_concern
-            work.save
+            if work.depositor != curation_concern.depositor
```
jcoyne: wasn't this updated to use CanCan in CC?

mjgiarlo: It was, and it was merged about ten minutes after I made this PR. I'll try including that commit in this branch but if it gets gnarly, I will be inclined to put in a follow-up PR instead.

.rubocop_todo.yml
mjgiarlo: I'll take another look at this.

mjgiarlo:
@jcoyne All changes made.

mjgiarlo:
@jcoyne re-review?

PR #519, opened 2017-03-01
Adding logging for workflow importing

### LINKED ISSUE #342: rake curation_concerns:workflow:load should have some output
This task should tell you what it loaded. For example, if you grab a default file with wget and fail to change the extension from .erb, the task says nothing. If you fix it, and the task successfully loads the .json file, it still says nothing.

I would prefer to see success messages for each load, or a message that no valid .json files were found in config/workflows/

jeremyf:
When run via the rake task, we use the STDOUT logger. Otherwise, default
to Rails.logger

Closes #342


hackmastera:
Awesome; thanks @jeremyf!!

PR #526, opened 2017-03-02
Merge commits from Sufia 7.3.0.rc2

mjgiarlo:


app/services/hyrax/workflow/complete_notification.rb
```diff
@@ -0,0 +1,7 @@
+module Hyrax
+  module Workflow
+    class CompleteNotification < DepositedNotification
```
jcoyne: Do we want to carry this forward?

mjgiarlo: Oops, no, we don't want any trace of this in Hyrax. Removing.

mjgiarlo:
@jcoyne OK, it's ready now.

PR #531, opened 2017-03-02
google_analytics_id is not a boolean

jcoyne:
Revert part of f6fe82f948733db3f0887cf87b175b41b5fdfa65


jeremyf:
@jcoyne I also caught this in https://github.com/projecthydra-labs/hyrax/pull/532

PR #544, opened 2017-03-03
Using fully-qualified Hyrax::Workflow::PermissionQuery in submodules

escowles:


jcoyne:
Can you say why you made this change (ideally in the commit message?)

escowles:
@jcoyne I was getting `Error: uninitialized constant Workflow::PermissionQuery` when my works called WorkIndexer.generate_solr_document.  I can add that to the commit message.

PR #548, opened 2017-03-07
Adding a migration to remove :workflow_id

jeremyf:
In the previous migration, due to an exception being raised, the
remove_column was not happening for the PermissionTemplate

`Hyrax::PermissionTemplate.each` raises an exception

@projecthydra-labs/hyrax-code-reviewers

PR #549, opened 2017-03-07
853 fix missing attribute

laritakr:
Remove obsolete update to permission template

Resolves projecthydra-labs/hyku#853

Work on permission template visibility options conflicted with work on workflows, and resulted in an obsolete update remaining in the code. 

Now that the workflow migration patch (#548) has been merged, there is a way to create a failing test. Prior to this migration, even the existence of this spec would not have caused a failing test because the issue was caused by an incompatibility between the permission template in Hyrax vs the one in Hyku (which had gotten migrated). This test was most likely another victim of the merge conflict which caused the problematic code to remain in the first place. But regardless, the issue is now resolved.

### DIFF
--- a/app/forms/hyrax/forms/permission_template_form.rb
+++ b/app/forms/hyrax/forms/permission_template_form.rb
@@ -48,7 +48,7 @@ module Hyrax
         when "visibility"
           error_code = update_visibility_options(attributes)
         when "workflow"
-          update_workflow_options(attributes)
+          grant_workflow_roles(attributes)
         end
         return_info[:error_code] = error_code if error_code
         return_info[:updated] = error_code ? false : true
@@ -77,12 +77,6 @@ module Hyrax
           update_permission_template(attributes)
         end
 
-        # @return [Void]
-        def update_workflow_options(attributes)
-          update_permission_template(attributes)
-          grant_workflow_roles(attributes)
-        end
-
         def activate_workflow_from(attributes)
           new_active_workflow_id = attributes[:workflow_id] || attributes['workflow_id']
           if active_workflow
@@ -95,6 +89,7 @@ module Hyrax
 
         # If the workflow has been changed, ensure that all the AdminSet managers
         # have all the roles for the new workflow
+        # @return [Void]
         # @todo Instead of granting the manage users all of the roles (which means lots of emails), can we agree on a Managing role that all workflows should have?
         def grant_workflow_roles(attributes)
           new_active_workflow = activate_workflow_from(attributes)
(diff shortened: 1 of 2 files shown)


laritakr:
@jcoyne done

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -79,7 +79,6 @@ def update_visibility_options(attributes)
 
         # @return [Void]
         def update_workflow_options(attributes)
-          update_permission_template(attributes)
```
jcoyne: Do we need to have two methods or can `grant_workflow_roles` and `update_workflow_options` be combined into one method?

laritakr:
@jcoyne done again.

spec/forms/hyrax/forms/permission_template_form_spec.rb
```diff
@@ -195,12 +195,14 @@
 
     context "when a new workflow has been chosen" do
       it "gives the managers workflow roles" do
+        expect(form).not_to receive(:update_permission_template)
```
jcoyne: I don't see these as being particularly useful assertions.  There are an infinite variety of messages it is not receiving, why pick this one?

jeremyf: @jcoyne This was to fulfill a previous request of yours:

> Is there a test that could be written that would fail before this change and would succeed after this change?

jcoyne: Yes, but what I meant by that was there a set of inputs we could use to trigger a specific bug, that is fixed by our changes to the code?  Not "does it follow a specific code path".

laritakr: I agree. I thought it was a rather worthless test to add, which was why I didn't add it initially. But since you requested it, it's there now and I'm not particularly inclined to remove it again. I think this one-line fix has taken more of my time than it was worth and there are more important things to work on.

laritakr: The line of code was a leftover from an error during merging two pull requests. It has nothing to do with any particular input.

laritakr: Now that the workflow migration patch (https://github.com/projecthydra-labs/hyrax/pull/548) has been merged, there is a way to create a failing test. Prior to this migration, this spec would not have caused a failing test because the issue was caused by an incompatibility between the permission template in Hyrax vs the one in Hyku (which had gotten migrated).  This test was most likely another victim of the merge conflict which caused the problematic code to remain in the first place. But regardless, the issue is now resolved.

PR #555, opened 2017-03-10
554 edit admin set redirect

laritakr:
When an admin set title or description is updated, the redirect now sends an update message and remains on the same tab of the form.

Fixes https://github.com/projecthydra-labs/hyrax/issues/554

Also adds a translation for the permission template update message when a participant's rights are removed.


jcoyne:
Can you squash these into one commit?  You are testing for the translated value in the first commit and changing the translated value in the controller on the second commit, so two commits is a little awkward.

laritakr:
@jcoyne done. :)

PR #561, opened 2017-03-15
Change "Rights" to "License"

jcoyne:
This just affects the labeling and the property name. The underlying
predicates are unchanged so there is no data migration. However, a solr
field changed, so reindexing all works, collections and FileSets is required.

This will make way for the "Rights Statement" which is different than
the license.

Ref https://github.com/projecthydra-labs/hyku/issues/293

### DIFF
--- a/app/presenters/hyrax/file_set_presenter.rb
+++ b/app/presenters/hyrax/file_set_presenter.rb
@@ -25,7 +25,7 @@ module Hyrax
 
     # Metadata Methods
     delegate :title, :label, :description, :creator, :contributor, :subject,
-             :publisher, :language, :date_uploaded, :rights,
+             :publisher, :language, :date_uploaded,
              :embargo_release_date, :lease_expiration_date,
              :depositor, :keyword, :title_or_label, :keyword,
              :date_created, :date_modified, :itemtype,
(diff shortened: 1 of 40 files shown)


app/presenters/hyrax/file_set_presenter.rb
```diff
@@ -25,7 +25,7 @@ def initialize(solr_document, current_ability, request = nil)
 
     # Metadata Methods
     delegate :title, :label, :description, :creator, :contributor, :subject,
-             :publisher, :language, :date_uploaded, :rights,
+             :publisher, :language, :date_uploaded,
```
mjgiarlo: `:rights` was delegated but `:license` doesn't need to be?

jcoyne: It was delegated, but then the delegate was overridden below, so it was never necessary to delegate it.

hannahfrost:
:+1:

mcritchlow:
@jcoyne @mjgiarlo so is the referenced hyku ticket in the description the plan for the final implementation? I ask because we've been having this conversation about rights vs copyright (vs access control) at UCSD and want to try and line up w/ hyrax, ideally. I think the proposed change will map well as described, just wanted to check.

mjgiarlo:
@mcritchlow Yep, it is.

jcoyne:
@mjgiarlo unblock?

mjgiarlo:
@jcoyne :white_check_mark:

PR #571, opened 2017-03-16
Provide a way for the user to change the default colors.

jcoyne:
screen shot 2017-03-16 at 1 40 43 pm

### DIFF
--- a/app/controllers/hyrax/admin/appearances_controller.rb
+++ b/app/controllers/hyrax/admin/appearances_controller.rb
@@ -1,0 +1,33 @@
+module Hyrax
+  module Admin
+    class AppearancesController < ApplicationController
+      before_action :require_permissions
+      layout 'dashboard'
+
+      def show
+        @form = form_class.new
+      end
+
+      def update
+        form_class.new(update_params).update!
+        redirect_to({ action: :show }, notice: t('.flash.success'))
+      end
+
+      private
+
+        def update_params
+          params.require(:admin_appearance).permit(:header_background_color,
+                                                   :header_text_color,
+                                                   :primary_button_background_color)
+        end
+
+        def form_class
+          Hyrax::Forms::Admin::Appearance
+        end
+
+        def require_permissions
+          authorize! :update, :appearance
+        end
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


ggeisler:
@mjgiarlo approves but:

* How does the user specify a color?
* Displaying the selected colors in a full-width controls seems unnecessary and rather garish (even if the selected colors were more reasonable)
* Showing the hex color code of the selected color, in addition to the visual representation of the color, would make it more clear to the user exactly what the selected color is, which could be important if the user is matching known colors (e.g. an institution's branding guidelines or style guide)
* The left-aligned save button doesn't match the right-aligned save buttons used elsewhere in the Admin UI, at least in the Hyku version of the Admin UI (though save button placement is already inconsistent and will need to be reconciled in a ticket at some point, so maybe it doesn't matter here)
* What if the user decides the choices she made were poor ones,, and wants to go back to the default colors?

jcoyne:
@ggeisler do you have any mockups that could help me with this?  All I found was https://ggeisler.github.io/lerna-dashboard/config-appearance.html

The fields are using html5 "color" input type, so I haven't been able to exert a lot of control over that.

There is no ticket stating that there should be a "reset to defaults" would you like to create one?

ggeisler:
@jcoyne As far as I know or remember, this was never a ticketed feature, which means it was never in the queue for design. So, no, there's no mockup. I will bring it up with @hannahfrost later today and see where she thinks it goes in the list of priorities.

jcoyne:
Ticket is here: https://github.com/projecthydra-labs/hyku/issues/286

ggeisler:
Okay. Note that in Waffle it is way down in the list of design needed tickets, far enough down that (in my memory) @hannahfrost and I have never discussed it as something I should prepare a mockup for in the near-term timeframe.

hannahfrost:
@jcoyne @ggeisler This is not a priority, so let's pick this up a bit later, post MVP

mjgiarlo:
Discussed with @hannahfrost today and though we know this doesn't meet our design standards yet, it will be compelling to demo the ability to dynamically alter the UI via the UI. We will revisit this implementation after MVP when we work on mock-ups. Thanks all!

PR #572, opened 2017-03-16
Further cleaning violations

jeremyf:
## Removing rubocop violation

@132cbc391d01a6b858dcdc06f55df747dbc4a720

In removing the violation, I'm also exposing a method that can be
overridden or extended in a downstream implementation.

## Removing exclusion that no longer applies

@b8c1aa9504f873c358f301ba4aa9e7152ea1b9dc

## Removing duplicate Rubocop key

@32dae1a1a4485357e82d5ccdd2b82517d1a4fd4d

With the key being declared twice, the last declaration was overriding
the initial declaration. So I'm taking the later declaration and moving
it up.

### DIFF
--- a/app/models/hyrax/uploaded_file.rb
+++ b/app/models/hyrax/uploaded_file.rb
@@ -6,8 +6,6 @@ module Hyrax
     mount_uploader :file, UploadedFileUploader
     belongs_to :user, class_name: '::User'
 
-    before_destroy do |obj|
-      obj.remove_file!
-    end
+    before_destroy :remove_file!
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/hyrax/uploaded_file.rb
```diff
@@ -6,8 +6,12 @@ class UploadedFile < ActiveRecord::Base
...
 
-    before_destroy do |obj|
-      obj.remove_file!
-    end
+    before_destroy :remove_file_before_destroy!
+
+    protected
+
+      def remove_file_before_destroy!
+        obj.remove_file!
```
jcoyne: `obj` is not defined here. Could you just have `before_destroy :remove_file!`?

PR #573, opened 2017-03-16
Tidying up Yard warning

jeremyf:
```console
[warn]: @param tag has unknown parameter name: the
    in file `lib/hyrax/configuration.rb' near line 430
```

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -425,10 +425,10 @@ module Hyrax
 
     private
 
-      # @param [Symbol] the symbol representing the model
+      # @param [Symbol, #to_s] model_name - symbol representing the model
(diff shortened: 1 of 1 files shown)


lib/hyrax/configuration.rb
```diff
@@ -425,10 +425,10 @@ def model_to_create
 
     private
 
-      # @param [Symbol] the symbol representing the model
+      # @param [Symbol, #to_s] concern_name - symbol representing the model
```
jcoyne: This is certainly better, but isn't the yard name supposed to match the parameter name?

jeremyf: Good catch…I changed my mind on the parameter name but forgot to change the documentation.

PR #581, opened 2017-03-17
Merge latest commits from Sufia (take three)

mjgiarlo:


app/views/hyrax/base/_form_relationships.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if Flipflop.assign_admin_set? %>
   <%= f.input :admin_set_id, as: :select,
-      collection: Hyrax::AdminSetService.new(controller).select_options(:deposit),
-      include_blank: true,
+      include_blank: false,
+      collection: Hyrax::AdminSetOptionsPresenter.new(Hyrax::AdminSetService.new(controller)).select_options,
```
jeremyf: Not a fan of this initialization sequence in a view.

Consider:

```ruby
Hyrax::AdminSetOptionsPresenter.select_options_for(controller: controller)
```

mjgiarlo: I don't disagree; this PR merges the latest stuff from Sufia over and does not otherwise introduce new code or patterns. Inclined to take this up later. Assuming it passes. ;)

mjgiarlo:
I'm seeing some failures that I'm guessing are due to the difference between how Sufia and Hyrax handle workflows. (Thinking of this change which was only made in Hyrax: https://github.com/projecthydra-labs/hyrax/pull/305/files)

app/views/hyrax/base/_form_relationships.html.erb
jeremyf: Can you add a `TODO` in the code to reflect the requested change?

jcoyne:
Still seeing the workflow_name problem here.

app/views/hyrax/base/_form_relationships.html.erb
mjgiarlo: Added. Patch forthcoming.

PR #583, opened 2017-03-17
Show non-approved status works

laritakr:
[WIP] 

The issue is to show all statuses of works on the admin/admin_set menu (which only admin can see), and add a column to the document list to show the column.

References https://github.com/projecthydra-labs/hyku/issues/726

### DIFF
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -15,7 +15,7 @@ module Hyrax
     self.single_item_search_builder_class = Hyrax::SingleAdminSetSearchBuilder
 
     # Used to get the members for the show action
-    self.member_search_builder_class = Hyrax::AdminSetMemberSearchBuilder
+    self.member_search_builder_class = Hyrax::AdminAdminSetMemberSearchBuilder
 
     # Used to get a list of admin sets for the index action
     self.list_search_builder_class = Hyrax::AdminSetSearchBuilder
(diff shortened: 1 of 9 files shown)


laritakr:
I have two questions on this issue:

1. The document list partials are shared the public collections view. Adding a status column to the admin/admin_sets show page also will add the same column to the collections show page. Because of this, I added a test so it would only show the status for the admin user, but is this the desired behavior since the collections list doesn't include anything other than deposited status?
2. If we do want to limit the status column to only show for the admin user, is this the correct way to do it? It seems as if it might be better to define a method somewhere so we could just say current_user.admin? but I wasn't sure where that should go or if it was even wanted?

Since the above questions will determine what I need to test for in the specs, I'm holding off on working on those until I get the answers.

Sample page views as it now stands (for an admin user):

screen shot 2017-03-17 at 3 14 40 pm

screen shot 2017-03-17 at 3 16 24 pm

ggeisler:
@laritakr My opinion on question 1 is no, there is no point in showing the status column to any user who will only see works with status of "deposited".  It seems safe to assume that any user who sees the list of works in that context will assume the works have a status of deposited.

laritakr:
Thanks @ggeisler I didn't think so either but wanted to double-check re: the intent. 

So that would mean that I need to create new partials to show the document list on the admin/admin_sets show page, and negates the need for answering question #2.

mjgiarlo:
@jcoyne thoughts on whether this should be 1.0.0 or 2.0.0?

jcoyne:
@mjgiarlo 1.0

PR #584, opened 2017-03-17
Update the AdminSet permissions when removing a PermissionTemplateAccess

### LINKED ISSUE #570: Removing AdminSet manager does not update edit_access in solr doc
### Descriptive summary

When a user is removed as manger of an AdminSet via the UI, the user is not removed from the edit_access_person_ssim in the solr doc.  (Likely the same is true for group managers.)

### Expected behavior

When a manager is removed, the edit_access_person/group_ssim in the solr doc for the AdminSet should have the user's id removed.

### Actual behavior

When a manager is removed, the edit_access_person_ssim is unchanged.

#### Prerequisite:
1. Create 1 admin user and 1 non-admin user
1. Create an Admin Set
  1. Login as an admin user
  1. Create an Admin Set and Save
1. Add non-admin user as a manager
  1. Click the Participants tab
  1. Add Participants -> Add user: -> set to non-admin user
  1. Set role to Manager
  1. Click Add button
  1. **VERIFY:** Check solr document for the new Admin Set in Solr and verify that the selected user is in the `edit_access_person_ssim` list.

#### Reproduce the issue

1. Remove the user as manager
  1. From the Participants tab, click the Remove button next to the non-admin user just added
  1. **NOTICE:** Check solr document for the Admin Set in Solr and notice that the selected user is still in the `edit_access_person_ssim` list.

**EXPECTED:** I would expect the `edit_access_person_ssim` to no longer include the removed manager.

jcoyne:
Fixes #570

### DIFF
--- a/app/models/concerns/hyrax/admin_set_behavior.rb
+++ b/app/models/concerns/hyrax/admin_set_behavior.rb
@@ -83,6 +83,13 @@ module Hyrax
       Sipity::Workflow.find_active_workflow_for(admin_set_id: id)
     end
 
+    # Calculate and update who should have edit access based on who
+    # has "manage" access in the PermissionTemplateAccess
+    def update_access_controls!
+      update!(edit_users: permission_template.access_grants.where(access: 'manage', agent_type: 'user').pluck(:agent_id),
+              edit_groups: permission_template.access_grants.where(access: 'manage', agent_type: 'group').pluck(:agent_id))
+    end
+
     private
 
       def destroy_permission_template
--- a/app/controllers/hyrax/admin/permission_template_accesses_controller.rb
+++ b/app/controllers/hyrax/admin/permission_template_accesses_controller.rb
@@ -5,11 +5,23 @@ module Hyrax
 
       def destroy
         @permission_template_access.destroy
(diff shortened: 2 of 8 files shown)


app/models/concerns/hyrax/admin_set_behavior.rb
```diff
@@ -83,6 +83,13 @@ def active_workflow
       Sipity::Workflow.find_active_workflow_for(admin_set_id: id)
     end
 
+    # Calculate and update who should have edit access based on who
+    # has "manage" access in the PermissionTemplateAccess
+    def update_access_controls!
+      update!(edit_users: permission_template.access_grants.where(access: 'manage', agent_type: 'user').pluck(:agent_id),
```
jeremyf: Can you extract local variables for these two lines. They are rather long and hard to quickly parse.

```ruby
new_edit_users = permission_template.access_grants.where(access: 'manage', agent_type: 'user').pluck(:agent_id)
new_edit_groups = permission_template.access_grants.where(access: 'manage', agent_type: 'group').pluck(:agent_id)
update!(edit_users: new_edit_users, edit_groups: new_edit_groups)
```

PR #585, opened 2017-03-17
Adding documentation and specs

jeremyf:
## Clarifying Hyrax::Operation

@7d793190752e63c32310b43186c2cddd598c3b17

Adding documentation and additional specs to better understand these
classes.

## Renaming parameter to reflect its implementation

@c3c811fd9dbb079f601ecb9d68f9f4731204f35e

Given that we have a class named Operation but the variable was "log",
rename the variable to "operation" to better reflect its usage.

### DIFF
--- a/app/models/hyrax/operation.rb
+++ b/app/models/hyrax/operation.rb
@@ -1,4 +1,5 @@
 module Hyrax
+  # The database storage of inter-related jobs and their states.
   class Operation < ActiveRecord::Base
     PENDING = 'pending'.freeze
     PERFORMING = 'performing'.freeze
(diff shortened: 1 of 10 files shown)


spec/models/hyrax/operation_spec.rb
```diff
@@ -12,6 +12,19 @@
       end
     end
 
+    describe "with only failng processes" do
```
jcoyne: typo here.

app/models/hyrax/operation.rb
```diff
@@ -11,33 +12,57 @@ class Operation < ActiveRecord::Base
...
 
     # If this is a batch job (has children), check to see if all the children are complete
+    # @see #fail! when none of the children are in a PENDING nor PERFORMING state but at least one is in a FAILURE state
+    # @see #success! none of the children are in a PENDING, PERFORMING, nor FAILURE state
     def rollup_status
       with_lock do
-        stats = children.pluck(:status)
+        # We don't need all of the status of the children, just need to see
+        # if there is at least one PENDING, PERFORMING, or FAILURE.
+        stats = children.select(:status).distinct.pluck(:status)
```
mjgiarlo: Can you say more about why this change is needed?

spec/models/hyrax/operation_spec.rb
mjgiarlo: I'd also like to see `s/describe/context/` here, but it's no big.

app/models/hyrax/operation.rb
jeremyf: Added comments inline.

The gist is: instead of transferring all of the statuses of the children from the database, to Rails, and then instantiating an array and leveraging the include that checks the whole array, opt instead to rely on the database to minimize the return values to that which is needed.

jeremyf:
Added comments inline.

The gist is: instead of transferring all of the statuses of the children from the database, to Rails, and then instantiating an array and leveraging the include that checks the whole array, opt instead to rely on the database to minimize the return values to that which is needed.

PR #587, opened 2017-03-17
AdminSetCreateService should not duplicate knowledge in the ASBehavior

mjgiarlo:


app/services/hyrax/admin_set_create_service.rb
```diff
@@ -35,7 +35,7 @@ def self.create_default_admin_set(admin_set_id:, title:)
     # @see AdminSet
     # @raise [RuntimeError] if you attempt to create a default admin set via this mechanism
     def self.call(admin_set:, creating_user:, **kwargs)
-      raise "Use .create_default_admin_set to create a default admin set" if admin_set.id == AdminSet::DEFAULT_ID
+      raise "Use .create_default_admin_set to create a default admin set" if AdminSet.default_set?(admin_set.id)
```
jeremyf: What you have is an improvement, but what about creating a method `AdminSet#default_set?`

jeremyf:
@mjgiarlo I pushed up my suggested change to this branch

PR #590, opened 2017-03-20
Adding constants

jeremyf:
## Reworking ProxyDepositRequest#transfer_to=

@ca104a651b8ce4297ffc634d76606a4bae1a28ef

Instead of relying setting an instance variable, instead prefer the more
direct method. This commit also includes documentation about the methods
usage as well as expectations.

## Extracting methods to reduce complexity

@b39c45885186d3a2173726c3615edca7f962761e

## Extracting ProxyDepositRequest constants and enum

@ebb16743fc414f23408325cff71838574dd29a6e

* Leveraging the enum to enforce the valid states for a
  ProxyDepositRequest.
* Creating constants to remove "magic strings"
* Extracting a `#fulfill!` method to encapsulate the common logic
* Adding spec to verify that we are persisting strings in the database
  for the enum values

### DIFF
--- a/app/models/proxy_deposit_request.rb
+++ b/app/models/proxy_deposit_request.rb
@@ -43,16 +43,26 @@ class ProxyDepositRequest < ActiveRecord::Base
 
   def send_request_transfer_message
     if updated_at == created_at
+      send_request_transfer_message_as_part_of_create
+    else
+      send_request_transfer_message_as_part_of_update
+    end
+  end
+
+  private
+
+    def send_request_transfer_message_as_part_of_create
       user_link = link_to(sending_user.name, Hyrax::Engine.routes.url_helpers.profile_path(sending_user.user_key))
       transfer_link = link_to('transfer requests', Hyrax::Engine.routes.url_helpers.transfers_path)
       message = "#{user_link} wants to transfer a work to you. Review all #{transfer_link}"
       User.batch_user.send_message(receiving_user, message, "Ownership Change Request")
-    else
+    end
+
+    def send_request_transfer_message_as_part_of_update
       message = "Your transfer request was #{status}."
       message += " Comments: #{receiver_comment}" unless receiver_comment.blank?
       User.batch_user.send_message(sending_user, message, "Ownership Change #{status}")
     end
-  end
 
   def pending?
     status == 'pending'
--- a/app/models/proxy_deposit_request.rb
+++ b/app/models/proxy_deposit_request.rb
@@ -37,7 +37,7 @@ class ProxyDepositRequest < ActiveRecord::Base
   end
 
   def should_not_be_already_part_of_a_transfer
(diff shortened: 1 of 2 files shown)


app/models/proxy_deposit_request.rb
```diff
@@ -31,55 +37,74 @@ def sending_user_should_not_be_receiving_user
...
 
-  def accepted?
-    status == 'accepted'
-  end
+  ACCEPTED = 'accepted'.freeze
+  PENDING = 'pending'.freeze
+  CANCELED = 'canceled'.freeze
+  REJECTED = 'rejected'.freeze
+
+  enum(
```
jcoyne: The documentation for enum explicitly states that it's for an integer column:
> Declare an enum attribute where the values map to integers in the database, but can be queried by name.

Are we sure this is a supported use case?  Should we submit a patch to the rails docs?

jeremyf: @jcoyne I've been doing the enum with non-integer values in Sipity (Rails 4.2.x). If it only supports integer, then I don't want to do the enum (not a fan of obfuscated values in the database)

jeremyf: I'm verifying now.

jeremyf: @jcoyne I added a verifying test that we are indeed persisting a human readable string (instead of the integer as the documentation says it supports).

jcoyne: A working use case doesn't mean it's a supported use case. :)

jeremyf: @jcoyne True…but if it fails downstream, we'll now know.

jeremyf: I'll submit a patch to the documentation and a spec to Rails.

jeremyf: There is this piece later in the existing documentation:

```ruby
  # Finally, it's also possible to explicitly map the relation between attribute and
  # database integer with a hash:
  #
  #   class Conversation < ActiveRecord::Base
  #     enum status: { active: 0, archived: 1 }
  #   end
```

jcoyne: Yes, but it still says "database integer".

jeremyf: This exists, but I'm not seeing tests supporting it https://github.com/rails/rails/blob/master/activerecord/test/models/book.rb#L18

jeremyf: I stand corrected, this exists in a test https://github.com/rails/rails/blob/master/activerecord/test/cases/enum_test.rb#L476-L479

jeremyf: Submitted documentation https://github.com/rails/rails/pull/28490

jcoyne:
@jeremyf You're the best!

PR #592, opened 2017-03-20
Directions for loading workflows

jcoyne:


mjgiarlo:
I didn't add this to the README because earlier we recommend installing via template, which already includes this task: https://github.com/projecthydra-labs/hyrax/blob/master/template.rb

I'm OK with being explicit about this, though, even if it means we may have users call it twice. (Shouldn't harm anything.)

PR #597, opened 2017-03-20
Fix no hyrax/editor and no admin set js error

little9:
JS fixes

This fixes two JS errors from the latest Sufia merge. 

One of the errors was `Error: No hyrax/editor` , which I fixed by changing the name of the class to editor_control and the other error was caused by not having an admin set, which I fixed by doing some checks. 


jcoyne:
Significant portions of this look like the fix currently in Sufia that is scheduled to be ported in the next merge.  See https://github.com/projecthydra/sufia/pull/3155  I prefer we use that to keep a consistent history with Sufia.

PR #603, opened 2017-03-20
Adding code to ensure workflows can be reloaded

jeremyf:
As reported by @treypendragon, when Princeton went to reload a workflow,
having just added methods, it encounter issues in new workflow states
being created and breaking the Sipity::Entity relationship

I believe this addresses the major problem, by ensuring that we don't
delete both the states that we transition_to as well as the states
that we are transitioning from.


mjgiarlo:
NOTE: We'd like @tpendragon to take this for a spin before merging.

PR #609, opened 2017-03-21
fix the bad link to the waffleboard

carrickr:
The README has a link pointing to 'waffle.io/projecthydra/hyrax' and is missing the '-labs'.  This fixes that.


mjgiarlo:
Thank you, @carrickr

PR #614, opened 2017-03-21
Remove hard coded autocomplete fields

little9:
Right now, there are hard coded arrays in the JS that are used for activating autocomplete fields. 
This removes them in favor of activating all the fields that have a `data-autocomplete-url` attribute on them. 

If a person wants to add autocomplete to a new field in Hyray they would need to initialize the Autocomplete class and pass it an array of the fields they want to be activated. This removes that step.

### DIFF
--- a/app/assets/javascripts/hyrax/authority_select.es6
+++ b/app/assets/javascripts/hyrax/authority_select.es6
@@ -44,8 +44,6 @@ export default class AuthoritySelect {
 	observer.observe(document.body, config);
     }
 
-    
-
     /**
      * Initialize bindings
      */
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/hyrax/autocomplete.es6
```diff
@@ -1,44 +1,37 @@
...
-    }
-    activateFields(autocompleteData, selector) {
-	for (let field in this.autocompleteFields) {
-	    if (autocompleteData === this.autocompleteFields[field]) {
-		this.autocomplete(selector);
-	    }
-	}
+  // This is the initial setup for the form.
+  setup () {
+    $('[data-autocomplete]').each((index, value) => {
```
jcoyne: Can we move this initialization code to the `app.js`. I'd like to pass the selected object to the initializer, then this class doesn't need to know about `$`. Ideally it would pass the result of `.data()` too.

little9:
Thanks for that suggestion! It's nice to be able to just do `Hyrax.autocomplete()`

app/assets/javascripts/hyrax/authority_select.es6
```diff
@@ -60,16 +58,5 @@ export default class AuthoritySelect {
  * intialize the Hyrax autocomplete with the fields that you are using
  */
 function setupAutocomplete() {
```
jcoyne: Why is this function in the global scope?

little9: It's not exported so only that class can use it. When activating the autocomplete was more involved I needed it that way, but may not anymore.

PR #615, opened 2017-03-21
Don't allow creating works without having an AdminSet to deposit to

### LINKED ISSUE #601: If a user is not a depositor or manager for any admin set, then do not show them the ‘Create new work’ page
Only people who can deposit into any admin set should be able to create a new work.

jcoyne:
Fixes #601

This should probably wait until 2.0


jcoyne:
@mjgiarlo 1.0 branch has been created, so this could be reviewed for merge.

PR #622, opened 2017-03-22
Move microdata out of translation files. Fixes #536

### LINKED ISSUE #536: Remove the `schema_org` key from the i18n files
the schema_org stuff is configuration, but it's not language dependent, so we should put it elsewhere.  When more languages that added, it makes it challenging to configure schema.org microdata as you have to make the changes N times.

jcoyne:
```console
Hyrax::Microdata
  .load_paths
    should be a kind of Array
    can be appended to
  .load_paths=
    replaces the existing
  .fetch
    with one file in the load path
      with a default
        and a miss on the key
          should be the given default value
        and hit on the key
          should be the registered value
      without a default
        and a miss on the key
          should be nil
        and hit on the key
          should be the registered value
    with multiple files in the load path
      and a miss on the key
        should be the given default value
      and what would be a hit on the first and second loaded file
        should be the second value loaded
      and a hit on the first loaded file but not the second
        should be the first loaded value
      and a hit on the second loaded file but not the first
        should be the first loaded value
```

### DIFF
--- a/app/presenters/hyrax/presents_attributes.rb
+++ b/app/presenters/hyrax/presents_attributes.rb
@@ -34,7 +34,7 @@ module Hyrax
 
     def microdata_type_to_html
       return "" unless display_microdata?
-      value = I18n.t(microdata_type_key, default: Hyrax.config.microdata_default_type)
+      value = Microdata.fetch(microdata_type_key, default: Hyrax.config.microdata_default_type)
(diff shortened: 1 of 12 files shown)


jcoyne:
That's thoughtful. There is lots of coverage in the renderer specs, so maybe just move that to a unit test for this class?

jeremyf:
@jcoyne - I'll look to those

Issue #621, opened 2017-03-22
Review submissions shows works you are not authorized to approve


jcoyne:
When I go have granted myself the "one_step_mediated_deposit - approving" role (and no others) and I go to http://localhost:3000/admin/workflows?locale=en

I see works that need approval, however when I click on them I get the unauthorized message.

This is because the document has the following solr field:

```
"actionable_workflow_roles_ssim"=>["one_step_mediated_deposit-depositing", "one_step_mediated_deposit-approving"]
```

jeremyf:
I'm not familiar with how the SOLR query is used in abilities. My gut says that it relates to entity level permissions not being found. I'm looking into this.

jeremyf:
@jcoyne And this is where I realize…I have no idea how permissions happen in this application. I see it is in the Ability class, but wow!

jcoyne:
@jeremyf the bigger questions is, should a deposited work that is waiting for review have `one_step_mediated_deposit-depositing` as one of the roles that can do something

jeremyf:
I can't duplicate.

mjgiarlo:
@jcoyne Do you suspect this is now close-able?

PR #624, opened 2017-03-22
Extract a common method for WorkflowFlow#label

jeremyf:
This is a stop-gap solution for #621

banners_and_alerts_and_index_workflow_role____hyrax

Note: I did not get the AdminSet's title as that would require yet another query.


app/models/sipity/workflow_role.rb
```diff
@@ -9,5 +9,12 @@ class WorkflowRole < ActiveRecord::Base
     has_many :workflow_responsibilities, dependent: :destroy, class_name: 'Sipity::WorkflowResponsibility'
     has_many :workflow_state_action_permissions, dependent: :destroy, class_name: 'Sipity::WorkflowStateActionPermission'
     has_many :entity_specific_responsibilities, dependent: :destroy, class_name: 'Sipity::EntitySpecificResponsibility'
+
+    # @todo This is a hack; I don't want to include reference to the admin set; However based on the current UI this is required.
```
jcoyne: Can you update this comment and explain why "based on the current UI this is required"?

PR #633, opened 2017-03-23
Fixing microdata introduced error

jeremyf:
I needed to both clear and reload the data.


spec/services/hyrax/microdata_spec.rb
```diff
@@ -9,7 +9,8 @@ module Hyrax
 
     after do
       # Need to reset as this is an instance variable
-      described_class.instance_variable_set("@load_paths", nil)
+      described_class.clear_load_paths!
+      described_class.reload!
```
jcoyne: `clear_load_paths!` already calls `reload!`

PR #642, opened 2017-03-24
Display Rights Statement on the form

jcoyne:


app/services/hyrax/curation_concern.rb
```diff
@@ -1,5 +1,6 @@
 module Hyrax
   class CurationConcern
+    Rails.logger.warn "INIT CC"
```
mjgiarlo: For debugging, or ... ?

jcoyne: Oops, need to remove that.

jcoyne:
Let's get #646 merged first, then we can remove that rubocop exception.

PR #648, opened 2017-03-24
Actors now only take Ability

jcoyne:
Passing a User was previously deprecated


jeremyf:
👍  for removing additional setup mocks. That tells me things are moving in a better direction.

PR #654, opened 2017-03-24
introduce json response to workflow_actions

### LINKED ISSUE #651: workflow_actions_controller respond_to JSON
A data migration project were working on at OSU would greatly benefit from the workflow_actions_controller handling JSON. JSON should return the updated work and a 200 on success, or error. Currently the controller returns a 302 redirect to the works `#show` page, which makes sense for HTML requests.

revgum:
fixes #651

### DIFF
--- a/app/controllers/hyrax/workflow_actions_controller.rb
+++ b/app/controllers/hyrax/workflow_actions_controller.rb
@@ -3,16 +3,19 @@ module Hyrax
     before_action :authenticate_user!
 
     def update
-      work = ActiveFedora::Base.find(params[:id])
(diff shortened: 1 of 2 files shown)


app/controllers/hyrax/workflow_actions_controller.rb
```diff
@@ -1,18 +1,22 @@
 module Hyrax
   class WorkflowActionsController < ApplicationController
     before_action :authenticate_user!
+    attr_accessor :curation_concern
```
jcoyne: why are we making this attr_accessor and not using it?

PR #662, opened 2017-03-27
Refactor ActorStack to use ActionDispatch:MiddlewareStack

jcoyne:
This allows a more customizable actor stack (add, remove, reorder) and leverages code that is well tested and maintained (rather than a custom solution).  This approach uses a functional programming style, where no state is stored in the actors.  This allows us to greatly reduce object allocations.


app/actors/hyrax/actors/interpret_visibility_actor.rb
```diff
@@ -204,19 +211,19 @@ def parse_date(date_string)
         end
 
         # If they want a lease, we can assume it's valid
-        def apply_lease
-          return true unless @intention.wants_lease?
-          curation_concern.apply_lease(*@intention.lease_params)
-          return unless curation_concern.lease
-          curation_concern.lease.save # see https://github.com/projecthydra/hydra-head/issues/226
+        def apply_lease(intention, env)
```
jeremyf: It appears that most other methods with `env` put `env` as the first attribute. Switch the order of these parameters.

Note: This is also a placeholder for a more in-depth review. In theory, I like leveraging the MiddlewareStack, I want to review this a bit more.

Also, adding documentation about modifying the stack would be great (unless that is a follow-up PR).

jcoyne:
@jeremyf please revisit. I've added documentation about customizing the actor middleware stack.

PR #667, opened 2017-03-28
adds json respond_to for hyrax controller

### LINKED ISSUE #666: /users/sign_in fails on POST using JSON
Devise expects the controller to respond_to json but `Hyrax::Controller` doesn't have this explicitly set. 

Two approaches I've tested that work, but have different implications.

1. Just tell `Hyrax::Controller` to respond to json by adding `respond_to :html, :json` to the `included` block. I've tested this locally and it works as expected with user login using json. 

2. Create a `Hyrax::SessionsController` to handle it.

```
module Hyrax
  class SessionsController < Devise::SessionsController
    respond_to :html, :json
  end
end
```

Along with this new controller, Hyrax would need a route update. This approach was also tested locally and works.
```  
devise_for :users, skip: [:sessions]
  as :user do
    get '/users/sign_in', to: 'hyrax/sessions#new', as: :new_user_session
    post '/users/sign_in', to: 'hyrax/sessions#create', as: :user_session
    get '/users/sign_out', to: 'hyrax/sessions#destroy', as: :destroy_user_session
  end
```

Looking for thoughts/feedback.

revgum:
fixes #666

### DIFF
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -8,6 +8,7 @@ module Hyrax::Controller
     include Hydra::Controller::ControllerBehavior
     helper_method :create_work_presenter
     before_action :set_locale
+    respond_to :html, :json
   end
 
   # Provide a place for Devise to send the user to after signing in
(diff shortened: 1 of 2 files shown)


jcoyne:
I don't think this fix is scoped too broadly.  If you need this just to affect the Sessions controller, then you should do it in the sessions controller.  That should probably be an application concern.

revgum:
@jcoyne did you read the details in issue #666 by chance? How do you feel about a SessionsController within Hyrax that supports json for login, which solves the underlying issue and allows for institutions to override it for themselves?

jcoyne:
@revgum yep, read it. I don't want to have a tight dependency on Devise in Hyrax, because that precludes the possibility of using other authentication gems.

PR #691, opened 2017-04-03
Don't show create work button to users who don't have the depositor role

### LINKED ISSUE #677: A new repo allows a user without any admin sets to deposit
This should have been fixed in #615, but it seems like it's not working.

jcoyne:
Fixes #677

### DIFF
--- a/app/presenters/hyrax/homepage_presenter.rb
+++ b/app/presenters/hyrax/homepage_presenter.rb
@@ -6,8 +6,19 @@ module Hyrax
       @current_ability = current_ability
     end
 
+    # @return [Boolean] If the current user is a guest and the display_share_button_when_not_logged_in?
+    #   is activated, then return true. Otherwise return true if the signed in
+    #   user has permission to create at least one kind of work.
     def display_share_button?
-      Hyrax.config.always_display_share_button? || current_ability.can_create_any_work?
+      (user_unregistered? && Hyrax.config.display_share_button_when_not_logged_in?) ||
+        current_ability.can_create_any_work?
     end
+
+    private
+
+      def user_unregistered?
+        current_ability.current_user.new_record? ||
+          current_ability.current_user.guest?
+      end
   end
 end
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -347,12 +347,18 @@ module Hyrax
       @cache_path ||= ->() { Rails.root + 'tmp' + 'uploads' + 'cache' }
     end
 
-    # Should a button with "Share my work" show on the front page to all users (even those not logged in)?
-    attr_writer :always_display_share_button
-    def always_display_share_button?
-      return true if @always_display_share_button.nil?
-      @always_display_share_button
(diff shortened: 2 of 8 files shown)


app/presenters/hyrax/homepage_presenter.rb
```diff
@@ -6,8 +6,19 @@ def initialize(current_ability)
       @current_ability = current_ability
     end
 
+    # @return [Boolean] If the current user is a guest and the always_display_share_button
+    #   is activated, then return true. Otherwise return true if the signed in
+    #   user has permission to create at least one kind of work.
     def display_share_button?
-      Hyrax.config.always_display_share_button? || current_ability.can_create_any_work?
+      (user_unregistered? && Hyrax.config.always_display_share_button?) ||
```
jeremyf: I'm a bit confused by the logic, is it the same as the following? If so, I prefer the following as it flows better when reading. There is a nuanced difference in that the `Hyrax.config.always_display_share_button?` has a higher order of precedence.

The logic as it stands seems to say only care about the `always_display_share_button?` if we have an unregistered user.

```ruby
return true if Hyrax.config.always_display_share_button?
return false if user_unregistered?
current_ability.can_create_any_work?
```

jcoyne: No, it is not the same.  `Hyrax.config.always_display_share_button?` is only a factor if the user is not logged in.

jeremyf: Is that feature a bit of a misnomer then? I read "always_display_share_button?" as if this is true then method "display_share_button?" should also be true.

jcoyne: It was a good name once upon a time, when all registered users could create all work types. It's no longer a good name.  Want me to change that now?

jeremyf: If you could change it that would help clear up what this means. Is that a breaking change?

jcoyne: We could probably do it with some deprecation in the configuration.

lib/hyrax/configuration.rb
```diff
@@ -365,14 +365,21 @@ def cache_path
...
-    deprecation_deprecate always_display_share_button: "use always_display_share_button? instead"
+    # Should a button with "Share my work" show on the front page to users who are not logged in?
+    attr_writer :display_share_button_when_not_logged_in
+    def display_share_button_when_not_logged_in?
+      return true if @display_share_button_when_not_logged_in.nil?
+      @display_share_button_when_not_logged_in
+    end
+
+    alias always_display_share_button= display_share_button_when_not_logged_in=
+    deprecation_deprecate :always_display_share_button= => "use always_display_share_button= instead"
```
jeremyf: Should this say:

```ruby
deprecation_deprecate :always_display_share_button= => "use display_share_button_when_not_logged_in = instead"
```

jcoyne: Yep. Change made.

PR #699, opened 2017-04-04
Put transfers on the admin dashboard

jcoyne:


app/views/hyrax/transfers/index.html.erb
```diff
@@ -1,7 +1,13 @@
-<h1><%= t("hyrax.dashboard.transfer_of_ownership") %></h1>
+<% provide :page_header do %>
```
atz: @jcoyne can you give a word here why `provide` instead of `content_for`?  Might not downstream consumers wish to further append to `:page_header`?

jcoyne: See https://github.com/projecthydra-labs/hyrax/pull/706 for more information.  It's the way all the transfers to the template are done now in Hyrax

PR #706, opened 2017-04-04
Prefer provide over content_for

jcoyne:
Provide knows to expect no more input for this block after it's been
called once. This enables streaming responses.

See:
http://api.rubyonrails.org/classes/ActionController/Streaming.html#module-ActionController::Streaming-label-Communication+between+layout+and+template

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

* 
* 
* 


atz:
I am unconvinced that nobody downstream will ever want to append additional `content_for` to a given element like `:page_header`.  Pretty sure this will bite us in the ass soon.

PR #709, opened 2017-04-04
Reuse common point of override for labels

atz:
We know the `product_name` (via `application_name` helper method) is
going to be overridden every time.  So let's not force the users to
override additional labels that also hardcode the same "Hyrax" string.

See: https://github.com/projecthydra-labs/hyku/issues/904


atz:
Ah, crap, that didn't work.  Not sure why `%{application_name}` is not being interpreted like in other labels.

PR #713, opened 2017-04-05
Remove deprecated configuration methods

jcoyne:
These were deprecated in Hyrax 1.0, so we can remove them for 2.0


atz:
Should we defer this until closer to 2.0 release time, or are we OK to merge it now?

jcoyne:
@atz I think this can go now, the master is already declaring it is 2.0.0.alpha

PR #715, opened 2017-04-05
Update header/footer text and link colors for better contrast

ggeisler:
Current text and link colors in the header and footer do not pass AA minimum contrast:

before

This PR brightens the colors to pass at AA (I'm not super excited by this color palette, but maybe we can revisit it in the future; presumably implementors will change the colors anyway):

after

### DIFF
--- a/app/assets/stylesheets/hyrax/_footer.scss
+++ b/app/assets/stylesheets/hyrax/_footer.scss
@@ -22,6 +22,11 @@ body {
   bottom: 0;
   height: $footer-height;
   width: 100%;
+
+  a,
+  &.navbar-inverse .navbar-link {
+    color: #e27800;
+  }
 }
 
 @media (max-width: $screen-sm) {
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/hyrax/_footer.scss
```diff
@@ -22,6 +22,11 @@ body {
   bottom: 0;
   height: $footer-height;
   width: 100%;
+
+  a,
+  &.navbar-inverse .navbar-link {
```
jcoyne: How is this changing the header color?  All of this is within a `.site-footer` parent selector?

jcoyne: I see, you're changing the customization defaults below.  How about I add a customizable footer link color into the appearance menu instead?

ggeisler: Sure, if you want to do that, it definitely makes sense.

PR #717, opened 2017-04-05
Default admin set grants the registered group deposit access

### LINKED ISSUE #602: the default admin set should by default give every logged in user deposit access
When the permission template for the default admin set is created it should give the `registered` group deposit access.

jcoyne:
Fixes #602


spec/models/hyrax/permission_template_spec.rb
```diff
@@ -43,7 +43,7 @@
     it 'queries the underlying access_grants' do
       template = create(:permission_template)
       to_find = template.access_grants.create!(agent_type: 'user', access: 'manage', agent_id: '123')
-      template.access_grants.create!(agent_type: 'user', access: 'read', agent_id: '456')
+      template.access_grants.create!(agent_type: 'user', access: 'view', agent_id: '456')
```
jeremyf: Good catch

PR #722, opened 2017-04-06
Turn off turbolinks for admin set edit page with tabs.

### LINKED ISSUE #254: Back button gets broken
If you go to https://nacho.demo.hydrainabox.org/admin/admin_sets/9p2909328/edit

Then click the workflow tab (https://nacho.demo.hydrainabox.org/admin/admin_sets/9p2909328/edit#workflow)
and then click the notifications (https://nacho.demo.hydrainabox.org/notifications?locale=en), then click the back button, it doesn't go back.

andjsmit:
Add body attributes to hyrax and dashboard templates.

Fixes #254

Allows back button to function on pages with tabs.

Currently, turbolinks and bootstrap tabs conflict. This PR allows for disabling turbolinks on a single page by adding a body tag attribute of `data-turbolinks=false`. Doing so on a page with tabs enables the browser back button to function when leaving the page. The tab enabled before leaving the page is returned to on back button click. Prior, clicking the back button did nothing.

* Add support for adding body attributes
* Disable turbolinks for pages using bootstrap tabs
* If accepted, additional issues for disabling turblolinks on pages with tabs should be created.


jcoyne:
@andjsmit have you tried just putting the `data-no-turbolink='true'` attribute on the tab links rather than turning it off for every link on the page?

andjsmit:
@jcoyne Yes, did try putting `data-turbolinks=false` and `data-no-turbolink=true` on tab links. Unfortunately, it needs to be on the outbound links when leaving the page.

app/views/hyrax/admin/admin_sets/edit.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for(:body_attributes, 'data-turbolinks="false"'.html_safe) %>
```
jcoyne: use `provide` rather than content for.  Additionally add a comment about why you are doing this (perhaps reference the issue)

app/views/layouts/hyrax.html.erb
```diff
@@ -5,7 +5,7 @@
     <%= content_for(:head) %>
   </head>
 
-  <body>
+  <body <%= yield(:body_attributes) %>>
```
jcoyne: Is this necessary? the admin set edit page never uses this layout.

app/views/hyrax/admin/admin_sets/edit.html.erb
```diff
@@ -1,3 +1,6 @@
+<%# Needed for bootstrap tabs use on this page (https://github.com/projecthydra-labs/hyrax/issues/254) %>
```
jcoyne: Can you be even more descriptive?  Why are we turning off turbolinks?  There is nothing in this ticket about turbolinks: https://github.com/projecthydra-labs/hyrax/issues/254

jcoyne:
It looks to me that this code is supposed to be handling the case you are looking at:
https://github.com/projecthydra-labs/hyrax/blob/master/app/assets/javascripts/hyrax/tabs.js#L28

Have you investigated this code at all?

jcoyne:
As far as I can tell this behavior was a regression introduced by https://github.com/projecthydra/sufia/commit/9e485c798d0b7580e0058f2beebf38088303da29.  I'm able to make the behavior go away by commenting out the `window.addEventListener("popstate", tabNavigation);` line.

jcoyne:
Maybe we could replace the popstate (when Turbolinks is defined) with:
```
Blacklight.onLoad(tabNavigation);
```

jcoyne:
I think this will just fix the problem rather than working around it: https://github.com/projecthydra-labs/hyrax/pull/732  does that work for you?

andjsmit:
@jcoyne I tried several changes to tabs.js but either got no change in behavior or rendered the code useless as the #tab was not returned to on back button. #732 works great for me. I'm going to close this PR and remove the branch in favor of it. Thanks.

PR #723, opened 2017-04-06
Don't delegate representative_id twice

jcoyne:


jeremyf:
Upon further inspection there are 4 delegations to the solr_document, each with comments regarding their purpose. I looked at extracting constants but it seemed overkill.

PR #725, opened 2017-04-06
Separate the indexing from data modeling

jcoyne:


app/indexers/hyrax/basic_metadata_indexer.rb
```diff
@@ -0,0 +1,53 @@
...
+      def stored_searchable_index_config
+        stored_fields.each_with_object({}) do |name, hash|
+          hash[name] = ActiveFedora::Indexing::Map::IndexObject.new(name) do |idx|
+            idx.as :stored_searchable
+          end
+        end
+      end
+
+      def symbol_index_config
+        { import_url: ActiveFedora::Indexing::Map::IndexObject.new(:import_url) do |idx|
```
jeremyf: This is a bit confusing regarding syntax. I'm going to add a commit that extracts the value to a local variable to assist in legibility.

PR #735, opened 2017-04-07
update form permission partial name

tampakis:
The `form_permission_under_embargo` and `form_permission_under_lease` partials don't exist. I think it's a typo.

Edit: Copied the missing partials over from CurationConcerns (no need to modify [form permission](https://github.com/projecthydra-labs/hyrax/blob/a744cffd70e93151f7cbe88dc620da495e955a47/app/views/hyrax/base/_form_permission.html.erb#L2-L5) partial).

### DIFF
--- a/app/views/hyrax/base/_form_permission_under_embargo.html.erb
+++ b/app/views/hyrax/base/_form_permission_under_embargo.html.erb
@@ -1,0 +1,16 @@
+<fieldset class="set-access-controls">
+  <legend>
+    Visibility
+    <small>Who should be able to view or download this content?</small>
+  </legend>
+
+  <section class="help-block">
+    <p>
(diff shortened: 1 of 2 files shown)


tampakis:
@jcoyne - updated

app/views/hyrax/base/_form_permission_under_embargo.html.erb.txt
```diff
@@ -0,0 +1,16 @@
+<fieldset class="set-access-controls">
```
jcoyne: Why is this a `.txt` file? Shouldn't it be a `.erb`?

app/views/hyrax/base/_form_permission_under_lease.html.erb.txt
```diff
@@ -0,0 +1,18 @@
+<fieldset class="set-access-controls">
```
jcoyne: Same with this file.

tampakis:
Oops. That’s what happen when I put in changes while I’m sleepy

PR #744, opened 2017-04-07
Update layout of Workflow Roles page

### LINKED ISSUE #721: Revise layout of Workflow Roles page
The details of this page might be affected by the outcome of #674, but generally the layout and appearance of the Workflow Roles page should be updated to more closely match other Admin pages. Specifically, the current Workflow Roles page doesn't have a proper page heading and uses a non-Bootstrap save button. Also, I think the page elements could be rearranged to make the page more usable when there are many existing roles assigned (by putting the list at the bottom of the page) and more closely match the layout of the Admin Set Participants page.

### Current
current

### Suggested

edit_admin_set____hyrax

ggeisler:
Fixes #721 

Improves the layout of the Workflow Roles page:

* Adds standard page header
* Moves assign role form to top of page, so it'll be easier to get to when the number of assigned roles becomes long
* Better separates the Assign Role and Current Roles elements
* Improves presentation of the roles currently assigned to a user

I think we can further improve the Assign Role section (e.g., I don't think the required labels are really necessary here and it would be nice for the select menus to display "Select user..." and "Select workflow role..." instead of being blank) but this is probably fine for the time being.

### Before
before

### After
after

### DIFF
--- a/app/assets/stylesheets/hyrax/dashboard.scss
+++ b/app/assets/stylesheets/hyrax/dashboard.scss
@@ -162,6 +162,16 @@ body.dashboard {
     margin-bottom: 6px;
   }
 
+  .workflow-roles {
+    list-style-type: none;
+    margin-bottom: 0;
+    padding-left: 0;
+
+    .fa-remove {
+      color: $brand-danger;
+    }
+  }
+
   .navbar {
     margin-bottom: 0;
   }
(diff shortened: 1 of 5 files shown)


atz:
Seems like just one conflict/rebase question in a scss file.

ggeisler:
@jcoyne @mjgiarlo @atz Can one of you double-check this conflict? I get no conflict when I rebase locally. Is it just an extra blank line it is saying is a conflict? I'm not sure how to interpret this one.

jcoyne:
@ggeisler I can check it.

jcoyne:
@ggeisler rebased.

Issue #753, opened 2017-04-10
Non-admin user can not submit a work


jcoyne:
Pressing the submit button goes to a 404 page. Seems to work when logged in as an admin though.

jcoyne:
Waiting on  https://github.com/projecthydra-labs/hyku/pull/939 for further diagnosis.

mjgiarlo:
Unblocked.

jcoyne:
This is confusing.
Admin gets this:
```
10.0.4.194 - - [11/Apr/2017:20:34:42 +0000] "POST /concern/generic_works?locale=en HTTP/1.1" 302 176 "https://hyphy.demo.hydrainabox.org/concern/generic_works/new?locale=en" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:52.0) Gecko/20100101 Firefox/52.0" "97.116.190.118"
```

Regular user gets:
```
10.0.4.194 - - [11/Apr/2017:20:32:05 +0000] "POST /concern/generic_works?locale=en HTTP/1.1" 404 1564 "https://hyphy.demo.hydrainabox.org/concern/generic_works/new?locale=en" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:52.0) Gecko/20100101 Firefox/52.0" "97.116.190.118"
```

jcoyne:
Looks like this was just a problem in my version, which was created before active_workflow was a thing.  Not a problem in a fresh instance.

mjgiarlo:
@jcoyne does this mean we ought to capture this in release notes somewhere?

jcoyne:
@mjgiarlo only if you created an admin set before admin sets automatically loaded workflows.  Sometime back in January?

mjgiarlo:
:ok_hand:

PR #755, opened 2017-04-10
Raise a custom error when the work form can't find a workflow

jcoyne:
The default error raised by find_by! is `ActiveRecord::RecordNotFound`.
Rails middleware catches this error and renders 404.  This makes it
particulary hard to debug the case where we haven't set up a workflow.
So, a custom type of error will make it easier to find this particular
problem

Ref #753


lib/hyrax/errors.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  # Error that is raised when an active workflow can't be found
+  class MissingWorkflowError < HyraxError
+  end
+
+  class WorkflowAuthorizationException < HyraxError
+  end
+
+  class SingleUseError < StandardError; end
```
mjgiarlo: Should this inherit from `HyraxError` like the others? Also, should we break the `end` onto the next line like the others?

spec/forms/hyrax/forms/work_form_spec.rb
```diff
@@ -182,4 +182,16 @@
     subject { described_class.required_fields }
     it { is_expected.to eq [:title, :creator, :keyword, :rights_statement, :license] }
   end
+
+  describe ".workflow_for" do
+    subject { described_class.send(:workflow_for, admin_set_id: admin_set.id) }
```
atz: Why do you have to do `.send` here for a non-private method?

atz: Ah, I see you have it declared a `private_class_method`.

PR #757, opened 2017-04-10
Add migration task to switch direction of collection membership

hackmastera:
for upgrade path from sufia 7.3

### DIFF
--- a/lib/hyrax/collections_migration.rb
+++ b/lib/hyrax/collections_migration.rb
@@ -4,10 +4,14 @@ module Hyrax
       ::Collection.all.each do |collection|
         collection.members.each do |member|
           member.member_of_collections << collection
-          member.save
+          unless member.save
+            raise "Work #{member.id} failed to save after recording its membership in collection #{collection.id}"
+          end
         end
         collection.members = []
-        collection.save
+        unless collection.save
+          raise "Collection #{collection.id} failed to save after emptying its member list"
+        end
       end
     end
   end
--- a/lib/hyrax/collections_migration.rb
+++ b/lib/hyrax/collections_migration.rb
@@ -1,0 +1,14 @@
+module Hyrax
+  class CollectionsMigration
+    def self.run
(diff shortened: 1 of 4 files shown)


mjgiarlo:
@hackmastera once merged, would you mind adding a reference to this rake task in the 1.0.0 release notes?

hackmastera:
Sure thing.

@mjgiarlo You may want to re-assert approval since I used the nuclear option on rubocop.

mjgiarlo:
@jcoyne I'm inclined to merge. Do you have druthers?

jcoyne:
@mjgiarlo @hackmastera should this merge to the 1-0-stable branch, or should we backport after merging to master?

hackmastera:
It's in cc already, so maybe just 1.0? That would mean people can't use an alternate upgrade path (i.e. must go first to cc2). Not sure if it's worth having in master just in case a cc person wants to do it differently.

lib/hyrax/collections_migration.rb
```diff
@@ -0,0 +1,14 @@
+module Hyrax
+  class CollectionsMigration
+    def self.run
+      ::Collection.all.each do |collection|
+        collection.members.each do |member|
+          member.member_of_collections << collection
+          member.save
```
jcoyne: should we consider save! or checking the return values?

lib/hyrax/collections_migration.rb
```diff
@@ -0,0 +1,14 @@
...
+        end
+        collection.members = []
+        collection.save
```
jcoyne: same here.

lib/hyrax/collections_migration.rb
atz: Of course if you throw exception or abort, your repo data is stuck in an unworkable state.  But it is better to make it obvious than silently propagate that error.

hackmastera:
closing to move to 1-0-stable branch

PR #761, opened 2017-04-10
Create a :review, :submission ability

### LINKED ISSUE #910: Non-admin reviewers need to be able to review all works in their admin set
### Descriptive summary

Hyrax 1-0-Stable branch workflow process does not give a non-admin user the ability to approve anything other than public works.

It is possible that there is a way to manage this that I didn't find (through permissions on the admin set perhaps?) but if so, the steps need to be clearly documented, and a user should not be allowed to be assigned the reviewer role without the correct permissions. 

### Rationale

Only the admin user has the review submissions menu. Other users with the approving role can only approve by clicking the link to the work in the notifications.  I don't know whether the review submissions menu allows the approving role to have access to view works they are otherwise not authorized to see, but they definitely can't access them via the link in the notification.

### Expected behavior

If a user has the approving role, they should be able to view and approve the work.

### Actual behavior

The user cannot view or review a work unless it is public.
[Spreadsheet of details from testing of workflows](https://docs.google.com/spreadsheets/d/1zQsWlYMVs7gz6dHaOiXICygLj_5GTFOmbpI9MHRAq5o/edit#gid=0) on both stable and master branches. 

[The gist shows the steps taken, and the highlighted line shows the point where this issue was reached.](https://gist.github.com/laritakr/3020bdd70609a7fc224078dca1871fd8#file-gistfile1-txt-L50)

### Related work

https://github.com/projecthydra-labs/hyrax/issues/882

jcoyne:
Previously one could be an admin set magager, but they would never see
the link to "Review Submissions"

Fixes #910

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -125,6 +125,10 @@ module Hyrax
         can [:create, :edit, :update, :destroy], Hyrax::PermissionTemplateAccess do |access|
           test_edit(access.permission_template.admin_set_id)
         end
+
+        can :review, :submissions do
+          can_review_submissions?
+        end
       end
 
       def operation_abilities
@@ -155,6 +159,7 @@ module Hyrax
 
       def admin_permissions
         return unless admin?
+        # TODO: deprecate this. We no longer have a dashboard just for admins
         can :read, :admin_dashboard
         alias_action :edit, to: :update
         alias_action :show, to: :read
@@ -187,6 +192,16 @@ module Hyrax
 
       def curation_concerns_models
         [::FileSet, ::Collection] + Hyrax.config.curation_concerns
+      end
+
+      def can_review_submissions?
+        # Are there any workflows where this user has the "approving" responsibility
+        approving_role = Sipity::Role.find_by(name: 'approving'.freeze)
+        return false unless approving_role
+        Hyrax::Workflow::PermissionQuery.scope_processing_agents_for(user: current_user).any? do |agent|
+          agent.workflow_responsibilities.joins(:workflow_role)
+               .where('sipity_workflow_roles.role_id' => approving_role.id).any?
+        end
       end
   end
 end
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -195,8 +195,13 @@ module Hyrax
       end
 
       def can_review_submissions?
+        # Short-circuit logic for admins, who should have the ability
+        # to review submissions whether or not they are explicitly
+        # granted the approving role in any workflows
+        return true if admin?
+
         # Are there any workflows where this user has the "approving" responsibility
-        approving_role = Sipity::Role.find_by(name: 'approving'.freeze)
+        approving_role = Sipity::Role.find_by(name: Hyrax::RoleRegistry::APPROVING)
         return false unless approving_role
         Hyrax::Workflow::PermissionQuery.scope_processing_agents_for(user: current_user).any? do |agent|
           agent.workflow_responsibilities.joins(:workflow_role)
--- a/app/controllers/hyrax/admin/workflows_controller.rb
+++ b/app/controllers/hyrax/admin/workflows_controller.rb
@@ -1,7 +1,7 @@
 module Hyrax
   # Presents a list of works in workflow
   class Admin::WorkflowsController < ApplicationController
-    before_action :ensure_admin!
+    before_action :ensure_authorized!
     layout 'dashboard'
 
     def index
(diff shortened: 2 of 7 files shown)


app/models/concerns/hyrax/ability.rb
```diff
@@ -125,6 +125,15 @@ def admin_set_abilities
         can [:create, :edit, :update, :destroy], Hyrax::PermissionTemplateAccess do |access|
           test_edit(access.permission_template.admin_set_id)
         end
+
+        can :review, :submissions do
```
mjgiarlo: Tagging @jeremyf for review here.

app/models/concerns/hyrax/ability.rb
```diff
@@ -155,6 +164,7 @@ def curation_concerns_permissions
 
       def admin_permissions
         return unless admin?
+        # TODO: deprecate this. We no longer have a dashboard just for admins
```
mjgiarlo: :clap:

app/models/concerns/hyrax/ability.rb
```diff
@@ -188,5 +193,15 @@ def user_is_depositor?(document_id)
       def curation_concerns_models
         [::FileSet, ::Collection] + Hyrax.config.curation_concerns
       end
+
+      def can_review_submissions?
+        # Are there any workflows where this user has the "approving" responsibility
+        approving_role = Sipity::Role.find_by(name: 'approving'.freeze)
```
jeremyf: I'm not a fan of hard-coding a magic role in an opaque ability class. I'm thinking that we need a declaration of "magic roles" in the application: :admin, :superadmin, :approving, :depositing.

I also believe that there should be a method in the PermssionQuery that handles this query without requiring looping through.

app/models/concerns/hyrax/ability.rb
```diff
@@ -188,5 +193,15 @@ def user_is_depositor?(document_id)
...
+        return false unless approving_role
+        Hyrax::Workflow::PermissionQuery.scope_processing_agents_for(user: current_user).any? do |agent|
```
jeremyf: I also don't believe that we should be asking does the user have the role, but instead be asking what action are we trying to see if they can do?

jcoyne: @jeremyf the action they are trying to do is "go to the page that shows me the list of stuff I have to approve".

So, we have class of people who are responsible for doing approvals and we need to know if the current user is in that class.

app/models/concerns/hyrax/ability.rb
jeremyf: I've been contemplating the need to add MAGIC_ROLE_NAMES to Sipity::Role.

app/models/concerns/hyrax/ability.rb
jeremyf: I'm wondering why this is hidden for other people? Instead of asking do I have the "approving" role, which may not exist on some workflows, why not always render this page?

Anyone may have one or more roles for any of the submissions - they may have the approving role, or the metadata approving role, or copyright reviewing role. And shouldn't they, in all cases see the page for reviewing submissions?

jeremyf: As I've been thinking about this it sounds like we need the following (so as to avoid magic role names that may or may not exist):

This method should return true if the given user has one or more roles in a workflow that are at the workflow level (and not the work specific level).

However, consider if someone is granted permission to review a specific work, they would perhaps want a dashboard for review. I personally believe rethinking this UI as "My Tasks" is a better mechanism. Then any authenticated user has the page. In many cases it is empty. We could also query "How many tasks do I have to work on" and put a numeric badge by "My Tasks".

mjgiarlo:
@jeremyf Any chance you'll have time this month to take a look at this and help it along?

app/models/concerns/hyrax/ability.rb
mjgiarlo: @jeremyf I'm pushing a small change to this branch in hopes that we can get *something* merged in the short-term while we consider what you have suggested, which I wrote up here: https://github.com/projecthydra-labs/hyrax/issues/956

mjgiarlo:
@projecthydra-labs/hyrax-code-reviewers ?

jeremyf:
I'm looking at this, but I continue to be dissatisfied with the implementation, because there is now a requirement for a magic Sipity::Role, which smells bad, as Role names are arbitrary (e.g. you create them in the JSON that gets loaded for a workflow).

This can be an interstitial solution as the ability refactor discussion continues. I'm looking for time this month, and it is not looking likely.

I believe the more complete solution is:

- Do I have any workflow roles associated? If so, show the dashboard.
- On the dashboard, show all objects for which I have a workflow role
- On the dashboard, show links to review for any object in which I have an action that I can now take on that object

mjgiarlo:
:clap:, @jeremyf 

I've copied your suggested complete solution to the new ticket created to track this.

PR #776, opened 2017-04-12
Use a remote authority (geonames) for based_near

jcoyne:
Ref projecthydra-labs/hyku#878

### DIFF
--- a/app/indexers/hyrax/deep_indexing_service.rb
+++ b/app/indexers/hyrax/deep_indexing_service.rb
@@ -1,0 +1,102 @@
+require 'linkeddata' # we need all the linked data types, because we don't know what types a service might return.
+module Hyrax
+  class DeepIndexingService < BasicMetadataIndexer
+    # We're overiding the default indexer in order to index the RDF labels. In order
+    # for this to be called, you must specify at least one default indexer on the property.
+    # @param [Hash] solr_doc
+    # @param [String] solr_field_key
+    # @param [Hash] field_info
+    # @param [ActiveTriples::Resource, String] val
+    def append_to_solr_doc(solr_doc, solr_field_key, field_info, val)
+      return super unless object.controlled_properties.include?(solr_field_key.to_sym)
+      case val
+      when ActiveTriples::Resource
+        append_label_and_uri(solr_doc, solr_field_key, field_info, val)
+      when String
+        append_label(solr_doc, solr_field_key, field_info, val)
+      else
+        raise ArgumentError, "Can't handle #{val.class}"
+      end
+    end
+
+    def add_assertions(*)
+      fetch_external
+      super
+    end
+
+    private
+
+      # Grab the labels for controlled properties from the remote sources
+      def fetch_external
+        object.controlled_properties.each do |property|
+          object[property].each do |value|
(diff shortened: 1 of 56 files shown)


atz:
Weird CI is not being skipped here.

atz:
Does this really require a 5,500-line delta?  That seems huge.

jcoyne:
@atz most of that is the addition of handlebars (a JS template library)

jcoyne:
@dunn @no-reply @little9 @tpendragon @hackmastera Do any of you have an interest in reviewing this PR?

jcoyne:
Merging #893 and rebasing should fix the remaining test failure.

app/indexers/hyrax/deep_indexing_service.rb
```diff
@@ -0,0 +1,102 @@
...
+    # We're overiding the default indexer in order to index the RDF labels. In order
+    # for this to be called, you must specify at least one default indexer on the property.
+    # @param [Hash] solr_doc
+    # @param [String] solr_field_key
+    # @param [Hash] field_info
+    # @param [ActiveTriples::Resource, String] val
+    def append_to_solr_doc(solr_doc, solr_field_key, field_info, val)
+      return super unless object.controlled_properties.include?(solr_field_key.to_sym)
+      case val
+      when ActiveTriples::Resource
```
dunn: These checks break if you use the 
```ruby
class Thing
  include ActiveTriples::RDFSource
```
style instead of 
```ruby
class Thing < ActiveTriples::Resource
```
Apparently the former is now recommended?

app/inputs/controlled_vocabulary_input.rb
```diff
@@ -0,0 +1,83 @@
...
+      hidden_value = value.node? ? '' : value.rdf_subject
+      @builder.hidden_field(attribute_name, name: name, id: id, value: hidden_value, data: { id: 'remote' })
+    end
+
+    def build_options(value, index, options)
+      options[:name] = name_for(attribute_name, index, 'hidden_label')
+      options[:data] ||= {}
+      options[:data][:attribute] = attribute_name
+      options[:id] = id_for_hidden_label(index)
+      if value.node?
```
dunn: I'm in the middle of upgrading ADRL to use the RDF 2.0 gems and the new ActiveTriples is causing problems with its pseudo-arrays.  What I'm trying here now is
```ruby
    def build_field(value, index)
      options = input_html_options.dup
      value = value.resource if value.is_a? ActiveFedora::Base

      if value.respond_to?(:empty) && value.empty?
        build_options_for_new_row(attribute_name, index, options)
      else
        options[:name] = name_for(attribute_name, index, "hidden_label")
        options[:data] = { attribute: attribute_name }
        options[:id] = id_for_hidden_label(index)

        build_options_for_existing_row(
          attribute_name,
          index,
          (value.respond_to?(:rdf_label) ? value.rdf_label : value),
          options
        )
      end
```
If you want to take a closer look I can get you access to our repo.

app/indexers/hyrax/deep_indexing_service.rb
jcoyne: I wasn't aware there were two ways.  @no-reply what was is correct?

PR #777, opened 2017-04-13
Relax redlock dependency to allow for 0.2.0

jcoyne:


hyrax.gemspec
```diff
@@ -68,7 +68,7 @@ EOF
   spec.add_dependency 'dry-equalizer', '~> 0.2'
   spec.add_dependency 'dry-struct', '~> 0.1'
   spec.add_dependency 'active_attr', '~> 0.9.0'
-  spec.add_dependency 'redlock', '~> 0.1.2'
+  spec.add_dependency 'redlock', '>= 0.1.2'
```
mjgiarlo: Might `'>= 0.1.2', '~> 0.1'` be preferable, to prevent 1.0.0 from rolling in and potentially breaking the build unpredictably?

PR #803, opened 2017-04-19
Add migration task to switch direction of collection membership

hackmastera:
for upgrade path from sufia 7.3
rework of pr #757

@atz is this what you had in mind?

### DIFF
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -7,7 +7,7 @@ namespace :hyrax do
     desc "Move membership from collection#members work#member_of_collections"
     task collections: :environment do
       require 'hyrax/collections_migration'
-      CollectionsMigration.run
+      Hyrax::CollectionsMigration.run
     end
   end
 end
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -4,5 +4,10 @@ namespace :hyrax do
       require 'hyrax/move_all_works_to_admin_set'
       MoveAllWorksToAdminSet.run(AdminSet.find(AdminSet::DEFAULT_ID))
     end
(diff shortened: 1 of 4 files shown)


lib/tasks/migrate.rake
```diff
@@ -4,5 +4,10 @@ namespace :hyrax do
       require 'hyrax/move_all_works_to_admin_set'
       MoveAllWorksToAdminSet.run(AdminSet.find(AdminSet::DEFAULT_ID))
     end
+    desc "Move membership from collection#members work#member_of_collections"
+    task collections: :environment do
+      require 'hyrax/collections_migration'
+      CollectionsMigration.run
```
carrickr: In testing, we found this needed to be 

`Hyrax::CollectionsMigration.run`

carrickr:
Just for reference, we ran this script on a Sufia 7.3 to Hyrax RC 1 upgrade and it was success.  All data was preserved, permissions, etc.  We did somehow lose thumbnails, but we weren't able to narrow that down to this script vs some weirdness with the `/tmp` dir on the developer's machine.  If other people lose thumbnails it might be an issue.  We'll be repeating the migration on our staging server in the near future and see if we can replicate that bug there or not.

spec/lib/hyrax/collections_migration_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'spec_helper'
```
atz: Minor detail, fyi: because the require is already done in `.rspec`, this line is unnecessary here.

hackmastera: Hm, other specs have it. I'll try it without!

hackmastera:
@carrickr Awesome, thank you for testing!
Thanks @carrickr and @atz for the review.

PR #804, opened 2017-04-19
Replace hardcoded CarrierWave::Storage::Fog::File that causes crashes

### LINKED ISSUE #802: Fog reference causes exception at runtime
Fog is not a dependency of Hyrax, but we try to sniff for it anyway:
https://github.com/projecthydra-labs/hyrax/blame/v1.0.0.rc1/app/jobs/attach_files_to_work_job.rb#L29

This causes exceptions to be thrown at job runtime:
https://app.honeybadger.io/projects/48327/faults/33327615#notice-summary

The error would have been discoverable with tests, but the testing arbitrarily defined the namespace in question (something that doesn't happen in production!):
https://github.com/projecthydra-labs/hyrax/blob/v1.0.0.rc1/spec/jobs/attach_files_to_work_job_spec.rb#L28-L32

atz:
Fixes #802

- Hyrax should not have to sniff classes that are not included in its dependencies.
- Moved `import_url` to the `FileSetActor` class, since every line was about the internals of the class.  Now it can just take one URL-like argument.  
- BUT, that made rubocop complain that the class was too long, so I tightened up some whitespace, comments and phrasing.  
- Important improvement to the test, by removing the ad hoc declaration of `CarrierWave::Storage::Fog::File` namespace, something that this gem does not require at runtime.

### DIFF
--- a/app/jobs/attach_files_to_work_job.rb
+++ b/app/jobs/attach_files_to_work_job.rb
@@ -21,24 +21,14 @@ class AttachFilesToWorkJob < ActiveJob::Base
   private
 
     # @param [Hyrax::Actors::FileSetActor] actor
-    # @param [UploadedFileUploader] file
+    # @param [Hyrax::UploadedFileUploader] file file.file must be a CarrierWave::SanitizedFile or file.url must be present
     def attach_content(actor, file)
-      case file.file
-      when CarrierWave::SanitizedFile
+      if file.file.is_a? CarrierWave::SanitizedFile
         actor.create_content(file.file.to_file)
-      when CarrierWave::Storage::Fog::File
-        import_url(actor, file)
+      elsif file.url.present?
+        actor.import_url(file.url)
       else
-        raise ArgumentError, "Unknown type of file #{file.class}"
+        raise ArgumentError, "#{file.class} received with #{file.file.class} object and no URL"
       end
-    end
-
-    # @param [Hyrax::Actors::FileSetActor] actor
-    # @param [UploadedFileUploader] file
-    def import_url(actor, file)
-      actor.file_set.update(import_url: file.url)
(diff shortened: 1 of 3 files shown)


app/jobs/attach_files_to_work_job.rb
```diff
@@ -23,22 +23,12 @@ def perform(work, uploaded_files)
     # @param [Hyrax::Actors::FileSetActor] actor
     # @param [UploadedFileUploader] file
     def attach_content(actor, file)
-      case file.file
-      when CarrierWave::SanitizedFile
+      if file.file.is_a? CarrierWave::SanitizedFile
         actor.create_content(file.file.to_file)
-      when CarrierWave::Storage::Fog::File
-        import_url(actor, file)
+      elsif file.url.present?
```
mjgiarlo: Does this need to be:

```ruby
elsif file.respond_to?(:url) && file.url.present?
```

?

mjgiarlo:
Thanks for taking this on, @atz.

app/jobs/attach_files_to_work_job.rb
atz: No.  I was confused by that at first also.  

`file` is of our `Hyrax::UploadedFileUploader` class, and that inherits from `CarrierWave::Uploader::Base`, so `file.respond_to?(:url)` is always true.  This is easy to demonstrate:
```ruby
> z=CarrierWave::Uploader::Base.new
 => #<CarrierWave::Uploader::Base:0x007f9de4783028 @model=nil, @mounted_as=nil> 
> z.respond_to?(:url)
 => true 
```

That is not to be confused with `file.file`, of course, which may be different things.

atz:
@mjgiarlo ready

mjgiarlo:
Thx, @atz!

PR #805, opened 2017-04-19
Add inflection to test app generator to "fix" Atlas paths

### LINKED ISSUE #693: Generator is creating bad routes for atlas
### Descriptive summary

In Master when you create the test app it creates bad routes for Atlas.  If I had to hazard a guess the plural to singular is causing the atla instead of the atlas.  I do not see the same behavior when I generate a class not ending in `s`.

screen shot 2017-04-03 at 2 00 43 pm

### Expected behavior

should get routes for atlas

### Actual behavior

get routes for atla

1. run rake engin_cart:generate
1. cd .internal_test_app
1. run rake routes

### Related work

mjgiarlo:
Also, clarified this in the documentation: https://github.com/projecthydra/sufia/wiki/Customizing-your-work-types

Fixes #693

### DIFF
--- a/spec/test_app_templates/lib/generators/test_app_generator.rb
+++ b/spec/test_app_templates/lib/generators/test_app_generator.rb
@@ -10,9 +10,19 @@ class TestAppGenerator < Rails::Generators::Base
   def create_generic_work
     generate 'hyrax:work GenericWork'
   end
-  
+
   def create_atlas_work
     generate 'hyrax:work RareBooks/Atlas'
+    # ActiveSupport interprets "atlas" as plural which causes
+    # counter-intuitive route paths. Add an inflection to correct
+    # these paths
+    append_file 'config/initializers/inflections.rb' do
+      <<~HEREDOC
+        ActiveSupport::Inflector.inflections(:en) do |inflect|
+          inflect.irregular 'atlas', 'atlases'
+        end
+      HEREDOC
+    end
   end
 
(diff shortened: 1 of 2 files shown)


spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -10,9 +10,16 @@ def install_engine
   def create_generic_work
     generate 'hyrax:work GenericWork'
   end
-  
+
   def create_atlas_work
     generate 'hyrax:work RareBooks/Atlas'
+    append_file 'config/initializers/inflections.rb' do
+      <<~HEREDOC
```
cam156: @mjgiarlo for future people look at this code can you add a comment that this was done becuase it created a bad plural?

mjgiarlo:
@cam156 There's only one "broken" set of paths that the test app sets up, and that's for atlases. We can't anticipate every single irregular noun and ship all those inflections. Rails has a pattern for handling this, which we now link to from our documentation: https://github.com/projecthydra/sufia/wiki/Customizing-your-work-types

spec/test_app_templates/lib/generators/test_app_generator.rb
mjgiarlo: Done!

atz:
One half of CI failed, possibly on an intermittent error.  Restarted job.

PR #817, opened 2017-04-21
Update Rubocop to use version of Ruby we test with

mjgiarlo:


.rubocop.yml
```diff
@@ -1,7 +1,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
-  TargetRubyVersion: 2.1
```
jcoyne: This is the language spec version. Which is concident, but not the same as the implementation version.  MRI 2.4 supports the language spec for 2.1 (because it's backwards compatible).  This check shows that we aren't using any language constructs newer than 2.1. I belive this is a good thing as people may not be deploying on 2.4.  If we want to start using 2.4 language spec, we should have a discussion about that separately.

app/controllers/concerns/hyrax/download_behavior.rb
```diff
@@ -69,7 +71,7 @@ def default_file
                                    DownloadsController.default_content_path
                                  end
         association = dereference_file(default_file_reference)
-        association.reader if association
+        association&.reader
```
jcoyne: This change means that no one could use hyrax without using ruby 2.3+

PR #819, opened 2017-04-21
Warn user if Rails interprets work type as plural

### LINKED ISSUE #693: Generator is creating bad routes for atlas

mjgiarlo:
Follow-on to #805 

Fixes #693 (for reals)

### DIFF
--- a/lib/generators/hyrax/work/work_generator.rb
+++ b/lib/generators/hyrax/work/work_generator.rb
@@ -1,6 +1,8 @@
 require 'rails/generators'
 
 class Hyrax::WorkGenerator < Rails::Generators::NamedBase
+  include Rails::Generators::ModelHelpers
+
   source_root File.expand_path('../templates', __FILE__)
 
   argument :attributes, type: :array, default: [], banner: 'field:type field:type'
--- a/lib/generators/hyrax/work/work_generator.rb
+++ b/lib/generators/hyrax/work/work_generator.rb
@@ -1,6 +1,10 @@
 require 'rails/generators'
 
 class Hyrax::WorkGenerator < Rails::Generators::NamedBase
+  # ActiveSupport can interpret models as plural which causes
+  # counter-intuitive route paths. Pull in ModelHelpers from
+  # Rails which warns users about pluralization when generating
+  # new models or scaffolds.
   include Rails::Generators::ModelHelpers
 
   source_root File.expand_path('../templates', __FILE__)
--- a/lib/generators/hyrax/work/work_generator.rb
+++ b/lib/generators/hyrax/work/work_generator.rb
@@ -1,4 +1,5 @@
 require 'rails/generators'
+require 'rails/generators/model_helpers'
 
 class Hyrax::WorkGenerator < Rails::Generators::NamedBase
   # ActiveSupport can interpret models as plural which causes
--- a/spec/test_app_templates/lib/generators/test_app_generator.rb
+++ b/spec/test_app_templates/lib/generators/test_app_generator.rb
@@ -12,17 +12,15 @@ class TestAppGenerator < Rails::Generators::Base
   end
 
   def create_atlas_work
-    generate 'hyrax:work RareBooks/Atlas'
     # ActiveSupport interprets "atlas" as plural which causes
     # counter-intuitive route paths. Add an inflection to correct
     # these paths
     append_file 'config/initializers/inflections.rb' do
-      <<~HEREDOC
-        ActiveSupport::Inflector.inflections(:en) do |inflect|
-          inflect.irregular 'atlas', 'atlases'
-        end
-      HEREDOC
+      "ActiveSupport::Inflector.inflections(:en) do |inflect|\n" \
+      "  inflect.irregular 'atlas', 'atlases'\n" \
+      "end\n"
     end
+    generate 'hyrax:work RareBooks/Atlas'
   end
 
   def comment_out_web_console
--- a/lib/generators/hyrax/install_generator.rb
+++ b/lib/generators/hyrax/install_generator.rb
@@ -29,11 +29,10 @@ module Hyrax
          """
 
     def run_required_generators
-      say_status('warning', '[Hyrax] GENERATING BLACKLIGHT', :yellow)
+      say_status('info', '[Hyrax] GENERATING BLACKLIGHT', :blue)
       generate 'blacklight:install --devise'
-      say_status('warning', '[Hyrax] GENERATING HYDRA-HEAD', :yellow)
+      say_status('info', '[Hyrax] GENERATING HYDRA-HEAD', :blue)
       generate 'hydra:head -f'
-      say_status('warning', '[Hyrax] GENERATING MODELS', :yellow)
       generate "hyrax:models#{options[:force] ? ' -f' : ''}"
     end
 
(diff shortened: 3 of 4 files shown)


lib/generators/hyrax/work/work_generator.rb
```diff
@@ -1,6 +1,9 @@
 require 'rails/generators'
 
 class Hyrax::WorkGenerator < Rails::Generators::NamedBase
+  PLURAL_MODEL_NAME_WARN_MESSAGE = "[EXITING] The work type name '%s' was recognized as a plural, using the singular '%s' instead. " \
```
cam156: Should be I18N since we are supporting multiple languages

mjgiarlo:
@cam156 Hrm. How do you set the locale when executing a generator? FWIW, Rails itself doesn't use I18n here.

lib/generators/hyrax/work/work_generator.rb
atz: I don't think so, since the plural-recognition is language-based.  To me, we should just be avoiding building anything that attempts to solve this problem.  It is not tractable and it is not our domain space.

cam156:
@mjgiarlo good point!  Quick google search does not yield an answer.

atz:
I don't think this is a wise direction to follow.  If you want to provide functionality that covers all cases, have the generator accept additional optional argument(s) to specify singular and plural forms, passing those to rails as needed.  

We are certainly not solving this for all words in all languages.

mjgiarlo:
@atz No, we're not, but we are reusing a related Rails generator behavior in our own generator behavior.

mjgiarlo:
Reusing this Rails logic in our work generator helps us catch when a user's expectation ("'lens' is singular!") doesn't match what Rails determines ("'lens' is plural"). It isn't perfect but this change at least makes our generator as good as Rails' generators.

mjgiarlo:
Will address the Travis failures ​on Monday.

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -15,6 +18,25 @@ def banner
...
+    # generated before this new logic was added
+    return if revoking?
+
+    # Mimic logic from Rails:
+    # https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/railties/lib/rails/generators/model_helpers.rb#L16
+    # To prevent problems with types named, e.g., "Atlas" or "Lens"
+    #
+    if file_name == plural_file_name && file_name.singularize != file_name
+      say_status("error", format(PLURAL_MODEL_NAME_WARN_MESSAGE, file_name, file_name.singularize), :red)
+      exit
```
atz: You want to unilaterally `exit`?  When this engine (as seems inevitable in our stack) is called through a chain of 3 other engines, you want to **prevent** them from being able to swallow an exception and keep on going?

jcoyne:
I'm with Joe on this.  This is a rails issue.  We are handling it the same way it would be handled if you did `rails generate model Atlas`

mjgiarlo:
@jcoyne @atz When you generate a model named `Atlas` in Rails, it does this:

```bash
[WARNING] The model name 'Atlas' was recognized as a plural, using the singular 'Atla' instead. Override with --force-plural or setup custom inflection rules for this noun before running the generator.
      invoke  active_record
      create    db/migrate/20170424165921_create_atlas.rb
      create    app/models/atla.rb
      invoke    rspec
      create      spec/models/atla_spec.rb
      invoke      factory_girl
      create        spec/factories/atlas.rb
```

We're not currently providing that warning with our generator and I suspect that merely doing that (and getting rid of the admittedly heavy-handed `exit`) would be an improvement over what we have now, and enough to fix #693. If I remove the `exit`, are you cool with this PR going forward?

jcoyne:
Can we just include this mixin? https://github.com/rails/rails/blob/5-0-stable/railties/lib/rails/generators/model_helpers.rb

mjgiarlo:
@jcoyne Good call. Now it does this:

```bash
$ rails g hyrax:work Lens
[WARNING] The model name 'Lens' was recognized as a plural, using the singular 'Len' instead. Override with --force-plural or setup custom inflection rules for this noun before running the generator.
        info  GENERATING WORK MODEL: Len
      create  app/actors/hyrax/actors/len_actor.rb
      create  app/controllers/hyrax/lens_controller.rb
      create  app/indexers/len_indexer.rb
      create  app/forms/hyrax/len_form.rb
      create  app/models/len.rb
      create  app/views/hyrax/lens/_len.html.erb
...

$ rails g hyrax:work Lens --force-plural
        info  GENERATING WORK MODEL: Lens
   identical  app/actors/hyrax/actors/lens_actor.rb
      create  app/controllers/hyrax/lens_controller.rb
   identical  app/indexers/lens_indexer.rb
   identical  app/forms/hyrax/lens_form.rb
   identical  app/models/lens.rb
   identical  app/views/hyrax/lens/_lens.html.erb
...
```

mjgiarlo:
@cam156 :point_up: that OK?

mjgiarlo:
@jcoyne @atz re-review?

mjgiarlo:
Once merged, this should be backported to `1-0-stable`.

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails/generators'
 
 class Hyrax::WorkGenerator < Rails::Generators::NamedBase
+  include Rails::Generators::ModelHelpers
```
jcoyne: @mjgiarlo how about adding a comment here explaining why we're including this.

spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -12,17 +12,15 @@ def create_generic_work
...
     # ActiveSupport interprets "atlas" as plural which causes
     # counter-intuitive route paths. Add an inflection to correct
     # these paths
     append_file 'config/initializers/inflections.rb' do
-      <<~HEREDOC
-        ActiveSupport::Inflector.inflections(:en) do |inflect|
-          inflect.irregular 'atlas', 'atlases'
-        end
-      HEREDOC
+      "ActiveSupport::Inflector.inflections(:en) do |inflect|\n" \
```
mjgiarlo: Made this change so that this works in Rubies < 2.3.

spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -12,17 +12,15 @@ def create_generic_work
...
+      "  inflect.irregular 'atlas', 'atlases'\n" \
+      "end\n"
     end
+    generate 'hyrax:work RareBooks/Atlas'
```
mjgiarlo: Also, setting up the inflections *after* generating the model was kinda silly. Caused test failures.

lib/generators/hyrax/work/work_generator.rb
mjgiarlo: Got it. Take another look?

PR #828, opened 2017-04-22
adds json serialization for admin sets listing

### LINKED ISSUE #827: AdminSetsController isn't able to render #index as json
Our work developing a DSpace to Hydra migration application could benefit from the ability to check if an admin set exists prior to beginning the process of publishing a work (first uploading files, then associating them to the new work, and finally publishing the work). 

The most direct approach would be to query a list of the admin sets on the server to verify the existence of the admin sets that would be used during the migration processing.

Without this jbuilder view, a default blacklight catalog view is failing to render because it's aware of documents and blacklight related things (not the controller variable admin_sets).

revgum:
fixes #827

### DIFF
--- a/spec/views/hyrax/admin/admin_sets/index.json.erb_spec.rb
+++ b/spec/views/hyrax/admin/admin_sets/index.json.erb_spec.rb
@@ -1,0 +1,29 @@
+RSpec.describe "hyrax/admin/admin_sets/index.json.jbuilder" do
+  context "when no admin sets exists" do
+    it "renders no admin sets" do
+      render
+      json = JSON.parse(rendered)
+      expect(json['admin_sets']).to eq []
+    end
+  end
+
+  context "when an admin set exists" do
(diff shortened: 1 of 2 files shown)


spec/views/hyrax/admin/admin_sets/index.json.erb_spec.rb
```diff
@@ -0,0 +1,29 @@
+RSpec.describe "hyrax/admin/admin_sets/index.json.jbuilder" do
+  context "when no admin sets exists" do
+    it "renders no admin sets" do
+      render
```
jcoyne: Just a style suggestion, but you could have put the `render` in a `before` block and the `json` in a `let` block. then only assertions are in the `it` block.

revgum: good call! Thanks @jcoyne

PR #869, opened 2017-05-02
Update to blacklight 6.9.0

### LINKED ISSUE #546: Blacklight::SearchState and Hyrax::SearchState have incompatible initialize method signatures
In Blacklight 6.8.0 we have the following:

```ruby
    def params_for_search(params_to_merge={}, &block)
      # params hash we'll return
      my_params = params.dup.merge(self.class.new(params_to_merge, blacklight_config))
      # MORE STUFF HERE …
```

https://github.com/projectblacklight/blacklight/blob/v6.8.0/lib/blacklight/search_state.rb#L123

In Hyrax the SearchState uses the following initialize signature:

```ruby
module Hyrax
  class SearchState < Blacklight::SearchState
    include ActionDispatch::Routing::RouteSet::MountedHelpers

    def initialize(view_context)
      super(view_context.params, view_context.blacklight_config)
      @view_context = view_context
    end
  end
end
```
https://github.com/projecthydra-labs/hyrax/blob/a14f72e6073ac0270980060880cb3bd105c9e746/lib/hyrax/search_state.rb#L5

The problem arises when we have a `Hyrax::SearchState` object, and call `#params_for_search`. This method is defined on `Blacklight::SearchState` (see above). When called it reinstantiates the Hyrax::SearchState but passes a hash and a blacklight_config instead of the view_context.

Short term fix:

* Freeze Blacklight to ~> 6.7.3 (as the error was introduced in Blacklight 6.8.0).

Mid-fix:

* Create an alternate initialize method for Hyrax::SearchState that is used within Hyrax (however, this information is not passed along).

jcoyne:
Fixes #546

### DIFF
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -4,6 +4,8 @@ module Hyrax::Controller
   included do
     class_attribute :create_work_presenter_class
     self.create_work_presenter_class = Hyrax::SelectTypeListPresenter
+    self.search_state_class = Hyrax::SearchState
+
     # Adds Hydra behaviors into the application controller
     include Hydra::Controller::ControllerBehavior
     helper_method :create_work_presenter
@@ -24,11 +26,6 @@ module Hyrax::Controller
   # Ensure that the locale choice is persistent across requests
   def default_url_options
     super.merge(locale: I18n.locale)
-  end
-
-  # Override Blacklight to use the Hyrax::SearchState
-  def search_state
-    @search_state ||= Hyrax::SearchState.new(self)
   end
 
   private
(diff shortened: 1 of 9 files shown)


app/controllers/concerns/hyrax/controller.rb
```diff
@@ -26,11 +28,6 @@ def default_url_options
     super.merge(locale: I18n.locale)
   end
 
-  # Override Blacklight to use the Hyrax::SearchState
```
mjgiarlo: Does removing a public method mean this PR would force a major version bump?

jcoyne: This method already exists in Blacklight, so we're not removing a public method, just this implementation of it.

PR #871, opened 2017-05-02
Fix documentation links

mjgiarlo:


.gitignore
```diff
@@ -1,3 +1,8 @@
 _site/
 .sass-cache/
 .ruby-*
+.byebug_history
+.internal_test_app
+Gemfile.lock
+fcrepo4-test-data
+spec
```
atz: I agree w/ the rest of these, but how does `spec` make sense?

mjgiarlo: This is for the `gh-pages` branch, @atz.

mjgiarlo: Re-review, @atz?

PR #888, opened 2017-05-04
Pass the correct attribute to sharingTabVisibility.

### LINKED ISSUE #884: Share tab disappearing
### Descriptive summary

targeting Hyrax v1.0.0

### Rationale

> Also, please see @hortongn's comment on this same thread from 4/24.  We found (with @hackmastera's help) that the `allow_access_grant` workflow setting would make the share tab show up for both work creation and editing, for someone depositing to the default admin set with default workflow, in Sufia 7.3.  We assume, but haven't tested yet, if the same workflow setting solves the same problem in Hyrax 1.0.

### Related work

mjgiarlo:
Fixes #884 

Backport of #719


mjgiarlo:
Blocked by https://github.com/projecthydra-labs/hyrax/pull/889

PR #891, opened 2017-05-04
Only use one copy of jquery

jcoyne:
including `browse_everything` pulls in a second copy of jquery.  This
makes it impossible to switch the local application to use a different
version of jquery (2 or 3) than what browse_everything is using.


atz:
https://travis-ci.org/projecthydra-labs/hyrax/builds/228820204#L2387-L2388
```
ActiveRecord::RecordNotFound: Couldn't find Hyrax::PermissionTemplate
```

PR #893, opened 2017-05-04
Load workflows for tests that require it

jcoyne:


spec/spec_helper.rb
```diff
@@ -168,6 +168,8 @@ def main_app
   end
 
   config.before :each do |example|
+    # using :workflow is preferable to :clean_repo, use the former if possible
+    insure_deposit_available_for(user) if example.metadata[:workflow]
```
mjgiarlo: `s/insure/ensure/`

PR #894, opened 2017-05-04
Expect at least the created expired leases

jcoyne:
There could be other expired leases from other test runs

### DIFF
--- a/spec/features/admin_admin_set_spec.rb
+++ b/spec/features/admin_admin_set_spec.rb
@@ -2,8 +2,9 @@ require 'spec_helper'
 
 RSpec.describe "The admin sets, through the admin dashboard" do
   let(:user) { create :admin }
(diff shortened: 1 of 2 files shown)


spec/features/admin_admin_set_spec.rb
```diff
@@ -15,10 +16,7 @@
     login_as(user, scope: :user)
     visit '/dashboard'
     click_link "Administrative Sets"
-
-    expect(page).to have_link "Create new administrative set"
```
mjgiarlo: No longer an expectation?

jcoyne: The next line says to click that link. If it wasn't present the "click_link" would fail. So this is a duplicate assertion.

mjgiarlo: :shipit:

PR #907, opened 2017-05-05
Adds the rights statement to the show page

### LINKED ISSUE #886: Right statement not showing up on show page
One of the pieces of metadata that may be quite important for a user to look at is the right statement. Currently, the rights statement is not showing up in the show page. 

One question would be, does hyrax have a usecase for showing rights  statements, would it be useful to the user? I have a fix locally which includes adding 

```
<%= presenter.attribute_to_html(:rights_statement) %>
```
to the `attribute_rows.html.erb` file then adding

```
attribute :rights_statement, Solr::Array, solr_name('rights_statement')
```
to metadata.rb in app/models/concerns/hyrax/solr_document

Then adding the `:rights_statement` symbol to the list of delegated methods in the `WorkShowPresenter`

I think this would fix it in a way that would allow people to see the rights statement in the show page.

straleyb:
Fixes #886 

This gets the rights statement showing up on the show page. Very basic and mvp. Discussion about whether people want external links and what not to show up can happen, but this is the most basic fix for this issue.

### DIFF
--- a/app/views/hyrax/base/_attribute_rows.html.erb
+++ b/app/views/hyrax/base/_attribute_rows.html.erb
@@ -10,3 +10,4 @@
 <%= presenter.attribute_to_html(:related_url, render_as: :external_link) %>
 <%= presenter.attribute_to_html(:resource_type, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:source) %>
+<%= presenter.attribute_to_html(:rights_statement) %>
--- a/spec/models/solr_document_spec.rb
+++ b/spec/models/solr_document_spec.rb
@@ -26,6 +26,17 @@ RSpec.describe ::SolrDocument, type: :model do
     end
   end
 
+  describe "rights_statement" do
+    let(:attributes) { { 'rights_statement_tesim' => ['A rights statement'] } }
+
+    it "responds to rights_statement" do
+      expect(document).to respond_to(:rights_statement)
+    end
+    it "returns the proper data" do
+      expect(document.rights_statement).to eq ['A rights statement']
+    end
+  end
+
   describe "create_date" do
     let(:attributes) { { 'system_create_dtsi' => '2013-03-14T00:00:00Z' } }
     subject { document.create_date }
--- a/app/models/concerns/hyrax/solr_document/metadata.rb
+++ b/app/models/concerns/hyrax/solr_document/metadata.rb
@@ -64,6 +64,7 @@ module Hyrax
         attribute :license, Solr::Array, solr_name('license')
         attribute :source, Solr::Array, solr_name('source')
         attribute :date_created, Solr::Array, solr_name('date_created')
+        attribute :rights_statement, Solr::Array, solr_name('rights_statement')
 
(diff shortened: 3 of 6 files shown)


straleyb:
I was looking in the solr_document about where I would test it, but it doesnt seem like most of the attributes have a test. At least from what I could tell.

straleyb:
@jcoyne I added a test that verifies that `SolrDocument` Responds to the proper method. Should I also check its return value?

spec/models/solr_document_spec.rb
```diff
@@ -26,6 +26,14 @@
     end
   end
 
+  describe "rights_statement" do
+    let(:attributes) { { 'rights_statement_sim' => ['A rights statement'] } }
```
jcoyne: I think the field you want is `rights_statement_tesim`, and so maybe it would be a good idea to check the return value.

straleyb:
Updated to check return value @jcoyne. 👍

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -10,3 +10,4 @@
 <%= presenter.attribute_to_html(:related_url, render_as: :external_link) %>
 <%= presenter.attribute_to_html(:resource_type, render_as: :faceted) %>
 <%= presenter.attribute_to_html(:source) %>
+<%= presenter.attribute_to_html(:rights_statement) %>
```
mjgiarlo: How is this rendering by default? I wonder if we need to handle rights_statements similar to how we handle licenses, using a new `:rights_statement` renderer? Here's what we use for licenses: https://github.com/projecthydra-labs/hyrax/blob/master/app/renderers/hyrax/renderers/license_attribute_renderer.rb

mjgiarlo: This renderer uses the [license service](https://github.com/projecthydra-labs/hyrax/blob/master/app/services/hyrax/license_service.rb), and we can hook a new renderer up to the existing [rights statement service](https://github.com/projecthydra-labs/hyrax/blob/master/app/services/hyrax/rights_statements.rb).

straleyb: It renders the URI for the rights statement. So something like `https://rightstatements.org/2.0/blah` and that links out to the actual copyrights page. I was looking into the renderers and was wondering if thats how you would like it implemented. With that being said, im not going to have time to work on that architecture. If you all want those to be put in place, maybe it could be made into a ticket as an enhancement?

straleyb:
@mjgiarlo @jcoyne Just wanted to revisit this PR. It's holding up some tickets in SA and was just curious if we wanted to merge it in.

mjgiarlo:
@straleyb Could you include a view spec to cover the change you added there? Then I'll be happy to merge. Thank you, sir.

mjgiarlo:
@straleyb Tossed a view spec up in this branch. Will look to merge later.

mjgiarlo:
@straleyb feel free to merge once this goes :green_heart: 

Thanks for the contribution!

PR #915, opened 2017-05-08
Grant all workflow roles to admin set creator.

### LINKED ISSUE #914: When I flip an admin sets workflow to "mediated_deposit" the admin users are not given the "approving" role
1. Create an new admin set
1. Flip to the "Workflow" tab
1. Select "One-step mediated deposit workflow"
1. Save the admin set.
1. Click "Workflow Roles" in the sidebar. You will see that the current user has no new roles.

Note. If you add the manager to the admin set after making the workflow mediated deposit, then the workflow role is added correctly.

mjgiarlo:
Fixes #914.

### DIFF
--- a/app/services/hyrax/admin_set_create_service.rb
+++ b/app/services/hyrax/admin_set_create_service.rb
@@ -80,7 +80,19 @@ module Hyrax
 
       def create_workflows_for(permission_template:)
         workflow_importer.call(permission_template: permission_template)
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@jcoyne @jeremyf Your reviews would be most welcome. (I don't like the `allow_any` either.)

jcoyne:
LGTM, please remove the `[WIP]` commit message when you're done.

mjgiarlo:
@jcoyne WIP removed.

PR #916, opened 2017-05-08
Generate the presenter into the local application

jcoyne:
This makes it easier to add custom behavior to the local application in
the correct place. This follows the behavior of the form object.

### DIFF
--- a/lib/generators/hyrax/work/work_generator.rb
+++ b/lib/generators/hyrax/work/work_generator.rb
@@ -38,6 +38,10 @@ class Hyrax::WorkGenerator < Rails::Generators::NamedBase
     template('form.rb.erb', File.join('app/forms/hyrax', class_path, "#{file_name}_form.rb"))
   end
 
+  def create_presenter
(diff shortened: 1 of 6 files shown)


mjgiarlo:
@jcoyne Should we add a note to the release notes that says "run the work generator again to generate presenter/presenter spec" ?

jcoyne:
@mjgiarlo that's not necessary. Things will continue to work just fine without a generated presenter. This just makes it easier on newcomers.

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -91,6 +95,11 @@ def create_form_spec
     template('form_spec.rb.erb', File.join('spec/forms/hyrax/', class_path, "#{file_name}_form_spec.rb"))
   end
 
+  def presenter_spec
+    return unless rspec_installed?
+    template('presenter_spec.rb.erb', File.join('spec/presenters/hyrax/', class_path, "#{file_name}_form_spec.rb"))
```
atz: You want `form` in the spec name for the presenter?

mjgiarlo: Good :eye:

PR #917, opened 2017-05-08
Update travis build matrix

cbeer:


atz:
Dunno why coveralls has been hanging up on reporting back green.

PR #921, opened 2017-05-09
Update RSolr exception class stubbing for rsolr 2.0

cbeer:


cbeer:
I'd also entertain objections that this is an upstream problem in Blacklight with RSolr 2.x. I'm not completely convinced the distinction rsolr 1.x made between socket errors and http errors is all that useful, though.

app/controllers/concerns/hyrax/homepage_controller_behavior.rb
```diff
@@ -35,14 +35,14 @@ def fetch_admin_sets
                                             .rows(5)
       response = repository.search(builder)
       response.documents
-    rescue Blacklight::Exceptions::ECONNREFUSED
+    rescue Blacklight::Exceptions::ECONNREFUSED, Blacklight::Exceptions::InvalidRequest
```
jcoyne: Can the existing error go away if we use RSolr 2?  If so, can you add a comment about this.

cbeer: Depends whether we try to fix this in Blacklight or not.

cbeer:
This should fix the RSolr regression: https://github.com/rsolr/rsolr/pull/176

PR #929, opened 2017-05-09
views: fix Flipflop typo

dunn:
flipflipflip

### DIFF
--- a/app/views/hyrax/my/works/index.html.erb
+++ b/app/views/hyrax/my/works/index.html.erb
@@ -29,7 +29,7 @@
               class: 'btn btn-primary'
             ) %>
       <% else # simple link to the first work type %>
-        <% if Flipflip.enable_batch_upload? %>
+        <% if Flipflop.enable_batch_upload? %>
           <%= link_to(
               t(:'helpers.action.batch.new'),
               hyrax.new_batch_upload_path(payload_concern: create_work_presenter.first_model),
--- a/spec/views/hyrax/my/works/index.html.erb_spec.rb
+++ b/spec/views/hyrax/my/works/index.html.erb_spec.rb
@@ -1,8 +1,6 @@
 require 'spec_helper'
 
 RSpec.describe 'hyrax/my/works/index.html.erb', type: :view do
-  let(:presenter) { instance_double(Hyrax::SelectTypeListPresenter, many?: true) }
-
   before do
     allow(view).to receive(:current_ability).and_return(ability)
     allow(view).to receive(:provide).and_yield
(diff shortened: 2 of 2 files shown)


mjgiarlo:
Yikes! Would you mind tossing up a view spec that fails without this change (like I should have done earlier :disappointed_relieved:).

dunn:
Would that mean a feature test to load that view?

mjgiarlo:
@dunn You can do this with a view spec -- even faster than a feature spec. I was thinking it'd be a small change to this file: https://github.com/projecthydra-labs/hyrax/blob/master/spec/views/hyrax/my/works/index.html.erb_spec.rb

Thanks! Btw, you're now :100: licensed.

dunn:
Ah, I see what you mean; I was confused because I didn't realize that constant isn't evaluated if that fork of the conditional isn't reached.

I figure the presenter double needs to be something like this for the new test, but I'm not sure what model would work:
```ruby
      let(:presenter) do
        instance_double(
          Hyrax::SelectTypeListPresenter,
          many?: false,
          first_model: AdminSet
        )
      end
```
That and the other few I tried lead to this failure:
```
  1) hyrax/my/works/index.html.erb when the user can add works without many presenters the line item displays the work and its actions
     Failure/Error: new_polymorphic_path([main_app, create_work_presenter.first_model]),
     
     ActionView::Template::Error:
       undefined method `new_admin_set_path' for #<ActionDispatch::Routing::RoutesProxy:0x007fb85fbc2568>
     # ./app/views/hyrax/my/works/index.html.erb:41:in `block in ___sers_alex_clones_hyrax_app_views_hyrax_my_works_index_html_erb__2734889455683821681_70214954258920'
     # ./app/views/hyrax/my/works/index.html.erb:12:in `___sers_alex_clones_hyrax_app_views_hyrax_my_works_index_html_erb__2734889455683821681_70214954258920'
     # ./spec/views/hyrax/my/works/index.html.erb_spec.rb:16:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method `new_admin_set_path' for #<ActionDispatch::Routing::RoutesProxy:0x007fb85fbc2568>
     #   ./app/views/hyrax/my/works/index.html.erb:41:in `block in ___sers_alex_clones_hyrax_app_views_hyrax_my_works_index_html_erb__2734889455683821681_70214954258920'

mjgiarlo:
@dunn Hrmm. Try using one of the work types already generated, maybe? `GenericWork`. If that doesn't work, can you push up the broken code? Also, feel free to push your branch to the projecthydra-labs org (you now have access), and that way I can toss a commit at your branch later. Cool?

dunn:
> Also, feel free to push your branch to the projecthydra-labs org (you now have access), and that way I can toss a commit at your branch later.

I'd have to open a new PR in that case, though, since it'd be a different branch?

mjgiarlo:
@dunn :speech_balloon: 
> I'd have to open a new PR in that case, though, since it'd be a different branch?

Yep, you're right.

dunn:
Using `GenericWork` seems to work, thanks! I was just looking at the Hyrax-defined models; where is `GenericWork` defined?

jcoyne:
@dunn it's not part of hyrax, but our test suite generates it for testing purposes.

mjgiarlo:
@dunn It's generated here: https://github.com/projecthydra-labs/hyrax/blob/master/spec/test_app_templates/lib/generators/test_app_generator.rb#L11

PR #945, opened 2017-05-11
Show mediated works

### LINKED ISSUE #855: Works in workflow should be listed with All Works
Works in active workflow (undergoing review) are not listed on the dashboard under Repository Contents > Works - All Works tab. They should be viewable to the repository admin on this tab, and filterable along with published works in the existing filter.

jcoyne:
Fixes #855

### DIFF
--- a/app/search_builders/hyrax/works_search_builder.rb
+++ b/app/search_builders/hyrax/works_search_builder.rb
@@ -1,0 +1,7 @@
+module Hyrax
+  # Returns all works, either active or suppressed.
+  # This should only be used by an admin user
+  class WorksSearchBuilder < ActiveWorksSearchBuilder
+    self.default_processor_chain -= [:only_active_works]
+  end
+end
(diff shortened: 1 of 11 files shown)


app/search_builders/hyrax/works_search_builder.rb
```diff
@@ -1,13 +1,7 @@
...
-  class WorksSearchBuilder < Hyrax::SearchBuilder
-    include Hyrax::FilterByType
-
-    def only_works?
-      true
-    end
+  # Returns all works, either active or suppressed.
+  # This should only be used by an admin user
+  class WorksSearchBuilder < ActiveWorksSearchBuilder
+    self.default_processor_chain -= [:only_active_works]
```
mjgiarlo: I would expect to see this in the `ActiveWorksSearchBuilder` rather than the `WorksSearchBuilder`. Did those get mixed up, or is it just me?

jcoyne: This is removing the filter.

PR #948, opened 2017-05-11
Extract the notification poll interval to an engine configuration setting

cbeer:


mjgiarlo:
@cbeer For this config and the uploader, do you think we shouldn't have them in `lib/hyrax/configuration.rb` and included in the `lib/generators/hyrax/templates/config/hyrax.rb` template so folks can see the configs they are able to tweak in their apps? I thought I'd ask since it looks like we're storing config options in a couple different places now, and it may not be clear to other developers when to place configs in one place vs. the other.

cbeer:
`Hyrax::Configuration` is news to me. Moving this there sounds reasonable.

mjgiarlo:
@cbeer are you inclined to move the new uploader options as a follow-up?

PR #950, opened 2017-05-11
Workflow notifications should be sent individually

### LINKED ISSUE #944: Incorrect behavior with notifications
### Descriptive summary

Hyrax master branch: irregular notification behavior is occuring

### Expected behavior

Workflows send notifications to multiple users. These notifications should be scoped individually to each user. 

- Deleting a notification should only affect the initiating user's notifications. 
- Delete all should redisplay the empty list.

### Actual behavior

When viewing notifications, "Delete all" does not change the view... all notifications remain in the list (even after refreshing the browser), until you click the trash icon icon next to one of the notifications, then the page redisplays as empty.

Additionally, notifications do not appear to be correctly scoped, as deleting a notification removed it from ALL of the users.

1. Do things with workflows that cause notifications to multiple users.
2. Clear notifications for one of the users. They appear to not clear.
3. Click the trash can next to one of the notifications. The page redisplays as empty.
4. Sign on as one of the other users who got the same notification. Their notifications list is also empty.

mjgiarlo:
Otherwise, when a user removes one or more messages, they disappear from all user inboxes.

Fixes #944

### DIFF
--- a/app/services/hyrax/workflow/abstract_notification.rb
+++ b/app/services/hyrax/workflow/abstract_notification.rb
@@ -19,7 +19,9 @@ module Hyrax
       end
 
       def call
-        user.send_message(users_to_notify.uniq, message, subject)
+        users_to_notify.uniq.each do |recipient|
+          user.send_message(recipient, message, subject)
+        end
       end
 
       protected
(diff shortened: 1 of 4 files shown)


laritakr:
This addresses one aspect of #944. Should the other item be moved to its own issue or are you making that changes as well ("Delete all" should redisplay the empty list)?

mjgiarlo:
@laritakr I couldn't reproduce the Delete All bug in our Hyku instance.

PR #951, opened 2017-05-11
Add a base class for all Hyrax jobs

jcoyne:


app/jobs/hyrax/background_job.rb
```diff
@@ -0,0 +1,7 @@
+module Hyrax
+  # A common base class for all Hyrax jobs.
+  # This allows upstream applications to manipulate all the hyrax jobs by
```
atz: downstream?

atz:
Why not just `Hyrax::Job`?

app/jobs/hyrax/background_job.rb
atz: Or if stream-directionality is confusing, "Hyrax consumers" is what we mean here.

jcoyne:
Because ActiveJob implies it's happening in the background. So BackgroundJob so as not to be confusing.

mjgiarlo:
coveralls :speech_balloon: 
> Coverage increased (+0.0007%) to 96.77%

NAILED IT, @JCOYNE

atz:
@jcoyne ActiveJob doesn't imply it is happening in the background, since you can call `perform_now` AND you can configure your adapter to `:inline`.

atz:
Even using a proper queue and workers, "Background" isn't the best metaphor here.  That suggests a forked process (like shell) or spawned task on the same server.  In reality nothing more is being _accurately_ communicated than to say this is `Hyrax::ActiveJob`.  Having no other base than `ActiveJob`, that is the same as `Hyrax::Job`.  Backgroundness or not is not part of the definition more than to say this is a job.

cbeer:
Perhaps `Hyrax::ApplicationJob` to align with the new Rails 5.1 stuff?

mjgiarlo:
@atz @cbeer How much do y'all care about the name?

cbeer:
I agree with @atz that `BackgroundJob` is problematic and confusing. I suspect it's worth getting right so we don't need to try to fix it later.

mjgiarlo:
Ok, let's go with ApplicationJob, then?

Issue #225, opened 2017-01-10
Style for workflow state


mjgiarlo:
<a href="https://github.com/jcoyne"><img src="https://avatars.githubusercontent.com/u/92044?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jcoyne](https://github.com/jcoyne)**
_Friday Sep 23, 2016 at 15:15 GMT_
_Originally opened as https://github.com/projecthydra/curation_concerns/issues/1029_

----

On the show page under the title. `.show`

mjgiarlo:
<a href="https://github.com/mjgiarlo"><img src="https://avatars.githubusercontent.com/u/131982?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mjgiarlo](https://github.com/mjgiarlo)**
_Friday Oct 14, 2016 at 21:03 GMT_

----

Item show page, @jcoyne?

mjgiarlo:
<a href="https://github.com/jcoyne"><img src="https://avatars.githubusercontent.com/u/92044?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jcoyne](https://github.com/jcoyne)**
_Friday Oct 14, 2016 at 22:06 GMT_

----

Yep

mjgiarlo:
@projecthydra-labs/hyrax-ui-ux-advisors Do you folks have suggestions for how we should display the current workflow state of items on their show pages?

newmanld:
@mjgiarlo – can you add some examples of workflow states that might be displayed?  And would part of a given workflow configuration be whether or not a work, and thus a show page for the work,  is itself private, at any given workflow stage?

mjgiarlo:
@newmanld You bet. I'll toss a screenshot here this afternoon. You'll just see the one state, but there are a handful of possible labels you could see in its place.

And, yes, workflow configuration includes which steps in the workflow "activate" or "deactivate" a work. When a work is activated, the usual Hydra access controls are used. When a work is deactivated, the code also factors in workflow roles of the requesting user.

mjgiarlo:
See the "complete" label beneath the work title, @newmanld @projecthydra-labs/hyrax-ui-ux-advisors:

![workflow_label](https://cloud.githubusercontent.com/assets/131982/21829044/c910c864-d748-11e6-936f-af90dfe0c0ff.png)

mjgiarlo:
[...]

PR #972, opened 2017-05-12
Render workflow state as a label

mjgiarlo:
Prior to this commit, workflow states displayed as unstyled text immediately below the work title. We discussed alternatives in #225, and this is not that full solution; this is a quick (arguable) improvement. 

Refs #225.

Before:

workflow_label

After: 

screenshot from 2017-05-12 15-11-34

@projecthydra-labs/hyrax-code-reviewers @projecthydra-labs/hyrax-ui-ux-advisors

### DIFF
--- a/app/views/hyrax/file_sets/_file_set_title.erb
+++ b/app/views/hyrax/file_sets/_file_set_title.erb
@@ -1,7 +1,7 @@
 <% presenter.title.each_with_index do |title, index| %>
     <% if index == 0 %>
         <h1>
-            <%= title %> <%= presenter.permission_badge %> <%= presenter.workflow.badge %>
+            <%= title %> <%= presenter.permission_badge %>
         </h1>
     <% else %>
         <h1><%= title %></h1>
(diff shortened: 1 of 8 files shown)


app/presenters/hyrax/presents_attributes.rb
```diff
@@ -28,6 +28,18 @@ def permission_badge_class
       PermissionBadge
     end
 
+    def workflow_badge
```
jcoyne: Why is this in this module if it can't be shared? Why not just on the presenter?

mjgiarlo: Good question. I'll rework it.

app/views/hyrax/file_sets/_file_set_title.erb
```diff
@@ -0,0 +1,9 @@
+<% presenter.title.each_with_index do |title, index| %>
```
mjgiarlo: FileSets used to use the `work_title` partial.

mjgiarlo:
@jcoyne pls re-review?

PR #974, opened 2017-05-15
Fixes alignment of radio button for thumbnail in file manager.

### LINKED ISSUE #928: Radio button for 'thumbnail' in file manager displays on wrong line
### Descriptive summary

Hyrax master; Chrome 58.0.3029.96

### Expected behavior

Radio button should be next to 'thumbnail'

### Actual behavior

Radio button is up by filename box.

1. Create a work with two files
1. Go to the file manager

screen shot 2017-05-09 at 2 32 47 pm

andjsmit:
Fixes #928

Fixes position of radio button for thumbnail option in file manager.

### DIFF
--- a/app/assets/stylesheets/hyrax/_file_manager.scss
+++ b/app/assets/stylesheets/hyrax/_file_manager.scss
@@ -105,6 +105,7 @@
       }
       input[type=radio] {
         width: 20px;
+        position: relative;
       }
     }
   }


mjgiarlo:
Thoughts on this, @jcoyne @ggeisler @hackmastera? LGTM, but stylesheets are not my strong suit.

ggeisler:
Screenshots for things like this would be helpful for those of us not running the app all the time...

andjsmit:
@ggeisler there are before and after screenshots in issue #928. Thx.

ggeisler:
@andjsmit Sure, thanks for pointing me to them. I guess my point, though, is when a PR involves a change to the UI it is helpful for reviewers (at least those who are more interested in the effect on the UI, rather than the changes to the underlying code) to be able see screenshots in the PR itself (i.e., here, for this PR) so we can offer an opinion on how things look without having to actually pull and run your branch locally. (My local dev environment is active with another project I'm working on so doing that is not trivial.)

Looking at #928 I'm not really clear what's going on. Do the last screenshots in the comment thread reflect what is in this PR? If so, I like the reorganization and think things look much cleaner.

This might be beside the point of #928 but in one of those screenshots I see two radio buttons selected. If you are using radio buttons only one should be selectable at a time. If the user can select both, or neither, at the same time, you should use checkboxes instead of radio buttons. (I haven't paid much attention to the File Manager feature yet so if my comments don't make sense, please ignore them and go with whatever the others on this thread want to do.)

mjgiarlo:
@ggeisler I believe they should be radio buttons, but the zoomed screenshot is making it difficult to see why. The file manager displays a bunch of files within a work, and for each file you can set it as the thumb or the representative media. So you don't have to choose between the two options for a file. Make sense?

andjsmit:
@ggeisler So like @mjgiarlo said, the radio buttons work more in a horizontal way. With multiple files one and only one Thumbnail can be selected and one and only one Representative Media is selected. Both radio buttons were originally in the code. It was just that the Thumbnail one was showing up at the top (in chrome) or outside the box in Firefox.

file_manager_3
file_manager_2
file_manager_1

ggeisler:
@andjsmit @mjgiarlo Okay, thanks for the explanation. As @mjgiarlo said, the fuller context wasn't clear to me from the screenshot in the other ticket.

The bottom version still feels odd to me, where page1.jp2 has both the radio buttons for Thumbnail and Representative Media selected, but you guys understand what's going on better than I.

jcoyne:
@ggeisler having a separate thumbnail and representative media makes more sense for an audio or video object, where you want the representative media to be a sound or video and you want the thumbnail to be cover art or a screen grab.

ggeisler:
@jcoyne Yeah, I get that part. It's just the fact that you can select two radio buttons (or select none) that appear to be in the same group that I think is off. With radio buttons you should only be able to select one in a related group of choices.

jcoyne:
@ggeisler So you are saying, we shouldn't interleave two groups of radio buttons because that's visually indistinguishable from many groups of two radio buttons.

ggeisler:
@jcoyne Yeah, I guess visually it just appears that the radio groups are per "page" (the blocks that look like independent "cards") in this case; the visual separation of the cards makes it less obvious that the groups span the cards.

Anyway, I'm not trying to stop this from being merged into the app, just wanted to make sure we were using radio buttons appropriately. It seems like we are, it's just a little less than ideal visually. But maybe it's perfectly fine in practice.

mjgiarlo:
Thanks, @ggeisler. Should we write up a new "design needed" issue for aligning the file manager design with the rest of the site?

tpendragon:
@ggeisler Oh, that's my fault. It looks silly here, but it works well in practice (users seem to know what's happening immediately.) In our case we actually have two sets of radio buttons too: some are per FileSet, and then there's the thumbnail/representative selector.

screen shot 2017-05-16 at 9 40 09 am

ggeisler:
@mjgiarlo I think we can just go with this and wait and see if it surfaces as anything that needs more design and/or development. @tpendragon says it works well in practice so that's good enough for me! Again, I think some of my concerns were due to seeing a more limited view of what was going on; not seeing things in the full context.

PR #983, opened 2017-05-16
Fixing documentation

jeremyf:


app/models/sipity/method.rb
```diff
@@ -1,5 +1,6 @@
 module Sipity
   class Method < ActiveRecord::Base
     self.table_name = 'sipity_workflow_methods'
+    belongs_to :workflow_action, class_name: 'Sipity::WorkflowAction'
```
jcoyne: This seems like more than documentation.

jeremyf: Yeah. I'll pull that out of this one.

mjgiarlo:
Thank you, @jeremyf

PR #986, opened 2017-05-17
Fix visibility form collapse/expand for batches

### LINKED ISSUE #930: Visibility options for Embargo don't expand on batch create form
### Descriptive summary

When a user is on the batch create form and clicks the radio button for "Embargo" the form does not expand to reveal the "Restricted to..." or "then open it up to.." options like it does on the single work form.  

We experienced the bug on 1.0.0.rc1, but it looks like the bug is still in master as well.

### Expected behavior

Clicking on the "Embargo" radio button in the Visibility section of the batch create form should reveal additional form options.

### Actual behavior

Clicking on the "Embargo" radio button in the Visibility section of the batch create form has no effect other than selecting the radio button itself.

1. Go to the form to create a new batch work
1. Click on the "Embargo" radio button

### Related work

University of Cincinnati has fixed this locally here: https://github.com/uclibs/scholar_uc/pull/1373/files

The problem is that the batch form doesn't include the `data-behavior="work-form"` in the <form> tag.  So the JS that expands the options doesn't know what to act on.

hortongn:
Fixes #930 

This fixes a bug in the visibility pane on the new batch form.  Clicking on the `Embargo` radio button will now expand and reveal the "Restricted to..." or "then open it up to.." options like it does on the single work form.  

### DIFF
--- a/app/views/hyrax/batch_uploads/_form.html.erb
+++ b/app/views/hyrax/batch_uploads/_form.html.erb
@@ -1,4 +1,9 @@
-<%= simple_form_for [hyrax, @form], html: { multipart: true } do |f| %>
+<%= simple_form_for [hyrax, @form],
+                    html: {
+                      data: { behavior: 'work-form',
+                              'param-key' => @form.model_name.param_key },
+                      multipart: true
+                    } do |f| %>
   <% provide :files_tab do %>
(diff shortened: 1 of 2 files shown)


mjgiarlo:
Thanks, @hortongn!

PR #992, opened 2017-05-17
Fix Browse Everything when work is being edited

### LINKED ISSUE #931: Browse everything only attaches files on work create (not edit)
### Descriptive summary

Attaching cloud files to a work behaves as expected when creating a brand new work.  However, if an existing work is edited and the user attempts attach cloud-based files, and exception is thrown upon saving the work.  Files are not attached.

We experienced the bug on 1.0.0.rc1, but it looks like the bug is still in master as well.

### Expected behavior

The user should be able to attach cloud files either when creating a work or editing a work.

### Actual behavior

Editing and saving an existing work with new cloud-based files attached throws an error.

1. Create a work
1. Edit the work
1. Add some new files from a cloud provider
1. Save the work

### Related work

University of Cincinnati has fixed this locally here: 
https://github.com/uclibs/scholar_uc/pull/1371/files

The fix is in the last line of the partial.

The problem is that when editing a work, browse_everything is looking for a form id of `edit_generic_work`. However the actual form id has the pid attached (e.g. `edit_generic_work_h989r335q`)

hortongn:
Fixes #931 

Appends the work pid to the Browse Everything `data-target` when editing a work so that browse everything can properly target the form (e.g. ` edit_generic_work_h989r335q` instead of `edit_generic_work`.  Without this fix, cloud files can't be added to an existing work.

### DIFF
--- a/app/views/hyrax/base/_browse_everything.html.erb
+++ b/app/views/hyrax/base/_browse_everything.html.erb
@@ -3,4 +3,4 @@
 </div>
 <%= button_tag(t('hyrax.upload.browse_everything.browse_files_button'), type: 'button', class: 'btn btn-lg btn-success', id: "browse-btn",
   'data-toggle' => 'browse-everything', 'data-route' => browse_everything_engine.root_path,
-  'data-target' => "##{f.object.persisted? ? 'edit' : 'new'}_#{f.object.model.model_name.param_key}" ) %>
+  'data-target' => "#{f.object.persisted? ? "#edit_#{f.object.model.model_name.param_key}_#{f.object.model.id}" : "#new_#{f.object.model.model_name.param_key}"}" ) %>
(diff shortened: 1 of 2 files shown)


app/views/hyrax/base/_browse_everything.html.erb
```diff
@@ -3,4 +3,4 @@
 </div>
 <%= button_tag(t('hyrax.upload.browse_everything.browse_files_button'), type: 'button', class: 'btn btn-lg btn-success', id: "browse-btn",
   'data-toggle' => 'browse-everything', 'data-route' => browse_everything_engine.root_path,
-  'data-target' => "##{f.object.persisted? ? 'edit' : 'new'}_#{f.object.model.model_name.param_key}" ) %>
+  'data-target' => "##{f.object.persisted? ? 'edit' : 'new'}_#{f.object.model.model_name.param_key}#{f.object.persisted? ? '_'+f.object.model.id : ''}" ) %>
```
jcoyne: Can we do this instead?
```ruby
f.object.persisted? ? "#edit_#{f.object.model.model_name.param_key}_#{f.object.model.id}" :
"#new_#{f.object.model.model_name.param_key}"
```

hortongn: @jcoyne Yeah, that's cleaner.  I've made the change.  Please re-review when you get a chance.

PR #1002, opened 2017-05-18
Update active-fedora ~> 11.2

darrenleeweber:

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -67,7 +67,7 @@ EOF
   spec.add_dependency 'active_attr', '~> 0.9.0'
   spec.add_dependency 'redlock', '>= 0.1.2'
   spec.add_dependency 'retriable', '>= 2.9', '< 4.0'
-  spec.add_dependency 'active-fedora', '>= 11.1.3'
+  spec.add_dependency 'active-fedora', '~> 11.2'
   spec.add_dependency 'linkeddata' # Required for getting values from geonames
 
   spec.add_development_dependency 'engine_cart', '~> 1.0'


hyrax.gemspec
```diff
@@ -67,7 +67,7 @@ EOF
   spec.add_dependency 'active_attr', '~> 0.9.0'
   spec.add_dependency 'redlock', '>= 0.1.2'
   spec.add_dependency 'retriable', '>= 2.9', '< 4.0'
-  spec.add_dependency 'active-fedora', '>= 11.1.3'
+  spec.add_dependency 'active-fedora', '~> 11.2'
```
jcoyne: `11.2.0` satisfies the existing constraint of`>= 11.1.3`. Is there some reason we need to force the upgrade here?

darrenleeweber: I chose this one to pin to any `11.x` (using the `~>` operator).  I'm not attached any form in particular, I just find that one easier to read as an `11.x` constraint.

jcoyne: @darrenleeweber when you specify `~> 11.2` that means any `11.x` where `x >= 2`.  So this  is forcing people running `11.1.x` to update.  Is there any changes we need in hyrax that requires this update at this time?

darrenleeweber: The intention was to have it in hyku, at least.  I don't understand the stack entirely and what all the community use cases are.  I assumed the update belongs on hyrax and then it gets pulled into hyku.  Also, this PR at least prompts consideration for this update in hyrax now that 11.2 is released.  I assumed an 11.x release is backward compatible and only releases new features and bug fixes.

darrenleeweber: So now I'm wondering if there is something special about `11.1.3` that fixed critical bugs or added critical features.  In other words, I'm trying to understand what the most permissive constraint could be, like `>= 11.0.0` or say `~> 11.0`, and why the current constraint is not that permissive.

darrenleeweber: Aside from understanding, I suppose there is a case for allowing `11.1.x`, which implies closing this PR unmerged.  AFAIK, at the moment, there's nothing in particular that I need from this.

mjgiarlo: OK, why don't we close this PR, @darrenleeweber, and if you want AF 11.2.0 in Hyku, let's pin it there. Cool?

PR #1008, opened 2017-05-19
Gemfile - remove engine cart block from git repo

darrenleeweber:
The master branch contains an older engine-cart block in the Gemfile.  When a more recent engine-cart runs, it appends a new block to the end of the Gemfile, adding to the existing block.  (This _might_ be a bug in engine cart, I dunno.)  Anyhow, this block doesn't belong in the gem - I know it is supposed to be smart about not interfering with the gem, but this block is only a development dependency.

I want to suggest the `spec/test_app_templates` stuff is a) moved to devel-dep gems and b) converted to rake tasks with some kind of engine-ish namespace that are only enabled in development/test.

Here's an example git-diff for the appended engine-cart block that prompted this PR:
```
 unless File.exist?(file)
   eval_gemfile File.expand_path('spec/test_app_templates/Gemfile.extra', File.dirname(__FILE__))
 end
+# BEGIN ENGINE_CART BLOCK
+# engine_cart: 1.1.0
+# engine_cart stanza: 0.10.0
+# the below comes from engine_cart, a gem used to test this Rails engine gem in the context of a Rails app.
+file = File.expand_path('Gemfile', ENV['ENGINE_CART_DESTINATION'] || ENV['RAILS_ROOT'] || File.expand_path('.internal_test_app', File.dirname(__FILE__)))
+if File.exist?(file)
+  begin
+    eval_gemfile file
+  rescue Bundler::GemfileError => e
+    Bundler.ui.warn '[EngineCart] Skipping Rails application dependencies:'
+    Bundler.ui.warn e.message
+  end
+else
+  Bundler.ui.warn "[EngineCart] Unable to find test application dependencies in #{file}, using placeholder dependencies"
+
+  if ENV['RAILS_VERSION']
+    if ENV['RAILS_VERSION'] == 'edge'
+      gem 'rails', github: 'rails/rails'
+      ENV['ENGINE_CART_RAILS_OPTIONS'] = '--edge --skip-turbolinks'
+    else
+      gem 'rails', ENV['RAILS_VERSION']
+    end
+  end
+
+  case ENV['RAILS_VERSION']
+  when /^4.2/
+    gem 'responders', '~> 2.0'
+    gem 'sass-rails', '>= 5.0'
+    gem 'coffee-rails', '~> 4.1.0'
+  when /^4.[01]/
+    gem 'sass-rails', '< 5.0'
+  end
+end
+# END ENGINE_CART BLOCK
```


Gemfile
```diff
@@ -10,41 +10,8 @@ group :development, :test do
   gem "simplecov", require: false
 end
 
-# BEGIN ENGINE_CART BLOCK
```
jcoyne: This should always be in the git repo. The way you use engine cart is to run the installer, and then check in the results.

darrenleeweber: engine-cart auto-detects this block and adds it when it is missing, it does not have to be in the repo.

cbeer: @jcoyne is right; we expect gem maintainers to check in this stanza.

darrenleeweber:
I might have to abandon this PR, it's too complicated.  I might switch to an issue on engine cart to note that it does not replace an existing Gemfile stanza when it needs to be updated.

PR #1010, opened 2017-05-19
Gemfile - update engine cart stanza for v1.1.0

darrenleeweber:
Related to https://github.com/cbeer/engine_cart/issues/79 (does not fix it in engine_cart)

@mjgiarlo - I just discovered that this could undo a prior hack on the engine-cart stanza:
  - https://github.com/projecthydra-labs/hyrax/commit/8071e331a5d522c42dc104bbfaa7f64b7cb422c8
  - Is there another way to achieve that without hacking the engine-cart stanza?
    - attempted to do so in https://github.com/projecthydra-labs/hyrax/pull/1010/commits/fefd4901a85d4be13ec2ff2cb5a418a1d66d6b99
  - review commentary resolved to ignore overriding this gem version for rails 4.2 because hyrax needs rails 5.x now anyway, so the commit ^^ was removed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,9 +36,9 @@ else
 
   case ENV['RAILS_VERSION']
   when /^4.2/
+    gem 'coffee-rails', '~> 4.1.0'
     gem 'responders', '~> 2.0'
     gem 'sass-rails', '>= 5.0'
-    gem 'coffee-rails', '~> 4.1.0'
   when /^4.[01]/
     gem 'sass-rails', '< 5.0'
   end
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ group :development, :test do
 end
 
 # BEGIN ENGINE_CART BLOCK
-# engine_cart: 0.10.0
+# engine_cart: 1.1.0
 # engine_cart stanza: 0.10.0
 # the below comes from engine_cart, a gem used to test this Rails engine gem in the context of a Rails app.
 file = File.expand_path('Gemfile', ENV['ENGINE_CART_DESTINATION'] || ENV['RAILS_ROOT'] || File.expand_path('.internal_test_app', File.dirname(__FILE__)))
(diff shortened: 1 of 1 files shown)


darrenleeweber:
@cbeer - add rubocop checks on engine_cart so it's stanza passes rubocop checks.

```
Gemfile:41:5: C: Bundler/OrderedGems: Gems should be sorted in an alphabetical order within their section of the Gemfile. Gem coffee-rails should appear before sass-rails.

    gem 'coffee-rails', '~> 4.1.0'

    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Gemfile
```diff
@@ -8,10 +8,15 @@ group :development, :test do
   gem 'pry' unless ENV['CI']
   gem 'pry-byebug' unless ENV['CI']
   gem "simplecov", require: false
+  # Override the coffee-rails version in engine_cart for rails 4.2
+  case ENV['RAILS_VERSION']
+  when /^4.2/
+    gem 'coffee-rails', '~> 4.2.0'
```
jcoyne: This seems like it conflicts with line 44.  Can you say more about why we need both?

darrenleeweber: The line-44 code is maintained by engine-cart, it should not be modified by anything but engine cart and we should assume that engine-cart will not respect any changes in the block it maintains.  To override anything in that block, it must be declared separately.  So, in this PR, to preserve the intention of a prior commit, this block was added in the last commit.

darrenleeweber: A separate issue is whether this override is actually needed at all any more, so I did not intend to address that in this PR.

cbeer: > The line-44 code is maintained by engine-cart, it should not be modified by anything but engine cart

I wouldn't go that far -- instead, I'd say: by modifying it, the onus is on the project maintainers to ensure it is maintained the way they expect.

I thought having duplicate gem declarations (particularly conflicting ones) makes bundler unhappy?

darrenleeweber: Good question - I have not tested what happens when the rails version hits this code.  If hyrax is no longer compatible with rails 4.x, then I should just remove the last commit and leave that stanza as engine-cart creates it (i.e. discard the override entirely).

cbeer: 👍

darrenleeweber: Removed the last commit to override anything on rails 4.2

PR #1031, opened 2017-05-24
[WIP] exclude "admin" from groups sharing form

### LINKED ISSUE #996: Remove admin group from sharing section of collection form
Check the work form for the same.

screen shot 2017-05-22 at 1 56 35 pm

darrenleeweber:
WIP, needs:
- clarification on the extent of excluding 'admin' group
- feature specs
- code cleanup

Fixes #996

### DIFF
--- a/app/views/hyrax/base/_form_share.html.erb
+++ b/app/views/hyrax/base/_form_share.html.erb
@@ -62,10 +62,21 @@
 <h2>Share work with groups of users</h2>
 <p class="sr-only">Use the add button to give access to one group at a time (it will be added to the list below).</p>
 <div class="form-group row">
+  <!--<div class="col-sm-5">-->
+    <!--<label for="new_group_name_skel" class="sr-only">Group</label>-->
+    <!--<%=# select_tag 'new_group_name_skel', options_for_select(["Select a group"] + current_user.groups), class: 'form-control' %>-->
+  <!--</div>-->
+
   <div class="col-sm-5">
+    <%
+      # https://github.com/projecthydra-labs/hyrax/issues/996
+      groups = current_user.groups
+      groups = groups.without('admin') if current_user.ability.admin?
+    %>
     <label for="new_group_name_skel" class="sr-only">Group</label>
-    <%= select_tag 'new_group_name_skel', options_for_select(["Select a group"] + current_user.groups), class: 'form-control' %>
+    <%= select_tag 'new_group_name_skel', options_for_select(["Select a group"] + groups), class: 'form-control' %>
   </div>
+
   <div class="col-sm-4">
     <label for="new_group_permission_skel" class="sr-only">Access type to grant</label>
     <%= select_tag 'new_group_permission_skel', options_for_select(Hyrax.config.permission_levels), class: 'form-control' %>


app/views/hyrax/base/_form_share.html.erb
```diff
@@ -62,10 +62,21 @@
 <h2>Share work with groups of users</h2>
 <p class="sr-only">Use the add button to give access to one group at a time (it will be added to the list below).</p>
 <div class="form-group row">
+  <!--<div class="col-sm-5">-->
+    <!--<label for="new_group_name_skel" class="sr-only">Group</label>-->
+    <!--<%=# select_tag 'new_group_name_skel', options_for_select(["Select a group"] + current_user.groups), class: 'form-control' %>-->
+  <!--</div>-->
+
   <div class="col-sm-5">
+    <%
```
mjgiarlo: What would you think of an alternative approach to exposing this logic in a view, like one of the following two:

1. Expose a new method -- on the form object or the user object? not sure where the best place for this is -- called something intention-revealing like `groups_for_sharing` that returns a list of groups a user belongs to with `'admin'` filtered out. This would also allow us to filter out other system-controlled groups such as `'registered'` and `'public'`. That way the logic lives somewhere reusable, outside of a view.
2. Alternatively, make this a one-line change:

```ruby
<%= select_tag 'new_group_name_skel', options_for_select(["Select a group"] + current_user.groups.without('admin'), class: 'form-control' %>
```

This latter approach improves upon the current code because you can safely call `.without('admin')` even if the current user if not an admin, and saves the cost of creating and mutating a new view-specific variable. The `if` there is pretty innocuous, I admit, but I'm remembering our Sandi Metz course and keeping in mind that branches/conditions increase complexity.

mjgiarlo:
@darrenleeweber Re: extent, ultimately we will want to hide the `'admin'` group in all sharing forms (for works, like you've done, and also for collections), at least for this PR. I don't think we surface groups elsewhere.

Re: feature specs, I think it would suffice to write a view spec (vs. a feature spec, which is more costly in terms of test suite run time). Either in https://github.com/projecthydra-labs/hyrax/blob/master/spec/views/hyrax/base/_form.html.erb_spec.rb or create a new _form_share.html.erb_spec.rb.

jcoyne:
> Expose a new method -- on the form object or the user object? not sure where the best place for this is -- called something intention-revealing like groups_for_sharing that returns a list of groups a user belongs to

This is a good idea.

> with 'admin' filtered out. 

Do we really need to do this?  Why not allow this?

> This would also allow us to filter out other system-controlled groups such as 'registered' and 'public'.

We wouldn't need to do this as `registered` and `public` are pseudo groups and are never returned by `User#groups`

> That way the logic lives somewhere reusable, outside of a view.

👍

mjgiarlo:
Thanks for the PR, @darrenleeweber! You're not far off.

mjgiarlo:
@jcoyne :speech_balloon: 
> Do we really need to do this? Why not allow this?

We don't really *need* to do this, no. I thought that was the intent of #996.

darrenleeweber:
Several questions to help clarify:

If the `admin` group is to be excluded from sharing:
- does this only apply to an `admin` user?
- should a non-admin user be able to share stuff with the admin users?
  - do admin users get access to everything regardless of sharing?

mjgiarlo:
@darrenleeweber 

* Non-admin users should not see the `'admin'` group because that won't be in their list o' groups.
* Let's discuss these in standup.

app/views/hyrax/base/_form_share.html.erb
darrenleeweber: The point of the two lines of code was to be clear that the admin-group was only excluded for an admin user, this seemed to be the minimal change that could be made.  I'm not yet clear about all the details on how the admin-group is defined and to be used (additional questions below).  If the one-liner is used - the logic changes - it excludes the admin-group from sharing, period.  I need to clarify whether non-admin users should be able to share content with an admin-group and, if so, that would imply that the sharing groups could include groups that a user does not belong to.

darrenleeweber:
Discarding this PR.   At standup today, we concluded that the admin-group is not to be treated as a special case.

Issue #1028, opened 2017-05-23
declare rails, rather than railties, as a dependency


darrenleeweber:
### Descriptive summary

Rather than manage many various dependencies on rails components, manage the rails version(s) supported by the hyrax gem.

### Rationale

- The rails version is not a sem-ver, so it can introduce breaking changes on minor/patch releases, which can be difficult to track without explicit rails version dependencies.

- hyrax is a rails engine and declaring the supported rails version(s) makes it easier for bundler to resolve dependencies and avoid conflicts in rails components; it's also clearer to the gem consumers, including the engine-cart and test framework, what version(s) of rails are going to work.

### Related work

https://github.com/projecthydra-labs/hyrax/issues/1007

PS, I may not have captured the essence of the rationale here, see #1007 for some detailed discussions.

darrenleeweber:
When creating a new `rails plugin`, the generator creates a gemspec that includes the rails version used to create the engine, e.g. using rails 5.1.1 results in a gemspec with:
```
s.add_dependency "rails", "~> 5.1.1"
```

But we want a range of known-good rails versions.

darrenleeweber:
Given that travis is already testing a `5.0.x` and a `5.1.x`, it should be safe to say that hyrax is good on any `5.0.x` and any `5.1.x`.  So we need to a way to specify this without allowing any other minor version like `5.2.x` (until that too is tested by travis, and so on for every minor version release).

darrenleeweber:
For reference:  http://guides.rubygems.org/patterns/#declaring-dependencies

- we cannot use any optimistic constraints, like `>= 5.0.0` alone
- adding a pessimistic constraint is essential, like:
  - `>= 5.0.0, < 5.1` allows only `5.0.x` versions
  - `>= 5.0.0, < 5.2` allows only `5.0.x` and `5.1.x` versions, nothing above 5.1.x

We run into significant problems anytime there is a breaking change in the minor version.  At that point, the application needs to freeze all releases when it worked for a given rails version-range and move forward with bug fixes that work on the new minor version and then pin the app so it works with rails releases above that minor version (up the next minor version).  Then back-porting new features becomes extra work to support those features for the prior app releases that worked on prior minor versions of rails.  AFAIK, ruby gems and bundler require the gem-spec to specify a continuous range of versions for any dependency (it's not possible to depend on `~> 5.0.0` but exclude just one patch version in that range, like `5.0.2`).

PR #1038, opened 2017-05-24
Gemspec - replace railties with rails

darrenleeweber:
Fixes https://github.com/projecthydra-labs/hyrax/issues/1028 

It's all described in #1028 and associated issues noted there.  The current range of rails versions allowed is any `5.0.x` up to `5.1.x` (excluding anything beyond `5.2` until it is tested).

Tagging commentators from #1007 - @mjgiarlo, @jcoyne, @cbeer, @jrochkind 

I get a successful bundle install and travis is testing this range of rails versions already, i.e.
```
$ rm -rf Gemfile.lock .internal_test_app
$ bundle install --quiet && echo 'success' || echo 'failed'
success
```

- [x] debate over the details of the rails version pin resolved to use `~> 5.0`
- [x] bump the hyrax version and release a new gem for this PR?
  - @mjgiarlo notes in slack that we don't need to bump for this
- [x] squash commits
- [x] cut a branch/release that is pinned to rails `~> 4.0`, using a release point that works on 4.x?
  - @mjgiarlo notes below that hyrax will not support rails 4.x

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -26,6 +26,11 @@ EOF
   spec.version       = Hyrax::VERSION
   spec.license       = 'Apache2'
 
+  # Note: rails does not follow sem-ver conventions, it's
+  # minor version releases can include breaking changes; see
+  # http://guides.rubyonrails.org/maintenance_policy.html
+  spec.add_dependency 'rails', '~> 5.0'
+
   spec.add_dependency 'hydra-head', '>= 10.4.0'
   spec.add_dependency 'hydra-editor', '~> 3.3'
   spec.add_dependency 'hydra-works', '~> 0.16'
@@ -52,7 +57,6 @@ EOF
   spec.add_dependency 'flipflop', '~> 2.3'
   spec.add_dependency 'jquery-datatables-rails', '~> 3.4.0'
   spec.add_dependency 'rdf-rdfxml' # controlled vocabulary importer
-  spec.add_dependency 'railties', '~> 5.0'
   spec.add_dependency 'clipboard-rails', '~> 1.5'
   spec.add_dependency 'rails_autolink', '~> 1.1'
   spec.add_dependency 'active_fedora-noid', '~> 2.0', '>= 2.0.2'
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,14 +16,21 @@ before_install:
   - "if [ $(phantomjs --version) != '2.1.1' ]; then wget https://assets.membergetmember.co/software/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2; fi"
   - "if [ $(phantomjs --version) != '2.1.1' ]; then tar -xvf $PWD/travis_phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 -C $PWD/travis_phantomjs; fi"
   - "phantomjs --version"
+
 rvm:
   - 2.4.1
+
 env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
+  # Travis should check every minor version in a range of supported versions, because
+  # rails does not follow sem-ver conventions, see http://guides.rubyonrails.org/maintenance_policy.html
+  # It should be sufficient to test only the latest of the patch versions for a minor version, they
+  # should be compatible across patch versions (only bug fixes are released in patch versions).
   matrix:
     - "RAILS_VERSION=5.0.3"
     - "RAILS_VERSION=5.1.1"
+
 services:
   - redis-server
 before_script:


darrenleeweber:
Is there any reason why hyrax would fail on rails `5.0.0` through `5.0.2`?  Travis is testing `5.0.3` and I wonder if the lowest range of rails compatibility should be `>= 5.0.3` to exclude anything `5.0.x` below that?

darrenleeweber:
Also, are there documentation or other cleanup operations required for any releases of hyrax that can and should support versions of rails `< 5.0.0`.  If anyone is currently using hyrax in any rails `4.x`, this change here will break them.

- should this PR be accompanied with a major version bump and release of hyrax?

jcoyne:
Why do we pessimistically believe that 5.2 would not work?

darrenleeweber:
@jcoyne - read #1028 details and links therein.  Basically, rails warns you that minor versions are breaking versions.  So, yea, `5.2.x` could break stuff; be pessimistic about rails upgrades on minor versions.

darrenleeweber:
Possibly the last time that hyrax explicitly supported rails 4.x was in
- https://github.com/projecthydra-labs/hyrax/blame/19f8fd7ba93920f8e96534836620f89875ed2998/.travis.yml

The travis builds removed explicit rails 4.x support in
- https://github.com/projecthydra-labs/hyrax/blame/c9ced606cb13d136d6e2be1db3eafed1b85d8698/.travis.yml

hyrax.gemspec
```diff
@@ -26,6 +26,12 @@ EOF
   spec.version       = Hyrax::VERSION
   spec.license       = 'Apache2'
 
+  # Be pessimistic, rails does not follow sem-ver conventions and
+  # minor version releases can include breaking changes; see
+  # http://guides.rubyonrails.org/maintenance_policy.html
+  # https://github.com/projecthydra-labs/hyrax/issues/1028
+  spec.add_dependency 'rails', '>= 5.0.0', '< 5.2'
```
mjgiarlo: After discussing this on stand-up yesterday, it sounds like the team was :-1: on including `'< 5.2'` in the Hyrax gemspec, because the cost of having to update the gemspec whenever Rails updates outweighs the benefit. I believe we should change the dependency to `'~> 5.0'` cc: @jcoyne @cbeer @atz

jcoyne: Agree. 👍

cbeer: I agree with @mjgiarlo -- we shouldn't pin an upper-bound. I don't think we've seen significant problems with the gem code itself; most of our problems are in getting a testing setup that works across a variety of rails versions.

darrenleeweber: Obviously, I coded and prefer the pessimistic constraint to avoid potential breaking changes downstream.  The safe upgrade process is to add a new minor release version to travis when it becomes available, read the release notes (esp. any deprecation warnings etc.) and see it the specs pass.  If or when the specs pass, bump up the pessimistic constraint to allow the new minor version.

darrenleeweber: Note that `rails plugin` itself creates a fairly pessimistic constraint, like:
- `s.add_dependency "rails", "~> 5.1.1"`

darrenleeweber: I guess the point to add is what do you want to maintain?  Do you want to take the time to track rails releases and manage the upgrade path proactively, which avoid spurious reactive responses to failures whenever someone tries to use hyrax with an un-tested version of rails.  Or, do you want to be reacting to any number of "bug" reports from users of hyrax trying to use it with a rails version that was never tested?  The latter 'reactive' maintenance seems like more work to me and possibly work that never gets done in a planned work cycle and drags developers away from planned work to try to put out fires.

darrenleeweber: Looks like the votes are in and the accepted practice is to respond to upgrade failures when they happen, so the gemspec will become `~> 5.0` to allow any `5.x` rails version.  That implies that breaking changes in rails minor versions are to be expected and fixed ad-hoc.

darrenleeweber: Once travis is done with the `~> 5.0` builds, I'll push up a squashed commit.

darrenleeweber:
For some reason, the PR specs had one failure in the rails 5.1.1 build, i.e.
```
  10) BlacklightHelper render_index_field_value description_tesim should eq "<span itemprop=\"description\">This links to <a href=\"http://example.com/\"><span class=\"glyphicon glyphicon-new-window\"></span> http://example.com/</a> What about that?</span>"
      # need a different way to test description
      Failure/Error:

        is_expected.to eq '<span itemprop="description">This links to ' \
                          '<a href="http://example.com/"><span class="glyphicon glyphicon-new-window"></span>' \
                          ' http://example.com/</a> What about that?</span>'
      
        expected: "<span itemprop=\"description\">This links to <a href=\"http://example.com/\"><span class=\"glyphicon glyphicon-new-window\"></span> http://example.com/</a> What about that?</span>"

             got: "This links to <a href=\"http://example.com/\"><span class=\"glyphicon glyphicon-new-window\"></span> http://example.com/</a> What about that?"

        (compared using ==)

      # ./spec/helpers/blacklight_helper_spec.rb:40:in `block (4 levels) in <top (required)>'
```

atz:
Looks like it is just missing the `<span itemprop=\"description\">` wrapper.  No idea why though.

darrenleeweber:
Squashed and rebased on master.

hyrax.gemspec
```diff
@@ -26,6 +26,12 @@ EOF
   spec.version       = Hyrax::VERSION
   spec.license       = 'Apache2'
 
+  # Be pessimistic, rails does not follow sem-ver conventions and
+  # minor version releases can include breaking changes; see
+  # http://guides.rubyonrails.org/maintenance_policy.html
+  # https://github.com/projecthydra-labs/hyrax/issues/1028
+  spec.add_dependency 'rails', '~> 5.0'
```
mjgiarlo: I'm not sure the comment is still relevant here. What do you think?

atz: I agree, it isn't pessimistic anymore, really.

darrenleeweber: It isn't pessimistic now (but I argue that it should be, and lost).  When it's not pessimistic, this comment serves as a reminder and a warning for anyone looking at this line of code.  It's easy to fall into a habit of thinking everything is sem-ver, but rails is not.  I think it's useful to have this reminder right here where it counts.

If y'all want it gone or changed, please indicate "delete line x->y" or "delete all comments" or give me exactly what you want the comment to be?

atz: I'd be fine w/ replacing `Be pessimistic,` with `Note:`

mjgiarlo: Ditto what @atz said. Thanks, @darrenleeweber

mjgiarlo:
@darrenleeweber I missed the stuff above about Rails 4. Hyrax does not support Rails 4.

Issue #799, opened 2017-04-18
Module architecture used incorrectly


atz:
This is symptomatic of all the Hydra ruby stack, but we need to stop abusing the module pattern.  That means more classes and more OO, less chains of `include`.  

A module is designed to be a set of methods that get mixed-in to a consuming class **without knowing anything about that class**.  

Ways to recognize a badly formed module:
- Is only `include`d once anywhere.  Premature optimization: use `concerning` or similar for structure.  
- Makes assumptions about consumers: contains calls to things not guaranteed to exist in an otherwise empty class.  
- Directly defines Class-level variables: That is the domain of the class.
- Directly defines Instance-level variables: Again, the domain of the class.
- Untraceable logic due to dozens of modules-including-modules.   

I realize `rails` is all-in on module magic, adding capabilities directly to `Module` to do things like [define class-and-instance level accessors](http://api.rubyonrails.org/classes/Module.html#method-i-mattr_accessor).  These are also terrible for our purposes.

mjgiarlo:
@atz does it make more sense to start working on this from AF up, or not really? I'm wondering what we can do to put this in action. Do we need a more detailed analysis that can, e.g., list a top ten or twenty modules to concentrate work ("bang-for-buck" analysis)? I want folks to take this on, and I'm not sure we have a shared sense of what done looks like yet. Thanks for writing this up and starting to list characteristics of antipatterns for modules!

atz:
First, I just want to stop propagating a broken pattern where development is happening, i.e. here.  Yes, the task is more difficult because the patterns seem to be already set by dependencies, but it is also more necessary.  If you want me to file similar tickets on core gems, I can do that, but pretty much everything can buck-pass to something underneath it.  Rails, devise, blacklight, BL access controls, Hydra access controls, certainly ActiveFedora, etc.  At some point I have to accept those as problematic and just work on the layer that I'm working on.  

IMO, it would be counterproductive to say "we'll only fix it here when we get new major versions from AF, BL, etc.".  Working in parallel is fine, but with a stack this deep, I doubt I will still be working on this project by the time changes bubbled up from the bottom, by which point a whole lot more Hyrax and Hyku code will have been written, likely abusing modules as before if nothing prevents it.

mjgiarlo:
@atz Yes, I'm inclined as you are. My primary concern here is that we make this actionable. It doesn't much matter to me whether we take a top-down or bottom-up approach.

tpendragon:
[...]

PR #1041, opened 2017-05-24
Remove controller behavior modules

jcoyne:
This pattern was established in order to allow local applications to
override the upstream gem classes.  In ruby 2 we can do this using
Module#prepend instead.

I believe that merging these files will make the development effort
simpler.

Ref #799


jcoyne:
@mjgiarlo you have to prepend a module, not a class, so:
```
FoosController.prepend MyFoosControllerOverrides
```

mjgiarlo:
@jeremyf Yes, it would be great to see documentation about how folks can use this technique to customize their apps.

darrenleeweber:
I tried to apply a similar change for embargo/lease controllers in https://github.com/samvera-labs/hyrax/pull/1042, but it broke specs in various ways.  If #1042 is merged prior to this PR, apply/fix this pattern on the embargo/lease controllers (either in this PR or another one).  I'll try to document this in an issue, also.

Issue #492, opened 2017-02-27
How should we use the installation template?


mjgiarlo:
How should Hyrax use the [installation template](https://github.com/projecthydra-labs/hyrax/blob/master/template.rb)? @jcoyne and I went back and forth on this for Sufia 7.x, with the two options being:

1. README always points at the template in the master branch, and the template is updated with each release to point at the latest release. This means that the README will always install the latest stable release of Hyrax.
1. README always points at the template corresponding to the current release. So if you're reading the README for Hyrax 1.2.0, you get Hyrax 1.2.0 installed. If you're reading the README for 0.1.0, you get 0.1.0 installed. If you're reading the README in master, you get the latest stable release of Hyrax. (In order to do this, we'd need to update the README and the template with each release, as ticketed here: projecthydra/sufia#2784)

I'd welcome @projecthydra-labs/hyrax-code-reviewers thoughts on this.

# Past discussions

* https://github.com/projecthydra/sufia/pull/2783#pullrequestreview-3965724
* https://github.com/projecthydra/sufia/issues/3075
* https://github.com/projecthydra/sufia/issues/2784

tpendragon:
I vote for #2.

mcritchlow:
+1 on #2. Seems like a reasonable step in the release workflow, and I (personally) like that it's more explicit.

jcoyne:
I'm for #1. I don't believe there is a use case where someone wants to use the template, knows the version they want to install, and it's not the most recent version.

jcoyne:
I'm actually going to propose a 1.1 option. The template should just say `gem 'hyrax'`. Bundler will grab the most up to date version.

mcritchlow:
@jcoyne While we're adding other options ;), would a 1.2 option be possible? Essentially: `require 'lib/hyrax/version'` in the template and interpolating the gem version?

mjgiarlo:
@jcoyne It seems like you're (perhaps reasonably) making a YAGNI argument. But, given the :+1:s above (and my own :+1:), and the very low cost of going with option number 2, is there another reason not to do this?

jcoyne:
Yes, YAGNI. It's just one more little thing to do.  It's fine to outvote me, but now you know my mind

.

mcritchlow:
As one of the 👍 'ers, I should say I'm also happy to help with this process also. Wouldn't be very fair to say "yes, someone **else** should do this extra thing" :)

no-reply:
[...]

PR #1045, opened 2017-05-25
Relax template so it points to latest released Hyrax

cjcolvar:
This would be more in line with Blacklight's approach (https://github.com/projectblacklight/blacklight/blob/master/template.demo.rb#L1) and follows the decision made in #492.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -1,4 +1,4 @@
-gem 'hyrax', '1.0.0.rc2'
+gem 'hyrax'
 
 run 'bundle install'
 


mjgiarlo:
@cjcolvar @jcoyne @jeremyf Hrmm, I'm not sure this does follow the decision made in #492. Once v1.0.2 is released, and I am following the instructions in v1.0.1's README (which points at the template in `master`), I will get v1.0.2 installed instead of v1.0.1. Steps 3 and 4 of Hyrax's release management documentation were updated after #492 was closed: https://github.com/samvera-labs/hyrax/wiki/Release-management-process It's too late to make any changes to 1.0.1, but I'd like to be faithful to our release docs in the future (though I understand that mistakes will be made because we're all overburdened and pushing through a lot of work very quickly).

jcoyne:
@mjgiarlo what's the process for changing that?  I'd be in favor of striking 4 and modifying 3 so that they are no longer necessary.  Do we need a review step? Does that belong in a wiki or under version control?

mjgiarlo:
@jcoyne The only approach for "fixing" this in v1.0.1 is to tweak `template.rb` in `1-0-stable` and re-cut the tag. I don't like that idea very much, myself, because it introduces variation (however slight) between the tag and its corresponding release. This is more about how we handle the installation template and README going forward.

As for striking 4 and modifying 3, that's what we discussed in #492, and most folks were onboard with the approach that we currently have documented so I'm not sure why we'd go back on that. (Though your strong preference for the "#1 option" on that issue is clear!) If you feel that steps 3 and 4 are too cumbersome, I'd be happy to help out when it's release time. I'm not sure we need a review step but I'm open to it?

Tagging @samvera-labs/hyrax-code-reviewers for their input here too.

jcoyne:
@mjgiarlo Can you explain how you'd hit that problem that requires this fix?

> is to tweak template.rb in 1-0-stable and re-cut the tag

I imagine that the only use case for the template is:

```
rails _5.0.3_ new my_app -m https://raw.githubusercontent.com/samvera-labs/hyrax/master/template.rb
```
In which case it's retrieving the fixed template from the master branch.

mjgiarlo:
@jcoyne If I follow you, here's what I have in mind. 2.0.0 is cut tomorrow. A person coming over from Sufia 7 wants to install Hyrax 1.0.1. They go to the 1.0.1 README and follow instructions. README says install using the template in `master`. That installs 2.0.0, not 1.0.1.

jcoyne:
If you're coming from something like sufia, you wouldn't use a template, you'd update your existing Gemfile.

mjgiarlo:
`s/coming over from Sufia 7//`, then.

hackmastera:
Just noticing. This looks merged but unresolved.

mjgiarlo:
@hackmastera Here's my take of where things stand: v1.0.1 is already cut, so we can't do much there, but we have a release management document that should help make sure we follow the process for future releases.

jcoyne:
@mjgiarlo I still don't know who would want to come from not the latest version of sufia to an old version of hyrax and use a template to do so.  If you are going to start from scratch (the use case for a template), you'd most certainly want the latest thing to avoid an upgrade in the near future.

mjgiarlo:
@jcoyne :speech_balloon: 
> If you are going to start from scratch (the use case for a template), you'd most certainly want the latest thing to avoid an upgrade in the near future.

My experience in the Samvera Community tells me this is often, but not always, true. I'm thinking about Sufia usage patterns around the time of 5 -> 6 -> 7. And the cost of the extra step is super low. I admit this is a "just in case" sort of thing, but the courtesy doesn't put us out very much.

PR #1047, opened 2017-05-25
cast column to a type that is appropriate for the adapter being used

### LINKED ISSUE #1043: Unable to use MySQL database
The line at https://github.com/projecthydra-labs/hyrax/blob/master/app/services/hyrax/workflow/permission_query.rb#L261 seems to be one that causes the query to fail when using MySQL as the Hyrax database. 

```
 ActiveRecord::StatementInvalid (Mysql2::Error: You have an error in your SQL syntax; check the
manual that corresponds to your MySQL server version for the right syntax to use near 'integer)
 FROM `sipity_agents` WHERE (`sipity_agents`.`id` IN (SELECT `sipity_wor' at line 1: SELECT
 `users`.* FROM `users` WHERE `users`.`id` IN (SELECT CAST(`sipity_agents`.`proxy_for_id` AS
 integer) FROM `sipity_agents` WHERE (`sipity_agents`.`id` IN (SELECT
 `sipity_workflow_responsibilities`.`agent_id` FROM `sipity_workflow_responsibilities` WHERE
 `sipity_workflow_responsibilities`.`workflow_role_id` IN (SELECT `sipity_workflow_roles`.`id`
FROM `sipity_workflow_roles` WHERE `sipity_workflow_roles`.`role_id` IN (3))) OR
`sipity_agents`.`id` IN (SELECT `sipity_entity_specific_responsibilities`.`agent_id` FROM
`sipity_entity_specific_responsibilities` WHERE
`sipity_entity_specific_responsibilities`.`workflow_role_id` IN (SELECT `sipity_workflow_roles`.`id`
FROM `sipity_workflow_roles` WHERE `sipity_workflow_roles`.`role_id` IN (3)) AND
`sipity_entity_specific_responsibilities`.`entity_id` = '2')) AND `sipity_agents`.`proxy_for_type` =
'User')):
```

revgum:
fixes #1043

### DIFF
--- a/app/services/hyrax/workflow/permission_query.rb
+++ b/app/services/hyrax/workflow/permission_query.rb
@@ -257,8 +257,11 @@ module Hyrax
             .and(entity_responsibilities[:entity_id].eq(entity.id))
         )
 
-        # PostgreSQL requires an explicit cast from string to integer
-        cast = Arel::Nodes::NamedFunction.new "CAST", [agent_table[:proxy_for_id].as("integer")]
+        # Default to "integer" for adapters like Postgres and Sqlite, but cast
+        # to "signed" for Mysql. The type CAST causes a SQL syntax error for an
+        # unsupported type depending on which adapter is being used.
+        type = ActiveRecord::Base.connection.adapter_name.casecmp("mysql2").zero? ? "signed" : "integer"
+        cast = Arel::Nodes::NamedFunction.new "CAST", [agent_table[:proxy_for_id].as(type)]
 
         sub_query_for_user = agent_table.project(cast).where(
           agent_table[:id].in(workflow_agent_id_subquery)


app/services/hyrax/workflow/permission_query.rb
```diff
@@ -257,8 +257,8 @@ def scope_users_for_entity_and_roles(entity:, roles:)
             .and(entity_responsibilities[:entity_id].eq(entity.id))
         )
 
-        # PostgreSQL requires an explicit cast from string to integer
-        cast = Arel::Nodes::NamedFunction.new "CAST", [agent_table[:proxy_for_id].as("integer")]
+        # PostgreSQL requires an explicit cast from string, MySQL allows for signed
+        cast = Arel::Nodes::NamedFunction.new "CAST", [agent_table[:proxy_for_id].as("signed")]
```
jcoyne: Have you tested this on Postgres?  Will this cause a regression?

jcoyne:
@revgum can you update your commit message to tell us about the bug you fixed?

How about "cast column to a type that is recognized by MySQL"

app/services/hyrax/workflow/permission_query.rb
jcoyne:
```
ERROR:  type "signed" does not exist
LINE 1: SELECT CAST('1' as signed)
                           ^
```

revgum:
@jcoyne This update is tested against mysql and postgres, and seems like a reasonable solution that defaults to "integer" but works for mysql.

app/services/hyrax/workflow/permission_query.rb
```diff
@@ -257,8 +257,11 @@ def scope_users_for_entity_and_roles(entity:, roles:)
             .and(entity_responsibilities[:entity_id].eq(entity.id))
         )
 
-        # PostgreSQL requires an explicit cast from string to integer
-        cast = Arel::Nodes::NamedFunction.new "CAST", [agent_table[:proxy_for_id].as("integer")]
+        # Default to "integer" for adapters like Postgres and Sqlite, but cast
+        # to "signed" for Mysql. The type CAST causes a SQL syntax error for an
+        # unsupported type depending on which adapter is being used.
+        type = ActiveRecord::Base.connection.adapter_name.casecmp("mysql2").zero? ? "signed" : "integer"
```
jcoyne: How about giving this a try: `ActiveRecord::Base.connection.type_to_sql(:integer)` that returns `"integer"` in postgres.  I suspect it'll return `unsigned` in mysql.

revgum: Fails with syntax error:

`ActiveRecord::StatementInvalid (Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'int) `

jcoyne:
Looks like we need this merged first https://github.com/projecthydra-labs/hyrax/pull/1055

PR #1049, opened 2017-05-25
[WIP] Deprecate Hyrax's Noid functionality

jcoyne:
NOID has been the favored ID generator since Sufia's early days.  But there are some issues with NOID:

1. It may be faster and less error prone to use the default Fedora UUIDs.
1. Some institutions may favor their own service for generating identifiers.  
1. Adding another mandatory dependency slows the build.

Deprecating NOID integration will enable us to address these issues.

*NOTE* If you were using noids, you must now explicitly turn them on: `config.enable_noids = true`. Previously this was the default.

### DIFF
--- a/spec/models/file_set_spec.rb
+++ b/spec/models/file_set_spec.rb
@@ -222,6 +222,7 @@ RSpec.describe FileSet do
   describe 'noid integration', :clean_repo do
     let(:service) { instance_double(ActiveFedora::Noid::Service, mint: noid) }
     before do
+      Hyrax.config.enable_noids = true
       allow(ActiveFedora::Noid::Service).to receive(:new).and_return(service)
     end
 
@@ -496,29 +497,6 @@ RSpec.describe FileSet do
       end
       specify 'title is set' do
         expect(terms[title_key]).to eql(title)
-      end
-    end
-  end
-
-  describe 'assign_id' do
-    let(:service) { instance_double(ActiveFedora::Noid::Service) }
-    before do
-      allow(ActiveFedora::Noid::Service).to receive(:new).and_return(service)
-    end
-    context 'with noids enabled (by default)' do
-      it 'uses the noid service' do
-        expect(service).to receive(:mint).once
-        subject.assign_id
-      end
-    end
-
-    context 'with noids disabled' do
-      before { Hyrax.config.enable_noids = false }
-      after { Hyrax.config.enable_noids = true }
-
-      it 'does not use the noid service' do
-        expect(service).not_to receive(:mint)
-        subject.assign_id
       end
     end
   end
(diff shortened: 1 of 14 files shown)


jcoyne:
I'm not sure how to get these tests to run. Travis says: "The repository at projecthydra-labs/hyrax was not found."

mjgiarlo:
@jcoyne have you tried pushing a change to the branch?

spec/models/file_set_spec.rb
```diff
@@ -500,29 +501,6 @@ def paranoid_edit_permissions
     end
   end
 
-  describe 'assign_id' do
```
jcoyne: This block tests the same stuff as the block starting at line 222

escowles:
I see reasons to make using Fedora's UUIDs the default (faster, less error prone, etc.) — but I don't see any reason for deprecating NOIDs.  I've heard tell of errors minting NOIDs at scale, but have never seen these personally.

jcoyne:
@escowles It does not deprecate the use of noids, It just deprecates the tight coupling between hyrax and AF-noid.  You can still use AF noid by including `ActiveFedora::Noid::Model` on your model classes.

hackmastera:
Link to background discussion on this change? I would expect implications for the maintenance and stability of the NOID codebase over time.

jcoyne:
@hackmastera I'm not sure I understand this statement:

> I would expect implications for the maintenance and stability of the NOID codebase over time.

The only way I can see decoupling components would affect maintenance is that people who are not using those components would not be involved in their maintenance.  That seems fair, right?

I'm not sure how stability would be affected in any way.

jcoyne:
@hackmastera  I don't think there has been any prior discussion.  In hyku we had to stop using AF-noid because it was a bottleneck on large imports.

hackmastera:
Can you say more about what you observed with hyku?

mjgiarlo:
We'll be discussing on the Tech call momentarily, for folks who can dial in.

mjgiarlo:
We decided not to move forward with deprecating NOID usage in Hyrax given the community's desire to keep using NOIDs.

atz:
@mjgiarlo so we close this?

jcoyne:
@atz please don't. The feedback from this meeting was that certain parts of the PR were acceptable, but outright deprecation was not.  So I'd like to rework it when I come back to Hyrax at some point.

jcoyne:
Working on an approach to this in:
https://github.com/samvera/active_fedora/pull/1295

Which will require:
https://github.com/samvera/hydra-head/pull/425
And:
https://github.com/samvera/hyrax/pull/2256

jenlindner:
Where are things at with this? samvera/hydra-head#425 was merged, https://github.com/samvera/hyrax/pull/2256 had been blocked by two pulls that have been merged. Looks like only samvera/active_fedora#1295 is left, where there's an open question but no opposition, it seems?

cjcolvar:
@no-reply @vantuyls This issue has been open for a year.  Given that it was decided to not merge this PR as a whole is there a plan to incorporate some of these changes or can this PR be closed?

PR #1058, opened 2017-05-25
solr & fedora configs for wrappers

darrenleeweber:
Hyrax should manage it's own Solr/Fedora configs for the solr_wrapper and fcrepo_wrapper.

- add .fcrepo_wrapper
  - copied some values from `.internal_test_app/.fcrepo_wrapper`
- add .solr_wrapper
  - copied some values from `.internal_test_app/.solr_wrapper`

This makes it possible to call the wrappers from the hyrax repo checkout, rather than cd into the internal test app, to run Solr/Fedora in a separate window.

### DIFF
--- a/.fcrepo_wrapper
+++ b/.fcrepo_wrapper
@@ -1,0 +1,4 @@
+# Place any default configuration for fcrepo_wrapper here
+port: 8984
+enable_jms: false
+fcrepo_home_dir: tmp/fcrepo4-development-data
--- a/.solr_wrapper
+++ b/.solr_wrapper
@@ -1,0 +1,6 @@
+# Place any default configuration for solr_wrapper here
+# port: 8983
+collection:
+  persist: true
+  dir: solr/config/
+  name: hydra-development


.fcrepo_wrapper
```diff
@@ -0,0 +1,10 @@
+# Place any default configuration for fcrepo_wrapper here
+port: 8984
+enable_jms: false
+
+download_dir: tmp/fcrepo_wrapper
```
mjgiarlo: The `port` and `enable_jms` seem good to include here. I want to set my own `download_dir` and not have it overridden by Hyrax. You cool with removing that?

.fcrepo_wrapper
```diff
@@ -0,0 +1,10 @@
...
+
```
mjgiarlo: The blank lines look unsightly to me. Open to yanking those out? I don't think they aid readability or organization here.

.fcrepo_wrapper
```diff
@@ -0,0 +1,10 @@
...
+fcrepo_home_dir: tmp/fcrepo4-development-data
```
mjgiarlo: Could you share why we need to specify this?

.fcrepo_wrapper
```diff
@@ -0,0 +1,10 @@
...
+
+version: 4.7.3-RC-1
```
mjgiarlo: I don't think we want `master` Hyrax on the bleeding edge quite yet. I'd prefer if folks specify versions themselves, or if we pull the latest version allowed by `fcrepo_wrapper`. This is going to be a pain to remember to update, too.

.solr_wrapper
```diff
@@ -0,0 +1,12 @@
+# Place any default configuration for solr_wrapper here
+# port: 8983
+
+download_dir: tmp/solr_wrapper
```
mjgiarlo: I'm :-1: on including `download_dir` and `instance_dir`. I'd prefer people handle this locally with their own config files.

.fcrepo_wrapper
darrenleeweber: I think your own `~/.solr_wrapper` and `~/.fcrepo_wrapper` might override these files?  If not, then, yea, got to remove that here.  (Dragged it over from the travis-cache PR.)

.fcrepo_wrapper
darrenleeweber: I like em, but I don't care, they can go and I'll take em out.

.fcrepo_wrapper
darrenleeweber: I dunno actually, it's from the engine_cart version and included here in case.

.fcrepo_wrapper
darrenleeweber: OK, this is coming out.

.fcrepo_wrapper
mjgiarlo: Given the order of the files here, and the use of `Hash#merge!`, I believe app-specific config will always override `~` config: https://github.com/cbeer/fcrepo_wrapper/blob/master/lib/fcrepo_wrapper/configuration.rb#L121-L125

darrenleeweber:
Some changes pushed up, leaving commits to see changes, to be squashed when we arrive at the final versions.

.fcrepo_wrapper
```diff
@@ -0,0 +1,4 @@
+# Place any default configuration for fcrepo_wrapper here
+port: 8984
+enable_jms: false
+fcrepo_home_dir: tmp/fcrepo4-development-data
```
darrenleeweber: @cbeer - do we need to set this `fcrepo_home_dir` here?

jcoyne:
I'd like some justification in the commit message. Why do we want this change?

.solr_wrapper
```diff
@@ -0,0 +1,6 @@
+# Place any default configuration for solr_wrapper here
+# port: 8983
+collection:
+  persist: true
+  dir: solr/config/
+  name: hydra-development
```
jcoyne: Why development?  Wouldn't you want this to be for hydra-test?

.fcrepo_wrapper
jcoyne: Why development and not test?

jcoyne:
This PR seems to encourage yet another way to do testing/dev.  Will that add confusion to an already confused topic?

darrenleeweber:
I see these files in other projects and they allow running solr_wrapper and fcrepo_wrapper from the project root directory.  In hyrax, these files do not exist yet, they only appear to be auto-generated in the engine-cart .internal_test_app.  So when I want to run the solr/fcrepo wrappers, I cd into that directory.  I get comments from Chris that this is a weird practice and he seemed to suggest that it's strange these files are not in the hyrax repo already.  When they are added here, they conform to an expectation on how to configure these wrappers in the code repo.  So, now that this is being questioned too, I'm totally confused.   I thought this is the best practice for hyrax to take control over these settings for the wrappers.

darrenleeweber:
There's no extensive commit message because we use github.  If you use github features to track the blame for a change and also note the merge PR number, you get back to this commentary etc. and the description in this PR.

jcoyne:
@darrenleeweber I don't want to tie us to github.  It's nice when you're away from the network to see the rational in the commit messages.  Furthermore, we can't guarantee that github is going to offer us a free account (or even exist) as long as we need it.  Putting everything in git would enable us to go elsewhere.

mjgiarlo:
FWIW, I'm fine with adding `.solr_wrapper` and `.fcrepo_wrapper` to Hyrax, as long as we follow up by editing the appropriate wrapper-related sections of https://github.com/samvera-labs/hyrax/wiki/Hyrax-Development-Guide such that we now have a less confusing state of affairs. Sorry for the twists and turns here, @darrenleeweber.

darrenleeweber:
Squashed and pushed up commit with more detailed commit message.

darrenleeweber:
Cannot edit the development guide until we merge this.  When we do, it becomes something like these patches:

- https://github.com/samvera-labs/hyrax/wiki/Hyrax-Development-Guide#solr
```
- solr_wrapper -d solr/config/ --collection_name hydra-development
+ solr_wrapper # using .solr_wrapper config
```

- https://github.com/samvera-labs/hyrax/wiki/Hyrax-Development-Guide#fedora
```
- fcrepo_wrapper -p 8984
+ fcrepo_wrapper # using .fcrepo_wrapper config
```

PR #1063, opened 2017-05-26
Use private access

jcoyne:
I think we had a misunderstanding about what private access means in
Ruby.  It's different rules than in Java

### DIFF
--- a/app/actors/hyrax/actors/apply_permission_template_actor.rb
+++ b/app/actors/hyrax/actors/apply_permission_template_actor.rb
@@ -10,7 +10,7 @@ module Hyrax
         next_actor.create(env)
       end
 
-      protected
+      private
 
         def add_edit_users(env)
           return if env.attributes[:admin_set_id].blank?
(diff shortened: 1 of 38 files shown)


jcoyne:
@mjgiarlo it does say:

```
I think we had a misunderstanding about what private access means in
Ruby.  It's different rules than in Java
```

mjgiarlo:
@jcoyne I know, but that doesn't tell me what this gains us. Extra security? etc. Like I said, I won't block over this.

jcoyne:
@mjgiarlo it gains us nothing. It was a mistake to use protected in the first place.

jcoyne:
Say we have an class Foo: 

```ruby
class Foo
   def call_bar(obj)
     obj.bar
   end

   protected
   def bar
   end
end

foo1 = Foo.new
foo2 = Foo.new

foo1.call_bar(foo2)
#=> nil

class Baz
   def call_bar(obj)
     obj.bar
   end

   private
   def bar
   end
end


baz1 = Baz.new
baz2 = Baz.new

baz1.call_bar(baz2)
NoMethodError: private method `bar' called for #<Baz:0x007ffcef8afe78>
```

mjgiarlo:
@jcoyne OK. Thanks for clarifying.

PR #1083, opened 2017-05-28
Visindex

### LINKED ISSUE #485: Index visibility as a facet field
This will allow us to filter on different visibility levels as well as simplify SolrDocumentBehavior.

This will be used for https://github.com/projecthydra-labs/hyku/issues/447

atz:
fixes #485

### DIFF
--- a/app/indexers/hyrax/collection_indexer.rb
+++ b/app/indexers/hyrax/collection_indexer.rb
@@ -6,6 +6,8 @@ module Hyrax
 
     self.thumbnail_path_service = Hyrax::CollectionThumbnailPathService
 
+    # @yield [Hash] calls the yielded block with the solr document
+    # @return [Hash] the solr document WITH all changes
     def generate_solr_document
       super.tap do |solr_doc|
         # Makes Collections show under the "Collections" tab
@@ -13,6 +15,7 @@ module Hyrax
         # Index the size of the collection in bytes
         solr_doc[Solrizer.solr_name(:bytes, STORED_LONG)] = object.bytes
         solr_doc['thumbnail_path_ss'] = thumbnail_path
+        solr_doc['visibility_ssi'] = object.visibility
 
         object.in_collections.each do |col|
           (solr_doc['member_of_collection_ids_ssim'] ||= []) << col.id
--- a/app/indexers/hyrax/file_set_indexer.rb
+++ b/app/indexers/hyrax/file_set_indexer.rb
@@ -17,7 +17,8 @@ module Hyrax
         solr_doc['all_text_timv'] = object.extracted_text.content if object.extracted_text.present?
         solr_doc['height_is'] = Integer(object.height.first) if object.height.present?
         solr_doc['width_is']  = Integer(object.width.first) if object.width.present?
-        solr_doc['mime_type_ssi'] = object.mime_type
+        solr_doc['visibility_ssi'] = object.visibility
+        solr_doc['mime_type_ssi']  = object.mime_type
         solr_doc['thumbnail_path_ss'] = thumbnail_path
         # Index the Fedora-generated SHA1 digest to create a linkage between
         # files on disk (in fcrepo.binary-store-path) and objects in the repository.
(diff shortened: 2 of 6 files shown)


atz:
CI is intermittently but frequently busted by:
~ timing out because build matrix now has 2 rails versions (execution time doubled)
~ flapping capybara tests

Now that it takes 90 minutes for a build just to fail randomly, we need serious attention to test performance, matrix dimensions and dependability.

jcoyne:
It doesn't seem like "convert `Solrizer.solr_name` calls into literals" is required to fix #485. Can you move that to a separate pull request?  I think that might be a change that should be considered by a wider community and I don't want that to hold up #485.

jcoyne:
@atz regarding the timeout, one of the builds on this PR only took 35min for both jobs to complete: https://travis-ci.org/samvera-labs/hyrax/builds/236578030

atz:
Not interested in moving it to a separate PR, and I will continue to implement the same kind of changes at every opportunity.

atz:
@jcoyne Our times across all branches vary, but are recently averaging more than an hour: https://travis-ci.org/samvera-labs/hyrax/builds

atz:
@jcoyne Two of the three indexers I touched already used some literal field names, so what is the "wider community" issue at hand here?  Should we feel *obligated* to mix and match as many possible methods with as much indirection and contrived terminology as possible... for doing the same thing?  For whose benefit?

jcoyne:
@atz I think you could put out an email to the list and explain your reasoning and ask "has anyone customized the behavior of `solr_name` so that it returns something custom"  I think it's rather unlikely that anyone has. My concern is more that PR should focus on changing one thing and be small if possible. That makes it easier on the reviewer.

mjgiarlo:
I'd like to get more folks to sound off on moving away from `solr_name`. Tagging @samvera-labs/hyrax-code-reviewers

escowles:
Unless we can find someone who's using customization of `solr_name` to good effect, I'm 👍  to just using the real Solr field names.  I personally find `solr_name` more trouble than it's worth.

cjcolvar:
I feel like I've been bitten by using `solr_name` before and always forget what the descriptors map to so I decided to try and stay away from it.  But I found I needed to use the descriptors even more when we switched most of our metadata to RDF and used `property` `index.as`.  So I would like to see the change to literal field names happen there as well.  That all being said, I agree with @jcoyne that this issue feels like a separate PR than the current one.

awead:
This PR says it fixes _Index visibility as a facet field_, but now it's morphed into replacing `solr_name`, so I think it should be a separate issue and PR with more discussion.

elrayle:
We don't customize solr_name, but it is worth noting, that the whole index.as functionality and the generation of the solr_name is to provide accurate and consistent extensions off the property name such that the correct dynamic field is applied when the field is added to solr.  If the user always writes their own solr name, then they must know that `index.as stored_searchable` and `multiple=true` translates to the `_tesim`, in that order.  

Ref:  http://www.rubydoc.info/gems/solrizer/3.4.0/Solrizer/DefaultDescriptors

elrayle:
+1 to PRs that tackle only one thing at a time.  Smaller and easier to review.  I'm working on doing that myself.

atz:
This affects `solr_name` **only** in the classes I edited.  It would be a separate task to take it out of the rest of the application as a whole, which I am perfectly happy to do.  But I also don't feel compelled to maintain multiple different ways of doing the same thing in the files I touch here, because that is inherently bad form.  

I hear some preference for separate PRs, but literally nobody defending the use of `solr_name` (correct me if I am mischaracterizing anyone).  With due sympathy for reviewers, this is among the simplest PRs to review.  All at the same level.  Well defined commits.  The fact that it *could* be simpler doesn't seem compelling when it is already *quite simple enough*.  Like please let's not split hairs about how the diff took 3 minutes to read instead of 2.  

@elrayle: A developer **still has access** to use `solr_name` interactively if they are uncertain about naming a field.  But if we want the field names to be consistent **to the application** over its lifetime, then using anything dynamic at runtime is counter-productive.

mjgiarlo:
In general I concur with the folks in this thread about keeping your changes as few as necessary. In practice, since I'm hearing we should move away from using `solr_name` in Hyrax, I am not inclined to ask @atz to make revisions. Especially if @atz would like to put in a follow-on PR to nuke the remainder of `solr_name` usage.

PR #1099, opened 2017-05-30
Unify files that live in both root directory and .github directory

mjgiarlo:


jeremyf:
@mjgiarlo How does this play with the https://github.com/samvera/hydra/blob/master/script/push_contributing_document.rb

(Not that I want to block this PR, just curious about effort we could put forward for all other repositories)

atz:
@jeremyf It appears not to.  Also seems odd that folks were just claiming they "don't want to be tied to github" on Darren's PR, but then don't mind using github-specific features for their own ends.  That being said, I'm fine w/ the changes if we can just be consistent w/ our tooling and expectations.

mjgiarlo:
@atz As I see it, there's a difference between sequestering a PR rationale in a GitHub comment (vs. the git commit message) and `git mv CONTRIBUTING.md .github/`. The docs in `.github/` are still present in your working tree and are still tracked by version control. In a post-GitHub scenario, the latter is trivial to revert; getting comments back out of GitHub comments and into commit messages, OTOH, is costly.

@jeremyf Perhaps the push script should include a step to move the file to `.github/` if it's not already there?

PR #1109, opened 2017-05-31
Add facet path for dashboard/works.

### LINKED ISSUE #1107: "More" facet link shows incorrect facets/counts for "All works" in dashboard.
### Descriptive summary

(Hyrax HEAD)
Clicking the "More" link for facets in the "All Works" tab under Dashboard -> Works renders a modal with incorrect facet values and/or counts.

### Expected behavior

Should render facets and counts relative to "all works".

### Actual behavior

Renders facets and counts relative to "My Works".

1. Login as Admin
2. Create 6 admin sets.
3. Create a work for each admin set.
4. Go to Dashboard -> Works -> All Works and click "more" under the collection facet dropdown.

jonathandixon:
Fixes #1107 

Clicking the "More" link for facets in the "All Works" tab under Dashboard -> Works renders a modal with incorrect facet values and/or counts.

* Define `search_facet_path` for `dashboard/works_controller.rb`
* Do not use `dropdown-menu` class inside facet modal.

### DIFF
--- a/app/controllers/hyrax/dashboard/works_controller.rb
+++ b/app/controllers/hyrax/dashboard/works_controller.rb
@@ -19,6 +19,11 @@ module Hyrax
         def search_action_url(*args)
           hyrax.dashboard_works_url(*args)
         end
+
+        # The url of the "more" link for additional facet values
+        def search_facet_path(args = {})
+          hyrax.dashboard_works_facet_path(args[:id])
+        end
     end
   end
 end
--- a/app/views/hyrax/my/_facet_limit.html.erb
+++ b/app/views/hyrax/my/_facet_limit.html.erb
@@ -1,4 +1,4 @@
-  <ul class="<%= 'dropdown-menu ' if request.format.html? %>facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
+  <ul class="<%= 'dropdown-menu ' if action_name == 'index' %>facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
     <% paginator = facet_paginator(facet_field, display_facet) %>
     <%= render_facet_limit_list paginator, facet_field.key %>
 
--- a/app/views/hyrax/my/_facet_limit.html.erb
+++ b/app/views/hyrax/my/_facet_limit.html.erb
@@ -1,4 +1,4 @@
-  <ul class="dropdown-menu facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
+  <ul class="<%= 'dropdown-menu ' if request.format.html? %>facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
     <% paginator = facet_paginator(facet_field, display_facet) %>
     <%= render_facet_limit_list paginator, facet_field.key %>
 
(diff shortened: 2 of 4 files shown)


app/views/hyrax/my/_facet_limit.html.erb
```diff
@@ -1,4 +1,4 @@
-  <ul class="dropdown-menu facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
+  <ul class="<%= 'dropdown-menu ' if request.format.html? %>facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
```
atz: We're in an HTML template.  So how would we get here if the request format wasn't HTML?

jonathandixon: This template is also used by AJAX requests to render facets in a modal panel. In that case, the request format is `json`.

jcoyne: How about testing for `action_name` instead? I think that would be a lot clearer as to the intention.

jonathandixon: You are right, I think that does make my intentions clearer.

darrenleeweber:
Seems OK to me.

spec/controllers/hyrax/dashboard/works_controller_spec.rb
```diff
@@ -0,0 +1,11 @@
+RSpec.describe Hyrax::Dashboard::WorksController, type: :controller do
+  describe "#search_builder_class" do
+    subject { controller.search_builder_class }
+    it { is_expected.to eq Hyrax::WorksSearchBuilder }
+  end
```
darrenleeweber: This spec ^^ might conflict with #769

jonathandixon: Should I just remove this spec?

jcoyne: no, don't remove the spec.. GH shows you when there is merge conflicts in a PR and won't allow merging if there will be a problem.

app/views/hyrax/my/_facet_limit.html.erb
```diff
@@ -1,4 +1,4 @@
-  <ul class="dropdown-menu facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
+  <ul class="<%= 'dropdown-menu ' if action_name == 'index' %>facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
```
jcoyne: I'd like to see a comment added here explaining why we need this check. Otherwise someone might come along and remove this in the future because they don't understand it's purpose.

app/controllers/hyrax/dashboard/works_controller.rb
```diff
@@ -19,6 +19,11 @@ def ensure_admin!
         def search_action_url(*args)
           hyrax.dashboard_works_url(*args)
         end
+
+        # The url of the "more" link for additional facet values
+        def search_facet_path(args = {})
+          hyrax.dashboard_works_facet_path(args[:id])
```
jeremyf: Should this leverage `args.fetch(:id)`? It would fail with KeyError. I don't know how `dashboard_works_facet_path` handles a `nil` parameter.

jonathandixon: Without an `id` parameter `dashboard_works_facet_path` throws `ActionController::UrlGenerationError`.  I'm not opposed to this change, but it's worth noting that are other controllers that use `args[:id]` in `search_facet_path`.

PR #1115, opened 2017-06-01
Params permitted in ContentBlocks controller should match those sent by form

### LINKED ISSUE #1072: Content blocks not saving content
### Descriptive summary
Adding content to content blocks and saving does not actually save them. 

### Expected behavior
On save, the content of a content block should be saved.

### Actual behavior
On save, the content of the content block is not saved. 

1. enter text into a content block
2. save the block
3. go to the view for the content block and note that it is not saved

I have confirmed that this works on the Hyku test environment. One major difference i see between the two systems is that when one selects Content Block from the Dashboard, one is not immediately entered into one of the content blocks for editing. In Hyku, one is dropped right into the About Page content block.

mjgiarlo:
This was a goof of mine -- I changed the controller without adjusting its tests, so we missed this.

Fixes #1072

### DIFF
--- a/app/controllers/hyrax/content_blocks_controller.rb
+++ b/app/controllers/hyrax/content_blocks_controller.rb
@@ -23,9 +23,9 @@ module Hyrax
     private
 
       def permitted_params
-        params.require(:content_block).permit(:marketing_text,
-                                              :announcement_text,
-                                              :featured_researcher)
+        params.require(:content_block).permit(:marketing,
+                                              :announcement,
+                                              :researcher)
       end
 
       # When a request comes to the controller, it will be for one and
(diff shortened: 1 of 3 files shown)


atz:
What explains this difference in Travis?
```
 3282.1 Ruby: 2.4.1 RAILS_VERSION=5.0.3 18 min 45 sec
 3282.2 Ruby: 2.4.1 RAILS_VERSION=5.1.1 49 min 58 sec
```

mjgiarlo:
Is it the Rails 5.1 builds with the timeout issues, consistently? If so, I don't care that much about being on the bleeding edge.

jcoyne:
I think it has far more to do with how busy the vm that our travis build is running on than anything we have control over.

jcoyne:
I re-ran the 5.1 build and it completed in under 24 minutes.

PR #1118, opened 2017-06-02
Add --fail-fast option to .rspec configuration

### LINKED ISSUE #1117: Development - add a fail-fast option to the travis builds
### Descriptive summary

The travis builds take a very long time.  Adding a fail-fast option might help to speed things up.

darrenleeweber:
Fixes #1117 

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --color
 --require spec_helper
+--fail-fast


jcoyne:
I'd prefer we don't do this. If we pay the cost for test set up (about 12 min) I want to see all the things that are failing

atz:
I agree w/ Justin on this one.  Might as well see all the broken things at once.  Otherwise, you are having to fail many times serially just to get the same information.

jeremyf:
I also agree with @jcoyne. I find the `.rspec` to be a place of surprises and not immediately obvious behavior. I would rather provide documentation that explains how to add the `--fail-fast` to the integration test.

From https://apidock.com/rspec/Spec/Rake/SpecTask

```
$ bundle exec rake SPEC_OPTS="--fail-fast"
```

mjgiarlo:
Given the above feedback, this is not a change we're going to pursue at this time. Thanks for your efforts to speed up the build, @darrenleeweber!

PR #1119, opened 2017-06-02
Remove unused ClassifyConcern class

jcoyne:
This class mostly consisted of unused code that was originally from Curate.


mjgiarlo:
@tpendragon @jeremyf ?

tpendragon:
We use `Hyrax::ClassifyConcern.new.all_curation_concern_classes`, but we can always change it to what's in this PR. Breaking change, though.

jcoyne:
@tpendragon can you show me where you're using it?

mjgiarlo:
https://github.com/pulibrary/plum/blob/ce34a948414084fc49c8e34cf66ef28d9bcd20f9/app/controllers/welcome_controller.rb

tpendragon:
@jcoyne It's trivial, I'm not blocking this PR over it: https://github.com/pulibrary/plum/blob/master/app/controllers/welcome_controller.rb#L3. It's just the definition of breaking - seems safe to ship with 2.0 though.

jcoyne:
@tpendragon Sounds good. I think `Hyrax.config.curation_concerns` would be a reasonable replacement, right?

tpendragon:
@jcoyne Yes, like I said I can change it to that.

jeremyf:
Yes. This is something to remove. At one point we, at Notre Dame, needed to handle what work types existed and what were "on the horizon". This class can go away!

atz:
Consensus achieved.

PR #1120, opened 2017-06-02
RSpec configuration to fail fast when running on travis.ci

### LINKED ISSUE #1117: Development - add a fail-fast option to the travis builds

darrenleeweber:
An alternative implementation to speed up travis builds, fix #1117

*  rspec configuration to fail-fast, only on travis

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -147,6 +147,8 @@ require 'active_fedora/cleaner'
 RSpec.configure do |config|
   include ActiveFedora::Noid::RSpec
 
+  config.fail_fast = true if ENV['TRAVIS']
+
   config.disable_monkey_patching!
   config.include Shoulda::Matchers::ActiveRecord, type: :model
 


mjgiarlo:
@darrenleeweber I don't think we want to fail fast in CI as discussed on yesterday's related PR. That's fine if you want to make rspec fail fast locally, though. Thanks.

darrenleeweber:
I wasn't entirely clear about whether the concerns were an implementation detail or not.  The previous implementation applied the fail-fast option everywhere and it was in a dot-file (.rspec) where it may not be obvious.  The intention of this PR is to revise the implementation so it is specific to travis, where builds are timing out and we do not know if they pass or fail.  The result of this implementation is that local/laptop specs run to completion, but travis will fail fast.  When travis fails, run it locally if you want more details.  So, additional commentary on this revised implementation would be appreciated.

mjgiarlo:
@darrenleeweber I hear you, but I was hearing on the other PR that folks expect Travis builds to run the full suite. This is a technical solution that runs counter to our collective expectations, so I'm inclined to keep this closed for now.

darrenleeweber:
Yes, indeed, if the consensus is for travis to run all the specs every time, then any implementation details on this are irrelevant.  But I did hear some concerns that sounded to me like implementation details.  Anyhow, happy to defer to your judgment on the commentary.  Ask for additional verdicts or close at will, no problem.

mjgiarlo:
Thanks for the flexibility, @darrenleeweber.

PR #1124, opened 2017-06-02
Don't create a user proxy that exists

jcoyne:
Fixes https://github.com/samvera-labs/hyku/issues/1199


atz:
Seems like the single CI failing test might actually be related?
https://travis-ci.org/samvera-labs/hyrax/jobs/238903484#L2376-L2384

PR #1144, opened 2017-06-05
Make collection edit page wording consistent

laritakr:
Fixes https://github.com/samvera-labs/hyrax/issues/1020

* Change "save" button text to refer to "update collection".
* Make "Edit" and "Edit Access" showing of permissions both say
"Edit access"
* Add a new method (permission_options), to be used for the dropdown selection of rights.
* Apply the new method to all partials which use a dropdown selection of rights.

### DIFF
--- a/app/views/hyrax/collections/_form_share.html.erb
+++ b/app/views/hyrax/collections/_form_share.html.erb
@@ -6,7 +6,7 @@
   </legend>
 
   <div class="alert alert-info hidden" id="save_perm_note">Permissions are
-  <strong>not</strong> saved until the &quot;Save&quot; button is pressed at the bottom of the page.</div>
+  <strong>not</strong> saved until the &quot;Update Collection&quot; button is pressed at the bottom of the page.</div>
 
   <div class="alert alert-warning hidden" role="alert" id="permissions_error">
     <span id="permissions_error_text"></span>
@@ -52,7 +52,7 @@
     </div>
     <div class="col-sm-4">
       <label for="new_user_permission_skel" class="sr-only">Access type to grant</label>
-      <%= select_tag 'new_user_permission_skel', options_for_select(Hyrax.config.permission_levels), class: 'form-control' %>
+      <%= select_tag 'new_user_permission_skel', options_for_select(Hyrax.config.permission_options), class: 'form-control' %>
     </div>
     <div class="col-sm-3">
       <button class="btn btn-default" id="add_new_user_skel">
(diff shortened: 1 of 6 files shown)


app/views/hyrax/collections/_form_share.html.erb
```diff
@@ -6,7 +6,7 @@
   </legend>
 
   <div class="alert alert-info hidden" id="save_perm_note">Permissions are
-  <strong>not</strong> saved until the &quot;Save&quot; button is pressed at the bottom of the page.</div>
+  <strong>not</strong> saved until the &quot;Update Collection&quot; button is pressed at the bottom of the page.</div>
```
mjgiarlo: Would you mind i18nizing this and adding `es` and `zh` translations, @laritakr?

lib/hyrax/configuration.rb
```diff
@@ -381,14 +381,20 @@ def google_analytics_id
 
     attr_writer :permission_levels
     def permission_levels
-      @permission_levels ||= { "Choose Access" => "none",
-                               "View/Download" => "read",
-                               "Edit" => "edit" }
+      @permission_levels ||= { "View/Download" => "read",
```
mjgiarlo: Would you also consider i18nizing the label strings in the methods you touched here too?

This may require undoing the memoization, to allow for translation to occur every time the method is called.

mjgiarlo:
Thanks for the PR, @laritakr. I have a few quick questions :point_up:

lib/hyrax/configuration.rb
laritakr: Yeah the i18n work was a much larger change... that's actually why I brought up the question about it in standup before I went on vacation. I can do that if it's what you want, but was just trying to get this change through quickly.

app/views/hyrax/collections/_form_share.html.erb
laritakr: This would be a fairly simple change, so I could add it in... the only reason I didn't was because redoing the configuration changes were a bit more of a change and I wasn't sure how much time I had to spend on this issue. It seemed to make sense to do the translations all at once.  I assume this is a change you want made for all three partials?

lib/hyrax/configuration.rb
laritakr: Also... in order to satisfy @ggeisler 's text differences in the two cases (dropdowns to add a new person/group vs showing authority previously given to a user/group,) I needed to have the additional method. 

However there may be a cleaner way that I could use via i18n if I do the other changes for the other languages. I'll take a look at that in the next day or two as I have time between the other work.

mjgiarlo: OK, thanks!

app/views/hyrax/collections/_form_share.html.erb
mjgiarlo: That'd be excellent!

laritakr:
Given that I had already added this to the spanish translation issue, I just added a link to this PR for additional reference on that issue. Translations should be able to be done at a later point.

PR #1151, opened 2017-06-06
Add new feature toggle allowing disabling account creation

mjgiarlo:
When account creation is disabled, redirect to the homepage with an alert. When enabled, let Devise do its thing. This implementation relies on a small routing change that lets Hyrax override two methods in the Devise RegistrationsController. I am on the fence as to whether Hyrax should automatically handle this for users (i.e., we should move the code from the test_app_generator to the install_generator), or if we should rely on documentation and release notes for this. There is a chance that this flipper will be made available in instances that have not made this routing change, which would be misleading for users.

Refs samvera-labs/hyku#1203

### DIFF
--- a/spec/controllers/hyrax/registrations_controller_spec.rb
+++ b/spec/controllers/hyrax/registrations_controller_spec.rb
@@ -1,0 +1,43 @@
+RSpec.describe Hyrax::RegistrationsController, type: :controller do
+  routes { Rails.application.routes }
+
+  before do
+    allow(Flipflop).to receive(:account_signup?).and_return(account_signup_enabled)
+    # Recommended by Devise: https://github.com/plataformatec/devise/wiki/How-To:-Test-controllers-with-Rails-3-and-4-%28and-RSpec%29
+    @request.env['devise.mapping'] = Devise.mappings[:user]
+  end
+
+  context 'with account signup enabled' do
+    let(:account_signup_enabled) { true }
+    describe '#new' do
+      it 'renders the form' do
+        get :new
+        expect(response).to render_template('devise/registrations/new')
+      end
+    end
+    describe '#create' do
+      it 'processes the form' do
+        post :create, params: { user: { email: "user@example.org", password: "password", password_confirmation: "password" } }
+        expect(response).to redirect_to Hyrax::Engine.routes.url_helpers.dashboard_path(locale: 'en')
+        expect(flash[:notice]).to eq 'Welcome! You have signed up successfully.'
+      end
+    end
+  end
+  context 'with account signup disabled' do
+    let(:account_signup_enabled) { false }
+    describe '#new' do
+      it 'redirects with a flash message' do
+        get :new
+        expect(response).to redirect_to root_path
+        expect(flash[:alert]).to eq 'Account registration is disabled'
+      end
+    end
+    describe '#create' do
+      it 'redirects with a flash message' do
+        post :create
+        expect(response).to redirect_to root_path
+        expect(flash[:alert]).to eq 'Account registration is disabled'
+      end
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -67,4 +67,8 @@ def enable_arkivo_api
   def relax_routing_constraint
     gsub_file 'config/initializers/arkivo_constraint.rb', 'false', 'true'
   end
+
+  def override_devise_registrations_controller
```
mjgiarlo: See also: https://github.com/samvera-labs/hyrax/pull/1151#issue-233985835

This could arguably move into Hyrax's install generator to "automatically" activate this flipper in new installs. (And should be documented in the 2.0.0 release notes, so folks who are upgrading know to make this change in their installations to make this flipper work.) Would like in put on that.

spec/controllers/hyrax/registrations_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
+RSpec.describe Hyrax::RegistrationsController, type: :controller do
+  routes { Rails.application.routes }
+
+  before do
+    allow(Flipflop).to receive(:account_signup?).and_return(account_signup_enabled)
+    # Recommended by Devise: https://github.com/plataformatec/devise/wiki/How-To:-Test-controllers-with-Rails-3-and-4-%28and-RSpec%29
+    @request.env['devise.mapping'] = Devise.mappings[:user]
```
jcoyne: Isn't this recommended only if you have more than one mapping?

jcoyne: Ah, this is if you are not using https://github.com/samvera-labs/hyrax/blob/8d44caab1ec373a82fe7b079fb670193263d6308/spec/spec_helper.rb#L211

jcoyne:
I'm wondering if we want this as a flipper? It would be less complex to just comment out `:registerable` in the user.rb, right?

mjgiarlo:
@jcoyne It's super convenient in dev/testing to have `:registerable` in the user model, though. And I'm not sure we have zero adopters who use `:registerable` in production. I'm reluctant to make that change without discussing on a tech call, at least.

jcoyne:
@mjgiarlo it's something that would be in your own local hyrax install, so I don't see why we'd want to discuss it on the tech call.

mjgiarlo:
@hannahfrost @ggeisler @bbranan How do y'all feel about turning user sign-up completely off in Hyku (for local, AWS, and Docker) installs?

mjgiarlo:
@jcoyne That's going a bit beyond what the issue specified, so I've tagged some others to see if they're cool with expanding the scope to remove all user signups.

bbranan:
@mjgiarlo Turning off self-sign-up completely in hyku seems reasonable as long as there is still a path via the UI to create the initial admin user when creating a tenant (likely via the super-admin flow) and for admins to create new user accounts in the tenant.

mjgiarlo:
Discussed on Samvera Tech call today. Decided to disable `:registerable` ability in Hyku depending on an config setting.

@jcoyne I *believe* disabling registerable also removes the ability for users to edit their passwords. Do we care about that? Perhaps not since we never even expose that link in the UI? Cc: @hannahfrost @ggeisler

bbranan:
Issue https://github.com/samvera-labs/hyku/issues/435 appears to provide a way for users to reset their password. As noted, there does not appear to be a way to do this at the moment. The ability for users to reset their own passwords is necessary however, especially if we're going down the path of admins creating user accounts.

ggeisler:
@mjgiarlo Not sure if I'm completely up-to-speed with the issue being discussed here but I'd also be concerned if there wasn't a way for users to change their passwords. More generally than the case @bbranan points out, if an application enables someone to create an account with a password, I think they need to be able to update that password.

mjgiarlo:
Superseded by https://github.com/samvera-labs/hyku/pull/1233.

PR #1158, opened 2017-06-07
Copy the import-export job from CurationConcerns

jcoyne:
Update it to work with the released 0.1.0 version of
fcrepo-import-export


jcoyne:
@mjgiarlo how should I word that?  If you uncommented and set the properties `binaries_directory` or `descriptions_directory` (which weren't actually used), you now need to set `bagit_directory` instead?

mjgiarlo:
@jcoyne :shipit:

PR #1161, opened 2017-06-07
Remove unused configuration option

jcoyne:


mjgiarlo:
@jcoyne I guess this means we'll have to change Hyku here too: https://github.com/samvera-labs/hyku/blob/master/app/jobs/import_work_from_purl_job.rb#L90

jcoyne:
@mjgiarlo Yes, I'm fine with hard coding something there.

mjgiarlo:
@jcoyne OK.

PR #1182, opened 2017-06-09
Add link allowing users to change their passwords

mjgiarlo:


app/views/_user_util_links.html.erb
```diff
@@ -16,6 +16,7 @@
         <li><%= link_to t("hyrax.toolbar.dashboard.menu"), hyrax.dashboard_path %></li>
 
         <li class="divider"></li>
+        <li><%= link_to t("hyrax.toolbar.profile.edit_registration"), main_app.edit_user_registration_path %></li>
```
jcoyne: This should be configurable. This isn't a valid path for someone who is doing SSO.

mjgiarlo: How about I check if `:registerable` is in the devise mappings, rather than commit to new SSO-related configuration in Hyrax?

jcoyne: @mjgiarlo would `:database_authenticatable` be more appropriate?

mjgiarlo:
@jcoyne The change password link is now only rendered if `:registerable` is included among Devise's mappings.

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -22,6 +22,10 @@ def available_translations
 
     delegate :name, :name_full, to: :institution, prefix: :institution
 
+    def registration_available?
```
jcoyne: Can we make this `change_password?`?

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -22,6 +22,10 @@ def available_translations
...
+      Devise.mappings[:user].registerable?
```
jcoyne: How about `database_authenticatable?`?

jcoyne: Should we test for the presence of devise?  What if someone uses omniauth (without devise)?

mjgiarlo: What if `database_authenticatable` is on and `registerable` is off?

mjgiarlo: Good call. I'm now wondering if I just should do this in Hyku, tbh, to keep this simple.

jcoyne: :+1:

PR #1184, opened 2017-06-09
Display collections rather than admin sets on the homepage

### LINKED ISSUE #1067: List collections on home page instead of admin sets
### Descriptive summary

Promoting user collections as a means of browsing the repository content, instead of admin sets

### Rationale

Admin sets as an organizing mechanism should not be made visible on the public page of the application. User collections -- created by any user permitted to create a collection -- should be used instead. 

### Expected behavior

As a public viewer of the repository, when I go to the home page, I should see user collections, not admin sets, listed under 'Explore Collections' and available in a list after I click on 'View all collections'

### Actual behavior

As a public viewer of the repository, when I go to the home page, I see admin sets when I click on 'Explore Collections" and on the Admin Sets list page after I click on 'View all collections'

jcoyne:
Fixes #1067


mjgiarlo:
@jcoyne Needs rebase.

app/views/hyrax/homepage/_explore_collections.html.erb
```diff
@@ -1,16 +1,16 @@
 <table class="table table-striped collection-highlights">
   <tbody>
-    <% @admin_sets.each do |admin_set| %>
+    <% collections.each do |collection| %>
```
darrenleeweber: `@collections`?

darrenleeweber:
A littel surprised to see the complete removal of admin-sets CV (in MVC) and wondering how a user accesses the admin-sets.  I was expecting to only see minor changes to the V (in MVC) for this PR.

mjgiarlo:
@darrenleeweber Good catch -- the AdminSet CV was removed only from the "public" side of the UI, not from the "admin" side. The requirement here is to make collections more visible to the public and make AdminSets less visible to the public.

PR #1186, opened 2017-06-09
[DEBUGGING] Log options passed to AddFileToFileSet

jcoyne:
This will aid us in debugging #1235

### DIFF
--- a/app/jobs/ingest_file_job.rb
+++ b/app/jobs/ingest_file_job.rb
@@ -10,6 +10,7 @@ class IngestFileJob < Hyrax::ApplicationJob
   def perform(file_set, filepath, user, opts = {})
     relation = opts.fetch(:relation, :original_file).to_sym
 
+    logger.debug("Performing ingest of \"#{filepath}\" with options: #{opts.inspect}")
     # Wrap in an IO decorator to attach passed-in options
     local_file = Hydra::Derivatives::IoDecorator.new(File.open(filepath, "rb"))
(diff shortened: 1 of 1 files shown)


jcoyne:
I'm not sure if we need to merge this. I suspect the problem is actually in characterization because of this log message:
```
D, [2017-06-08T16:51:13.735391 #25523] DEBUG -- : [19f5b4bd-949b-4d36-bb70-ac1c3ad7ea6d] [ActiveJob] [CharacterizeJob] [5ceb499f-5d0f-4787-9054-41c192a1634f] Ran characterization on 40e8273e-1ea8-48ec-916c-e78bc06ba1ea/files/b5ce5a15-4942-45c1-9e47-05b6a8cc0dc1 (cannot open `/tmp/40e8273e-1ea8-48ec-916c-e78bc06ba1ea20170608-25523-m6g42x' (No such file or directory))
```

mjgiarlo:
@jcoyne :speech_balloon: 
> I'm not sure if we need to merge this. I suspect the problem is actually in characterization because of this log message:

Should we keep this PR around still?

darrenleeweber:
Can we mark the title of this PR with something to indicate that it doesn't require review?  Perhaps adding `[EXPERIMENTAL]` or something like that to the title?

jcoyne:
Let's keep this until we get the mime type issue sorted. I'm hoping @atz's work means I don't have to do anything

atz:
Closing this since the job itself was deleted.

PR #1204, opened 2017-06-12
Combine noid tests

jcoyne:
Ensure that the enabled bit is always left as it was before the test.

Prior to this test, some tests would fail depending on the order they
were run.

### DIFF
--- a/spec/models/file_set_spec.rb
+++ b/spec/models/file_set_spec.rb
@@ -221,34 +221,50 @@ RSpec.describe FileSet do
 
   describe 'noid integration', :clean_repo do
     let(:service) { instance_double(ActiveFedora::Noid::Service, mint: noid) }
+    let(:noid) { 'wd3763094' }
+    let!(:default) { Hyrax.config.enable_noids? }
+
     before do
       allow(ActiveFedora::Noid::Service).to receive(:new).and_return(service)
     end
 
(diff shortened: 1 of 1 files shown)


spec/models/file_set_spec.rb
```diff
@@ -221,34 +221,52 @@ class AltFile < ActiveFedora::Base
 
   describe 'noid integration', :clean_repo do
     let(:service) { instance_double(ActiveFedora::Noid::Service, mint: noid) }
+    let(:noid) { 'wd3763094' }
+    let(:default) { Hyrax.config.enable_noids? }
```
darrenleeweber: Maybe use `let!` here to ensure this happens before anything in specs (as it in a before block)?

spec/models/file_set_spec.rb
```diff
@@ -221,34 +221,52 @@ class AltFile < ActiveFedora::Base
...
-    let(:noid) { 'wd3763094' }
+    after { Hyrax.config.enable_noids = default }
 
-    subject do
-      described_class.create { |f| f.apply_depositor_metadata('mjg36') }
-    end
+    context 'with noids enabled' do
+      before do
+        Hyrax.config.enable_noids = true
+      end
```
darrenleeweber: Collapse into a one-line block, as in `before { Hyrax.config.enable_noids = true }`

Issue #1089, opened 2017-05-30
Character encoding problems in full-text indexing


ojlyytinen:
I created a simple Word document with this content "Aaa bbb åäö ÅÄÖ – aaa bbb." and tried to upload that to standard Hyrax 1.0.1 installation. I got an error with a lengthy stack trace which I've copied to https://gist.github.com/ojlyytinen/80d678701abbcc15a003a3f3386c7439. 

See also the discussion at hydra-tech at https://groups.google.com/forum/#!topic/hydra-tech/G61sJUym8VA.

I did some research into the cause of this and I believe the problem is that ```file_set.to_solr['all_text_timv'].encoding``` is ASCII-8BIT. When serialising that to json, the conversion to UTF-8 fails. The contents are actually UTF-8 encoded, it's just that it's got the wrong encoding set. If you do ```.force_encoding('UTF-8')```, you get the correct string back with all the international characters intact. (See output of a Rails console session below)

Looking further into this, the reason the encoding is ASCII-8BIT is that files read from Fedora get that encoding if they contain any characters not in 7-bit ASCII. The full-text file contains the international characters which comes back from Fedora as ASCII-8BIT. That then gets added to the Solr document and converting that to UTF-8 json fails and produces the error.

I think the way to fix this is to set the encoding correctly in ActiveFedora when loading the file from Fedora, based on what charset the mime_type has. Then Hyrax could explicitly set the mime_type of the full-text document to "text/plain; charset=utf-8' and get the contents back in the right encoding.

```ruby
f=ActiveFedora::File.new
f.content = 'abc'
f.save
f.reload
# 7-bit ASCII contents work fine and come back as UTF-8
f.content # => "abc"
f.content.encoding # => #<Encoding:UTF-8> 
f.content = 'åäö'
f.content.encoding # => #<Encoding:UTF-8>
f.save
f.reload
# With international characters, the contents come back as ASCII-8BIT but the bytes are the UTF-8 encoding of the string
f.content # => "\xC3\xA5\xC3\xA4\xC3\xB6"
f.content.encoding # => #<Encoding:ASCII-8BIT>
f.content.force_encoding('UTF-8') # => "åäö"
f.mime_type # => "text/plain"
# mime_type could be set to contain the charset
f.mime_type = 'text/plain; charset=utf-8'
f.content = 'öäå'
f.save
f.reload
f.content # => "\xC3\xB6\xC3\xA4\xC3\xA5"
[...]

PR #1205, opened 2017-06-12
Add configuration option to disable full text extract

aaron-collier:
This is related to issue #1089.

This provides an option that can be set to allow the skipping of full text generation when desired.

Note: This does not address issue #1089, but provides a work around for instances where full text extraction is causing encoding issues and full text is not required.

### DIFF
--- a/app/services/hyrax/file_set_derivatives_service.rb
+++ b/app/services/hyrax/file_set_derivatives_service.rb
@@ -49,8 +49,7 @@ module Hyrax
       def create_pdf_derivatives(filename)
         Hydra::Derivatives::PdfDerivatives.create(filename,
                                                   outputs: [{ label: :thumbnail, format: 'jpg', size: '338x493', url: derivative_url('thumbnail') }])
-        Hydra::Derivatives::FullTextExtract.create(filename,
-                                                   outputs: [{ url: uri, container: "extracted_text" }])
+        extract_full_text(filename, uri)
       end
 
       def create_office_document_derivatives(filename)
@@ -58,8 +57,7 @@ module Hyrax
                                                        outputs: [{ label: :thumbnail, format: 'jpg',
                                                                    size: '200x150>',
                                                                    url: derivative_url('thumbnail') }])
-        Hydra::Derivatives::FullTextExtract.create(filename,
-                                                   outputs: [{ url: uri, container: "extracted_text" }])
+        extract_full_text(filename, uri)
       end
 
       def create_audio_derivatives(filename)
@@ -82,6 +80,16 @@ module Hyrax
 
       def derivative_path_factory
         Hyrax::DerivativePath
+      end
+
+      # Calls the Hydra::Derivates::FulltextExtraction unless the extract_full_text
+      # configuration option is set to false
+      # @param [String] filename of the object to be used for full text extraction
+      # @param [String] uri to the file set (deligated to file_set)
+      def extract_full_text(filename, uri)
+        return unless Hyrax.config.extract_full_text?
+        Hydra::Derivatives::FullTextExtract.create(filename,
+                                                   outputs: [{ url: uri, container: "extracted_text" }])
       end
   end
 end
--- a/lib/generators/hyrax/templates/config/hyrax.rb
+++ b/lib/generators/hyrax/templates/config/hyrax.rb
@@ -74,6 +74,10 @@ Hyrax.config do |config|
   # Path to the file derivatives creation tool
   # config.libreoffice_path = "soffice"
 
+  # Option to enable/disable full text extraction from PDFs
+  # Default is true, set to false to disable full text extraction
+  # config.extract_full_text = true
+
   # How many seconds back from the current time that we should show by default of the user's activity on the user's dashboard
   # config.activity_to_show_default_seconds_since_now = 24*60*60
 
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -373,6 +373,12 @@ module Hyrax
       @subject_prefix ||= "Contact form:"
     end
 
+    attr_writer :extract_full_text
+    def extract_full_text?
+      return @extract_full_text unless @extract_full_text.nil?
+      @extract_full_text = true
+    end
+
     private
 
       # @param [Symbol, #to_s] model_name - symbol representing the model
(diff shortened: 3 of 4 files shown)


aaron-collier:
There appears to be an existing DRY issue here that was likely before code climate (I suppose) that I'm testing the cleanup for now.

app/services/hyrax/file_set_derivatives_service.rb
```diff
@@ -83,5 +81,11 @@ def create_image_derivatives(filename)
       def derivative_path_factory
         Hyrax::DerivativePath
       end
+
+      def extract_full_text(filename, uri)
```
jcoyne: Will you please add some yard doc about the expected types.

lib/generators/hyrax/templates/config/hyrax.rb
```diff
@@ -74,6 +74,10 @@
   # Path to the file derivatives creation tool
   # config.libreoffice_path = "soffice"
 
+  # Option to enable/disable full text extraction from PDFs
+  # Default is false, set to true to skip extraction
+  config.extract_full_text = true
```
jcoyne: In the template typically we comment out the default options just to show what they are.  The default should be set in `Configuration`

lib/hyrax/configuration.rb
```diff
@@ -373,6 +373,11 @@ def subject_prefix
       @subject_prefix ||= "Contact form:"
     end
 
+    attr_writer :extract_full_text
+    def extract_full_text?
+      @extract_full_text ||= false
```
jcoyne: For backwards compatibility the default should be true, right?  I don't think you can use the standard memorization idiom with booleans (consider what happens if `@extract_full_text = false`)

lib/generators/hyrax/templates/config/hyrax.rb
```diff
@@ -74,6 +74,10 @@
   # Path to the file derivatives creation tool
   # config.libreoffice_path = "soffice"
 
+  # Option to enable/disable full text extraction from PDFs
+  # Default is false, set to true to skip extraction
```
jcoyne: This seems backwards to me 🤔 .

app/services/hyrax/file_set_derivatives_service.rb
aaron-collier: Will do.

lib/hyrax/configuration.rb
aaron-collier: Yeah - that's what I thought. I was having issues testing until this setting. I'll work it the other way and see what was up with my test.

lib/generators/hyrax/templates/config/hyrax.rb
aaron-collier: Agreed. See comment above. Was having weird testing and felt odd about the naming vs. the true/false value. I'll update this and rework it in the AM. ;-)

lib/hyrax/configuration.rb
jcoyne: Look at `@enable_noids` for ideas.

app/services/hyrax/file_set_derivatives_service.rb
aaron-collier: Done.

lib/generators/hyrax/templates/config/hyrax.rb
aaron-collier: Fixed.

lib/hyrax/configuration.rb
aaron-collier: Thanks, that resolved the odd testing issues I had.

aaron-collier:
I believe all comments were addressed.

app/services/hyrax/file_set_derivatives_service.rb
```diff
@@ -83,5 +81,15 @@ def create_image_derivatives(filename)
       def derivative_path_factory
         Hyrax::DerivativePath
       end
+
+      # Calls the Hydra::Derivates::FulltextExtraction unless the extract_full_text
+      # configuration option is set to false
+      # @param [String] filename of the object to be used for full text extraction
+      # @param [String] uri to the files set (deligated to file_set)
```
darrenleeweber: "uri to the file set (delegated to file_set)"

lib/hyrax/configuration.rb
```diff
@@ -373,6 +373,12 @@ def subject_prefix
       @subject_prefix ||= "Contact form:"
     end
 
+    attr_writer :extract_full_text
+    def extract_full_text?
+      return @extract_full_text unless @extract_full_text.nil?
+      @extract_full_text = true
+    end
```
darrenleeweber: Replace lines 378-379 inclusive with `@extract_full_text ||= true`

darrenleeweber: An alternative pattern for this method body could use a ternary expression, i.e.
```ruby
x.nil? ? x = true : x
```
The assignment in the ternary expression is valid and it returns `true`.

spec/lib/hyrax/configuration_spec.rb
```diff
@@ -55,4 +55,5 @@
   it { is_expected.to respond_to(:translate_uri_to_id) }
   it { is_expected.to respond_to(:upload_path) }
   it { is_expected.to respond_to(:work_requires_files?) }
+  it { is_expected.to respond_to(:extract_full_text?) }
 end
```
darrenleeweber: Can we also check that the default is actually `true`, both when the config value is explicitly set and when it is not?

lib/hyrax/configuration.rb
```diff
@@ -373,6 +373,12 @@ def subject_prefix
       @subject_prefix ||= "Contact form:"
     end
 
+    attr_writer :extract_full_text
```
darrenleeweber: I wonder if the writer is necessary and whether the config value should be the only way to initialize this?

app/services/hyrax/file_set_derivatives_service.rb
aaron-collier: Ugh, typo. Fixed.

lib/hyrax/configuration.rb
aaron-collier: I think this would be a reasonable change, but the above follows the pattern used throughout this particular source. I would suggest a specific change/issue to fix this seperately IMHO.

lib/hyrax/configuration.rb
aaron-collier: Good question. Following convention here, so I haven't tested otherwise but can. This would have wider implications on the file as it's the pattern used throughout.

spec/lib/hyrax/configuration_spec.rb
aaron-collier: Agree - and happy to work in a change like this. However, should it be throughout? As it's not currently setup that way?

mjgiarlo:
Agree with @aaron-collier that the changes he's made are consistent with past boolean config options, so changes to that pattern should be in a separate PR. Thx for the review, @darrenleeweber.

PR #1206, opened 2017-06-13
Replace FileIngestJob w/ synchronous method (since it failed async anyway)

### LINKED ISSUE #1155: IngestFileJob cannot be used async & parallel
`IngestFileJob#perform` takes a String `filepath` argument.  It immediately tries to do:

```ruby
File.open(filepath, "rb")
```
That ONLY works on when the file happens to be in temp, which is only possible on the same webserver that received the (uploaded) file.  In an asynchronous multi-worker setup (the kind jobs are intended to support) or even a simple setup with a single separate worker box, the file is not there.  There is no failover: an exception is thrown.  Even on a single box, there is no guarantee a slowly processing queue will service the job via `perform_later` before temp is cleaned up.  

So this job violates the constraints that allow a job to be used synchronously or asynchronously.  It must be rewritten.  It is effectively just a Service that only works synchronously.

atz:
Fixes #1155 

A major change necessitated by `IngestFileJob`'s reliance on local path information.  Jobs do not get to pick what worker services them or when. So the logic cannot depend on the caller prepping a local file in temp for the job to execute upon. The caller can supply the path as a hint, like "Use this cached content IF it exists for you", but it cannot be guaranteed (must be retrievable by other means/identifiers).

Notable changes:
- Use `Hydra::Derivatives::IoDecorator`
- sync/async boolean goes away, since async was not fully supported
- More robust testing
- `FileActor` ONLY acts upon the relation it was initialized with, no more backdoor kludgey overrides.  If you want a different relation, just get a new actor.  They're cheap.
- Better mime-type sniffing, mostly affecting local file uploads of the type end users cannot perform.


app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
...
-                 else
-                   :perform_now
-                 end
-
-        IngestFileJob.send(method,
-                           file_set,
-                           working_file(file),
-                           user,
-                           ingest_options(file))
+      # Spawns a job to characterize and create derivatives.
```
jcoyne: Yes, it does do this, but more importantly, it persists the file in the repository.

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
...
+      # @param [Hydra::Derivatives::IoDecorator] io the file to save in the repository, with mime_type and original_name
+      # @return [Boolean] true on success (note: does NOT mean spawned job completed), false otherwise
+      # @example for File
+      #   file_actor = Hyrax::Actors::FileActor.new(fileset, :original_file, user)
+      #   file = File.open('/tmp/mydir/upload.jpg', 'rb')
+      #   io = Hydra::Derivatives::IoDecorator.new(file, 'image/jpeg', 'spirostomum_2400x1200.jpg')
+      #   file_actor.ingest_file(io)
+      # @example for Tempfile
+      #   tempfile = get_tempfile_from_request # presume you get a Tempfile from Rack, Rails or wherever
```
jcoyne: `get_` would not be idiomatic Ruby.

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
...
+      #   io = Hydra::Derivatives::IoDecorator.new(tempfile, 'image/jpeg', File.basename(tempfile.path))
+      #   file_actor.ingest_file(io)
+      # @example for ActionDispatch::Http::UploadedFile
+      #   file = ActionDispatch::Http::UploadedFile.new(filename: '汉字.jpg', type: 'image/jpeg', tempfile: tempfile)
+      #   io = Hydra::Derivatives::IoDecorator.new(file, file.content_type, file.original_filename)
+      #   file_actor.ingest_file(io)
+      # @note the primary requirement of the file passed to Hydra::Derivatives::IoDecorator is to support #read
+      # @todo create a job to monitor the temp directory (or in a multi-worker system, directories!) to prune old files that have made it into the repo
+      def ingest_file(io)
```
jcoyne: You've changed the number of methods.  Perhaps you should keep two and deprecate usage of the second.  e.g. "calling ingest_file with two arguments is deprecated. The second argument is no longer used..."

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
...
-      private
-
-        def working_file(file)
-          path = file.path
-          return path if File.exist?(path)
-          Hyrax::WorkingDirectory.copy_file_to_working_directory(file, file_set.id)
-        end
+      # @return [Hydra::PCDM::File] the file referenced by relation
+      def related_file
+        file_set.send(relation) || raise("No .#{relation} returned for FileSet #{file_set.id}")
```
jcoyne: This should probably use `public_send`. Is the period in the message intentional?

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
...
+      def ingest_file(io)
+        # Skip versioning because versions will be minted by VersionCommitter as necessary during save_characterize_and_record_committer.
+        Hydra::Works::AddFileToFileSet.call(file_set,
+                                            io,
+                                            relation,
+                                            versioning: false)
+        return false unless file_set.save
+        repository_file = related_file
+        Hyrax::VersioningService.create(repository_file, user)
+        CharacterizeJob.perform_later(file_set, repository_file.id, path_for(io)) # path hint in case next worker is on same box
```
jcoyne: `s/box/file system/`

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
...
+        file_set.send(relation) || raise("No .#{relation} returned for FileSet #{file_set.id}")
+      end
 
-        def ingest_options(file, opts = {})
-          opts[:mime_type] = file.content_type if file.respond_to?(:content_type)
-          opts[:filename] = file.original_filename if file.respond_to?(:original_filename)
-          opts.merge!(relation: relation)
-        end
+      # @param [Hydra::Derivatives::IoDecorator] io
+      def path_for(io)
```
jcoyne: Add `# @return [String]`

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
       attr_reader :file_set, :relation, :user
 
       # @param [FileSet] file_set the parent FileSet
-      # @param [String] relation the type/use for the file.
+      # @param [Symbol, #to_sym] relation the type/use for the file
```
jcoyne: I don't understand why the type of this argument is changing.

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -5,62 +5,66 @@ class FileActor
...
 
-      private
-
-        def working_file(file)
-          path = file.path
-          return path if File.exist?(path)
-          Hyrax::WorkingDirectory.copy_file_to_working_directory(file, file_set.id)
-        end
+      # @return [Hydra::PCDM::File] the file referenced by relation
+      def related_file
```
jcoyne: Should this be a private method?

app/actors/hyrax/actors/file_actor.rb
jcoyne: Should this method be private?

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -31,14 +31,14 @@ def create_metadata(file_set_params = {})
 
       # Called from FileSetsController, AttachFilesToWorkJob, ImportURLJob, IngestLocalFileJob
       # @param [File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user.
-      # @param [String] relation ('original_file')
-      # @param [Boolean] asynchronous (true) set to false if you don't want to launch a new background job.
-      def create_content(file, relation = 'original_file', asynchronous = true)
+      # @param [Symbol, #to_sym] relation
+      # @return [Boolean] true on success, false otherwise
+      def create_content(file, relation = :original_file)
```
jcoyne: Should there be a deprecation of the third argument?  Why did the type of the second argument change?

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -110,12 +112,28 @@ def build_file_actor(relation)
           file_actor_class.new(file_set, relation, user)
         end
 
+        def io_decorator(file)
```
jcoyne: private methods?

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -110,12 +112,28 @@ def build_file_actor(relation)
...
+        # If the file was imported via URL, parse the original filename.
+        # If all else fails, use the basename of the file where it sits.
         def label_for(file)
-          # For the label, use the original filename if it's there.
-          # If the file was imported via URL, parse the original filename
-          # If all else fails, use the basename of the file where it sits
-          if file.respond_to?(:original_filename)
+          if file.respond_to?(:original_filename) # e.g. ActionDispatch::Http::UploadedFile, CarrierWave::SanitizedFile
             file.original_filename
+          elsif file.respond_to?(:original_name) # e.g. Hydra::Derivatives::IoDecorator
```
jcoyne: the `file` argument to `update_content()` and `create_content()` do not permit `IoDecorator` in the YARD doc.  Do we need this clause?

spec/factories/file_sets.rb
```diff
@@ -22,39 +22,6 @@
       read_groups ["registered"]
     end
 
-    factory :public_pdf do
```
jcoyne: 👏

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -110,12 +112,28 @@ def build_file_actor(relation)
           file_actor_class.new(file_set, relation, user)
         end
 
+        def io_decorator(file)
+          Hydra::Derivatives::IoDecorator.new(file, mime_for(file), label_for(file))
+        end
+
+        def mime_for(file)
+          if file.respond_to?(:content_type)
```
jcoyne: Can you add documentation about which classes have which methods as you've done in `label_for`?

app/actors/hyrax/actors/file_actor.rb
atz: It's not materially different.  String is fine. Symbol is fine.  The point is that it is only used as a thing that gets `.to_sym` called on it anyway.

app/actors/hyrax/actors/file_actor.rb
atz: This seriously concerns you at the level of PR review?

app/actors/hyrax/actors/file_set_actor.rb
atz: Same as above.  Not a breaking change, just a clarification of documentation.

app/actors/hyrax/actors/file_actor.rb
atz: I changed everywhere we call it.  I would consider changing the synchronicity/asynchronicity of an invocation a breaking change anyway.  A deprecation warning doesn't really solve that any better ("You asked for X, here's Y").  I opted for the breaking change that provides the cleanest method.

app/actors/hyrax/actors/file_actor.rb
atz: It could be.

app/actors/hyrax/actors/file_actor.rb
atz: Previous code used send in multiple places.  I just consolidated it to this method.  I guess I agree, but also don't feel like it is required of this work. Easy to change though.

Yes, intentional message, though perhaps not the clearest.  Open to rephrasing.

app/actors/hyrax/actors/file_actor.rb
atz: I'm amenable to both those.

app/actors/hyrax/actors/file_actor.rb
jcoyne: Yeah, so why not use a realistic example. Is it so hard to remove 4 characters?

app/actors/hyrax/actors/file_actor.rb
jcoyne: There is a good chance that this is called outside of the Hyrax context.  This is a public method in a public class.  You should provide deprecations.

app/actors/hyrax/actors/file_actor.rb
jcoyne: Is there any reason to keep it public?

app/actors/hyrax/actors/file_actor.rb
atz: Because "realism" is not the point here.

app/actors/hyrax/actors/file_actor.rb
atz: It's still a breaking change to ignore the argument for something as important as *when* and *where* execution happens.  Deprecation doesn't help that.

app/actors/hyrax/actors/file_actor.rb
mjgiarlo: Agree that `get_` is generally non-idiomatic Ruby, though the example here isn't egregious because it's not using `get` in a `getter` context. It's using `get` like `extract`. I think this is acceptable.

app/actors/hyrax/actors/file_actor.rb
mjgiarlo: `s/file system/filesystem/`

app/actors/hyrax/actors/file_set_actor.rb
atz: @jcoyne Already private.  See line 105.

app/actors/hyrax/actors/file_set_actor.rb
atz: I found it useful.  In the long term I hope to get rid of `IoDecorator` altogether though.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -116,12 +116,13 @@ def io_decorator(file)
           Hydra::Derivatives::IoDecorator.new(file, mime_for(file), label_for(file))
         end
 
+        # @return [String, nil] mime type explicitly set or sniffed
         def mime_for(file)
-          if file.respond_to?(:content_type)
+          if file.respond_to?(:content_type) # e.g. ActionDispatch::Http::UploadedFile, CarrierWave::SanitizedFile
             file.content_type
-          elsif file.respond_to?(:mime_type)
+          elsif file.respond_to?(:mime_type) # e.g. Hydra::Derivatives::IoDecorator
```
jcoyne: I don't think we need to support `IoDecorator` here.

jcoyne: But, this was already here. 👍

atz:
@tpendragon Then you are committed to making IngestFileJob work for *our* use case, which it claims to support but doesn't.

atz:
@tpendragon: It is unclear what changes you are requesting other than "worksforme/wontfix".  You seem to be saying "cannot change IngestFileJob", but IngestFileJob **must** change.  Since the file is only identified by the path, it cannot be retrieved independently and the job cannot be used without a filesystem shared with the caller.  

This PR fixes a deterministic failure (thrown exception) from a user submitted bug report.  Can you identify what changes you are requesting?

escowles:
@atz Right now there are three scenarios:
1. Single worker (works)
2. Multiple workers, with shared filesystem (works)
3. Multiple workers, without a shared filesystems (doesn't always work)

IMHO, a much less disruptive way of fixing #3 is to subclass `FileActor` with the inline ingest functionality, and configure your app to use that instead of the default `FileActor`.

atz:
@escowles: (3) is *deterministically fails*

mjgiarlo:
My sense of the concerns isn't so much around the approach taken in this PR but about when it lands in `master`, with folks asking that the follow-on work begin before it's merged.

hackmastera:
I'm not sure it's clear to me what the follow-on work is, though. Can someone explain that either here or in an issue?

I thought this PR removed the job entirely, but then found out it replaces it with an inline invocation. So is a matter of adding the job back in or something else?

atz:
I might counter-propose that users who rely on the undocumented shared worker filesystem behavior should add configuration and docs to _optionally_ support that, and then rework all the jobs to still work without it.  That would get us full circle to the work done here.  

@hackmastera the `IngestFileJob` class is removed here, but the functionality is moved into an actor.  Sub-jobs are still async, but the top level is now only synchronous (previously `IngestFileJob.perform_later`).  That is because only the synchronous process is guaranteed to have access to the local file the use just uploaded.  

I am similarly confused about what work is imagined to follow on.  

The subsequent work *might* be to write a new job that wraps the functionality to allow async at the top level again.  It **cannot** be `IngestFileJob` as it stands, because it still relies on a local path.  IF we use `Hyrax::UploadedFile` for the new job, that gives us the carrierwave abstraction over storage (see `BatchCreateJob`).  This is what I am prepared to undertake.

Alternatively, the subsequent work might be to propose Hyrax **requires** shared filesystem amongst all workers.  Despite everyone w/ legacy systems relying on it and seemingly enjoying it, and specifically those who object to the timing of this merge, I have yet to hear anyone actually propose it.

atz:
@mjgiarlo Nobody is telling me what threshold they actually consider sufficient though.  `IngestFileJob` is not a valid `ActiveJob`.  It cannot be made valid without a breaking change. 
 Claims of reliance on `IngestFileJob` are therefore unpersuasive to me.  After acknowledging that it is broken, If you want to keep it as is, you *should* be forced onto a fork.  

@tpendragon The logic in this PR allows user uploads/updates for all worker configurations.  The current code does not.  As a platform, Hyrax has already claimed to support separate workers (w/o shared filesystems) and currently fails.  Therefore, if you consider the current code "releasable" and this PR "unreleasable", you are speaking only for your own institutional app/context and not for Hyrax.  My philosophy is to fix the platform/architecture first and worry about the apps second.

escowles:
@atz I have [suggested](https://github.com/samvera/hyrax/pull/1206#issuecomment-308503511) an alternative that I think would be less disruptive, maintain the background file ingest currently used by most apps, but still allow people with multiple-workers-without-shared-filesystems.  Is there a reason why having an alternate FileActor with the inline ingest functionality wouldn't work?

atz:
@escowles As I said, `IngestFileJob` would still be invalid.

tpendragon:
@atz I'm wondering if I just have a fundamental misunderstanding. Could you expand on why `IngestFileJob is not a valid ActiveJob.` ?

escowles:
@atz It would be invalid if the shared filesystem assumption wasn't true.  I appreciate that we've identified a new scenario where our old assumption isn't true, but I am 👎  on breaking existing functionality if a less disruptive solution is available.

atz:
@escowles It isn't existing functionality to a configuration we (all, apparently) intend to support, namely ours.  I appreciate the legacy context w/ Sufia/CC, but **Hyrax has no shared filesystem assumption**.  That is to say, the job is invalid **because** that assumption is false.  Any disruption is because we fooled ourselves into thinking we had something we didn't.  

@tpendragon It relies on a filepath argument, unconditionally calling `File.open` on it (and throwing exception) here: https://github.com/samvera/hyrax/blob/958c205d359a63166842d773e05ee157595a4128/app/jobs/ingest_file_job.rb#L14

That means the job expects to share filesystem with the caller, which it cannot do.  The job effectively decides that Hyrax does not support independent worker systems, despite our stated intentions.  

Subsequent Hyrax jobs also take a filepath argument, but it is treated as a "hint", and there are means to retrieve the file from storage if the file isn't found.  But as the first step of ingest, the file hasn't hit storage yet, so this functionality cannot use the same workaround.  There's nowhere else for a worker to get the file, it's only on the server that received the upload.  (Using carrierwave with shared or remote-backed storage solves this problem.)  

TL;DR: Passing a local tempfile path to a separate box like "do something with this!" is meaningless, but `IngestFileJob` *requires* it.

atz:
Having individual jobs dictate architectural choices, especially implicit undocumented ones, is a severely detrimental path to go down.

escowles:
> That is to say, the job is invalid because that assumption is false. Any disruption is because we fooled ourselves into thinking we had something we didn't.

As I said initially, there are two scenarios where that assumption is true:
1. Rails/workers running on a single machine
2. Rails/workers running on multiple machines with a shared filesystems

You keep repeating that the job is invalid because the assumption is false, but that simply isn't true in those two scenarios.  The reason why the job has worked that way for years, and multiple people have built their production repositories around the functionality you're throwing away here, is because the assumption is true for us.

You are trying to support a new scenario where the shared filesystem assumption isn't true, and I completely support that.  But acting like this is the only scenario that matters is not going to be persuasive with the dozens of sites that have been running apps in the first two scenarios for years.

If you don't want to have a different FileActor like I've outlined, the other way I can think of addressing this is to have shared temp space.  I'll admit I don't know much about AWS, but isn't there some way to have a shared filesystem?  Can the Rails uploads go into a shared S3 bucket or something?

hackmastera:
If possible, I'd prefer a scenario where we aren't maintaining more FileActors where two of them do essentially the same thing but for different architectures. I'm hoping we can think a little bit more about whether there's a way to make the new approach to work for everyone.

It sounds like @atz is ready to do additional work to create a new job that will take the place of the file ingest job. It's clear that this would be a breaking change. But I don't think that disqualifies it, as long as the functionality still exists and an upgrade path is documented to a reasonable extent. @escowles I'm not sure I totally understand your position. Can you say more about why this approach doesn't work for you? From your or @tpendragon's perspective is there additional work that could build on this PR toward a solution that works for you?

@atz can you or @mjgiarlo say more about hyrax explicitly targeting an architecture without a shared filesystem? I'm also on AWS and in favor of supporting that architecture, but I am not entirely sure this was clear about the project and it seems like that has not necessarily been a shared assumption.

escowles:
@hackmastera I agree that having a bunch of duplicated code isn't a good solution here.  But I would like to preserve the ability to do background file ingest, while it sounds like that definitely won't work with multiple machines without shared temp space.

I wonder if we can just check for a config variable and then call `perform_later` if there is shared temp space and `perform_now` if there isn't?

atz:
Yes, we already put the uploads into S3.  The mechanism for that is carrierwave.

@escowles:
(1) Running on the same server isn't enough.  The processes have to have access to the same temp directory AND be the same user or comparably permissioned users.  
(2) Tautology: yes, the assumption is true when it is true.  Speaking of permissions, are you just `0777` masking all your tempfiles?  In a shared directory?  

Your position is that you have accommodated the concessions required by the job, so that makes it a valid job.  This misunderstands my point.  What you have is code that does something, and I'm not arguing that it has no utility.  But that code doesn't accept the constraints expected by `ActiveJob`.  And your accommodations are not part of Hyrax, or else we would be using them already.  I would still make the same arguments, but we wouldn't have our pilot users throwing exceptions and there would be less urgency.

By more roundabout path, allow me to describe the kind of things a valid `ActiveJob` cannot do:
- only work synchronously (fail async)
- only work asynchronously (fail sync)
- require a given backend (e.g. must use sidekiq)
- require custom queue processing behavior (like "I want `ingest` queue, but I always need to be processed first")
- require that the job is serviced by a particular worker, or require that its worker have:
  - more memory than other workers on the same queue
  - special dependencies, binaries or hardware
  - a specific OS
  - a specific OS configuration
  - a specific UID/GID
  - a shared filesystem

If you agree with all the other points, you should be able to identify why the last one is unacceptable also.  The proposal to toggle `perform_now` and `perform_later` reveals the problem.  We all want `perform_later`: if it were a valid `ActiveJob`, we would have it.  (Separately, we also need to have a whole conversation about idempotence, another requirement we often shirk.) 

## Does Hyrax require the shared filesystem workaround or not? 

Strictly speaking no, but it is the only way to get beyond trivial usage of worker architecture (otherwise, only single server or `:inline` works).  So, *de facto*, yes.  That's a problem.

No other job has this requirement, only `FileIngestJob`.  That is exactly what I referred to as an individual job dictating an architectural choice. Despite shared filesystem having the benefit of being implemented by prominent users and *de facto* required, that choice is currently not documented, not proposed and not accepted **by Hyrax**.  Likewise, it isn't part of Hyrax testing or development environments.  As far as I can tell, it isn't even documented [in Sufia](https://github.com/samvera/sufia/wiki/Background-Workers).  Therefore, I don't see how the code I write or fix or accept into this project could rely on it.

atz:
Side note: it is odd that we already have [`IngestLocalFileJob`](https://github.com/samvera/hyrax/blob/master/app/jobs/ingest_local_file_job.rb).  It obviously has the same failings as `IngestFileJob`, but at least the name hints at its limitations.  By my reading, that means you still have a vehicle to rely on local files for your apps.

tpendragon:
@atz We talked locally, and `IngestLocalFileJob` provides the migration path we would need if we update in the mean-time. Unblocked.

atz:
thx, @tpendragon, trying to wrap up to replacement job soon

PR #1207, opened 2017-06-13
Remove depositors access to FileSets in mediated deposit

### LINKED ISSUE #1091: Depositor can edit the fileset after deposit has been approved (mediated workflow)
### Descriptive summary

After a depositor has made a deposit and this has been approved, they can no longer edit the work. But they are still able to edit the fileset.

[ observed in Hyku v1.0.0.beta1 ]

### Rationale

The behaviour needs to be consistent across works and their contained filesets.

### Expected behavior

User can no longer edit Work and contained FileSets.

### Actual behavior

User can edit FileSet.

1. Submit as a 'depositing' user to the mediated worfklow
2. Approver approves the deposit
3. Depositing user returns to the work and edits fileset

jcoyne:
Fixes #1091

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/hyrax/revoke_edit_from_members_job.rb
+++ b/app/jobs/hyrax/revoke_edit_from_members_job.rb
@@ -1,0 +1,23 @@
+module Hyrax
+  # Revokes edit access for the supplied user for the members attached to a work
+  class RevokeEditFromMembersJob < ApplicationJob
+    queue_as Hyrax.config.ingest_queue_name
+
+    # @param [ActiveFedora::Base] work - the work object
+    # @param [String] user_key - the user to remove
+    def perform(work, user_key)
+      # Iterate over ids because reifying objects is slow.
+      file_set_ids(work).each do |file_set_id|
+        RevokeEditJob.perform_now(file_set_id, user_key)
+      end
+    end
+
+    private
+
+      # Filter the member ids and return only the FileSet ids (filter out child works)
+      # @return [Array<String>] the file set ids
+      def file_set_ids(work)
+        ::FileSet.search_with_conditions(id: work.member_ids).map(&:id)
+      end
+  end
+end
--- a/app/jobs/hyrax/revoke_edit_job.rb
+++ b/app/jobs/hyrax/revoke_edit_job.rb
@@ -1,0 +1,14 @@
+module Hyrax
+  # Revokes the user's edit access on the provided FileSet
+  class RevokeEditJob < ApplicationJob
+    queue_as Hyrax.config.ingest_queue_name
+
+    # @param [String] file_set_id - the identifier of the object to revoke access from
+    # @param [String] user_key - the user to remove
+    def perform(file_set_id, user_key)
+      file_set = ::FileSet.find(file_set_id)
+      file_set.edit_users -= [user_key]
+      file_set.save!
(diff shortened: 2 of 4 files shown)


app/jobs/hyrax/revoke_edit_from_members_job.rb
```diff
@@ -0,0 +1,15 @@
...
+  # Revokes edit access for the supplied user for the members attached to a work
+  class RevokeEditFromMembersJob < ApplicationJob
+    queue_as Hyrax.config.ingest_queue_name
+
+    # @param [ActiveFedora::Base] work - the work object
+    # @param [String] user_key - the user to remove
+    def perform(work, user_key)
+      # Iterate over ids because reifying objects is slow.
+      work.member_ids.each do |file_set_id|
+        RevokeEditJob.perform_now(file_set_id, user_key)
```
mjgiarlo: Will this also act on child works? If so, `file_set_id` may cause confusion in the future.

app/jobs/hyrax/revoke_edit_job.rb
```diff
@@ -0,0 +1,14 @@
+module Hyrax
+  # Revokes the user's edit access on the provided FileSet
+  class RevokeEditJob < ApplicationJob
+    queue_as Hyrax.config.ingest_queue_name
+
+    # @param [String] file_set_id - the identifier of the object to revoke access from
+    # @param [String] user_key - the user to remove
+    def perform(file_set_id, user_key)
```
mjgiarlo: Same comment as above.

app/jobs/hyrax/revoke_edit_from_members_job.rb
darrenleeweber: If the method in `RevokeEditJob` is always called with `perform_now` and only ever called here, then move that perform method into a private method on this job and dispense with the extra job.

jcoyne: @darrenleeweber If we do that, that means a single failure in updating one of the members would halt the whole batch. This way, the failure is isolated and can be restarted independently.

jcoyne: @mjgiarlo I'm not sure what we should do about child works. Should I explicitly include them or exclude them?

darrenleeweber: @jcoyne - so your saying the whole thing does not need to be an atomic "transaction"?  It's OK if any one of the users is not removed?  As it is, I _guess_ this parent job will never fail.

jcoyne: @darrenleeweber There is no way in ActiveFedora to use a transaction (mostly because there is no support for transactions in Solr).

mjgiarlo: @jcoyne I suspect that excluding them until we have to analyze the effects and side-effects of that change can be explored. Write up a new issue and focus on FSes here?

jcoyne: Okay, I'll make some changes. Here's the ticket for child works: https://github.com/samvera/hyrax/issues/1213

PR #1215, opened 2017-06-14
Change service `working_directory` to not slurp file into memory

### LINKED ISSUE #1128: Hyrax::WorkingDirectory puts object in memory as string
`Hyrax::WorkingDirectory` in the line:
```ruby
copy_stream_to_working_directory(id, file.original_name, StringIO.new(file.content))
```
calling `file.content` (potentially) puts the entire file in memory at once.  See [ActiveFedora::File#content](https://github.com/samvera/active_fedora/blob/96e8a63de5a8c1881fd4b40dc3448c7f827c668d/lib/active_fedora/file.rb#L154-L156) and the various behaviors based on different `@content` classes (`local_or_remote_content` and `behaves_like_io?`).  

The point is that `WorkingDirectory` doesn't know whether it is getting in IO-like object, or an internal `@ds_content`, `ldp_source.get.body` or what.  @jcoyne identified this problem in commit 5db7d0c5699e707e9dee2ddfcd6574c8f058cbe1 with the line:
```
TODO: this causes a load into memory, which we'd like to avoid
```

billdueber:
Fixes #1128

Method `#copy_stream_to_working_directory` previously used ruby's
`copy_stream`, which reads the whole source stream into memory.

New `#copy_stream_to_file` uses `IO#readpartial` to ferry it around in
4k chunks.

`IO#readpartial` is preferred to `#read` because it doesn't assume the stream
is complete when first referenced. 4k is a common heap page size and
should be fairly efficient.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/working_directory.rb
+++ b/app/services/hyrax/working_directory.rb
@@ -37,6 +37,23 @@ module Hyrax
 
       private
 
+        # @param [#read] stream The stream to copy from
+        # @param  [String] filename The file name to copy to
+        # @param  [Number] chunk_size How many bytes to copy at once
+        def copy_stream_to_file(stream, filename, chunk_size = 4096)
+          File.open(filename, 'wb') do |outfile|
+            loop do
+              begin
+                outfile.write(
+                  stream.readpartial(chunk_size) # throws EOFError when done
+                )
+              rescue EOFError
+                # all done. I hate using error handling for flow control
+              end
+            end
+          end
+        end
+
         # @param [String] id the identifier
         # @param [String] name the file name
         # @param [#read] stream the stream to copy to the working directory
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@atz mind reviewing?

jcoyne:
My understanding was that the problem was on this line https://github.com/samvera/hyrax/blob/4e8b1b065be175262ace3ff18e6210b36b10bf70/app/services/hyrax/working_directory.rb#L35 

specifically: `StringIO.new(file.content)`  I wonder if we could stop using `Hydra::PCDM::File#content` and instead create a `Hydra::PCDM::File#content_as_stream`

jcoyne:
@billdueber Could we do something like:
```
fs.original_file.stream.each { |chunk| f.write(chunk) }
```

atz:
@jcoyne: cannot assume `original_file` is what matters.  

My first reaction is that I don't care, I plan on removing `Hyrax::WorkingDirectory` anyway as it becomes unnecessary with carrierwave.  But secondly, why is the exception catching inside the loop?  Don't you keep looping after the `rescue`?

billdueber:
The exception should not be inside the loop. The code doesn't do what I set out to address. I feel like I'm in one of those "wanna get away" SouthWest commercials.

PR #1216, opened 2017-06-14
increase the number of AdminSets rendered on #index

revgum:
by default, when visiting the AdminSet #index view, the number of rows rendered is limited to 10. The select list for adjusting viewable rows in the table is not intending on making a request of the server for more/less rows, it just adjusts the visible rows and pagination on the page. 

This patch increased the default number of rows to be rendered so that the table is capable of rendering and paginating up to 100 AdminSets.

I don't think this is the best long term solution for this issue, however it does somewhat fall in line with how this is done in the CollectionSearchBuilder. https://github.com/samvera/hyrax/blob/master/app/search_builders/hyrax/collection_search_builder.rb

### DIFF
--- a/app/services/hyrax/admin_set_service.rb
+++ b/app/services/hyrax/admin_set_service.rb
@@ -46,7 +46,7 @@ module Hyrax
 
       # @param [Symbol] access :read or :edit
       def builder(access)
-        search_builder.new(context, access)
+        search_builder.new(context, access).rows(100)
       end
 
       # Count number of files from admin set works
--- a/spec/services/hyrax/admin_set_service_spec.rb
+++ b/spec/services/hyrax/admin_set_service_spec.rb
@@ -40,6 +40,7 @@ RSpec.describe Hyrax::AdminSetService do
     let(:search_builder_instance) { double }
 
     it "calls the injected search builder" do
+      expect(search_builder_instance).to receive(:rows).and_return(search_builder_instance)
       expect(search_builder_instance).to receive(:reverse_merge).and_return({})
       subject
     end


app/search_builders/hyrax/admin_set_search_builder.rb
```diff
@@ -8,6 +8,7 @@ class AdminSetSearchBuilder < ::SearchBuilder
     # @param [Symbol] access one of :edit, :read, or :deposit
     def initialize(context, access)
       @access = access
+      @rows = 100
```
jcoyne: Is it possible to scope this change to `AdminSetService#search_results_with_work_count` ?

revgum: In the case I was addressing (/admin/admin_sets?locale=en) `#search_results_with_work_count?` isn't involved. The request comes in by way of: https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/admin/admin_sets_controller.rb#L36

jcoyne: Okay, `AdminSetService#search_results` then?

revgum: I had thought about that as well but went with this approach to be consistent with the CollectionSearchBuilder.. So would are you thinking, 
https://github.com/samvera/hyrax/blob/da079af2148204f580830377bb91038fc0198108/app/services/hyrax/admin_set_service.rb#L18 :

`response = context.repository.search(builder(access, 100))` 

and then setting @rows as an argument in `AdminSetSearchBuilder.initialize`?

jcoyne: I do not think we should set `@rows` at all if we can avoid it. We can do `AdminSetSearchBuilder.new.rows(100)` instead.  That's how you're supposed to use that API.

jcoyne: @revgum see https://github.com/samvera/hyrax/pull/1221

PR #1219, opened 2017-06-14
Fix collection count

### LINKED ISSUE #1081: Count of items in collection is broken
The collection show page displays the number of "total items" in the collection. The number is consistently shown as "0" despite when there are one or more works who are members of the collection.

This should be fixed so that the number shown reflects the accurate number of works in the collection.

(Replace "item" with "work" is ticketed separately.)

jcoyne:
It needed to be changed to use the new inverted collection model.
Fixes #1081 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/collections.js
+++ b/app/assets/javascripts/hyrax/collections.js
@@ -1,13 +1,12 @@
 Blacklight.onLoad(function () {
 
   // change the action based which collection is selected
-  $('input.updates-collection').on('click', function() {
- 
-      var form = $(this).closest("form"); 
-      var collection_id = $(".collection-selector:checked")[0].value;      
-      form[0].action = form[0].action.replace("collection_replace_id",collection_id);
+  // This expects the form to have a path that includes the string 'collection_replace_id'
+  $('[data-behavior="updates-collection"]').on('click', function() {
+      var string_to_replace = "collection_replace_id"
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/hyrax/collections.js
```diff
@@ -1,13 +1,12 @@
...
 
   // change the action based which collection is selected
-  $('input.updates-collection').on('click', function() {
- 
-      var form = $(this).closest("form"); 
-      var collection_id = $(".collection-selector:checked")[0].value;      
-      form[0].action = form[0].action.replace("collection_replace_id",collection_id);
+  // This expects the form to have a path that includes the string 'collection_replace_id'
+  $('[data-behavior="updates-collection"]').on('click', function() {
+      var string_to_replace = "collection_replace_id"
```
atz: Care to toss the dog a semicolon?

jcoyne: If you feel strongly I can. I think this is silly because we have a eslint rule that says, we don't care about semicolins: https://github.com/samvera/hyrax/blob/master/.eslintrc#L183 It's just that hound isn't reading it right now.

jcoyne: https://github.com/samvera/hyrax/pull/1228

PR #1227, opened 2017-06-15
Update dependencies which saves and retreives charset

### LINKED ISSUE #1089: Character encoding problems in full-text indexing

jcoyne:
Specifically for the extracted text.

Fixes #1089 
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/persist_directly_contained_output_file_service.rb
+++ b/app/services/hyrax/persist_directly_contained_output_file_service.rb
@@ -5,22 +5,37 @@ module Hyrax
     # This method conforms to the signature of the .call method on Hydra::Derivatives::PersistOutputFileService
     # * Persists the file within the DirectContainer specified by :container
     #
-    # @param [String] the data to be persisted
-    # @param [Hash] directives directions which can be used to determine where to persist to.
+    # @param content [String] the data to be persisted
+    # @param directives [Hash] directions which can be used to determine where to persist to.
     # @option directives [String] url URI for the parent object.
     # @option directives [String] container Name of the container association.
-    def self.call(string, directives)
-      file = Hydra::Derivatives::IoDecorator.new(string, new_mime_type(directives.fetch(:format)))
-      o_name = determine_original_name(file)
-      m_type = determine_mime_type(file)
-      uri = URI(directives.fetch(:url))
-      raise ArgumentError, "#{uri} is not an http(s) uri" unless uri.is_a?(URI::HTTP)
-      file_set = ActiveFedora::Base.find(ActiveFedora::Base.uri_to_id(uri.to_s))
-      remote_file = file_set.send("build_#{directives.fetch(:container)}".to_sym)
+    def self.call(content, directives)
+      file = io(content, directives)
+      file_set = retrieve_file_set(directives)
+      remote_file = retrieve_remote_file(file_set, directives)
       remote_file.content = file
-      remote_file.mime_type = m_type
-      remote_file.original_name = o_name
+      remote_file.mime_type = determine_mime_type(file)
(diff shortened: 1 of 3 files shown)


app/services/hyrax/persist_directly_contained_output_file_service.rb
```diff
@@ -5,22 +5,37 @@ class PersistDirectlyContainedOutputFileService < Hydra::Derivatives::PersistBas
...
-    def self.call(string, directives)
-      file = Hydra::Derivatives::IoDecorator.new(string, new_mime_type(directives.fetch(:format)))
-      o_name = determine_original_name(file)
-      m_type = determine_mime_type(file)
-      uri = URI(directives.fetch(:url))
-      raise ArgumentError, "#{uri} is not an http(s) uri" unless uri.is_a?(URI::HTTP)
-      file_set = ActiveFedora::Base.find(ActiveFedora::Base.uri_to_id(uri.to_s))
-      remote_file = file_set.send("build_#{directives.fetch(:container)}".to_sym)
+    def self.call(content, directives)
+      file = io(content, directives)
```
jeremyf: I'm missing where the `io` method comes from. Otherwise, things look good.

jcoyne: New version of hydra-derivatives 3.3.0

Issue #790, opened 2017-04-14
When I select a thumbnail_id for an admin set it doesn't show up on the front page.


jcoyne:
It just shows the default admin set.

atz:
@andjsmit, did you want to have this one assigned?

ggeisler:
Note that the selected thumbnail also does not display on the Administrative Set show page in the admin UI; the file icon is shown when it should be the selected thumbnail image.

![show_admin_set____hyrax](https://cloud.githubusercontent.com/assets/101482/25396554/da328bc0-2999-11e7-8257-c8f33364dd69.png)

randalldfloyd:
This might be related to ability on the fileset when the thumbnail is rendered.  If you create a work that is open access (and the fileset is open access) then set one of its images to be the thumbnail, then the admin set views display it correctly. If the work's fileset is private, it will not display, even if the user's role is admin.

randalldfloyd:
@mjgiarlo @jcoyne  I need a some guidance or opinions on this one...

Here's the problem on this issue.  If user1 can enumerate files in an admin set and select one of them to be the representative thumbnail, then there's a pretty good chance that other users will not have ability to view the thumbnail (i.e. the file_set/files are private).  Those users therefore get the default image asset per the ability check and the thumbnail helper. As it turns out, being able to select the file while editing the admin set as user1 is actually no guarantee that user1 can even view the thumbnail, hence the behavior as described in this issue. 

So, I was trying to approach this as an ability issue on the thumbnail rendering end, but really this is a problem of even being able to select a representative image that may not be public.  It seems like a design oversight. Either the admin set editor should not be able to assign non-public files, or there should be a bit of magic to render a nicer thumbnail representing an admin set instead of getting the ugly-too-big standin that Blacklight returns. Thoughts?

jcoyne:
@randalldfloyd that sounds like a different issue, and I suggest opening a ticket for that.  For this ticket lets presume that they've selected a publicly visible file as the thumbnail.

randalldfloyd:
@jcoyne I might be missing something or misunderstanding you, but from my testing, if the selected file is publicly visible, then this problem doesn't exist. The admin set editor will see the selected thumbnail in all relevant views, and any other user (or even nobody) can see the thumbnail:

Nobody logged in and selected file is Open Access
![image](https://cloud.githubusercontent.com/assets/6411186/26502470/e9488bbc-420a-11e7-9fe2-c0179c8060c6.png)

[...]

PR #1230, opened 2017-06-16
Skipping authorize for file download for admins that can already read parent file_set

### LINKED ISSUE #1163: Admin user has no read ability on files from within download controller.
### Expected behavior

Admin user should be able to view files attached to other users' private works, especially if they are able to view or enumerate the private file_sets they are attached to.  

### Actual behavior

Admin user can view other users' private works and related file_sets but not the actual files via the download controller.  For example, an admin user can enumerate and select any member file in an admin set to be the representative thumbnail even though the admin user itself may not have ability to view it (see #790).

In the example below, a regular depositor has attached a file to a private work and the admin user can search, view, and edit the private work and its file_set.  However, when rendering the thumbnail or attempting to download, it fails the ability check to get the file via the download controller and gets the default asset instead:
```
Started GET "/downloads/gq67jr16q?file=thumbnail" for 127.0.0.1 at 2017-06-07 10:57:48 -0400
Processing by Hyrax::DownloadsController#show as HTML
  Parameters: {"file"=>"thumbnail", "id"=>"gq67jr16q"}
  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? ORDER BY "users"."id" ASC LIMIT ?  [["id", 3], ["LIMIT", 1]]
Usergroups are ["public", "admin", "registered"]
  Hyrax::Feature Load (0.1ms)  SELECT  "hyrax_features".* FROM "hyrax_features" WHERE "hyrax_features"."key" = ? ORDER BY "hyrax_features"."id" ASC LIMIT ?  [["key", "transfer_works"], ["LIMIT", 1]]
  Hyrax::Feature Load (0.1ms)  SELECT  "hyrax_features".* FROM "hyrax_features" WHERE "hyrax_features"."key" = ? ORDER BY "hyrax_features"."id" ASC LIMIT ?  [["key", "proxy_deposit"], ["LIMIT", 1]]
[CANCAN] Checking read permissions for user: admin@example.com with groups: ["public", "admin", "registered"]
[CANCAN] download_groups: []
[CANCAN] read_groups: []
[CANCAN] edit_groups: []
[CANCAN] read_groups: []
[CANCAN] download_users: []
[CANCAN] read_users: []
[CANCAN] edit_users: ["depositor@example.com"]
[CANCAN] read_users: ["depositor@example.com"]
Redirected to http://localhost:3000/assets/default-f936e9c3ea7a38e2c2092099586a71380b11258697b37fb4df376704495a849a.png
Filter chain halted as :authorize_download! rendered or redirected
Completed 302 Found in 10ms (ActiveRecord: 0.5ms)
```
### Related work

Disparity in ability first described in #167
Affects #790

randalldfloyd:
Fixes #1163 

Bypasses the `authorize!` check for admin user in the downloads controller so that the file(s) in the direct container can be downloaded/viewed (i.e. thumbnails, etc.)

There's a problem I can't figure out where the admin user has already passed an ability check to read a private file_set, but then when the downloads controller authorizes that user for that same file_set, in order to allow its file content to be viewed, it fails.  For now recommending bypassing the check when `user_ability.admin?` to get this issue unstuck so that #790 is not affected by this problem.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/downloads_controller.rb
+++ b/app/controllers/hyrax/downloads_controller.rb
@@ -41,6 +41,7 @@ module Hyrax
       # Hydra::Ability#download_permissions can't be used in this case because it assumes
       # that files are in a LDP basic container, and thus, included in the asset's uri.
       def authorize_download!
+        return if current_ability.admin?
         authorize! :read, params[asset_param_key]
       rescue CanCan::AccessDenied
         redirect_to default_image


app/controllers/hyrax/downloads_controller.rb
```diff
@@ -41,6 +41,7 @@ def derivative_download_options
       # Hydra::Ability#download_permissions can't be used in this case because it assumes
       # that files are in a LDP basic container, and thus, included in the asset's uri.
       def authorize_download!
+        return if current_ability.admin?
```
jeremyf: Can this be moved into the Ability class? It seems like the short-circuit should be there.

jcoyne:
I'd like to have a look at this next week and see if there's anything I can do. I believe @randalldfloyd is on vacation.

app/controllers/hyrax/downloads_controller.rb
randalldfloyd: Yeah I agree but tried all kinds of ways to do that and my CanCan(Can) knowledge isn't enough to see what to do.  The thing is, the admin user already has [declared ability on ::FileSet in Ability](https://github.com/samvera/hyrax/blob/master/app/models/concerns/hyrax/ability.rb#L207), and presumably that's what gets the admin user access to the parent file_set in the file_set controller context.  It should just work without the shortcut but I can try a different shortcut in Ability if someone can tell me what to do.

mjgiarlo:
Thanks for taking a look, @jcoyne, and for getting this started, @randalldfloyd!

mjgiarlo:
This has been taken on in https://github.com/samvera/hyrax/pull/1232. Thanks for getting this started!

PR #1261, opened 2017-06-22
Rename no-empty-label to no-label

jcoyne:
This rule was removed and replaced in eslint 2.0
See http://eslint.org/docs/rules/no-empty-label


atz:
Any reason we see that failing test w/ `homepage_controller_spec`?

jcoyne:
@atz I think that was a spurious error since one build passed and one failed.  I've restarted the failing one.

PR #1262, opened 2017-06-22
Remove update profile

### LINKED ISSUE #1096: User profile: ‘refresh directory info’
### Descriptive summary

In the User Profile  ‘refresh directory info’ isn’t very intuitive - could it be renamed?

Also it should be hidden when not relevant, eg. when local user accounts are being used.

jcoyne:
This didn't actually have a working implementation in Hyrax.

Fixes #1096

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/profiles_controller.rb
+++ b/app/controllers/hyrax/dashboard/profiles_controller.rb
@@ -23,9 +23,7 @@ module Hyrax
 
       # Process changes from profile form
       def update
-        conditionally_set_user_attributes
-
-        if @user.save
+        if conditionally_update
           handle_successful_update
           redirect_to hyrax.dashboard_profile_path(@user.to_param), notice: "Your profile has been updated"
         else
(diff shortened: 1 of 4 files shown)


mjgiarlo:
The effect of this PR is to remove a checkbox in Hyrax that does nothing and probably confuses users. It also removes some functionality that ScholarSphere may be using.

@awead @cam156 @mtribone How much will this impact the ability in ScholarSphere to update attributes from LDAP? I'm not sure how much of this code ever belonged in Sufia, or was used -- I suspect most of the working code here is in ScholarSphere. If we remove this, can you still satisfy the "update attrs from LDAP" use case via e.g., overriding `ProfilesController#update`? I want to gauge the amount of work this is going to cause you folks to do before approving this PR.

awead:
@mjgiarlo @cam156 is out until next week, but I don't see a reason to keep this in. Directory information is going to be different per implementor and seems more locally-oriented. That said, it's still a feature we use and would need to re-implement if we were to move to Hyrax.

mjgiarlo:
@awead No rush. Happy to wait on @cam156's review!

mjgiarlo:
@cam156 Thoughts?

jcoyne:
@cam156 there is no hook. You'd have to prepend a module to the `ProfilesController` that overrode some methods.  I don't think this is a common enough use case (anyone besides PSU?) to justify writing in some sort of configuration for this.

mjgiarlo:
@jcoyne The pre-pended module would likely only have to override `#update`, right?

jcoyne:
@mjgiarlo Yes, just that method.

mjgiarlo:
@cam156 Hi! Do you care to follow up given the method of extension has now been described?

cam156:
@mjgiarlo We are using prepend to override other things in hyrax, so if that is the new described approach for making updates go for it.  I must say I am not entirely in love with the PFM of prepends, but they do work in a pinch and are better than copying over the entire class.

mjgiarlo:
@cam156 OK. Can you say what you don't like about the prepend method of extension? (I'm not sure what "PFM" is, sorry!)

jcoyne:
Well, surely PFM = pure fantastic magic, right? 😉

mjgiarlo:
Hmmm, how is `Module#prepend` more :sparkles: than `Module#include`?

cam156:
@mjgiarlo becuase it is configured away from the code.  It is monkey patching with rails support.

mjgiarlo:
@cam156 OK.

PR #1270, opened 2017-06-22
Move stray config from engine.rb to configuration class

mjgiarlo:
And add the option to the generated initializer so users know they can tweak it.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -379,12 +379,31 @@ module Hyrax
       @extract_full_text = true
     end
 
+    attr_writer :uploader
+    def uploader
+      @uploader ||= if Rails.env.development?
+                      # use sequential uploads in development to avoid database locking problems with sqlite3.
+                      default_uploader_config.merge(limitConcurrentUploads: 1, sequentialUploads: true)
+                    else
+                      default_uploader_config
+                    end
+    end
+
     private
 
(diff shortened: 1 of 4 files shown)


lib/hyrax/configuration.rb
```diff
@@ -379,12 +379,28 @@ def extract_full_text?
       @extract_full_text = true
     end
 
+    attr_writer :uploader
+    def uploader
+      return default_uploader_config unless Rails.env.development?
+      # use sequential uploads in development to avoid database locking problems with sqlite3.
+      default_uploader_config.merge!(limitConcurrentUploads: 1, sequentialUploads: true)
```
jcoyne: why `merge!`, This is returning a Hash, so no need to modify state.

lib/hyrax/configuration.rb
```diff
@@ -379,12 +379,28 @@ def extract_full_text?
       @extract_full_text = true
     end
 
+    attr_writer :uploader
+    def uploader
+      return default_uploader_config unless Rails.env.development?
```
jcoyne: Shoudn't this return `@uploader` otherwise what's the point of the writer?

lib/hyrax/configuration.rb
mjgiarlo: The implemention I am moving uses `merge!`. I don't need to use it, but I wasn't planning to change it. I can do that in my next push.

PR #1271, opened 2017-06-22
Use layer 0 when making image derivatives

jcoyne:
Fixes problem where pyramidal tiffs get flattend into the thumbnail.
Thanks to @hackmastera for the fix:
https://github.com/chemheritage/chf-sufia/blob/b8d3d8747cb2a77e141febdac2f7d80ca5704a45/app/models/file_set.rb#L14

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/file_set_derivatives_service.rb
+++ b/app/services/hyrax/file_set_derivatives_service.rb
@@ -74,8 +74,13 @@ module Hyrax
       end
 
       def create_image_derivatives(filename)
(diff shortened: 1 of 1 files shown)


atz:
Is there some reason that layer 0 is conventionally or per-spec the correct layer and not one of the (potential) others, or is this just a best guess?

hackmastera:
@atz i'm not sure. It makes sense for it to be layer 0 since the smaller tiffs are created from the largest one it and then added in.

Issue #1226, opened 2017-06-15
Intermittent test failure 


jcoyne:
I suspect we need to change `empty?` to `blank?`
```
  1) leases managing leases shows lists of objects under lease

     Failure/Error: <td class="title"><%= link_to curation_concern, edit_lease_path(curation_concern) %></td>

     

     ActionView::Template::Error:

       undefined method `empty?' for nil:NilClass

     # ./app/views/hyrax/leases/_list_active_leases.html.erb:9:in `block in __home_travis_build_samvera_hyrax_app_views_hyrax_leases__list_active_leases_html_erb__1918967764296208783_226972580'

     # ./app/views/hyrax/leases/_list_active_leases.html.erb:6:in `each'

     # ./app/views/hyrax/leases/_list_active_leases.html.erb:6:in `__home_travis_build_samvera_hyrax_app_views_hyrax_leases__list_active_leases_html_erb__1918967764296208783_226972580'

     # ./app/views/hyrax/leases/index.html.erb:14:in `__home_travis_build_samvera_hyrax_app_views_hyrax_leases_index_html_erb__2642505232786911392_227003440'

     # ./spec/features/lease_spec.rb:39:in `block (3 levels) in <top (required)>'

     # ------------------

     # --- Caused by: ---

     # NoMethodError:

     #   undefined method `empty?' for nil:NilClass

     #   ./app/views/hyrax/leases/_list_active_leases.html.erb:9:in `block in __home_travis_build_samvera_hyrax_app_views_hyrax_leases__list_active_leases_html_erb__1918967764296208783_226972580'

Finished in 16 minutes 13 seconds (files took 13.29 seconds to load)

2314 examples, 1 failure, 11 pending

Failed examples:

rspec ./spec/features/lease_spec.rb:38 # leases managing leases shows lists of objects under lease

Randomized with seed 15788
```

PR #1278, opened 2017-06-22
[rubocop] update to 0.49

dunn:
* 0.49 introduced cops for ensuring that jobs and models don't subclass directly from ActiveJob and ActiveModel; but they are `--auto-correct` cops it's only safe to change `ActiveJob::Base` to `ApplicationJob` etc. in an application that was generated as a Rails 5 app.

* Some cops were moved from Style to Layout.

* `.rubocop_todo.yml` is meant to be auto-generated by rubocop, so `require: rubocop-rails` should be in `.rubocop.yml` itself.

* `--auto-gen-config` sets a maximum `Metrics/MethodLength` in `.rubocop_todo.yml` instead of listing exceptions; that max value is overridden by the maximum in `.rubocop.yml` so I removed the latter for now.

@samvera/hyrax-code-reviewers


.rubocop_todo.yml
```diff
@@ -1,154 +1,106 @@
...
   Exclude:
-    - 'app/jobs/batch_create_job.rb'
+    - 'Rakefile'
 
+# Offense count: 104
 Metrics/AbcSize:
-  Exclude:
-    - 'app/indexers/hyrax/file_set_indexer.rb'
-    - 'app/services/hyrax/workflow/permission_query.rb'
+  Max: 58
```
jcoyne: This is increasing from the limit we previously set of 28.  There were only 2 exceptions to that. I prefer we put that back so that we maintain a relatively low ceiling with a few outliers.

.rubocop_todo.yml
```diff
@@ -1,154 +1,106 @@
...
-    - 'app/controllers/concerns/hyrax/collections_controller_behavior.rb'
-    - 'app/controllers/concerns/hyrax/works_controller_behavior.rb'
-    - 'app/helpers/hyrax/hyrax_helper_behavior.rb'
-    - 'app/models/concerns/hyrax/ability.rb'
-    - 'app/services/hyrax/workflow/permission_query.rb'
-    - 'spec/services/hyrax/workflow/permission_query_spec.rb'
+# Offense count: 8
+# Configuration parameters: CountComments.
+Metrics/ClassLength:
+  Max: 278
```
jcoyne: This is much higher than what we previously were using (100 I believe). Please revert this and keep the exceptions.  We don't want new classes at > 200 lines.

.rubocop_todo.yml
```diff
@@ -1,154 +1,106 @@
...
-    - 'app/controllers/hyrax/my_controller.rb'
-    - 'app/controllers/concerns/hyrax/works_controller_behavior.rb'
-    - 'app/indexers/hyrax/file_set_indexer.rb'
-    - 'app/inputs/multi_value_select_input.rb'
-    - 'app/models/concerns/hyrax/solr_document/export.rb'
-    - 'app/services/hyrax/workflow/permission_query.rb'
-    - 'lib/generators/hyrax/install_generator.rb'
-    - 'lib/hyrax/rails/routes.rb'
-    - 'spec/support/**/*'
+  Max: 51
```
jcoyne: Same here.

.rubocop_todo.yml
```diff
@@ -1,154 +1,106 @@
...
 
-Style/LineEndConcatenation:
-  Exclude:
-    - 'spec/test_app_templates/lib/generators/**/*'
-    - 'lib/generators/**/*'
+# Offense count: 6
+# Configuration parameters: CountComments.
+Metrics/ModuleLength:
+  Max: 230
```
jcoyne: And here.

.rubocop_todo.yml
```diff
@@ -1,154 +1,106 @@
...
+# Configuration parameters: CountKeywordArgs.
+Metrics/ParameterLists:
+  Max: 6
 
-Rails/Output:
-  Exclude:
-    - 'lib/generators/**/*'
+# Offense count: 37
+RSpec/AnyInstance:
+  Enabled: false
```
jcoyne: I'd prefer we enable this and work towards removing those exceptions we've already granted.

.rubocop_todo.yml
```diff
@@ -1,154 +1,106 @@
...
 
+# Offense count: 227
+# Configuration parameters: Max.
+RSpec/ExampleLength:
+  Enabled: false
```
jcoyne: Why are you turning this off?

dunn:
> I do not think we should abandon our customized .rubocop_todo.yml for the generated one.

Then it should be named something else, so it's not overwritten by `--auto-gen-config`.

.rubocop.yml
```diff
@@ -99,25 +101,32 @@ Style/ZeroLengthPredicate:
 Rails:
   Enabled: true
 
+Rails/ApplicationJob:
+  Enabled: false
+
+Rails/ApplicationRecord:
+  Enabled: false
```
jcoyne: Can we turn this on and autofix?

.rubocop.yml
```diff
@@ -99,25 +101,32 @@ Style/ZeroLengthPredicate:
 Rails:
   Enabled: true
 
+Rails/ApplicationJob:
+  Enabled: false
```
jcoyne: Can we turn this on and autofix?

app/presenters/hyrax/fixity_status_presenter.rb
```diff
@@ -34,7 +34,6 @@ def render_file_set_status
 
       # A weird display, cause we need it to fit in that 180px column on
       # FileSet show, and have no real UI to link to for files/versions :(
-      # rubocop:disable Metrics/MethodLength
```
jcoyne: Is there a reason you're removing these and moving to the global?  These are preferred as it scopes the problem to a few lines rather than the whole file.

app/services/hyrax/default_middleware_stack.rb
```diff
@@ -1,6 +1,5 @@
 module Hyrax
   class DefaultMiddlewareStack
-    # rubocop:disable Metrics/MethodLength
```
jcoyne: Same question here.

lib/hyrax/arkivo/actor.rb
```diff
@@ -55,7 +55,6 @@ def destroy_work(work)
       private
 
         # @return [Hash<String, Array>] a list of properties to set on the work. Keys must be strings in order for them to correctly merge with the values from arkivio (in `@item`)
-        # rubocop:disable Metrics/MethodLength
```
jcoyne: and here.

app/presenters/hyrax/fixity_status_presenter.rb
dunn: Oh, that was a result of the mixup with overwriting the `.rubocop_todo.yml`.  I'll fix it.

.rubocop.yml
dunn: Possibly; I was going to leave it for a future PR though.

jcoyne: 👌

.rubocop.yml
```diff
@@ -99,25 +101,32 @@ Style/ZeroLengthPredicate:
...
-    - 'app/helpers/hyrax/hyrax_helper_behavior.rb'
-    - 'app/helpers/hyrax/collections_helper.rb'
-    - 'app/helpers/hyrax/content_block_helper_behavior.rb'
     - 'app/builders/hyrax/form_builder.rb'
     - 'app/helpers/hyrax/citations_behaviors/formatters/apa_formatter.rb'
     - 'app/helpers/hyrax/citations_behaviors/formatters/chicago_formatter.rb'
     - 'app/helpers/hyrax/citations_behaviors/formatters/mla_formatter.rb'
+    - 'app/helpers/hyrax/collections_helper.rb'
+    - 'app/helpers/hyrax/content_block_helper_behavior.rb'
+    - 'app/helpers/hyrax/hyrax_helper_behavior.rb'
```
jcoyne: Remove these two lines & 129 please.

dunn:
```
$ bundle exec rubocop --format simple --config .rubocop.yml --parallel
== app/helpers/hyrax/hyrax_helper_behavior.rb ==
C: 38:133: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
C: 43:135: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
C: 79: 83: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
C:195: 87: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
== app/presenters/hyrax/fixity_status_presenter.rb ==
C: 40: 53: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
C: 44: 23: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
C: 50: 24: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
== spec/views/hyrax/my/works/_list_works.html.erb_spec.rb ==
C: 17:123: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
C: 18:117: Rails/OutputSafety: Tagging a string as html safe may be a security risk.

1028 files inspected, 9 offenses detected
```

.rubocop_fixme.yml
```diff
@@ -47,6 +47,10 @@ Metrics/MethodLength:
     - 'lib/hyrax/rails/routes.rb'
     - 'spec/support/**/*'
 
+Style/FormatStringToken:
```
dunn: rubocop says
```
== config/routes.rb ==
C: 10: 58: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over template tokens (like %{foo}).
```
But that doesn't seem to work, at least in the routes definitions.

dunn: Possibly https://github.com/bbatsov/rubocop/issues/4539

dunn:
One of the two runs failed; maybe it's just flaky? https://travis-ci.org/samvera/hyrax/builds/246198232?utm_source=github_status&utm_medium=notification

mjgiarlo:
@dunn Yup, it's a flappy feature spec (https://github.com/samvera/hyrax/issues/1226). I've restarted the build.

Thanks for the contribution!

dunn:
🍏

mjgiarlo:
@jcoyne re-review?

.rubocop_fixme.yml
jcoyne: Can you add a comment with that issue here?

.rubocop.yml
dunn: It was an issue; I just fixed the sorting of those lines.  Sorry for the confusion.

.rubocop_fixme.yml
dunn: Done.

dunn:
OK, I think I addressed Justin's last comments.

.rubocop.yml
jcoyne: fixity_status_presenter.rb was not in the list before. 🤔

PR #1280, opened 2017-06-23
Move My* search builders into the My:: module

jcoyne:
Extract a My::SearchBuilder for searches that need to be constrained by
depositor

This pulls the good stuff out of  #769 and leaves the controversial changes.
@samvera/hyrax-code-reviewers


jcoyne:
@mjgiarlo added a note to the *Search builder module changes* section

PR #1286, opened 2017-06-23
converted batch select to es6

cam156:
Did this to test my knowledge after running through the es6 tutorial
Additionally added a jasmine test to make sure this feature does not break in the future

* convert batch_select.js to batch_select.es6
* Include a jasmine test for batch_select.js

@samvera/hyrax-code-reviewers


app/assets/javascripts/hyrax/app.js
```diff
@@ -134,5 +135,12 @@ Hyrax = {
     statistics: {},
 
     // initialized in hyrax/config.js
-    config: {}
+    config: {},
+
+    // Per Page select will submit its form to change records shown
+    batchSubmit: function () {
+        var BatchSubmit = require('hyrax/batch_select');
```
cam156: @jcoyne This matches the other requires in the system.  Any idea why hound is yelling for this one? Should it just be ignored?

jcoyne: Yeah, we're just ignoring these. I guess eventually we could convert app.js to app.es6 and do away with these altogether.

app/assets/javascripts/hyrax/app.js
```diff
@@ -134,5 +135,12 @@ Hyrax = {
     statistics: {},
 
     // initialized in hyrax/config.js
-    config: {}
+    config: {},
+
+    // Per Page select will submit its form to change records shown
```
jcoyne: I'm not sure I understand "Per Page select".  Is there another way to say this?

app/assets/javascripts/hyrax/app.js
jcoyne: Why are we calling the file `batch_select` but the constant `BatchSubmit`? Can we decide on submit or select and stick with it?

spec/javascripts/batch_select_spec.js
```diff
@@ -0,0 +1,15 @@
+var BatchSubmit = require('hyrax/batch_select');
+
+describe("BatchSelect", function() {
```
jcoyne: 👏 for adding a test.

jcoyne:
Looks mostly good. Just a few questions around clarity.

mjgiarlo:
Looks like this is close but could use a couple small tweaks from @cam156

cam156:
@jcoyne I updated the comment and decided to go with batch_select as that was the original file name.

PR #1301, opened 2017-06-24
Asynchronous IngestJob and JobIoWrapper

### LINKED ISSUE #1239: Callback timing broken w/ asynchronous jobs
Here we have a method, `update_content`: https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/file_set_actor.rb#L74-L78
It calls `ingest_file` w/ async set true, tossing a job in the queue…. but it immediately calls the callback.
```ruby
      def update_content(file, relation = 'original_file')
        build_file_actor(relation).ingest_file(file, true)
        Hyrax.config.callback.run(:after_update_content, file_set, user)
        true
      end
```
That boolean was changed by @mjgiarlo 2 months ago. The problem is that in realistic scenarios, `Hyrax.config.callback.run(:after_update_content, file_set, user)` happens BEFORE the update.  By toggling it async, the logical relationship between itself and callback was broken.  Also exposed #1155 and https://github.com/samvera-labs/hyku/issues/1182.  

What do we want to do?

1. switch back to synchronous.  Princeton/PSU has been saying “we must retain async”, so that seems out.
2. remove callback
3. replace it w/ different callback from inside the new ingest job

The latter has the advantage of also covering `create_content` (not just `update_content`)… which otherwise doesn’t seem to get a callback, despite being more common and presumably more important.  Depending on what actions are expected in the callbacks, it makes more sense for them to execute in the same thread as the job... or the opposite.  If you are adding a user-notification to flash in the page response, for example, it does nothing from the job context on a worker box.

But, the only way to allow async and a logical `after` is to tie to the job, not the invocation.  Otherwise you risk retrieving and operating on (say, building a secondary index) with the un-updated object.  Note: you don't get another chance later, you just get the wrong object.  I think this is universally unacceptable.

atz:
Fixes #1239 
Follows #1206 

`IngestJob` replaces the removed `IngestFileJob`, and is wired in for asynchronous use as before.

`JobIoWrapper` is persistence-backed, allowing it to be passed to asynchronous jobs.

In actors, where async job is spawned, return it (instead of `true`), in case caller wants to wait for completion or otherwise track/reference it. This also signifies the difference between "action successfully completed" like from `.save` and "a deferred thing will determine success/failure".

Replace one-off `after_update_content` callback with `after_perform` block.  Otherwise, the callback can happen **before** the async event.

In general, I propose that the rails framework callbacks (especially on ActiveJob) are a universally better approach than our bespoke magic in `config/initializers/hyrax_callbacks.rb`, giving us:
- `before` callbacks, symmetric to each `after` callback
- `around` callbacks, combining `before` and `after` actions
- a complete range of other callbacks
- the ability to use asynchronicity without breaking timing logic

If the problem is that some job callers want the callback to be optional while others do not, see the implementation of `:notification` in `IngestJob`'s `after_perform` block for an example that handles that.

TODO:
- update downstream jobs to accept `JobIoWrapper` for completeness
- update last two jobs using `Hyrax::WorkingDirectory` and remove it

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -10,6 +10,40 @@ module Hyrax
         @user = user
       end
 
+      # @!group Asynchronous Operations
+
+      # Spawns asynchronous IngestJob
+      # Called from FileSetsController, AttachFilesToWorkJob, ImportURLJob, IngestLocalFileJob
+      # @param [Hyrax::UploadedFile, File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user
+      # @param [Symbol, #to_s] relation
+      # @return [IngestJob, FalseClass] false on failure, otherwise the queued job
+      def create_content(file, relation = :original_file)
+        # If the file set doesn't have a title or label assigned, set a default.
+        file_set.label ||= label_for(file)
+        file_set.title = [file_set.label] if file_set.title.blank?
+        return false unless file_set.save # Need to save to get an id
+        IngestJob.perform_later(wrapper!(file, relation))
+      end
+
+      # Spawns asynchronous IngestJob with user notification afterward
+      # @param [Hyrax::UploadedFile, File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user
+      # @param [Symbol, #to_s] relation
+      # @return [IngestJob] the queued job
+      def update_content(file, relation = :original_file)
+        IngestJob.perform_later(wrapper!(file, relation), notification: true)
+      end
+
+      # Spawns async ImportUrlJob to attach remote file to fileset
+      # @param [#to_s] url
+      # @return [IngestUrlJob] the queued job
+      def import_url(url)
+        file_set.update(import_url: url.to_s)
+        operation = Hyrax::Operation.create!(user: user, operation_type: "Attach File")
+        ImportUrlJob.perform_later(file_set, operation)
+      end
+
+      # @!endgroup
+
       # Adds the appropriate metadata, visibility and relationships to file_set
       # @note In past versions of Hyrax this method did not perform a save because it is mainly used in conjunction with
       #   create_content, which also performs a save.  However, due to the relationship between Hydra::PCDM objects,
@@ -29,29 +63,16 @@ module Hyrax
         yield(file_set) if block_given?
       end
 
-      # Called from FileSetsController, AttachFilesToWorkJob, ImportURLJob, IngestLocalFileJob
-      # @param [File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user.
-      # @param [Symbol, #to_sym] relation
-      # @return [Boolean] true on success, false otherwise
-      def create_content(file, relation = :original_file)
-        # If the file set doesn't have a title or label assigned, set a default.
-        file_set.label ||= label_for(file)
-        file_set.title = [file_set.label] if file_set.title.blank?
-        return false unless file_set.save # Need to save to get an id
-        build_file_actor(relation).ingest_file(io_decorator(file))
-        true
-      end
-
       # Adds a FileSet to the work using ore:Aggregations.
       # Locks to ensure that only one process is operating on the list at a time.
       def attach_to_work(work, file_set_params = {})
         acquire_lock_for(work.id) do
           # Ensure we have an up-to-date copy of the members association, so that we append to the end of the list.
           work.reload unless work.new_record?
-          copy_visibility(work, file_set) unless assign_visibility?(file_set_params)
+          file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
           work.ordered_members << file_set
-          set_representative(work, file_set)
-          set_thumbnail(work, file_set)
+          work.representative = file_set if work.representative_id.blank?
+          work.thumbnail = file_set if work.thumbnail_id.blank?
           # Save the work so the association between the work and the file_set is persisted (head_id)
           # NOTE: the work may not be valid, in which case this save doesn't do anything.
           work.save
@@ -70,15 +91,6 @@ module Hyrax
         true
       end
 
-      # @param [File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user.
-      # @param [Symbol, #to_sym] relation
-      # @return [Boolean] true on success, false otherwise
-      def update_content(file, relation = :original_file)
-        return false unless build_file_actor(relation).ingest_file(io_decorator(file))
-        Hyrax.config.callback.run(:after_update_content, file_set, user)
-        true
-      end
-
       def update_metadata(attributes)
         env = Actors::Environment.new(file_set, ability, attributes)
         CurationConcern.file_set_update_actor.update(env)
@@ -94,14 +106,6 @@ module Hyrax
         Hyrax::Actors::FileActor
       end
 
-      # Spawns async job to attach file to fileset
-      # @param [#to_s] url
-      def import_url(url)
-        file_set.update(import_url: url.to_s)
-        operation = Hyrax::Operation.create!(user: user, operation_type: "Attach File")
-        ImportUrlJob.perform_later(file_set, operation)
-      end
-
       private
 
         def ability
@@ -112,26 +116,36 @@ module Hyrax
           file_actor_class.new(file_set, relation, user)
         end
 
-        def io_decorator(file)
-          Hydra::Derivatives::IoDecorator.new(file, mime_for(file), label_for(file))
+        # uses create! because object must be persisted to serialize for jobs
+        def wrapper!(file, relation)
+          JobIoWrapper.create!(wrapper_params(file, relation))
         end
 
-        # @return [String, nil] mime type explicitly set or sniffed
-        def mime_for(file)
-          if file.respond_to?(:content_type) # e.g. ActionDispatch::Http::UploadedFile, CarrierWave::SanitizedFile
-            file.content_type
-          elsif file.respond_to?(:mime_type) # e.g. Hydra::Derivatives::IoDecorator
-            file.mime_type
-          elsif file.respond_to?(:path) # e.g. File (note: will be meaningless for Tempfile)
-            MIME::Types.type_for(File.extname(file.path)).first.content_type
-          end # else nil
+        # helps testing
+        # @return [Hash] params for JobIoWrapper (new or create)
+        def wrapper_params(file, relation)
+          args = { user: user, relation: relation.to_s, file_set_id: file_set.id }
+          if file.is_a?(Hyrax::UploadedFile)
+            args[:uploaded_file] = file
+            args[:path] = file.uploader.path
+          elsif file.respond_to?(:path)
+            args[:path] = file.path
+            args[:original_name] = file.original_filename if file.respond_to?(:original_filename)
+            args[:original_name] ||= file.original_name if file.respond_to?(:original_name)
+          else
+            raise "Require Hyrax::UploadedFile or File-like object, received #{file.class} object: #{file}"
+          end
+          args
         end
 
         # For the label, use the original_filename or original_name if it's there.
         # If the file was imported via URL, parse the original filename.
         # If all else fails, use the basename of the file where it sits.
+        # @note This is only useful for labeling the file_set, because of the recourse to import_url
         def label_for(file)
-          if file.respond_to?(:original_filename) # e.g. ActionDispatch::Http::UploadedFile, CarrierWave::SanitizedFile
+          if file.is_a?(Hyrax::UploadedFile)
+            file.uploader.filename
+          elsif file.respond_to?(:original_filename) # e.g. ActionDispatch::Http::UploadedFile, CarrierWave::SanitizedFile
             file.original_filename
           elsif file.respond_to?(:original_name) # e.g. Hydra::Derivatives::IoDecorator
             file.original_name
@@ -143,34 +157,8 @@ module Hyrax
           end
         end
 
-        # Takes an optional block and executes the block if the save was successful.
-        # @return [Boolean] false if the save was unsuccessful
-        def save
-          on_retry = ->(exception, _, _, _) { ActiveFedora::Base.logger.warn "Hyrax::Actors::FileSetActor#save Caught RSOLR error #{exception.inspect}" }
-          Retriable.retriable on: RSolr::Error::Http, on_retry: on_retry, tries: 4, base_interval: 0.01 do
-            return false unless file_set.save
-          end
-          yield if block_given?
-          true
-        end
-
         def assign_visibility?(file_set_params = {})
           !((file_set_params || {}).keys.map(&:to_s) & %w[visibility embargo_release_date lease_expiration_date]).empty?
-        end
-
-        # copy visibility from source_concern to destination_concern
-        def copy_visibility(source_concern, destination_concern)
-          destination_concern.visibility =  source_concern.visibility
-        end
-
-        def set_representative(work, file_set)
-          return if work.representative_id.present?
-          work.representative = file_set
-        end
-
-        def set_thumbnail(work, file_set)
-          return if work.thumbnail_id.present?
-          work.thumbnail = file_set
         end
 
         def unlink_from_work
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -14,7 +14,7 @@ module Hyrax
 
       # Spawns asynchronous IngestJob
       # Called from FileSetsController, AttachFilesToWorkJob, ImportURLJob, IngestLocalFileJob
-      # @param [Hyrax::UploadedFile, File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user
+      # @param [Hyrax::UploadedFile, File, ActionDigest::HTTP::UploadedFile] file the file uploaded by the user
       # @param [Symbol, #to_s] relation
       # @return [IngestJob, FalseClass] false on failure, otherwise the queued job
       def create_content(file, relation = :original_file)
@@ -26,7 +26,7 @@ module Hyrax
       end
 
       # Spawns asynchronous IngestJob with user notification afterward
-      # @param [Hyrax::UploadedFile, File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user
+      # @param [Hyrax::UploadedFile, File, ActionDigest::HTTP::UploadedFile] file the file uploaded by the user
       # @param [Symbol, #to_s] relation
       # @return [IngestJob] the queued job
       def update_content(file, relation = :original_file)
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -39,7 +39,15 @@ module Hyrax
 
     # routed to /files (POST)
     def create
-      create_from_upload(params)
+      file = params.fetch(:file_set, {}).fetch(:files, []).detect { |f| f.respond_to?(:original_filename) }
+      return render_json_response(response_type: :bad_request, options: { message: 'Error! No file uploaded', description: 'missing file' }) unless file
+      return empty_file_response(file) if empty_file?(file)
+      process_non_empty_file(file: file)
+    rescue RSolr::Error::Http => error
+      logger.error "FileSetController::create rescued #{error.class}\n\t#{error}\n #{error.backtrace.join("\n")}\n\n"
+      render_json_response(response_type: :internal_error, options: { message: 'Error occurred while creating a FileSet.' })
+    ensure
+      file.tempfile.delete if file.respond_to?(:tempfile) # remove tempfile (only if it is a temp file)
     end
 
     # routed to /files/:id
@@ -80,33 +88,12 @@ module Hyrax
 
     private
 
-      def create_from_upload(params)
-        # check error condition No files
-        return render_json_response(response_type: :bad_request, options: { message: 'Error! No file to save' }) unless params.key?(:file_set) && params.fetch(:file_set).key?(:files)
-
-        file = params[:file_set][:files].detect { |f| f.respond_to?(:original_filename) }
-        attempt_process_file_for_create_from_upload(file)
-      rescue RSolr::Error::Http => error
-        logger.error "FileSetController::create rescued #{error.class}\n\t#{error}\n #{error.backtrace.join("\n")}\n\n"
-        render_json_response(response_type: :internal_error, options: { message: 'Error occurred while creating a FileSet.' })
-      ensure
-        # remove the tempfile (only if it is a temp file)
-        file.tempfile.delete if file.respond_to?(:tempfile)
-      end
-
-      def attempt_process_file_for_create_from_upload(file)
-        if !file
-          render_json_response(response_type: :bad_request, options: { message: 'Error! No file for upload', description: 'unknown file' })
-        elsif empty_file?(file)
-          empty_file_response(file)
-        else
-          process_non_empty_file(file: file)
-        end
-      end
-
       def process_non_empty_file(file:)
-        update_metadata_from_upload_screen
-        if process_file(actor, file)
+        # Relative path is set by the jquery uploader when uploading a directory
+        curation_concern.relative_path = params[:relative_path] if params[:relative_path]
+        actor.create_metadata(params[:file_set])
+        actor.attach_to_work(find_parent_by_id)
+        if actor.create_content(file)
           response_for_successfully_processed_file
         else
           msg = curation_concern.errors.full_messages.join(', ')
@@ -115,20 +102,12 @@ module Hyrax
         end
       end
 
-      def process_file(actor, file)
-        actor.create_metadata(params[:file_set])
-        actor.attach_to_work(find_parent_by_id)
-        actor.create_content(file)
-      end
-
       def empty_file_response(file)
-        render_json_response(response_type: :unprocessable_entity,
-                             options: {
-                               errors: {
-                                 files: "#{file.original_filename} has no content! (Zero length file)"
-                               },
-                               description: t('hyrax.api.unprocessable_entity.empty_file')
-                             })
+        options = {
+          errors: { files: "#{file.original_filename} has no content! (Zero length file)" },
+          description: t('hyrax.api.unprocessable_entity.empty_file')
+        }
+        render_json_response(response_type: :unprocessable_entity, options: options)
       end
 
       def response_for_successfully_processed_file
@@ -207,14 +186,9 @@ module Hyrax
       end
 
       def initialize_edit_form
-        @version_list = version_list
-        @groups = current_user.groups
-      end
-
-      def version_list
         original = @file_set.original_file
-        versions = original ? original.versions.all : []
-        Hyrax::VersionListPresenter.new(versions)
+        @version_list = Hyrax::VersionListPresenter.new(original ? original.versions.all : [])
+        @groups = current_user.groups
       end
 
       def actor
@@ -238,11 +212,8 @@ module Hyrax
         params.key?(:revision) && params[:revision] != curation_concern.latest_content_version.label
       end
 
-      # Override this method to add additional response
-      # formats to your local app
-      def additional_response_formats(_)
-        # nop
-      end
+      # Override this method to add additional response formats to your local app
+      def additional_response_formats(_); end
 
       def file_set_params
         params.require(:file_set).permit(
@@ -252,12 +223,6 @@ module Hyrax
 
       def empty_file?(file)
         (file.respond_to?(:tempfile) && file.tempfile.size == 0) || (file.respond_to?(:size) && file.size == 0)
-      end
-
-      # this is provided so that implementing application can override this behavior and map params to different attributes
-      def update_metadata_from_upload_screen
-        # Relative path is set by the jquery uploader when uploading a directory
-        curation_concern.relative_path = params[:relative_path] if params[:relative_path]
       end
 
       # This allows us to use the unauthorized and form_permission template in hyrax/base,
--- a/app/jobs/ingest_job.rb
+++ b/app/jobs/ingest_job.rb
@@ -2,6 +2,7 @@ class IngestJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
 
   after_perform do |job|
+    # We want the lastmost Hash, if any.
     opts = job.arguments.reverse.detect { |x| x.is_a? Hash } || {}
     wrapper = job.arguments.first
     ContentNewVersionEventJob.perform_later(wrapper.file_set, wrapper.user) if opts[:notification]
--- a/app/jobs/ingest_job.rb
+++ b/app/jobs/ingest_job.rb
@@ -1,0 +1,17 @@
+class IngestJob < Hyrax::ApplicationJob
+  queue_as Hyrax.config.ingest_queue_name
+
+  after_perform do |job|
+    opts = job.arguments.reverse.detect { |x| x.is_a? Hash } || {}
+    wrapper = job.arguments.first
+    ContentNewVersionEventJob.perform_later(wrapper.file_set, wrapper.user) if opts[:notification]
+  end
+
+  # @param [JobIoWrapper] wrapper
+  # @param [Boolean] notification send the user a notification, used in after_perform callback
+  # @see 'config/initializers/hyrax_callbacks.rb'
+  # rubocop:disable Lint/UnusedMethodArgument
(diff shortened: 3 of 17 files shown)


app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -143,36 +157,10 @@ def label_for(file)
           end
         end
 
-        # Takes an optional block and executes the block if the save was successful.
-        # @return [Boolean] false if the save was unsuccessful
-        def save
```
atz: Removed this unused weirdo method.

atz:
@tpendragon, @escowles: I think this accomplishes what we wanted.

atz:
Weird.  I don't understand how Travis can get this error that I don't get locally:
```
Failure/Error: expect { described_class.create!(file_set_id: file_set_id, path: path) }.to raise_error ActiveRecord::RecordInvalid
       expected ActiveRecord::RecordInvalid but nothing was raised
```
The model has:
```
belongs_to :user
```

`belongs_to` is a [default required relationship](http://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html) in rails 5+, and the version of rails is [5.0.4](https://travis-ci.org/samvera/hyrax/jobs/246863681#L725).  Therefore, `user` is required.  So how can `described_class.create!(file_set_id: file_set_id, path: path)` succeed?  Only on Travis?

escowles:
@atz I'm not sure I understand exactly what's going on here — this is modeling the uploaded file in ActiveRecord, so it can be serialized/deserialized by an ActiveJob?  If so, then the shared filesystem requirement is still there.

The benefit then, is that the ActiveRecord object can be used to figure out when the background job has been run, as opposed to simply passing the local filename (where you would have no way of knowing if it was safe to remove the temp file).

Assuming I'm understanding this correctly, that seems like a nice improvement.

atz:
@escowles: Yes, serialization is indispensable.  The model also allows `#read` (and `#size`) to use either a File opened on the provided pathname or a referenced `Hyrax::UploadedFile`.  This is basically similar to what `Hyrax::WorkingDirectory` was trying to do, but starting from the data and reducing the complexity of arguments/logic inside the job.

atz:
Also the main point is this gets back async at the top level of ingest (`IngestJob.perform_later` instead of inline).

atz:
@escowles Added some additional testing to make that flexibility more obvious.  Let me know if you have any questions.

mjgiarlo:
@atz I see two TODOs listed in the description above. Are these coming in this PR, or do you see them as follow-up work?

mjgiarlo:
@atz If we wanted to test this branch in Hyku (deployed in AWS0, would we need to make any changes at that layer? Or should it work in that environment?

atz:
@mjgiarlo That's follow-on work.  No known problems w/ Hyku (but written and tested exclusively at Hyrax level).

app/models/job_io_wrapper.rb
```diff
@@ -0,0 +1,79 @@
...
+# for the metadata that might travel w/ the actor's various supported types of @file.
+# For example, we cannot just do:
+#
+#   SomeJob.perform_later(arg1, arg2, File.new('/path/to/file'))
+#
+# Because we'll get:
+#
+#   ActiveJob::SerializationError: Unsupported argument type: File
+#
+# This also applies to Hydra::Derivatives::IoDecorator, Tempfile, etc., pretty much any IO.
```
jcoyne: How could this apply to a Tempfile?  That would be removed as soon as the process ended, right?

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -10,6 +10,40 @@ def initialize(file_set, user)
         @user = user
       end
 
+      # @!group Asynchronous Operations
+
+      # Spawns asynchronous IngestJob
+      # Called from FileSetsController, AttachFilesToWorkJob, ImportURLJob, IngestLocalFileJob
+      # @param [Hyrax::UploadedFile, File, ActionDigest::HTTP::UploadedFile, Tempfile] file the file uploaded by the user
```
jcoyne: You can't accept Tempfile here. Because you can't run an asynchronous job on a file that is temporary.

spec/actors/hyrax/actors/file_set_actor_spec.rb
```diff
@@ -223,80 +247,6 @@
...
-    context "when file set params with visibility are specified with strings as keys" do
-      let(:file_set_params) { { "visibility" => Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC } }
-
-      it "does need to assign visibility" do
-        expect(actor.send(:assign_visibility?, file_set_params)).to eq true
-      end
-    end
-  end
-
-  describe "#set_representative" do
```
jcoyne: Why was this test coverage removed? Previously we covered thumbnail set and unset. By removing this method, which was unnecessary to acchieving the JobIoWrapper, you've made the code less testable.  I'd like to see these tests and their methods restored.

mjgiarlo:
@atz I tested this branch in Hyku, specifically the docker-compose instance, and uploaded files never seemed to get attached. I didn't have enough time to troubleshoot, so I pushed this up for further poking: https://github.com/samvera-labs/hyku/pull/1303

atz:
@mjgiarlo: that is also a complaint about the *current* Hyrax code.

spec/actors/hyrax/actors/file_set_actor_spec.rb
atz: The functionality is still tested.  Expecting a one-line like:
```ruby
``` 
to require its own broken out test coverage is **absurd**.

mjgiarlo:
@atz Yeah, that's true. We pointed Hyku at a Hyrax commit before #1206 landed so the demo instance works for folks: https://github.com/samvera-labs/hyku/pull/1292 I was testing the hypothesis that this PR would get Hyrax working for Hyku again, and that's not yet the case.

app/actors/hyrax/actors/file_set_actor.rb
atz: That's what `wrapper!` is for.  It's the point of this PR.  We can accept `Tempfile` here, because this is not the job.  We pass the non-temp `JobIoWrapper` object to the job.

app/models/job_io_wrapper.rb
atz: This identifies that `Tempfile` (and other IO classes) cannot be serialized for a job, let alone deserialized in another process.

atz:
Rebased and re-rubocopped.  

From all the folks who cared strongly about having an async top-level ingest job a week ago, I might expect more review attention here, since this is what delivers it.

mjgiarlo:
@atz I'm going to take this for a test-drive today. Thanks for yr patience, sir.

mjgiarlo:
@atz I want to correct my last comment on this PR. Namely:

> With samvera-labs/hyku#1303, I was testing the hypothesis that this branch would get Hyrax working in Hyku again, and that's not yet the case.

This was incorrect -- while I did point Hyku at the branch for this PR, I failed to notice that I needed to generate and load a new Hyrax database migration. When I did that, this PR restored the ability to attach files in Hyku. Thanks for this work, @atz, and apologies for the misunderstanding I caused earlier.

jcoyne:
@atz my concern here is that you are changing code that does not need to be changed to accomplish the goal of this PR. Someone submitted this code (@val99erie) in it's current format and it went through review.  They probably had a reason for writing it that way and defended it in the review.  So it's disrespectful of their efforts to just change it without a reason.  If you feel that change is important to make, please break it into a different pull request and submit it with you reason for changing it.

app/actors/hyrax/actors/file_set_actor.rb
mjgiarlo: AFAIK, this code is nearly five years old to the day: https://github.com/samvera/sufia/commit/530868950a314e4dfdd78f2f02c262636bc70129#diff-47b795ef43d49d2d98f2e4a700c306b5R163

I'm not sure we still need it.

mjgiarlo: Can you also remove the dependency on `retriable` from the gemspec? This was the only code that used that gem, AFAIK, so now that it's gone we no longer need the dependency.

atz:
I think the code was unnecessarily indirect and would have gladly objected at that time.  Even if it was defensible then on some premise, I don't think it is now.  I also removed an unused method, for which the same arguments about provenance, precedence and previous intent could be made.  

@val99erie is free to offer comment here, if it is considered important/defensible, and I would certainly consider the author's position on any given code with unique interest (but it has been 11 days already, I won't wait forever).  Broadly, it isn't disrespectful to maintain and improve others work, and much of the value of private methods is that they allow for seemless refactoring or removal.  On the other hand, it is disrespectful to hold up diligent and material work over hypothetical stylistic concerns offered by proxy, in my opinion.  The problem is not that we lack deference to legacy code, but much the opposite.  

@jcoyne Do you personally think it is important to atomically test (what is effectively) individual lines of code?  Or just that since somebody once thought it was wise, we should stick to that indefinitely?  The cleanup was useful to me in understanding what I needed to understand to accomplish and test this PR.  Its inclusion is not arbitrary.  As its author, I can tell you the "goal of this PR" includes **many** improvements to structure, reliability, documentation, testing and logic, including that cleanup.

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -254,12 +225,6 @@ def empty_file?(file)
         (file.respond_to?(:tempfile) && file.tempfile.size == 0) || (file.respond_to?(:size) && file.size == 0)
       end
 
-      # this is provided so that implementing application can override this behavior and map params to different attributes
```
mjgiarlo: Do we think that the need expressed in this comment is no longer around? This ability to override the controller has been around for a long time, since https://github.com/samvera/sufia/commit/92a3788ec799183e862e1a636cdabebf305e5374.

app/jobs/ingest_job.rb
```diff
@@ -0,0 +1,17 @@
+class IngestJob < Hyrax::ApplicationJob
+  queue_as Hyrax.config.ingest_queue_name
+
+  after_perform do |job|
+    opts = job.arguments.reverse.detect { |x| x.is_a? Hash } || {}
```
mjgiarlo: Why is the `#reverse` needed here?

app/jobs/ingest_job.rb
```diff
@@ -0,0 +1,17 @@
...
+    wrapper = job.arguments.first
+    ContentNewVersionEventJob.perform_later(wrapper.file_set, wrapper.user) if opts[:notification]
```
mjgiarlo: I'm not sure what a better approach would look like, but I'm wondering if this line will get us into trouble later because of where and when this coupling is expressed. For instance, a downstream app may want to add notifications with other jobs, but we don't want to assume all jobs with `notification: true` correspond to new version events. The callbacks used before -- though invoked at the incorrect time 'til this PR -- had the advantage of giving us more flexibility, which could be overridden by downstream consumers. Can you suggest an improvement here? FWIW, I wouldn't block the PR over this, but I would like to ticket this for follow-on work.

app/models/job_io_wrapper.rb
```diff
@@ -0,0 +1,79 @@
...
+  def mime_type
+    super || extracted_mime_type
+  end
+
+  def file_set
+    FileSet.find(file_set_id)
+  end
+
+  def file_actor
+    Hyrax::Actors::FileActor.new(file_set, relation.to_sym, user)
```
mjgiarlo: Can/ought this be private too?

app/jobs/ingest_job.rb
atz: We want the "lastmost hash", if any.  In this case is isn't particularly important, now that the number of arguments is small (1-2), but it minimizes the number of times the `detect` block is evaluated (max once).

spec/actors/hyrax/actors/file_actor_spec.rb
```diff
@@ -34,7 +34,7 @@ class FileSetWithExtras < FileSet
     end
     it 'uses the relation from the actor' do
       expect(CharacterizeJob).to receive(:perform_later).with(file_set, String, io.tempfile.path)
-      expect(actor.ingest_file(io)).to be_truthy
+      actor.ingest_file(io)
```
mjgiarlo: Do we expect this to at least not return `false` here, or do the tests on the preceding and succeeding lines suffice? (I suspect the latter, that if the `CharacterizeJob` is called successfully and the reloaded mime type is as we expect it, we no longer need to check the return value of `#ingest_file`.)

app/models/job_io_wrapper.rb
atz: I considered that, but other jobs receiving `JobIoWrapper` objects are going to act directly via this `file_actor`.  The only reason it isn't currently directly invoked is the convenience method for `ingest_file`.  All _other_ actions (via the actor) would invoke `file_actor` or require their own convenience method.  I didn't want to preclude any methods in the actor from being used in jobs or force jobs to manually reinstantiate their own file actor.

spec/actors/hyrax/actors/file_set_actor_spec.rb
```diff
@@ -25,17 +71,9 @@
...
-      # NOTE: we have to use this unweidly block instead of `.with(io)`
-      # because Hydra::Derivatives::IoDecorator objects cannot be compared
-      # https://github.com/samvera/hydra-derivatives/issues/165
-      # We reuse this pattern throughout this file.
-      # TODO: replace IoDecorator with a regular Hyrax::Io class
-      expect(file_actor).to receive(:ingest_file) do |input|
-        expect(input.mime_type).to eq(uploaded_file.content_type)
-        expect(input.original_name).to eq(uploaded_file.original_filename)
-      end
+      # expect(actor).to receive(:wrapper).with(file, relation)
```
mjgiarlo: Can this comment be removed?

jcoyne:
@atz I don't think line coverage is important. However, I believe that we ought to cover all the interesting scenarios (with a thumbnail, without one, with a representative and without).  Testing this when the logic all in one method is pretty difficult.  I find that it is easier to expect that `set_thumbnail` and `set_representative` are invoked in one tests, and then test their variant behavior in other tests.

spec/actors/hyrax/actors/file_set_actor_spec.rb
```diff
@@ -227,80 +254,6 @@
     end
   end
 
-  describe "#assign_visibility?" do
```
mjgiarlo: Why'd you remove the `#assign_visibility?` block of tests? It doesn't look like you touched this method's implementation.

mjgiarlo: Nevermind, I now see these tests moved here: https://github.com/samvera/hyrax/pull/1301/files#diff-1950b382cdd95aed6b463a2f079106cdR52 Right?

spec/actors/hyrax/actors/file_set_actor_spec.rb
mjgiarlo: @atz Can you point me at where this is still tested?

app/actors/hyrax/actors/file_set_actor.rb
atz: I'd be happy to attempt that.

app/jobs/ingest_job.rb
mjgiarlo: @atz OK. Could you add that as a comment above this line? That's helpful to know.

app/jobs/ingest_job.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  # @param [JobIoWrapper] wrapper
+  # @param [Boolean] notification send the user a notification, used in after_perform callback
+  # @see 'config/initializers/hyrax_callbacks.rb'
+  # rubocop:disable Lint/UnusedMethodArgument
```
mjgiarlo: Do we need to re-enable this rule after line 16, or will Rubocop do the right thing? I'd prefer to be explicit here, but don't care *too* much.

app/models/job_io_wrapper.rb
mjgiarlo: K, works for me, @atz.

app/controllers/hyrax/file_sets_controller.rb
atz: This description, the method name and the method content seem unrelated to each other.  It being a private method means it is **not** part of the public contract of the class, regardless of the comment's claim.  The approach that says (several times over in this class) "override this private method" is a fundamental OO error we should avoid.  Basic OO approaches or well-known Rails callbacks would be appropriate directions for Hyrax to point users towards instead of overriding private methods.

mjgiarlo: They may seem unrelated, but I believe they're being used by folks to make overrides such as:

* https://github.com/duke-libraries/course-assets/blob/2c8a64687649eb2ebc8cf6777edf270ada1e3ed2/lib/course_assets/files_controller_behavior.rb#L5-L7
* https://github.com/ualbertalib/HydraNorth/blob/d4a1c87cf34e376f69138c411cc21819f1edb2f2/app/controllers/concerns/hydranorth/files_controller_behavior.rb#L61-L65
* https://github.com/ihrnexuslab/research-repo/blob/572865c2125c18891762abaa0efdeb2d9895da7d/app/controllers/generic_files_controller.rb#L121-L132
* https://github.com/sephirothkod/Sufia-Head/blob/7994b6eb6eeb29ac40b3d243af8f65bc07d43947/app/controllers/concerns/sufia/files_controller_behavior.rb#L286-L297

Given that folks are still using this seam, and it's not necessary for this PR (methinks?), would you consider reverting this change?

app/jobs/ingest_job.rb
atz: It wasn't more flexibility at all.  `ActiveJob` callbacks are *far* more flexible than our one-offs **and actually work** without logical errors.  We had `before` callbacks with no symmetric `after`.  We don't provide matching `around` callbacks.  Rails has already worked this stuff out to a vast degree, and already has symmetric and complete before/after/around hooks.  In the context of jobs, I don't see our one-off callbacks as ever preferable.  

But there are various approaches to this that I considered (some, even implemented) before this proposed one that might be useful to restate:
- separate jobs: `IngestJob` and `IngestJobWithNotification`.  The latter extends the former and unilaterally applies the `after_perform` block.  PRO: options fall away.  CON: would need a descendant job for each existing job.  If a second (or more) such runtime option emerged, it would become a nonsensical matrix of classes, like `IngestJobWithBellsWithNotification` and `IngestJobWithBellsWithWhistlesWithNotification`
- some refinement, where the `after_perform` block moves into a base class (or even the main `Hyrax::ActiveJob` class).  I would consider implementing this when a second job adopts the same pattern, not before.
- adding to `JobIoWrapper`: did not pursue this because it seemed distinct from the local filesystem problem the class intends to solve.  The *same* `JobIoWrapper` (uploaded file) might be expected to be passed to a series of jobs, some of them with notifications, some without.  We shouldn't affect the persisted representation of the wrapper just to reflect that.  Indeed we CANNOT, if a given job wants to invoke two separate `perform_later`s, it would be, e.g.: 
```
Job1.perform_later(wrapper)
wrapper.notification = false
wrapper.save
Job2.perform_later(wrapper)
```
But given that we don't know when the jobs process, i.e. when the persisted wrapper is retrieved from DB, this will allow `Job1` to retrieve the state we intended for `Job2`.

app/controllers/hyrax/file_sets_controller.rb
atz: Not really.  It's fundamentally bad form and I cannot support it.  If you want, I would remove ALL other references to overriding private methods.  I would *consider* someone else's PR to make all such methods `public`, but only with significant skepticism given how scattershot they seem to be.  

We are trying to be **better** than sufia.  Keeping people dependent on hacks is not doing them any favors.

spec/actors/hyrax/actors/file_actor_spec.rb
atz: Right, doesn't matter much and we get into circular (not particularly informative) testing logic.  `CharacterizeJob.perform_later` is the last thing in the method.  We don't actually want that job to execute here, and we want to be sure it gets called, so we have an `expect` catch that.  At that point, we have assumed control of what the "return value" would be, so testing for it would be unnecessary.

app/jobs/ingest_job.rb
atz: No, it only affects this file, which is now so terse as to only include 1 method.

spec/actors/hyrax/actors/file_set_actor_spec.rb
atz: Right, I grouped all the private methods being tested together in the spec.

spec/actors/hyrax/actors/file_set_actor_spec.rb
atz: ok

app/controllers/hyrax/file_sets_controller.rb
mjgiarlo: I don't disagree with any of that, @atz, but I wonder how removing the code in question helps this PR.

Or maybe we're not as concerned with this specific override capability in Hyrax as we were in Sufia? Thoughts welcome from @jcoyne, the original author of this method.

app/jobs/ingest_job.rb
mjgiarlo: @atz :speech_balloon: 
> separate jobs: IngestJob and IngestJobWithNotification. The latter extends the former and unilaterally applies the after_perform block. PRO: options fall away. CON: would need a descendant job for each existing job. If a second (or more) such runtime option emerged, it would become a nonsensical matrix of classes, like IngestJobWithBellsWithNotification and IngestJobWithBellsWithWhistlesWithNotification

Yeah, :-1: 

> some refinement, where the after_perform block moves into a base class (or even the main Hyrax::ActiveJob class). I would consider implementing this when a second job adopts the same pattern, not before.

:+1: to the "after, not before" comment. Now's not the right time for that.

> adding to JobIoWrapper: did not pursue this because it seemed distinct from the local filesystem problem the class intends to solve. The same JobIoWrapper (uploaded file) might be expected to be passed to a series of jobs, some of them with notifications, some without. We shouldn't affect the persisted representation of the wrapper just to reflect that.

OK. What about adding an arg to, or changing the second arg in, `IngestJob` to allow consumers to inject a notification class or the like? That way the code could look like:

```ruby
IngestJob.perform_later(wrapper, notification: true, notification_class: ContentNewVersionEvent)
```

Though maybe we'd need to pass that as a string rather than class name, and the complicating factor here is that event jobs don't all receive the same arguments, so we'd need some way to deal with that. Thinking aloud here.

spec/actors/hyrax/actors/file_actor_spec.rb
mjgiarlo: WFM. :+1:

spec/actors/hyrax/actors/file_set_actor_spec.rb
mjgiarlo: :cool:

spec/actors/hyrax/actors/file_set_actor_spec.rb
mjgiarlo: Nice. :beers:

app/jobs/ingest_job.rb
mjgiarlo: K.

app/jobs/ingest_job.rb
atz: I'm not sure what you're asking for.  `We want the lastmost hash`?

app/jobs/ingest_job.rb
atz: Correct, event jobs do not take the same args or the same number of args.  That latter idea sounds like making the thing you're objecting to worse, no?  Also the caller doesn't get to decide what *type* of action this job considers itself to implement.  If we wanted to go that way (I don't), we would cut out the boolean, and just use the presence of a `notification_class` to signal `true`.

app/controllers/hyrax/file_sets_controller.rb
atz: I'm sick of dancing around completely busted patterns that are somehow still regarded as sacred.  It is funny because normally in this situation, legacy status quo folks will argue "we need a deprecation warning!".   But that reveals how messed up this situation is, since **you don't get deprecation warnings for private methods**!  IF this hack is important to somebody, it is on them to provide an implementation here that is not an abomination.  We move them in the right direction by removing it (and the rest).

spec/actors/hyrax/actors/file_set_actor_spec.rb
atz: It was in integrations.  But for ease of mind over what is, effectively, two lines of code and to get away from overuse of integration testing, I added some explicit tests in the actor spec for `#attach_to_work`:
```ruby
    context 'without representative and thumbnail' do
      it 'assigns them (with persistence)' do
        actor.attach_to_work(work)
        expect(work.representative).to eq(file_set)
        expect(work.thumbnail).to eq(file_set)
        expect { work.reload }.not_to change { [work.representative, work.thumbnail] }
      end
    end

    context 'with representative and thumbnail' do
      it 'does not (re)assign them' do
        allow(work).to receive(:thumbnail_id).and_return('ab123c78h')
        allow(work).to receive(:representative_id).and_return('zz365c99q')
        expect(work).not_to receive(:representative=)
        expect(work).not_to receive(:thumbnail=)
        actor.attach_to_work(work)
      end
    end
```

I trust this is the last we'll have to talk about how the logic is.

app/controllers/hyrax/file_sets_controller.rb
atz: @mjgiarlo The code you link to is all forked anyway:
- `ihrnexuslab`'s `update_metadata_from_upload_screen` requires a param!  That means they are already overriding the method that invokes it, too (and not calling `super`).  
- `sephirothkod`: same thing
- `HyrdraNorth` decided everything in their `FilesControllerBehavior` including this method would be `protected`... via a module, no less!  
- Duke has the cleanest implementation, but makes the method `public`.  Like others, it has decided an instance variable named `@generic_file` means something.  But it doesn't in Hyrax.

app/jobs/ingest_job.rb
mjgiarlo: @atz How about: `# We want the "lastmost hash", if any. It isn't particularly important, since the number of arguments is small (1-2), but it minimizes the number of times the detect block is evaluated (max once).`

mjgiarlo: Feel free to rework/shorten/simplify.

app/jobs/ingest_job.rb
mjgiarlo: @atz What's happening here, as I currently understand it, is that we use `IngestJob` for two tasks: 1) to create a new work and 2) to upload a new version of a file to a work. We want notifications for both, but currently they're handled in different spots in the code, with only the latter triggered within the job. (The former is triggered by `Hyrax::Actors::BaseActor`.) The code we have now retains past behavior (triggering new version events when new versions are uploaded), so I'd be willing to punt this concern to a later date if it becomes problematic.

app/controllers/hyrax/file_sets_controller.rb
mjgiarlo: @atz Those are admittedly all on Sufia, some older than others.

@jcoyne How much do you care about this change? Based on what I'm seeing in other repos, I'm not too worried about this complicating the CC to Hyrax bridge and thus I'm inclined to stop caring about it in the context of this PR.

spec/actors/hyrax/actors/file_set_actor_spec.rb
mjgiarlo: Great, thanks, @atz!

app/actors/hyrax/actors/file_set_actor.rb
jcoyne: Since this runs an asynchronous job, won't this end up with the Tempfile unavailable for the job?

app/controllers/hyrax/file_sets_controller.rb
jcoyne: I don't really care about this change.  I do think we should endeavor to keep our PRs small and not change things that aren't related to the stated purpose of the PR.  I would have liked to have seen this change in a separate PR like "Reduce indirection in FileSetsController"

mjgiarlo: OK, let's say :+1:  to this change for now, then.

app/actors/hyrax/actors/file_set_actor.rb
atz: You are right that `IngestJob` won't work on a `Tempfile`.  But unlike its predecessor, this method doesn't attempt to pass the `Tempfile` directly to the job.  Instead it calls `IngestJob.perform_later(wrapper!(file, relation))`, where `wrapper!` effectively converts `file` to a persisted `JobIoWrapper` to match the signature of `IngestJob`.  Make sense?

app/jobs/ingest_job.rb
atz: @mjgiarlo: The job is invoked for both `create_content` and `update_content`.  Only the latter wants notifications (presumably because it is more targeted than the 100 notifications you might get from a single initial directory upload).  I don't know if it overlaps w/ `BaseActor` or not, but the two distinct uses is what prompted the option/param.

app/actors/hyrax/actors/file_set_actor.rb
jcoyne: Yes, the wrapper thing makes sense. I'm just calling out a potential problem with the Yard documentation.  I don't think we want to accept a Tempfile, because they aren't going to have much success with it (even though the exception wouldn't occur until the background job)

app/jobs/ingest_job.rb
mjgiarlo: @atz Right. At the moment, there are six event jobs around content changes:

https://github.com/samvera/hyrax/blob/master/config/initializers/hyrax_callbacks.rb#L2-L24

* `:after_create_concern` creates an event for every new work, and it fires from the actor stack: https://github.com/samvera/hyrax/blob/68527862df498aa9f407b909d4f41d781ba146dc/app/actors/hyrax/actors/base_actor.rb#L14
* `:after_create_fileset` creates an event when a fileset is attached to a work using `attach_to_work`: https://github.com/samvera/hyrax/blob/77140b38907158aaa4e30aae442ea5f04d0b1be2/app/actors/hyrax/actors/file_set_actor.rb#L58
* `:after_revert_content` creates an event when a fileset is reverted to a past version: https://github.com/samvera/hyrax/blob/77140b38907158aaa4e30aae442ea5f04d0b1be2/app/actors/hyrax/actors/file_set_actor.rb#L69
* `:after_update_content` you're well acquainted with
* `:after_update_metadata` creates an event when a work is updated (as distinct from when a fileset is updated), invoked by the actor stack: https://github.com/samvera/hyrax/blob/68527862df498aa9f407b909d4f41d781ba146dc/app/actors/hyrax/actors/base_actor.rb#L22
* `:after_destroy` creates an event when a work is destroyed, called from the works controller: https://github.com/samvera/hyrax/blob/45b988145d76dd5c2e659aa0cd53737e7d138c3a/app/controllers/concerns/hyrax/works_controller_behavior.rb#L117

Part of why we're using the callbacks in different spots in the codebase could be that some of our actors are meant to be used in a stack and others are meant to be used alone (more akin to "service objects"?). I'm definitely not suggesting this PR needs to tackle this problem. ;) I'm leaving this comment here to show that we considered changes to how we approach this but that we decided to punt for good reason.

app/actors/hyrax/actors/file_set_actor.rb
mjgiarlo: @atz @jcoyne Do y'all have a sense of when `Tempfile`s are used in this context? For instance, is this going to hose us in AWS-hosted environments?

jcoyne: Tempfiles used to be used when we copied something from s3 to local. I don't know how they would be used now after the refactoring.

atz: In practice, as currently used in the web app, we mostly get the fancier variety of params, so I'm OK with removing this from the YARD.  However, there is existing code referencing Tempfiles related to arkivo and `ImportUrlJob`.  If we want a technical fix here, it might involve: 
1. raise if given a Tempfile (even though on simplistic single-box inline systems, it would otherwise work)
1. automatically convert Tempfile to a `Hyrax::UploadedFile`
1. convert any remaining callers to passing non-Tempfile objects

I might be for some combination of 1 and 3 in a follow-up.

Note: there may be other kinds of "self destructing" file classes that will otherwise successfully pass through this method (and present the same issue jcoyne is talking about).

jcoyne: I'm okay with just removing the documentation.  I understand this is Ruby and we can't prevent the user from shooting themselves in the foot, but we should at least not advertise something we know to be bad.

PR #1302, opened 2017-06-27
Fix label of open access

mjgiarlo:
CurationConcerns provided the ability to differentiate between labels and text for visibility layers: https://github.com/samvera/curation_concerns/blob/master/config/locales/curation_concerns.en.yml#L145-L147 and used the term "Open Access" (which we now understand is problematic).

You can see how these are used in CC:

* https://github.com/samvera/curation_concerns/blob/master/app/helpers/curation_concerns/ability_helper.rb#L37
* https://github.com/samvera/curation_concerns/blob/b1d34b15f233da1df1ed07421fffc5a8959eb18f/app/views/curation_concerns/base/_form_permission.html.erb#L16

Sufia overrode this behavior from CC, but only overrode *part* of it (the label). But, as far as I can tell, Sufia failed to override the text because this i18n key:

* https://github.com/samvera/sufia/blob/master/config/locales/sufia.en.yml#L222

Which is used here:

* https://github.com/samvera/sufia/blob/master/app/helpers/sufia/sufia_helper_behavior.rb#L255

is overridden by an identically named key here: https://github.com/samvera/sufia/blob/master/config/locales/sufia.en.yml#L495

Since, as we have learned that the term "Open Access" is problematic because it has baggage, I propose that we change this label back to "Public".

### DIFF
--- a/config/locales/hyrax.de.yml
+++ b/config/locales/hyrax.de.yml
@@ -72,6 +72,8 @@ de:
           header: Verwaltungssatz bearbeiten
         form:
           cancel: Stornieren
+          permission_destroy_errors:
+            admin_group: Die Gruppe der Repository-Administratoren kann nicht entfernt werden
           permission_update_errors:
             error: Ungültige Update-Option für Berechtigungsvorlage.
             no_date: Für die gewählte Release-Option ist ein Datum erforderlich.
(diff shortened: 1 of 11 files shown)


vantuyls:
i support this message.

PR #1304, opened 2017-06-28
Streamlined specs

atz:
Simple improvements

### DIFF
--- a/spec/actors/hyrax/actors/generic_work_actor_spec.rb
+++ b/spec/actors/hyrax/actors/generic_work_actor_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe Hyrax::Actors::GenericWorkActor do
   let(:env) { Hyrax::Actors::Environment.new(curation_concern, ability, attributes) }
   let(:user) { create(:user) }
   let(:ability) { ::Ability.new(user) }
-  let(:admin_set) { create(:admin_set, with_permission_template: { with_active_workflow: true }) }
+  let(:admin_set) { build(:admin_set, with_permission_template: { with_active_workflow: true }) }
   # stub out redis connection
   let(:redlock_client_stub) do
     client = double('redlock client')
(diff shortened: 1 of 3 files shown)


spec/actors/hyrax/actors/generic_work_actor_spec.rb
```diff
@@ -242,6 +236,7 @@
         expect(old_parent.reload.members).to eq []
       end
     end
+
```
jcoyne: I think the breaks between context blocks leads to readability.  Do you feel strongly that  we need to compress everything?

spec/actors/hyrax/actors/generic_work_actor_spec.rb
```diff
@@ -204,11 +201,8 @@
   describe '#update' do
     let(:curation_concern) { create(:generic_work, user: user, admin_set_id: admin_set.id) }
 
-    subject { Hyrax::CurationConcern.actor }
-
     context 'failure' do
       let(:attributes) { {} }
-
```
jcoyne: Later versions of rubocop-rspec suggests having blank lines after the subject and final let block.

spec/actors/hyrax/actors/generic_work_actor_spec.rb
atz: @jcoyne I don't know if it is the jet lag or what, but this line is **added** by me, not deleted.  We're in agreement here.

jcoyne: Sorry, must be the jet lag.

spec/actors/hyrax/actors/generic_work_actor_spec.rb
atz: @jcoyne We can deal with that when we get to later versions of rubocop-rspec with one big autocorrect, if we care.  For this terse little `let`/`it` pair though, I like it together.

PR #1305, opened 2017-06-28
Update rubocop-rspec

jcoyne:
@samvera/hyrax-code-reviewers


atz:
Definitely 👍  on the `let/before/after/subject` ordering.  

A solid **meh** on the mechanical application of thousands of empty lines in places where they really don't improve readability to my eye (mostly smaller examples).

PR #1313, opened 2017-06-30
update README.md with brew install note

revgum:
adds a note for using homebrew to install fits

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ Hyrax requires the following software to work:
 
 ### Characterization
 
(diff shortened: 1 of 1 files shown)


jcoyne:
You might want to say that home brew is a solution only for OS X users

mjgiarlo:
@revgum Do you have cycles for this this week?

mjgiarlo:
@revgum @jcoyne I've pushed a patch into this branch and I will be merging it in a few.

PR #1314, opened 2017-07-05
1.0-stable encoding fix

dheles:
Backports the fix for encoding issues from 2.x to 1.0 stable

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/microdata.rb
+++ b/app/services/hyrax/microdata.rb
@@ -1,7 +1,7 @@
 module Hyrax
   class Microdata
     include Singleton
-    FILENAME = 'config/schema_org.yml'.freeze
+    FILENAME = Hyrax::Engine.root + 'config/schema_org.yml'
     TOP_KEY = 'schema_org'.freeze
 
     def self.fetch(key, options = {})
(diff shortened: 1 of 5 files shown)


jcoyne:
I'm curious as to why the second commit isn't a cherry-pick of https://github.com/samvera/hyrax/commit/e0b5359c23ec46706132e55bcffa3af25d698a13 @dheles Can you explain?

dheles:
@jcoyne no reason other than that I had seen the result (adding "Hyrax::Engine.root" to the filename path) and the need, but had not noted the commit as a discrete and cherry-pick-able one we could use here. i'm happy to make changes, if that would be preferable.

jcoyne:
@dheles I recommend we do that as it's a little easier to diff the two branches that way.

dheles:
yikes! been sleeping on this... the CLAs are in, can we get this merged, plz?

mjgiarlo:
@dheles Got it! And @hortongn's 8th backport as well.

dheles:
woot! much obliged, @mjgiarlo

Issue #1100, opened 2017-05-30
Rename 'Technical' in Admin Dashboard


geekscruff:
The label for the flip-flop feature in the admin dashboard is called 'Technical', whilst the page itself is called 'Feature'. I suggest aligning the terminology and renaming the dashboard item 'Feature'.

hannahfrost:
@ggeisler Are you cool with this proposal? I am.

ggeisler:
I agree we need some renaming here but want to confirm we're talking about the same thing. From what I see the two labels that need alignment are "Technical" and "Settings" (not "Feature"):

![index_feature____hyrax](https://user-images.githubusercontent.com/101482/27052548-d7f1c5ba-4f6e-11e7-83a0-638c2e99c597.png)

Assuming A and B in the screenshot above are the things we're trying to align, "Features" (plural) seems like a good choice?

mjgiarlo:
Yes, that's what we're talking about and :+1: to your suggestion!

hannahfrost:
👍

laritakr:
Closed via PR #1316

PR #1316, opened 2017-07-06
Rename Technical page

laritakr:
The flip-flop features page in the admin dashboard was called 'Technical', while the page itself was titled 'Settings'. Both were renamed to 'Features', the icon was changed to be consistent, and I18n capability was added for each of the headings (via google translate).

Resolves issue https://github.com/samvera/hyrax/issues/1100

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.it.yml
+++ b/config/locales/hyrax.it.yml
@@ -165,6 +165,12 @@ it:
         update:
           flash:
             success: L'aspetto è stato aggiornato con successo
+      features:
+        index:
+          header:           Caratteristiche
+          feature:          Caratteristica
+          description:      Descrizione
+          action:           Azione
       sidebar:
         activity: Attività
         admin_sets: Set amministrativi
@@ -179,7 +185,7 @@ it:
         settings: impostazioni
         statistics: Rapporti
         tasks: Compiti
-        technical: Tecnico
+        technical: Caratteristiche
         transfers: trasferimenti
         user_activity: La tua attività
         users: Gestisci utenti
--- a/config/locales/hyrax.zh.yml
+++ b/config/locales/hyrax.zh.yml
@@ -167,6 +167,12 @@ zh:
         update:
           flash:
             success: 表面已更新
+      features:
+        index:
+          header:  特征
+          feature:          特征
+          description:      描述
+          action:           行动
       sidebar:
         activity: 活动
         admin_sets: 管理集
@@ -181,7 +187,7 @@ zh:
         settings: 设置
         statistics: 统计
         tasks: 任务
-        technical: 技术
+        technical: 特征
         transfers: 转让
         user_activity: 你的操作
         users: 管理用户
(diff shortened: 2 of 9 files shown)


config/locales/hyrax.it.yml
```diff
@@ -158,13 +158,18 @@ it:
           breadcrumb: Visualizza set
           confirm_delete: Sei sicuro di voler eliminare questo set amministrativo? Questa azione non può essere annullata.
           header: Impostazione amministrativa
-          item_list_header: Funziona in questo set
```
mjgiarlo: Was this removed on purpose?

config/locales/hyrax.zh.yml
```diff
@@ -8,6 +8,7 @@ zh:
       messages:
         conflict: 您的更改无法保存，因为另一用户(或后台作业）在您开始编辑后更新了%{model}。 请确保所有文件附件已成功完成之后再试一次。这已经是最新刷新后保存的副本%{model}
   blacklight:
+    application_name: must be here for features page to open
```
mjgiarlo: Why would this only be needed in the Chinese translation?

laritakr: Haven't tracked that down yet, but it was definitely the case. I pushed it up this way and am looking into that next.

config/locales/hyrax.it.yml
laritakr: No! I wonder if I hit a bad keystroke in there?!? Good catch... I'll fix it.

config/locales/hyrax.zh.yml
laritakr: Apparently because Chinese is the only language we have in Hyrax that is not included in blacklight? https://github.com/projectblacklight/blacklight/tree/master/config/locales

mjgiarlo: Ah, good call! So in Hyku, we did this: https://github.com/samvera-labs/hyku/commit/e9430e27cd82343fcf7ed08d051b7a78629dce35

I'd ask the Blacklight folks for advice on how we should be handling this. Should we find someone to add translations directly to Blacklight, or in the meantime should Hyrax ship some Blacklight translations? @jkeck @mejackreed @cbeer @jcoyne

jcoyne: I think we should move this to https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/templates/config/locales/hyrax.zh.yml, because that file gets generated into the application, and it makes it easier for people to change it.  Long term it would be great to get a Chinese translation into Blacklight.

laritakr: Thanks @jcoyne I knew someone would be able to tell me the right way to handle this! :)

jcoyne:
@laritakr this shouldn't impact these changes, but here's the long term solution to the `application_name` problem: https://github.com/projectblacklight/blacklight/pull/1724

jcoyne:
Let's make sure to do the "Squash and merge" since we don't need this to be 2 commits.

PR #1318, opened 2017-07-06
Use retriable in UserStatImporter

jcoyne:
We don't need to roll our own retry logic.
@samvera/hyrax-code-reviewers


mjgiarlo:
#1301 removes `retriable` as a dependency, FYI!

mjgiarlo:
@jcoyne Rebased on `master` and added `retriable` back to the gemspec.

PR #1319, opened 2017-07-06
adds pagination to the notifications table

### LINKED ISSUE #328: Notifications should be paginated
currently they all show on one page. This gets to be a lot when there are several hundred.

revgum:
fixes #328

screen shot 2017-07-06 at 3 32 45 pm

### DIFF
--- a/app/views/hyrax/notifications/_notifications.html.erb
+++ b/app/views/hyrax/notifications/_notifications.html.erb
@@ -1,27 +1,33 @@
-<table class="table table-striped"> 
-  <thead> 
-    <tr>
-      <th><%= t('hyrax.mailbox.date') %></th>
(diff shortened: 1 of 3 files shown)


app/views/hyrax/notifications/_notifications.html.erb
```diff
@@ -1,27 +1,33 @@
...
-        </td>
-      </tr>
-    <% end %> 
-  </tbody> 
-</table>
+      </tbody>
+    </table>
+  </div>
+<% else %>
+  <p>No notifications found.</p>
```
mjgiarlo: Can you add this string to `config/locales/hyrax.en.yml` so we can translate this into other languages?

Issue #1296, opened 2017-06-23
New collection form should have the layout of the edit form


jcoyne:

atz:
To clarify, you're saying that `Hyrax::CollectionsController` should be updated?  It currently does:
```ruby
layout :decide_layout
...
      def decide_layout
        case action_name
        when 'show'
          theme
        else
          'dashboard'
        end
      end
```

So AFAICT, that means that the layout IS the same for `create` as for `edit` (namely 'dashboard').  What am I missing?

mjgiarlo:
I assume what @jcoyne means is the difference between https://github.com/samvera/hyrax/blob/master/app/views/hyrax/collections/new.html.erb and https://github.com/samvera/hyrax/blob/master/app/views/hyrax/collections/edit.html.erb. So the actual layout, not the Rails layout. Terminology is so overloaded.

atz:
Ok, the view/template.

PR #1326, opened 2017-07-10
Make collections add & edit consistent

laritakr:
1. Add breadcrumbs which were missing
2. Handle i18n where missing on edit and new pages

Resolves https://github.com/samvera/hyrax/issues/1296

Note: the issue implied that the form being rendered was not the same, but in actuality it was... the only difference was that new rendered form without qualifying the path, while edit qualified the path to the form. For consistency purposes, I changed this in new.html.erb, but this does not affect behavior.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/collections_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/collections_controller_behavior.rb
@@ -59,6 +59,9 @@ module Hyrax
     end
 
     def new
+      add_breadcrumb t(:'hyrax.controls.home'), root_path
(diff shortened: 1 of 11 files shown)


mjgiarlo:
@laritakr looks like you have a related feature spec failing.

laritakr:
@mjgiarlo Yes, I accidentally changed the title slightly differently between the new page and the feature spec. Was fixing it when you reviewed. Should be good now but will require re-approval. Thanks.

Issue #167, opened 2016-12-21
Administrator group should be able to modify all content


mjgiarlo:
<a href="https://github.com/vantuyls"><img src="https://avatars.githubusercontent.com/u/9534576?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [vantuyls](https://github.com/vantuyls)**
_Thursday Sep 01, 2016 at 22:10 GMT_
_Originally opened as https://github.com/projecthydra/sufia/issues/2614_

----

Administrator can: 
- [ ] search and view all items (including private items not owned by me), 
- [ ] edit all items (including items I do not own), 
- [ ] delete all items, 
- [ ] transfer ownership of all items (including items I do not own), 
- [ ] share all items (including items I do not own), 
- [ ] move items to and from all collections and admin sets

Decomposes #2421

mjgiarlo:
<a href="https://github.com/mjgiarlo"><img src="https://avatars.githubusercontent.com/u/131982?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mjgiarlo](https://github.com/mjgiarlo)**
_Friday Oct 28, 2016 at 23:09 GMT_

----

Out of scope for current mediated deposit sprints.

mjgiarlo:
@hannahfrost @vantuyls @scholarworks How high a priority is this issue?

hannahfrost:
@mjgiarlo I think it is high priority.

jcoyne:
Has anyone done analysis on this to determine which of these already work and which do not?

jcoyne:
What does this mean? 
> move items to and from all communities and sets

jcoyne:
What does "ownership" mean?  Hyrax doesn't model ownership, only people with edit access.

vantuyls:
@jcoyne i fixed that "communities and sets" stuff. but, basically, if an items is in the "wrong" admin set or collection, an admin can move them.

As for "ownership", i'm referring to who has the object as "My Work" - see attachment:

![screen shot 2017-02-02 at 12 22 42 pm](https://cloud.githubusercontent.com/assets/9534576/22566851/7114ae62-e942-11e6-8a02-62ccbf8e023f.png)

jcoyne:
[...]

PR #1331, opened 2017-07-10
Default Admin Set not getting all access grants and workflow responsibilities created

mjgiarlo:
This brings the Default Admin Set in line with other Admin Sets. In particular, works deposited by other users to the Default Admin Set with mediation turned on would not appear in the Review Submissions view for repository administrators.

Refs #167

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/admin_set_create_service.rb
+++ b/app/services/hyrax/admin_set_create_service.rb
@@ -70,11 +70,14 @@ module Hyrax
     private
 
       def access_grants_attributes
-        return [] unless creating_user
         [
-          { agent_type: 'user', agent_id: creating_user.user_key, access: Hyrax::PermissionTemplateAccess::MANAGE },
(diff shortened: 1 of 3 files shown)


app/services/hyrax/admin_set_create_service.rb
```diff
@@ -91,28 +94,37 @@ def create_workflows_for(permission_template:)
...
-                                               creating_user,
-                                               Hyrax::Group.new(admin_group_name)
-                                             ])
+                                             agents: workflow_agents)
           end
         end
       end
 
-      # Gives deposit access to registered users and manage access to admins to default AdminSet
+      # Default admin set has a nil creating_user; guard against that condition
```
jcoyne: The guard clause was on line 96. I don't think it makes sense to move this comment here.

mjgiarlo: Changed.

mjgiarlo:
@jcoyne Addressed the issue you raised in code review re: an inaccurate comment.

PR #1368, opened 2017-07-19
Lock down dependencies in gemspec

dheles:
In response to output from running:
  gem build hyrax.gemspec

https://gist.github.com/dheles/910a6e0ea8f423b8a65dc3f8e6aeb16f

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -25,34 +25,34 @@ EOF
   spec.version       = Hyrax::VERSION
   spec.license       = 'Apache-2.0'
 
-  spec.add_dependency 'hydra-head', '>= 10.4.0'
+  spec.add_dependency 'hydra-head', '~> 10.4', '>= 10.4.0'
   spec.add_dependency 'hydra-editor', '~> 3.2'
   spec.add_dependency 'hydra-works', '~> 0.16'
   spec.add_dependency 'hydra-derivatives', '~> 3.3'
-  spec.add_dependency 'browse-everything', '>= 0.10.5'
+  spec.add_dependency 'browse-everything', '~> 0.10', '>= 0.10.5'
   spec.add_dependency 'blacklight', '~> 6.9'
   spec.add_dependency 'blacklight-gallery', '~> 0.7'
   spec.add_dependency 'tinymce-rails', '~> 4.1'
   spec.add_dependency 'tinymce-rails-imageupload', '~> 4.0.17.beta'
   spec.add_dependency 'font-awesome-rails', '~> 4.2'
-  spec.add_dependency 'select2-rails', '~> 3.5.9'
-  spec.add_dependency 'json-schema' # for Arkivio
+  spec.add_dependency 'select2-rails', '~> 3.5', '~> 3.5.9'
+  spec.add_dependency 'json-schema', '~> 0' # for Arkivio
   spec.add_dependency 'nest', '~> 2.0'
   spec.add_dependency 'mailboxer', '~> 0.12'
   spec.add_dependency 'carrierwave', '~> 1.0'
-  spec.add_dependency 'oauth'
+  spec.add_dependency 'oauth', '~> 0'
   spec.add_dependency 'oauth2', '~> 1.2'
-  spec.add_dependency 'signet'
+  spec.add_dependency 'signet', '~> 0'
   spec.add_dependency 'legato', '~> 0.3'
-  spec.add_dependency 'posix-spawn'
+  spec.add_dependency 'posix-spawn', '~> 0'
   spec.add_dependency 'jquery-ui-rails', '~> 5.0'
-  spec.add_dependency 'redis-namespace', '~> 1.5.2'
+  spec.add_dependency 'redis-namespace', '~> 1.5', '~> 1.5.2'
   spec.add_dependency 'flot-rails', '~> 0.0.6'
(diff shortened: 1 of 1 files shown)


hyrax.gemspec
```diff
@@ -84,10 +84,10 @@ EOF
   spec.add_development_dependency "equivalent-xml", '~> 0.5'
   spec.add_development_dependency "jasmine", '~> 2.3'
   spec.add_development_dependency 'rubocop', '~> 0.47.0'
-  spec.add_development_dependency 'rubocop-rspec', '~> 1.10.0'
+  spec.add_development_dependency 'rubocop-rspec', '~> 1.10', '~> 1.10.0'
```
mjgiarlo: What is the effect of having two twiddle-wakkas on one line? I believe we want rubocop and rubocop-rspec pinned such that only patch-level updates come in.

hyrax.gemspec
```diff
@@ -25,34 +25,34 @@ EOF
...
-  spec.add_dependency 'browse-everything', '>= 0.10.5'
+  spec.add_dependency 'browse-everything', '~> 0.10', '>= 0.10.5'
   spec.add_dependency 'blacklight', '~> 6.9'
   spec.add_dependency 'blacklight-gallery', '~> 0.7'
   spec.add_dependency 'tinymce-rails', '~> 4.1'
   spec.add_dependency 'tinymce-rails-imageupload', '~> 4.0.17.beta'
   spec.add_dependency 'font-awesome-rails', '~> 4.2'
-  spec.add_dependency 'select2-rails', '~> 3.5.9'
-  spec.add_dependency 'json-schema' # for Arkivio
+  spec.add_dependency 'select2-rails', '~> 3.5', '~> 3.5.9'
```
mjgiarlo: Should the twiddle-wakka you added be `>=` instead?

hyrax.gemspec
```diff
@@ -25,34 +25,34 @@ EOF
...
+  spec.add_dependency 'oauth', '~> 0'
   spec.add_dependency 'oauth2', '~> 1.2'
-  spec.add_dependency 'signet'
+  spec.add_dependency 'signet', '~> 0'
   spec.add_dependency 'legato', '~> 0.3'
-  spec.add_dependency 'posix-spawn'
+  spec.add_dependency 'posix-spawn', '~> 0'
   spec.add_dependency 'jquery-ui-rails', '~> 5.0'
-  spec.add_dependency 'redis-namespace', '~> 1.5.2'
+  spec.add_dependency 'redis-namespace', '~> 1.5', '~> 1.5.2'
```
mjgiarlo: Should the twiddle-wakka you added be `>=` instead?

hyrax.gemspec
```diff
@@ -25,34 +25,34 @@ EOF
...
   spec.add_dependency 'flot-rails', '~> 0.0.6'
   spec.add_dependency 'almond-rails', '~> 0.1'
   spec.add_dependency 'qa', '~> 1.0' # questioning_authority
   spec.add_dependency 'flipflop', '~> 2.3'
-  spec.add_dependency 'jquery-datatables-rails', '~> 3.4.0'
-  spec.add_dependency 'rdf-rdfxml' # controlled vocabulary importer
+  spec.add_dependency 'jquery-datatables-rails', '~> 3.4', '~> 3.4.0'
```
mjgiarlo: Should the twiddle-wakka you added be `>=` instead?

dheles:
my bad. uncareful editing, not rare ruby-fu

mjgiarlo:
@dheles Rubocop balked. You should set the rubocop-rspec line back to `'~> 1.10.0'` The pessimistic dependency there is intentional. (If you want to toss a comment around that line to make that explicit, go for it!) You may want to (though it's up to you) run rubocop and rspec locally to save some time, so you know if your branch is merge-able. I've found it helpful to create a file in my Hyrax workspace called `.git/hooks/pre-push` that looks like the following:

```shell
#!/bin/sh
bundle exec rubocop
```

That prevents me from pushing code to Hyrax that fails a rubocop check.

dheles:
wilco

hyrax.gemspec
```diff
@@ -25,34 +25,34 @@ EOF
...
+  spec.add_dependency 'browse-everything', '~> 0.10', '>= 0.10.5'
   spec.add_dependency 'blacklight', '~> 6.9'
   spec.add_dependency 'blacklight-gallery', '~> 0.7'
   spec.add_dependency 'tinymce-rails', '~> 4.1'
   spec.add_dependency 'tinymce-rails-imageupload', '~> 4.0.17.beta'
   spec.add_dependency 'font-awesome-rails', '~> 4.2'
-  spec.add_dependency 'select2-rails', '~> 3.5.9'
-  spec.add_dependency 'json-schema' # for Arkivio
+  spec.add_dependency 'select2-rails', '~> 3.5', '>= 3.5.9'
+  spec.add_dependency 'json-schema', '~> 0' # for Arkivio
```
mjgiarlo: This line appears to cause bundler to pick up version 0.9.12 of json-schema. I'm seeing the most recent version (2.8.0) used before this change. That's why Travis is failing, methinks. This makes me wonder if the other `~> 0` declarations in the gemspec are potentially problematic as well, and I'm not sure how `~> 0` is any clearer or more tightly specified than no version at all.

PR #1373, opened 2017-07-21
AWSFile/S3 integrations

atz:
- Use shared_examples in test
- Cut out unnecessary recourse to `ImportUrlJob` for Carrierwave-AWS (S3) objects since the OO API still does what we want
- Validate all files in job *before* beginning, to avoid dying mid-job.
- Failover for `filename` of uncached remote file, since that is `nil` until fetched (and we don't want to fetch just to learn the filename)

See https://github.com/samvera-labs/hyku/issues/1371

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -143,8 +143,8 @@ module Hyrax
         # If all else fails, use the basename of the file where it sits.
         # @note This is only useful for labeling the file_set, because of the recourse to import_url
         def label_for(file)
-          if file.is_a?(Hyrax::UploadedFile)
-            file.uploader.filename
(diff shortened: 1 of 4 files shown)


atz:
@mjgiarlo This should be merged as is currently.  It doesn't fix "the problem", it fixes a discrete portion of the problem.

mjgiarlo:
@atz the one concern I have about merging this PR is that it eliminates the code path to calling `import_url`, which I believe will break BrowseEverything integration:

mjgiarlo:
I have confirmed that BrowseEverything integration is already broken in `master`. I was able to connect with my Dropbox account, select four files, and populate the "Files" section of the New Work form. So far, so good. Submitting the form results in the right number of FileSets underneath the work, but zero of these files are downloadable, none of them have thumbnails, and only 1 of the 4 appears to have been characterized. I believe these bugs are already ticketed. I am approving this PR and we can come back to fixing BE integration.

PR #1384, opened 2017-07-25
Handle resource_type # => nil case in SolrDocumentBehavior#itemtype

no-reply:
`SolrDocumentBehavior#itemtype` tries to access `resource_type`. By default, this is fine, but `resource_type` is allowed to return `nil` in client apps. This handles this case by using an empty array as the default resource_type value within the method.

`ResourceTypesService#microdata_type` handles the `nil` case, so `[].first` is an acceptable parameter. The `#microdata_type` behavior is tested, and this PR adds documentation making it more clear to clients.

* Make `SolrDocumentBehavior#itemtype` more resilient to return values of the (non-core) `resource_type`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/solr_document_behavior.rb
+++ b/app/models/concerns/hyrax/solr_document_behavior.rb
@@ -11,7 +11,8 @@ module Hyrax
 
     # Add a schema.org itemtype
     def itemtype
-      ResourceTypesService.microdata_type(resource_type.first)
+      types = resource_type || []
+      ResourceTypesService.microdata_type(types.first)
     end
 
     def title_or_label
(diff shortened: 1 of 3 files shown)


app/models/concerns/hyrax/solr_document_behavior.rb
```diff
@@ -11,7 +11,8 @@ module SolrDocumentBehavior
 
     # Add a schema.org itemtype
     def itemtype
-      ResourceTypesService.microdata_type(resource_type.first)
+      types = resource_type || []
+      ResourceTypesService.microdata_type(types.first)
```
jcoyne: Please add a test for this behavior change.

no-reply: Incoming. I missed that these methods are tested in the `SolrDocument` context. I was looking for explicit tests of the Behavior mixin.

no-reply: The test is in; I squashed and force pushed.

PR #1387, opened 2017-07-25
Update abstract_actor.rb docs

jcoyne:


app/actors/hyrax/actors/abstract_actor.rb
```diff
@@ -1,18 +1,15 @@
...
     #
-    # * next_actor
-    # * curation_concern
-    # * user
     #
-    # it must instantiate the next actor in the chain and instantiate it.
-    # it should respond to curation_concern, user and attributes.
-    # it ha to next_actor
+    # In order to continue the stack it must instantiate the next actor in the chain and call it
+    #   OR to exit from the stack, return a truthy or falsey value.
```
no-reply: Is it acceptable to return truthy but *not* call the `next_actor`? What does a `truthy` or `fasley` return value indicate, if anything?

mjgiarlo: @jcoyne I'm inclined to get this PR merged soon, but I see there's an outstanding question. Have time to clarify this in the documentation?

PR #1393, opened 2017-07-25
Update Actor docs to reflect current requirements for Actors

no-reply:
Extends @jcoyne's work on #1387. I believe the text is basically correct, but would appreciate serious review for both accuracy and clarity.

@samvera/hyrax-code-reviewers


jcoyne:
Great stuff!

mjgiarlo:
Thx, @no-reply!

PR #1396, opened 2017-07-26
Workflow::PermissionQuery extract method refactor

### LINKED ISSUE #936: Notifications for reviewer role going beyond the admin set
### Descriptive summary

Anyone with reviewer role on an admin set is getting the notifications for works submitted on ALL admin sets.
This issue exists in both the 1-0 stable and the master branch.

### Rationale

Users with the approving role should only get useful notifications. 

### Expected behavior

Anyone with the workflow role of reviewer on an admin set should be notified of works submitted to that admin set which they should review.

### Actual behavior

Anyone with the workflow role of reviewer on any admin set are getting notifications of works submitted to all admin sets.

I created two separate admin sets and 6 users with varying participant roles and workflow roles, and submitted works from each user who was able to submit to each admin set. Reviewers for only admin set 1 received notifications for works submitted to both admin sets. 

[Spreadsheet of users, roles, and what they were able to do in the app.](https://docs.google.com/spreadsheets/d/1zQsWlYMVs7gz6dHaOiXICygLj_5GTFOmbpI9MHRAq5o/edit#gid=0)

darrenleeweber:
Refs #936

- `Hyrax::Workflow::PermissionQuery`
  - extract method for Sipity::EntitySpecificResponsibility.arel_table
  - extract method for Sipity::WorkflowResponsibility.arel_table
  - extract method for Sipity::WorkflowRole.arel_table

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/workflow/permission_query.rb
+++ b/app/services/hyrax/workflow/permission_query.rb
@@ -20,6 +20,10 @@ module Hyrax
     module PermissionQuery
       module_function
 
+      def entity_responsibilities
+        @entity_responsibilities ||= Sipity::EntitySpecificResponsibility.arel_table
+      end
+
       # @api public
       #
       # For the given :user and :entity return only workflow actions that meet all of the following:
@@ -58,7 +62,6 @@ module Hyrax
         role = PowerConverter.convert_to_sipity_role(role)
         workflow_roles = Sipity::WorkflowRole.arel_table
         workflow_responsibilities = Sipity::WorkflowResponsibility.arel_table
-        entity_responsibilities = Sipity::EntitySpecificResponsibility.arel_table
 
         agents = Sipity::Agent.arel_table
 
@@ -184,7 +187,6 @@ module Hyrax
         workflow_states = Sipity::WorkflowState.arel_table
         workflow_state_action_permissions = Sipity::WorkflowStateActionPermission.arel_table
         workflow_responsibilities = Sipity::WorkflowResponsibility.arel_table
-        entity_responsibilities = Sipity::EntitySpecificResponsibility.arel_table
 
         user_agent_scope = scope_processing_agents_for(user: user)
         user_agent_contraints = user_agent_scope.arel_table.project(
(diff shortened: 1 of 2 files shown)


darrenleeweber:
The AR method seems to return the same instance every time.  For example, if we run the specs with an extra `puts` in the method, i.e.
```
      def entity_responsibilities
        puts Sipity::EntitySpecificResponsibility.arel_table.inspect
        @entity_responsibilities ||= Sipity::EntitySpecificResponsibility.arel_table
      end
```

Then we get the same instance-ID for every call of that method.  As a _guess_, this instance is created by AR when rails is initialized.

mjgiarlo:
@darrenleeweber I'm reluctant to approve this PR given the complexity in this module without feedback from either @jeremyf or @jcoyne. I'm not specifically opposed to this refactoring, though.

jcoyne:
Why are these changes necessary? Are they fixing some bug or in support of some feature?

JeremyFriesen:
Might want to make sure you're sending these messages to the correct Jeremy
Friesen, as the one you've tagged here is definitely the wrong one.  Would
you mind removing me from your conversation please?

Cheers and good luck with your project!
Jeremy Friesen (also an Application Developer, just the wrong one... what a
coincidence!)

mjgiarlo:
@JeremyFriesen Yes, I corrected the reference ten seconds after autocomplete did the wrong thing. Sorry about that. You shouldn't see any more of these (after this one). Happy coding to you, too. :)

mjgiarlo:
@darrenleeweber In the meantime, can you change the description of this PR to read `Refs #936` instead of `Fixes #936`?

darrenleeweber:
@mjgiarlo - changed Fix to Ref on the issue; removed the WIP status on this PR to isolate these method refactors before moving forward with further investigation and issue resolution.

@jcoyne - this refactor is aimed to DRY the methods in this class, that's all.  I did it quickly and simply as part of getting familiar with this class.  It's in this PR because I want to keep the review for this simple without confusing WIP on the issue resolution (which is far more complex).

PR #1399, opened 2017-07-31
There is no need to declare indexer as it's part of ActiveFedora

jcoyne:


hackmastera:
The change to the test seems to indicate that this is moving the method from instance to class. Should this commit be called in in some way in upgrade docs?

jcoyne:
@hackmastera It's no longer an instance method. I'm not sure that anyone would be using that though as it's called by ActiveFedora itself. Were you?

hackmastera:
@jcoyne I don't think so. Thanks for the explanation!

PR #1412, opened 2017-08-02
AdminSet - include GlobalID::Identification

### LINKED ISSUE #1411: Sipity entities require global-ID but AdminSet does not have one
### Descriptive summary

The Sipity workflow permissions use an entity as a proxy for works etc. and, in doing so, they keep a reference to the global-id for the object.  To use AdminSet with Sipity, it needs a global-id.

### Related work

This is related to investigation work on https://github.com/samvera/hyrax/issues/936

darrenleeweber:
Fixes #1411 

This adds a global-id to the AdminSet (as suggested by @mjgiarlo in slack commentary on Aug 1, 2017)

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/admin_set.rb
+++ b/app/models/admin_set.rb
@@ -16,6 +16,7 @@
 # @see Hyrax::DefaultAdminSetActor
 # @see Hyrax::ApplyPermissionTemplateActor
 class AdminSet < ActiveFedora::Base
+  include GlobalID::Identification
   include Hydra::AccessControls::WithAccessRight
   include Hyrax::Noid
   include Hyrax::HumanReadableType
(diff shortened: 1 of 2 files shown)


atz:
Even if Sipity doesn't require it, GlobalID might still be useful for AdminSets.  I don't see the reason to block this.

mjgiarlo:
After pairing with @darrenleeweber yesterday, I believe we discovered that this change was not necessary given the tight and reliable relationship between AdminSets and PermissionTemplates. OK to close this for now, @darrenleeweber? I'll catch up with your latest work on this soon, too. Thanks!

darrenleeweber:
@mjgiarlo - in a console session, with this commit available, the `AdminSet` was still not compatible with a `Sipity::Entity`, so this PR doesn't serve that purpose (although it may still be a good idea to enable that somehow).  That said, @atz has made a point that this PR could serve additional purpose(s) and I'd like to now defer a decision about that, pending commentary/debate about that purpose.  In any case, my focus will shift elsewhere.

PR #1413, opened 2017-08-02
AWS S3 file ingest abstraction/testing

atz:
This is necessary to support configuration in Hyku w/ carrierwave-aws.


atz:
Failing test is unrelated in `spec/features/embargo_spec.rb` (passes locally, passes in the other half of the PR).  Recommend merge.

PR #1414, opened 2017-08-02
consolidating requires into app manifest

csyversen:
We wanted to override some stuff in `uploader.js` but were running into issues with dependencies in our child app. I figured our dependencies should all be in the application manifest anyway (plus it allows us to override `uploader.js`successfully now) 

I didn't make this change for the master branch of hyrax since it looks like maxFileSize is going to be configurable in an initializer, but i can consolidate js dependencies in that branch too if you want.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax.js
+++ b/app/assets/javascripts/hyrax.js
@@ -28,6 +28,14 @@
 //= require morris/raphael-min
 //= require morris/morris.min
 
+// Moved from app/assets/javascripts/hyrax/uploader.js
+//= require fileupload/tmpl
+//= require fileupload/jquery.iframe-transport
+//= require fileupload/jquery.fileupload.js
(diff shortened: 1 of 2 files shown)


atz:
Seems like the change would make more sense on master, then *maybe* backported to stable.  Starting a change on a stable branch is anti-pattern.

csyversen:
@atz sounds good, i'll create another PR today for the master branch. 

We're running a hyrax 1.x application right now with users who need to upload larger files, so it'd be nice to have this change backported.

mjgiarlo:
@csyversen Once we get this merged to `master`, it should be straightforward to cherry-pick the relevant commit to a new branch off of `1-0-stable` for a new PR. Thanks for your patience!

PR #1415, opened 2017-08-02
Workflow notifications for pending reviews

### LINKED ISSUE #936: Notifications for reviewer role going beyond the admin set

darrenleeweber:
Fixes #936 

This ensures that notifications to review/approve a new submission that belongs to an AdminSet are sent only to the managers associated with the AdminSet (cc to the depositor).

* Use the entity workflow to query the AdminSet/PermissionTemplate associated with that workflow

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/workflow/notification_service.rb
+++ b/app/services/hyrax/workflow/notification_service.rb
@@ -48,10 +48,18 @@ module Hyrax
       # @return [Hash<String, Array>] a hash with keys being the strategy (e.g. "to", "cc") and
       #                               the values are a list of users.
       def recipients(notification)
-        notification.recipients.each_with_object({}) do |r, h|
-          h[r.recipient_strategy] ||= []
-          h[r.recipient_strategy] += PermissionQuery.scope_users_for_entity_and_roles(entity: entity,
-                                                                                      roles: r.role)
+        case entity.workflow_state.name
+        when 'pending_review'
+          # notify the managers to review the deposit (CC depositors)
+          {
+            to: entity.workflow.permission_template.agent_ids_for(access: 'manage',  agent_type: 'user'),
+            cc: entity.workflow.permission_template.agent_ids_for(access: 'deposit', agent_type: 'user')
+          }
+        else
+          notification.recipients.each_with_object({}) do |r, h|
+            h[r.recipient_strategy] ||= []
+            h[r.recipient_strategy] += PermissionQuery.scope_users_for_entity_and_roles(entity: entity, roles: r.role)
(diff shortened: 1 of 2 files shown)


app/services/hyrax/workflow/notification_service.rb
```diff
@@ -48,10 +48,18 @@ def send_notification(notification)
...
-          h[r.recipient_strategy] ||= []
-          h[r.recipient_strategy] += PermissionQuery.scope_users_for_entity_and_roles(entity: entity,
-                                                                                      roles: r.role)
+        case entity.workflow_state.name
+        when 'pending_review'
+          # notify the managers to review the deposit (CC depositors)
+          {
+            to: entity.workflow.permission_template.agent_ids_for(access: 'manage',  agent_type: 'user'),
+            cc: entity.workflow.permission_template.agent_ids_for(access: 'deposit', agent_type: 'user')
+          }
```
darrenleeweber: Not sure if this Hash return value supports indifferent access or if the keys must be strings.  If the latter, this PR needs to be updated to use strings.

darrenleeweber: Also not sure if the notification recipients should include groups.  I've left out groups because there are no email addresses associated with a Hyrax::Group (AFAICT).

jcoyne: I think this is fine for now. As far as I know @laritakr is still working on the group abstraction.

laritakr: @jcoyne not at the moment (I'm on the collections sprints), but we consider it to be part of the roles & permissions work.

PR #1420, opened 2017-08-03
Move #present_terms to the CollectionsHelper

jcoyne:
This method is only used in the collections show action.


cjcolvar:
@mjgiarlo I restarted the failed travis build and it is 💚 .

mjgiarlo:
Thx, @cjcolvar!

PR #1424, opened 2017-08-03
Remove incorrect TODO comment

mjgiarlo:
This is a documented service that folks can use in their own rake tasks, etc.

### DIFF
--- a/app/services/hyrax/repository_fixity_check_service.rb
+++ b/app/services/hyrax/repository_fixity_check_service.rb
@@ -1,6 +1,5 @@
 module Hyrax
   class RepositoryFixityCheckService
-    # TODO: this never seems to be used. Remove?
     def self.fixity_check_everything
       ::FileSet.find_each do |gf|
         Hyrax::FileSetFixityCheckService.new(gf).fixity_check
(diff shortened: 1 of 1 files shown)


app/services/hyrax/repository_fixity_check_service.rb
```diff
@@ -1,6 +1,5 @@
 module Hyrax
   class RepositoryFixityCheckService
-    # TODO: this never seems to be used. Remove?
     def self.fixity_check_everything
```
jcoyne: Can you add a comment here: `This is service that folks can use in their own rake tasks, etc.`  otherwise, it's not really "documented" 😉

mjgiarlo: There is this: https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#fixity-checking Good enough?

jcoyne: I think that public methods should have documentation.  Want to ask the WG 😉 ?

mjgiarlo: Done. Incoming.

Issue #1401, opened 2017-07-31
Translation errors on Works listing


laritakr:
### Descriptive summary

On the works list, there are some missing translations on both tabs (/dashboard/my/works and /dashboard/works).

In addition, attempting to view it in Chinese results in a Blacklight exception.

### Actual behavior

Several items are not translated. Others work differently between languages. The page crashes with a blacklight error for Chinese.

### Steps to reproduce the behavior

1. Select "Works" from sidebar.
2. View each tab, and select different languages. 

I found the following errors on these two tabs:

- [x] Label: "Status" is not translated in the "All Works" tab in all languages (ok for Spanish).
==> t("blacklight.search.fields.facet.suppressed_bsi") is used appropriately... google translation of Status is apparently status in several languages. 
- [x] Label: "Highlighted" is not translated in the "My Works" tab.
==> Added lookup, and translation to each yml file.
- [x] The Visibility option only has the badge background in English and Spanish. In other languages it has white text and no background.
==> The badge class was being found via a lookup in the translation yml files, and the class name had at some point gotten translated to the various languages, so it was no longer a valid class. Moved the class names out of the yml and into a hash lookup.
- [x] The Visibility option throws a blacklight error in Chinese.
==> added required Blacklight translations for Chinese only into Hyrax, since blacklight has no chinese translation (I believe that application_name could be removed if the updated Blacklight version were released)
- [x] Action: "Highlight work on profile" has no translation.
==> Added lookup, and translation to each yml file.

PR #1431, opened 2017-08-04
Fix various translation errors on Works list

laritakr:
## 2. Move class names out of I18n files. 
They were getting translated into various languages, causing them to not work.

## 3. Add missing Chinese Blacklight translations. 
Blacklight does not include Chinese translation, so any terms used in Hyrax need to be included in the Hyrax I18n files until Blacklight gets updated.

Fixes https://github.com/samvera/hyrax/issues/1401

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/permission_badge.rb
+++ b/app/presenters/hyrax/permission_badge.rb
@@ -2,6 +2,14 @@ module Hyrax
   class PermissionBadge
     include ActionView::Helpers::TagHelper
 
+    VISIBILITY_LABEL_CLASS = {
+      authenticated: "label-info",
+      embargo: "label-warning",
+      lease: "label-warning",
+      open: "label-success",
+      restricted: "label-danger"
+    }.freeze
+
     # @param visibility [String] the current visibility
     def initialize(visibility)
       @visibility = visibility
@@ -15,7 +23,7 @@ module Hyrax
     private
 
       def dom_label_class
-        I18n.t("hyrax.visibility.#{@visibility}.class")
+        VISIBILITY_LABEL_CLASS.fetch(@visibility.to_sym)
       end
 
       def text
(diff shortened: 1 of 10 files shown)


app/presenters/hyrax/permission_badge.rb
```diff
@@ -15,7 +23,7 @@ def render
     private
 
       def dom_label_class
-        I18n.t("hyrax.visibility.#{@visibility}.class")
```
mjgiarlo: :+1: 

Gosh, why did we used to do this?

app/views/hyrax/my/_work_action_menu.html.erb
```diff
@@ -27,7 +27,7 @@
 
     <li role="menuitem" tabindex="-1">
       <%= display_trophy_link(current_user, document.id) do |text| %>
-        <i class='glyphicon glyphicon-star'></i> <%= text %>
+        <i class='glyphicon glyphicon-star'></i> <%= t("hyrax.dashboard.my.action.highlight") %>
```
mjgiarlo: Hmm, in this case we throw away the value of `text`, which I *think* we care about? See here:

https://github.com/samvera/hyrax/blob/8cd726b1d16fa1b85eec02d2fbdaa2f71c31178a/app/helpers/hyrax/trophy_helper.rb#L9-L10

We can return one of two strings in this block. Maybe the translations should be added to the above helper instead of here, and then we can include a translation for the `"Unhighlight Work"` case?

laritakr: Ah good catch! I missed that it switched according to context... easy fix. Thanks!

app/presenters/hyrax/permission_badge.rb
laritakr: Haha I don't know... looking up by visibility is right... but not via I18n!

laritakr:
@mjgiarlo This is ready to be re-reviewed.

app/presenters/hyrax/permission_badge.rb
atz: @mjgiarlo I think whatever changes you requested are changed

PR #1440, opened 2017-08-08
Refactoring FileSetActor

jeremyf:
* Moving logic to create JobIoWrapper into a factory-like method in the
  JobIoWrapper class. The clue for this refactor was that there were
  several specs for a private method. Also, the private method in
  FileSetactor had intimate knowledge of files and the attributes of
  JobIoWrapper.
* Marking #import_url as deprecated, as it does not appear to be called
  in any tests in Hyrax.
* Adding a class attribute for .file_actor_class; I believe this creates
  a clearer definition of external collaborators.

The original impetus was to remove a Rubocop violation regarding class
length.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -22,7 +22,7 @@ module Hyrax
         file_set.label ||= label_for(file)
         file_set.title = [file_set.label] if file_set.title.blank?
         return false unless file_set.save # Need to save to get an id
-        IngestJob.perform_later(wrapper!(file, relation))
+        IngestJob.perform_later(wrapper!(file: file, relation: relation))
       end
 
       # Spawns asynchronous IngestJob with user notification afterward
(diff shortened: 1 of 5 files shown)


app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -22,25 +22,27 @@ def create_content(file, relation = :original_file)
...
       # @return [IngestJob] the queued job
       def update_content(file, relation = :original_file)
-        IngestJob.perform_later(wrapper!(file, relation), notification: true)
+        IngestJob.perform_later(wrapper!(file: file, relation: relation), notification: true)
       end
 
       # Spawns async ImportUrlJob to attach remote file to fileset
       # @param [#to_s] url
       # @return [IngestUrlJob] the queued job
+      # @todo Remove as it appears to be untested and not called
```
jcoyne: Is this used by BrowseEverything?

jcoyne: Looks like it was introduced here https://github.com/samvera/sufia/commit/26eed26343bf41a321bef125c4a8f39106f78bd8 for S3/Fog imports.  So I think you're right that this is no longer necessary.

PR #1442, opened 2017-08-08
Refactoring to more compact methods

jeremyf:
## Refactoring to more compact methods

@90eab8d604c981f839e3fe3028b94602c36a9b9d

Rubocop has a 14 line method length enforcer. These two files exceeded
that method length. By refactoring the methods they are below the 14
line method length.

In the case of the `batch_uploads_controller.rb` the refactoring
introduces a seam in the code that allows Hyrax implementors to more
easily change the response behavior when batches are disabled.

I believe we should aspire to a 10 line method length as the general
rule, with exceptions for data mapping of attributes and other similar
methods.

## Reducing Rubocop allowed method length to 13

@9f2019363a679d347c7cfcd842ea234cc2e16439

Following the advice of Sandi Metz, aiming for fewer lines per method.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -47,7 +47,7 @@ Metrics/BlockLength:
     - 'spec/**/*.rb'
 
 Metrics/MethodLength:
-  Max: 14
+  Max: 13
 
 Style/AsciiComments:
   Enabled: false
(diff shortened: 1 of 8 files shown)


jcoyne:
@jeremyf did you want to change the Rubocop config in this PR too?

jeremyf:
Not yet. I'm just going hunting for possible refactorings.

PR #1449, opened 2017-08-08
consolidating requires into app manifest

csyversen:
refs #1414 

Let's put this change into `master` first!

We wanted to override some stuff in uploader.js but were running into issues with dependencies in our child app. I figured our dependencies should all be in the application manifest anyway (plus it allows us to override uploader.jssuccessfully now)

I didn't make this change for the master branch of hyrax since it looks like maxFileSize is going to be configurable in an initializer, but i can consolidate js dependencies in that branch too if you want.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax.js
+++ b/app/assets/javascripts/hyrax.js
@@ -37,6 +37,14 @@
 //= require webcomponentsjs/0.5.4/CustomElements.min
 //= require time-elements
 
+// Moved from app/assets/javascripts/hyrax/uploader.js
+//= require fileupload/tmpl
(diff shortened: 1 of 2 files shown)


jcoyne:
I rather like it the way it was before.  That shows what files require those dependencies.  If we want to replace the uploader, we currently know that all those dependencies can be removed.  If you put them all in application.js, we would loose that.

Why not back-port the changes you need? Where the values are set in the initializer instead?

csyversen:
cool, that works. i'll close this

PR #1453, opened 2017-08-09
Work form should display selected collections based on params

mjgiarlo:
Fixes samvera-labs/hyku#1374

Users should be able to browse to a collection they created, click "Add Works," and create a new work that belongs to the collection. The current implementation uses a param `add_files_to_collection` (renamed to `add_works_to_collection` in this PR) that is passed to the `/dashboard/my/works` view but is lost when the select work type modal pops up and submits, creating the new work form.

The work on this PR passes on the `add_works_to_collection` param value to the "my" controller. To accomplish this, it adds a new method to the work form that selects one or more collections (based on the IDs passed in params) automatically in the Relationships tab of the rendered form. (Note that this does allow arbitrary collection IDs to be passed in, but the `CollectionOptionsPresenter` ensures the user only sees collections to which they have edit access so this should not introduce security or authorization issues.)

The implementation in this PR is working for both single works and batches of works.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/my_controller.rb
+++ b/app/controllers/hyrax/my_controller.rb
@@ -70,7 +70,7 @@ module Hyrax
         @result_set_size = @response.response["numFound"]
         @empty_batch = batch.empty?
         @all_checked = (count_on_page == @document_list.count)
-        @add_files_to_collection = params.fetch(:add_files_to_collection, '')
+        @add_works_to_collection = params.fetch(:add_works_to_collection, '')
       end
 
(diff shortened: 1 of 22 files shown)


mjgiarlo:
I want to iterate on the work in this PR to move it more in the direction @awead and the ScholarSphere team did here: https://github.com/psu-stewardship/scholarsphere/commit/773960cd15e973469ccb8018bf9344c2aca167e5

PR #1465, opened 2017-08-10
Send workflow notifications only to roles associated with specific workflow

### LINKED ISSUE #1455: QA for pending review notifications
### Descriptive summary

Confirm that notifications for pending reviews are sent to the managers for an AdminSet and cc to the depositor.

Some related workflow notes are in https://github.com/samvera/hyrax/issues/936#issuecomment-319154569

### Related work

A bug in notifications was reported in https://github.com/samvera/hyrax/issues/936

The notification was fixed in https://github.com/samvera/hyrax/pull/1415, but it was merged before a complete QA was done to confirm the behavior is correct.

darrenleeweber:
Fixes #1455

This was super gnarly to untangle, partly because the root problem wasn't about notifications but about how workflow roles were being handled. We also had some prior work that brought us closer to a solution but that added new machinery instead of using the (substantial) Sipity machinery already in place, and this PR backs that work out. Here are the different components contained within this pull request.

* Admin Set Managers should be granted *all* registered roles on all available workflows for an Admin Set, not just the approving role on the active workflow
* Fix bug with `#grant_workflow_roles` that prevents creating a new corresponding workflow role when switching workflows
* Scope `PermissionQuery#scope_users_for_entity_and_roles` to the entity's workflow roles, not all roles (this method is only used in the context of this PR, so I'm not concerned about side effects)
* Disentangle `PermissionTemplateForm#update_management`, which was doing too much/being used in one more context than it needed to be
* Stop caring about hash keys in workflow notifications. (This has caused a small handful of workflow notification bugs over the past few months, and I don't think we need to care much about symbol vs. string keys since we can use indifferently accessed hashes).

This isn't perfect, but it leaves the code better than I found it and gets us closer to a more rational solution.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/workflow/notification_service.rb
+++ b/app/services/hyrax/workflow/notification_service.rb
@@ -48,18 +48,9 @@ module Hyrax
       # @return [Hash<String, Array>] a hash with keys being the strategy (e.g. "to", "cc") and
       #                               the values are a list of users.
       def recipients(notification)
-        case entity.workflow_state.name
-        when 'pending_review'
-          # notify the managers to review the deposit (CC depositors)
-          {
-            to: entity.workflow.permission_template.agent_ids_for(access: 'manage',  agent_type: 'user'),
-            cc: entity.workflow.permission_template.agent_ids_for(access: 'deposit', agent_type: 'user')
-          }
-        else
-          notification.recipients.each_with_object({}) do |r, h|
-            h[r.recipient_strategy] ||= []
-            h[r.recipient_strategy] += PermissionQuery.scope_users_for_entity_and_roles(entity: entity, roles: r.role)
-          end
+        notification.recipients.each_with_object({}) do |r, h|
+          h[r.recipient_strategy] ||= []
+          h[r.recipient_strategy] += PermissionQuery.scope_users_for_entity_and_roles(entity: entity, roles: r.role)
         end
       end
 
--- a/app/controllers/hyrax/admin/permission_template_accesses_controller.rb
+++ b/app/controllers/hyrax/admin/permission_template_accesses_controller.rb
@@ -6,7 +6,7 @@ module Hyrax
       def destroy
         ActiveRecord::Base.transaction do
           @permission_template_access.destroy
-          update_management if @permission_template_access.manage?
+          remove_access!
         end
 
         if @permission_template_access.destroyed?
@@ -28,8 +28,9 @@ module Hyrax
           @admin_set_id ||= @permission_template_access.permission_template.admin_set_id
         end
 
-        def update_management
-          Forms::PermissionTemplateForm.new(@permission_template_access.permission_template).update_management
+        def remove_access!
+          Forms::PermissionTemplateForm.new(@permission_template_access.permission_template)
+                                       .remove_access!(@permission_template_access)
         end
     end
   end
--- a/app/forms/hyrax/forms/permission_template_form.rb
+++ b/app/forms/hyrax/forms/permission_template_form.rb
@@ -20,6 +20,9 @@ module Hyrax
       # Stores the selected
       attr_writer :workflow_id
 
+      # Adding attributes hash to state to avoid having to pass it around
+      attr_accessor :attributes
+
       def workflow_id
         @workflow_id || active_workflow.try(:id)
       end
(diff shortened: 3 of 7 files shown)


app/services/hyrax/workflow/notification_service.rb
```diff
@@ -51,10 +51,10 @@ def recipients(notification)
         case entity.workflow_state.name
         when 'pending_review'
           # notify the managers to review the deposit (CC depositors)
-          {
-            to: entity.workflow.permission_template.agent_ids_for(access: 'manage',  agent_type: 'user'),
-            cc: entity.workflow.permission_template.agent_ids_for(access: 'deposit', agent_type: 'user')
-          }
+          permissions = entity.workflow.permission_template
+          managers = Array.wrap(permissions.agent_ids_for(access: 'manage', agent_type: 'user'))
```
mjgiarlo: I'm not sure we need the `Array.wrap` calls here; `agent_ids_for` should always return an array.

app/services/hyrax/workflow/notification_service.rb
```diff
@@ -51,10 +51,10 @@ def recipients(notification)
...
+          depositors = Array.wrap(permissions.agent_ids_for(access: 'deposit', agent_type: 'user'))
+          { 'to' => managers, 'cc' => depositors }
```
mjgiarlo: This change may be the more pertinent one. It could be the underlying code cares about what form its keys are in.

mjgiarlo:
When you make these changes, what change in behavior do you see? I know you've been doing a ton of QA on this.

app/services/hyrax/workflow/notification_service.rb
darrenleeweber: Also wanted to checked that the `else` block returns a hash, so I tried something similar in the REPL and it looks like it does, i.e.
```ruby
> [*1..10].each_with_object({}) {|x, h| h['a'] ||= []; h['a'].push(x); }
=> {"a"=>[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}
```

darrenleeweber:
Doing some QA just now (will update this comment as details follow).

- not in scope for this PR, but:
  - managerB gets notifications about approvals by managerA and vice-versa

darrenleeweber:
AR exception:
```
  Load LDP (72.0ms) http://127.0.0.1:8984/rest/dev/ff/36/55/24/ff365524z Service: 47081432854280
   (0.5ms)  rollback transaction

Completed 500 Internal Server Error in 3879ms (ActiveRecord: 62.3ms)

NoMethodError - undefined method `primary_key' for String:Class:
  activerecord (5.1.3) lib/active_record/reflection.rb:679:in `primary_key'
  activerecord (5.1.3) lib/active_record/reflection.rb:456:in `association_primary_key'
  activerecord (5.1.3) lib/active_record/associations/belongs_to_association.rb:79:in `replace_keys'
  activerecord (5.1.3) lib/active_record/associations/belongs_to_polymorphic_association.rb:13:in `replace_keys'
  activerecord (5.1.3) lib/active_record/associations/belongs_to_association.rb:13:in `replace'
  activerecord (5.1.3) lib/active_record/associations/singular_association.rb:15:in `writer'
  activerecord (5.1.3) lib/active_record/associations/builder/association.rb:119:in `receiver='
  activemodel (5.1.3) lib/active_model/attribute_assignment.rb:46:in `_assign_attribute'
  activemodel (5.1.3) lib/active_model/attribute_assignment.rb:40:in `block in _assign_attributes'
  activemodel (5.1.3) lib/active_model/attribute_assignment.rb:39:in `_assign_attributes'
  activerecord (5.1.3) lib/active_record/attribute_assignment.rb:26:in `_assign_attributes'
  activemodel (5.1.3) lib/active_model/attribute_assignment.rb:33:in `assign_attributes'
  activerecord (5.1.3) lib/active_record/core.rb:337:in `initialize'
  activerecord (5.1.3) lib/active_record/inheritance.rb:66:in `new'
  activerecord (5.1.3) lib/active_record/reflection.rb:154:in `build_association'
  activerecord (5.1.3) lib/active_record/associations/association.rb:281:in `build_record'
  activerecord (5.1.3) lib/active_record/associations/collection_association.rb:96:in `build'
  activerecord (5.1.3) lib/active_record/associations/collection_proxy.rb:316:in `build'
  mailboxer (0.15.1) app/models/mailboxer/message.rb:30:in `block in deliver'
  mailboxer (0.15.1) app/models/mailboxer/message.rb:29:in `deliver'
  mailboxer (0.15.1) lib/mailboxer/models/messageable.rb:78:in `send_message'
  /data/src/dlss/samvera/hyrax/app/services/hyrax/workflow/abstract_notification.rb:51:in `block in call'
  /data/src/dlss/samvera/hyrax/app/services/hyrax/workflow/abstract_notification.rb:50:in `call'
  /data/src/dlss/samvera/hyrax/app/services/hyrax/workflow/abstract_notification.rb:29:in `send_notification'
  /data/src/dlss/samvera/hyrax/app/services/hyrax/workflow/notification_service.rb:42:in `send_notification'
  /data/src/dlss/samvera/hyrax/app/services/hyrax/workflow/notification_service.rb:35:in `block in call'

```

mjgiarlo:
@darrenleeweber Here are my thoughts re: the following:
* `managerB gets an invite to approve AdminSetA works that cannot be reviewed` This is a bug that this work should address. This was the original bug report, I think?
* `managerB gets notifications about approvals by managerA and vice-versa` These are OK as long as they're actionable
* `managerB gets a redundant notification about their own approval actions` This is OK too, though perhaps annoying.

darrenleeweber:
@giarlo - what was I thinking when I said "managerB gets an invite to approve AdminSetA works that cannot be reviewed" is not in scope?  Sorry about that, that's the whole point of these changes.

I've released this from [WIP] status because the coding is done and passes the specs.  The larger workflow QA should be done before merging it.  While I'm busy on other things, if someone gets that sorted, this could be OK to merge.  As I ran into strange, likely unrelated errors, I had to stop work on the QA and haven't had time to get back to it.

mjgiarlo:
@darrenleeweber Thanks for clarifying. I'm going to begin tossing more commits at this branch in hopes of working through some of what I'm seeing locally.

no-reply:
This build failed again with a different error. I'll punt it back to @mjgiarlo to diagnose.

mjgiarlo:
@no-reply I've seen this happen with sqlite on occasion. I'm 99% confident it is unrelated to the work in this PR.

Yes, the PTForm is a monster. :unamused: Thanks for reviewing.

PR #1466, opened 2017-08-10
Fixing an occasional failing test

jeremyf:
## Fixing an occasional failing test

@391283a896f05e82c458ee09d0c83f081da6b7db

To verify the fix, you need to run the full spec suite with the
following random seed.

```console
rspec --seed 20087
```

PS: I'm not certain why this fails for some test runs but not others.

## Fixing an occasional failing test

@1819dc2f32b2de3e211ba40a81d004cb353b9013

To verify the fix, you need to run the full spec suite with the
following random seed.

```console
rspec --seed 48638
```

PS: I'm not certain why this fails for some test runs but not others.

There appear to be other problems with these views as well. [This build](https://travis-ci.org/samvera/hyrax/jobs/263066487) demonstrates another issue.

### DIFF
--- a/app/views/hyrax/embargoes/_embargo_history.html.erb
+++ b/app/views/hyrax/embargoes/_embargo_history.html.erb
@@ -1,5 +1,6 @@
+<%# To render this view pass `embargo_history` (e.g. `object: curation_concern.embargo_history`) %>
 <ul>
-  <% curation_concern.embargo_history.each do |entry| %>
+  <% embargo_history.each do |entry| %>
     <% unless entry.empty? %>
       <li><%= entry %></li>
     <% end %>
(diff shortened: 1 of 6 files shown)


jeremyf:
And now `rspec --seed 48638` is failing on a different spec. I'm updating the `./app/views/hyrax/embargoes/_list_deactivated_embargoes.html.erb` and `./app/views/hyrax/embargoes/_embargo_history.html.erb` using a similar solution.

escowles:
So sometimes `curation_concern` is available in the child partials implicitly, and other times you need to explicitly pass it?  Could we update the calling code to explicitly pass `curation_concern` as a parameter, leaving the child partials unchanged?

jeremyf:
@escowles I went back and forth on that. I can pass `{ locals: { curation_concern: curation_concern } }` to the view partials. I personally favor passing an object instead of a local. Should the called partial have knowledge that it needs to collaborate with a `curation_concern` to get the `lease_history` (eg. `curation_concern.lease_history`)? Or is it better to say "I'm responsible for rendering a `lease_history` so pass me a `lease_history` object. I the latter is a better approach in regards to predictability.

There appear to be other problems with these views as well. [This build](https://travis-ci.org/samvera/hyrax/jobs/263066487) demonstrates another issue.

escowles:
@jeremyf That's a good point — passing the `lease_history` object does remove the need for the partial to know about the curation_concern at all.  That was my only concern, so I'm 👍  on these updates.

PR #1471, opened 2017-08-10
Add collection_type to model

### LINKED ISSUE #1343: Collection Types: Create model for collection types
### Descriptive summary

Add code to implement the model part of the Rails MVC for managing collection types.

### Rationale

Site administrators can configure various types of collections as needed for their implementation.  Examples of collection types include User Collections, Exhibits, Controlled Collections, etc.

These are typically configured such that...
* User Collections - can be created by anyone, are not discoverable, works can belong to multiple user collections, optionally may be nested
* Exhibits - can be created by select staff, are discoverable, works can belong to multiple exhibits, typically are not nested
* Controlled Collections - can be created by select staff, typically discoverable, works can belong to multiple controlled collections, typically are nested.  Examples include DSpace type communities and collections, collections tied to organizational units like university departments, etc.

See [user stories](https://wiki.duraspace.org/x/YQCzB) and [requirements](https://wiki.duraspace.org/x/aojBB) for collection extensions.

### Expected behavior

Hold configuration specific to a collection type in the database.

### Implementation

**File:** app/models/hyrax/collection_type.rb
**Locaton of Data:** DB in table collection_type with an index on machine_id
**Fields:** See table below for names, descriptions, and values.

Provide general descriptive metadata fields for a collection type.

| field name | description | example values |
| ------------ | ------------ | -------------------|
| :title | Human readable label shown to users when they select collection type for new collections and in the summary table showing all collections and their types | User Collection, Exhibit, Nested Collections, etc. |
| :description: | A short description that is shown to users when they select collection type for new collections. | "A curated collection for public users." |
| :machine_id | A unique id that can be used in the future to create a class of the same name that can add additional behaviors and properties used with collections of this type.  Not used during this sprint. | user_collection, exhibit, nested_collections, etc. |

Provide property fields to hold configuration values for a collection type configuration.

| field name | description | default value | possible values |
| ------------ | ------------ | -------------| ------------------ |
| :nestable | Allow collections to be nested (a collection can contain other collections; a collection cannot have itself as an ancestor) | true | true, false |
| :discovery | Allow collections to be discoverable | true | true, false |
| :sharing | Allow users to assign collection managers, depositors, and viewers for collections they manage | true | true, false |
| :multiple_membership | Allow a work to belong to multiple collections of this type | true | true, false |
| :require_membership | A work must below to at least one collection of this type | false | true, false |
| :workflow | Allow collections of this type to assign workflow to a new work | false | true, false |
| :visibility | Allow collections of this type to assign initial visibility settings to a new work | false | true, false |

Provide ability to assign participants to a collection type.  
* collection managers
* collection creators

#### Outstanding Questions:
How do permission templates for participants fit into this work?  Are they created as part of the collection type configuration?  Or are they used by the individual collections to hold participants? or both? Are they created when the collection type is created?  See Code for Permission Templates in the wiki for an exploration of this code... https://wiki.duraspace.org/x/2aA2BQ

### Related work

| Collection Types feature | Issue |
| Admin Set Generator      | #1342 |
| User Collection Generator | #1485  |
| Routes                   | #1347 |
| Controller               | #1341 |
| Model                    | this issue |
| View - Index             | #1349 |
| View -  New/Edit         | #1348 |
| Partial -  Description   | #1344 |
| Partial -  Settings      | #1346 |
| Partial -  Participants  | #1345 |

### LINKED ISSUE #1383: Collections: Modify model to include Collection Type and settings methods
### Descriptive summary

Add field collection_type to the model part of the Rails MVC for managing collections.

### Rationale

This effort is expanding the existing Collections implementation to accommodate Collection Types and the new [requirements](https://wiki.duraspace.org/x/g402BQ) that are part of the [Collection Extensions](https://wiki.duraspace.org/x/5JM2BQ) work.

### Expected behavior

Hold existing metadata properties and new :collection_type property.  Provide access to collection type configurations.

### Implementation

**File:** [app/models/concerns/hyrax/collection_behavior.rb](https://github.com/samvera/hyrax/blob/master/app/models/concerns/hyrax/collection_behavior.rb)
**Location of Data:** Fedora object with model Collection
**Fields:** See table below for names, descriptions, and values.

Provide property fields to hold configuration values for a collection configuration by adding the property definition to collection_behavior.rb.

| field name | predicate | description | example value | comment |
| ------------ | ----------- | --------------| ---------------- | ----------- |
| collection_type | _See question below_ | Machine id | user_collection, exhibit, nested_collection, etc. | When instantiated, should this return the machine_id or an instance of the collection_type? | 

question_24 **QUESTION:** What predicate should be used to store the collection_type for a collection?

NOTE: There is no metadata defined in collection_behavior.rb.  Collections get metadata through [Hyrax::CoreMetadata](https://github.com/samvera/hyrax/blob/master/app/models/concerns/hyrax/core_metadata.rb) included in collection_behavior.rb.  And through [Hyrax::BasicMetadata](https://github.com/samvera/hyrax/blob/master/app/models/concerns/hyrax/basic_metadata.rb) included in the generated Collection class in the app.  Both core and basic metadata are also used by Works and are not good locations for adding the collection_type, which is specific to collections only.

Provide methods that the collection can use to determine it's configuration based on the collection type by adding the methods to collection_behavior.rb.

| method name | description | returns |
| ------------ | ------------ | -------------|
| nestable? | Can this collection be nested in others of its type? | true, if it can; otherwise, false |
| discoverable? | Can this collection be discovered by public users? | true, if it can; otherwise, false |
| shareable? | Can this collection be shared with other users? | true, if it can; otherwise, false |
| multiple_membership? | Can works belong to more than one collection of this type? | true, if it can; otherwise, false |

NOTE: These methods should get this information from the collection's collection type.  

Implementation Question:  We have the machine_id and can query the DB for the setting.  Would it be more efficient to have an instance of the collection_type accessible from the collection in order to answer these questions for the collection?  Or a pointer to a shared instance of the collection_type?

#### SPECIAL NOTE on Admin Sets:

For the initial implementation, the Admin Set underlying code will not change.  But to a user, the Admin Set will appear to be a collection.

To make the illusion happen...
* Two possible approaches:
  * OPTION 1: Add collection_type property and the methods to the Admin Sets code.  Then use the same processing mechanism as for regular collections.
  * OPTION 2: Admin Sets can have the same methods added, but they can be hardcoded to return the appropriate responses as follows...
    * nestable? false
    * discoverable? - get from collection type
    * shareable? true
    * multiple_membership? false

Option 1 might save us some time down the road.

### Related work

| Model                    | (this issue) |

elrayle:
Partially Fixes #1383

* Adds collection_type_gid as a property that will be stored in Fedora and identify the collection_type of a specific collection.
* Adds collection_type instance and delegates settings methods to the instance.

Remaining work:
* Still needs collection type processing in controller to convert between controller_type instance and gid, and vice versa. I have some of that work done, but don’t want to include it in this PR.
* Admin set work - see Issue #1383 for more information

### DIFF
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -14,7 +14,14 @@ module Hyrax
     included do
       validates_with HasOneTitleValidator
       self.indexer = Hyrax::CollectionIndexer
+
+      property :collection_type_gid, predicate: ::RDF::Vocab::SCHEMA.additionalType, multiple: false do |index|
+        index.as :symbol, :stored_searchable
+      end
     end
+
+    attr_accessor :collection_type
+    delegate :nestable?, :discoverable?, :sharable?, :allow_multiple_membership?, :require_membership?, :assigns_workflow?, :assigns_visibility?, to: :collection_type
 
     # Add members using the members association.
     def add_members(new_member_ids)
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -16,5 +16,9 @@ module Hyrax
     deprecation_deprecate workflow: "prefer #assigns_workflow instead"
     alias_attribute :visibility, :assigns_visibility
     deprecation_deprecate visibility: "prefer #assigns_visibility instead"
+
+    def gid
+      URI::GID.build app: GlobalID.app, model_name: model_name.name.parameterize.to_sym, model_id: id unless id.nil?
+    end
   end
 end
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -1,5 +1,5 @@
 RSpec.describe Collection do
-  let(:collection) { create(:public_collection) }
+  let(:collection) { build(:public_collection) }
 
   it "has open visibility" do
     expect(collection.read_groups).to eq ['public']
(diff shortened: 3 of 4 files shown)


elrayle:
READY TO MERGE.  According to @mjgiarlo, the error is in flappy feature specs, unrelated to this PR.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -14,8 +14,15 @@ module CollectionBehavior
     included do
       validates_with HasOneTitleValidator
       self.indexer = Hyrax::CollectionIndexer
+
+      property :collection_type_gid, predicate: ::RDF::Vocab::SCHEMA.additionalType, multiple: false do |index|
+        index.as :symbol, :stored_searchable
+      end
     end
 
+    attr_accessor :collection_type
```
jeremyf: Should there be validation that the `collection_type` is set and persisted?

Should the `nestable?`, `discoverable?` etc. methods return a default value (based on the column defaults)? If so, I have some ideas on implementation.

I don't think that `collection_type=` should be a method on the collection. We set the collection via the `collection_type_gid`.

```ruby
def collection_type
  if collection_type_gid.present?
    CollectionType.find(collection_type_gid) # or however we retrieve collection types
  else
    nil
  end
end
```

elrayle: RE: validation - That makes sense.  There are some loose ends around this because the code for setting this in the controller isn't in place yet.  The load happens for edit, but is pending some work on the New Collection to incorporate the modal selection of collection type process.  And we don't have all the collection_type code done to be able to create a collection type to be selected.  Bottom line, we can't put the set and persisted test in place until some of these other parts fall into place.  I'll make a note in the main issue for this.

RE: nestable?, etc. - These methods are delegates to the collection_type.  If the collection_type doesn't exist, then teh object is in an error state.  Any defaults should be controlled by the collection_type.

RE: collection_type= - Depends on how we want to set the collection type for new collections?  Thinking through the implications...  this means that the collection_type needs to provide its gid AND the new collection modal for selecting a collection_type should have the gid for each collection_type such that when the new collection object is created, it is passed the gid from the modal.

jeremyf: Thanks for the responses. This is a great next step, and we'll continue to iterate.

Issue #1349, opened 2017-07-14
Collection Types: Create index views for managing collection types


elrayle:
### Descriptive summary

Add code to implement the index view for listing collection types.

### Rationale

See model issue for rationale.

### Implementation

**File:** app/views/hyrax/admin/collection_types/index.html.erb

#### Expected user interactions

Create the UI view for listing collection types.

From this page, a user can...
- [x] see a list of existing collection types
- [x] click a button to create a new collection type goes to `/admin/collection_types/new`
- [x] click edit for an existing collection type goes to `/admin/collection_types/{:id}/edit`
- [x] click delete for an existing collection type
  - [x] disallow if there are any collections of this type
  - [x] show a confirmation box stating this cannot be undone

Exact text for the page is in the mockup below.

#### SPECIAL NOTE on Admin Sets:

For the initial implementaiton, the Admin Set underlying code will not change.  But to a user, the Admin Set will appear to be a collection type.

To make the illusion happen...
* as part of the hyrax generation process, a collection type will be created for Administrative Sets and the id will be recorded in a variable that can be checked by this UI (see issue #1342 for Admin Set Generator)
* if current_collection_type == admin_set_collection_type_id, do not include a remove button

### Mockup

![collection-types](https://user-images.githubusercontent.com/101482/28342644-22a99ada-6bce-11e7-9532-1903a1e9bb0c.png)

### Related work

| Collection Types feature | Issue |
| ------------------------ | ----- |
| Admin Set Generator      | #1342 |
| User Collection Generator | #1485  |
| Routes                   | #1347 |
| Controller               | #1341 |
| Model                    | #1343 |
| View - Index             | this issue |
| View -  New/Edit         | #1348 |
| Partial -  Description   | #1344 |
| Partial -  Settings      | #1346 |
| Partial -  Participants  | #1345 |

adamjarling:
@elrayle Should updating the sidebar be part of this story, or just the new view/page itself?

elrayle:
@adamjarling - The sidebar link to the Collection Types index page was added by PR #1446.

elrayle:
This is complete but will have one minor change.  See issue #1486.

elrayle:
From PR #1472 

#### Issues remaining:
 -  We added a `collections?` method to the CollectionsType model so that we could simulate different modals for empty vs full collection types. This will need to be changed when we have implemented a way to find collections that belong to a collection type. (being addressed by Issue #1341)
[...]

PR #1472, opened 2017-08-10
Collections #1349 index views

adamjarling:
Moves forward issue #1349 ; 

The initial Collection Types index page, markup, and basic UI wired up.  Still needs completion of other stories to be fully functional.

@samvera/hyrax-code-reviewers

Issues remaining:
 -  We added a `collections?` method to the CollectionsType model so that we could simulate different modals for empty vs full collection types. This will need to be changed when we have implemented a way to find collections that belong to a collection type.
 - Added `collection_type.machine_id != 'admin_set'` as a placeholder to check if the collection type was admin set. Not sure what the final check will be yet.
 - Actual delete functionality not wired up yet
 - Button to take you to 'View collections of this type' (if unable to delete because collection type has collections) not wired up yet

### DIFF
--- a/app/assets/javascripts/hyrax/app.js
+++ b/app/assets/javascripts/hyrax/app.js
@@ -12,6 +12,7 @@ Hyrax = {
         this.datatable();
         this.adminSetEditor();
         this.collectionEditor();
+        this.collectionTypes();
         this.adminStatisticsGraphs();
         this.tinyMCE();
         this.perPage();
@@ -40,6 +41,11 @@ Hyrax = {
       var controls = new CollectionControls($('#collection-controls'));
     },
 
+    // Collection types
+    collectionTypes: function() {
+      var CollectionTypes = require('hyrax/collection_types');
+      var collection_types = new CollectionTypes($('.collection-types-wrapper'))
+    },
 
     // Pretty graphs on the dashboard page
     adminStatisticsGraphs: function() {
--- a/spec/views/hyrax/admin/collection_types/index.html.erb_spec.rb
+++ b/spec/views/hyrax/admin/collection_types/index.html.erb_spec.rb
@@ -1,0 +1,20 @@
+require 'spec_helper'
+
+RSpec.describe 'hyrax/admin/collection_types/index.html.erb', type: :view do
+  before do
+    assign(:collection_types, [
+             FactoryGirl.create(:collection_type, title: 'Test Title 1', machine_id: 'test_title_1'),
+             FactoryGirl.create(:collection_type, title: 'Test Title 2', machine_id: 'test_title_2')
+           ])
+    render
+  end
+
+  it 'lists all the collection_types' do
+    expect(rendered).to have_content('Test Title 1')
+    expect(rendered).to have_content('Test Title 2')
+  end
+
+  it 'displays the collection type count correctly' do
+    expect(rendered).to have_content '2 collection types in this repository'
+  end
+end
--- a/app/assets/javascripts/hyrax.js
+++ b/app/assets/javascripts/hyrax.js
@@ -82,6 +82,7 @@
 //= require hyrax/relationships
 //= require hyrax/select_work_type
 //= require hyrax/collections
+//= require hyrax/collection_types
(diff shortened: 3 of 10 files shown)


adamjarling:
Sorry, didn't know about the houndci task, that you can run locally or that it even existed.  Working on updating now...

app/assets/javascripts/hyrax/app.js
```diff
@@ -40,6 +41,14 @@ Hyrax = {
       var controls = new CollectionControls($('#collection-controls'));
     },
 
+    // Collection types
+    collectionTypes: function() {
+      var element = $('.collection-types-wrapper')
+      if (element.length > 0) {
+        var CollectionTypes = require('hyrax/collection_types');
```
jeremyf: @adamjarling Do you know what is happening here? I assume that requires should not be part of conditional logic.

adamjarling: @jeremyf Yea agree.  I followed the lead of how the 'workEditor' function (line 60 in app.js) created a new Editor instance.  I can move this logic into the actual class itself though.  Would that work?

elrayle: There are requires all through this file that follow the same pattern.  Now that doesn't necessarily mean it is the best approach.  @jeremyf do you have an alternative suggestion?

elrayle:
I ran the changes in my local app and the UI looks good.  I'll wait for @jeremyf to complete the review based on his js comments.

app/assets/javascripts/hyrax/app.js
adamjarling: One alternative way would be to perform the element exists check within class constructor.   It would look cleaner.

jeremyf: @elrayle I don't have recommendations. Just an observation. I defer to @adamjarling's JS experience.

adamjarling: I'll push up a quick change.  I agree with @jeremyf original assumption, it may establish a better pattern.

spec/views/hyrax/admin/collection_types/index.html.erb_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+RSpec.describe 'hyrax/admin/collection_types/index.html.erb', type: :view do
+  before do
+    assign(:collection_types, [
+             FactoryGirl.create(:collection_type, title: 'Test Title 1'),
+             FactoryGirl.create(:collection_type, title: 'Test Title 2')
```
elrayle: The failing test...
```
Failure/Error: FactoryGirl.create(:collection_type, title: 'Test Title 1'),
     ActiveRecord::RecordInvalid:
       Validation failed: Machine can't be blank
```
is happening because machine_id is required.  You can fix by adding `machine_id: "test_title_1"` to your factory.  See [predefined user_collection factory](https://github.com/samvera/hyrax/blob/collections-sprint/spec/factories/collection_types.rb#L3-L6) as an example.

Also, will the test run correctly with `build` instead of `create`?  `build` is faster than `create` and always preferable unless saving to the DB is required for the test.

adamjarling: @elrayle @kdid Thanks for the feedback.  I'll make these adjustments now and see if we can get it green.

spec/views/hyrax/admin/collection_types/index.html.erb_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+RSpec.describe 'hyrax/admin/collection_types/index.html.erb', type: :view do
+  before do
+    assign(:collection_types, [
+             FactoryGirl.create(:collection_type, title: 'Test Title 1', machine_id: 'test_title_1'),
```
adamjarling: @elrayle I think your previous comments were erased when I squashed commits, but just pointing out we updated this, and in fact we do need `FactoryGirl.create` here instead of build for the test to pass.

PR #1474, opened 2017-08-10
Speeding up tests

jeremyf:
## Aligning spec description with spec body

@e55e9a25a7b7fe12d1c7ad86752cd1d82d7badad

I don't believe that testing the existence of a title and description
should require saving AdminSet.

## Reducing LDP requests in spec/collections_spec.rb

@404ee271a8a61bce049ad02a2784e097101f6942

After:

```console
Collection#validates_with ensures the collection has a title
  Total LDP: 3 {"GET"=>2, "POST"=>1}
...
Finished in 8.98 seconds (files took 6.93 seconds to load)
```

Before:

```console
Collection#validates_with ensures the collection has a title
  Total LDP: 16 {"HEAD"=>2, "GET"=>10, "POST"=>4}
...
Finished in 10.18 seconds (files took 7.07 seconds to load)
```

## Moving `let!` directives closer to their uses

@0e8cbb4c671e28215b9f4dd4ca6e6af1b3aca96e

This is a minor speed increase, as the three `let!` statements are not
fired before each and every `it` block in this spec file.

Before:

```console
Finished in 0.71341 seconds (files took 7.8 seconds to load)
```

After:

```console
Finished in 0.52115 seconds (files took 7.16 seconds to load)
```

## Removing an ActiveFedora create in spec

@16ac4e53685b156a009fddde5ece7fbafa13279f

We can instead rely on the identifier. This drops 14 LDP requests and
shaves 0.5 seconds from the specs.

Before:

```console
Examples with the most LDP requests
  FeaturedWorkList#featured_works_attributes= sets order
    Total LDP: 14 {"HEAD"=>2, "GET"=>9, "POST"=>3}
  FeaturedWorkList featured_works is a list of the featured work objects, each with the generic_work's solr_doc
    Total LDP: 29 {"HEAD"=>5, "GET"=>18, "POST"=>6}
  FeaturedWorkList featured_works when one of the files is deleted is a list of the remaining featured work objects, each with the generic_work's solr_doc
    Total LDP: 32 {"HEAD"=>4, "GET"=>19, "POST"=>6, "DELETE"=>3}

Finished in 5.21 seconds (files took 7.4 seconds to load)
```

After:

```console
Examples with the most LDP requests
  FeaturedWorkList featured_works is a list of the featured work objects, each with the generic_work's solr_doc
    Total LDP: 29 {"HEAD"=>5, "GET"=>18, "POST"=>6}
[...]

PR #1475, opened 2017-08-10
make labels in discovery pane bold.

blancoj:
This makes the labels in the discovery pane bold.


jeremyf:
The spec failed for a known, yet unrelated issue. See https://github.com/samvera/hyrax/pull/1474 for the fix. I'm going to go ahead and merge.

PR #1478, opened 2017-08-10
Access-granting workflow services should work on member file sets

### LINKED ISSUE #1477: In one-step workflow, grant read/edit to depositor down to members
The workflow service that is run when a work entered the one-step workflows revokes edit access from the depositor and propagates that down to the work's member file sets. When granting read and edit access back to the user later in workflow, those changes are *not* propagated down to the work's file sets, so the depositor cannot access the files they uploaded.

mjgiarlo:
Fixes #1477

Also, `RevokeEditFromDepositor` service should invoke job asynchronously if its intent is to run in the background.

@samvera/hyrax-code-reviewers


mjgiarlo:
@jeremyf Done. Re-review when you get a chance, pls?

mjgiarlo:
Note that the failures are due to flappy feature specs.

Issue #1460, opened 2017-08-09
Collections: collections with collection_type.discoverable? false are always Private


elrayle:
### Descriptive summary

For new/edit, set collection visibility to Private when collection_type.discoverable? is false

### Rationale

Implementing the simplest approach for the first pass of always making the visibility Private when a collection is not discoverable.  This can be revisited in the future if use cases arise for allowing setting the value of institution or private.

### Implementation

See related issues for view files involved.

- [x] Hide Discovery tab on new/edit collection form is discoverable? false  (this task is part of issue #1374 collection new/edit form)
- [x] Automatically set visibility=Private for collections when discoverable? false

Question: Where should the auto-setting of visibility been done?
* Option 1: Include a hidden field in the new/edit form that holds the visibility value.  This has the advantage of not requiring any other processing changes.
* Option 2: In the actor for the collection where save is processed.  This has the advantage of ensuring the parameter cannot be changed by some other means between the submit and the processing of the submit.

I did not find an actor class for collections under [app/actors/hyrax/actors](https://github.com/samvera/hyrax/tree/master/app/actors/hyrax/actors).  The processing of #create happens in the [app/controllers/concerns/hyrax/collections_controller_behavior.rb](https://github.com/samvera/hyrax/blob/master/app/controllers/concerns/hyrax/collections_controller_behavior.rb#L96-L111)

More exploration is needed.

### Related work

Issue #1374
Issue #1379

FAQ for Discovery - [samvera_docs/collection_extensions_faq_discovery.html](http://elr37-dev.library.cornell.edu/samvera_docs/collection_extensions_faq_discovery.html)

elrayle:
Fixed by PR #1556

PR #1484, opened 2017-08-11
default to private when coll is not discoverable.

blancoj:
@elrayle could you take a look at this and let me know if this will work for #1460 once dicoverable? becomes available.


app/controllers/concerns/hyrax/collections_controller_behavior.rb
```diff
@@ -103,6 +103,8 @@ def create
       @collection.attributes = collection_params.except(:members)
       @collection.apply_depositor_metadata(current_user.user_key)
       add_members_to_collection unless batch.empty?
+      @collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE
```
elrayle: Should this be...
```
@collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE unless @collection.discoverable?
```
Otherwise, aren't you setting all collections to private?  Am I missing something?

elrayle: I wonder if for security purposes, the same check should happen for #update, even though it should already be set there.

blancoj: Should it be set to open if if @collection.dicoverable? is set to true here at create method?

As for the update method, I think that the default setting at creation should take care of it.

elrayle: I took a look at this and can't find where the default visibility is set.  But I did confirm that it comes into the new form with visibility='restricted' (aka private) and it always leaves the form that one because the Discovery tab is not visible until the Collection is created.

I do think that after update should force the visibility to private if discoverable? false

mjgiarlo:
This work is now being done in another branch.

PR #1487, opened 2017-08-14
Addressing todo items

jeremyf:
## Addressing TODO item

@17bb5d194723311b5d36d8c157115ba1da4fd817

There exists a view spec (eg. `/app/views/hyrax/users/index.html.erb`)
that covers the rendering feature test.

## Removing spec that is skipped

@3761156c14231978716dafa1c81ca7e11effdcf3

Avoid paying the setup cost of a spec that is skipped. On a quick check
the issue (https://github.com/samvera/hyrax/issues/1699) is missing (eg
404).

## Adding documentation to scenario

@2ffae937ed06692e9d16abd6d36633f4135878ee

Instead of relying on comments, rely on documentation.


spec/features/users_spec.rb
```diff
@@ -5,15 +5,6 @@
   let(:user) { create(:user) }
   let(:profile_path) { Hyrax::Engine.routes.url_helpers.user_path(user, locale: 'en') }
 
-  context 'when clicking all users' do
-    # TODO: Move this to a view test
```
jcoyne: Did this move to a view test?

jeremyf: The following view spec (`app/views/hyrax/users/index.html.erb`) already existed.

jcoyne: 👍

PR #1436, opened 2017-08-07
Swap out poltergeist for selenium-webdriver and headless chrome for JS testing

mjgiarlo:
@samvera/hyrax-code-reviewers

PR #1491, opened 2017-08-14
Use upstream implementation of `#accept_confirm`

### LINKED ISSUE #1445: Remove custom accept_confirm implementation
#1436 changed our feature specs such that they now use Selenium with headless Chrome instead. Using Capybara with this stack does not yet support the `#accept_confirm` method (and related methods) for interacting with alert/confirm browser dialogs. We've added a custom handler for this in the meantime at ` spec/support/features/confirmation.rb` -- when this is supported in the Capybara-atop-Selenium-with-Chrome-driver stack, remove this custom code.

mjgiarlo:
@twalpole (from the Capybara team) commented on #1445 that support for `#accept_confirm` for headless Chrome is already present in Capybara. I am not sure how my testing last week failed to turn this up... This commit removes the temporary workaround added in #1436 and uses the upstream implementations of both `#accept_confirm` and the headless Chrome javascript driver.

Fixes #1445

(Thanks a bunch for commenting over here, @twalpole, and for all the work you do on Capybara!)

@samvera/hyrax-code-reviewers


twalpole:
@mjgiarlo You're welcome. One quick comment on this PR.  You should still be calling `accept_confirm { click_link 'Transfer Ownership of Work' }` otherwise if you ever switch to non-headless chrome (for debugging, etc) your test will fail.

mjgiarlo:
@twalpole I moved that call because the `'Transfer Ownership of Work'` link does not even use a confirmation dialog (whereas the `'Transfer'` button does). That was an unseen mistaken in the prior PR.

twalpole:
@mjgiarlo Ah, perfect then :)

PR #989, opened 2017-05-17
[WIP] Port batch edit spec changes from ScholarSphere

mjgiarlo:
Do this to address the transient CI failures caused by batch_edit_spec.rb.

Fixes #815

This is still very much in progress, and I need to move on to other work. Pushing up this branch for later completion (by me or anyone who's interested). Any chance you'll get cycles to put towards Hyrax, @awead? :smiley:

Issue #815, opened 2017-04-21
Consider adopting changes to batch edit spec from ScholarSphere


mjgiarlo:
@awead :speech_balloon: 
> @atz @mjgiarlo we cleaned up our batch edit feature test. Hyrax has a version of our old test. Refactoring to match ours might help: https://github.com/psu-stewardship/scholarsphere/blob/develop/spec/features/batch_edit_spec.rb
> 
> One of the key bits was disabling animations: https://github.com/psu-stewardship/scholarsphere/blob/develop/spec/features/batch_edit_spec.rb#L10

This may address the intermittent failure issue we've been seeing.

PR #1492, opened 2017-08-14
Nudging JS wait times and testing sequence

jeremyf:
Instead of removing the sleep time, rely instead of Capybara's wait
time. Hyrax is encountering regular failures on finding this content as
part of the Travis build. Hopefully this tweak can assist in reducing
the number of Capybara induced failures in the Travis specs.

Incrementing from 5 to 15 to see if we can prevent erratic failures in
Travis. These failures result in a broken build and require a restart
of the specs. This restart and waiting for feedback slows the overall
trajectory of iterating on Hyrax.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/batch_edit_spec.rb
+++ b/spec/features/batch_edit_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe 'Batch management of works', type: :feature do
   end
 
   describe 'editing and viewing multiple works' do
-    before do
+    it 'edits each field and displays the changes', clean_repo: true, js: true do
       check 'check_all'
       click_on 'batch-edit'
       fields.each do |f|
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@jeremyf I wouldn't reject this PR, but I will say I believe the Penn Staters found a better solution:

https://github.com/samvera/hyrax/issues/815

I began making that change here: https://github.com/samvera/hyrax/pull/989 (I am unlikely to get back to that any time soon, so folks are free to pick that up if they are inclined!)

jeremyf:
@mjgiarlo I like your change…however, I won't have time for awhile. I believe this is still a nominal improvement and would merit merging.

PR #1490, opened 2017-08-14
Collections: Refactor CRUD moving hyrax/collections to hyrax/dashboard/collections

elrayle:
Partially addresses #1380 Collections: Adjust routes to accommodate a public landing page and permission templates.

 Refactor CRUD moving hyrax/collections to hyrax/dashboard/collections.  Also adds a second show page under hyrax/dashcoard/collections that only those with edit access can view and use to perform management tasks for the collection.  The original show page under hyrax/collections has had management tasks removed in preparation of becoming the public branded landing page.
    
There are a few errors that were not fully resolved as part of this commit.  See...
* #1488 Collection CRUD refactor - test 'removes all works' fails
* #1489 Collection CRUD refactor - Throwing ActionController::InvalidAuthenticityToken when trying to add works to an existing collection

Impacts issues related to CRUD for Collections, including...
* #1417 Collections: Update UI for index views (All/Your) of MVC for Managing Collections
* #1427 Collections: Create an admin show view
* #1428 Collections: Update public landing page to include branding and other UI recommendations
* #1374 Collections: Modify existing new/edit view to include partials for new tabs

Primary changes that impact these issues...
* All issues...
  * many paths are updated to include dashboard in the path
* Effecting 1374 new/edit view
  * management views and partials (e.g. new, edit, + partials) moved to [views/app/hyrax/dashboard/collections](https://github.com/samvera/hyrax/tree/collections-sprint/app/views/hyrax/dashboard/collections)
  * [app/controllers/concerns/hyrax/collections_controller_behavior.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/concerns/hyrax/collections_controller_behavior.rb) no longer controls new/edit/destroy CRUD.
  * [app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb) now controls new/edit/destroy CRUD.  NOTE: This controller behavior is included in [app/controllers/hyrax/dashboard/collections_controller.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/hyrax/dashboard/collections_controller.rb) which inherits from [app/controllers/hyrax/my/collections_controller.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/hyrax/my/collections_controller.rb)
* Effecting 1427 & 1428 both show pages
[...]

Issue #1380, opened 2017-07-24
Collections: Adjust routes to accommodate a public landing page and permission templates


elrayle:
### Descriptive summary

Adjust existing routes and define new routes to separate management of collections from public Define the routes for permission templates usage in collections.  I'm not sure what is needed.

### Rationale

See model issue for rationale.

### Implementation

The primary management routes for collections are currently defined at...
- new/edit/show - https://github.com/samvera/hyrax/blob/master/config/routes.rb#L49-L59
- dashboard (All) - https://github.com/samvera/hyrax/blob/master/config/routes.rb#L124
- dashboard (Yours) - https://github.com/samvera/hyrax/blob/master/config/routes.rb#L136-L138

There may be other related definitions.

Changes needed...
- [x] Need show route that is accessible from admin menu only (e.g. /dashboard/collections/:id) (#1427)
- [x] Need show route for (landing page) that is accessible by public users (e.g. /collections/:id) (#1428)
- [x] Need to move CrUD routes for collections to /dashboard/collections
  - [x] Move new/edit and partials to app/views/dashboard/collections (#1374 & partials in related work below)
- [ ] Need routes related to permission templates - Moved to issue #1523

More exploration is needed on how to integrate with permission templates and their impact on routes.

![question_24](https://cloud.githubusercontent.com/assets/6855473/13064236/f2f04cbe-d41e-11e5-9674-e9a56a6326e6.png) **QUESTION:** Should all management routes be moved under the admin path, leaving only the landing page directly off home path (e.g. /admin/collections/...)?  What would be the impact of this change?

_Answer:_ Move to /dashboard/collections.  See comments for more details.

### Related work

| Collections feature | Issue |
| ------------------------ | ----- |
| Routes                   | (this issue) |
| Controller               | # |
| Model                    | #1383 |
| View - Index            | #1417 |
| View - Show - Admin      | #1427 |
| View - Show - Landing    | #1428 |
| View - New/Edit          | #1374 |
| Partial -  Description   | #1375 |
| Partial -  Branding      | #1376 |
| Partial -  Relationships | #1377 |
| Partial -  Sharing       | #1378 |
| Partial -  Discovery     | #1379 |

mjgiarlo:
> QUESTION: Should all management routes be moved under the admin path, leaving only the landing page directly off home path (e.g. /admin/collections/...)? What would be the impact of this change?

What routes do you include in "all management routes" here?

elrayle:
[...]

PR #1499, opened 2017-08-15
Refactor CRUD for collections from hyrax/collections to hyrax/dashboard/collections

### LINKED ISSUE #1493: Refactor Collections CRUD from hyrax/collections to hyrax/dashboard/collections
### Descriptive summary

Refactor Collections CRUD from hyrax/collections to hyrax/dashboard/collections.  This applies to new/edit and creates a new show page that provides access to management tasks.  The original show page will remain at hyrax/collections, but will not include access to management tasks.  Your/All Collection index pages will send uses to the management show page when collection titles are clicked.

### Rationale

This separates management functions of the UI from publicly consumed pages of the UI.  The Collection Extensions work is adding in a landing page with branding that will become the public show page.  This work sets the stage for the collection extensions work and keeps the two development efforts in sync.

### Implementation

See the issues listed under Related Issues for discussions that led to this implementation approach.

Proposed views and controllers...

| view | controller#action | comments |
| ----- | ------------------- | ------------ |
| views/hyrax/dashboard/collections/index | controllers/hyrax/dashboard/collection_controller#index | admin All Collections tab in dashboard |
| views/hyrax/dashboard/collections/show | controllers/hyrax/dashboard/collection_controller#show | manager's view of the collection |
| views/hyrax/dashboard/collections/new | controllers/hyrax/dashboard/collection_controller#new | |
| views/hyrax/dashboard/collections/edit | controllers/hyrax/dashboard/collection_controller#edit | |
| views/hyrax/my/collections/index | controllers/hyrax/dashboard/collection_controller#index | all users Your Collections tab in dashboard |
| views/hyrax/collections/show | controllers/hyrax/collection_controller#show | public landing page |

### Related work

#1380 Collections: Adjust routes to accommodate a public landing page and permission templates
#1490 Collections: Refactor CRUD moving hyrax/collections to hyrax/dashboard/collections

elrayle:
Fixes #1493

Refactor CRUD for collections from hyrax/collections to hyrax/dashboard/collections

See collections-sprint PR #1490 provides impact information and is essentially the same PR against a different branch.  That PR will be reworked when this is committed to keep collections-sprint branch and master in sync.

@samvera/hyrax-code-reviewers


.rubocop_fixme.yml
```diff
@@ -23,7 +23,7 @@ Metrics/AbcSize:
 
 Metrics/ModuleLength:
   Exclude:
-    - 'app/controllers/concerns/hyrax/collections_controller_behavior.rb'
+    - 'app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb'
```
jcoyne: I believe this should be `admin` rather than `dashboard` for consistency.

app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb
```diff
@@ -0,0 +1,279 @@
+module Hyrax
+  module Dashboard
```
jcoyne: I think this should be `Admin` rather than `Dashboard`.  This follows the existing pattern.

app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb
```diff
@@ -0,0 +1,279 @@
...
+    module CollectionsControllerBehavior
```
jcoyne: I think we made the call to do away with the `*ControllerBehavior` pattern as now Ruby has `Module#prepend` which can be used to achieve the same goals. This also removes some indirection from the gem.

.rubocop_fixme.yml
elrayle: See the discussion with @mjgiarlo in issue #1380.  I was originally going with admin, but he suggested dashboard because this functionality is not limited to admins.

app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb
elrayle: Same as above.

app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb
elrayle: Can we make this a separate issue?  That would put a major change in that does not correspond to the original code that was being moved.  I tried to keep the structure of the code in the new location as close to the original as possible to make the PR easier to review.  It's already a bear with all the changes required for basic refactoring to a new location.

.rubocop_fixme.yml
jcoyne: This is already a discussion I had with @ggeisler who I believe indicated we should us "admin" for any work that had to do with administration of repository objects. This is not an area that is restricted to administrators.

mjgiarlo: My answer about using `/dashboard/` was about the routes/URLs. (We don't see `/admin/` in many routes.) @jcoyne's point is about module namespaces.

app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb
jcoyne: I don't think we should because you've created a whole new module here. It's more challenging to track history if we have to do code moves between files multiple times.

.rubocop_fixme.yml
elrayle: More comments from slack...

jcoyne: 
> he (@ggeisler) cares about whether the url is `/admin/collections` or `/dashboard/collections`
> And it’s confusing to have modules use different namespaces from the conventions

mjgiarlo:
> we're currently in an inconsistent state, it seems?
> Put otherwise, should @elrayle's PR be consistent with this? https://github.com/samvera/hyrax/blob/master/config/routes.rb#L124  Or this? https://github.com/samvera/hyrax/blob/master/config/routes.rb#L200

elrayle:
> Here is the distinction that I groked from our original conversation.  
> * Managing admin_sets, users, workflows, settings, etc is done by an admin. (use `/admin/`)
> * Managing works, collections, viewing stats can be done by ordinary users. (use `/dashboard`)

mjgiarlo:
> The move from admin to dashboard makes sense to me given that we went from a rigid distinction between "admin dashboard" and "user dashboard" to having a single dashboard that shows users information based on their abilities.
> That sounds reasonable to me, @elrayle.

elrayle:
> Seems to me that works management should have a similar refactor away from hyrax/concerns/generic_work to hyrax/dashboard/works/generic_work.

@ggeisler Want to provide your perspective on  collections under admin vs dashboard?  Potential works refactor is not on the table right now.

app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb
elrayle: @jcoyne Can you point me to an example usage in Hyrax?  I searched for 'prepend' but don't see anything that looks equivalent to replacing behavior modules.  The search included hyrax code and the test app.

elrayle: My inclination is to not make the refactor even more complex.  I will ticket this for follow-on work, given the codebase does not consistently adopt the new behavior-less approach.

.rubocop_fixme.yml
ggeisler: @jcoyne is correct that in the past I argued for referring to anything a user can do administratively (that is, things that involved managing objects) as "administration" and thus the logical path would include `/admin`.

However, I think @mjgiarlo made a suggestion more than once (not sure whether it was in Slack, Github tickets, or both, and hope I'm not misrepresenting you, @mjgiarlo!) to use "dashboard" to refer to what I was calling "administration," I believe to remove the potential confusion over what a repository administrator can do and what a user with edit privileges (for her own works, for example) can do, since some people seem to consider only what a repository administrator can do as "administration."

So I personally prefer "administration" (see my last paragraph below for more explanation of why) but I don't think my preference is strong enough to counter the argument for "dashboard" if @mjgiarlo and most people in the community think that is a more intuitive term to use. That's why I didn't object to switching to "dashboard." I don't want to stand in the way if the general consensus is that "dashboard" makes more sense.

I am a little worried about trying to use both terms, as outlined in the quoted text in @elrayle's comment above, because repository administrators can also manage another user's work, but using that scheme that action would come under `/dashboard` not `/admin` even though other things the repository admin does (e.g., manage users) would come under `/admin`. The line seems blurry so using both paths might be confusing.

The reason I prefer "administration" to refer to anything the user does on the non-public side is because the term "dashboard" in interaction design has had a more specific meaning, to refer to quick displays of critical information about a website or application:

> "The term “dashboard” is a metaphor for a car dashboard, which provides essential information that the driver can absorb quickly, without having to think." 
[link](https://www.nngroup.com/articles/dashboards-preattentive/)

Our "Activity Summary" page is what has more commonly been called a "dashboard." So calling the whole admin side a dashboard conflates the specific dashboard page with all the other administrative features. That's what I don't feel comfortable with. On the other hand, in recent years I think it has become somewhat more common to refer to user administration UIs as the "dashboard." So I don't really like it, but I also accept that the term has possibly evolved a more general meaning and thus don't feel like I can make too strong an argument based on the traditional meaning of a "dashboard."

elrayle: @ggeisler I see your point about the term dashboard.  I'm wondering if this should be reviewed as part of a larger refactoring of the dashboard/admin features.  I would not want to hold up this PR for something that is a larger discussion and potentially work for another sprint.

ggeisler: That sounds fine to me @elrayle, although given I'm not involved in the technical details associated with this the opinions of @jcoyne and @mjgiarlo matter much more than mine.

app/controllers/concerns/hyrax/dashboard/collections_controller_behavior.rb
jcoyne: I think this would make the refactor much less complex as you'd move code to an existing file rather than creating a new module. We aren't using prepend in Hyrax itself, but you can use it to inject behavior in your local app.  It's a core language feature available since ruby 2.0 http://ruby-doc.org/core-2.0.0/Module.html#method-i-prepend

atz: You don't need an example because `#prepend` is not what Hyrax would use, it is what a consumer would use to get the effect they previously might need the separate Behavior module for.  Our stack has been counterproductively indirect with Behavior modules that are used only once and still tightly bound to the context of the consuming class.  That is to say, they are not modular modules.  Basically, all you need to do is move the guts back into the proper Controller.

elrayle:
@jcoyne @mjgiarlo I rebased with latest changes.  Tests have completed.  @ggeisler defers to you.  Can you clear this for merge?

mjgiarlo:
@elrayle :white_check_mark: :tada:

mjgiarlo:
@jcoyne Given the discussion above, I'm inclined to push this through. If you feel *strongly* about any of the changes you suggested, could you list them here? I will work with @elrayle today on those, if there are any. Else, I'm inclined to unblock the Collection Extensions folks.

mjgiarlo:
@elrayle :policeman: found issues.

.rubocop_fixme.yml
jcoyne: Is this exemption still relevant?

elrayle:
I did one more refactor to move all code in the dashboard collection_controller_behavior module into the dashboard collection_controller (per @atz and @jcoyne) suggestion.  Had to repush to fix rubocop errors introduced by that refactor.  Tests running now.

@jcoyne Can you remove your block, such that this can be merged once the tests complete?

elrayle:
@jcoyne - Can you unblock?  This is ready to merge.

NOTE:  The failing test is not related to this PR.  It is one of the known errors that randomly occurs.

```
batch editing changes the value of each field for all selected works
     Failure/Error: expect(page).to have_content 'Changes Saved'
       expected to find text "Changes Saved" in "Creator Cancel". (However, . it was found 1 time including non-visible text.)
     # ./spec/support/features/batch_edit_actions.rb:15:in `block (2 levels) in fill_in_batch_edit_fields_and_verify!'
```

Issue #1341, opened 2017-07-14
Collection Types: Create controller for collection types


elrayle:
### Descriptive summary

Add code to implement the controller part of the Rails MVC for managing collection types.

### Rationale

See model issue for rationale.

### Expected behavior

Serve as the controller for collection type views limiting access as appropriate.

### Implementation

The contoller needs to handle...
* authentication and abilities enforcement
* define methods for controller actions
  - [ ] #index
  - [ ] #new
  - [ ] #create
  - [ ] #edit
  - [ ] #update
  - [ ] #destroy

#### General Notes

* values for descriptive metadata, settings, and participants are stored and read from the database
* participants will be handled using permission templates (I think - need more info - see Admin Sets implementation)

#### Notes on #create

* automatically assign the machine_id property using `new_collection_type.title.parameterize.underscore.to_sym`

#### Notes on #update action

* disallow changes to settings IF a collection of this type already exists
* for Admin Sets, don't allow any changes EXCEPT for participants and description.

#### Notes on #destroy action

* FAIL if collections of this type exist
* delete this collection type from the database

### Related work

| Collection Types feature | Issue |
| ------------------------ | ----- |
| Admin Set Generator      | #1342 |
| User Collection Generator | #1485  |
| Routes                   | #1347 |
| Controller               | this issue |
| Model                    | #1343 |
| View - Index             | #1349 |
| View -  New/Edit         | #1348 |
| Partial -  Description   | #1344 |
| Partial -  Settings      | #1346 |
| Partial -  Participants  | #1345 |

hortongn:
FYI, I'm creating a shell of CollectionTypesController as part of #1347 so I have a controller to point to in the routing specs.

gkostin1966:
First Iteration: Skeleton Controller

elrayle:
NOTE: For create, there needs to be a validation that the title, and machine_id generated from the title, are unique.

elrayle:
Probably needs to be done in the model with `validate :machine_id, uniqueness: true`

hortongn:
PR merged, but leaving this issue open since there's more to do here.

elrayle:
Work remaining after PR #1542 is being addressed by...

* #1345 anything dealing with participants
* #1547 - for Admin Sets, don't allow any changes EXCEPT for participants and description.
* #1547 - Saving a collection type should allow some changes even when collections exist

PR #1504, opened 2017-08-16
Build out basic controller

cjcolvar:
refs #1341

Builds out the basic CrUD functionality of the collection type controller as well as some refactoring to DRY things up along with more testing based upon the rspec controller template.  This leaves the notes about special handling still to be handled in another PR.

```
General Notes

- participants will be handled using permission templates (I think - need more info - see Admin Sets implementation)

Notes on #create

- automatically assign the machine_id property using -
- new_collection_type.title.parameterize.underscore.to_sym

Notes on #update action

- disallow changes to settings IF a collection of this type already exists
- for Admin Sets, don't allow any changes EXCEPT for participants and description.

Notes on #destroy action

- FAIL if collections of this type exist
- delete this collection type from the database
```

### DIFF
--- a/app/controllers/hyrax/admin/collection_types_controller.rb
+++ b/app/controllers/hyrax/admin/collection_types_controller.rb
@@ -3,6 +3,7 @@ module Hyrax
     before_action do
       authorize! :manage, :collection_types
     end
+    before_action :set_collection_type, only: [:edit, :update, :destroy]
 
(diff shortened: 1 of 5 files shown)


elrayle:
FYI... I added the machine_id generation because save doesn't work without it.  I also added :description to the acceptable params because the form save was rejected in the UI without it.

PR #1516, opened 2017-08-18
Injecting progressive sleep into find elements

jeremyf:
In Notre Dame's extensive Capybara usage for our [QA_tests][1]
ecosystem, we have discovered intermittent failures of Capybara
expectations even when we have very large
`Capybara.max_default_wait_time` values.

Our solution has been to auto-inject sleep behavior into each of the
Capybara matchers ([see examples][2]). We have found that this behavior
eliminates most all of our unresponsive javascript element not found
errors.

In doing this, I've moved from `expect(page).to have_css` to the
simplified `page.find`. Note the outward behavior is still somewhat
similar (instead of an RSpec failure message we would instead get an
RSpec exception message).

If desired, I can refactor to auto-inject all of the sleep injectors
that we have included.

[1]:https://github.com/ndlib/qa_tests
[2]:https://github.com/ndlib/qa_tests

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/batch_edit_spec.rb
+++ b/spec/features/batch_edit_spec.rb
@@ -25,30 +25,40 @@ RSpec.describe 'batch', type: :feature, clean_repo: true, js: true do
       # Reload the form and verify
       visit '/dashboard/my/works'
       check 'check_all'
-      click_on 'batch-edit'
+      find('#batch-edit').click
       expect(page).to have_content('Batch Edit Descriptions')
-      batch_edit_expand("creator")
-      expect(page).to have_css "input#generic_work_creator[value*='NEW creator']"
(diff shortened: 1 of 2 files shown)


jeremyf:
Assuming this is accepted, I want to propose creating a gem that handles this sleep injection behavior (taking what we have in ND's QA_tests repository) and making it a shared resource.

jeremyf:
@jenlindner I'm working on a gem today!

mjgiarlo:
Thanks for the PR and the review. I put in a new PR that should supersede this one. The new PR uses some code from ScholarSphere (thanks, @awead) that auto-expands the batch edit accordion using Javascript that is injected in a middleware that is only used in the test application. See #1522

Issue #1376, opened 2017-07-24
Collections: Create branding metadata partial of the new/edit form


elrayle:
### Descriptive summary

Add code to implement the Branding tab partial for the new/edit form views for collections.  This partial is shared by both the new and edit form views.

### Rationale

See model issue for rationale.

### Implementation

**Files:**
* Edit [app/views/hyrax/collections/_form.html.erb](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/collections/_form.html.erb)
* Create [app/views/hyrax/collections/_form_branding.html.erb](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/collections/_form_branding.html.erb)

#### Expected user interactions

Add the tab to the form by editing _form.html.erb.

Create the UI view for uploading branding files for a collection by adding partial _branding.html.erb.

Able to upload files...
- [ ] Banner (one file)
  - [ ] ability to replace the banner file
  - [ ] ability to remove the banner file
- [ ] Logos (multiple files)
  - [ ] ability to remove each logo
  - [ ] ability to set link URL
  - [ ] ability to set alt text

Exact text for the page is in the mockup below.

NOTES:
* Need to verify that the files uploaded are JPG, GIF, or PNG.
* The mockup does not show using jquery uploader which allows for drag-n-drop.  That might be a nice bit to add.

NOTE: This includes changes to the [Collection controller](https://github.com/samvera/hyrax/blob/master/app/controllers/concerns/hyrax/collections_controller_behavior.rb) to process the new landing page values.  I'm not sure where the code is that actually does the processing for new/edit.  There is some documentation in the wiki identifying code locations for [Managing Collections](https://wiki.duraspace.org/x/1qA2BQ).  Also, there is additional information about [Permission Templates](https://wiki.duraspace.org/x/2aA2BQ) in the wiki.  I would expect the save process to be similar to that for [What happens when I deposit a work?](http://samvera.github.io/what-happens-deposit-2.0.html)

![question_24](https://cloud.githubusercontent.com/assets/6855473/13064236/f2f04cbe-d41e-11e5-9674-e9a56a6326e6.png) **QUESTION:** Should the new landing page information go in Fedora or the database?  If it becomes part of the collection object, then there are questions with regard to what predicates to use to associate the images with the collection object.

#### SPECIAL NOTE on Admin Sets:

For the initial implementaiton, the Admin Set underlying code will not change.  But to a user, the Admin Set will appear to be a collection.

[...]

PR #1518, opened 2017-08-18
collection branding

blancoj:
@elrayle I just pushed this over take a look.  I think it does most of the requirements from #1376. It also displays the banner on the collection page.


elrayle:
Closing... replaced by PR #1569

PR #1520, opened 2017-08-18
Connect Collections to Collection Type

### LINKED ISSUE #1486: Collections: Set the collection_type instance variable for the Collection
### Descriptive summary

All instances, new and resumed, should have an instance of Collection Type in the collection_type instance variable of the collection object based on the value of collection_type_gid property.  If the collection_type_gid property is nil, set the collection type to the default User Collection.

### Rationale

Keeping the instance of the Collection Type in the collection object allows for delegation of settings methods to the collection type.

Collection type may not be assigned for any collections that existed prior to this implementation.  There is an existing implementation of the Collections concept that is generally referred to as User Collections.  A default collection type of User Collection is being created by Issue #1485.  Any collection that does not have a collection type assigned will be assigned the User Collection the first time it is accessed.  This removes the need for an upfront migration process for collection types.
 
This work is expanding the existing Collections implementation to accommodate Collection Types and the new [requirements](https://wiki.duraspace.org/x/g402BQ) that are part of the [Collection Extensions](https://wiki.duraspace.org/x/5JM2BQ) work.

### Implementation

**Files:** 
* [app/models/concerns/hyrax/collection_behavior.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/concerns/hyrax/collection_behavior.rb)
* [app/controllers/concerns/hyrax/collections_controller_behavior.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/concerns/hyrax/collections_controller_behavior.rb)
* [app/models/hyrax/collection_type.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb)

When a collection is read back in, an instance of the collection_type is automatically set in the collection_type instance variable from the gid stored in the collection_type_gid property.  If the gid is nil, then use the default User Collection collection type created by Issue #1485.

#### In [collections controller](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/concerns/hyrax/collections_controller_behavior.rb):

For all actions...
```
before_action :load_collection_type_instance
```

With the method definition...
```
      def load_collection_type
        # TODO: For new collections, the type comes from the modal OR uses the one and only one type the user can create.  This TODO will be removed when the processing for modals is in place.
        if @collection.collection_type_gid.nil?
          # gid will only be nil for non-migrated collections created before collection types existed
          @collection.collection_type = Hyrax::CollectionType.default_type
          @collection.collection_type_gid = @collection.collection_type.gid
          # TODO: Question -- should we save here to persist the gid setting?
        else
          @collection.collection_type = Hyrax::CollectionType.find_by(machine_id: GlobalID.new(RDF::URI.new(@collection.collection_type_gid)).model_id)
        end
      end
```

#### In [collection_type model](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb):

```
      def self.default_type
        # TODO Return generated User Collection.  See Issue #1485.
        find_by_machine_id('user_collection')
      end

      def self.admin_set_type
        # TODO Return generated Admin Set.  See Issue #1342.
        find_by_machine_id('admin_set')
      end

      def self.find_by_machine_id
        # TODO Create an instance of CollectionTypes given the machine_id
      end

      def admin_set?
        machine_id == 'admin_set'
      end
```

NOTE: Not sure we need a method to instantiate the admin_set collection_type. We do need a way to tell if a collection_type is the admin_set.

### Related work

Generator for User Collection #1485 
Generator for Admin Set #1342 

| Model                    | #1383 |

elrayle:
Fixes #1486

Changes to Hyrax::CollectionBehavior model module
* add load_collection_type in collection model which...
  * loads the collection type based on the gid in the collection model
  * if a gid isn't in the collection model, loads the default collection type
* add collection_type reader that loads the collection_type the first time it is accessed if it is not already set
* add protected collection_type writer
    
Changes to Hyrax::CollectionType model
* add method that generates gid from model id
* add method to find a type by gid


elrayle:
Set up against wrong branch.

PR #1521, opened 2017-08-18
Connect Collections to Collection Type

### LINKED ISSUE #1486: Collections: Set the collection_type instance variable for the Collection

elrayle:
Fixes #1486

Changes to Hyrax::CollectionBehavior model module
* add load_collection_type in collection model which...
  * loads the collection type based on the gid in the collection model
  * if a gid isn't in the collection model, loads the default collection type
* add collection_type reader that loads the collection_type the first time it is accessed if it is not already set
* add protected collection_type writer
    
Changes to Hyrax::CollectionType model
* add method that generates gid from model id
* add method to find a type by gid

### DIFF
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -16,12 +16,21 @@ module Hyrax
       self.indexer = Hyrax::CollectionIndexer
 
       property :collection_type_gid, predicate: ::RDF::Vocab::SCHEMA.additionalType, multiple: false do |index|
-        index.as :symbol, :stored_searchable
+        index.as :symbol
       end
+
+      after_find { |col| load_collection_type_instance(col) }
+      # before_update { |col| validate_collection_type_gid(col) } # TODO: check that gid is not nil AND check that gid has not changed since it was read in (It is ok to go from nil to a value.)
     end
 
-    attr_accessor :collection_type
     delegate :nestable?, :discoverable?, :sharable?, :allow_multiple_membership?, :require_membership?, :assigns_workflow?, :assigns_visibility?, to: :collection_type
+
+    # Get (and set) the collection_type when accessed
+    def collection_type
+      return @collection_type if @collection_type
+      return nil if collection_type_gid.nil?
+      @collection_type = Hyrax::CollectionType.find_by_gid!(collection_type_gid)
+    end
 
     # Add members using the members association.
     def add_members(new_member_ids)
@@ -78,7 +87,30 @@ module Hyrax
       ActiveFedora::Base.search_with_conditions("member_of_collection_ids_ssim:#{id}").map(&:id)
     end
 
+    protected
+
+      def collection_type=(collection_type)
+        unless collection_type && collection_type.is_a?(Hyrax::CollectionType) && collection_type.persisted?
+          raise ActiveRecord::AssociationTypeMismatch, 'Collection type must be a valid Hyrax::CollectionType'
+        end
+        @collection_type = collection_type
+      end
+
     private
+
+      # Load the collection_type attribute with an instance of the collection type of this collection based on the gid stored in the
+      # collection object model.  Defaults to the default collection type (i.e. user_collection) so that all collections existing
+      # before the addition of collection types do not need to be migrated.  They are assumed to be the default.
+      # @param collection [Collection] an instance of the Collection model
+      def load_collection_type_instance(collection)
+        if collection.collection_type_gid.nil?
+          collection.collection_type = Hyrax::CollectionType.find_or_create_default_collection_type
+          collection.collection_type_gid = collection.collection_type.gid
+          collection.save # do this one time on the first read by saving the gid in the collection object
+        else
+          collection.collection_type = Hyrax::CollectionType.find_by_gid!(collection.collection_type_gid)
+        end
+      end
 
       # Calculate the size of all the files in the work
       # @param work_id [String] identifer for a work
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -22,8 +22,34 @@ module Hyrax
     alias_attribute :visibility, :assigns_visibility
     deprecation_deprecate visibility: "prefer #assigns_visibility instead"
 
+    # Find the collection type associated with the Global Identifier (gid)
+    # @param [String] gid - Global Identifier for this collection_type (e.g. gid://internal/hyrax-collectiontype/3)
+    # @return [Hyrax::CollectionType] if record matching gid is found, an instance of Hyrax::CollectionType with id = the model_id portion of the gid (e.g. 3)
+    # @return [False] if record matching gid is not found
+    def self.find_by_gid(gid)
+      find(GlobalID.new(gid).model_id)
+    rescue ActiveRecord::RecordNotFound
+      false
+    end
+
+    # Find the collection type associated with the Global Identifier (gid)
+    # @param [String] gid - Global Identifier for this collection_type (e.g. gid://internal/hyrax-collectiontype/3)
+    # @return [Hyrax::CollectionType] an instance of Hyrax::CollectionType with id = the model_id portion of the gid (e.g. 3)
+    # @raise [ActiveRecord::RecordNotFound] if record matching gid is not found
+    def self.find_by_gid!(gid)
+      find(GlobalID.new(gid).model_id)
+    rescue ActiveRecord::RecordNotFound
+      raise ActiveRecord::RecordNotFound, "Couldn't find Hyrax::CollectionType matching GID '#{gid}'"
+    end
+
+    # Return the Global Identifier for this collection type.
+    # @return [String] Global Identifier (gid) for this collection_type (e.g. gid://internal/hyrax-collectiontype/3)
     def gid
-      URI::GID.build app: GlobalID.app, model_name: model_name.name.parameterize.to_sym, model_id: id unless id.nil?
+      return nil if id.nil?
+      # TODO: From a modeling perspective, I'm not sure which of these we should make the <object_value>.  For now, we will store in Fedora object as string, but might want to revisit.
+      # rdf_uri = RDF::URI.new(uri_gid) # TODO: Decide if this should return RDF::URI - ActiveTriples converts this to an ActiveTriples relationship that is a new triple with <subject> = gid uri
+      uri_gid = URI::GID.build(app: GlobalID.app, model_name: model_name.name.parameterize.to_sym, model_id: id) unless id.nil? # ActiveTriples won't accept this as an <object_value>
+      uri_gid.to_s # ActiveTriples treats this string version as a literal <object_value>
     end
 
     def collections?
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -120,9 +120,89 @@ RSpec.describe Collection do
   describe '#collection_type_gid' do
     it 'has a collection_type_gid' do
       subject.title = ['title']
-      subject.collection_type_gid = 'gid://internal/CollectionType/5'
+      subject.collection_type_gid = 'gid://internal/hyrax-collectiontype/5'
       subject.save!
-      expect(subject.reload.collection_type_gid).to eq 'gid://internal/CollectionType/5'
+      expect(subject.reload.collection_type_gid).to eq 'gid://internal/hyrax-collectiontype/5'
+    end
+  end
+
+  describe '#load_collection_type_instance' do
+    context 'when gid exists in collection object' do
+      let(:collection) { described_class.new(title: ['title']) }
+      let(:collection_type) { Hyrax::CollectionType.new(id: 5) }
+
+      before do
+        allow(Hyrax::CollectionType).to receive(:find_by_gid!).with('gid://internal/hyrax-collectiontype/5').and_return(collection_type)
+        allow(collection_type).to receive(:persisted?).and_return(true)
+      end
+
+      it 'loads instance of collection type based on gid' do
+        collection.collection_type_gid = 'gid://internal/hyrax-collectiontype/5'
+        collection.save!
+        col = described_class.find(collection.id)
+        expect(col.collection_type).to be_a Hyrax::CollectionType
+        expect(col.collection_type.id).to eq 5
+      end
+    end
+
+    context 'when gid in collection object is nil' do
+      let(:collection) { described_class.new(title: ['title']) }
+
+      before do
+        allow(collection).to receive(:collection_type_gid).and_return(nil)
+      end
+
+      it 'loads default collection type' do
+        collection.save!
+        col = described_class.find(collection.id)
+        expect(col.collection_type).to be_a Hyrax::CollectionType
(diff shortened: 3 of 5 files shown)


app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -16,13 +16,24 @@ module CollectionBehavior
...
     end
 
-    attr_accessor :collection_type
     delegate :nestable?, :discoverable?, :sharable?, :allow_multiple_membership?, :require_membership?, :assigns_workflow?, :assigns_visibility?, to: :collection_type
 
+    # Get (and set) the collection_type when accessed
+    def collection_type
+      return @collection_type if @collection_type
+      return nil if collection_type_gid.nil?
+      collection_type = Hyrax::CollectionType.find_by_gid(collection_type_gid)
```
jeremyf: I believe it is preferable to replaces lines 32 to 34 with `@collection_type = Hyrax::CollectionType.find_by_gid!`. (Have `.find_by_gid!` raise the consistent exception)

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -78,8 +89,39 @@ def member_object_ids
       ActiveFedora::Base.search_with_conditions("member_of_collection_ids_ssim:#{id}").map(&:id)
     end
 
+    protected
+
+      def collection_type=(collection_type)
+        @collection_type = collection_type
```
jeremyf: I believe this should raise a `Runtime` exception if the given `collection_type` is not a Hyrax::CollectionType? Or an `ActiveRecord::AssociationTypeMismatch` error? It appears you could move the `collection_type_valid?` method logic into this method.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -78,8 +89,39 @@ def member_object_ids
...
+      # before the addition of collection types do not need to be migrated.  They are assumed to be the default.
+      # @param collection [Collection] an instance of the Collection model
+      def load_collection_type_instance(collection)
+        if collection.collection_type_gid.nil?
+          collection.collection_type = Hyrax::CollectionType.find_or_create_default_collection_type
+          collection.collection_type_gid = collection.collection_type.gid
+          collection.save # do this one time on the first read by saving the gid in the collection object
+        else
+          collection.collection_type = Hyrax::CollectionType.find_by_gid(collection.collection_type_gid)
+          # TODO: What to do if collection_type isn't valid?
```
jeremyf: If it's not valid, I believe we should raise an uncaught exception. This is a data integrity issue.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -78,8 +89,39 @@ def member_object_ids
...
+          logger.warn("WARNING: Failed to load collection type instance.  Collection type is invalid.") unless collection_type_valid?(collection)
+        end
+      end
+
+      # Validate that the collection has a valid instance of the collection type.
+      # @param collection [Collection] an instance of the Collection model
+      # @return true if valid; otherwise, false
+      def collection_type_valid?(collection)
+        return false if collection.collection_type.nil?
```
jeremyf: Should the collection's collection_type be checked for persistence? I believe that appears to be a requirement.

app/models/hyrax/collection_type.rb
```diff
@@ -22,8 +22,23 @@ class CollectionType < ActiveRecord::Base
     alias_attribute :visibility, :assigns_visibility
     deprecation_deprecate visibility: "prefer #assigns_visibility instead"
 
+    # Find the collection type associated with the Global Identifier (gid)
+    # @param [String] gid - Global Identifier for this collection_type (e.g. gid://internal/hyrax-collectiontype/3)
+    # @return [Hyrax::CollectionType] an instance of Hyrax::CollectionType with id = the model_id portion of the gid (e.g. 3), or nil if gid not found
+    def self.find_by_gid(gid)
```
jeremyf: It would be helpful to have both a `find_by_gid` and a `find_by_gid!`.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -78,8 +89,39 @@ def member_object_ids
...
+      # collection object model.  Defaults to the default collection type (i.e. user_collection) so that all collections existing
+      # before the addition of collection types do not need to be migrated.  They are assumed to be the default.
+      # @param collection [Collection] an instance of the Collection model
+      def load_collection_type_instance(collection)
+        if collection.collection_type_gid.nil?
+          collection.collection_type = Hyrax::CollectionType.find_or_create_default_collection_type
+          collection.collection_type_gid = collection.collection_type.gid
+          collection.save # do this one time on the first read by saving the gid in the collection object
+        else
+          collection.collection_type = Hyrax::CollectionType.find_by_gid(collection.collection_type_gid)
```
jeremyf: Prefer self.class.find_by_gid! instead of Hyrax::CollectionType.find_by_gid there could be subclassing or extension which could cause unexpected behavior.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -78,8 +89,39 @@ def member_object_ids
...
+
     private
 
+      # Load the collection_type attribute with an instance of the collection type of this collection based on the gid stored in the
+      # collection object model.  Defaults to the default collection type (i.e. user_collection) so that all collections existing
+      # before the addition of collection types do not need to be migrated.  They are assumed to be the default.
+      # @param collection [Collection] an instance of the Collection model
+      def load_collection_type_instance(collection)
+        if collection.collection_type_gid.nil?
+          collection.collection_type = Hyrax::CollectionType.find_or_create_default_collection_type
```
jeremyf: Prefer `self.class.find_by_gid!` instead of `Hyrax::CollectionType.find_by_gid` there could be subclassing or extension which could cause unexpected behavior.

app/models/hyrax/collection_type.rb
```diff
@@ -22,8 +22,23 @@ class CollectionType < ActiveRecord::Base
...
+    end
+
+    # Return the Global Identifier for this collection type.
+    # @return [String] Global Identifier (gid) for this collection_type (e.g. gid://internal/hyrax-collectiontype/3)
     def gid
-      URI::GID.build app: GlobalID.app, model_name: model_name.name.parameterize.to_sym, model_id: id unless id.nil?
+      return nil if id.nil?
+      # TODO: From a modeling perspective, I'm not sure which of these we should make the <object_value>.  For now, we will store in Fedora object as string, but might want to revisit.
+      uri_gid = URI::GID.build(app: GlobalID.app, model_name: model_name.name.parameterize.to_sym, model_id: id) unless id.nil? # ActiveTriples won't accept this as an <object_value>
+      # rdf_uri = RDF::URI.new(uri_gid) # ActiveTriples converts this to an ActiveTriples relationship that is a new triple with <subject> = gid uri
```
jeremyf: This appears to be commented out code. Remove it?

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -78,8 +89,39 @@ def member_object_ids
...
+          collection.collection_type = Hyrax::CollectionType.find_by_gid(collection.collection_type_gid)
+          # TODO: What to do if collection_type isn't valid?
+          logger.warn("WARNING: Failed to load collection type instance.  Collection type is invalid.") unless collection_type_valid?(collection)
+        end
+      end
+
+      # Validate that the collection has a valid instance of the collection type.
+      # @param collection [Collection] an instance of the Collection model
+      # @return true if valid; otherwise, false
+      def collection_type_valid?(collection)
```
jeremyf: The method name implies that you are validating a `collection_type` parameter, and not a collection. Consider either adding a named parameter to clarify, or replacing the parameters from the caller.

app/models/concerns/hyrax/collection_behavior.rb
elrayle: I had it there and then moved it out.  And now it's back again.  It makes more sense.  Do you have a suggestion for the exception to throw?  I catch ActiveRecord::RecordNotFound.  I could stay with that exception and just change the message to reference gid instead of id.

app/models/concerns/hyrax/collection_behavior.rb
elrayle: The `Hyrax::CollectionBehavior` gets included in the `Collection` class.  This is using `Hyrax::CollectionType` class to get the default collection type and to find_by_gid.  I'm not sure your comment applies as `self.class` would evaluate to `Collection.find_by_gid` which doesn't have that method.

app/models/concerns/hyrax/collection_behavior.rb
elrayle: same as above

app/models/concerns/hyrax/collection_behavior.rb
elrayle: Now that `find_by_gid` raises an exception, I don't think this applies.  An exception will be raised if the gid is invalid.

app/models/concerns/hyrax/collection_behavior.rb
elrayle: This method was removed and the validation happens in the `collection_type=` method.

app/models/hyrax/collection_type.rb
elrayle: It is there associated with the TODO question to document how to generate an RDF::URI version of the gid and what ActiveTriples does with it.

The way the code is written right now, the gid gets stored in Fedora as a string with format `'gid://internal/hyrax-collectiontype/:id'`.  I'm not sure that is what we want from a modeling perspective.  But ActiveTriples doesn't seem too happy with it.  So I'm punting it to later. 

I'll move it up one line, closer to the TODO and add a TODO marker to it.

app/models/concerns/hyrax/collection_behavior.rb
elrayle: I decided to go with ActiveRecord::RecordNotFound with the updated message.

app/models/concerns/hyrax/collection_behavior.rb
elrayle: I used ActiveRecord::AssociationTypeMismatch because it seems close in concept.  But there isn't a real associate between the two other than a runtime association.

elrayle:
@jeremyf Your requested changes are complete.  PR ready for review.

app/models/concerns/hyrax/collection_behavior.rb
elrayle: I did change from find_by_gid to find_by_gid!

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -21,18 +21,18 @@
...
 
-    it 'returns the gid when id is nil' do
+    it 'returns nil when id is nil' do
       collection_type.id = nil
       expect(collection_type.gid).to be_nil
     end
   end
 
-  describe ".find_or_create_default_collection_type" do
+  describe "#find_or_create_default_collection_type" do
```
jeremyf: This should be `. find_or_create_default_collection_type` (as it is a class method). Instance methods are documented as `#find_or_create_default_collection_type`.

spec/models/collection_spec.rb
```diff
@@ -120,9 +120,89 @@ class Member < ActiveFedora::Base
...
+      collection.collection_type_gid = gid99
+      expect(collection.collection_type).to be_kind_of(Hyrax::CollectionType)
+      expect(collection.collection_type.gid).to eq gid89
+    end
+
+    it 'throws StandardError if cannot find collection type for the gid' do
+      gid = 'gid://internal/hyrax-collectiontype/999'
+      collection.collection_type_gid = gid
+      # TODO: Should we capture the ActiveRecord error and produce something nicer?
+      expect { collection.collection_type }.to raise_error(StandardError, "Couldn't find Hyrax::CollectionType matching GID '#{gid}'")
```
jeremyf: I believe this should be an `ActiveRecord::RecordNotFound` error (which is a descendant of `StandardError`)

PR #1522, opened 2017-08-19
Auto-expand batch edit accordion in test environment

mjgiarlo:
Do this to avoid sleeping which slows down the test suite. Supersedes #1516

@samvera/hyrax-code-reviewers


jeremyf:
I believe PR #1537 provides a more robust response.

mjgiarlo:
@jeremyf 👍👏

Issue #1345, opened 2017-07-14
Collection Types: Create participants partial of the new/edit form


elrayle:
### Descriptive summary

Add code to implement the Participants tab partial for the new/edit form views for collection types.  This partial is shared by both the new and edit form views.

### Rationale

See model issue for rationale.

### Implementation

**Files:**
* app/views/hyrax/admin/collection_types/_form_participants.html.erb

#### Expected user interactions

Create the UI view for adding participants for a collection type.

The form fields allow the user to select a group or user and apply a participant role to that group/user.  The roles for collection types includes...
- [ ] Collection Manager
- [ ] Collection Creator

The form will also include a listing of groups/users assigned to each participant role.

Exact text for the page is in the mockup below.

#### SPECIAL NOTE on Admin Sets:

For the initial implementaiton, the Admin Set underlying code will not change.  But to a user, the Admin Set will appear to be a collection type.

To make the illusion happen...
* as part of the hyrax generation process, a collection type will be created for Administrative Sets and the id will be recorded in a variable that can be checked by this UI (see issue #1342 for Admin Set Generator)
* the initial values for these settings are set by the generator to be the set of admins
* just like other types, these values can be changed at any time

### Mockup

![new-collection-type-participants-tab](https://user-images.githubusercontent.com/101482/28341990-f5c646ba-6bca-11e7-8f2d-ca6ac363a07e.png)

### Related work

| Collection Types feature | Issue |
| ------------------------ | ----- |
| Admin Set Generator      | #1342 |
| User Collection Generator | #1485  |
| Routes                   | #1347 |
| Controller               | #1341 |
| Model                    | #1343 |
| View - Index             | #1349 |
| View -  New/Edit         | #1348 |
| Partial -  Description   | #1344 |
| Partial -  Settings      | #1346 |
| Partial -  Participants  | this issue |

elrayle:
@laritakr is pondering an alternative to this.  Here is my proposed implementation.

Make use of existing permission_templates infrastructure with some modifications...
* Modify permission_templates table to...
  * add column source_type with values (admin_set, collection_type, collection)
  * rename column admin_set_id to source_id and it will hold the object id of the admin set, collection type, or collection
* permission_templates table is unchanged

Impact on admin_sets
* Code that queries the permission_templates table may need to update queries to check `source_type=='admin_set'`.  If they only access by id, then this won't be an issue.

Impact on collection_type
[...]

PR #1542, opened 2017-08-22
Generate machine_id from title; Check for collections before save/destroy

cjcolvar:
Relates to #1341.

Handles the rest of #1341 except the following:
- #1345 anything dealing with participants
- #1547 - for Admin Sets, don't allow any changes EXCEPT for participants and description.
- #1547 - Saving a collection type should allow some changes even when collections exist

### DIFF
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -3,9 +3,16 @@ module Hyrax
     self.table_name = 'hyrax_collection_types'
     validates :title, presence: true, uniqueness: true
     validates :machine_id, presence: true, uniqueness: true
+    before_save :ensure_no_collections
+    before_destroy :ensure_no_collections
 
     DEFAULT_ID = 'user_collection'.freeze
     DEFAULT_TITLE = 'User Collection'.freeze
+
+    def title=(value)
+      super
+      assign_machine_id
+    end
 
     # These are provided as a convenience method based on prior design discussions.
     # The deprecations are added to allow upstream developers to continue with what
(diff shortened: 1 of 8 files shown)


spec/models/hyrax/collection_type_spec.rb
```diff
@@ -56,28 +57,86 @@
...
+      expect(Collection.count > 0).to be_truthy
+      expect(build(:collection_type).collections).to eq []
+    end
+  end
+
+  describe "collections?" do
+    let(:collection_type) { create(:collection_type) }
+
+    it 'returns true if there are any collections of this collection type' do
+      create(:collection, collection_type_gid: collection_type.gid.to_s)
```
elrayle: `collection_type.gid` returns a string, so `to_s` is not needed

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -56,28 +57,86 @@
...
+  end
+
+  describe "save" do
+    before do
+      allow(collection_type).to receive(:collections?).and_return(true)
+    end
+
+    it "fails if collections exist of this type" do
+      expect(collection_type.save).to be_falsey
+      expect(collection_type.errors).not_to be_empty
```
elrayle: This is too strict.  Should only fail to save Settings.

app/models/hyrax/collection_type.rb
```diff
@@ -3,10 +3,17 @@ class CollectionType < ActiveRecord::Base
     self.table_name = 'hyrax_collection_types'
     validates :title, presence: true, uniqueness: true
     validates :machine_id, presence: true, uniqueness: true
+    before_save :ensure_no_collections
+    before_destroy :ensure_no_collections
```
elrayle: To be clear, the `before_destroy` check is good as is.

elrayle:
NOTE: Merging even though there is a request to loosen the restrictions on changes during save.  See Issue #1547 for more information.

app/models/hyrax/collection_type.rb
cjcolvar: I can't recall, but I think I was doing this as first pass.  So I think the follow on ticket you wrote to be 💯 .

spec/models/hyrax/collection_type_spec.rb
cjcolvar: 👍 I think this may have changed in one of my rebases and I missed catching it.

PR #1548, opened 2017-08-22
Remove outdated task from template

mjgiarlo:
One no longer needs to manually load workflows; workflows are automatically loaded when an Admin Set is created.

@samvera/hyrax-code-reviewers


mjgiarlo:
Failure is unrelated to work in this PR. Merging.

PR #1556, opened 2017-08-24
fixes 1460 - do not display Discovery tab if discoverable? false

### LINKED ISSUE #1460: Collections: collections with collection_type.discoverable? false are always Private

elrayle:
Fixes #1460

Work in this PR:
* add collection_type_gid to permitted terms for collection form
* add collection_type_gid assignment to collection factory
* add traits for settings on/off to collection type factory
* force visibility=private in the controller for create and update when discoverable? false
* do not show Discovery tab in collection edit form when discoverable? false
* do not show Sharing tab in collection edit form when sharable? false

### DIFF
--- a/spec/controllers/hyrax/dashboard/collections_controller_spec.rb
+++ b/spec/controllers/hyrax/dashboard/collections_controller_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe Hyrax::Dashboard::CollectionsController do
+RSpec.describe Hyrax::Dashboard::CollectionsController, :clean_repo do
   routes { Hyrax::Engine.routes }
   let(:user)  { create(:user) }
   let(:other) { build(:user) }
@@ -79,6 +79,28 @@ RSpec.describe Hyrax::Dashboard::CollectionsController do
         expect(asset_results["response"]["numFound"]).to eq 1
         doc = asset_results["response"]["docs"].first
         expect(doc["id"]).to eq asset1.id
+      end
+    end
+
+    context 'when setting collection type' do
+      let(:collection_type) { create(:collection_type) }
+
+      it "creates a Collection of default type when type is nil" do
+        expect do
+          post :create, params: {
+            collection: collection_attrs
+          }
+        end.to change { Collection.count }.by(1)
+        expect(assigns[:collection].collection_type.machine_id).to eq Hyrax::CollectionType::USER_COLLECTION_MACHINE_ID
+      end
+
+      it "creates a Collection of specified type" do
+        expect do
+          post :create, params: {
+            collection: collection_attrs.merge(collection_type_gid: collection_type.gid)
+          }
+        end.to change { Collection.count }.by(1)
+        expect(assigns[:collection].collection_type_gid).to eq collection_type.gid
       end
     end
 
@@ -182,7 +204,7 @@ RSpec.describe Hyrax::Dashboard::CollectionsController do
     end
 
     context "when update fails" do
-      let(:collection) { Collection.new(id: '12345') }
+      let(:collection) { create(:collection, id: '12345') }
       let(:repository) { instance_double(Blacklight::Solr::Repository, search: result) }
       let(:result) { double(documents: []) }
 
--- a/spec/factories/collection_types.rb
+++ b/spec/factories/collection_types.rb
@@ -1,7 +1,8 @@
 FactoryGirl.define do
   factory :collection_type, class: Hyrax::CollectionType do
-    sequence(:id) { |n| n }
+    sequence(:id) { |n| format("%010d", n) }
     sequence(:title) { |n| "Title #{n}" }
+
     description 'Collection type with all options'
     nestable true
     discoverable true
@@ -14,6 +15,38 @@ FactoryGirl.define do
     factory :user_collection_type do
       title 'User Collection'
       description 'A user oriented collection type'
+    end
+
+    trait :nestable do
+      nestable true
+    end
+
+    trait :not_nestable do
+      nestable false
+    end
+
+    trait :discoverable do
+      discoverable true
+    end
+
+    trait :not_discoverable do
+      discoverable false
+    end
+
+    trait :sharable do
+      sharable true
+    end
+
+    trait :not_sharable do
+      sharable false
+    end
+
+    trait :allow_multiple_membership do
+      allow_multiple_membership true
+    end
+
+    trait :not_allow_multiple_membership do
+      allow_multiple_membership false
     end
   end
 end
--- a/spec/features/dashboard/collection_spec.rb
+++ b/spec/features/dashboard/collection_spec.rb
@@ -203,7 +203,7 @@ RSpec.describe 'collection', type: :feature do
         find('button.dropdown-toggle').click
         click_link('Edit Collection')
       end
-      # URL: /collections/collection-id/edit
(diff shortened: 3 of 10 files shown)


spec/factories/collection_types.rb
```diff
@@ -15,5 +16,37 @@
...
+    trait :not_sharable do
+      sharable false
+    end
+
+    trait :allow_multiple_membership do
+      allow_multiple_membership true
+    end
+
+    trait :not_allow_multiple_membership do
+      allow_multiple_membership false
```
mjgiarlo: I'll want to use this in my WIP PR. Great!

spec/features/dashboard/collection_spec.rb
```diff
@@ -226,6 +226,36 @@
...
+    context 'with discoverable set' do
+      let(:discoverable_collection_id) { create(:collection, user: user, collection_type_settings: [:discoverable]).id }
+      let(:not_discoverable_collection_id) { create(:collection, user: user, collection_type_settings: [:not_discoverable]).id }
+
+      it 'to true, it shows Discovery tab' do
+        visit "/dashboard/collections/#{discoverable_collection_id}"
+        expect(page).to have_link('Discovery', href: '#discovery')
+      end
+
+      it 'to false, it shows Discovery tab' do
```
mjgiarlo: `s/shows/hides/`?

spec/features/dashboard/collection_spec.rb
```diff
@@ -226,6 +226,36 @@
...
+        visit "/dashboard/collections/#{not_discoverable_collection_id}"
+        expect(page).not_to have_link('Discovery', href: '#discovery')
+      end
+    end
+
+    context 'with sharable set' do
+      let(:sharable_collection_id) { create(:collection, user: user, collection_type_settings: [:sharable]).id }
+      let(:not_sharable_collection_id) { create(:collection, user: user, collection_type_settings: [:not_sharable]).id }
+
+      it 'to true, it shows Discovery tab' do
```
mjgiarlo: `s/Discovery/Sharable/` ?

spec/features/dashboard/collection_spec.rb
```diff
@@ -226,6 +226,36 @@
...
+        visit "/dashboard/collections/#{sharable_collection_id}"
+        expect(page).to have_link('Sharable', href: '#sharable')
+      end
+
+      it 'to false, it shows Discovery tab' do
```
mjgiarlo: `s/Discovery/Sharable/` ?

`s/shows/hides/` ?

spec/models/collection_spec.rb
```diff
@@ -146,7 +146,7 @@ class Member < ActiveFedora::Base
     end
 
     context 'when gid in collection object is nil' do
-      let(:collection) { create(:collection, title: ['title']) }
+      let(:collection) { create(:unmigrated_collection, title: ['title']) }
```
mjgiarlo: Can we rename this factory to `:typeless_collection`?

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe Hyrax::CollectionType, type: :model do
+RSpec.describe Hyrax::CollectionType, clean_repo: true, type: :model do
```
mjgiarlo: Did you add this to address remnants that cause problems with repeated test runs?

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -82,6 +82,28 @@
...
+        expect(assigns[:collection].collection_type.machine_id).to eq Hyrax::CollectionType::DEFAULT_ID
+      end
+
+      it "creates a Collection of specified type" do
+        expect do
+          post :create, params: {
+            collection: collection_attrs.merge(collection_type_gid: collection_type.gid)
+          }
+        end.to change { Collection.count }.by(1)
+        expect(assigns[:collection].collection_type_gid).to eq collection_type.gid
```
elrayle: In the previous test, I am specifically checking that the special 'user_collection' was created.  If I check the gid, it could be that the factory happened to create the collection_type with gid and I would never know it.  Where as in this second test, the factory should be creating a new type with gid.  I just need them to be the same.

spec/features/dashboard/collection_spec.rb
elrayle: doh

spec/models/collection_spec.rb
elrayle: unmigrated_collection is also aliased to untyped_collection.  That ok?

spec/models/hyrax/collection_type_spec.rb
elrayle: Yes, the clean_repo addresses the issue where if there are any collections of a type, then the type can't be saved.  Since the repo doesn't get cleaned between tests AND factory girl happily generates collection types with ids used previously, tests fail all over the place without the clean_repo.  Yuck.

spec/factories/collection_types.rb
```diff
@@ -1,7 +1,8 @@
 FactoryGirl.define do
   factory :collection_type, class: Hyrax::CollectionType do
-    sequence(:id) { |n| n }
-    sequence(:title) { |n| "Title #{n}" }
+    sequence(:id) { |n| format("%010d", n + 2) }
```
elrayle: Factory Girl starts the sequence at 1.  If the Admin Set and User Collection get created via engine cart generation process, there will already be collection types and the new collection type won't save because you can't change those.  I'm not sure what happens when the db is cleared between tests to thing generated before the tests run.

elrayle: Duh, this is not needed because the special types will be created in the dev db of the test app and not the test db.  I took this out and tests pass.

elrayle:
@mjgiarlo Requested changes in and tests passing.  Ready for review and merge.

PR #1558, opened 2017-08-24
Collection Types: Settings should be read only if any collections of a type exist

### LINKED ISSUE #1531: Collection Types: Settings should be read only if any collections of a type exist
### Descriptive summary

Make sure that in the collection type edit form's Settings tab, when a collection type is being edited AND there are collections of that type THEN all settings are readonly.

### Rationale

For at least the initial implementation, there will be no code that provides a migration path for existing collections to change behaviors associated with settings once the collection exists.  Therefore, we are disallowing changes to settings for collection types once a collection of that type is created.

### Expected behavior

When editing a collection type and viewing the Settings tab, all settings will be READONLY (i.e., disabled and grayed out) if there are any collections of that type.

### Implementation Note

[Hyrax::CollectionType](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb) #collections? method can be used to determine if collections exist for a type.

1. Dashboard -> Settings -> Collection Type -> Create collection type
  1. add metadata and save
  1. on Settings tab, note settings values
  1. save (may need to cancel to get back out to index)
1. Dashboard -> Settings -> Collection Type -> click Edit beside type created in first step
  1. click Settings tab
  1. change settings
  1. save (may need to cancel to get back out to index)

Verify you were able to change settings.

1. Dashboard -> Repository Contents -> Collections -> Create a collection -> in modal, select the type created in the first step
  1. add title and save
1. Dashboard -> Settings -> Collection Type -> click Edit beside type created in first step
  1. click Settings tab

Verify that all checkbox settings are disabled and grayed out.

### Dependency

#1506 Collections: Show modal for selecting collection type when New Collection
#1341 adding implementation of Hyrax::Collections #collections (list of all collections of a type) and #collections? (count of all collections of a type).  PR coming soon.

### Related Work

#1346 Collection Types: Create configuration settings partial of the new/edit form
#1530 Collection Types: Admin Set Settings should always be readonly

respinos:
Fixes #1531

If a collection type has collections, its settings fields should be read-only:

* CollectionType: limit before_save to prevent updates to settings
* force settings inputs to be disabled if collectiontype has collections
* fix collection type settings partial tests

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -113,28 +113,28 @@ module Hyrax
 
       def ensure_no_collections
         return true unless collections?
-        errors[:base] << I18n.t('hyrax.admin.collection_types.error_not_empty')
+        errors[:base] << I18n.t('hyrax.admin.collection_types.errors.not_empty')
         throw :abort
       end
 
       def ensure_no_settings_changes_for_admin_set_type
         return true unless admin_set? && exists_for_machine_id?(ADMIN_SET_MACHINE_ID)
         return true unless collection_type_settings_changed?
-        errors[:base] << I18n.t('hyrax.admin.collection_types.error_settings_changes')
+        errors[:base] << I18n.t('hyrax.admin.collection_types.errors.no_settings_change_for_admin_sets')
         throw :abort
       end
 
       def ensure_no_settings_changes_for_user_collection_type
         return true unless user_collection? && exists_for_machine_id?(USER_COLLECTION_MACHINE_ID)
         return true unless collection_type_settings_changed?
-        errors[:base] << I18n.t('hyrax.admin.collection_types.error_settings_changes')
+        errors[:base] << I18n.t('hyrax.admin.collection_types.errors.no_settings_change_for_user_collections')
         throw :abort
       end
 
       def ensure_no_settings_changes_if_collections_exist
         return true unless collections?
         return true unless collection_type_settings_changed?
-        errors[:base] << I18n.t('hyrax.admin.collection_types.error_settings_changes')
+        errors[:base] << I18n.t('hyrax.admin.collection_types.errors.no_settings_change_if_not_empty')
         throw :abort
       end
 
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -3,7 +3,9 @@ module Hyrax
     self.table_name = 'hyrax_collection_types'
     validates :title, presence: true, uniqueness: true
     validates :machine_id, presence: true, uniqueness: true
-    before_save :ensure_no_settings_changes
+    before_save :ensure_no_settings_changes_for_admin_set_type
+    before_save :ensure_no_settings_changes_for_user_collection_type
+    before_save :ensure_no_settings_changes_if_collections_exist
     before_destroy :ensure_no_collections
     has_many :collection_type_participants, class_name: 'Hyrax::CollectionTypeParticipant', foreign_key: 'hyrax_collection_type_id', dependent: :destroy
 
@@ -65,11 +67,7 @@ module Hyrax
     # Return the Global Identifier for this collection type.
     # @return [String] Global Identifier (gid) for this collection_type (e.g. gid://internal/hyrax-collectiontype/3)
     def gid
-      return nil if id.nil?
-      # TODO: From a modeling perspective, I'm not sure which of these we should make the <object_value>.  For now, we will store in Fedora object as string, but might want to revisit.
-      # rdf_uri = RDF::URI.new(uri_gid) # TODO: Decide if this should return RDF::URI - ActiveTriples converts this to an ActiveTriples relationship that is a new triple with <subject> = gid uri
-      uri_gid = URI::GID.build(app: GlobalID.app, model_name: model_name.name.parameterize.to_sym, model_id: id) unless id.nil? # ActiveTriples won't accept this as an <object_value>
-      uri_gid.to_s # ActiveTriples treats this string version as a literal <object_value>
+      URI::GID.build(app: GlobalID.app, model_name: model_name.name.parameterize.to_sym, model_id: id).to_s if id
     end
 
     def collections
@@ -119,17 +117,33 @@ module Hyrax
         throw :abort
       end
 
-      def ensure_no_settings_changes
-        return true unless collections?
-        return true unless collection_type_properties_changed?
+      def ensure_no_settings_changes_for_admin_set_type
+        return true unless admin_set? && exists_for_machine_id?(ADMIN_SET_MACHINE_ID)
+        return true unless collection_type_settings_changed?
         errors[:base] << I18n.t('hyrax.admin.collection_types.error_settings_changes')
         throw :abort
       end
 
-      def collection_type_properties_changed?
-        ['nestable', 'discoverable', 'sharable', 'allow_multiple_membership', 'require_membership', 'assigns_workflow', 'assigns_visibility'].any? do |key|
-          key.in?(changes)
-        end
+      def ensure_no_settings_changes_for_user_collection_type
+        return true unless user_collection? && exists_for_machine_id?(USER_COLLECTION_MACHINE_ID)
+        return true unless collection_type_settings_changed?
+        errors[:base] << I18n.t('hyrax.admin.collection_types.error_settings_changes')
+        throw :abort
+      end
+
+      def ensure_no_settings_changes_if_collections_exist
+        return true unless collections?
+        return true unless collection_type_settings_changed?
+        errors[:base] << I18n.t('hyrax.admin.collection_types.error_settings_changes')
+        throw :abort
+      end
+
+      def collection_type_settings_changed?
+        (changes.keys & ['nestable', 'discoverable', 'sharable', 'allow_multiple_membership', 'require_membership', 'assigns_workflow', 'assigns_visibility']).any?
+      end
+
+      def exists_for_machine_id?(machine_id)
+        Hyrax::CollectionType.where(machine_id: machine_id).exists?
       end
   end
 end
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -3,7 +3,7 @@ module Hyrax
     self.table_name = 'hyrax_collection_types'
     validates :title, presence: true, uniqueness: true
     validates :machine_id, presence: true, uniqueness: true
-    before_save :ensure_no_collections
+    before_save :ensure_no_settings_changes
     before_destroy :ensure_no_collections
     has_many :collection_type_participants, class_name: 'Hyrax::CollectionTypeParticipant', foreign_key: 'hyrax_collection_type_id', dependent: :destroy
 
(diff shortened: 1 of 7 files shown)


app/models/hyrax/collection_type.rb
```diff
@@ -109,5 +109,22 @@ def ensure_no_collections
         errors[:base] << I18n.t('hyrax.admin.collection_types.error_not_empty')
         throw :abort
       end
+
+      def ensure_no_settings_changes
+        return true unless collections?
+        return true unless
+            self.changes.has_key?('nestable') or
```
mjgiarlo: Would you consider refactoring this to look like the following?

```ruby
['nestable', 'discoverable', 'sharable', 'allow_multiple_membership', 'require_membership', 'assigns_workflow', 'assigns_visibility'].any? do |key|
  key.in?(changes)
end
```

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -129,13 +129,25 @@
     end
   end
 
+  describe "save (no settings changes)" do
+    before do
+      allow(collection_type).to receive(:collections?).and_return(true)
+    end
+
+    it "succeeds no changes to settings are being made" do
+      expect(collection_type.save).to be_truthy
```
mjgiarlo: You can make this even more reliable by changing `be_truthy` to `be true`

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -129,13 +129,25 @@
...
   describe "save" do
     before do
       allow(collection_type).to receive(:collections?).and_return(true)
+      allow(collection_type).to receive(:changes).and_return({'nestable' => false})
     end
 
-    it "fails if collections exist of this type" do
-      expect(collection_type.save).to eq false
+    it "fails if collections exist of this type and settings are changed" do
+      expect(collection_type.save).to be_falsey
```
mjgiarlo: Would `be false` also work here? (Sorry, I am a `truthy`/`falsey` hater!)

spec/views/hyrax/admin/admin_sets/_form.html.erb_spec.rb
```diff
@@ -2,6 +2,8 @@
   let(:admin_set) { stub_model(AdminSet) }
   let(:form) { Hyrax::Forms::AdminSetForm.new(admin_set, double, double) }
 
+  context ""
```
mjgiarlo: Cruft? TODO?

spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb
```diff
@@ -8,24 +21,47 @@
...
+    it "renders the intructions and warning" do
+      expect(rendered).to match(I18n.t("hyrax.admin.collection_types.form_settings.instructions"))
+      expect(rendered).to match(I18n.t("hyrax.admin.collection_types.form_settings.warning"))
+    end
+
+    INPUT_IDS.each do |id|
+      it "renders the #{id} checkbox to be disabled" do
+        match = rendered.match(/(<input.*id="#{id}".*>)/)
+        expect(match).not_to be_nil
+        next if match.nil?
```
mjgiarlo: If line 40 will fail is `match` is `nil`, do we need this `next` on line 41?

spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb
```diff
@@ -8,24 +21,47 @@
...
+      assign(:form, collection_type_form)
+      allow(view).to receive(:f).and_return(form)
+      render
+    end
+
+    INPUT_IDS.each do |id|
+      it "renders the #{id} checkbox to be disabled" do
+        match = rendered.match(/(<input.*id="#{id}".*>)/)
+        expect(match).not_to be_nil
+        next if match.nil?
```
mjgiarlo: If line 60 will fail if `match` is `nil`, do we need this `next` on line 61?

spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb
```diff
@@ -8,24 +21,47 @@
...
 
-  it "renders the checkboxes" do
-    expect(rendered).to have_selector("input#collection_type_nestable")
-    expect(rendered).to have_selector("input#collection_type_discoverable")
-    expect(rendered).to have_selector("input#collection_type_sharable")
-    expect(rendered).to have_selector("input#collection_type_require_membership")
-    expect(rendered).to have_selector("input#collection_type_allow_multiple_membership")
-    expect(rendered).to have_selector("input#collection_type_assigns_workflow")
-    expect(rendered).to have_selector("input#collection_type_assigns_visibility")
+  context "collection_type.collections? is true" do
```
mjgiarlo: Would you consider renaming the `context` to `when collection_type.collections? is true`? The rspec output reads more idiomatically when `context` blocks begin with e.g., `when` or `with`.

spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb
```diff
@@ -8,24 +21,47 @@
     end
   end
 
-  before do
-    assign(:form, collection_type_form)
-    allow(view).to receive(:f).and_return(form)
-    render
-  end
+  context "collection_type.collections? is false" do
```
mjgiarlo: Would you consider renaming the `context` to `when collection_type.collections? is false`? The rspec output reads more idiomatically when `context` blocks begin with e.g., `when` or `with`.

app/models/hyrax/collection_type.rb
```diff
@@ -109,5 +109,18 @@ def ensure_no_collections
         errors[:base] << I18n.t('hyrax.admin.collection_types.error_not_empty')
         throw :abort
       end
+
+      def ensure_no_settings_changes
+        return true unless collections?
+        return true unless
```
mjgiarlo: The indenting looks a bit funky here. I would consider reworking this as follows:

```ruby
def ensure_no_settings_changes
  # ...
  return true if collection_type_properties_changed?
  errors[:base] << ...
  # ...
end

def collection_type_properties_changed?
  ['nestable', 'discoverable', 'sharable', 'allow_multiple_membership', 'require_membership', 'assigns_workflow', 'assigns_visibility'].any? do |key|
     key.in?(changes)
  end
end
```

spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb
respinos: To avoid a Ruby error with the following line (`match[1]`) if `match` is indeed nil. May not be necessary?

spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb
respinos: OK.

app/models/hyrax/collection_type.rb
respinos: Sure. That's a `return true unless...`, though, right?

respinos: (That's what the spec says :-)

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -129,13 +129,25 @@
     end
   end
 
+  describe "save (no settings changes)" do
+    before do
+      allow(collection_type).to receive(:collections?).and_return(true)
+    end
+
+    it "succeeds no changes to settings are being made" do
+      expect(collection_type.save).to be true
```
cjcolvar: I think the preferred version of this is now `be_truthy`.

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -129,13 +129,25 @@
...
   describe "save" do
     before do
       allow(collection_type).to receive(:collections?).and_return(true)
+      allow(collection_type).to receive(:changes).and_return({'nestable' => false})
     end
 
-    it "fails if collections exist of this type" do
-      expect(collection_type.save).to eq false
+    it "fails if collections exist of this type and settings are changed" do
+      expect(collection_type.save).to be false
```
cjcolvar: I think the preferred version of this is now `be_falsey`.

app/models/hyrax/collection_type.rb
```diff
@@ -109,5 +109,19 @@ def ensure_no_collections
...
+
+      def ensure_no_settings_changes
+        return true unless collections?
+        return true unless collection_type_properties_changed?
+        errors[:base] << I18n.t('hyrax.admin.collection_types.error_settings_changes')
+        throw :abort
+      end
+
+      def collection_type_properties_changed?
+        ['nestable', 'discoverable', 'sharable', 'allow_multiple_membership', 'require_membership', 'assigns_workflow', 'assigns_visibility'].any? do |key|
```
cjcolvar: I think a slightly simpler version of this would be using the intersection instead of any?:
(property_keys & changes.keys).present?

spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb
mjgiarlo: @respinos Yeah, I think it is indeed unnecessary. You shouldn't ever get to `match[1]` if `match` is indeed `nil`.

app/models/hyrax/collection_type.rb
mjgiarlo: @cjcolvar Agree. That was similar to my most recent suggestion, but I fear that editing comments multiple times (as I am wont to do as I change my mind...) means that they're easy to miss. I might prefer `any?` to `present?` here, but I really dig the intersection approach.

elrayle:
@mjgiarlo @cjcolvar You want to take another look at this.  I added a second commit that does special processing for admin_sets and user_collections.

spec/models/hyrax/collection_type_spec.rb
```diff
@@ -142,13 +142,44 @@
...
+        allow(collection_type).to receive(:collections?).and_return(false)
+      end
+
+      it 'fails if settings are changed' do
+        expect(collection_type.save).to be false
+        expect(collection_type.errors).not_to be_empty
+      end
+    end
+
+    context 'for admin set collection type' do
```
cjcolvar: This should be `user collection type`.

app/models/hyrax/collection_type.rb
```diff
@@ -110,17 +108,33 @@ def ensure_no_collections
         throw :abort
       end
 
-      def ensure_no_settings_changes
+      def ensure_no_settings_changes_for_admin_set_type
+        return true unless admin_set? && exists_for_machine_id?(ADMIN_SET_MACHINE_ID)
```
cjcolvar: I don't quite understand what is being checked here.  Is this the same as `unless admin_set? && persisted?`?  Why do you need to check for existence because modifying admin_set settings is always forbidden, right?

If you don't need to check existence I think these three cases could be collapsed together.

elrayle:
Failure was randomly occurring known batch edit failure.

PR #1572, opened 2017-08-25
When batch upload fails due to multiple membership checking, notifications should include the reason why this happened

mjgiarlo:
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/operation.rb
+++ b/app/models/hyrax/operation.rb
@@ -41,6 +41,18 @@ module Hyrax
       end
     end
 
+    # Roll up messages for an operation and all of its children
+    def rollup_messages
+      [].tap do |messages|
+        messages << message if message.present?
+        if children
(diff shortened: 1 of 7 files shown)


app/models/hyrax/operation.rb
```diff
@@ -58,10 +70,10 @@ def success!
     # Mark this operation as a FAILURE. If this is a child operation, roll up to
     # the parent any failures.
     #
-    # @param [String, nil] message - record any failure message
+    # @param [String, nil] message record any failure message
     # @see Hyrax::Operation::FAILURE
     # @see #rollup_status
-    # @note This will run any registered :success callbacks
+    # @note This will run any registered :failure callbacks
```
elrayle: Just a subtle distinction between :success callback and :failure callback.

elrayle:
This is really good.  Glad you were able to add the messaging in.

Issue #1533, opened 2017-08-21
Collection Types: Hook user collection generator into the engine_cart test app generation process


elrayle:
### Descriptive summary

Hook existing user collection generator into the engine_cart test app generation process, such that the User Collection type will exist in the test app once the engine_cart process is complete.

### Rationale

Provide the rationale or user story that describes "why" this issue should be addressed. Especially if this is a new feature or significant change to the existing implementation.

### Expected behavior

When the engine_cart process completes, the default User Collection will exist in the test app.

### Implementation

Notes by @randalldfloyd from PR #1513:

> Regarding initialization of the default collection type in the generated test app: this is currently a problem because the creation method described cannot be invoked until the database table exists. However, database migrations do not occur within the flow of the main Hyrax install generator (migration files are simply staged by various sub-generators.) It seems that the Engine Cart gem is currently responsible for the actual database migration (see https://github.com/cbeer/engine_cart/blob/master/lib/engine_cart/tasks/engine_cart.rake#L102 ). The end result is that you can't run database dependent tasks before the migration, and there's no place to tuck additional tasks in after it.
> 
> Note that this is mostly a problem for developers who like to use the generated test app in their development workflow - actual tests can always invoke the find_or_create_default_collection_type method as needed. This timing problem isn't unique to this PR and presents the same problem for other database dependent use cases. The possible solutions are:
> 
> * Hyrax could redefine the EC :generate task in a way that would allow post processing
> * Make a contribution to EC to allow for post processing options or optional placement of migrations
> * The find_or_create_default_collection_type method can be invoked by any code (from test suite or application) that requires the default collection type to exist
> * Rely on documentation and knowing to run the hyrax:default_collection_type:create rake task

### Steps to reproduce the behavior

1. run `rake engine_cart:generate`
1. start up test app
1. Dashboard -> Settings -> Collection Types

Verify that User Collection is listed as a type

1. Click Edit beside User Collection

Verify that metadata and settings are correct.  See Issue #1485 for details on expected values.

### Related work

Issue #1485-PR #1513 Added method `#self.create_default_collection_type` to [Hyrax::CollectionType](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb)

elrayle:
All requirements for this issue are met.  Closing.

Issue #1342, opened 2017-07-14
Collection Types: Update Hyrax-generation process to create a collection type for Administrative Sets


elrayle:
### Descriptive summary

Create a mostly read-only collection type for Administrative Sets.  This will allow users to interact with Admin Sets as though they are a collection type.

### Rationale

For the initial implementaiton, the Admin Set underlying code will not change much, if any.  But to a user, the Admin Set will appear to be a collection type.

See model issue for collection types rationale.

### Implementation

**File:** (lib/generators/hyrax/install_generator.rb)[https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/install_generator.rb]

Write a generator that...
- [x] creates a new collection type named: Administrative Sets, with machine_id: admin_set, and with a brief description of what an Admin Set is and the requirement that there be at least one
- [x] save the collection type machine_id in a globally accessible structure (most likely ENV variable, but maybe Hyrax has its own way of handling this - NEEDS INVESTIGATION)
- [x] assign default settings that correspond to the expected behaviors and restrictions of the current implementation for Administrative Sets
  - [x] nestable=false
  - [x] discovery=true
  - [x] sharing=true
  - [x] multiple_membership=false
  - [x] require_membership=true
  - [x] workflow=true
  - [x] visibility=true
- [ ] assign default settings for participants
  - [ ] collection_mangers -> group: admins
  - [ ] collection_creators -> group: admins

QUESTION: Is the group admin or admins?

### Related work

| Collection Types feature | Issue |
| ------------------------ | ----- |
| Admin Set Generator      | this issue |
| User Collection Generator | #1485  |
| Routes                   | #1347 |
| Controller               | #1341 |
| Model                    | #1343 |
| View - Index             | #1349 |
| View -  New/Edit         | #1348 |
| Partial -  Description   | #1344 |
| Partial -  Settings      | #1346 |
| Partial -  Participants  | #1345 |

dlpierce:
Regarding "save the collection type machine_id in a globally accessible structure", would following the existing patterns for config in /lib/hyrax/configuration.rb be sufficient?

elrayle:
@dlpierce I'm wondering if we need to do that if we define up front via the generator that the machine_id for the admin set type is `'admin_set'`.  The proposed check for `collection_type.admin_set?`  checks for `machine_id == 'admin_set'`.  See issue #1486.

elrayle:
Remaining Work will be addressed in issues:
[...]

PR #1588, opened 2017-08-29
Collections #1342 generate Admin Set collection type

randalldfloyd:
refs #1342

Adds method in CollectionType to find or create the Admin Set collection type.

Also refactors the create collection type method to: refactor it out into a service to alleviate class size; refactors to be reusable so that `machine_id`, `title`, and `options` can all be sent to method and thus allowing it to be usable to create the default user collection type, admin set type, etc.  

Also changes the rake task for creating the default user collection type so that it additionally creates the Admin Set type.  Note the change the namespace of the rake task to be of plural form:

`rake hyrax:default_collection_types:create`

As with #1485, this one is also dependent upon #1533 to create the collection types in the test app via the Hyrax install generator because the rake task is dependent upon final database migration, which is done by the engine_cart gem and does not occur until the very end.


app/services/hyrax/collection_type_create_service.rb
```diff
@@ -0,0 +1,30 @@
+module Hyrax
+  class CollectionTypeCreateService
```
elrayle: I am glad to see this move toward a service.  The collection model is getting full.  We have some other services that are going to be coming on board too.

With that in mind, can you move the service to...  `app/services/hyrax/collection_type/create_service.rb`.  That way all the services can be gathered under that directory.

randalldfloyd: A new push is coming soon.  The service module name had to be `CollectionTypes` to avoid namespace collision with the model, but it looks like plurality was a pattern followed for other Hyrax:: modules (i.e. Collection vs. Collections module).

randalldfloyd:
CI failure from latest push is due to a known random failure and not related to commit:

```Creating a new Work when the user is a proxy allows on-behalf-of deposit
     Failure/Error: raise ActionController::RoutingError, "No route matches [#{env['REQUEST_METHOD']}] #{env['PATH_INFO'].inspect}"
     
     ActionController::RoutingError:
       No route matches [GET] "/concern/generic_works"```

PR #1517, opened 2017-08-18
set old licenses to inactive and add refs to new licenses

csyversen:
Fixes #1135  

We set the old licenses to inactive so you can still pull up works that use an old license, but they aren't selectable for new works.  

@samvera/hyrax-code-reviewers

PR #1598, opened 2017-08-31
Change 'All rights reserved' to active:false

### LINKED ISSUE #589: Rights value of 'All rights reserved' uses Europeana RR-R URI that is no longer supported
### Descriptive summary

Europeana no longer supports the Rights Reserved - Restricted Access statement used in Hyrax as 'All rights reserved." http://www.europeana.eu/portal/rights/rr-r.html  (gives 'Forbidden' error)

https://github.com/projecthydra-labs/hyrax/blob/master/lib/generators/hyrax/templates/config/authorities/licenses.yml

Looks like they actually dropped support in 2014 (last paragraph here: http://pro.europeana.eu/blogpost/europeana-digital-objects-to-have-valid-rights-statement-by-july )

(Oregon Digital recently switched to an equivalent form at opaquenamespace.org: http://opaquenamespace.org/ns/rights/rr-r  as a temporary solution, until we can fully evaluate how rightsstatements.org figures in. The main factor for this was having a URI with a label that was fetchable. I'm not advocating Hyrax or anyone else use this.) 

A likely replacement is In Copyright: http://rightsstatements.org/vocab/InC/1.0/

wickr:
Fixes #589

Change 'All rights reserved' URI to active:false.
The RR-R URI is used in a couple specs, but it seemed they were more for formatting and thus didn't need updating.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/templates/config/authorities/licenses.yml
+++ b/lib/generators/hyrax/templates/config/authorities/licenses.yml
@@ -25,4 +25,4 @@ terms:
       active: true
     - id: http://www.europeana.eu/portal/rights/rr-r.html
       term: All rights reserved
-      active: true
+      active: false


wickr:
Closing since this change was included in #1517

PR #1599, opened 2017-08-31
replacing unreliable capybara assertions and adding chromedriver-helper

### LINKED ISSUE #1592: Feature tests always pass
### Descriptive summary

A check for content in a feature test always passes even when the content is not there.

### To test

I edited `spec/features/users_spec.rb` and added the following tests.

```
    it 'passes to have_content check when it should fail because the content does not exist' do
      visit profile_path
      expect(page).to have_content('GARBAGE_GARBAGE')     # Passes, but shouldn't
      expect(page).not_to have_content('GARBAGE_GARBAGE') # Also passes
      result = page.has_content?('GARBAGE_GARBAGE')       # correctly is false
      expect(result).to be false                          # so this passes, as expected
      expect(result).to be true                           # and this fails, as expected
    end

    it 'passes to have_content check when the content does exist' do
      visit profile_path
      expect(page).to have_content(user.email)     # Passes, as expected
      expect(page).not_to have_content(user.email) # Fails, as expected
    end
```

### Related work

I ran into this in trying to fix Issue #1591.  I was trying to write a failing test first so I could verify the fix.  But I couldn't get the test to fail.

I wonder if this is also related to the periodically failing batch_edit test.

jenlindner:
Fixes #1592 ; refs ##1592

Replace have_content and have_text with assert_test in feature tests, as they've proven unreliable, and the latter has proven reliable, to the best of our present knowledge. And adding chromedriver-helper gem, for easy use of chromedriver, added a few sleeps to adjust for slowness of persistence. 
  
@laritakr's comments: 
Based on testing within numerous spec files, have_content and have_text are consistently returning false passing results. In my testing, I did not find a single case where have_content or have_text produced errors as expected.

Acting on this issue involves the following:

    search for all instances of have_content and have_text
    replace all instances with assert_text as in the comment above.

* Adding Chromedriver-helper gem
* Replacing have_content and have_text with asset_text
* added a few sleeps in places where code updates backend and then checks it 

@samvera/hyrax-code-reviewers
@elrayle 
@laritakr 
@jcoyne
@mjgiarlo

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :development, :test do
+  gem "chromedriver-helper"
   gem 'coveralls', require: false
   gem 'pry' unless ENV['CI']
   gem 'pry-byebug' unless ENV['CI']
--- a/lib/generators/hyrax/work/templates/feature_spec.rb.erb
+++ b/lib/generators/hyrax/work/templates/feature_spec.rb.erb
@@ -27,7 +27,7 @@ RSpec.feature 'Create a <%= class_name %>', js: false do
       # choose "payload_concern", option: "<%= class_name %>"
       # click_button "Create work"
 
-      expect(page).to have_content "Add New <%= human_name %>"
+      page.assert_text "Add New <%= human_name %>"
     end
   end
 end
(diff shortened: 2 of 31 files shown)


jcoyne:
I want to hold off merging this until we hear from the capybara maintainers about what they think the problem is.

jenlindner:
Sounds good, I didn't realize we might get some answers from them.

jeremyf:
I have a general solution for the Capybara sleep injector that I am pushing up. We have verified proper `to have_content?` behavior (and `not_to have_content?`)

PR #1603, opened 2017-08-31
adds CollectionTypes service 

### LINKED ISSUE #1582: Collection Types: Service for collection types a user can create
### Descriptive summary

Create a service that returns the information about the collection types a user can create.

NOTE: Check with @laritakr who may have stubbed out a method to return all collection types.

### Rationale

Users are limited in which types of collections they can create based on the participant roles assigned to the user.

### Expected behavior

**Basic Rules:**  A user can create...
    * collection types where user has `manage` access for collection types by being a user with `manage` access
    * collection types where user has `manage` access for collection types by being in a group with `manage` access
    * collection types where user has `create` access for collections types by being a user with `create` access
    * collection types where user has `create` access for collections types by being in a group with `create` access

**Special cases:**
    * all collection types when user is an admin (i.e., in the `admin` group).  This will be achieved by setting access: `'manage'` agent_type: `'group'` agent_id: `'admin`' for each collection type as they are created.  This grant then follows the rules above.  For performance, we may want to take advantage of the fact that admins can manage all and just return all collection types if user.admin? is true
    * registered user without any special access can create `user_collection`.  This will be achieved by setting  access: `'create'` agent_type: `'group'` agent_id: `'registered`' for `user_collection`.  This grant then follows the rules above.

1. Create users `admin@ex.com`, `mgr1@ex.com`, `mgr2@ex.com` , `cr1@ex.com`, `cr2@ex.com`, and `other@ex.com`
1. Need groups `mgr_group` and `cr_group`
  1. add `mgr2@ex.com` to `mgr_group`
  1. add `cr2@ex.com` to `cr_group`
1. Login as an `admin@ex.com`
1. Dashboard -> Settings -> Collection Types
1. Create a collection, save, and edit
1. On Participants tab
  1. add user `mgr1@ex.com` as a manager
  1. add user `cr1@ex.com` as a creator
  1. add group `mgr_group` as a manager
  1. add group `cr_group` as a creator
1. Save

Calling service should return ids for...
* for `admin@ex.com`: [admin_set, user_collection, new_collection]
* for `mgr1@ex.com`: [user_collection, new_collection]
* for `mgr2@ex.com`: [user_collection, new_collection]
* for `cr1@ex.com`: [user_collection, new_collection]
* for `cr2@ex.com`: [user_collection, new_collection]
* for `other@ex.com`: [user_collection]

NOTE: 
* By default, admin_set can only be created by admins.  
* By default, user_collections can be created by all registered users.

### Dependencies

The following issue should be completed first in order to not require revisiting this issue.

#1581 Collection Types: Change manager to manage and creator to create for participants access

### Related work

#1345 Collection Types: Create participants partial of the new/edit form

kdid:
Fixes #1582

provides service for requesting collection types a user has access to
Hyrax::Collections::CollectionTypesService.collection_types_for_user

- pass the user and access type to check ('create', 'manage', or 'both').
- the service checks collection_type_participants table to find Collection Types where user has access of the given type(s), or a group the user belongs to has group access of the given type(s)
- Returns array of CollectionTypes

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/dashboard/collection_spec.rb
+++ b/spec/features/dashboard/collection_spec.rb
@@ -1,7 +1,7 @@
 RSpec.describe 'collection', type: :feature, clean_repo: true do
   let(:user) { create(:user) }
   let(:admin_user) { create(:admin) }
-  let(:collection_type) { create(:collection_type) }
+  let(:collection_type) { create(:collection_type, creator_user: user) }
   let(:user_collection_type) { create(:user_collection_type) }
 
   let(:collection1) { create(:public_collection, user: user, collection_type_gid: collection_type.gid) }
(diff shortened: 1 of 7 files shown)


app/services/hyrax/collections/collection_types_service.rb
```diff
@@ -0,0 +1,49 @@
...
+  module Collections
+    module CollectionTypesService
+      # @api public
+      #
+      # What types of collection can the user create or manage
+      #
+      # @param user [User] user - The user requesting to create/manage a Collection
+      # @param roles [String] type of access, 'manage' and/or 'create'
+      # @return [Array<Hyrax::CollectionType>]
+      def self.collection_types_for_user(user, roles)
```
jeremyf: Would you be able to switch to using named parameters?

```ruby
self.collection_types_for_user(user:, roles:)
```

Then callers wouldn't need to remember the parameter order when calling.

(This is a non-blocker)

elrayle: I was envisioning this method a bit differently.  The service that is needed is one that returns collection types that can be created by a user.  Perhaps this would be an additional class method in this service class.  I was thinking something like this...

```
      # @api public
      #
      # Get a list of collection types that a user can create
      #
      # @param user [User] the user that will be creating a collection (default: current_user)
      # @return [Array<Hyrax::CollectionType>] array of collection types the user can create
      def self.can_create_collection_types(user=current_user)
        # Stubbed to return all types.
        return Hyrax::CollectionType.all if user.present? # condition test added to allow paramter to be present
        # TODO: implement the following per issue #1582
        # all collection types when user is an admin (i.e., in the admin group)
        # collection types where user has manage access for collection types by being a user with manage access
        # colleciton types where user has manage access by being in a group with manage access
        # collection types where user has create access for collections types by being a user with create access
        # collection types where user has create access by being in a group with create access
        []
      end
```

kdid: Both these requests are addressed now @elrayle and @jeremyf, if you want to take a look.

spec/features/dashboard/collection_spec.rb
```diff
@@ -73,38 +73,79 @@
...
-      click_button "New Collection"
-      expect(page).to have_content 'Select type of collection'
+        sign_in user
+        visit '/dashboard/my/collections'
+      end
 
-      choose('User Collection')
-      click_on('Create collection')
+      it 'makes a new collection' do
+        click_link "New Collection"
```
elrayle: @jeremyf I'm a little concerned that this test is twitchy.  [Hyrax::Dashboard::CollectionsController#new](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/hyrax/dashboard/collections_controller.rb#L70) method seems to fail randomly at line `@collection.apply_depositor_metadata(current_user.user_key)`.  The test just passed in Travis for rails 5.0, but I had to re-run it once for rails 5.1.  I don't think it is the rails version.  An earlier commit had it fail on both.

elrayle:
@jeremyf This could use a review when you get a chance.  I had to adjust some of the tests to sync up @laritakr's work with stubbed methods that now return real values in @kdid's work.   See the one question about the collection_spec feature test.

spec/features/dashboard/collection_spec.rb
elrayle: For what it is worth, it passes on my local machine every time.

jeremyf: @elrayle I have time after 2pm today to review and look at possible solutions.

PR #1606, opened 2017-08-31
Collection #1533 engine_cart:after_generate

### LINKED ISSUE #1533: Collection Types: Hook user collection generator into the engine_cart test app generation process

dlpierce:
Fixes #1533 

Provides a way to automatically run tasks after engine_cart:generate.

Some actions (e.g. seeding data) need to run after the test_app has run db:migrate during generation, but engine_cart does not provide a builtin hook for doing so. Defining the engine_cart:generate task again in hyrax_dev.rake appends the action block to the existing task, and will be run after the original task's action block completes.

This feature is used to run hyrax:default_collection_type:create in the development test_app.

To avoid possible out of order execution of the task actions, if the engine_cart:generate task is not defined when hyrax_dev.rake is loaded, an exception will be raised.

Rubocop suggests the `$?` variable be replaced with a variable from the `English` library, but this appears to not already be in use in hyrax, so an exception was added.

@elrayle

### DIFF
--- a/tasks/hyrax_dev.rake
+++ b/tasks/hyrax_dev.rake
@@ -30,6 +30,27 @@ task :i18n_sorter do
   end
 end
 
+if Gem.loaded_specs.key? :engine_cart
+  namespace :engine_cart do
+    # This generate task should only add its action to an existing engine_cart:generate task
+    raise 'engine_cart:generate task should already be defined' unless Rake::Task.task_defined?('engine_cart:generate')
+    task :generate do |_task|
+      puts 'Running post-generation operations...'
+      Rake::Task['engine_cart:after_generate'].invoke
+    end
+
+    desc 'Operations that need to run after the test_app migrations have run'
+    task :after_generate do
+      puts 'Creating default collection type...'
+      EngineCart.within_test_app do
+        unless system "bundle exec rake hyrax:default_collection_types:create"
+          raise "EngineCart failed on with: #{$?}"
+        end
+      end
+    end
+  end
+end
+
 desc 'Generate the engine_cart and spin up test servers and run specs'
 task ci: ['rubocop', 'engine_cart:generate'] do
   puts 'running continuous integration'
--- a/.rubocop_fixme.yml
+++ b/.rubocop_fixme.yml
@@ -51,6 +51,10 @@ Style/PredicateName:
   Exclude:
     - 'app/helpers/hyrax/collections_helper.rb'
 
+Style/SpecialGlobalVars:
+  Exclude:
+    - 'tasks/hyrax_dev.rake'
+
 Rails/Output:
   Exclude:
     - 'lib/generators/**/*'


tasks/hyrax_dev.rake
```diff
@@ -30,6 +30,25 @@ task :i18n_sorter do
   end
 end
 
+namespace :engine_cart do
+  # This generate task should only add its action to an existing generate task
+  raise 'engine_cart:generate task should already be defined' unless Rake::Task.task_defined?('engine_cart:generate')
```
mjgiarlo: When would this `raise` happen? When this rake file is loaded? If so, will that break production apps when this code is evaluated? https://github.com/samvera/hyrax/blob/master/Rakefile#L20

tasks/hyrax_dev.rake
```diff
@@ -30,6 +30,25 @@ task :i18n_sorter do
...
+  task :generate do |_task|
```
mjgiarlo: Where is the original EC:generate task being run in the redeclared one? (Or is that not necessary due to Rake :sparkles:?)

dlpierce: [Rake :sparkles:.](https://ruby.github.io/rake/doc/rakefile_rdoc.html#label-Multiple+Definitions)
A task's block is appended to a list of existing action blocks.

cbeer:
How is this different than adding steps to the test app install generator (https://github.com/samvera/hyrax/blob/master/spec/test_app_templates/lib/generators/test_app_generator.rb)?

dlpierce:
As mentioned in #1533, these steps (essentially seeding data in this case) need to happen after the db:migrate that is called towards the end of engine_cart:generate.
Alternatively, engine_cart could be changed to call db:seed or provide a builtin task hook at this point.

PR #1610, opened 2017-08-31
Modify permission_templates for collections

### LINKED ISSUE #1593: Collections: modify permissions template controller to receive collection IDs
### Descriptive summary

This is follow up work for #1378.  The PermissionTemplatesController needs to be modified to receive collection IDs.  It has only received admin set IDs in the past, but needs to be able to handle both moving forward.

### Expected behavior

Both admin sets and collections should be able to use the same PermissionTemplatesController to manage permission templates.

hortongn:
Fixes #1593 

(collections sprint)

Makes the necessary modifications for the "Sharing" tab on the collections edit page to function.  Managers, depositors, and viewers can be added and deleted from collections.  The Participants tab for admin sets (which same the same permissions template controller with collections) continues to work properly as well.

Using the same permission templates controller for both admin sets and collections requires a lot of conditional logic.  Normally that might indicate a separate controller for collection permission templates, but given that admin sets should eventually morph into just another type of collection, this is probably the best approach.

### DIFF
--- a/app/controllers/hyrax/admin/permission_template_accesses_controller.rb
+++ b/app/controllers/hyrax/admin/permission_template_accesses_controller.rb
@@ -10,14 +10,9 @@ module Hyrax
         end
 
         if @permission_template_access.destroyed?
-          redirect_to hyrax.edit_admin_admin_set_path(source_id,
-                                                      anchor: 'participants'),
-                      notice: translate('participants', scope: 'hyrax.admin.admin_sets.form.permission_update_notices')
+          redirect_to_edit_path
         else
-          redirect_to hyrax.edit_admin_admin_set_path(source_id,
-                                                      anchor: 'participants'),
-                      alert: @permission_template_access.errors.full_messages.to_sentence
-
+          redirect_to_edit_path_with_error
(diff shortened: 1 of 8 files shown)


app/controllers/hyrax/admin/permission_template_accesses_controller.rb
```diff
@@ -31,6 +26,30 @@ def source_id
         def update_management
           Forms::PermissionTemplateForm.new(@permission_template_access.permission_template).update_management
         end
+
+        def redirect_to_edit_path
+          if Hyrax::PermissionTemplate.find(@permission_template_access.permission_template_id).source_type == 'admin_set'
```
jeremyf: The following query is duplicated: 
`Hyrax::PermissionTemplate.find(@permission_template_access.permission_template_id).source_type`

Please extract a common method for this.

hortongn: I've made the requested changes and the PR is green again

elrayle:
@jeremyf Thanks for merging!

PR #1613, opened 2017-09-01
Replace banner image with one that does not require attribution

### LINKED ISSUE #1551: original URL for default background image 404s
The current value for [`background_attribution_html`](https://github.com/samvera/hyrax/blob/master/config/locales/hyrax.en.yml#L259) as of 181141da95e7f9efe5d5d7a1b8b38fed6c7c1ff9 links back to the Flickr source for the background image using https://flic.kr/p/eirxaf, which no longer resolves.

mjgiarlo:
Fixes #1551

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.de.yml
+++ b/config/locales/hyrax.de.yml
@@ -263,7 +263,7 @@ de:
       unprocessable_entity:
         default: Die Ressource, die Sie zu ändern versucht haben, kann nicht entsprechend Ihrer Anfrage geändert werden.
         empty_file: Die hochgeladene Datei hat keinen Inhalt.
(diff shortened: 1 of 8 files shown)


mjgiarlo:
Failure above is transient (and only present in one of the two builds).

cam156:
@mjgiarlo Is the url your own picture?  I can get to the link directly, but there is no information on who owns the image.

mjgiarlo:
@cam156 That photo is one that @ggeisler grabbed and it requires no attribution. See https://github.com/samvera/hyrax/issues/1551#issuecomment-326624909

PR #1618, opened 2017-09-03
Raise error when attempting to set collection_type when collection ha…

### LINKED ISSUE #1529: Collections: Prevent changes to collection type gid
### Descriptive summary

Need to prevent collection_type_gid property in Collection model from being changed once it is assigned.

### Rationale

A collection is created of a specific collection type.  The type is identified by assigning the collection type's global identifier (gid) to the collection's `collection_type_gid` property.  A collection cannot be converted to a different type by changing its `collection_type_gid`.  So attempts to change the `collection_type_gid` should be disallowed. 

### Expected behavior

* Setting the `collection_type_gid` when it is `nil` should be allowed.  
* If this is a new collection, the `collection_type_gid` can be assigned any type.  
* If this is an edit of an existing collection and the `collection_type_gid` is `nil`, it will be assigned the default `collection_type_gid` (e.g. `user_collection`).  
* In all other scenarios, an exception will be thrown if an attempt is made to change the `collection_type_gid` from one `gid` to another.

Side effect: When `collection_type_gid` is assigned, also load `collection_type` into the collection object.

### Actual behavior

There is currently no check in place for `collection_type_gid`.

### Notes

This is complicated by the fact that `collection_type_gid` is a property that is controlled by `ActiveTriples`.  I wasn't able to find an easy way to capture `collection.collection_type_gid=` and add code to check if it already has a value before assigning a new one.  

The side effect of assigning `collection_type` when setting `collection_type_gid` can only be accomplished if we can override `collection.collection_type_gid=`.  Currently, loading of `collection_type` is done the first time it is accessed.

Create a test with the following...
1. create 2 Collection Types (in a test, you can fake this by just creating two gids, e.g. `gid://internal/hyrax-collectiontype/89` and `gid://internal/hyrax-collectiontype/99`)
1. create a new Collection and assign it one of the type's gid using `collection.collection_type_gid=` method
1. attempt to assign the second type's gid to the collection using `collection.collection_type_gid=` method

At the very least, it should not change the value.  Even better, it should throw an exception stating that `collection_type_gid` cannot be changed.

### Related work

Issue #1486  - PR #1521 - Add Collection Type to Collection 
* added Hyrax::CollectionType [#gid](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb#L45-L53), [#find_by_gid!](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb#L35-L43) methods
* added Hyrax::CollectionBehavior [#load_collection_type_instance](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/concerns/hyrax/collection_behavior.rb#L101-L113), [#collection_type](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/concerns/hyrax/collection_behavior.rb#L29-L33), [#collection_type=](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/concerns/hyrax/collection_behavior.rb#L92-L97)

cjcolvar:
…s already been persisted with a collection_type

Fixes #1529 

This PR has a good deal of refactoring in order to clean up and make easier the actual work on #1529.  I believe these changes will help clear up some pain points mentioned [here](https://github.com/samvera/hyrax/blob/collections-sprint/app/presenters/hyrax/collection_presenter.rb#L23-L43) and previously [here](https://github.com/samvera/hyrax/commit/ad9b00a87e76c8b86072d574f36feaa25a6e619b#diff-47461a9012b685da44afe3d1a87111ffR12) as well as others present in the spec tests.

Guiding principals for this work were:
* Collections need a collection type for delegated methods
* Every collection must have a collection type
* Once a collection is persisted with a collection type it cannot be changed
* Attempt to mimic ActiveRecord where possible

Changes proposed in this pull request beyond the original issue:
* Change assignment of unassigned collections (including `new?` collections) to a collection type from `after_find` to `after_initialize`
* Assignment of collection type of a previously persisted collection does not force saving
* Reassignment of collection type is valid up until the point of saving
* Validate the presence of collection_type_gid
* Add `with.library rails` to the shoulda-matchers configuration block per their documentation.  This fixed an issue where `validate_presence_of` wasn't visible in the collection model test.

Work still to do:
* `:clean_repo` could probably be removed from more spec blocks
* More validations may be needed to ensure that all aspects of a collection type are being adhered to since this flow might be possible: create new collection, set collection type, add works, change collection type, save collection
* `collection_type=` is a convenience method for `collection_type_gid=` but is inefficient since it does a find to get the same collection type that was passed in to `collection_type=`
* I couldn't think of a good way to test the collection_type_gid presence validation since the code ensures that every collection will always have a collection_type_gid and the`validate_presence_of` shoulda-matcher attempts to set it to `nil` which throws an error causing the matcher to fail.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -109,7 +109,7 @@ module Hyrax
         @collection.attributes = collection_params.except(:members)
         @collection.apply_depositor_metadata(current_user.user_key)
         add_members_to_collection unless batch.empty?
-        @collection.collection_type_gid = params[:collection_type_gid]
+        @collection.collection_type_gid = params[:collection_type_gid] if params[:collection_type_gid]
         # TODO: There has to be a better way to handle a missing gid than setting to User Collection.
         # TODO: Via UI, there should always be one defined.  It is missing right now because the modal isn't implemented yet
         # TODO: But perhaps this is needed for the case when it gets called outside the context of the UI?
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -22,20 +22,31 @@ module Hyrax
         index.as(*index_collection_type_gid_as)
       end
 
-      after_find { |col| load_collection_type_instance(col) }
-      # @todo check that gid is not nil AND check that gid has not changed since
-      #       it was read in (It is ok to go from nil to a value.)
-      #
-      # before_update { |col| validate_collection_type_gid(col) }
+      # validates that collection_type_gid is present
+      validates :collection_type_gid, presence: true
+      after_initialize :ensure_collection_type_gid
+
+      # Need to define here in order to override setter defined by ActiveTriples
+      def collection_type_gid=(new_collection_type_gid)
+        if persisted? && !collection_type_gid_was.nil? && collection_type_gid_was != new_collection_type_gid
+          raise "Can't modify collection type of this collection"
+        end
+        new_collection_type = Hyrax::CollectionType.find_by_gid!(new_collection_type_gid)
+        super
+        @collection_type = new_collection_type
+        collection_type_gid
+      end
     end
 
     delegate(*Hyrax::CollectionType.collection_type_settings_methods, to: :collection_type)
 
-    # Get (and set) the collection_type when accessed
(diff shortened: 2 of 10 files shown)


jeremyf:
I give a +1 but want @elrayle to weigh in

Issue #1601, opened 2017-08-31
Collection Types: add default participants for admin set collection type


elrayle:
### Descriptive summary

The predefined Admin Set collection type was created by Issue #1342-PR #1588.  This included default metadata and settings.  It did not implement default participants.  This issue addresses this remaining work from that issue.

### Implementation

**File:** (lib/generators/hyrax/install_generator.rb)[https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/install_generator.rb]

Add default participants to [Hyrax::CollectionType](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb) `#self.find_or_create_admin_set_type` method.

- [ ] assign default settings for participants
  - [ ] collection_mangers -> group: admin
  - [ ] collection_creators -> group: admin

### Dependency

Issue #1345 Collection Types: Create participants partial of the new/edit form

### Related work

Issue #1342-PR #1588 Added method `#self.find_or_create_admin_set_type` to [Hyrax::CollectionType](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb)

elrayle:
Fixed by PR #1622

Issue #1532, opened 2017-08-21
Collection Types: add default participants for user collection type


elrayle:
### Descriptive summary

The default User Collection was created by Issue #1485-PR #1513.  This included default metadata and settings.  It did not implement default participants.  This issue addresses this remaining work from that issue.

### Implementation

**File:** (lib/generators/hyrax/install_generator.rb)[https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/install_generator.rb]

Add default participants to [Hyrax::CollectionType](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb) `#self.create_default_collection_type` method.

- [ ] assign default settings for participants
  - [ ] collection_mangers -> group: admin
  - [ ] collection_creators -> group: registered users

### Dependency

Issue #1345 Collection Types: Create participants partial of the new/edit form

### Related work

Issue #1485-PR #1513 Added method `#self.create_default_collection_type` to [Hyrax::CollectionType](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/hyrax/collection_type.rb)

elrayle:
Fixed by PR #1622

PR #1622, opened 2017-09-05
Adding functionality to add participants to collection types

randalldfloyd:
refs #1532 #1601 

Adds a method to the Collection Type Create Service to create participants for a collection type, which is then used when creating the default user collection type and the default admin set type.  The options for the participants are encapsulated in the collection type options sent to the create service.

### DIFF
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -83,6 +83,8 @@ module Hyrax
     # * USER_COLLECTION_DEFAULT_TITLE
     # * Hyrax::CollectionTypes::CreateService::DEFAULT_OPTIONS
     #
+    # @see Hyrax::CollectionTypes::CreateService
+    #
     # @return [Hyrax::CollectionType] where machine_id = USER_COLLECTION_MACHINE_ID
     def self.find_or_create_default_collection_type
       find_by(machine_id: USER_COLLECTION_MACHINE_ID) || Hyrax::CollectionTypes::CreateService.create_collection_type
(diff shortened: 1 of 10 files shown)


app/services/hyrax/collection_types/create_service.rb
```diff
@@ -40,6 +42,20 @@ def self.create_collection_type(machine_id: DEFAULT_MACHINE_ID, title: DEFAULT_T
...
+        add_participants(ct.id, opts[:participants])
+        ct
+      end
+
+      # @param collection_type_id [Integer]
+      # @param participants [Array]
+      def self.add_participants(collection_type_id, participants)
+        return unless collection_type_id && participants.count > 0
+        participants.each do |p|
+          agent_type = p[:agent_type]
```
jeremyf: If the given participant is `{ }`, we end up with `agent_type = nil` is that an acceptable value for the `.create` I'm wondering about using `p.fetch(:agent_type)` instead. (As well as for `p.fetch(:agent_id)` and `p.fetch(:access)`).

app/services/hyrax/collection_types/create_service.rb
```diff
@@ -40,6 +42,20 @@ def self.create_collection_type(machine_id: DEFAULT_MACHINE_ID, title: DEFAULT_T
...
+          agent_id = p[:agent_id]
+          access = p[:access]
+          Hyrax::CollectionTypeParticipant.create(hyrax_collection_type_id: collection_type_id, agent_type: agent_type, agent_id: agent_id, access: access)
```
jeremyf: I believe this should be `.create!` instead of `.create`. If the underlying validation fails, I think that is a major thing to consider.

elrayle: Another consideration is that a particular combination should not be added twice.  A record is considered a duplicate when all four fields are identical...  collection type id, agent type, agent id, and access.

app/models/hyrax/collection_type.rb
```diff
@@ -94,19 +96,20 @@ def self.find_or_create_default_collection_type
...
-        discoverable: true,
-        sharable: true,
-        allow_multiple_membership: false,
-        require_membership: true,
+        nestable: false, discoverable: true, sharable: true, allow_multiple_membership: false, require_membership: true,
         assigns_workflow: true,
-        assigns_visibility: true
+        assigns_visibility: true,
+        participants: [{ agent_type: 'group', agent_id: 'admin', access: 'manage' },
+                       { agent_type: 'group', agent_id: 'admin', access: 'manage' }]
```
elrayle: You are setting access to 'manage' twice.  This should be...

```
participants: [{ agent_type: 'group', agent_id: 'admin', access: 'manage' },
               { agent_type: 'group', agent_id: 'admin', access: 'create' }]
```

app/services/hyrax/collection_types/create_service.rb
elrayle: I would prefer to see add_participants as part of the [PermissionsService](https://github.com/samvera/hyrax/blob/collections-sprint/app/services/hyrax/collection_types/permissions_service.rb) since that service deals with managing and querying participants.

app/services/hyrax/collection_types/create_service.rb
kdid: There isn't any validation preventing adding the same combo twice. There doesn't seem to be in admin sets/permission templates either. Adding this validation would probably be a good idea.

app/services/hyrax/collection_types/create_service.rb
elrayle: RE: acceptable values
* agent_type must be either `'group'` or `'user'`
* access must be either `'manage'` or `'create'`
* none of the values can be nil

Validation should be added to the model for that.

elrayle: * Updated to use fetch.  Model already has `presence` validation of all fields.
* Created an issue to cover validation isssues..  #1649 Collection Types: Add validations to participants model 
* Method was moved to PermissionsService

app/services/hyrax/collection_types/create_service.rb
elrayle: * Method was switched to `create!`
* As mentioned above, validation issues are covered by #1649.  This includes preventing dups.

app/models/hyrax/collection_type.rb
elrayle: * verified now creates `manage` and `create` access

elrayle:
While testing this PR, I found a seed that would fail predictably with the text failure described in Issue #1641.  I was able to put in a fix for this failing test as part of this PR.

Fixes by Issue #1641

PR #1624, opened 2017-09-05
[WIP] Use websockets for notifications instead of polling

mjgiarlo:
This is quite WIP-y. Needs code review and tests. Sharing this now to get initial feedback on the approach I've taken.

@samvera/hyrax-code-reviewers


app/assets/javascripts/hyrax/cable.js
```diff
@@ -0,0 +1,12 @@
...
+// You can generate new channels where WebSocket features live using the rails generate channel command.
+//
+//= require action_cable
+//= require_self
+//= require_tree ./channels
+
+(function() {
+  this.App || (this.App = {});
+
+  App.cable = ActionCable.createConsumer("/notifications/endpoint");
```
mjgiarlo: This is hard-coded at the moment to the new path from `config/routes.rb`

app/assets/javascripts/hyrax/channels/notifications.js
```diff
@@ -0,0 +1,27 @@
+App.cable.subscriptions.create("Hyrax::NotificationsChannel", {
+    connected: function(data) {
+        this.perform("update_locale", { locale: $('html').attr('lang') })
```
mjgiarlo: This persists the current locale value the user has chosen to the User model, which the notifications can then use when rendering a new notification. This is needed because the code that processes notifications does not have access to session -- indeed, a user may not even be connected when a notification fires. This also moves us in the direction of being able to allow users to set their own preferred locale and have it 'stick'.

app/assets/javascripts/hyrax/channels/notifications.js
```diff
@@ -0,0 +1,27 @@
...
+    },
+
+    received: function(data) {
+        let count = data["notify_number"];
```
mjgiarlo: Whenever a notification is created for a user, two bits of data are passed via the websocket: the number of unread notifications and a message like "You have 12 notifications" that is translated into the user's preferred locale (or the default locale if no preference exists).

app/controllers/hyrax/depositors_controller.rb
```diff
@@ -38,8 +38,8 @@ def authorize_and_return_grantor
       def send_proxy_depositor_added_messages(grantor, grantee)
         message_to_grantee = "#{grantor.name} has assigned you as a proxy depositor"
         message_to_grantor = "You have assigned #{grantee.name} as a proxy depositor"
-        ::User.batch_user.send_message(grantor, message_to_grantor, "Proxy Depositor Added")
-        ::User.batch_user.send_message(grantee, message_to_grantee, "Proxy Depositor Added")
+        Hyrax::MessengerService.deliver(::User.batch_user, grantor, message_to_grantor, "Proxy Depositor Added")
```
mjgiarlo: I've taken the liberty of consolidating all uses of `send_message` (from Mailboxer) into a new service class, that way we have a single place where we need to tie notifications into ActionCable.

app/controllers/hyrax/notifications_controller.rb
```diff
@@ -8,6 +8,8 @@ def index
       add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
       add_breadcrumb t(:'hyrax.admin.sidebar.notifications'), hyrax.notifications_path
       @messages = user_mailbox.inbox
+      # Update the notifications now that there are zero unread
+      StreamNotificationsJob.perform_later(current_user)
```
mjgiarlo: This is the only place in the code where user notifications are marked as read. Calling the job here ensures the user's notifications clear out.

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -54,6 +54,23 @@ def browser_supports_directory_upload?
       user_agent.include? 'Chrome'
     end
 
+    # @param [User] user
```
mjgiarlo: I replaced the notify_number partial with this helper, since there's a fair bit of logic here.

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -54,6 +54,23 @@ def browser_supports_directory_upload?
...
+    def render_notifications(user:)
+      mailbox = UserMailbox.new(user)
+      unread_notifications = mailbox.unread_count
+      link_to(hyrax.notifications_path,
+              'aria-label' => mailbox.label(params[:locale]),
```
mjgiarlo: When the page is first rendered, before the actioncable subscriber connects, we don't know whether the user has just switched locales or not. So rather than deferring to the user's preferred locale, which hasn't been set since the *last* page load, we pass on the locale param from the current page load which has an overriding effect.

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -54,6 +54,23 @@ def browser_supports_directory_upload?
...
+              class: 'notify-number') do
+        capture do
+          concat content_tag(:span, '', class: 'fa fa-bell')
+          concat "\n"
```
mjgiarlo: It looks janky but the newline here preserves the prior appearance as set by the `notify_number` partial.

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -233,6 +250,16 @@ def user_agent
         request.user_agent || ''
       end
 
+      def count_classes_for(unread_count)
+        'count label '.tap do |classes|
```
mjgiarlo: Always use the `count` and `label` css classes. Otherwise, vary the classes depending on the unread count.

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -233,6 +250,16 @@ def user_agent
...
+          classes << if unread_count.zero?
+                       'invisible label-default'
```
mjgiarlo: Do not show the number badge if the count is zero.

app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -233,6 +250,16 @@ def user_agent
...
+                     else
+                       'label-danger'
```
mjgiarlo: Else, if the unread count is > 0, light it up with :red_circle:

app/services/hyrax/abstract_message_service.rb
```diff
@@ -1,5 +1,5 @@
 module Hyrax
-  class MessageUserService
+  class AbstractMessageService
```
mjgiarlo: Rename this to better reflect intent -- else, this may be easily confused with the new `MessengerService`

app/services/hyrax/messenger_service.rb
```diff
@@ -0,0 +1,8 @@
+module Hyrax
+  class MessengerService
+    def self.deliver(sender, recipients, body, subject, *args)
+      sender.send_message(recipients, body, subject, *args)
+      StreamNotificationsJob.perform_later(recipients)
```
mjgiarlo: The MessengerService is used to send all notifications. After every message is sent (possibly to multiple recipients), queue a new StreamNotificationsJob.

lib/hyrax/engine.rb
```diff
@@ -25,6 +25,11 @@ class Engine < ::Rails::Engine
       "Blacklight::Exceptions::RecordNotFound" => :not_found
     )
 
+    config.before_initialize do
+      # ActionCable should use Hyrax's connection class instead of app's
+      config.action_cable.connection_class = -> { 'Hyrax::ApplicationCable::Connection'.safe_constantize }
```
mjgiarlo: Without this, the ActionCable server will *always* use `ApplicationCable::Connection` which is in the domain of the application, not Hyrax.

app/assets/javascripts/hyrax/cable.js
```diff
@@ -0,0 +1,12 @@
+// Action Cable provides the framework to deal with WebSockets in Rails.
+// You can generate new channels where WebSocket features live using the rails generate channel command.
+//
+//= require action_cable
+//= require_self
+//= require_tree ./channels
+
+(function() {
+  this.App || (this.App = {});
```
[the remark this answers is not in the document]
mjgiarlo: This comes right from Rails. Meh.

mjgiarlo:
Another important note! I couldn't get this to work in the test app with the default (`async`) AC adapter. The consumer (Javascript client) would connect to the AC back-end, and events would fire on the server-side, but the consumer wouldn't receive data from the server.

Once I edited `.internal_test_app/config/cable.yml` to use Redis at `localhost:6379`, this was fixed.

app/assets/javascripts/hyrax/cable.js
barmintor: Are you trying to avoid erb'ing this file to pull a named route in, or is that TBD?

spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -70,6 +70,12 @@ def relax_routing_constraint
     gsub_file 'config/initializers/arkivo_constraint.rb', 'false', 'true'
   end
 
+  # def run_action_cable_within_test_app
```
barmintor: minor thing: unless you plan on bringing this method back, I recommend deleting it and letting the revision control keep your backup!

app/assets/javascripts/hyrax/cable.js
mjgiarlo: Laziness, to see if folks cared. I'll make this change. :)

spec/test_app_templates/lib/generators/test_app_generator.rb
mjgiarlo: Yep, that's already in this branch. You may be looking at a change in an earlier commit.

app/assets/javascripts/hyrax/cable.js.erb
```diff
@@ -8,5 +8,5 @@
 (function() {
   this.App || (this.App = {});
 
-  App.cable = ActionCable.createConsumer("/notifications/endpoint");
+  App.cable = ActionCable.createConsumer(<%= Hyrax::Engine.routes.url_helpers.notifications_endpoint_path %>);
```
barmintor: this is a nice change.

escowles:
@mjgiarlo This looks like a good direction to me: newer tech, avoid polling (which I always found annoying when I was trying to watch logfiles), consolidating the message sending, good i18n, etc.

app/assets/javascripts/hyrax.js
```diff
@@ -37,11 +37,16 @@
 //= require webcomponentsjs/0.5.4/CustomElements.min
 //= require time-elements
 
+//= require action_cable
+
 //= require hyrax/monkey_patch_turbolinks
+//= require hyrax/fileupload
```
mjgiarlo: Moved this before the `almond` require, because otherwise it is unhappy.

app/assets/javascripts/hyrax.js
```diff
@@ -37,11 +37,16 @@
...
+// Provide AMD module support
+//= require almond
```
mjgiarlo: Moved `almond` up because it's needed by `hyrax/notification`, which needs to be loaded before `hyrax/app` is.

mjgiarlo:
@barmintor @escowles Tests now written, if you have time to take another gander.

mjgiarlo:
Also: @jcoyne @cbeer @elrayle @samvera/hyrax-code-reviewers

app/helpers/hyrax/hyrax_helper_behavior.rb
jenlindner: lol - but looking better to the user is the important thing

app/services/hyrax/abstract_message_service.rb
jenlindner: Cool

app/assets/javascripts/hyrax.js
jenlindner: Might be nice to comment which components need to be before and after almond

spec/jobs/stream_notifications_job_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    context 'with a user' do
+      let(:users) { create(:user) }
+      let(:mailbox) { double('mailbox', unread_count: 7, label: "You've got mail!") }
+
+      before do
+        allow(UserMailbox).to receive(:new).and_return(mailbox)
+      end
+
+      it 'broadcasts to the user' do
+        # TODO: figure out why the heck this test fails
```
emanuil-tolev: Sorry to wade in, just following developments. Is this still valid? Your CI build is passing.

mjgiarlo: Ah! This is cruft. Will remove, thx!

mjgiarlo: Done.

app/assets/javascripts/hyrax.js
mjgiarlo: I like the idea. In practice, I don't know what needs almond and I'm not sure how I'd even figure that out. Maybe someone who is more knowledgeable about this part of the code could toss a new commit in here in the future that more rationally orders these requires with appropriate comments?

PR #1626, opened 2017-09-06
Fix detection of loaded gems.

dlpierce:
Quick fix for detecting loaded gems from the dev rake file. Ref #1606


elrayle:
The test failure is a previously identified randomly failing test.  It is ok to ignore this failure.

PR #1632, opened 2017-09-07
Add new work to collection from collection show page

### LINKED ISSUE #1552: Collection: Add button to create a work through a collection from the admin show page
### Descriptive summary

Add a button to the collection admin show page that creates a work that when saved is added to the collection.

### Implementation

**Edit File:** 
* [app/views/hyrax/dashboard/collections/show.html.erb](https://github.com/samvera/hyrax/blob/master/app/views/dashboard/hyrax/collections/show.html.erb)

- [x] Modify title of button `Add works`  to be `Add existing works`
- [x] Add a `Add new work` button near the existing renamed `Add existing works` button.  We'll rename and move it to match the mockup in another issue.
- [x] Process the button in a way similar to the following...

* Dashboard -> Repository Content -> Works -> New Work -> fill in requirements and save
* Dashboard -> Repository Content -> Works -> edit the work you just created -> Relationships tab
* click Attach New Work button

This will go through the new work process and when done, the new work will be a member of the first work you created.

We want to do the same thing with the process following...

* Dashboard -> Repository Content -> Collections -> Create a new collection -> fill in requirements and save 
* Dashboard -> Repository Content -> Collections -> click the collection you just created to go to the admin show page
* click `Add new work` button

This will go through the new work process and when done, the new work will be a member of the collection you created.

This means that under the covers, the collection id has to be passed to the new work form and has to be automatically assigned as the parent collection in the relationships tab.

### Mockup

Mockup shown for reference only as to the direction we are moving the admin show page's UI.

admin-collection-show

### Related work

#1427 Collections: Create an admin show view

elrayle:
Fixes #1552

Completes work started by @gordonleacock 

Include `Add new work to collection` button on the Collections show page.  It automatically selects the originating collection as the work's parent collection on the new form.  When the new work process is complete, the new work is added in the collection.

### DIFF
--- a/app/views/hyrax/dashboard/collections/_edit_actions.html.erb
+++ b/app/views/hyrax/dashboard/collections/_edit_actions.html.erb
@@ -1,8 +1,8 @@
 <h2 class="sr-only"><%= t('hyrax.collection.actions.header') %></h2>
 <div class="actions-controls-collections">
-  <%= link_to t('hyrax.collection.actions.add_works.label'),
+  <%= link_to t('hyrax.collection.actions.add_existing_works.label'),
               hyrax.my_works_path(add_works_to_collection: @form.id),
-              title: t('hyrax.collection.actions.add_works.desc'),
+              title: t('hyrax.collection.actions.add_existing_works.desc'),
               class: 'btn btn-default',
               data: { turbolinks: false } %>
 </div>
--- a/app/views/hyrax/dashboard/collections/_form.html.erb
+++ b/app/views/hyrax/dashboard/collections/_form.html.erb
@@ -19,9 +19,9 @@
       <% end %>
     <% end %>
   </ul>
-  <%= simple_form_for @form, url: [hyrax, :dashboard, @form], html: { class: 'editor' } do |f| %>
 
-  <div class="tab-content">
+  <%= simple_form_for @form, url: [hyrax, :dashboard, @form], html: { class: 'editor' } do |f| %>
+    <div class="tab-content">
       <div id="description" class="tab-pane active">
         <div class="panel panel-default labels">
           <div class="panel-body">
@@ -61,7 +61,7 @@
             <% end %>
           </div>
         </div>
-      </div>
+      </div> <!-- end description -->
 
       <% if @form.persisted? %>
         <div id="branding" class="tab-pane">
@@ -97,8 +97,7 @@
         <% end %>
         <%= link_to t(:'helpers.action.cancel'), main_app.root_path, class: 'btn btn-link' %>
       </div>
+    </div> <!-- end tab-content -->
+  <% end # simple_form_for %>
 
-  <% end %>
-  <% # end of form %>
-
-</div>
+</div> <!-- end collection-controls -->
--- a/app/views/hyrax/dashboard/collections/_show_add_items_actions.html.erb
+++ b/app/views/hyrax/dashboard/collections/_show_add_items_actions.html.erb
@@ -1,16 +1,21 @@
 <h2 class="sr-only"><%= t('hyrax.collection.actions.header') %></h2>
 <div class="actions-controls-collections">
   <% if can? :edit, presenter.solr_document # TODO: Shouldn't this be :deposit -- dependency on participants %>
-    <%= link_to t('hyrax.collection.actions.add_works.label'),
+    <%= link_to t('hyrax.collection.actions.add_existing_works.label'),
                 hyrax.my_works_path(add_works_to_collection: presenter.id),
-                title: t('hyrax.collection.actions.add_works.desc'),
+                title: t('hyrax.collection.actions.add_existing_works.desc'),
                 class: 'btn btn-default pull-right',
                 data: { turbolinks: false } %>
+    <%= link_to t('hyrax.collection.actions.add_new_work.label'),
+                '#',
+                title: t('hyrax.collection.actions.add_new_work.desc'),
+                data: { behavior: "select-work", target: "#worktypes-to-create", 'create-type' => 'single', add_works_to_collection: presenter.id },
+                class: 'btn btn-default  pull-right'  %>
     <% if presenter.collection_type_is_nestable? %>
-        <%= link_to t('hyrax.collection.actions.nest_collections.label'),
-                    hyrax.dashboard_new_nest_collection_within_path(child_id: presenter.id),
-                    title: t('hyrax.collection.actions.nest_collections.desc'),
-                    class: 'btn btn-default' %>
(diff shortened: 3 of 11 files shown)


app/views/shared/_select_work_type_modal.html.erb
```diff
@@ -7,6 +7,7 @@
       </div>
       <form class="new-work-select">
         <div class="modal-body">
+          <% params[:add_works_to_collection] = params['id'] if params['id'] && params['controller'] == 'hyrax/dashboard/collections' %>
```
jeremyf: Why was this added? An inline comment would be helpful to understand. I don't usually expect that the `params` object is updated in the view. I'm wondering if this is better done somewhere else?

elrayle: I agree this is not optimal.  I was attempting to make the minimal change instead of refactoring.  Bad rush job.  I will look at refactoring now.

The thing I really don't like about the current code (at master) is that this file is in views/shared and is included on every page.  As far as I know, the selector is only needed in places where works get created.

* Dashboard -> My/All Works
* Collection -> show page
* maybe from the edit work form
* any place else?

app/views/hyrax/dashboard/collections/_edit_actions.html.erb
```diff
@@ -1,8 +1,7 @@
 <h2 class="sr-only"><%= t('hyrax.collection.actions.header') %></h2>
 <div class="actions-controls-collections">
-  <%= link_to t('hyrax.collection.actions.add_works.label'),
+  <%= link_to t('hyrax.collection.actions.add_existing_works.label'),
               hyrax.my_works_path(add_works_to_collection: @form.id),
-              title: t('hyrax.collection.actions.add_works.desc'),
-              class: 'btn btn-default',
-              data: { turbolinks: false } %>
```
mjgiarlo: Did you mean to remove this line? Are we no longer seeing turbolinks-related issues?

app/views/hyrax/dashboard/collections/_form.html.erb
```diff
@@ -19,9 +19,9 @@
       <% end %>
     <% end %>
   </ul>
-  <%= simple_form_for @form, url: [hyrax, :dashboard, @form], html: { class: 'editor' } do |f| %>
 
-  <div class="tab-content">
+  <%= simple_form_for @form, url: [hyrax, :dashboard, @form], html: { class: 'editor' } do |f| %>
```
mjgiarlo: Can you say more about the changes made to this file?

app/views/hyrax/dashboard/collections/_show_add_items_actions.html.erb
```diff
@@ -1,16 +1,21 @@
...
+    <%= link_to t('hyrax.collection.actions.add_existing_works.label'),
                 hyrax.my_works_path(add_works_to_collection: presenter.id),
-                title: t('hyrax.collection.actions.add_works.desc'),
+                title: t('hyrax.collection.actions.add_existing_works.desc'),
                 class: 'btn btn-default pull-right',
                 data: { turbolinks: false } %>
+    <%= link_to t('hyrax.collection.actions.add_new_work.label'),
+                '#',
+                title: t('hyrax.collection.actions.add_new_work.desc'),
+                data: { behavior: "select-work", target: "#worktypes-to-create", 'create-type' => 'single', add_works_to_collection: presenter.id },
```
mjgiarlo: And this shouldn't require the `data-turbolinks` jazz, right?

app/views/hyrax/dashboard/collections/_show_add_items_actions.html.erb
```diff
@@ -1,16 +1,21 @@
...
+                class: 'btn btn-default  pull-right'  %>
     <% if presenter.collection_type_is_nestable? %>
-        <%= link_to t('hyrax.collection.actions.nest_collections.label'),
-                    hyrax.dashboard_new_nest_collection_within_path(child_id: presenter.id),
-                    title: t('hyrax.collection.actions.nest_collections.desc'),
-                    class: 'btn btn-default' %>
+    <%= link_to t('hyrax.collection.actions.nest_collections.label'),
```
mjgiarlo: Don't we want this indented so it's clear this is within the `if` block?

app/views/shared/_select_work_type_modal.html.erb
```diff
@@ -7,12 +8,14 @@
       </div>
       <form class="new-work-select">
         <div class="modal-body">
+          <% collection_id = params[:add_works_to_collection] %>
+          <% collection_id ||= params['id'] if params['id'] && params['controller'] == 'hyrax/dashboard/collections' %>
```
mjgiarlo: I would prefer to see this handled in the presenter if possible, rather than having this logic in a view partial. Now that the presenter branches on `collection_id`, you could define a new method in the presenter to handle this case?

app/views/hyrax/dashboard/collections/_edit_actions.html.erb
elrayle: Nope! Because we are indeed still seeing those failures.  I'm sure this was a copy and paste error I copied from a change that happened before your work.  Glad you caught it.

app/views/hyrax/dashboard/collections/_form.html.erb
elrayle: The tags were not aligned.  I added comments to make it clearer the tag closing an open.  I cleaned up the indentations to make things easier to read.  This helped me to verify that there was indeed a missing close `</div>`.  I added the missing close tag.

app/views/hyrax/dashboard/collections/_show_add_items_actions.html.erb
elrayle: There are some places in the code that seem to try and make the produces html indent correctly (which is what is seen here) and others that make the erb indent correctly (like in _form above).  I'm not sure which is 'better', so I'll go with your suggestion.  It's pretty inconsistent in the code.

app/views/hyrax/dashboard/collections/_show_add_items_actions.html.erb
elrayle: Here it is correct to not include `data-turbolinks`.  The work type selection modal will not display if you include it.

mjgiarlo:
Thanks, @elrayle!

app/views/shared/_select_work_type_modal.html.erb
elrayle: Moved to presenter.

elrayle:
Approval by @mjgiarlo  in slack on Sept 8 at 6:16pm

PR #1635, opened 2017-09-07
Share button should link to My Works

### LINKED ISSUE #1092: Share My Work redirects to dashboard after login
### Descriptive summary

If a user is not logged in, Share My Work redirects them to login (Good) but then takes them to the dashboard. From here it is not obvious how to create a new work.

### Rationale

As a user when I click on 'Share My Work' I want it to be obvious how to create a new work after login.

mjgiarlo:
When an unauthenticated user clicks the Share Work button on the homepage, they should not be taken to the root dashboard page after logging in, because that is not where works are shared. Instead, redirect the user to the My Works page where the controls for creating works live.

Fixes #1092

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/homepage/index.html.erb
+++ b/app/views/hyrax/homepage/index.html.erb
@@ -17,7 +17,7 @@
           <% end %>
         <% end %>
       <% else %>
-        <%= link_to hyrax.dashboard_path,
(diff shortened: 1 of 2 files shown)


mjgiarlo:
:clap: @escowles

PR #1638, opened 2017-09-08
Fix test failure

elrayle:
Tests at collections-sprint are failing.

TEST: hyrax/admin/collection_types/_form_settings.html.erb for non-special collection types when collection_type.collections? is false renders the all settings with checkbox to enabled
FAILING: because .collections? was not stubbed to return false


elrayle:
Set against master instead of collections-sprint.  Closing.

PR #1639, opened 2017-09-08
fix failing tests

elrayle:
Tests at collections-sprint are failing.

TEST: hyrax/admin/collection_types/_form_settings.html.erb for non-special collection types when collection_type.collections? is false renders the all settings with checkbox to enabled
FAILING: because .collections? was not stubbed to return false


elrayle:
The failing test is a pre-existing random failure.  I entered an issue to try and help track that one down too at a later time.  See #1641

PR #1642, opened 2017-09-08
[WIP] when adding works to collection fails, post an error using flash

blancoj:
If an error occurs when calling MultipleMembershipChecker, trap it and report the error via flash

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -406,7 +406,8 @@ module Hyrax
 
         def add_members_to_collection(collection = nil)
           collection ||= @collection
-          collection.add_member_objects batch
+          error_messages = collection.add_member_objects batch
+          flash[:notice] = error_messages.join(" ") if error_messages.any?
         end
 
         def remove_members_from_collection
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -57,17 +57,19 @@ module Hyrax
 
     # Add member objects by adding this collection to the objects' member_of_collection association.
     def add_member_objects(new_member_ids)
+      error_messages = []
       Array(new_member_ids).each do |member_id|
         member = ActiveFedora::Base.find(member_id)
-        # @note Ideally, this would be surfaced as a warning in a flash
-        #       message. Because the member is found and saved in this model
-        #       method, I am not sure it's worth the effort to rejigger things
-        #       such that this information bubbles up to the controller and
-        #       view.
-        next if Hyrax::MultipleMembershipChecker.new(item: member).check(collection_ids: id, include_current_members: true)
+        # will return and error message as a string, otherwise it will return a nil
+        message = Hyrax::MultipleMembershipChecker.new(item: member).check(collection_ids: id, include_current_members: true)
+        if message
+          error_messages << message
+          next
+        end
         member.member_of_collections << self
         member.save!
       end
+      error_messages
     end
 
     def member_objects


app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -55,19 +55,22 @@ def add_members(new_member_ids)
       members << ActiveFedora::Base.find(new_member_ids)
     end
 
+
```
mjgiarlo: Remove blank line, pls.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -55,19 +55,22 @@ def add_members(new_member_ids)
...
     # Add member objects by adding this collection to the objects' member_of_collection association.
     def add_member_objects(new_member_ids)
+      error_messages = ["THis is just a test"]
```
mjgiarlo: I'm guessing you want to remove the test string here. :)

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -55,19 +55,22 @@ def add_members(new_member_ids)
...
       Array(new_member_ids).each do |member_id|
         member = ActiveFedora::Base.find(member_id)
-        # @note Ideally, this would be surfaced as a warning in a flash
-        #       message. Because the member is found and saved in this model
-        #       method, I am not sure it's worth the effort to rejigger things
-        #       such that this information bubbles up to the controller and
-        #       view.
-        next if Hyrax::MultipleMembershipChecker.new(item: member).check(collection_ids: id, include_current_members: true)
+        #will return and error message as a string, otherwise it will return a nil
```
mjgiarlo: How would you feel about removing this and adding YARD documentation immediately before this method declaration:

`# @return [nil, String] nil if no conflicts; an error message string if so`

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -55,19 +55,22 @@ def add_members(new_member_ids)
...
+        message = Hyrax::MultipleMembershipChecker.new(item: member).check(collection_ids: id, include_current_members: true)
+        if message
+           error_messages << message
+           next
+        end
         member.member_of_collections << self
         member.save!
       end
+      return error_messages
```
mjgiarlo: Don't need the `return` here since it's the last line of the method.

mjgiarlo: I think the return behavior we want is to return `error_messages` if it's got anything in it (check `error_messages.any?` for instance), otherwise return the result of `member.save!`.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -406,7 +406,8 @@ def process_member_changes
 
         def add_members_to_collection(collection = nil)
           collection ||= @collection
-          collection.add_member_objects batch
+          error_messages = collection.add_member_objects batch
+          flash[:notice] = error_messages.join(" ") if error_messages.any?
```
mjgiarlo: Rather than manually populating the flash here, I would try the following instead:

`collection.errors.add(:collections, error_messages) if error_messages.any?`

And rely on code higher up to detect errors and render the right flash message.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -57,17 +57,19 @@ def add_members(new_member_ids)
...
+      error_messages = []
       Array(new_member_ids).each do |member_id|
         member = ActiveFedora::Base.find(member_id)
-        # @note Ideally, this would be surfaced as a warning in a flash
-        #       message. Because the member is found and saved in this model
-        #       method, I am not sure it's worth the effort to rejigger things
-        #       such that this information bubbles up to the controller and
-        #       view.
-        next if Hyrax::MultipleMembershipChecker.new(item: member).check(collection_ids: id, include_current_members: true)
+        # will return and error message as a string, otherwise it will return a nil
```
mjgiarlo: See prior code review for suggestions around this file. (They were auto-hidden by GitHub prematurely.)

app/models/concerns/hyrax/collection_behavior.rb
blancoj: I think we always want to return error_messages. In the collection controller it will check  error_messages.any? in order to add to the flash.  Right?  That's the way I read it.

app/controllers/hyrax/dashboard/collections_controller.rb
jcoyne: also, why `notice`? Typically errors should be in the `:error` flavor of  flash message.

blancoj:
This issue is still open, but I'm not working on it.  I believe my approach here is not quite right, so I'm going to close the PR.

PR #1646, opened 2017-09-08
Single instance of Hyrax::DefaultMiddlewareStack

### LINKED ISSUE #1643: The DefaultMiddlewareStack is not customizable through Hyrax::CurationConcern.actor_factory
### Descriptive summary

Hyrax 2.0.0beta1

According to comments in `Hyrax::CurationConcern` the middleware stack can be customized like so, `Hyrax::CurationConcern.actor_factory.use MyCustomActor`. This no longer works this way

### Expected behavior

Customizations can be made to the middleware stack from a `config.to_prepare` block in `application.rb` as described above.

### Actual behavior

Customizing the middleware stack has no effect, because `Hyrax::CurationConcern.actor_factor` returns a new instance of the DefaultMiddlewareStack each call.

jonathandixon:
Fixes #1643

Changes `Hyrax::CurationConcern.actor_factory` to return the same instance of `Hyrax::DefaultMiddlewareStack` to allow customization of the actor stack.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/curation_concern.rb
+++ b/app/services/hyrax/curation_concern.rb
@@ -17,7 +17,7 @@ module Hyrax
     # is used.  Once it is used, it becomes immutable.
     # @return [ActionDispatch::MiddlewareStack]
     def self.actor_factory
-      @middleware_stack ||= Hyrax::DefaultMiddlewareStack.build_stack
+      @actor_factory ||= Hyrax::DefaultMiddlewareStack.build_stack
     end
 
     # A consumer of this method can inject a different factory
--- a/app/services/hyrax/curation_concern.rb
+++ b/app/services/hyrax/curation_concern.rb
@@ -17,7 +17,7 @@ module Hyrax
     # is used.  Once it is used, it becomes immutable.
     # @return [ActionDispatch::MiddlewareStack]
     def self.actor_factory
-      Hyrax::DefaultMiddlewareStack.build_stack
+      @middleware_stack ||= Hyrax::DefaultMiddlewareStack.build_stack
     end
 
     # A consumer of this method can inject a different factory
--- a/spec/services/hyrax/curation_concern_spec.rb
+++ b/spec/services/hyrax/curation_concern_spec.rb
@@ -7,4 +7,12 @@ RSpec.describe Hyrax::CurationConcern do
 
     it { is_expected.to be_kind_of Hyrax::Actors::TransactionalRequest }
   end
+
+  describe ".actor_factory" do
+    subject { described_class.actor_factory }
+
+    it "returns same ActionDispatch::MiddlewareStack instance" do
+      is_expected.to eq described_class.actor_factory
+    end
+  end
 end


spec/services/hyrax/curation_concern_spec.rb
```diff
@@ -7,4 +7,12 @@
 
     it { is_expected.to be_kind_of Hyrax::Actors::TransactionalRequest }
   end
+
+  describe ".actor_factory" do
+    subject { described_class.actor_factory }
+
+    it "returns same ActionDispatch::MiddlewareStack instance" do
+      is_expected.to eq described_class.actor_factory
```
jonathandixon: I wanted to test this by adding an actor the middleware stack then confirming it's still there after calling `actor_factory`, but by the time the test runs the middleware stack has been finalized and is immutable. I'm hoping just checking that actor factor returns the same instance is good enough.

app/services/hyrax/curation_concern.rb
```diff
@@ -17,7 +17,7 @@ class CurationConcern
     # is used.  Once it is used, it becomes immutable.
     # @return [ActionDispatch::MiddlewareStack]
     def self.actor_factory
-      Hyrax::DefaultMiddlewareStack.build_stack
+      @middleware_stack ||= Hyrax::DefaultMiddlewareStack.build_stack
```
jeremyf: It's a bit surprising that the ivar is not the same name as the method name. (Not a blocker, just a surprise).

spec/services/hyrax/curation_concern_spec.rb
jeremyf: Would checking for the instance variable be a more complete test? I know it violates encapsulation…but it seems to better test what you are after.

```ruby
expect(described_class.instance_variable_get('@middleware_stack').to be_a(WhateverInstanceIsExpected™)
```

app/services/hyrax/curation_concern.rb
jonathandixon: I have no issue with renaming this. I think I just went with `middleware_stack` because it's an instance of `ActionDispatch::MiddlewareStack`. I didn't know it was a convention to mirror the method name.

spec/services/hyrax/curation_concern_spec.rb
jonathandixon: I don't know the answer to the question, but I have no problem with making this change as well. Let me know if I should make these changes.

app/services/hyrax/curation_concern.rb
jeremyf: It's not quite a convention, but something I personally would've expected (not that I would want to directly access the instance variable). At times, when debugging, I'll do `object.instance_variables` to see what all is there. And method names equalling instance variables is a "nice to have".

jeremyf:
@jonathandixon Let me know if you are pushing up changes to this. If not, go ahead and merge.

jonathandixon:
@jeremyf Thanks for the review. I'll be pushing the changes you suggested.

PR #1647, opened 2017-09-08
Fix redirects for collection/work edit save and cancel

### LINKED ISSUE #1578: Collections-related forms: Forwarding action for Save/Cancel buttons
Sub-ticket of #1527, which is about creating a more consistent user experience for new/edit forms for collection types, admin sets, collections, and works.

## New pages
The current behavior for the Save button for the new collection type, admin set, collection, and work is good. There's a different flow for New Work than the other three, but that's based on how a work is created so it's okay.

For the Cancel button, updates are needed to be consistent among pages. The general recommendation is Cancel on a new page is to return to the appropriate index page.

- New Admin Set - okay as is
- New Collection Type - okay as is
- [x] New Collection - cancel currently goes to site root, should be updated to go to Collections index page
- [x] New Work - cancel currently goes to dashboard index, should be updated to go to Works index page

## Edit pages
The current behavior for the Save button for the edit collection type, admin set, collection, and work is inconsistent, with a couple of the pages going to a show page. Because the user could want to do any number of things after editing a tab in one of the pages (including more editing), I don't think we should send the user to a new context. When the user clicks "Save changes" on a tab on one of these pages, we should save the updates but remain on that tab. Note that in I think one case the current behavior is to return to the first tab, even if the user saves from another tab. I think we should remain on the _current_ tab. So the necessary changes are:

- [x] Save changes from Edit Admin Set:  stay on current tab
- [x] Save changes from Edit Collection Type:  stay on current tab
- [x] Save changes from Edit Collection:  stay on current tab
- [x] Save changes from Edit Work:  stay on current tab

For the Cancel button, updates are needed to be consistent among pages. The general recommendation is Cancel on an edit page is to return to the appropriate index page.

- Cancel from Edit Admin Set:  okay as is
- Cancel from Edit Collection Type:  okay as is
- [x] Cancel from Edit Collection:  currently goes to site root, should be updated to go to Collections index page
- [x] Cancel from Edit Work:  currently goes to dashboard index, should be updated to go to Works index page

bkeese:
Fixes #1578 

Changes page forwarding for save/cancel actions on new/edit forms for collection types, admin sets, collections, and works

Edit pages have tabbed content. Those tabs are coded as anchors on the page. Since a request's referrer never includes an anchor in it's url, this change adds the anchor as a hidden input on the form. The controllers then use that anchor param when constructing urls for redirection back to the appropriate tab.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/save_work/save_work_control.es6
+++ b/app/assets/javascripts/hyrax/save_work/save_work_control.es6
@@ -3,6 +3,7 @@ import { ChecklistItem } from './checklist_item'
 import { UploadedFiles } from './uploaded_files'
 import { DepositAgreement } from './deposit_agreement'
 import VisibilityComponent from './visibility_component'
+import tabifyForm from 'hyrax/tabbed_form'
 
 /**
  * Polyfill String.prototype.startsWith()
@@ -92,6 +93,7 @@ export default class SaveWorkControl {
     this.preventSubmit()
     this.watchMultivaluedFields()
     this.formChanged()
+    tabifyForm(this.form)
   }
 
   preventSubmit() {
--- a/app/assets/javascripts/hyrax/save_work/save_work_control.es6
+++ b/app/assets/javascripts/hyrax/save_work/save_work_control.es6
@@ -3,7 +3,6 @@ import { ChecklistItem } from './checklist_item'
 import { UploadedFiles } from './uploaded_files'
 import { DepositAgreement } from './deposit_agreement'
 import VisibilityComponent from './visibility_component'
-import tabifyForm from 'hyrax/tabbed_form'
 
 /**
  * Polyfill String.prototype.startsWith()
@@ -93,7 +92,6 @@ export default class SaveWorkControl {
     this.preventSubmit()
     this.watchMultivaluedFields()
     this.formChanged()
-    tabifyForm(this.form)
   }
 
   preventSubmit() {
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -266,13 +266,21 @@ module Hyrax
         end
       end
 
+      def update_referer
+        if params[:referer_anchor]
+          edit_polymorphic_path([main_app, curation_concern]) + params[:referer_anchor]
+        else
+          polymorphic_path([main_app, curation_concern])
+        end
+      end
+
       def after_update_response
         if curation_concern.file_sets.present?
           return redirect_to hyrax.confirm_access_permission_path(curation_concern) if permissions_changed?
           return redirect_to main_app.confirm_hyrax_permission_path(curation_concern) if curation_concern.visibility_changed?
         end
         respond_to do |wants|
-          wants.html { redirect_to [main_app, curation_concern] }
+          wants.html { redirect_to update_referer, notice: "Work \"#{curation_concern}\" successfully updated." }
           wants.json { render :show, status: :ok, location: polymorphic_path([main_app, curation_concern]) }
         end
       end
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -266,21 +266,13 @@ module Hyrax
         end
       end
 
-      def update_referer
-        if params[:referer_anchor]
-          edit_polymorphic_path([main_app, curation_concern]) + params[:referer_anchor]
-        else
-          polymorphic_path([main_app, curation_concern])
-        end
-      end
-
       def after_update_response
         if curation_concern.file_sets.present?
           return redirect_to hyrax.confirm_access_permission_path(curation_concern) if permissions_changed?
           return redirect_to main_app.confirm_hyrax_permission_path(curation_concern) if curation_concern.visibility_changed?
         end
         respond_to do |wants|
-          wants.html { redirect_to update_referer, notice: "Work \"#{curation_concern}\" successfully updated." }
+          wants.html { redirect_to [main_app, curation_concern], notice: "Work \"#{curation_concern}\" successfully updated." }
           wants.json { render :show, status: :ok, location: polymorphic_path([main_app, curation_concern]) }
         end
       end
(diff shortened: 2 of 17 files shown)


app/assets/javascripts/hyrax/save_work/save_work_control.es6
```diff
@@ -92,6 +92,23 @@ export default class SaveWorkControl {
     this.preventSubmit()
     this.watchMultivaluedFields()
     this.formChanged()
+    this.refererAnchor = this.addRefererAnchor()
```
mjgiarlo: There's a lot of repeated code across the es6 files that have been changed in this branch. Have you considered creating a new es6 class with the common code and using it in each of these other es6 files?

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -266,13 +266,21 @@ def after_create_response
         end
       end
 
+      def update_referer
+        if params['referer_anchor'].present?
```
mjgiarlo: Two things here:

1. Not sure you need the `#present?` here. If the key isn't present, I believe `params[]` will return `nil`.
2. What would you think of replacing the `'referer_anchor'` string in lines 270 and 271 with a symbol?

app/controllers/hyrax/admin/collection_types_controller.rb
```diff
@@ -44,7 +44,8 @@ def edit
 
     def update
       if @collection_type.update(collection_type_params)
-        redirect_to hyrax.edit_admin_collection_type_path(@collection_type), notice: t(:'hyrax.admin.collection_types.update.notification', name: @collection_type.title)
+        referer = hyrax.edit_admin_collection_type_path(@collection_type) + (params['referer_anchor'] || '')
```
mjgiarlo: I realize this is a simpler case than above, but I might like to see this handled in a new private method, `update_referer`. Also prefer to use a symbol for the hash key if possible.

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -55,7 +55,8 @@ def files
 
     def update
       if @admin_set.update(admin_set_params)
-        redirect_to hyrax.edit_admin_admin_set_path(@admin_set), notice: I18n.t('updated_admin_set', scope: 'hyrax.admin.admin_sets.form.permission_update_notices', name: @admin_set.title.first)
+        referer = hyrax.edit_admin_admin_set_path(@admin_set) + (params['referer_anchor'] || '')
```
mjgiarlo: I realize this is a simpler case than above, but I might like to see this handled in a new private method, `update_referer`. Also prefer to use a symbol for the hash key if possible.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -218,6 +218,14 @@ def uploaded_files(uploaded_file_ids)
           UploadedFile.find(uploaded_file_ids)
         end
 
+        def update_referer
+          if params['referer_anchor'].present?
```
mjgiarlo: See above for feedback on this method.

elrayle:
### UI Interactive Testing

#### Expectations:
* new -> cancel --  forwards to index for all types
* new -> save -- forwards to edit for all types (except maybe work) and shows flash message
* edit -> cancel -- forwards to show for all types
* edit -> save -- stays on edit for all types (except maybe work) and shows flash message
* edit -> change tab -> save -- stays on edit on the same tab for all types (except maybe work) and shows flash message

#### Actual:

√ - meets expectations
f - shows expected flash message

| Type | new -> cancel | new -> save | edit -> cancel | edit -> save | edit -> change tab -> save |
| ------| ---------------- | -------------- | ---------------- | -------------- | ------------------------------ |
| Admin Set |    √         |      √  - f        |           √           |          √  - f    |                 √  - f                    |
| Collection Type |  √   |      √  - f        |           √           |          √  - f    |                 √  - f                    |
| Collection |    √         |  doesn't forward to edit  - f  |    √   |   √  - doesn't flash    |  √  - doesn't flash                    |
| Work         | doesn't forward to index; route error | forwards to show - f   |       √      | stays on edit - doesn't flash    |  see note below  - f     |

Work -> edit -> save 
* doesn't save changes on the Relationships tab regardless of which tab you are on
* does save values on Descriptions tab regardless of which tab you are on
* does stay on the tab your were on when you clicked save

I asked @ggeisler about expected forwarding for works in Issue #1578

bkeese:
There were a couple of feature tests with assumptions about forwarding which I had to change with the changing requirements. Clearly all the cases aren't tested though.

elrayle:
Based on discussion with @ggeisler...

Works should...
- [ ] new -> cancel -- forward to index
- [ ] new -> save -- forward to show
- [ ] edit -> cancel -- forward to show
- [ ] edit -> save -- forward to show

bkeese:
I think the requirements are met now, with a couple of exceptions that I don't think are really in the scope of this issue. 1. Flash messages for collection edit/save being set, but are not showing up for some reason that I have been unable to quickly identify. 2. Changing values in one tab then saving from another tab doesn't consistently save the changes in non-active tabs. I don't think this is related to forwarding actions.

elrayle:
@bkeese Can you give examples of which object type (e.g. Collection Types, Admin Sets, Collections, Works) where data is not consistently saved in non-active tabs?  I want to be able to verify if this is an issue at master as well, or new with our sprint work.  Either way, I want to open issues for those before merging this work.

bkeese:
Admin Set Edit. Enter a new value for "Abstract or Summary" on the Description tab. Switch to "Release and Visibility" tab and click Save. When the page reloads, switch back to the Description tab. The new value for "Abstract or Summary" has been lost.

elrayle:
#### Work -> Edit -> Save does not forward to show

1. create work and save
1. edit work, change title, save

Should forward to work show.  Same if you are on Relationships tab and save.

#### Collection -> Edit -> Cancel does not go to admin show page

I wasn't specific enough on the Collection -> cancel forward.  It should go to the admin show page instead of the public landing page.  I added the route to the checkbox list above.

Issue #1507, opened 2017-08-17
Collections: Update code to handle delete action in index pages


elrayle:
### Descriptive summary

Update code to handle deleting collections from the collection index pages (Dashboard All/Managed/Your Collections).

NOTE: This issue was extracted from the uber issue for index view changes.  #1417

### Rationale

See model issue for rationale.

### Implementation

**Edit Files:**
* index for All/Your Collections - [app/views/hyrax/my/collections/index.html.erb](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/my/collections/index.html.erb)
* all other files are under
  * [app/views/hyrax/my/collections](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/my/collections) - specific to collections
  * [app/views/hyrax/my](https://github.com/samvera/hyrax/blob/master/app/views/hyrax/my) - shared with works

#### Expected user interactions

**NOTE: Dialogs are the same here and on admin show page.  Create partials for the dialogs and share with #1841.** 

Update the processing that occurs when the Delete action is selected for a collection. 

From the All Collections and My Collections pages, a user can...
- [x] show delete in action menu whether or not the user can delete
- when user selects delete action
  - if the user does not have edit_access to the collection
    - [x] show dialog stating "You do not have sufficient privileges to edit this document" with only option Ok
  - if user can delete
    - AND collection has items 
      - [x] AND `collection_type.required_membership? true` (i.e. Admin Set) --> show dialog stating "This collection is defined as an `<%= collection_type.title %>` and is not empty. To delete an `<%= collection_type.title %>`, you must first remove (delete or move to another collection) all items from the `<%= collection_type.title %>`. with only option Ok
      - [x] AND `collection_type.required_membership? false` (i.e. all other types) --> show confirmation dialog warning "Deleting this collection will permanently remove the collection from the repository. Items in the collection will remain in the repository. Are you sure you want to delete this collection?" AND allow user to Delete or Cancel
    - [x] AND collection has 0 items (for both Admin Set and Collections) --> show confirmation dialog `"Are you sure you want to delete this collection? This action cannot be undone."` AND allow user to Delete or Cancel
  - [x] if user chooses to continue, perform the delete
    - [x] if collection type != admin set, execute Collection#destroy action
    - [x] if collection type == admin set, execute AdminSet#destroy action

Modal messages...
- [ ] collection or admin set is empty - show mockup message and allow delete or cancel
[...]

PR #1655, opened 2017-09-12
Include admin sets in collection search builders.

### LINKED ISSUE #1587: Collections: Modify My Collections search builder to return all editable collections and admin sets
### Descriptive summary

My Collections list of collections should include all collections and admin sets for which the current user has edit_access.

### Rationale

Per UX redesign, My Collections will include all editable collections and admin sets.

### Expected behavior

The list of collections should include all collections and admin set for which the current user has edit_access.

### Actual behavior

The list of collections includes the collections the current user deposited.

#### Current Implementation
Search builder specified:  [app/controllers/hyrax/my/collections_controller.rb](https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/hyrax/my/collections_controller.rb#L4-L6)

Current Search Builder:  [Hyrax::My::CollectionsSearchBuilder](https://github.com/samvera/hyrax/blob/collections-sprint/app/search_builders/hyrax/my/collections_search_builder.rb)

which inherits from: [Hyrax::My::SearchBuilder](https://github.com/samvera/hyrax/blob/collections-sprint/app/search_builders/hyrax/my/search_builder.rb)

#### Proposed Implementation

Rewrite the search builder such that each of the following are included:  (pseudocode)

```
solr_parameters[:fq] << "edit_access_group_ssim:#{current_user.groups}"
solr_parameters[:fq] << "edit_access_person_ssim:#{current_user}"
solr_parameters[:fq] << "has_model_ssim:AdminSet"
solr_parameters[:fq] << "has_model_ssim:Collection"
```

The challenging bit is that the `edit_access_person_ssim` for AdminSets does not include the person who created the admin set.  Collections do put the depositor in the `edit_access_person_ssim`. 

And the thing that makes this even more challenging is that the AdminSet keeps the id of the person who created the AdminSet in the `creator_ssim`.  And the Collection uses `creator_tesim` for the conceptual creator of the collection.  This is not tied to a user in the system and does not grant the person identified in `creator_tesim` any access to the Collection.

`has_model_ssim:AdminSet` uses field `creator_ssim:user@email.com`
`has_model_ssim:Collection` uses field `creator_tesim:John Smith`

The question is, can solr docs other than Admin Sets have a `creator_ssim` field?  Is it safe to assume that if we find `creator_ssim:user@email.com` AND our current user is `user@email.com` then this is a positive match?

Question:   How to get solr to do an OR on a field like has_model_ssim?

1. Create a few admin sets with various group and user participants
1. Create a few collections with various group and user participants
1. Login as a user that has edit access to multiple collections and admin sets
1. Dashboard -> Repository Contents -> Collections

Verify that all collections and admin sets for which the current user has edit_access are included in the list.

### Related work

#1509 Collections: Update UI/UX of the listing of collections in the All/My Collections index

jonathandixon:
Fixes #1587

This PR updates the search builders to include admin sets for the "all collections" and "my collections" dashboard pages.

* Update the my collection search builder to include all collections and admin sets the current user has manage access for.
* Create a new search builder for the "All Collections" tab to include admin sets.
* Update the corresponding presenters and views to support Admin sets.

@samvera/hyrax-code-reviewers


elrayle:
@jonathandixon I did testing with this and it behaved as expected.  Which is a bit unexpected.  The code that updates the solr-doc with correct edit and read access is not merged.  I haven't fully looked through the code to see why it worked anyway.  I would expect the search builder to try and answer all questions from solr unless it absolutely can't.  Does it make sense to you that the following scenario should work now without the solr doc getting proper access update?

* non-admin user A 
    * creates two collections (e.g. Col1 and Col2)  
    * shares Col1 with user B by adding as manager
* non-admin user B logs in
    * goes to Dashboard -> Collections
    * sees Col1, but not Col2

Looking in solr, Col1 does NOT have edit access for user B.  Looking in permissions table, user B is a manager.

FYI, I have most of the code done that will fix the edit access updates in the solr doc.  I just need to add more testing.

elrayle:
Per discussion at standup, it looks like the search builder is querying the permission tables for manage access.  This is how the admin set search builder works.  I have a question in slack dev to ask why admin set search builder doesn't just use the solr doc edit access.

config/locales/hyrax.en.yml
```diff
@@ -574,11 +576,14 @@ en:
         action:
           add_collection_to_collection: "Add collection to another collection"
           add_to_collection:        "Add to Collection"
+          admin_set_confirmation:  "Deleting this administrative set from is permanent. Click OK to delete this administrative set, or Cancel to cancel this operation"
```
elrayle: Typo in... "Deleting this administrative set from is permanent."  
Should be?...  ""Deleting this administrative set from **the repository** is permanent."

app/views/hyrax/my/_collection_action_menu.html.erb
```diff
@@ -14,27 +14,29 @@
...
+        <%= link_to hyrax.dashboard_collection_path(id),
+              class: 'itemicon itemtrash delete-collection-button',
+              title: t("hyrax.dashboard.my.action.delete_collection") do %>
+          <%= t("hyrax.dashboard.my.action.delete_collection") %>
+        <% end %>
+      </li>
+      <li role="menuitem" tabindex="-1">
+        <%= link_to '#',
+          class: 'itemicon add-collection-to-collection',
+          title: t("hyrax.dashboard.my.action.add_collection_to_collection"),
```
elrayle: Depositors and managers can add collection to collection.

config/locales/hyrax.en.yml
elrayle: There should be a check before adding the delete action.  The modal for delete should either show the above message with the option to delete or cancel OR it should show a message "The administrative set is not empty.  Remove all items before deleting."

Checking on actual wording of the can't delete modal.

elrayle: I think you can leave the message for now.  The delete action and modal messages have a separate issue #1507.  I would just fix the typo so it makes sense as it is right now.

app/views/hyrax/my/_collection_action_menu.html.erb
```diff
@@ -14,20 +14,23 @@
...
+      </li>
+      <li role="menuitem" tabindex="-1">
+        <%= link_to hyrax.dashboard_collection_path(id),
+              class: 'itemicon itemtrash delete-collection-button',
+              title: t("hyrax.dashboard.my.action.delete_collection") do %>
+          <%= t("hyrax.dashboard.my.action.delete_collection") %>
+        <% end %>
+      </li>
+    <% end %>
+    <%# TODO: Only managers and depositors can add collections to collections. %>
```
jonathandixon: I added this as a TODO for now. It looks like `Hyrax::Collections::PermissionService` has a method stubbed that can handle this.

PR #1659, opened 2017-09-12
during migration set source_type in permission_template to 'admin_set'

### LINKED ISSUE #1580: Collections: Migrate existing permission templates to use new source_type column
### Descriptive summary

Refs #1576

Issue #1538 created a new `source_type` column in the permission_template table.  If someone is upgrading their existing Hyrax-based app, all existing permission templates should have their `source_type` column set to "admin_set".  

### Expected behavior

We might be able to just do some type of automatic inline migration at the time the user runs db:migrate.  Otherwise, a separate rake can can accomplish this.

blancoj:
Fixes #1580

### DIFF
--- a/db/migrate/20170821152307_add_source_type_to_permission_templates.rb
+++ b/db/migrate/20170821152307_add_source_type_to_permission_templates.rb
@@ -1,7 +1,16 @@
 class AddSourceTypeToPermissionTemplates < ActiveRecord::Migration[4.2]
   def change
     add_column :permission_templates, :source_type, :string
-    rename_column :permission_templates, :admin_set_id, :source_id
+    add_column :permission_templates, :source_id, :string
+
+    Hyrax::PermissionTemplate.find_each do |permission_template|
+      permission_template.source_type = 'admin_set'
+      permission_template.source_id = permission_template.admin_set_id
+      permission_template.save!
+    end
+
+    # If the db is sqlite, don't drop the column or you will get a FK error
+    # because of limitation in sqlite
+    remove_column :permission_templates, :admin_set_id, :string unless connection.adapter_name.downcase.starts_with?('sqlite')
   end
 end
-


elrayle:
I wasn't able to run the migrations with this change.  I'm not sure the column rename is working properly.  You might want to check with @hortongn 

I see this problem whether running with branch coll-1580 or collections-sprint.

To reproduce:
* clone hyrax master
* generate test app
* create default admin set
* edit Gemfile in test app
  * comment out line including local hyrax engine
  * add line `gem 'hyrax', github: 'samvera/hyrax', branch: 'coll-1580'`
* in test app, run...
```
bundle update hyrax
rails hyrax:install:migrations
rails db:migrate
```

Migration dies when it tries to update rename the admin_set_id column to source_id.  I only see this if there is data already in the table, which is why you have to create the default admin set before running the migrations.

```
== 20170913132713 AddSourceTypeToPermissionTemplates: migrating ===============
-- add_column(:permission_templates, :source_type, :string)
   -> 0.0004s
-- rename_column(:permission_templates, :admin_set_id, :source_id)
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

SQLite3::ConstraintException: FOREIGN KEY constraint failed: DROP TABLE "permission_templates"
```

blancoj:
There's an index that is created prior to the rename, so I think the thing to do is this:
    remove_index :permission_templates, :admin_set_id
    rename_column :permission_templates, :admin_set_id, :source_id
    add_index :permission_templates, :source_id, unique: true

I'll test this out.

db/migrate/20170821152307_add_source_type_to_permission_templates.rb
```diff
@@ -1,7 +1,16 @@
 class AddSourceTypeToPermissionTemplates < ActiveRecord::Migration[4.2]
   def change
     add_column :permission_templates, :source_type, :string
-    rename_column :permission_templates, :admin_set_id, :source_id
+    add_column :permission_templates, :source_id, :string
+
+    Hyrax::PermissionTemplate.find_each do |permission_template|
+      permission_template.source_type = 'admin_set'
```
jeremyf: My understanding is that this should usually be in class name format (e.g. "AdminSet") as we will often constantize the string (`permission_template.source_type.constantize` would return the class `AdminSet`).

db/migrate/20170821152307_add_source_type_to_permission_templates.rb
```diff
@@ -1,7 +1,16 @@
...
+      permission_template.source_id = permission_template.admin_set_id
+      permission_template.save!
+    end
+
+    # If the db is sqlite, don't drop the column or you will get a FK error
+    # because of limitation in sqlite
+    remove_column :permission_templates, :admin_set_id, :string unless connection.adapter_name.downcase.starts_with?('sqlite')
```
jeremyf: Not keen on conditional migrations, because the SQLite database will keep the admin_set_id (and the foreign key constraint).

Take a look at http://api.rubyonrails.org/v4.2/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-remove_foreign_key for removing foreign key constraints. I believe you'll need to first remove the constraint, then remove the column.

blancoj: In Sqlite, the FK constraint does not get removed.  In other words, `remove_foreign_key` has no affect in Sqlite.

db/migrate/20170821152307_add_source_type_to_permission_templates.rb
blancoj: Presently there seems to be other code that makes use of 'admin_set' for source_type. Not saying your proposal is not valid, it may just be a separate Issue?  https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/hyrax/admin/permission_template_accesses_controller.rb#L31

elrayle: Two thoughts here...
* The use of `if source_type == 'admin_set'` was brought in by the addition of the source_type column, so we can certainly change this to define source_type as either Collection or AdminSet.
* Use of admin_set and collection does not prevent use of constantize...
```
>> 'admin_set'.camelize.constantize
=> AdminSet
>> 'collection'.camelize.constantize
=> Collection
```
@jeremyf Considering these observations, does that lean you strongly in one direction or another.  Or is this minor.  It will take a bit of refactoring to change all the places where we are currently checking source_type.  That said, if you think this will be a problem down stream, it is better to bite that bullet now rather than later.

db/migrate/20170821152307_add_source_type_to_permission_templates.rb
jgondron: It seems like this will be (or should be?) such a rare case that it's reasonable to expect manual intervention to migrate the table themselves. If so, I would suggest just removing the conditional. Let it fail when this happens. Perhaps output some text about this being a known issue and what to do when it happens. 

To migrate it themselves, they would have to:
- Recreate permission_template_accesses without the 't.references :permission_template, foreign_key: true'
- Run the migration
- Recreate permission_template_accesses with the 't.references :permission_template, foreign_key: true'

All because sqlite doesn't support dropping FKC in an ALTER TABLE statement :(

That, or do a conditional migration that does all of this...

jgondron: If you need to see an example of the migration I'm describing above, I have a working one here https://github.com/samvera/hyrax/compare/collections-sprint...coll-1580-sqlite-fix. My suggestion is still to not do this, and just fix the data problem. I've created a PR that focuses on only the data https://github.com/samvera/hyrax/pull/2310. If that one is accepted, perhaps we can just close this one.

elrayle: @mjgiarlo @vantuyls I agree with @jgondron's accessment of the migration script.  There was one missing piece from the original migration script that does need to be addressed which left the source_type blank instead of setting it to 'admin_set'.  That is remedied by PR #2310.

So the proposal is to close this PR without merging, merge PR #2310, and provide migration instructions that explain the problem with sqlite.  

The instructions can include an explanation that sqlite is not recommended for production systems.  If they are using sqlite for development and need to migrate, then follow the instructions as outlined by @jgondron above.

jcoyne:
The `source_type` column has been removed. Do we still need this?

elrayle:
Source type was removed by PR #2382

PR #1688, opened 2017-09-16
Skip creating an ActionCable consumer if the user is not logged in

mjgiarlo:
This prevents unauthenticated users from attempting to create consumers. This cuts down on network traffic and keeps noise out of the server logs.

@samvera/hyrax-code-reviewers


jeremyf:
I'm going to take a stab at getting the specs to pass.

mjgiarlo:
Thanks, @jeremyf! Appreciate it.

PR #1706, opened 2017-09-21
Do not ask for the notifications_endpoint_path when compiling assets

mjgiarlo:
If the feature has been disabled in config, this results in a `NoMethodError`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/app.js.erb
+++ b/app/assets/javascripts/hyrax/app.js.erb
@@ -91,6 +91,7 @@ Hyrax = {
         // Do not create a consumer if user is not logged in
         if ($("meta[name='current-user']").length === 0)
             return;
+        <% if Hyrax.config.realtime_notifications? %>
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/hyrax/app.js.erb
```diff
@@ -91,7 +91,7 @@ Hyrax = {
         // Do not create a consumer if user is not logged in
         if ($("meta[name='current-user']").length === 0)
             return;
-        var consumer = ActionCable.createConsumer("<%= Hyrax::Engine.routes.url_helpers.notifications_endpoint_path %>");
+        var consumer = ActionCable.createConsumer("<%= Hyrax::Engine.routes.url_helpers.notifications_endpoint_path if Hyrax.config.realtime_notifications? %>");
```
mjgiarlo: This code is never interpreted when realtime notifications are off, because the function returns on line 93.

PR #1715, opened 2017-09-22
BasicMetadata: use full namespace for Location vocab

dunn:
If the application has independently defined a ControlledVocabulary module, this will cause an error:

```
NameError (uninitialized constant ControlledVocabularies::Location):

/Users/alex/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/bundler/gems/hyrax-a8873d1c3647/app/models/concerns/hyrax/basic_metadata.rb:29:in
`block in <module:BasicMetadata>'
```

@samvera/hyrax-code-reviewers


dunn:
I ran into another error that required quoting it as a string instead 😭 ; once I get a better handle on things I may open another PR.

mjgiarlo:
@dunn okie dokie, thx! `:tissue_box:`

Issue #1461, opened 2017-08-09
work.is_part_of metadata attribute usage in forms conflicts with AdminSet behavior


little9:
### Descriptive summary

In Hyrax 2.x, isPartOf is used for AdminSet membership. If you create a Hyrax app that uses
isPartOf and include it in a form, you're unable to save the work and if you edit an existing object
the property is replaced with a string. 

### Rationale

isPartOf is a common metadata term and it would be reasonable to think that institutions may want 
to have the term be editable in forms. 

### Expected behavior

Adding is_part_of to the work's form object behaves like any other attribute. 

### Actual behavior

After adding work.is_part_of to the form:

![29122234-8d8446c6-7cdf-11e7-8b44-f4ac6bdaf885](https://user-images.githubusercontent.com/4324761/29144719-ef16c52e-7d27-11e7-9525-caaf7ad5dfba.png)

Trying to save a new work that has a `work.is_part_of` attribute raises this error: 

```ActiveRecord::RecordNotFound in Hyrax::AudiosController#create
Couldn't find Sipity::Workflow with [WHERE `sipity_workflows`.`active` = ? AND `sipity_workflows`.`permission_template_id` IN (SELECT `permission_templates`.`id` FROM `permission_templates` WHERE `permission_templates`.`admin_set_id` = '#<ActiveTriples::Resource:0x007f78451adc08>')]
```
### Steps to reproduce the behavior

1. In the work's Form class add `is_part_of` to `self.terms` 
2. Save or edit a work 

### Related work

https://github.com/curationexperts/epigaea/issues/149

mjgiarlo:
Short of rejiggering the Work/AdminSet relationship to use a different predicate -- which I'm not sure we want to change in Hyrax 2.0.0, and Hyrax 3.0.0 shouldn't come out until summer 2018 -- could folks use a different predicate for their descriptive metadata?

Consulting the @samvera/hyrax-metadataists!

mark-dce:
@mjgiarlo yes, folks obviously could use a different predicate; however, none that we are aware of, stand out as obvious or clear candiates.  Hence this ticket.

[...]

PR #1720, opened 2017-09-25
Allow configuration of the predicate used for AdminSet inclusion

no-reply:
Allow configuration of the predicate used for AdminSet inclusion

Allow a custom predicate to be used to relate objects to their administrative sets. The default remains `dcterms:isPartOf`. This is the first step in a process to change the default predicate for this relation. A follow-up PR, to be merged after the Hyrax 2.0.0 release, will provide warnings for changed usage
and add migration tooling.

This addresses "Phase 1" of #1461 as described at https://github.com/samvera/hyrax/issues/1461#issuecomment-331205110.

* Add a configuration option to `Hyrax::Configuration` to set the admin set predicate.
* Use the configured predicate for the `AdminSet#members` `has_many` and `belongs_to` relations.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/admin_set.rb
+++ b/app/models/admin_set.rb
@@ -40,7 +40,7 @@ class AdminSet < ActiveFedora::Base
   end
 
   has_many :members,
-           predicate: ::RDF::Vocab::DC.isPartOf,
+           predicate:  Hyrax.config.admin_set_predicate,
            class_name: 'ActiveFedora::Base'
 
   before_destroy :check_if_not_default_set, :check_if_empty
--- a/app/models/concerns/hyrax/in_admin_set.rb
+++ b/app/models/concerns/hyrax/in_admin_set.rb
@@ -3,7 +3,7 @@ module Hyrax
     extend ActiveSupport::Concern
 
     included do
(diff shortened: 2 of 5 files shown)


mjgiarlo:
Looks good, @no-reply! I've taken the liberty of documenting this in the 2.0.0 release notes and the Hyrax management guide. Thanks.

mjgiarlo:
@no-reply Here it is:

```markdown
By default, Administrative Sets and their member objects use the [Dublin Core Terms](http://dublincore.org/documents/dcmi-terms/) predicate `isPartOf` to express the membership relationship. Hyrax also allows use of a custom predicate for this relationship. To customize the predicate, point the Hyrax configuration option `admin_set_predicate` at another RDF term or a URI object.

(Note: allowing customization is the first step towards changing the default predicate for this relation. After the 2.0.0 release, Hyrax will warn users about using `isPartOf` for this purpose and add migration tooling for production instances using `isPartOf`. The default will change in Hyrax 3.0.0.)
```

mjgiarlo:
@no-reply Feel free to merge when this goes :green_heart: (unless there are other blocking reviews by that point, of course). That will make sure this gets into 2.0.0.rc1.

PR #1721, opened 2017-09-25
Stop granting public read access to Admin Sets by default

mjgiarlo:
This reflects an outdated understanding of how Admin Sets would be used, and we no longer require this special-casing.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/factories/admin_sets.rb
+++ b/spec/factories/admin_sets.rb
@@ -22,8 +22,4 @@ FactoryGirl.define do
       with_permission_template false
     end
   end
-
-  trait :public do
-    read_groups ['public']
-  end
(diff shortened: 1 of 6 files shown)


spec/factories/admin_sets.rb
```diff
@@ -22,8 +22,4 @@
       with_permission_template false
     end
   end
-
-  trait :public do
```
mjgiarlo: I removed this trait in favor of using `read_groups` explicitly in specs that need public Admin Sets, primarily because the presence of this trait could communicate that we expect public Admin Sets to be a common case. I don't think we do.

Issue #1725, opened 2017-09-25
DeepIndexingService shouldn't inherit from BasicMetadataIndexer


dunn:
Hyrax provides `BasicMetadata` as a mixin for object models so that institutions can replace it with their own metadata model.  Consequently, the indexing behaviors should not assume that `BasicMetadata` is being used; but currently they do.

The indexer templates include a mixin corresponding to `BasicMetadata`, called `IndexesBasicMetadata`.  The comment indicates that it can be removed if `BasicMetadata` is also removed: https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/work/templates/indexer.rb.erb#L4-L6

What `IndexesBasicMetadata` does is override the indexer's `rdf_service` (setting it to `BasicMetadataIndexer` https://github.com/samvera/hyrax/blob/master/app/indexers/hyrax/basic_metadata_indexer.rb) in order to do some special handling of the properties defined by `BasicMetadata`.

There is a second, also optional, indexer mixin: `IndexesLinkedMetadata`, which also overrides the `rdf_service`, setting it to `DeepIndexingService` instead: https://github.com/samvera/hyrax/blob/master/app/indexers/hyrax/deep_indexing_service.rb.  This enhances the indexer by having it fetch remote labels.

Since these indexer mixins work by overriding `ActiveFedora::IndexingService`'s `rdf_service` method, they aren't intended to be used simultaneously (see https://github.com/samvera/hyrax/issues/1077).  But since some models using `BasicMetadata` will want deep indexing, the `DeepIndexingService` subclasses `BasicMetadataIndexer`.

What this means, finally, is that institutions that _want_ deep indexing but _don't_ want `BasicMetadata` are in a bind, since adding `IndexesLinkedMetadata` will pull in the `BasicMetadataIndexer` behavior, which assumes the existence of the `BasicMetadata` properties.  This will cause ingest errors (e.g.: https://gist.github.com/anonymous/c7324c7265fdb7ac00174d25f62d734a).

The functionality from `BasicMetadataIndexer` should be moved out of an `rdf_service` override and into the indexer class(es) themselves, so that `DeepIndexingService` doesn't have to pull in `BasicMetadata`

### Steps to reproduce the behavior

1. Remove `BasicMetadata` from a model but leave `IndexesLinkedMetadata` in its indexer.
1. Try to ingest an object.

mjgiarlo:
Thank you for writing this up and taking it on, @dunn!

Issue #1077, opened 2017-05-26
Generated indexer contradicts itself


atz:
The [generator](https://github.com/samvera-labs/hyrax/blob/master/lib/generators/hyrax/work/templates/indexer.rb.erb) includes both:
```ruby
  include Hyrax::IndexesBasicMetadata
```
and:
```ruby
  include Hyrax::IndexesLinkedMetadata
```

That is useless, since [IndexesBasicMetadata](https://github.com/samvera-labs/hyrax/blob/master/app/indexers/concerns/hyrax/indexes_basic_metadata.rb#L5-L7) and [IndexesLinkedMetadata](https://github.com/samvera-labs/hyrax/blob/master/app/indexers/concerns/hyrax/indexes_linked_metadata.rb#L5-L7) both override `rdf_service` differently (`BasicMetadataIndexer` vs. `DeepIndexingService`).  And that's ALL they do.  Only the last one "wins", so there is no point including the first one.  

MOREOVER, why is this stuff buried in a module that the user has no understanding of anyway, just to hide the one thing the module actually does?  We should just put this in the generator:
```ruby
def rdf_service
  DeepIndexingService # Unlike BasicMetadataIndexer, fetches remote labels for based_near
end
```

And remove `DeepIndexingService` completely.  (Further work could be done to eradicate `BasicMetadataIndexer`).

jcoyne:
If you don't want to deal with linked metadata, then the procedure is to comment out or remove `include Hyrax::IndexesLinkedMetadata` and then it still works.

atz:
@jcoyne That isn't different than having a commented out redefinition of the method.  Or having it defined twice, which would make more obvious that what we have is backwards.  This is just bad form with obfuscation to hide hit.

no-reply:
This seems fine, and functions to provide the fallback @jcoyne described. Refactors are welcome (especially ones that use more flexible patterns than multiple/module inheritance), but I'm going to close this as a tracked issue.

PR #1729, opened 2017-09-26
separate deep indexing from Hyrax's basic metadata schema

dunn:
Fixes GH-1725.
Fixes GH-1077.

* move the functionality from BasicMetadataIndexer to BasicMetadata
* have DeepIndexingService inherit directly from AF::RDF::IndexingService
* override `rdf_service` directly instead of via a mixin (#1077)

@samvera/hyrax-code-reviewers


app/indexers/concerns/hyrax/indexes_basic_metadata.rb
```diff
@@ -1,9 +0,0 @@
-module Hyrax
-  # This module can be mixed in on an indexer in order to index the basic metadata fields
-  module IndexesBasicMetadata
```
dunn: I guess I should deprecate these instead of removing them outright, huh.

dunn:
I got some spurious Selenium failures locally: https://gist.github.com/anonymous/d79e9c281e75154619ac249d8cd5522b

Am I just running the wrong version of Chrome?

mjgiarlo:
@dunn Very possibly, yes. I believe @laritakr and some of the folks on the Collection Extensions sprints (@elrayle ?) saw similar a couple weeks ago. Tagging them in hopes they can help w/ the chromedriver/selenium errors.

mjgiarlo:
Once this is merged, could you update the 2.0.0 release notes to let folks know of this change and how to deal with it? Thx!

laritakr:
@dunn Yes, I had issues as well. I had to force-upgrade the latest version of chromedriver. 

The chromedriver-helper readme explains how to do it:

1. delete the directory $HOME/.chromedriver-helper
2. run chromedriver-update

mjgiarlo:
@dunn @laritakr Would one of y'all mind adding a blurb for this to the Hyrax dev guide? https://github.com/samvera/hyrax/wiki/Hyrax-Development-Guide Thank you!

dunn:
Done https://github.com/samvera/hyrax/wiki/Hyrax-Development-Guide/_compare/b26b50ee5c56334c839196e2d2c3cfd4adbc9f80...45d0a6a714a86a84498238e60f7dd46de0c95160

geekscruff:
What's the benefit of doing the indexing of basic metadata properties this way rather than with 'index as' within the property statements? Am trying to keep some of my own code aligned with Hyrax so curious about this.

dunn:
@geekscruff looks like it used to be that way: https://github.com/samvera/hyrax/commit/3a6c072f6dfad1bc8a7ec3c9d6c81f441e1ce6e9#diff-581f139fc4a6b844dd0b5fa1d9f5f213

I'm not sure what the relative merits are; @jcoyne ?

jcoyne:
@geekscruff I think it  was just to keep modeling code separate  from the indexing code.  Thus each class has a single responsibility.

dunn:
Any thoughts on this, @carrickr @jcoyne ?

geekscruff:
@dunn Hello, FWIW I do see the benefit of separating the indexing logic, especially with the potential move to valkyrie on the horizon. But I think that separate indexing logic should stay in a separate class, de-coupled from the deep indexing service (as per the topic of the PR).

carrickr:
I would agree with @geekscruff separate indexing logic for this

dunn:
So maybe just rename the mixin as something like `BasicMetadataIndexingBehavior` and include it in `BasicMetadata`?

dunn:
How would Valkyrie affect this? Is the current indexing code Fedora-specific?

geekscruff:
Could it just become part of IndexesBasicMetadata? 

I'm only just getting my head around Valkyrie, but the model and indexing behaviour are very separate, so I think it's cleaner to keep this separation.

dunn:
My concern is if this would be a mixin that's only designed to be included in this specific class.

geekscruff:
IndexesBasicMetadata is included in the generated indexer for a new work so I thought this was a sensible place for it? But I haven't got strong feelings about this and don't want to stall your PR!

app/indexers/concerns/hyrax/indexes_basic_metadata.rb
```diff
@@ -1,9 +1,10 @@
 module Hyrax
-  # This module can be mixed in on an indexer in order to index the basic metadata fields
   module IndexesBasicMetadata
-    # We're overriding a method from ActiveFedora::IndexingService
-    def rdf_service
-      BasicMetadataIndexer
-    end
+    extend Deprecation
+
+    Deprecation.warn(
```
jcoyne: Can you put this in an `included` block?  Otherwise it'll trigger when the code is eager loaded in production mode.

app/models/concerns/hyrax/basic_metadata.rb
```diff
@@ -36,6 +36,59 @@ module BasicMetadata
       class_attribute :controlled_properties
       self.controlled_properties = [:based_near]
       accepts_nested_attributes_for :based_near, reject_if: id_blank, allow_destroy: true
+
```
jcoyne: This is not a good place to include this logic. The intention here is that BasicMetadata is something one could remove from their models so they can add their own metadata.  We don't want to add any other logic to this module.

app/indexers/concerns/hyrax/indexes_basic_metadata.rb
dunn: Oh, sure.

app/models/concerns/hyrax/basic_metadata.rb
dunn: My point was that if you want to remove BasicMetadata, you're gonna want to remove this as well; this PR is needed because they are tightly coupled like that.

But I'll put it in a module if that gets things moving.

jcoyne:
@dunn can you explain where you’re seeing coupling? I’m not I see.

dunn:
@jcoyne these lines:
```ruby
    class_attribute :stored_and_facetable_fields, :stored_fields, :symbol_fields
    self.stored_and_facetable_fields = %i[resource_type creator contributor keyword publisher subject language based_near]
    self.stored_fields = %i[description license rights_statement date_created identifier related_url bibliographic_citation source]
self.symbol_fields = %i[import_url]
```
Those are in the BasicMetadataIndexer and assume the existence of those fields from BasicMetadata.

I suppose we could move these lines into BasicMetadata, but either way anyone who wants to use BasicMetadataIndexer but not BasicMetadata would have to redefine all those attributes.

jcoyne:
@dunn Okay there is coupling, but it is loose coupling. Those are the default indexing directions, but as they are variables they should be configurable. So you could get rid of BasicMetadata and then set `BasicMetadataIndexer.stored_and_facetable_fields = %i[my_field1 my_field2]`, etc.

app/indexers/hyrax/collection_with_basic_metadata_indexer.rb
```diff
@@ -1,5 +1,11 @@
 module Hyrax
+  extend Deprecation
+
   class CollectionWithBasicMetadataIndexer < CollectionIndexer
-    include Hyrax::IndexesBasicMetadata
+    Deprecation.warn(
```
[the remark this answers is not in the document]
dunn: You mean monkeypatch it in `config/initializers`?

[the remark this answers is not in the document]
jcoyne: no I mean:

```ruby
def initialize(*)
  super
  Deprecation.warn(self, 'message here')
end
```

dunn:
Re:
```ruby
      # This can be replaced by a simple merge once
      # https://github.com/samvera/active_fedora/pull/1227
      # is available to us
      # @param [ActiveFedora::Indexing::Map] first
      # @param [Hash] second
      def self.merge_config(first, second)
        first_hash = first.instance_variable_get(:@hash).deep_dup
        ActiveFedora::Indexing::Map.new(first_hash.merge(second))
      end
```
Can we pull the trigger on that?  Looks like it was part of 11.2.0, and my lockfile for Hyrax shows `~> 11.5`.

dunn:
So here are the methods from `BasicMetadataIndexer` 
```ruby
      # This method is passed to {ActiveFedora::RDF::IndexingService}
      # @return [ActiveFedora::Indexing::Map]
      def self.index_config
        merge_config(
          merge_config(
            merge_config(super, stored_and_facetable_index_config),
            stored_searchable_index_config
          ),
          symbol_index_config
        )
      end

      # This can be replaced by a simple merge once
      # https://github.com/samvera/active_fedora/pull/1227
      # is available to us
      # @param [ActiveFedora::Indexing::Map] first
      # @param [Hash] second
      def self.merge_config(first, second)
        first_hash = first.instance_variable_get(:@hash).deep_dup
        ActiveFedora::Indexing::Map.new(first_hash.merge(second))
      end

      def self.stored_and_facetable_index_config
        stored_and_facetable_fields.each_with_object({}) do |name, hash|
          hash[name] = index_object_for(name, as: [:stored_searchable, :facetable])
        end
      end

      def self.stored_searchable_index_config
        stored_fields.each_with_object({}) do |name, hash|
          hash[name] = index_object_for(name, as: [:stored_searchable])
        end
      end

      def self.symbol_index_config
        symbol_fields.each_with_object({}) do |name, hash|
          hash[name] = index_object_for(name, as: [:symbol])
        end
      end

      def self.index_object_for(attribute_name, as: [])
        ActiveFedora::Indexing::Map::IndexObject.new(attribute_name) do |idx|
          idx.as(*as)
        end
      end
```
@jcoyne your suggestion is to make a separate mixin for them, and have them use whatever values are set for `stored_and_facetable_fields`, `stored_fields`, `symbol_fields`, by the including class, right?

What should that mixin be called? If these indexer methods are generally useful, should we move them down to AF/Valkyrie?

jcoyne:
> your suggestion is to make a separate mixin for them

I'm not sure I'm suggesting that.  I guess I don't understand why that code needs changing.

> What should that mixin be called? If these indexer methods are generally useful, should we move them down to AF/Valkyrie?

It could be moved to AF, but you'd be blocked by https://github.com/samvera/hyrax/issues/2203.  Valkyrie has it's own way of indexing and doesn't use this code at all.

dunn:
@jcoyne this is what I meant, and which I thought you were asking for: https://github.com/samvera/hyrax/pull/1729/commits/cf28a9d38028fe5af290d348c708bf0c98b0e16c

dunn:
pinging @jcoyne and @mjgiarlo on this

straleyb:
Im really interested in having this, but could it be extended a bit further? Some fields are `stored_sortable` such as the date created. Having an option to make any of the fields sortable would be insanely helpful.

dunn:
@straleyb that's why I originally put the indexing behavior into the BasicMetadata concern; it's currently not really written to be used with anyone else's metadata model.  But at Justin's request I separated it out again: https://github.com/samvera/hyrax/pull/1729/commits/6b6d19d96309c22e731950ac5e504a71962c04cd

PR #1734, opened 2017-09-27
for admin_set don't allow title to be editable.

### LINKED ISSUE #1732: Collection Types: Don't allow users to change the name of Admin Sets type
### Descriptive summary

On Collection Types edit page for Admin Sets only, don't allow the Type Name to be edited.

### Rationale

The terminology for Admin Sets in the UI is hardcoded in the Admin Set new/edit forms and show pages.  There would need to be additional work to drive those off the collection type name.

### Expected behavior

Type Name for Admin Sets cannot be changed.  Description can be changed.

### Actual behavior

Type Name for Admin Sets can be changed.

1. Login as an admin
1. Dashboard -> Settings -> Collection Types -> click Edit for Admin Sets -> Description tab

Verify that the Type Name field is not editable.

blancoj:
Fixes #1732

### DIFF
--- a/app/views/hyrax/admin/collection_types/_form.html.erb
+++ b/app/views/hyrax/admin/collection_types/_form.html.erb
@@ -15,13 +15,23 @@
   <% form_url = @form.id.nil? ? hyrax.admin_collection_types_path : hyrax.admin_collection_type_path(@form.id) %>
   <%= simple_form_for @form, url: form_url, as: :collection_type do |f| %>
     <div class="tab-content">
-      <div id="metadata" class="tab-pane active">
-        <div class="panel panel-default labels">
-          <div class="panel-body">
-            <%= render 'form_metadata', f: f %>
+      <% if @collection_type.admin_set? %>
+        <div id="metadata" class="tab-pane active">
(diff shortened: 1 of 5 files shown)


app/views/hyrax/admin/collection_types/_form.html.erb
```diff
@@ -15,13 +15,23 @@
   <% form_url = @form.id.nil? ? hyrax.admin_collection_types_path : hyrax.admin_collection_type_path(@form.id) %>
   <%= simple_form_for @form, url: form_url, as: :collection_type do |f| %>
     <div class="tab-content">
-      <div id="metadata" class="tab-pane active">
-        <div class="panel panel-default labels">
-          <div class="panel-body">
-            <%= render 'form_metadata', f: f %>
+      <% if @collection_type.machine_id == "admin_set" %>
```
jeremyf: Push this check to the collection type model. In the collection type model, we should the method `@collection_type.admin_set?`. This way we have a consistent mechanism for asking a collection type if it is an admin_set (which has specific behavior in the current implementation).

PR #1743, opened 2017-09-28
ContentBlockHelperBehavior#display_content_block? guard against nils

dunn:
If you don't have a content block defined, you can get an error out of Hyrax:
```
ActionView::Template::Error (undefined method `value' for nil:NilClass
Did you mean?  evaluate):
    1: <% if display_content_block? @marketing_text %>
    2:   <div class="home_marketing_text">
    3:     <%= displayable_content_block @marketing_text, class: 'col-sm-offset-2 col-sm-9 col-md-offset-1 home_marketing_text' %>
    4:   </div>

/Users/alex/clones/hyrax/app/helpers/hyrax/content_block_helper_behavior.rb:9:in `display_content_block?'
```

### DIFF
--- a/app/helpers/hyrax/content_block_helper_behavior.rb
+++ b/app/helpers/hyrax/content_block_helper_behavior.rb
@@ -1,12 +1,12 @@
 module Hyrax
   module ContentBlockHelperBehavior
     def displayable_content_block(content_block, **options)
-      return if content_block.value.blank?
+      return unless display_content_block? content_block
       content_tag :div, raw(content_block.value), options
     end
 
     def display_content_block?(content_block)
-      content_block.value.present?
+      content_block.present? && content_block.value.present?
     end
   end
 end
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@dunn Hmm, under what conditions are Hyrax's default content blocks (like marketing text) nils?

dunn:
Under these conditions, I guess: https://github.com/ucsblibrary/geniza/tree/ADRL

Are there supposed to be default values defined?  Where should they be?

dunn:
That works for me.

mjgiarlo:
Perhaps the most intention-revealing form of this change may look like `content_block.respond_to?(:value) && content_block.value.present?` but I'm also happy with what @jonathandixon suggested and this somewhat simplified version: `content_block && content_block.value.present?`

I don't care *too* much by what form this takes, though, to be honest!

dunn:
Test added.

PR #1761, opened 2017-10-04
Spawn permission-copying jobs when ingesting via BrowseEverything

### LINKED ISSUE #1754: Item added to existing embargoed work using Browse cloud files (Box) is private
### Descriptive summary
(Chrome/Windows) Item added to existing embargoed work using Browse cloud files (Box as source) is private instead of embargoed.

### Expected behavior
When an item is added to an existing embargoed work using the Add files button, the newly added item maintains the same embargo visibility as the work. This behavior would also be expected for an item added using the Browse cloud files option.

### Actual behavior
The item added from Box has a visibility of private.

From a Work Edit, Files tab
1 select the “Browser cloud files” button
2 select a cloud source "Box"
3 select the Connect to the cloud source button
4 select a file, submit. Check deposit agreement, and save.
5 View visibility of items for the embargoed work. Items added via cloud files method will have Private visibility instead of embargoed visibility.

mjgiarlo:
Fixes #1754
Refs https://github.com/curationexperts/nurax/issues/81

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/actors/hyrax/actors/file_set_actor_spec.rb
+++ b/spec/actors/hyrax/actors/file_set_actor_spec.rb
@@ -136,10 +136,14 @@ RSpec.describe Hyrax::Actors::FileSetActor do
   describe '#create_content when from_url is true' do
     before do
       expect(JobIoWrapper).to receive(:create_with_varied_file_handling!).with(any_args).once.and_call_original
+      allow(VisibilityCopyJob).to receive(:perform_later).with(file_set.parent).and_return(true)
+      allow(InheritPermissionsJob).to receive(:perform_later).with(file_set.parent).and_return(true)
     end
 
-    it 'calls ingest_file' do
+    it 'calls ingest_file and kicks off jobs' do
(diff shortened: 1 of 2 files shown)


spec/actors/hyrax/actors/file_set_actor_spec.rb
```diff
@@ -136,10 +136,14 @@
   describe '#create_content when from_url is true' do
     before do
       expect(JobIoWrapper).to receive(:create_with_varied_file_handling!).with(any_args).once.and_call_original
+      allow(VisibilityCopyJob).to receive(:perform_later).with(file_set.parent).and_return(true)
+      allow(InheritPermissionsJob).to receive(:perform_later).with(file_set.parent).and_return(true)
```
no-reply: I think we should plan to migrate job queuing tests to the `have_equeued` matcher. It looks like we're using stubbing across the board, so :+1: to this for now.

PR #1764, opened 2017-10-05
Use correct route for collection delete in show actions.

### LINKED ISSUE #1750: Delete button doesn't work in collection show page
### Descriptive summary

If you click the red delete button on collection show page you'll get an error message and the collection is not deleted.

Selecting delete from the actions drop down menu in the dashboard collection list does work and delete the collection.

This happens on Hyrax v2.0.0.beta4.

### Rationale

The delete button should work the same as deleting from the dashoard list.

### Expected behavior

No error message, collection should be deleted.

### Actual behavior

Rails Routing error page and error message "No route matches [DELETE] "/collections/0z708w40c". The collection is not deleted.

1. Create a new collection
2. Go to dashboard, then select Collections on the menu and in the list click the collection you just created
3. On the page that opens (http://localhost:3000/dashboard/collections/0z708w40c?locale=en), click the red delete button and confirm deletion in the popup that opens

ojlyytinen:
Fixes #1750

Use the correct route in collection show actions. The collection_path route doesn't exist for delete method and resulted in a routing error page. The route I changed it to is the same that's used in app/views/hyrax/my/_collection_action_menu.html.erb which works.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/dashboard/collections/_show_actions.html.erb
+++ b/app/views/hyrax/dashboard/collections/_show_actions.html.erb
@@ -14,7 +14,7 @@
 
     <% if can? :destroy, presenter.solr_document %>
       <%= link_to t('hyrax.collection.actions.delete.label'),
-                  collection_path(presenter),
+                  hyrax.dashboard_collection_path(presenter),
                   title: t('hyrax.collection.actions.delete.desc'),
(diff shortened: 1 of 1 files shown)


mjgiarlo:
Thanks, @ojlyytinen!

PR #1770, opened 2017-10-05
set_read_groups is not in this codebase

jcoyne:
This test appears to be testing code that is not in this codebase.


mjgiarlo:
@jcoyne needs rebase

PR #1777, opened 2017-10-06
[WIP] Valkyrie

jcoyne:


app/actors/hyrax/actors/apply_order_actor.rb
```diff
@@ -58,6 +56,16 @@ def apply_order(curation_concern, new_order)
           curation_concern.list_source.order_will_change!
           true
         end
+
+        def find_resource(id)
+          query_service.find_by(id: Valkyrie::ID.new(id.to_s))
```
jeremyf: This looks like `Valkyrie` should be responsible for casting the given id to a `Valkyrie:ID`. This method does not know that the given query_service is Valkyrie. As such I believe the `find_by` method in Valkyrie should be responsible for casting to an id.

jcoyne:
Closing this in favor of #2178

Issue #1775, opened 2017-10-06
Collections factory does not behave as expected without a permission template


laritakr:
### Descriptive summary

The Collections factory allows you to create a collection without a permission template. Adding nested indexing identified a bug in the permissions given to a collection via this method. 

Without a permission template, at some point during creating the collection, edit permissions are assigned to the creator. But during the reindexing, ActiveFedora::Base.find(id).to_solr does not return permissions related fields, so the edit permissions are removed during nested indexing. 

### Rationale

To ensure that tests are valid, we need to ensure that the factory is producing an object in the same way that the UI does.

### Expected behavior

A collection created via the factory will behave in the same manner as a collection created through the UI.

### Actual behavior

When a collection is created without a permission template, while using nested indexing, the creator loses access to the collection. 

### Steps to reproduce the behavior

1. Create a spec to test a collection through the factory: 
`      let(:user) { create(:user) }
      let(:collection) { create(:named_collection, user: user) }`
2.  Test without nested indexing passes 
      `expect(user.can?(:edit, collection)).to be true`
3. Add `with_nested_reindexing: true`
4. Test with nested indexing fails

Breakpoints in the nesting_index_adapter method `self.write_document_attributes_to_index_layer` show that edit_access_person_ssim is present prior to re-saving the solr document, and absent after. 

### Related work

Nesting indexing PRs: 
- Nested indexing fix: https://github.com/samvera/hyrax/pull/1736
- Add indexing for nested collections: https://github.com/samvera/hyrax/pull/1668

elrayle:
Fixed by PR #1781

PR #1781, opened 2017-10-06
Update nested reindexing

laritakr:
Resolves https://github.com/samvera/hyrax/issues/1775
Resolves https://github.com/samvera/hyrax/issues/1785

Also removes ActiveFedora caching from the nesting index adapter.
Nested reindexing occurs midway through the transaction, so AF was
attempting to use the cache. However, this resulted in the updated solr
document not including all of the AF updates.

Reindexing changes have been separated into this one PR to isolate the
indexing changes from other changes.

In addition, unpins the solr version since incompatibilities have been
previously resolved.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.rubocop_fixme.yml
+++ b/.rubocop_fixme.yml
@@ -29,6 +29,7 @@ Metrics/ModuleLength:
     - 'app/models/concerns/hyrax/ability.rb'
     - 'app/services/hyrax/workflow/permission_query.rb'
     - 'spec/services/hyrax/workflow/permission_query_spec.rb'
+    # TODO:  extract CollectionAccessControls or something, so we don't have to skip this check?
     - 'app/models/concerns/hyrax/collection_behavior.rb'
 
 Metrics/MethodLength:
--- a/.rubocop_fixme.yml
+++ b/.rubocop_fixme.yml
@@ -29,6 +29,7 @@ Metrics/ModuleLength:
     - 'app/models/concerns/hyrax/ability.rb'
     - 'app/services/hyrax/workflow/permission_query.rb'
     - 'spec/services/hyrax/workflow/permission_query_spec.rb'
+    - 'app/models/concerns/hyrax/collection_behavior.rb'
 
 Metrics/MethodLength:
   Exclude:
--- a/app/services/hyrax/adapters/nesting_index_adapter.rb
+++ b/app/services/hyrax/adapters/nesting_index_adapter.rb
@@ -19,14 +19,14 @@ module Hyrax
       def self.find_preservation_parent_ids_for(id:)
         # Not everything is guaranteed to have library_collection_ids
         # If it doesn't have it, what do we do?
-        ActiveFedora::Base.uncached do
+        fedora_object = ActiveFedora::Base.uncached do
           fedora_object = ActiveFedora::Base.find(id)
+        end
 
-          if fedora_object.respond_to?(:member_of_collection_ids)
-            fedora_object.member_of_collection_ids
-          else
-            []
-          end
+        if fedora_object.respond_to?(:member_of_collection_ids)
+          fedora_object.member_of_collection_ids
+        else
+          []
         end
       end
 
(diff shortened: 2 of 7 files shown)


app/services/hyrax/adapters/nesting_index_adapter.rb
```diff
@@ -19,11 +19,14 @@ def self.find_preservation_document_by(id:)
       def self.find_preservation_parent_ids_for(id:)
         # Not everything is guaranteed to have library_collection_ids
         # If it doesn't have it, what do we do?
-        fedora_object = ActiveFedora::Base.find(id)
-        if fedora_object.respond_to?(:member_of_collection_ids)
-          fedora_object.member_of_collection_ids
-        else
-          []
+        ActiveFedora::Base.uncached do
+          fedora_object = ActiveFedora::Base.find(id)
```
elrayle: What is the performance impact of using `uncached`?

elrayle:
See bug #1785 BUG Collections: use of updated_index causes participants not to save on Add.  Does it have bearing on this PR?

app/services/hyrax/adapters/nesting_index_adapter.rb
laritakr: This is most likely an AF bug, but I'm not sure how to duplicate it lower in the stack. The indexing here is happening after the save and updating other indexing, but AF was retrieving the value from the cache and undoing the changes we just made.

laritakr: And yes I suspect it may be the reason for the problem in 1785. It was causing an error when removing a work from a collection, but it could have much broader implications as well.

laritakr:
I suspect that is the cause of the error in 1785, but I can't be certain without doing similar debugging that I did to track down the error during removing a work from a collection. I don't want to work on 1785 until after this is merged and we are sure the problem still exists.

app/services/hyrax/adapters/nesting_index_adapter.rb
laritakr: After researching #1785 bug, I realized the caching isn't an AF bug, it's just a result of the point in the process where the nested indexing was happening (mid-transaction for the update process). So removing the caching here is necessary, but not an AF caching bug.

app/services/hyrax/adapters/nesting_index_adapter.rb
```diff
@@ -65,14 +68,16 @@ def self.each_perservation_document_id_and_parent_ids(&block)
...
       def self.write_document_attributes_to_index_layer(id:, parent_ids:, ancestors:, pathnames:)
-        solr_doc = ActiveFedora::Base.find(id).to_solr # What is the current state of the solr document
-
-        # Now add the details from the nesting indexor to the document
-        solr_doc[solr_field_name_for_storing_ancestors] = ancestors
-        solr_doc[solr_field_name_for_storing_parent_ids] = parent_ids
-        solr_doc[solr_field_name_for_storing_pathnames] = pathnames
-        ActiveFedora::SolrService.add(solr_doc, commit: true)
-        solr_doc
+        ActiveFedora::Base.uncached do
```
jcoyne: I think having uncached around the whole block might be confusing.  It's only affecting line 72, right?  Can we minimize the scope?

app/services/hyrax/adapters/nesting_index_adapter.rb
jcoyne: Is the uncached needed around just the find or just the call to member_of_collection_ids or both?

.rubocop_fixme.yml
```diff
@@ -29,6 +29,7 @@ Metrics/ModuleLength:
     - 'app/models/concerns/hyrax/ability.rb'
     - 'app/services/hyrax/workflow/permission_query.rb'
     - 'spec/services/hyrax/workflow/permission_query_spec.rb'
+    - 'app/models/concerns/hyrax/collection_behavior.rb'
```
jcoyne: How about extracting CollectoinAccessControls or something, so we don't have to skip this check?

laritakr: Making this a TODO so we can get the indexing issues resolved ASAP

app/services/hyrax/adapters/nesting_index_adapter.rb
laritakr: It is only needed on the find. I am changing this.

PR #1787, opened 2017-10-09
Provide some Rails options that make the build faster

jcoyne:


.travis.yml
```diff
@@ -18,6 +18,7 @@ rvm:
 env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
+    - ENGINE_CART_RAILS_OPTIONS='--skip-git --skip-bundle --skip-listen --skip-spring --skip-yarn --skip-keeps --skip-action-cable --skip-coffee --skip-puma --skip-test'
```
mjgiarlo: What would you think about re-enabing ActionCable? This would allow folks to see notifications working in the test application, which may not be valuable in a CI context but we also know that folks tend to use the test application as a way to kick the tires on Hyrax.

jcoyne: This is only for travis.

mjgiarlo: Derp.

PR #1789, opened 2017-10-09
Validate local file ingestions against a whitelist of directories.

ojlyytinen:
Make CreateWithRemoteFilesActor to validate the list of file urls given. To pass the validation, each file path must start with a whitelisted directory path from which ingestions are allowed to happen. Without this check, users could ingest any files from the file system (as long as the server process can read them).

The default whitelist will be either empty or contain the Browse Everything file system mount point if it is defined. As far as I can tell, nothing else should need to ingest files directly from the local file system. However, people may have implemented custom ingestion mechanisms in which case they would need to add their ingestion directories in Hyrax.config.ingest_dirs.

The validation method currently allows all urls that use something else than the file: scheme. It would probably be a good idea to implement some kind of validation here as well. It is quite likely that on production systems the server can access network resources that the user shouldn't be able to access. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/create_with_remote_files_actor.rb
+++ b/app/actors/hyrax/actors/create_with_remote_files_actor.rb
@@ -18,12 +18,34 @@ module Hyrax
 
       private
 
+        def whitelisted_ingest_dirs
+          Hyrax.config.whitelisted_ingest_dirs
+        end
+
+        def validate_remote_url(url)
+          uri = URI.parse(URI.encode(url))
+          if uri.scheme == 'file'
+            path = File.absolute_path(URI.decode(uri.path))
+            whitelisted_ingest_dirs.any? do |dir|
+              path.start_with?(dir) && path.length > dir.length
+            end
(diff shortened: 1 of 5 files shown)


app/actors/hyrax/actors/create_with_remote_files_actor.rb
```diff
@@ -18,12 +18,34 @@ def update(env)
 
       private
 
+        def whitelisted_ingest_dirs
```
revgum: I feel like it would be beneficial to have the method name match the configuration or vice-versa.

ojlyytinen:
Good suggestions, I've changed those and also added a couple of lines in spec/lib/hyrax/configuration_spec.rb.

mjgiarlo:
@ojlyytinen Can you rebase this? And then I will merge.

mjgiarlo:
Merged via command-line. Thanks, @ojlyytinen!

PR #1796, opened 2017-10-09
Include form error for works visibility field.

### LINKED ISSUE #1772: Works "disappear" with deposit to Admin sets with combo: Release Embargo and Visibility Everyone
Descriptive summary

When submitting to admin sets 2 and 11, it defaults to "private setting" - so if the depositor doesn't actively click another setting, but hits save, the work disappears. As far as I can tell, the combo has to be embargo allowed and public visibility.

Here is the admin set setting.
screen shot 2017-10-05 at 11 58 38 am

Here is the work view, where the setting does not move from private, like the others do.
screen shot 2017-10-05 at 11 47 19 am

Steps to reproduce the behavior

Create admin set with Embargo allowed, public visibility only
deposit work to admin set - do not change access control.
Click save, work disappears.
Related work

I believe this is like curationexperts/nurax#43 Sorry I didn't catch it before.

jonathandixon:
Fixes #1772 

Render any visibility field errors on the work form.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_forms.scss
+++ b/app/assets/stylesheets/hyrax/_forms.scss
@@ -49,11 +49,7 @@ form {
 }
 
 .set-access-controls {
-  label {
-    color: #333;
-    font-weight: normal;
-  }
-
+  label { font-weight: normal; }
   & .form-group {
     padding: 0 1.75em;
 
--- a/app/assets/stylesheets/hyrax/_forms.scss
+++ b/app/assets/stylesheets/hyrax/_forms.scss
@@ -49,7 +49,11 @@ form {
 }
 
 .set-access-controls {
-  label { font-weight: normal; }
+  label {
+    color: #333;
+    font-weight: normal;
+  }
+
   & .form-group {
     padding: 0 1.75em;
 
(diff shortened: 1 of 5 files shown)


mjgiarlo:
Thoughts on this fix, @vantuyls @julesies ?

jonathandixon:
@mjgiarlo Do you think I should modify this to render the errors like this, https://github.com/samvera/hyrax/blob/master/app/views/hyrax/base/_form.html.erb#L12? Doing this would also make it easier to handle errors produced in #1801.

mjgiarlo:
@jonathandixon That makes sense, yes. I like that even better. Thanks for asking.

julesies:
I just want to check in on this, so we have consistent behavior. As far as I know, you should not be allowed to select an setting prohibited by the Admin Set "Release and Visibility" Policy. Is this consistent with that expectation?

jonathandixon:
@julesies This PR only ensures that if a validation error occurred related to the "Release and Visibility" policy of the AS, then the error message will be displayed to the user when the form reloads. So no, the behavior is still not consistent with your expectation. At least with this, the user will know why the work didn't save.

julesies:
@jonathandixon @mjgiarlo @vantuyls  so it sounds like we need the ticket "Allow deposits to admin sets with expired embargoes"?  is this 2.0 or 2.x?

jonathandixon:
@julesies I'm working on that now, see #1801

julesies:
@jonathandixon this is awesome...thank you so much! You have removed a ton of blockers for 2.0 - it's really great!! We are only down to a few...if you are interested ;)

PR #1798, opened 2017-10-09
Use the persister rather than ActiveFedora::Base#save

### LINKED ISSUE #1790: Stop using ActiveFedora::Base#save

jcoyne:
Fixes #1790


app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -21,7 +21,12 @@ def create_content(file, relation = :original_file, from_url: false)
         # If the file set doesn't have a title or label assigned, set a default.
         file_set.label ||= label_for(file)
         file_set.title = [file_set.label] if file_set.title.blank?
-        return false unless file_set.save # Need to save to get an id
+
+        persister = Valkyrie.config.metadata_adapter.persister
```
tpendragon: Are you planning to set this to IndexingPersister? If not, then this won't index into solr.

PR #1809, opened 2017-10-09
Use FactoryGirl everywhere

jcoyne:
This uses Valkyrie rather than `ActiveFedora::Base.create`


tpendragon:
@jcoyne Not `create_for_repository`?

Issue #1765, opened 2017-10-05
Rights in catalog_controller template - intentional or oversight?


geekscruff:
Hello. I notice that in `rights` appears several times in the `catalog_controller.rb` template at `lib/generators/hyrax/templates/`.

Should that be changed to `license` and `rights_statement` in line with changes to `BasicMetadata` and `WorkShowPresenter`?

I'm happy to pick this up, but wanted to check first that there isn't a reason for it being this way.

mjgiarlo:
I suspect you're right. Go for it. Thanks!

PR #1817, opened 2017-10-10
Cleanup legacy rights (#1765) and add a rights statement helper

geekscruff:
This PR is to cleanup some out-of-date code relating to the `rights` property, which has been replaced by `rights_statement` and `license` in `basic_metadata`.

* removed rights from `lib/generators/hyrax/templates/catalog_controller.rb` and added license and rights_statement
* removed rights from `lib/generators/hyrax/templates/config/locales/`  and added rights statement and license (using google translate for the various languages)

In addition, 

* renamed `app/services/rights_statements.rb` to `rights_statement_service.rb` to be consistent with `license_service`
* added a `rights_statement_links` method into hyrax_helper_behavior.rb to provide linked text for the rights statement, in the same way as the existing `license_links` method
* added this method as a render in the `catalog_controller` file referenced above

### DIFF
--- a/app/helpers/hyrax/hyrax_helper_behavior.rb
+++ b/app/helpers/hyrax/hyrax_helper_behavior.rb
@@ -1,5 +1,3 @@
-# coding: utf-8
-
 module Hyrax
   module HyraxHelperBehavior
     include Hyrax::CitationsBehavior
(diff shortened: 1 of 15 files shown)


jcoyne:
@geekscruff While you are working on this would you please consider providing a more descriptive commit message so that one doesn't have to lookup #1765 to see what the commit is about?  Thanks! 😃

PR #1824, opened 2017-10-10
Validate local file ingestions against a whitelist of directories

mjgiarlo:
Backports #1789 from `master` to `1-0-stable`. (Also backports #1696 to fix the build.)

From @ojlyytinen:
> Make CreateWithRemoteFilesActor to validate the list of file urls given. To pass the validation, each file path must start with a whitelisted directory path from which ingestions are allowed to happen. Without this check, users could ingest any files from the file system (as long as the server process can read them).
> 
> The default whitelist will be either empty or contain the Browse Everything file system mount point if it is defined. As far as I can tell, nothing else should need to ingest files directly from the local file system. However, people may have implemented custom ingestion mechanisms in which case they would need to add their ingestion directories in Hyrax.config.ingest_dirs.
> 
> The validation method currently allows all urls that use something else than the file: scheme. It would probably be a good idea to implement some kind of validation here as well. It is quite likely that on production systems the server can access network resources that the user shouldn't be able to access.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/create_with_remote_files_actor.rb
+++ b/app/actors/hyrax/create_with_remote_files_actor.rb
@@ -19,6 +19,10 @@ module Hyrax
         return true unless remote_files
         remote_files.each do |file_info|
           next if file_info.blank? || file_info[:url].blank?
+          unless validate_remote_url(file_info[:url])
+            Rails.logger.error "User #{user.user_key} attempted to ingest file from url #{file_info[:url]}, which doesn't pass validation"
+            return false
+          end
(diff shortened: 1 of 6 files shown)


mjgiarlo:
@ojlyytinen @revgum Care to review this backport of #1789 to `1-0-stable`?

cc: @vantuyls

ojlyytinen:
I realise you already merged this but just to confirm that this looks good to me as well.

mjgiarlo:
Thanks, @ojlyytinen!

PR #1830, opened 2017-10-10
Use change_sets instead of forms 

carrickr:
- Add change_set_persisters
- Update Hyrax::Actors::Environment to use Hyrax::ChangeSetPersister
- Update base_actor to use change_sets instead of forms

Fixes #1729


carrickr:
Todo/To review:

- Determine if DynamicChangeSet was brought over properly from Figgy
- Have actor stack only use change sets, or a combination of changesets and curation_concerns?
- Make sure ::BaseActor#create works correctly

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -16,9 +16,16 @@ class BaseActor < AbstractActor
       # @param [Hyrax::Actors::Environment] env
       # @return [Boolean] true if create was successful
       def create(env)
-        apply_creation_data_to_curation_concern(env)
-        apply_save_data_to_curation_concern(env)
-        save(env) && next_actor.create(env) && run_callbacks(:after_create_concern, env)
+        begin
+          change_set = Hyrax::DynamicChangeSet.new(env.curation_concern)
```
jcoyne: This seems like unnecessary indirection.  Can we make it `ChangeSetFactory.create(env.curation_concern)` instead?

jcoyne: Ugh, I see this is how figgy is doing it, so I'm not going to complain too hard.

jcoyne:
@carrickr should we merge this now, or do you want to resolve your todo list first?

app/actors/hyrax/actors/base_actor.rb
cjcolvar: I'm confused by this since it appears to be returning `Hyrax::DynamicChangeSet#initialize` with `env.curation_concern` being set as the `resource_param` when I'm guessing you're wanting the result of `Hyrax::DynamicChangeSet#new` and to get that right now you'd need to do `change_set = Hyrax::DynamicChangeSet.new.new(env.curation_concern)`.

cjcolvar: @carrickr I see the new changes :+1:

carrickr: We updated the DynamicChangeSet to use new as a class method instead of an instance method and avoid the `.new.new`

https://github.com/samvera/hyrax/blob/1729-changesets/app/change_sets/hyrax/dynamic_change_set.rb#L5

carrickr:
@jcoyne @tpendragon @mjgiarlo @cjcolvar 

Ready for review and merge

app/actors/hyrax/actors/base_actor.rb
tpendragon: @jcoyne You're welcome to use `ChangeSetFactory`. It's a `DynamicChangeSet` in Figgy because I tend to treat classes as factories.

@cjcolvar `DynamicChangeSet` overrides the class method `new`

lib/generators/hyrax/work/work_generator.rb
```diff
@@ -34,8 +34,8 @@ def create_indexer
     template('indexer.rb.erb', File.join('app/indexers', class_path, "#{file_name}_indexer.rb"))
   end
 
-  def create_form
-    template('form.rb.erb', File.join('app/forms/hyrax', class_path, "#{file_name}_form.rb"))
```
jcoyne: The spec/forms/hyrax/generic_work_form_spec.rb still references `GenericWorkForm` which is no longer being created.

carrickr: Removed the `GenericWorkForm` spec and added in a basic spec for `GenericWorkChangeSet`

carrickr: #1851 added to the backlog to enhance the features for change sets and spec for `GenericWorkChangeSet`

Issue #1823, opened 2017-10-10
Replace #find with #find_by(id:)


cjcolvar:
Use Valkyrie's query_service method to find resources by id.

PR #1831, opened 2017-10-10
Lookup using Valkyrie instead of ActiveFedora::Base.find

cjcolvar:
Start of work for #1823.

### DIFF
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -19,13 +19,13 @@ module Hyrax
     # Add members using the members association.
     def add_members(new_member_ids)
       return if new_member_ids.blank?
-      members << ActiveFedora::Base.find(new_member_ids)
+      members << new_member_ids
     end
 
     # Add member objects by adding this collection to the objects' member_of_collection association.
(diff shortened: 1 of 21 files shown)


app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -19,13 +19,13 @@ module CollectionBehavior
     # Add members using the members association.
     def add_members(new_member_ids)
       return if new_member_ids.blank?
-      members << ActiveFedora::Base.find(new_member_ids)
+      members << find_resource(new_member_ids)
```
jcoyne: I'm not sure that valkyrie's `find_by(id:)` can take/return an array.

tpendragon: It can't, but there's also no reason to find the resource if you have the member IDs. Just add the member IDs.

jcoyne:
@cjcolvar looks like this needs to be rebased.

app/models/concerns/hyrax/collection_behavior.rb
jcoyne: so switch this to `member_ids << new_member_ids`

PR #1838, opened 2017-10-11
Use data mapper pattern to persist the AdminSet

jcoyne:


app/services/hyrax/admin_set_create_service.rb
```diff
@@ -55,19 +55,20 @@ def initialize(admin_set:, creating_user:, workflow_importer: default_workflow_i
     def create
       admin_set.edit_groups = [admin_group_name]
       admin_set.creator = [creating_user.user_key] if creating_user
-      admin_set.save.tap do |result|
-        if result
```
cjcolvar: Do you need to worry about save not returning a persisted object? Or are you letting the exceptions bubble up instead (assuming `PermissionTemplate.create!` will throw an exception when passed a resource with `id: nil`)?

jcoyne: We checked the result here before due to validations, but Valkyrie doesn't have validations on the model, so I don't think there's any think to be gleaned from checking the return value.

PR #1854, opened 2017-10-11
Default option for admin set release settings.

### LINKED ISSUE #1799: Admin set release option can't return to defaults (disables lease option for the admin set)
**Descriptive summary**

When creating an admin set, you can opt not to set anything under Release and Visibility: Release. However, if you click any of the release settings here, you can not Unclick (return to default). AFAICT, the only way to allow the option of Lease to work in the Work deposit form is to have nothing selected under "Release".

So if you create a new admin set and take the “default setting” on the admin set for release and vis…meaning you don’t select any radio button…then a user can submit a work with a lease.

However, if the repo admin ever clicks any of those radio buttons, even by accident to click around…they’ve unknowingly disabled leases for that admin set completely and they can't get the setting back.

**Rationale**

Either have a way to clear settings in the admin set, or add an option "any setting" or something.

1.Create a new admin set, do not click any setting under release
2.submit work with lease.
3.Change admin set by clicking any radio button to any setting under Release
4.Can't submit leased works, can't unclick Release Setting.

jonathandixon:
Fixes #1799 

Add options to the Admin set release form to capture the default release settings. The default options allow the user to select any visibility for works, including lease and embargo.

Admin Set Release Form now includes these options, with the indicated items selected by default:

- No Delay
- Varies (DEFAULT)
  - Allow depositor to decide (DEFAULT)
  - Between now and `{DATE}`
  - Between now and `{PERIOD}`
- Fixed

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/permission_template_form.rb
+++ b/app/forms/hyrax/forms/permission_template_form.rb
@@ -186,6 +186,8 @@ module Hyrax
         # Removes release_varies and release_embargo from the returned attributes
         # These form fields are only used to update release_period
         # @return [Hash] attributes used to update the model
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/PerceivedComplexity
         def permission_template_update_params(raw_attributes)
           attributes = raw_attributes.except(:release_varies, :release_embargo)
           # If 'varies' before date option selected, then set release_period='before' and save release_date as-is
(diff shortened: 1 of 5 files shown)


jonathandixon:
@mjgiarlo @vantuyls @julesies The proposed change differs slightly from what was suggested in the issue. Please review this and let me know if you think this is better or worse. I have no issues changing to the suggested approach, this was just easier from an implementation standpoint.

mjgiarlo:
@tdonohue This is a small PR related to embargo/lease. Do you have cycles to review this?

mjgiarlo:
@jonathandixon that approach works for me.

PR #1862, opened 2017-10-12
Swap change set for form

jcoyne:


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -131,8 +131,8 @@ def inspect_work
 
     private
 
-      def build_form
-        @form = work_form_service.build(curation_concern, current_ability, self)
+      def build_change_set
+        @change_set = work_form_service.build(curation_concern, current_ability, self)
```
carrickr: Should we refactor `work_form_service` if we're looking at generally removing the concept of forms?

jcoyne: Yes, but I don't think that needs to go into this PR, does it?

PR #1863, opened 2017-10-12
Persist test objects using Valkyrie

jcoyne:
Fixed consistency of the factory name.

### DIFF
--- a/spec/actors/hyrax/actors/generic_work_actor_spec.rb
+++ b/spec/actors/hyrax/actors/generic_work_actor_spec.rb
@@ -246,7 +246,7 @@ RSpec.describe Hyrax::Actors::GenericWorkActor do
     end
 
     context 'without in_works_ids' do
-      let(:old_parent) { create_for_repository(:generic_work) }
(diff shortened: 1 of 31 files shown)


spec/actors/hyrax/actors/generic_work_actor_spec.rb
```diff
@@ -270,7 +270,7 @@
     end
 
     context 'with nil in_works_ids' do
-      let(:parent) { FactoryGirl.create(:generic_work) }
+      let(:parent) { FactoryGirl.create_for_repository(:work) }
```
cjcolvar: For consistency in this spec file, should the `FactoryGirl.` be removed or is that a different concern/PR?

jcoyne: I can change it if you want.

cjcolvar: I think if I care enough I'll make a PR later to make this consistent across the test suite.

PR #1868, opened 2017-10-12
Find all file sets using Valkyrie

cjcolvar:


app/services/hyrax/queries.rb
```diff
@@ -5,15 +5,15 @@ class Queries
     class_attribute :metadata_adapter
     self.metadata_adapter = Valkyrie.config.metadata_adapter
     class << self
-      delegate :find_all, :find_by, :find_members, :find_inverse_references_by, to: :default_adapter
+      delegate :find_all, :find_all_of_model, :find_by, :find_members, :find_inverse_references_by, to: :default_adapter
```
tpendragon: If `Hyrax::Queries` is going to be a standard practice, we should have its specs run the shared specs.

cjcolvar: I can work on that.  Do you think it should go in this PR?

spec/services/hyrax/repository_fixity_check_service_spec.rb
```diff
@@ -1,9 +1,5 @@
 RSpec.describe Hyrax::RepositoryFixityCheckService do
-  let!(:file) do
-    create(:file_set).tap do |file|
-      file.add_file(File.open(fixture_path + '/world.png'), path: 'content', original_name: 'world.png')
-    end
-  end
+  let!(:file) { create_for_repository(:file_set) }
```
jcoyne: Do we not need the attached file? Why?

cjcolvar: No, it isn't being used actually.  The test is just checking that `FileSetFixityCheckService#fixity_check` is being called:
https://github.com/samvera/hyrax/blob/2d8f81e1be8181644b6a57165a442aec4107b44d/spec/services/hyrax/repository_fixity_check_service_spec.rb#L8-L9

PR #1870, opened 2017-10-12
Refactor presenter factory to use Valkyrie queries

jcoyne:


app/presenters/hyrax/member_presenter_factory.rb
```diff
@@ -68,5 +54,10 @@ def presenter_factory_arguments
       def composite_presenter_class
         CompositePresenterFactory.new(file_presenter_class, work_presenter_class, ordered_ids & file_set_ids)
       end
+
+      def file_sets
+        @file_set_ids ||= Hyrax::Queries.metadata_adapter_query_service.find_members(resource: @work,
```
cjcolvar: Should the instance variable be `@file_sets` since it is the objects and not the ids?

cjcolvar: Why `Hyrax::Queries.metadata_adapter_query_service.find_members` instead of just `Hyrax::Queries.find_members`?

jcoyne:
@cjcolvar Thanks for the review. I've made the changes you suggested.

PR #1876, opened 2017-10-12
Change has_model to internal_model 

jcoyne:
and use Valkyrie::Persistence::Solr::Queries::MODEL for consistency

### DIFF
--- a/app/search_builders/hyrax/abstract_type_relation.rb
+++ b/app/search_builders/hyrax/abstract_type_relation.rb
@@ -14,7 +14,7 @@ module Hyrax
 
     def search_model_clause
       clauses = allowable_types.map do |k|
-        ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: k.to_s)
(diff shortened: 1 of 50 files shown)


app/search_builders/hyrax/abstract_type_relation.rb
```diff
@@ -14,7 +14,7 @@ def equivalent_class?(klass)
 
     def search_model_clause
       clauses = allowable_types.map do |k|
-        ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: k.to_s)
+        ActiveFedora::SolrQueryBuilder.construct_query_for_rel(Valkyrie::Persistence::Solr::Queries::MODEL => k.to_s)
```
cjcolvar: I don't think this is right since `construct_query_for_rel` is going to run `solr_name` and I think end up with "internal_resource_ssim_ssim".

jcoyne: Oops, looks like you're right. I thought it behaved differently for string keys.

PR #1878, opened 2017-10-12
Prune tests for features that are no longer supported.

jcoyne:
Fix FileSet#parents
Remove unused FileSet::Querying

Added a draft of release notes


app/models/concerns/hyrax/file_set/belongs_to_works.rb
```diff
@@ -7,8 +7,10 @@ module BelongsToWorks
       #   before_destroy :remove_representative_relationship
       # end
 
+      # Objects that call this FileSet a member
       def parents
-        in_works
+        query_service = Valkyrie::MetadataAdapter.find(:indexing_persister).query_service
+        query_service.find_inverse_references_by(resource: self, property: :member_ids)
```
cjcolvar: Could this be simplified to `query_service.parents(resource: self)`?  If so, do we even need this method/module at all?

spec/models/file_set_spec.rb
```diff
@@ -205,93 +137,9 @@ class AltFile < ActiveFedora::Base
...
-        m.apply_depositor_metadata('jcoyne')
-        m.permissions_attributes = [{ type: 'person', access: 'read', name: 'person1' },
-                                    { type: 'person', access: 'read', name: 'person2' },
-                                    { type: 'group', access: 'read', name: 'group-6' },
-                                    { type: 'group', access: 'read', name: 'group-7' },
-                                    { type: 'group', access: 'edit', name: 'group-8' }]
-      end
-    end
-
-    it 'has read groups accessor' do
```
cjcolvar: Did you mean to remove this test?

jcoyne: Yeah, this is a method on Valkyrie::Resource, not on this class.

app/models/concerns/hyrax/file_set/belongs_to_works.rb
jcoyne: I've made the suggested change and deprecated this method.

spec/models/file_set_spec.rb
```diff
@@ -462,6 +192,7 @@ def paranoid_edit_permissions
     end
   end
 
+<<<<<<< HEAD
```
cjcolvar: Looks like this get left here.

spec/models/file_set_spec.rb
```diff
@@ -509,6 +240,8 @@ def paranoid_edit_permissions
     end
   end
 
+=======
+>>>>>>> Prune tests for features that are no longer supported.
```
cjcolvar: Looks like this get left here.

PR #1889, opened 2017-10-13
Allow Fedora Schema to be configured.

tpendragon:


jcoyne:
Please fix the rubocop issue in `lib/generators/hyrax/templates/config/initializers/valkyrie.rb`

tpendragon:
@jcoyne Fixed

PR #1893, opened 2017-10-13
Use storage.upload() rather than Hydra::Works::UploadFileToFileSet

jcoyne:


spec/controllers/hyrax/downloads_controller_spec.rb
```diff
@@ -2,20 +2,26 @@
   routes { Hyrax::Engine.routes }
 
   describe '#show' do
-    let(:user) { FactoryGirl.create(:user) }
+    let(:user) { create(:user) }
+    let(:file) { fixture_file_upload('image.png', 'image/png') }
+    let(:storage_adapter) { Valkyrie::StorageAdapter.find(:disk) }
     let(:file_set) do
-      FactoryGirl.create(:file_with_work, user: user, content: File.open(fixture_path + '/image.png'))
```
cjcolvar: I see you got rid of the content param here but kept it below in the test for FileSetFixityCheckService as well as fixed the factory.  Is it preferred to pass content to the factory to avoid the before block you have here?  If so, could you use that pattern here?

PR #1896, opened 2017-10-14
Stop using AF::Base#where and #search_with_conditions

jcoyne:
Remove some collection tests for features we no longer support.

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -6,14 +6,6 @@ RSpec.describe Collection do
     expect(collection.read_groups).to eq ['public']
   end
 
-  describe "#validates_with" do
-    before { collection.title = nil }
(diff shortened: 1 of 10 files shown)


spec/models/collection_spec.rb
```diff
@@ -37,26 +29,15 @@
...
+      before do
         collection.add_member_objects [work1.id, work2.id]
-        collection.save!
-        expect(collection.reload.member_objects).to match_array [work1, work2]
+        persister.save(resource: collection)
       end
-    end
-  end
 
-  describe "#destroy", clean_repo: true do
```
cjcolvar: Can you explain why testing ` #destroy` is being removed?

jcoyne: Because in Valkyrie you don't destroy on a model.  You destroy on a change set.

PR #1897, opened 2017-10-14
Add AdminSetChangeSetPersister

jcoyne:


app/change_set_persisters/hyrax/admin_set_change_set_persister.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literals: true
+
+module Hyrax
+  class AdminSetChangeSetPersister < ChangeSetPersister
```
tpendragon: I don't usually do a change set for specific models, but maybe it's useful in this case.

PR #1898, opened 2017-10-16
Use layer 0 for PDF and document derivatives

mjgiarlo:
This is an innovation in place in more than one Sufia/Hyrax application downstream that makes derivatives work as expected.

I'd like to tag @kerchner, @awead, and @jenlindner to review this. Will this have unintended side effects?

Also, looking at [some other work](https://github.com/aic-collections/aicdams-lakeshore/commit/811c3b9e0d687c1e667d0bfdc3e765ac6248d7cf#diff-21fb19adf50f17aa1e5567a28fd37400) linked from the `garbled PDF derivatives` [thread](https://groups.google.com/d/topic/samvera-tech/DRTzwBG9WKg/discussion) on samvera-tech, should we look into switching from ImageMagick to GraphicsMagick by default either in (Hydra-Derivatives or Hyrax)?

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/file_set_derivatives_service.rb
+++ b/app/services/hyrax/file_set_derivatives_service.rb
@@ -48,15 +48,24 @@ module Hyrax
 
       def create_pdf_derivatives(filename)
         Hydra::Derivatives::PdfDerivatives.create(filename,
-                                                  outputs: [{ label: :thumbnail, format: 'jpg', size: '338x493', url: derivative_url('thumbnail') }])
+                                                  outputs: [{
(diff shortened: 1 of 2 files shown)


awead:
@mjgiarlo it shouldn't, although the `layer` directive defaults to 0 for pdfs https://github.com/samvera/hydra-derivatives/blob/master/lib/hydra/derivatives/processors/image.rb#L63

GraphicsMagick seemed better because it didn't fail on trivial tiff errors. Although, there's a setting in ImageMagick to make it less whiny.

PR #1900, opened 2017-10-17
Add missing translation

### LINKED ISSUE #1882: Collections: BUG - missing permission_update_notices.participants translation
### Descriptive summary

Error flashes when adding a participant...  `translation missing: en.hyrax.dashboard.collections.form.permission_update_notices.participants`

### Expected behavior

When a participant is added, the flash message should be... `The collection's sharing options have been updated.`

### Actual behavior

When a participant is added, the flash message get a translation missing error... `translation missing: en.hyrax.dashboard.collections.form.permission_update_notices.participants`

1. Dashboard -> Collections -> New Collection -> set title and Save
1. click Sharing tab -> set user with any role and click Add

- [ ] Verify that a flash message is displayed with text `The collection's sharing options have been updated.`

bkeese:
Fixes #1882 

Adds missing translation.

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.en.yml
+++ b/config/locales/hyrax.en.yml
@@ -496,6 +496,7 @@ en:
         form:
           permission_update_notices:
             sharing: "The collection's sharing options have been updated."
+            participants: "The collection's sharing options have been updated."
           permission_update_errors:
             error:   "Invalid update option for permission template."
           tabs:


config/locales/hyrax.en.yml
```diff
@@ -496,6 +496,7 @@ en:
         form:
           permission_update_notices:
             sharing: "The collection's sharing options have been updated."
+            participants: "The collection's sharing options have been updated."
```
jcoyne: Shouldn't it be something about participants?

elrayle: @jcoyne Collections implement sharing using participants in the same way as admin sets.  But the UI/UX design wants general user facing labels to match that of works.  So collections and works both use the terminology 'sharing' in tabs and messages to the user.

The original issue with the mockup using Sharing as a label is ##1378.  @ggeisler confirmed in an earlier discussion the desire for user to have a consistent experience.

PR #1919, opened 2017-10-23
Remove the `AssignRepresentativeActor`, which is a no-op/save

no-reply:
This actor assigned a variable to `nil`, then acted only `if variable`. It then called `save` on the object, making it expensive and leading to potential surprising impact on other actors. This removes it from the actor stack and deletes the actor file.

* Remove the `AssignRepresentativeActor` from the stack.
* Delete `AssignRepresentativeActor`.

Note that the second change is potentially breaking (`NameError`) for anyone who refers to this actor in a custom stack.

@samvera/hyrax-code-reviewers


no-reply:
Tagging @jcoyne since the `nil` assignment dates to a refactor of his. I'd like his formal :+1:.

mjgiarlo:
@no-reply When this is merged, can you add this change to the 2.0.0 release notes so folks know about the breaking change? Thanks!

no-reply:
> @no-reply When this is merged, can you add this change to the 2.0.0 release notes so folks know about the breaking change? Thanks!

Can do.

jcoyne:
I'm of two minds about this.  Clearly we've lived without it for 2 years, so NBD. On the other hand it was supposed to work like this: https://github.com/samvera/curation_concerns/commit/747888b8e05b3ec91d7fb63d5635ffbef5facb4e#diff-6adfb371362714d03d3c9cf353aae0b5L75 and it wouldn't be hard to restore that behavior.  I want someone else to decide.

mjgiarlo:
Noting for posterity: @no-reply will follow up with.

1. A blurb in the release notes about this change, and
2. A new issue to
> [c]onsider use cases along the lines of the content_type aware assignment discussed last week and think about reintroducing an actor in that context

PR #1922, opened 2017-10-23
Fix error message for unimplemented change_sets

### LINKED ISSUE #1888: Remove Apply Order Actor from the stack
No longer needed due to Valkyrie tracking order, remove from stack and middleware layer calls

carrickr:
Fixes #1888 

Error message for unimplemented change_sets is now passed up to env rather than silently swallowed
BaseActor now has a working attribute setter when attributes are passed into the initializer
Refactor change_set generator to include work class's fields by default
ApplyOrderActor is no longer needed due to Valkyrie retaining order, remove actor and spec for actor
Remove ApplyOrderActor references from the middleware stack and testing stack


jcoyne:
I feel like this could use a better commit message.

carrickr:
@jcoyne commit message updated

app/actors/hyrax/actors/apply_order_actor.rb
```diff
@@ -16,44 +17,28 @@ def can_edit_both_works?(env, work)
...
-        def sync_members(env, ordered_member_ids)
-          return true if ordered_member_ids.nil?
-          cleanup_ids_to_remove_from_curation_concern(env.curation_concern, ordered_member_ids)
-          add_new_work_ids_not_already_in_curation_concern(env, ordered_member_ids)
-          env.curation_concern.errors[:ordered_member_ids].empty?
+        def sync_members(env, member_ids)
+          return true if member_ids.nil?
+          env.change_set.validate(member_ids: member_ids.select { |work_id| can_add_work_id?(env, work_id) })
+          env.change_set.sync
+          persister.save(resource: env.curation_concern)
```
jcoyne: I'm not sure that it's clear that `sync_members` is going to cause a save operation.  Is it possible to do all the saving in the BaseActor?  That's what we use to do. Do we still need both a BaseActor and ApplyOrderActor?

mbklein: I don't even know if ApplyOrderActor is required any more, since order is handled farther down by Valkyrie. Whatever order you set the members in, that's the order they persist in.

jcoyne: I think we should yank it out then.

jcoyne:
It looks like this build fails on the rubocop step. Can you fix that?

carrickr:
@jcoyne 

Rubocop fixed.
Upon further review it was decided ApplyOrderActor has no purpose in life, not even to pass the butter, ince Valkyrie will return items to us in order so the actor been removed entirely.

jcoyne:
I feel like we need to move the `can_edit_both_works?` into a change_set now? At least put it in the `Valkyrie_README.md` as a TODO.

carrickr:
@jcoyne #1927 created for the `can_edit_both_works?` concern.  I'd like to get this PR in though because we need the general change_set improvements merged in the valkyrie branch so other devs can work on fixing other actors and not have to base off this branch to get that code

PR #1934, opened 2017-10-24
Support all abilities defined in blacklight-access-controls

cjcolvar:
Add read, discover, and download abilities for Valkyrie::Resource objects.


jcoyne:
this might be a bit misleading as these are defined in hydra-access-controls: https://github.com/samvera/hydra-head/blob/master/hydra-access-controls/lib/hydra/ability.rb#L59-L61

PR #1937, opened 2017-10-24
Use the correct factory for works

jcoyne:


spec/controllers/hyrax/api/items_controller_spec.rb
```diff
@@ -204,7 +204,7 @@
     end
 
     context 'with a valid item, matching token, authorized resource, but not Arkivo-deposited' do
-      let(:non_arkivo_gw) { create :generic_work, id: 'abc123xyz', arkivo_checksum: nil }
+      let(:non_arkivo_gw) { create :work, id: 'abc123xyz', arkivo_checksum: nil }
```
cjcolvar: Should this change to `create_for_repository`?

PR #1942, opened 2017-10-24
Refactor count queries away from ActiveFedora::Relation

jcoyne:


cjcolvar:
@jcoyne do you think these count queries will be moved to custom queries on Hyrax::Queries or something different?  I'm guessing we don't want to keep depending on ActiveFedora.

jcoyne:
@cjcolvar probably not because these are only ever backed by Solr. Although, I suspect we'll just swap the solr connection out with the one Blacklight provides.

PR #1946, opened 2017-10-24
Add a CollectionChangeSetPersister

jcoyne:


app/change_set_persisters/hyrax/collection_change_set_persister.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+      end
+
+      def move_members_between_collections(change_set:, resource:)
+        destination_collection = find_resource(change_set.destination_collection_id)
+        remove_members_from_collection(change_set: change_set, resource: resource)
+        add_members_to_collection(change_set: change_set, resource: destination_collection)
+      end
+
+      # This can be removed when add_member_objects is removed
```
cjcolvar: This comment doesn't appear to be correct anymore so it should probably be remove.  Same for the comment below.

app/change_set_persisters/hyrax/collection_change_set_persister.rb
```diff
@@ -0,0 +1,52 @@
...
+          member = find_resource(member_id)
+          member.member_of_collection_ids.delete resource.id
+          persister.save(resource: member)
+        end
+      end
+
+      def add_members_to_collection(change_set:, resource:)
+        Array(change_set.batch).each do |member_id|
+          member = find_resource(member_id)
+          member.member_of_collection_ids << resource.id
```
cjcolvar: Is `<<` okay here or should it be `+=`?

jcoyne: I'm not seeing any frozen array modification error, so I guess it's okay.

tpendragon: I think ChangeSets handle defaults differently.

cjcolvar:
@carrickr note that this conflicts with some of #1952 in ChangeSetPersister.

PR #1949, opened 2017-10-24
Remove more references to WorkRelation

cjcolvar:


app/services/hyrax/user_stat_importer.rb
```diff
@@ -60,13 +60,25 @@ def process_files(stats, user, start_date)
...
         work_ids_for_user(user).each do |work_id|
-          work = Hyrax::WorkRelation.new.find(work_id)
+          work = find_work(work_id)
           work_stats = extract_stats_for(object: work, from: WorkViewStat, start_date: start_date, user: user)
           stats = tally_results(work_stats, :work_views, stats) if work_stats.present?
           delay
         end
       end
 
+      def find_work(id)
```
jcoyne: Do you think we should move this to `Hyrax::Queries`?

cjcolvar: We definitely need a way to dry this out now that it is appearing many times.  Do you think we should have move `find_work`, `find_file_set`, and `find_collection` as is or make a generic `find(id:, model_type:)`?  I wonder if it would help to have a field on the models like `hydra_works_type` since we need to query as well as find and verify.

jcoyne: Either way is fine with me.  I believe there is a field called `internal_resource` we could use.  This sounds like a special case of the `FindAllQuery` https://github.com/samvera-labs/valkyrie/blob/1ca6b75a77b036d0cd8b21e127f3d27ef3d33a41/lib/valkyrie/persistence/active_fedora/queries/find_all_query.rb#L21

cjcolvar: I moved them into `Hyrax::Queries`.  Could you re-review please?

Issue #1951, opened 2017-10-24
Create Query to find all members in the tree of a resource


carrickr:
This can be used as a starting point

https://github.com/samvera/hyrax/blob/306ffbd91c1bd9c063e21cad7f1d64821d8774bc/app/change_set_persisters/hyrax/change_set_persister.rb#L57-L66

Create a query that when given a resource returns an array of Valkyrie::IDs that are members of that resource (or members of members, members of those members, etc).  

Also consider this for root Valkyrie ( @tpendragon thoughts?)

PR #1952, opened 2017-10-24
Remove CleanupFileSetsActor, replicate removed functionality in the ChangeSetPersister

### LINKED ISSUE #1932: Remove File Set Actor and Replicate Functionality
- [x] Remove the Actor
- [x] Implement a function for a resource that deletes all members and then itself

carrickr:
Remove CleanupFileSetsActor
Add Hyrax::FileSetChangeSetPersister
Add Hyrax::WorkChangeSetPersister
Implement clean_up_file_sets in Hyrax::WorkChangeSetPersister
Implemeent delete_all in Hyrax::ChangeSetPersister
save and delete in Hyrax::ChangeSetPersister no longer catches :abort or other throws
delete in Hyrax::AdminSetChangeSetPersister now catches :abort
Add before_delete hook in Hyrax::WorkChangeSetPersister to delete all file_sets attached to the resource about to be deleted
Add test coverage for Hyrax::ChangeSetPersister
Remove usage of Hyrax::DynamicChangeSet in tests
Update Hyrax::ChangeSetPersister to prepare for the coming removal of Hyrax::DynamicChangeSet

Fixes #1932


spec/change_set_persisters/hyrax/change_set_persister_spec.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literals: true
+
+RSpec.describe Hyrax::ChangeSetPersister, type: :model do
+  let(:adapter) { Valkyrie::MetadataAdapter.find(:indexing_persister) }
+  let(:persister) { adapter.persister }
+  let(:query_service) { adapter.query_service }
+  let(:storage_adapter) { Valkyrie.config.storage_adapter }
+  let(:change_set_persister) { Hyrax::ChangeSetPersister.new(metadata_adapter: adapter, storage_adapter: storage_adapter) }
```
jcoyne: Would you mind using `described_class` here?

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -32,16 +34,35 @@ def save_all(change_sets:)
       end
     end
 
-    def buffer_into_index
-      metadata_adapter.persister.buffer_into_index do |buffered_adapter|
-        with(metadata_adapter: buffered_adapter) do |buffered_changeset_persister|
-          yield(buffered_changeset_persister)
-        end
+    def delete_all(change_sets:)
```
jcoyne: Is this a public API? Can it be documented?

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -32,16 +34,35 @@ def save_all(change_sets:)
...
+      change_sets.map do |change_set|
+        delete(change_set: change_set)
       end
     end
 
-    def with(metadata_adapter:)
-      yield self.class.new(metadata_adapter: metadata_adapter, storage_adapter: storage_adapter)
+    def delete_self_and_members(change_set:)
```
jcoyne: Is this a public api?  can we document it?

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -32,16 +34,35 @@ def save_all(change_sets:)
...
+      ids_to_delete = get_ids_of_self_and_all_members(resource: change_set.resource)
```
jcoyne: Why not just `get_ids_of_members` as we throw away the self anyway?

carrickr: @jcoyne By default I'd prefer the function return self and members, since I think for other use cases where I just want an array of all Valkyrie::IDs in play.  In this case we're only shifting the self ID off due to the actor we're replacing having a specific code comment about deleting self last to avoid an issue with solr indexing.  In other cases where I'm using the array of ids to figure out what to load into a buffer or whatever, I'd want self in that array.

spec/change_set_persisters/hyrax/change_set_persister_spec.rb
carrickr: Done

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: Doc Added

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: Doc Added

jcoyne:
Looks like  rebase is required.

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,45 +3,107 @@
...
+    # Saves all changes in the ChangeSet to the resource contained in the ChangeSet
+    # using the metadata_adapter and storage_adapter
+    # @param change_set Any change_set created by Hyrax::DynamicChangeSet
+    # @return the saved resource, such as GenericWork, that is subclassed from Valkyrie::Resource
     def save(change_set:)
-      persister.save(resource: change_set.resource)
+      run_hook(:before_save, change_set: change_set)
+      resource = persister.save(resource: change_set.resource)
+      run_hook(:after_save, change_set: change_set)
+      query_service.find_by(id: resource.id)
```
jcoyne: I don't think we should reload here. That could get rather slow. Is there a specific case where you think this would be required?

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,45 +3,107 @@
...
+      persister.delete(resource: change_set.resource)
+      run_hook(:after_delete, change_set: change_set)
     end
 
+    # Calls save on all change_sets in the array passed in
+    # @param [Array] change_sets An array of any change_set created by Hyrax::DynamicChangeSet
+    # @return [Array] an array of the saved resources, such as GenericWork,
+    #  that is subclassed from Valkyrie::Resource
     def save_all(change_sets:)
+      objects = []
```
jcoyne: Do we need this accumulator? Can't we just return the result of the `map`?

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,45 +3,107 @@
...
+        change_sets << Hyrax::DynamicChangeSet.new(resource)
+      end
+      delete_all(change_sets: change_sets)
+      delete(change_set: change_set)
     end
+
+    private
+
+      def get_ids_of_self_and_all_members(resource:, accumulator: [])
+        accumulator << resource.id
```
jcoyne: We're pushing this on just to shift it off on line 86?  That seems unnecessary.

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,45 +3,107 @@
...
+        resource.member_ids.each do |member_id|
+          next if accumulator.include?(member_id)
+          accumulator << member_id
+          get_ids_of_self_and_all_members(resource: metadata_adapter.query_service.find_by(id: member_id), accumulator: accumulator)
+        end
+        accumulator.uniq!
```
jcoyne: we could just return `accumulator.uniq` here

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: @jcoyne No specific case if one of the after_hooks alters the `resource` though we should reload.  This is reloading out of the metadata_adapter (so solr normally, but maybe postgres for some configs) so while not ideal, not killer.

I thought the consensus of the call was to keep the reload here

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,45 +3,107 @@
...
+    # Given a resource in a change_set, recursively deletes all members for the entire
+    # tree of members.  Will delete all members first and then the resource itself
+    # to avoid issues with removing the root of the tree first.  However all other
+    # resources are deleted as found by the query, so issues may arise from the order
+    # the resources were found in, so depending on the complexity of your member tree
+    # you may want to execerise caution in calling this method.
+    # @param [Array] change_sets An array of any change_set created by Hyrax::DynamicChangeSet
+    # @return [Array] an array of the resources that failed to delete, if empty
+    #  all resources were succesfully deleted
+    def delete_self_and_members(change_set:)
```
jcoyne: Previously we were just deleting FileSets, but this is now deleting all members. Do we need to insure the members we are deleting are not members of other objects (not-deleted)?  This is one of the beneficial constraints on FileSets, they can only be a member of one object.

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: Removed

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: This is being with an eye toward moving this over to a query (#1951) where it would be assumed you want the self as well.  We're only shifting this specifically to delay deletion of the original resource to the end of the delete_all.  For non order specific uses (like buffering a resource and all its members somewhere) we'd want the original resource in the array so the function was written from that perspective.

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: updated

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: My personal feeling is if you chose to call `delete_self_and_members` it's going to do that.  If you want to apply additional restrictions (type, membership to something else, etc) you can run `get_ids_of_self_and_all_members` (or some other query) and then prune out the `Valkyrie::ID`'s you don't want to delete before sending the array to `delete_all`.  

Before we didn't have delete recursion, so it was just `delete_all`, so for FileSets it would be a query for all members of the resource that are FileSets and pass the result to `delete_all`.  I'm definitely fine with putting in an issue for a recursive delete with some filtering support though.

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,53 +3,104 @@
...
+      change_sets.map do |change_set|
+        delete(change_set: change_set)
       end
     end
 
-    def with(metadata_adapter:)
-      yield self.class.new(metadata_adapter: metadata_adapter, storage_adapter: storage_adapter)
+    # Given a resource in a change_set, recursively deletes all members for the entire
+    # tree of members.  Will delete all members first and then the resource itself
+    # to avoid issues with removing the root of the tree first.  However all other
```
jcoyne: It's not a tree. It's a graph.  This method is extremely dangerous in such a situation.

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -48,8 +92,24 @@ def buffer_into_index
       end
     end
 
-    def with(metadata_adapter:)
-      yield self.class.new(metadata_adapter: metadata_adapter, storage_adapter: storage_adapter)
-    end
+    private
+
+      def get_ids_of_self_and_all_members(resource:, accumulator: [])
```
jcoyne: This doesn't appear to be used. Please remove.

app/change_sets/hyrax/dynamic_change_set.rb
```diff
@@ -3,12 +3,12 @@
...
   class DynamicChangeSet
     def self.new(obj, *args)
-      change_set_class = if obj.is_a? Collection
-                           "Hyrax::#{obj.class}ChangeSet"
-                         else
-                           "#{obj.class}ChangeSet"
+      change_set_class = begin
+                           "::#{obj.class}ChangeSet".constantize
+                         rescue NameError
+                           "Hyrax::#{obj.class}ChangeSet".constantize
```
jcoyne: Does this case ever get used anymore?

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,43 +3,87 @@
...
+    def delete_all(change_sets:)
+      change_sets.map do |change_set|
+        delete(change_set: change_set)
+      end
+    end
+
+    # Deletes all file_sets that are members of the resource in the supplied
+    # change_set and and the resource in the change_set
+    # @param [Hyrax::DynamicChangeSet] the change_set that contains the resource whose member file_sets you wish to delete
+    def delete_self_and_file_sets(change_set:)
```
jcoyne: I'm not seeing where this is called.

Have you considered putting this in a `WorkChangeSetPersister`?

app/change_sets/hyrax/dynamic_change_set.rb
carrickr: This solves the problem with FileSet not being namespaced, ex:

```
[9] pry(#<Hyrax::ChangeSetPersister>)> metadata_adapter.query_service.find_members(resource: change_set.resource)[1].class
=> FileSet
[10] pry(#<Hyrax::ChangeSetPersister>)> metadata_adapter.query_service.find_members(resource: change_set.resource)[1].is_a?(FileSet)
=> false
[11] pry(#<Hyrax::ChangeSetPersister>)> metadata_adapter.query_service.find_members(resource: change_set.resource)[1].is_a?(::FileSet)
=> true
```

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: removed

app/change_set_persisters/hyrax/change_set_persister.rb
carrickr: removed

app/change_sets/hyrax/dynamic_change_set.rb
jcoyne: I don't think there are any namespaced changesets anymore (for things that use DynamicChangeSet)

app/change_set_persisters/hyrax/change_set_persister.rb
```diff
@@ -3,43 +3,79 @@
...
+    #  all resources were succesfully deleted
+    def delete_all(change_sets:)
+      change_sets.map do |change_set|
+        delete(change_set: change_set)
+      end
+    end
+
+    # Deletes all file_sets that are members of the resource in the supplied change_set
+    # @param [Hyrax::DynamicChangeSet] the change_set that contains the resource whose member file_sets you wish to delete
+    def cleanup_file_sets(change_set:)
```
jcoyne: Should this be called by some code (other than tests?) How about moving to a WorkChangeSetPersister?

app/change_sets/hyrax/dynamic_change_set.rb
jcoyne: Is there a reason we are ever allowing an instance of FileSet as a valid parameter to DynamicChangeSet#new?

app/change_sets/hyrax/dynamic_change_set.rb
```diff
@@ -1,14 +1 @@
 # frozen_string_literal: true
```
jcoyne: Is there a reason to leave this one line behind?

carrickr: Nope, it was just WIP when you reviewed it

PR #1958, opened 2017-10-25
Convert batch forms to ChangeSets

jcoyne:


app/change_sets/hyrax/batch_upload_change_set.rb
```diff
@@ -0,0 +1,63 @@
...
+      super - [:title]
+    end
+
+    # # On the batch upload, title is set per-file.
+    # def secondary_terms
+    #   super - [:title]
+    # end
+
+    # Override of ActiveModel::Model name that allows us to use our custom name class
+    # def self.model_name
```
jcoyne: I think we want to retain this for routing purposes, but it's unclear if it will be needed or not.

app/change_sets/hyrax/batch_edit_change_set.rb
```diff
@@ -0,0 +1,59 @@
...
+          work = find_resource(doc_id)
+          (schema.keys - ['batch_document_ids', 'append_id']).each do |field|
+            fields[field] ||= []
+            fields[field] = (fields[field] + work[field].to_a).uniq
+          end
+          names << work.to_s
+        end
+      end
+
+      def find_resource(id)
```
carrickr: We could move this over to work_change_set so everyone can get it

app/change_sets/hyrax/batch_edit_change_set.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Hyrax
+  class BatchEditChangeSet < WorkChangeSet
+    property :creator, virtual: true, multiple: true, required: false
```
carrickr: @mbklein this is somewhat parallel to your dynamic generation to about having to list these out, thoughts?

app/change_sets/hyrax/batch_edit_change_set.rb
jcoyne: I don't know why a single work would need  this.

app/change_sets/hyrax/batch_edit_change_set.rb
jcoyne: These can not be dynamic because you don't know which type of objects are in the batch.

jcoyne: Furthermore, the Batch edit page only allows a limited subset of fields to be edited.

PR #1959, opened 2017-10-25
Restore attributes for thumbnail and representative

jcoyne:


app/models/admin_set.rb
```diff
@@ -33,6 +32,7 @@ class AdminSet < Valkyrie::Resource
   attribute :title, Valkyrie::Types::Set
   attribute :description, Valkyrie::Types::Set
   attribute :creator, Valkyrie::Types::Set
+  attribute :thumbnail_id, Valkyrie::Types::SingleValuedString
```
cjcolvar: Is representative_id needed here as well?

jcoyne: Not for an AdminSet.

cjcolvar: Ok, just wanted to make sure since it had it with the mixin.

PR #1963, opened 2017-10-25
Move missing methods to the WorkChangeSet

jcoyne:


jcoyne:
@bmquinn good catch. I've updated to fix.

PR #1982, opened 2017-10-26
add fedora_schema

geekscruff:
Added fedora_schema for core_metadata, basic_metadata and a handful of other predicates scattered throughout Hyrax.

Added it to a separate class for readability.

Question: should I test these go into Fedora, or not?

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -415,12 +415,7 @@ module Hyrax
                     end
     end
 
-    attr_writer :fedora_schema
-    def fedora_schema
-      @fedora_schema ||= {
-        depositor: RDF::URI("http://id.loc.gov/vocabulary/relators/dpt")
-      }
-    end
+    delegate :fedora_schema, :fedora_schema=, to: Hyrax::FedoraSchema
 
     private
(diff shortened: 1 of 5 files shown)


lib/hyrax/configuration.rb
```diff
@@ -417,9 +417,7 @@ def uploader
 
     attr_writer :fedora_schema
     def fedora_schema
-      @fedora_schema ||= {
-        depositor: RDF::URI("http://id.loc.gov/vocabulary/relators/dpt")
-      }
+      @fedora_schema ||= Hyrax::FedoraSchema.fedora_schema
```
jcoyne: as this is already a class attribute. I see no need to memoize this any longer.

jcoyne: Could just be `delegate :fedora_schema, :fedora_schema=, to: Hyrax::FedoraSchema`

geekscruff: that won't mess up overriding it locally will it?

geekscruff: Done. Learning new things every day ... today 'memoize'!

PR #2014, opened 2017-10-28
Pass resource to the after_save persister hook

jcoyne:


cjcolvar:
Should it get passed to all hooks for consistency?

jcoyne:
It's only necessary in the after_save hook, because the save gives it the ID.  The change_set.resource does not have the ID.

cjcolvar:
Would an alternative be to reload the resource in the change set?  Or would that confuse the change set?
A small comment like what you just said might be useful if passing the resource.

carrickr:
Resource could be changed with `change_set.model[:resource] = `.  I suppose the question is are the use cases focused on having the change_set, previous resource, and the new resource for a hook or do they focus on just the new resource and changes?

jcoyne:
@cjcolvar A change set should not be mutable.  It represents the prior state.  The updated resource is the subsequent state.

cjcolvar:
:+1: that clears things up.

PR #2015, opened 2017-10-28
Works controller no longer needs ControllerResource

jcoyne:
The work controller can inherit from ResourceController to avoid code duplication


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -18,23 +18,6 @@ module WorksControllerBehavior
...
 
-    class_methods do
-      def cancan_resource_class
-        Hyrax::ControllerResource
-      end
-    end
-
-    def build_change_set(resource)
-      change_set_class.new(resource,
-                           depositor: current_user.user_key,
```
cjcolvar: Depositor isn't in the `ResourceController` equivalent.  Should it be?

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -18,23 +18,6 @@ module WorksControllerBehavior
       rescue_from WorkflowAuthorizationException, with: :render_unavailable
     end
 
-    class_methods do
-      def cancan_resource_class
```
cjcolvar: This method isn't in `ResourceController`.  Should it be?

jcoyne: No, we don't need this anymore.

app/controllers/concerns/hyrax/works_controller_behavior.rb
jcoyne: Good eye. I've restored this (but it's moved to the private section)

PR #2031, opened 2017-10-30
Refactor Hyrax::CollectionChangeSet terms methods and specs

bmquinn:
- Class methods for Hyrax::CollectionChangeSet terms methods (terms, primary, secondary)
- Fix specs for Hyrax::CollectionChangeSet

### DIFF
--- a/app/change_sets/hyrax/collection_change_set.rb
+++ b/app/change_sets/hyrax/collection_change_set.rb
@@ -4,7 +4,7 @@ module Hyrax
     attr_accessor :current_ability, :repository
 
     delegate :id, :depositor, :permissions, to: :model
-
+    delegate :primary_terms, :secondary_terms, to: :class
     delegate :human_readable_type, :member_ids, :representative_id, :thumbnail_id, to: :model
     property :resource_type, multiple: false, required: false
     property :title, multiple: false, required: true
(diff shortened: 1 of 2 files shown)


app/change_sets/hyrax/collection_change_set.rb
```diff
@@ -36,26 +36,36 @@ def select_files
       Hash[all_files_with_access]
     end
 
-    # Terms that appear above the accordion
-    def primary_terms
-      [:title]
-    end
+    class << self
+      # Combined primary and secondary terms
+      def terms
```
jcoyne: what is this for? Why do we need it?

app/change_sets/hyrax/collection_change_set.rb
```diff
@@ -36,26 +36,36 @@ def select_files
...
-       :publisher,
-       :date_created,
-       :subject,
-       :language,
-       :identifier,
-       :based_near,
-       :related_url,
-       :resource_type]
+      # Terms that appear above the accordion
+      def primary_terms
```
jcoyne: You just changed these instance methods to a class methods, why? That's going to mess up the views which are expecting to call this on the instance.

app/change_sets/hyrax/collection_change_set.rb
bmquinn: Just trying to green up a failing test, the spec was failing on `described_class.terms` with a `NoMethodError` but we can get of it and re-work the test if needed.

app/change_sets/hyrax/collection_change_set.rb
bmquinn: Ok thanks, will revert them.

app/change_sets/hyrax/collection_change_set.rb
jcoyne: I believe that would be taken care of if you ported `autocreate_fields!` into CollectionsChangeSet, but I'm really not sure if `fields` has any utility anymore.

bmquinn: Ok cool, should we just merge this PR now to keep it moving and talk about `fields` with the group after standup tomorrow?

PR #2036, opened 2017-10-30
Restore change_set_persister

cjcolvar:
Without this the controllers which include `WorksControllerBehavior` throw exceptions whenever trying to use the metadata_adapter since it is delegated to `change_set_persister`.  I just copied what was in `AdminSetsController`.  Since all of the controllers so far use this same configuration maybe it makes sense to pull it up into `ResourceController`.


cjcolvar:
Closed in favor of #2035

PR #2039, opened 2017-10-30
Add missing fields to the change set

jcoyne:


carrickr:
With the number of `TODOs:` being brought in here, perhaps drop some tickets or discussion items to track that?

jcoyne:
@carrickr there is now a ticket for each TODO

PR #2054, opened 2017-10-31
Add explanatory note for admin set permissions.

### LINKED ISSUE #2005: Make manager/depositor/viewer permissions more clear
It's very unclear that a manager, depositor, viewer can only see those works that are deposited after they gain permissions to the admin set. I think we need to add language in this screen to make it clear. This has been a surprise to many testers, so I thought I'd write this up as a minimal thing to do before a larger effort happens.

Suggest something like this:
Individuals granted permission below will only have access to items (works?) deposited after their permission is granted.

**More comments at:**

https://github.com/curationexperts/nurax/issues/120

lfarrell:
Fixes #2005 https://github.com/samvera/hyrax/issues/2005

Add explanatory note to make manager/depositor/viewer permissions more clear

### DIFF
--- a/config/locales/hyrax.en.yml
+++ b/config/locales/hyrax.en.yml
@@ -71,6 +71,7 @@ en:
           header:         "Edit Administrative Set"
         form:
           cancel:               "Cancel"
+          note:                 "Users granted a new role will only see works that are deposited after that role has been granted."
           permission_destroy_errors:
             admin_group:        "The repository administrators group cannot be removed"
           permission_update_notices:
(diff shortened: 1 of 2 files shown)


config/locales/hyrax.en.yml
```diff
@@ -71,6 +71,7 @@ en:
           header:         "Edit Administrative Set"
         form:
           cancel:               "Cancel"
+          note:                 "Users granted a new role will only see works that are deposited after that role has been granted."
```
jcoyne: What if they are granted "approver" but they are already a "viewer"?  Then "see works" won't be correct.  How about "...will only gain the role on works that are deposited..."

PR #2059, opened 2017-10-31
Persist instead of save

cjcolvar:


spec/actors/hyrax/actors/file_actor_spec.rb
```diff
@@ -116,7 +116,6 @@ class FileSetWithExtras < FileSet
         actor.revert_to(revision_id)
       end
       it 'does not rely on the default relation' do
-        pending "Hydra::Works::VirusCheck must support other relations: https://github.com/samvera/hyrax/issues/1187"
```
jcoyne: I suspect this is premature.  We currently aren't supporting any relations besides "original_file"

cjcolvar: Ok, the test passed so I thought I'd remove it.  I'll put it back in.

PR #2061, opened 2017-10-31
Persist instead of save

cjcolvar:


app/jobs/inherit_permissions_job.rb
```diff
@@ -18,7 +19,13 @@ def perform(work)
 
       # apply the new and deleted attributes
       file.permissions_attributes = attribute_map
-      file.save!
+      persister.save(resource: file)
```
carrickr: Is this just an result of systematically removing `.save!` in favor of the Valkyrie way?  I'm curious as to why not just go the rest of the way and use a `change_set` and a `change_set_persister` over this?  My understanding of the pattern was to keep the persister.save calls into change_set_persisters and prefer saving through them, eg `change_set_persister.save(change_set: change_set)`, whenever possible.  

In Figgy, I'm seeing the pattern for jobs:

https://github.com/pulibrary/figgy/blob/0d6aa5f62237537085e61922eb5e39d5bacd51ca/app/jobs/browse_everything_ingest_job.rb#L5-L15

app/jobs/inherit_permissions_job.rb
```diff
@@ -18,7 +19,13 @@ def perform(work)
...
     end
   end
+
+  private
+
+    def persister
+      Valkyrie::MetadataAdapter.find(:indexing_persister).persister
+    end
```
carrickr: I get that this is shorter than `Valkyrie::MetadataAdapter.find(:indexing_persister).persister`, but repeating in every job seems ugly.  Since most of these jobs inherit from Hyrax::ApplicationJob as a protected method?  Or ideally just create a `change_set_persister` as per my comment above there and have it for every job.

app/jobs/inherit_permissions_job.rb
cjcolvar: Thanks I'll look into that.  My intention right now is to make broken code not broken for the tests and then later come back to doing it the right way.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -163,7 +163,7 @@ def unlink_from_work
           # Although ActiveFedora clears the children nodes it leaves those fields in Solr populated.
           work.thumbnail = nil if work.thumbnail_id == file_set.id
           work.representative = nil if work.representative_id == file_set.id
-          work.save!
+          persister.save(resource: work)
```
carrickr: Is this just an result of systematically removing `.save!` in favor of the Valkyrie way?  I'm curious as to why not just go the rest of the way and use a `change_set` and a `change_set_persister` over this?  My understanding of the pattern was to keep the persister.save calls into change_set_persisters and prefer saving through them, eg `change_set_persister.save(change_set: change_set)`, whenever possible.  

In Figgy, I'm seeing the pattern for jobs:

https://github.com/pulibrary/figgy/blob/0d6aa5f62237537085e61922eb5e39d5bacd51ca/app/jobs/browse_everything_ingest_job.rb#L5-L15

app/jobs/inherit_permissions_job.rb
carrickr: @cjcolvar Sorry moved my comment over to FileSetActor since that was the first one and actually goes to Fedora

cjcolvar:
@carrickr I moved persister up into `ApplicationJob`.  Could you rereview please?

PR #2077, opened 2017-11-01
remove permission_template and permission_template_access entries whe…

### LINKED ISSUE #1673: at master -- Permissions are not deleted anytime there is 'group' 'admin' for any access
### Descriptive summary

Effected branch:  master

When an Admin Set is deleted, all permissions should also be deleted.  Currently, none of the permissions are deleted because of a check for `'group'` `'admin'`.

### Rationale

Completely deleting an admin set should also delete all related permissions.

### Expected behavior

Deleting an admin set also deletes all related entries in permission_templates table and permission_template_accesses table.

### Actual behavior

The admin set is deleted, but all permissions remain in the two tables.

### Code impacting the expected behavoir

The following prevents the delete from happening...

[app/models/hyrax/permission_template_access.rb#check_if_admin_group](https://github.com/samvera/hyrax/blob/master/app/models/hyrax/permission_template_access.rb#L39-L43)
```
      def check_if_admin_group
        return true unless admin_group?
        errors[:base] << I18n.t('hyrax.admin.admin_sets.form.permission_destroy_errors.admin_group')
        throw :abort
      end
```
with this line calling that method...
[app/models/hyrax/permission_template_access.rb#before_destroy](https://github.com/samvera/hyrax/blob/master/app/models/hyrax/permission_template_access.rb#L7)
```
 before_destroy :check_if_admin_group
```

And permission_template won't delete unless all the access entries delete.

1. Login as an admin
1. Create an admin set

Verify that there is one entry in permission_templates table with...
```
   admin_set_id=new_admin_set.id
```

Verify that there are two entries in permission_template_accesses table with...
``` 
   agent_type='group' agent_id='admin'; access='manage'; permission_template_id=id of new permission template
   agent_type='user' agent_id='creator user's key'; access='manage'; permission_template_id=id of new permission template; 
```
### Related work

#1664 Add collection permissions to solr doc read/edit access -- This PR tries to do a similar delete for Collections

blancoj:
Fixes #1673 

When an admin set is deleted remove entries from permission_templates and permission_template_accesses.

### DIFF
--- a/spec/models/hyrax/permission_template_access_spec.rb
+++ b/spec/models/hyrax/permission_template_access_spec.rb
@@ -25,19 +25,118 @@ RSpec.describe Hyrax::PermissionTemplateAccess do
       end
     end
     describe '#destroy' do
-      it 'aborts the destroy operation' do
+      it 'destroys the permission template access record' do
         subject.destroy
-        expect(subject).not_to be_destroyed
-        expect(subject.errors.messages[:base]).to include 'The repository administrators group cannot be removed'
+        expect(subject).to be_destroyed
       end
     end
   end
+
   context 'with an agent that is not the admin users group' do
     let(:agent_type) { 'user' }
     let(:agent_id) { 'foobar' }
 
     describe '#label' do
       it 'returns the repo admins label' do
+        expect(subject.label).to eq agent_id
+      end
+    end
+    describe '#admin_group?' do
+      it 'returns true' do
+        expect(subject).not_to be_admin_group
+      end
+    end
+    describe '#destroy' do
+      it 'carries out the destroy operation' do
+        subject.destroy
+        expect(subject).to be_destroyed
+        expect(subject.errors).to be_empty
+      end
+    end
+  end
+
+  context 'with a user that is an admin set viewer' do
+    let(:agent_type) { 'user' }
+    let(:agent_id) { 'foobar' }
+
+    let(:permission_template_access) do
+      create(:permission_template_access,
+             :view,
+             permission_template: admin_set.permission_template,
+             agent_type: agent_type,
+             agent_id: agent_id)
+    end
+
+    subject { permission_template_access }
+
+    describe '#label' do
+      it 'returns the user label' do
+        expect(subject.label).to eq agent_id
+      end
+    end
+    describe '#admin_group?' do
+      it 'returns true' do
+        expect(subject).not_to be_admin_group
+      end
+    end
+    describe '#destroy' do
+      it 'carries out the destroy operation' do
+        subject.destroy
+        expect(subject).to be_destroyed
+        expect(subject.errors).to be_empty
+      end
+    end
+  end
+
+  context 'with a user that is an admin set depositor' do
+    let(:agent_type) { 'user' }
+    let(:agent_id) { 'foobar' }
+
+    let(:permission_template_access) do
+      create(:permission_template_access,
+             :deposit,
+             permission_template: admin_set.permission_template,
+             agent_type: agent_type,
+             agent_id: agent_id)
+    end
+
+    subject { permission_template_access }
+
+    describe '#label' do
+      it 'returns the user label' do
+        expect(subject.label).to eq agent_id
+      end
+    end
+    describe '#admin_group?' do
+      it 'returns true' do
+        expect(subject).not_to be_admin_group
+      end
+    end
+    describe '#destroy' do
+      it 'carries out the destroy operation' do
+        subject.destroy
+        expect(subject).to be_destroyed
+        expect(subject.errors).to be_empty
+      end
+    end
+  end
+
+  context 'with a user that is an admin set manager' do
+    let(:agent_type) { 'user' }
+    let(:agent_id) { 'foobar' }
+
+    let(:permission_template_access) do
+      create(:permission_template_access,
+             :manage,
+             permission_template: admin_set.permission_template,
+             agent_type: agent_type,
+             agent_id: agent_id)
+    end
+
+    subject { permission_template_access }
+
+    describe '#label' do
+      it 'returns the user label' do
         expect(subject.label).to eq agent_id
       end
     end
--- a/app/models/hyrax/permission_template_access.rb
+++ b/app/models/hyrax/permission_template_access.rb
@@ -4,7 +4,6 @@ module Hyrax
   class PermissionTemplateAccess < ActiveRecord::Base
     self.table_name = 'permission_template_accesses'
     belongs_to :permission_template
-    before_destroy :check_if_admin_group
 
     VIEW = 'view'.freeze
     DEPOSIT = 'deposit'.freeze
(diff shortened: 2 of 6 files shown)


app/models/admin_set.rb
```diff
@@ -74,6 +74,11 @@ def permission_template
     Hyrax::PermissionTemplate.find_by!(admin_set_id: id)
   end
 
+  def delete_permission_template_access
+    Hyrax::PermissionTemplateAccess.where(:permission_template_id => permission_template.id).delete_all
```
mjgiarlo: Doesn't the PT have a `dependent: :destroy` relationship with the PTAs? I don't *think* this should be necessary but I could be mistaken.

mjgiarlo:
I believe the ultimate solution for this is that this behavior:

https://github.com/samvera/hyrax/blob/master/app/models/hyrax/permission_template_access.rb#L39

Should prevent users from crafting the right URL/params to delete admin groups from the UI, *but* should not prevent an AS deletion from causing PT and PTA deletions. Perhaps we should handle the validation in a higher layer?

elrayle:
I tested the proposed solution in the UI and it does work as expected.  We could merge with this fix, or we could address this at a different level.

As an example, for Collection Extensions, based on recommendations from @ggeisler, we have been allowing users to click Remove buttons (and others) and displaying a dialog when they can't perform the action stating why.  So in this example, the message would be something like 'Administrative users cannot be removed from the Manager role'.  This is seen as a teaching aid to educate the user why they can do something in some circumstance, but not others.  In this particular case, the UI would never try to destroy the access entry for admin users in the Manager role.

If we went this route, I believe it would be safe to modify the PermissionTemplateAccess model to remove the after_destroy check for admin group.

@mjgiarlo Is this the kind of thing you had in mind as validation in a higher layer?

elrayle:
Note that currently at master in the UI, the Remove button is disabled for Administrators in the manager role.  It is also disabled if you make the admin group a depositor or viewer, which probably isn't correct.  You should be able to Remove admins from the depositor and viewer roles.

elrayle:
I entered a related bug issue (#2089) that I discovered while testing this.  Adding the same participant twice results in a duplicate access entry in the permission_template_accesses table.  If the duplicate is the admin group, then Repository Administrators shows up in the list twice and neither can be removed.

image

elrayle:
### Options

* **Option A:** accept this PR when requested changes are applied
* **Option B:** depend on the UI to enforce the 'can not delete admin' condition by having the Remove button be disabled (for manager only) and remove the before_destroy code in permission_template_access
* **Option C:** allow Remove to be clicked and display a dialog telling the user why they cannot remove admins as managers

My vote would be to implement option B in master and option C in collections-sprint branch.

@mjgiarlo @vantuyls What is your preference?

blancoj:
I just want to add one more suggestion, but if it makes things more complicated, please let me know. Why even bother listing Repository/Admin in list of Managers for an Admin Set, since it's understood that the Repository/Admin has ALL privileges, then in permission_template_access.rb ( the model ) we would not have to do: `before_destroy :check_if_admin_group`

We could just destroy away when an Admin Set is `Deleted`, since the UI would never give users access to remove an entry in the permission_template_accesses table specifically for the Repository/Admin.  
So ....
(1) Change UI not to show  Repository/Admin as Manager for all the Admin Sets ( this is understood - admins can do anything! )
(2) Remove the `before_destroy` check.

And we're done!

vantuyls:
@mjgiarlo @elrayle thoughts on @blancoj's suggestion?

vantuyls:
@elrayle i think i agree with your preference for b an c.

mjgiarlo:
Don't have cycles to dig into this before Connect. Sorry!

elrayle:
@blancoj Decision from slack:  For 2.0, this PR should remove the before_delete callback from model/permission_template_access.rb and the method that before_delete calls.

spec/models/hyrax/permission_template_access_spec.rb
```diff
@@ -54,4 +54,106 @@
...
   end
+
+  context 'with a user that is an admin set viwer' do
+    let(:agent_type) { 'user' }
+    let(:agent_id) { 'foobar' }
+    let(:access) { 'view' }
+
+    let(:permission_template_access) do
+      create(:permission_template_access,
+             :view,
```
elrayle: Shouldn't this use the access variable you create above?

spec/models/hyrax/permission_template_access_spec.rb
```diff
@@ -54,4 +54,106 @@
...
+        expect(subject).to be_destroyed
+        expect(subject.errors).to be_empty
+      end
+    end
+  end
+
+  context 'with a user that is an admin set depositor' do
+    let(:agent_type) { 'user' }
+    let(:agent_id) { 'foobar' }
+    let(:access) { 'deposit' }
```
elrayle: Use Hyrax::PermissionTemplateAccess::DEPOSIT instead of 'deposit'

spec/models/hyrax/permission_template_access_spec.rb
```diff
@@ -54,4 +54,106 @@
       end
     end
   end
+
+  context 'with a user that is an admin set viwer' do
+    let(:agent_type) { 'user' }
+    let(:agent_id) { 'foobar' }
+    let(:access) { 'view' }
```
elrayle: Use Hyrax::PermissionTemplateAccess::VIEW instead of 'view'.  I know that other tests just use the string, but whenever we touch a test or add more, we should switch to using the constant.

spec/models/hyrax/permission_template_access_spec.rb
```diff
@@ -54,4 +54,106 @@
...
+        expect(subject).to be_destroyed
+        expect(subject.errors).to be_empty
+      end
+    end
+  end
+
+  context 'with a user that is an admin set manager' do
+    let(:agent_type) { 'user' }
+    let(:agent_id) { 'foobar' }
+    let(:access) { 'manage' }
```
elrayle: Use Hyrax::PermissionTemplateAccess::MANAGE instead of 'manage'

spec/models/hyrax/permission_template_access_spec.rb
```diff
@@ -54,4 +54,106 @@
...
+
+    let(:permission_template_access) do
+      create(:permission_template_access,
+             :view,
```
elrayle: Use `access` instead of `:view`

elrayle:
Hyrax 2.0 will be released any time now.  This fix will be applied to 2.0.1

Can you make this change here as well?

Edit:  https://github.com/samvera/hyrax/blob/master/app/views/hyrax/admin/admin_sets/_form_participant_table.html.erb#L18

CHANGE FROM:  `<% if g.admin_group? %>`
CHANGE TO:  `<% if g.admin_group? && g.access == Hyrax::PermissionTemplateAccess::MANAGE %>`

This allows a user to add admin to Depositor or Viewer role and be able to remove them from it.  It continues to disable the Delete button for Manager role.

PR #2080, opened 2017-11-01
Update collections_controller_behavior queries

bmquinn:
- `query_for_collection_members` relies on new `member_object_ids` relationship
- No longer populating `@response` in `#query_collection_members` due to a solr query needed for each member rather than one query to find all members as before.
- Removed `#params_for_members_query` method, which could be overridden in hyrax applications and may need a comment in release notes.

### DIFF
--- a/app/controllers/concerns/hyrax/collections_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/collections_controller_behavior.rb
@@ -81,24 +81,18 @@ module Hyrax
       end
 
       # Queries Solr for members of the collection.
-      # Populates @response and @member_docs similar to Blacklight Catalog#index populating @response and @documents
+      # Populates @member_docs similar to Blacklight Catalog#index populating @response and @documents
       def query_collection_members
-        params[:q] = params[:cq]
-        @response = repository.search(query_for_collection_members)
-        @member_docs = @response.documents
+        @member_docs = query_for_collection_members
       end
 
-      # @return <Hash> a representation of the solr query that find the collection members
+      # @return <Array> an array of the solr documents found by the query
       def query_for_collection_members
-        member_search_builder.with(params_for_members_query).query
-      end
-
-      # You can override this method if you need to provide additional inputs to the search
-      # builder. For example:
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/hyrax/collections_controller_behavior.rb
```diff
@@ -81,24 +81,18 @@ def collection_params
...
-      end
-
-      # You can override this method if you need to provide additional inputs to the search
-      # builder. For example:
-      #   search_field: 'all_fields'
-      # @return <Hash> the inputs required for the collection member search builder
-      def params_for_members_query
-        params.merge(q: params[:cq])
+        members = Hyrax::Queries.find_by(id: params[:id]).member_object_ids
+        members.collect do |member_id|
```
jcoyne: This seems like we've expanded a single query into a 2N + 1 query.  I'm not sure I understand why we'd favor this.

app/controllers/concerns/hyrax/collections_controller_behavior.rb
```diff
@@ -81,24 +81,18 @@ def collection_params
...
+          argz = { fq: "id:#{member_id}", q: params[:cq] }
```
jcoyne: This is no longer applying the access controls that member_search_builder was.

jcoyne: Nor is paginating.

app/controllers/concerns/hyrax/collections_controller_behavior.rb
bmquinn: Definitely agree that one query is better. Looking into making this work.

app/controllers/concerns/hyrax/collections_controller_behavior.rb
bmquinn: Didn't catch that the search_builder was doing all that work, will revert this.

carrickr:
Github weirdness when trying to forcepush to Brendan's branch from my box, closed in favor of #2099

PR #2083, opened 2017-11-01
Generic Type indexer

geekscruff:


app/indexers/hyrax/generic_type_indexer.rb
```diff
@@ -0,0 +1,34 @@
...
+      attr_reader :resource
+
+      def generic_type
+        if resource.is_a?(AdminSet) # must be first as doesn't have a falsey method for collection
+          { generic_type_field => 'Admin Set' }
+        elsif resource.collection?
+          { generic_type_field => 'Collection' }
+        elsif resource.work?
+          { generic_type_field => 'Work' }
+        else
```
cjcolvar: What about adding File Set?

spec/indexers/hyrax/generic_type_indexer_spec.rb
```diff
@@ -0,0 +1,48 @@
+RSpec.describe Hyrax::GenericTypeIndexer do
+  let(:user) { create(:user) }
+  let(:admin_set) { AdminSet.new(title: ['Some admin set']) }
```
cjcolvar: How about changing these defining lets into `build(:admin_set)`, `build(:work)`, `build(:collection)`, and `build(:file_set)`?  You don't appear to be using `title` so a simple `build` with no args besides factory should work I think.

spec/indexers/hyrax/generic_type_indexer_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      expect(work_solr_document.fetch(:generic_type_sim)).to eq 'Work'
+    end
+  end
+
+  context "when the object is a Fileset" do
+    let(:fs_service) { described_class.new(resource: fs) }
+
+    subject(:fs_solr_document) { fs_service.to_solr }
+
+    it "does not index generic_type" do
```
cjcolvar: Why not support File Set?
You could make another test for a plain old `Valkyrie::Resource` for testing an unknown model.

spec/indexers/hyrax/generic_type_indexer_spec.rb
```diff
@@ -0,0 +1,48 @@
+RSpec.describe Hyrax::GenericTypeIndexer do
+  let(:user) { create(:user) }
```
cjcolvar: Do you need this user?

spec/indexers/hyrax/generic_type_indexer_spec.rb
geekscruff: In current Hyrax, FileSet doesn't have a GenericType set. I could add it ... but I'm not sure what it's return would be, 'File Set' or 'File' (as human_readable_type for file_set gets set to File). So I think I'd rather leave it as is?

cjcolvar: 👍 makes sense.

PR #2085, opened 2017-11-01
human readable type indexer

geekscruff:
Moved indexing out of HumanReadableType into a HumanReadableTypeIndexer


spec/indexers/hyrax/human_readable_type_indexer_spec.rb
```diff
@@ -0,0 +1,12 @@
+RSpec.describe Hyrax::HumanReadableTypeIndexer do
+  subject(:solr_document) { service.to_solr }
+
+  let(:user) { create(:user) }
```
cjcolvar: Do you need user?

spec/indexers/hyrax/human_readable_type_indexer_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+  let(:service) { described_class.new(resource: work) }
+  let(:work) { create_for_repository(:work) }
```
cjcolvar: I think `build(:work)` would suffice here.

spec/indexers/hyrax/human_readable_type_indexer_spec.rb
geekscruff: You are right, no. I've changed this.

spec/indexers/hyrax/human_readable_type_indexer_spec.rb
geekscruff: Done.

PR #2090, opened 2017-11-02
Update the file_set_controller to use ChangeSets

jcoyne:


spec/change_sets/hyrax/file_set_change_set_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(subject.terms).not_to include(:date_uploaded)
+    end
+  end
+
+  describe "field initialization" do
+    before do
+      change_set.prepopulate!
+    end
+    it 'initializes multivalued fields' do
+      expect(subject.title).to eq ['']
```
bmquinn: Should this be `expect(subject.title).to eq []`?

jcoyne: Nope, it initializes it to an array with a single blank string. Otherwise it wouldn't draw any fields on the page.

bmquinn: Ah, gotcha. Thanks!

PR #2092, opened 2017-11-02
Use FileSetChangeSets in jobs

cjcolvar:


app/jobs/inherit_permissions_job.rb
```diff
@@ -6,23 +6,16 @@ class InheritPermissionsJob < Hyrax::ApplicationJob
...
     file_sets = Hyrax::Queries.find_members(resource: work, model: ::FileSet)
-    file_sets.each do |file|
-      file_change_set = Hyrax::FileSetChangeSet.new(file)
-      file_change_set.read_users = work.read_users
-      file_change_set.read_groups = work.read_groups
-      file_change_set.edit_users = work.edit_users
-      file_change_set.edit_groups = work.edit_groups
-      file_change_set.sync
-      change_set_persister.save(change_set: file_change_set)
+    change_set_persister.buffer_into_index do |persister|
```
jcoyne: Should this be FileSetChangeSetPersister?

cjcolvar: Yes, now that it exists. 😃

cjcolvar: @jcoyne  I wasn't sure who the user is in these job contexts so instead of skipping the notification I set user to be the file_set's depositor.  Does this seem right to you?

jcoyne: WFM.

PR #2093, opened 2017-11-02
Provide better color contrast.

### LINKED ISSUE #2076: Contrast Errors - Create New Generic Work
Originally reported in https://github.com/curationexperts/nurax/issues/127 by @mtribone 

**Descriptive summary**

When creating a new generic work and checking the contrast errors of the form using the WAVE accessibility toolbar, there are several contrast errors for things like hyperlinks and visibility badges/labels. Text is present that has a contrast ratio less than 4.5:1, or large text (larger than 18 point or 14 point bold) has a contrast ratio less than 3:1.

Suggestion: I believe that these colors from ScholarSphere will do the trick. The very specific CSS syntax is to override Bootstrap coming from 3 places. May be different in Hyrax. Using important! may work also, but to me is "more icky" than the very specific CSS syntax.

https://github.com/psu-stewardship/scholarsphere/pull/552/files#diff-64c7b543e3771590e928b0d4b47e00dbR31

**They are:**

Required and Institution (2.09:1)
Public (2.48:1)
Embargo and Lease (1.95:1)
Private (3.96:1)
Rationale

Very low contrast between foreground and background colors. Adequate contrast is necessary for all users, especially users with low vision.

**Expected behavior**

Expect sufficient contrast ratio for all text and UI elements

**Actual behavior**

Some text and UI elements do not meet requirements.

Select Works from the left-hand menu
On the works screen select "Add new work"
Select "generic Work" from the modal
Validate the form with the WAVE toolbar

lfarrell:
Fixes #2076 

Provides better color contrast for labels and form submit link. Selected colors pass WAVE validation. The breadcrumb link colors do not pass WAVE validation, but they seem outside the scope of this ticket.

### DIFF
--- a/app/assets/stylesheets/hyrax/_styles.scss
+++ b/app/assets/stylesheets/hyrax/_styles.scss
@@ -71,3 +71,13 @@ label.disabled {
   text-decoration:underline;
   background-color:#0088CC;
 }
+
+/* Label and Button overrides for better contrast ratio */
+html > body .label-success,
+html > body .btn-success { background-color: #387f38; }
+html > body .label-danger,
+html > body .btn-danger { background-color: #d33a35; }
+html > body .label-info,
+html > body .btn-info { background-color: #2c76c7; }
+html > body .label-warning,
+html > body .btn-warning { background-color: #565653; }
--- a/app/assets/stylesheets/hyrax/_styles.scss
+++ b/app/assets/stylesheets/hyrax/_styles.scss
@@ -71,16 +71,3 @@ label.disabled {
   text-decoration:underline;
   background-color:#0088CC;
 }
-
-// Label and Button overrides for better contrast ratio
-html > body .label-success,
-html > body .btn-success { background-color: $brand-success; }
-
-html > body .label-danger,
-html > body .btn-danger { background-color: $brand-danger; }
-
-html > body .label-info,
-html > body .btn-info { background-color: $brand-info; }
-
-html > body .label-warning,
-html > body .btn-warning { background-color: $brand-warning; }
--- a/app/assets/stylesheets/hyrax/_styles.scss
+++ b/app/assets/stylesheets/hyrax/_styles.scss
@@ -72,12 +72,15 @@ label.disabled {
   background-color:#0088CC;
 }
 
-/* Label and Button overrides for better contrast ratio */
+// Label and Button overrides for better contrast ratio
 html > body .label-success,
-html > body .btn-success { background-color: #387f38; }
+html > body .btn-success { background-color: $brand-success; }
+
 html > body .label-danger,
-html > body .btn-danger { background-color: #d33a35; }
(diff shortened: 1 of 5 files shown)


mjgiarlo:
I've approved this change, @lfarrell. Feel free to merge when the build passes.

Could you also attach a screenshot showing what the new colors look like? Thanks!

app/assets/stylesheets/hyrax/_styles.scss
```diff
@@ -71,3 +71,13 @@ label.disabled {
   text-decoration:underline;
   background-color:#0088CC;
 }
+
+/* Label and Button overrides for better contrast ratio */
+html > body .label-success,
+html > body .btn-success { background-color: #387f38; }
```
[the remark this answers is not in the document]
jcoyne: These should be set using the bootstrap sass variables.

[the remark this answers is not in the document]
jcoyne: See https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L18-L22

mjgiarlo:
@lfarrell Can you make the change @jcoyne requested and resubmit? Thank you.

lfarrell:
Updated the colors as SASS variables. Here's the current color scheme that passes accessibility report:

colors

app/assets/stylesheets/hyrax/_variables.scss
```diff
@@ -16,3 +16,9 @@ $marketing-text-shadow: 1px 1px 0 rgba(0, 0, 0, 0.1) !important;
 $footer-top-margin: 3 * $padding-small-vertical !default;
 $footer-dashboard-top-margin: 0 !default;
 $footer-height: 118px !default;
+
+// Brand colors
+$brand-danger: #d33a35;
```
jcoyne: These are bootstrap variables, not hyrax variables, so I wonder if this is the right location.  If you put them here they may not be used in bootstrap.  Also we should set `!default` so the consumer could provide their own defaults.

jcoyne: We need these variables set prior to this: https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/templates/hyrax.scss#L8 but the code you provided doesn't get included until https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/templates/hyrax.scss#L15

lfarrell: Yes, I ran into that. Removing the default allowed the variables to override the bootstrap defaults without changing the load order. I can create a new SASS file and load it before bootstrap. Does that seem like a reasonable alternative?

mjgiarlo:
@jcoyne Have time for another review?

mjgiarlo:
@jcoyne I'm merging for now. We can pop an amendment in if we need one.

PR #2099, opened 2017-11-02
Update the CollectionMemberSearchBuilder to use -id in member queries

carrickr:
Fixes up the collection_controller spec tests

Same goal as PR #2080 , but we had some weirdness when I tried to force push up from my machine, so we moved the changes to a new branch.


jcoyne:
👍

PR #2103, opened 2017-11-03
Add the saved work to the Actors::Environment

jcoyne:


app/actors/hyrax/actors/base_actor.rb
```diff
@@ -49,6 +49,17 @@ def destroy(env)
 
       private
 
+
+      private
```
cjcolvar: Double private?

PR #2109, opened 2017-11-04
Retreive files from Valkyrie

cjcolvar:
One of the tests (spec/controllers/hyrax/downloads_controller_spec.rb:84) randomly fails with `uninitialized constant Hyrax::ObjectNotFoundError`.  If you run it directly it always fails but when the whole file is run it only fails some of the time.  I can't find what would be different from a different run order.

### DIFF
--- a/app/controllers/hyrax/single_use_links_viewer_controller.rb
+++ b/app/controllers/hyrax/single_use_links_viewer_controller.rb
@@ -14,7 +14,12 @@ module Hyrax
 
     def download
       raise not_found_exception unless single_use_link.path == hyrax.download_path(id: @asset)
-      send_content
+      # send_content
+      # TODO: copied from DownloadsController#show and isn't handling HEAD or range requests
+      response.headers['Accept-Ranges'] = 'bytes'
+      response.headers['Content-Length'] = File.size(file).to_s
+      response_options = derivative_download_options.tap { |options| options[:disposition] = 'attachment' }
+      send_file file, response_options
     end
 
     def show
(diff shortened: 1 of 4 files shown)


app/controllers/hyrax/single_use_links_viewer_controller.rb
```diff
@@ -14,7 +14,12 @@ class SingleUseLinksViewerController < DownloadsController
 
     def download
       raise not_found_exception unless single_use_link.path == hyrax.download_path(id: @asset)
-      send_content
+      # send_content
```
jcoyne: remove?

spec/controllers/hyrax/downloads_controller_spec.rb
```diff
@@ -56,7 +56,7 @@
           end
 
           it 'sends requested file content' do
-            get :show, params: { id: file_set, file: 'thumbnail' }
+            get :show, params: { id: file_set.to_param, file: 'thumbnail' }
```
jcoyne: Why do we have to do `to_param`? Isn't that something rails does automatically?

cjcolvar:
@jcoyne I dealt with your comments.

spec/controllers/hyrax/downloads_controller_spec.rb
```diff
@@ -56,7 +56,7 @@
           end
 
           it 'sends requested file content' do
-            get :show, params: { id: file_set, file: 'thumbnail' }
+            get :show, params: { id: file_set.id, file: 'thumbnail' }
```
jcoyne: Why are we adding `.id` here? Is that necessary?  It should be fine to just say `id: file_set`

cjcolvar: Tests break if we don't have it.  It seems like we needed it for other controller tests too.

cjcolvar:
@jcoyne I did the same casting as in #2117.  Should we make abilities handle Valkyrie::ID instead of just String instead of this fix?

PR #2117, opened 2017-11-05
Request with id instead of resource

cjcolvar:


jcoyne:
Why is this necessary?  It should work the way it was before this change.

cjcolvar:
@jcoyne does this look better?

jcoyne:
I think https://github.com/samvera-labs/valkyrie/pull/305 would have fixed this.

PR #2123, opened 2017-11-06
Use WorksSearchBuilder instead of WorkRelation

cjcolvar:
After this PR there are two remaining references to WorkRelation: `Hyrax::Statistics::Works::OverTime` and `Hyrax::ResourceSync::ResourceListWriter`

### DIFF
--- a/app/change_sets/hyrax/work_change_set.rb
+++ b/app/change_sets/hyrax/work_change_set.rb
@@ -33,6 +33,8 @@ module Hyrax
     collection :permissions, virtual: true
     collection :work_members, virtual: true
 
+    property :admin_set_id, virtual: false
+
     validate :validate_lease
     validate :validate_embargo
     validate :validate_release_type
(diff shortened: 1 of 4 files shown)


app/services/hyrax/workflow/status_list_service.rb
```diff
@@ -38,7 +41,10 @@ def search_solr
           actionable_roles = roles_for_user
           logger.debug("Actionable roles for #{user.user_key} are #{actionable_roles}")
           return [] if actionable_roles.empty?
-          WorkRelation.new.search_with_conditions(query(actionable_roles), rows: 1000)
+          work_query = WorksSearchBuilder.new([:filter_models], self).query
+          work_query["fq"] += query(actionable_roles)
```
cjcolvar: I couldn't get the query to look right when using `WorksSearchBuilder.with(fq: query(actionable_roles).rows(1000)`.

app/change_sets/hyrax/work_change_set.rb
```diff
@@ -33,6 +33,8 @@ class WorkChangeSet < Valkyrie::ChangeSet
     collection :permissions, virtual: true
     collection :work_members, virtual: true
 
+    property :admin_set_id, virtual: false
```
jcoyne: `virtual: false` is the default, so this can be omitted.

PR #2128, opened 2017-11-07
Extends `ImportUrlJob` and `CreateWithRemoteFilesActor` to handle auth headers for cloud storage providers

### LINKED ISSUE #1850: Upload from Google Drive is not working
### Descriptive summary
Uploading from Cloud via Browse-Everything gem is not working for Google Drive:  See related ticket from @scherztc for description of issue https://github.com/samvera/browse-everything/issues/177

jrgriffiniii:
Extends ImportUrlJob#copy_remote_file and CreateWithRemoteFilesActor#create_file_from_url in order to handle the auth. header exposed by the Google API's

Fixes #1850 

Please note that this requires some additional testing against https://github.com/samvera/browse-everything/pull/178 before it can be submitted for review.

### DIFF
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -20,7 +20,7 @@ class ImportUrlJob < Hyrax::ApplicationJob
 
   # @param [FileSet] file_set
   # @param [Hyrax::BatchCreateOperation] operation
-  def perform(file_set:, operation:, headers: {})
+  def perform(file_set, operation, headers = {})
     operation.performing!
     user = User.find_by_user_key(file_set.depositor)
     uri = URI(file_set.import_url)
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -10,19 +10,23 @@ class ImportUrlJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
 
   before_enqueue do |job|
-    operation = job.arguments.last
     operation.pending_job(job)
+  end
+
+  # Retrieves the operation for the job
+  def operation
+    arguments.reduce(:merge).fetch(:operation)
   end
 
   # @param [FileSet] file_set
   # @param [Hyrax::BatchCreateOperation] operation
-  def perform(file_set, operation)
+  def perform(file_set:, operation:, headers: {})
     operation.performing!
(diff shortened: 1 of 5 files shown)


mjgiarlo:
@jrgriffiniii :clap: Looks good to me! Still WIP?

mjgiarlo:
Heads-up, @vantuyls @julesies

mjgiarlo:
@jrgriffiniii How WIP-y is this?

jrgriffiniii:
Sorry for the delay @mjgiarlo , I've managed to reproduce a bug which must be resolved within the scope of https://github.com/samvera/browse-everything/pull/178.  It requires testing against the Google API's (there are issues with the management of access codes within the Driver).

mjgiarlo:
OK, thanks, @jrgriffiniii

jrgriffiniii:
https://github.com/jnunemaker/httparty/blob/master/lib/httparty/request.rb#L387 is keeping `HTTParty` from properly downloading the file when the request is made against the Google Drive API.  When just the access token is sent (without additional cookies captured from the redirect), it appears to download the file.

app/jobs/import_url_job.rb
```diff
@@ -10,19 +10,23 @@ class ImportUrlJob < Hyrax::ApplicationJob
...
 
+  # Retrieves the operation for the job
+  def operation
+    arguments.reduce(:merge).fetch(:operation)
+  end
+
   # @param [FileSet] file_set
   # @param [Hyrax::BatchCreateOperation] operation
-  def perform(file_set, operation)
+  def perform(file_set:, operation:, headers: {})
```
cjcolvar: Is it possible to do this change without changing to keyword arguments?  I think the switch to keyword arguments would change the method signature and thus would be backwards incompatible.

spec/spec_helper.rb
```diff
@@ -72,6 +72,7 @@ def coverage_needed?
   browser_options.args << '--headless'
   browser_options.args << '--disable-gpu'
   browser_options.args << '--no-sandbox'
+  browser_options.args << '--window-size=1920, 1080'
```
cjcolvar: Can you please explain why this change is being added?  Was it needed for tests to pass? Or could it be removed?

jrgriffiniii: This was required to pass, I can rebase and attempt to remove this

PR #2135, opened 2017-11-08
Updating documentation

jeremyf:


app/models/sipity/method.rb
```diff
@@ -1,4 +1,14 @@
 module Sipity
+  # When a Sipity::Action is taken, each Sipity::Method is loaded and
+  # it's service_name is instantiated and called (all of this done via
+  # the Hyrax::Workflow::ActionTakenService).
+  #
+  # @example
```
no-reply: [`@example`](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#example) is meant for code blocks. Should this be `@note`?

PR #2136, opened 2017-11-08
[WIP] Remove Solr suggest fields because of @awead's presentation at …

mjgiarlo:
…SC2017, to see if 1) anything breaks, and 2) the test suite's performance is affected

### DIFF
--- a/solr/config/schema.xml
+++ b/solr/config/schema.xml
@@ -59,32 +59,32 @@
 
   <types>
     <fieldType name="string" class="solr.StrField" sortMissingLast="true" />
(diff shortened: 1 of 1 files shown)


jcoyne:
@awead does suggest power the Blacklight autocomplete functionality?

awead:
@jcoyne not sure. I didn't really investigate any further, other than noting that indexing performance improved when we disabled these fields. It be worth finding out if there's something in the `suggest` field configuration that's causing the problem, instead of disabling it entirely.

mjgiarlo:
@jcoyne does that work in Hyrax?

jcoyne:
@mjgiarlo it does, but it's not turned on in our default templates.

PR #2137, opened 2017-11-08
after deleting an admin set from the dashboard return to the appropriate place.

### LINKED ISSUE #1968: Collections: Delete admin set should forward to my collections
### Descriptive summary

When the user executes delete admin set from collection index action menu, should forward to Dashboard -> Collections -> Your Collections when delete completes.

### Rationale

The delete action is initiated from Dashboard -> Collections and should remain there once completed.

### Expected behavior

Dashboard -> Collections -> for any admin set -> action menu: Delete should forward to /dashboard/my/collections

### Actual behavior

Dashboard -> Collections -> for any admin set -> action menu: Delete forwards to /admin/admin_sets

1. Login as an admin user
1. Create an admin set
1. Dashboard -> Collections -> Your Collections
1. for the new admin set, click Delete in the action menu
1. click Delete in the confirmation dialog

Verify that when the delete completes forwards to Dashboard -> Collections -> Your Collections

### Related work

Issue #1507 Update code to handle delete on collection index
Issue #1972 Collections: admin sets action menu should match collections action menu
Issue #2069 BUG Collections: clicking Edit collection in the action menu for an admin set should forward to edit form

blancoj:
Fixes #1968

### DIFF
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -71,7 +71,7 @@ module Hyrax
 
     def destroy
       if @admin_set.destroy
-        redirect_to hyrax.admin_admin_sets_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
+        after_delete_success
       else
         redirect_to hyrax.admin_admin_set_path(@admin_set), alert: @admin_set.errors.full_messages.to_sentence
       end
@@ -139,6 +139,16 @@ module Hyrax
 
       def repository_class
         blacklight_config.repository_class
+      end
+
+      def after_delete_success
+        if request.referer.include? "my/collections"
+          redirect_to hyrax.my_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
+        elsif request.referer.include? "collections"
+          redirect_to hyrax.dashboard_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
+        else
+          redirect_to hyrax.my_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
+        end
       end
   end
 end
--- a/spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
+++ b/spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
@@ -148,9 +148,35 @@ RSpec.describe Hyrax::Admin::AdminSetsController do
 
       context "with empty admin set" do
         it "deletes the admin set" do
+          controller.request.set_header("HTTP_REFERER", "/admin/admin_sets")
           delete :destroy, params: { id: admin_set }
+
           expect(response).to have_http_status(:found)
-          expect(response).to redirect_to(admin_admin_sets_path)
+          expect(response).to redirect_to(my_collections_path)
+          expect(flash[:notice]).to eq "Administrative set successfully deleted"
+          expect(AdminSet.exists?(admin_set.id)).to be false
+        end
+      end
+
+      context "with empty admin set and referrer from the my/collections dashboard" do
+        it "deletes the admin set" do
+          controller.request.set_header("HTTP_REFERER", "/my/collections")
+          delete :destroy, params: { id: admin_set }
+
+          expect(response).to have_http_status(:found)
+          expect(response).to redirect_to(my_collections_path)
+          expect(flash[:notice]).to eq "Administrative set successfully deleted"
+          expect(AdminSet.exists?(admin_set.id)).to be false
+        end
+      end
+
+      context "with empty admin set and referrer from the /collections dashboard" do
+        it "deletes the admin set" do
+          controller.request.set_header("HTTP_REFERER", "/collections")
+          delete :destroy, params: { id: admin_set }
+
+          expect(response).to have_http_status(:found)
+          expect(response).to redirect_to(dashboard_collections_path)
           expect(flash[:notice]).to eq "Administrative set successfully deleted"
           expect(AdminSet.exists?(admin_set.id)).to be false
         end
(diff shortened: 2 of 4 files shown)


app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -71,7 +71,13 @@ def create
 
     def destroy
       if @admin_set.destroy
-        redirect_to hyrax.admin_admin_sets_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
+        if request.referer.include? "my/collections"
```
jcoyne: Please move this into a method called `after_delete_success`. That way it's easy to override the single method if these paths need to be changed in a local application.

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
```diff
@@ -1,6 +1,7 @@
 RSpec.describe Hyrax::Admin::AdminSetsController do
   routes { Hyrax::Engine.routes }
   let(:user) { create(:user) }
+  let(:request) { double }
```
jcoyne: I don't think this is used, is it?

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
```diff
@@ -147,6 +148,12 @@
       let(:admin_set) { create(:admin_set, edit_users: [user]) }
 
       context "with empty admin set" do
+        before do
+          allow_any_instance_of(ActionController::TestRequest).to receive(:referer) do
```
jcoyne: using `allow_any_instance_of` is discouraged. Can you do `allow(controller).to receive(:request).and_return(double(referer: '/path'))` ?

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
```diff
@@ -156,6 +163,23 @@
         end
       end
 
+      context "with empty admin set and referrer from the my/collecton dashboard" do
+        before do
+          allow_any_instance_of(ActionController::TestRequest).to receive(:referer) do
```
jcoyne: and here.

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
blancoj: This one does not work. Have tried many variations, but I'm not having any luck.

jcoyne: Okay, lets try: `request.env['HTTP_REFERER'] = '/path'`

blancoj: I found a solution:
controller.request.set_header("HTTP_REFERER", "/my/collections")

app/controllers/concerns/hyrax/collections/accepts_batches.rb
```diff
@@ -36,7 +36,7 @@ def filter_docs_with_edit_access!
 
         def filter_docs_with_access!(access_type = :edit)
           no_permissions = []
-          if batch.empty?
+          if batch.empty? && flash[:notice].nil?
```
jcoyne: Why the check for nil here? When is the notice ever not nil?

blancoj: This method gets called when drawing the dashboard for displaying all collections or my collections.  If you delete an Admin Set, flash[:notice] will contain: "Administrative set successfully deleted" when the code gets here, and you want this message to show-up, not "Select something first".  So this check makes sure the flash notice message does not get overwritten.

jcoyne: This method (`filter_docs_with_access!`) is called in a before filter, so no flash could be set at this point, right?  If you revert this change is there a test that will fail that can illustrate why this conditional is necessary?

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -140,5 +140,15 @@ def admin_set_params
...
         blacklight_config.repository_class
       end
+
+      def after_delete_success
+        if request.referer.include? "my/collections"
+          redirect_to hyrax.my_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
+        elsif request.referer.include? "collections"
+          redirect_to hyrax.dashboard_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
+        else
+          redirect_to hyrax.admin_admin_sets_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
```
elrayle: Can you redirect this the else to `redirect_to hyrax.my_collections_path`?  This executes when deleting from admin set show page.  It should redirect to collections listing.

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
```diff
@@ -148,14 +148,28 @@
...
           expect(response).to redirect_to(admin_admin_sets_path)
           expect(flash[:notice]).to eq "Administrative set successfully deleted"
           expect(AdminSet.exists?(admin_set.id)).to be false
         end
       end
 
+      context "with empty admin set and referrer from the my/collections dashboard" do
+        it "deletes the admin set" do
+          controller.request.set_header("HTTP_REFERER", "/my/collections")
+          delete :destroy, params: { id: admin_set }
```
elrayle: Can you test /dashboard/collections as a referer too?

app/controllers/concerns/hyrax/collections/accepts_batches.rb
blancoj: Even though this is a before action, it is passed a notice in the redirect, and you want that notice to get out to the user:

`redirect_to hyrax.dashboard_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')`

I can't get a test to fail if I remove the checkout, but I do see this behavior I described above when I test it.

blancoj: I found a way not to have to do this check.  going to push a new pr for it now.

elrayle:
1 failing test that is a known randomly failing test.

PR #2150, opened 2017-11-10
Updating documentation to clarify

jeremyf:
In reviewing the documentation, I believe there may have been confusing
inline documentation that lead to a statement that may be incorrect.

> Calling `#activate!` on a Sipity::Workflow a second time will
> *deactivate* that workflow. If you want your workflow setup script to
> be idempotent, you should add a check to see if it is already
> activated.

### DIFF
--- a/app/models/sipity/workflow.rb
+++ b/app/models/sipity/workflow.rb
@@ -37,14 +37,16 @@ module Sipity
     # @api public
     #
     # Within the given permission_template scope:
-    #   * Deactivate the current active workflow (if one exists)
-    #   * Activate the specified workflow_id
+    #   * Activate the specified workflow_id or workflow_name
+    #   * Deactivate the other workflows
+    #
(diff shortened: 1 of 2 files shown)


jeremyf:
@bess I see that you submitted the following: https://github.com/samvera/samvera.github.io/commit/30a31995769199d6ce5b06107cbda7d09d68e447#diff-c813ad58bd3e03065da783330bfbe7c8R77

Do you recall if this was documentation based on experience or based on a reading of the `Sipity::Workflow.activate!` method? (Or something else)

jeremyf:
These updates are aspirationally true; However there is a bug identified in the documentation. https://github.com/samvera/samvera.github.io/commit/30a31995769199d6ce5b06107cbda7d09d68e447#diff-c813ad58bd3e03065da783330bfbe7c8R77

PR #2155, opened 2017-11-10
Redirect to correct page

### LINKED ISSUE #2153: Changing "Show" (per page) drop down sends user to different page.
When I'm on my dashboard collection show page (`/dashboard/collections/gt54kn002`):

And I change the value in the "Show" widget:

screen shot 2017-11-10 at 10 23 25 am

Then I am redirected to this page: `/collections/gt54kn002?utf8=✓&sort=system_create_dtsi+asc&per_page=100&locale=en`

There are 2 issues here:
1. Consistency with behavior of "sort by". It does not redirect when the value is changed. We should pick one behavior for both dropdowns an stick with it.
1. The redirect should be back to the collection show page on my dashboard, not the public collection show page.

lfarrell:
Fixes #2153

Redirects collections results per page to correct route.
Makes sort by/results per page JS onChange event consistent.
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/app.js.erb
+++ b/app/assets/javascripts/hyrax/app.js.erb
@@ -135,7 +135,7 @@ Hyrax = {
     // Per Page select will submit its form to change records shown
     perPage: function () {
         var PerPage = require('hyrax/per_page');
-        $('#per_page').each(function () {
+        $('#sort, #per_page').each(function () {
             new PerPage($(this));
         });
     },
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/hyrax/app.js.erb
```diff
@@ -135,7 +135,7 @@ Hyrax = {
     // Per Page select will submit its form to change records shown
     perPage: function () {
         var PerPage = require('hyrax/per_page');
-        $('#per_page').each(function () {
+        $('#sort, #per_page').each(function () {
```
jcoyne: This seems incorrect to initialize the PerPage with a `#sort` identifier.  Can we change the name of `PerPage` to `SortAndPerPage` or `SearchOptions`?

lfarrell:
Updated the file and method names.

lfarrell:
Fixed merge conflict

PR #2159, opened 2017-11-10
Remove autocreate_fields! it makes it impossible to set required fields

jcoyne:


spec/change_sets/hyrax/generic_work_change_set_spec.rb
```diff
@@ -360,20 +360,16 @@
   describe "#fields" do
     subject { change_set.fields.keys }
 
-    # rubocop:disable RSpec/ExampleLength
     it do
-      is_expected.to eq ["created_at", "updated_at", "depositor", "title",
-                         "date_uploaded", "date_modified", "admin_set_id",
-                         "state", "proxy_depositor", "on_behalf_of",
```
cjcolvar: Was "state" removed deliberately?

bmquinn: @cjcolvar My guess is that's coming from the AASM workflows in Figgy and we don't need it in Hyrax.

jcoyne: @cjcolvar yes, because the state was automatically copied to the form from the model, but we don't need it there as far as I can tell.

PR #2172, opened 2017-11-13
update valkyrie.yml to use fedora adapter for dev/prod

geekscruff:


jcoyne:
Do we need this?  What if I don't intend to put my resources in Postgresql?  What if I'm using mysql?

geekscruff:
@jcoyne just realised that, was about to post a comment ... currently there's an error if trying to create anything from console or admin_set create task ... I guess we need a way round that?

jcoyne:
@geekscruff but that's only if you have configured it to use the postgresql adapter, right?  I say let's not ship in that configuration.

geekscruff:
@jcoyne I was trying this inside .internal_test_app ... wanted to see how far I could get bringing up rails s, failed at create_admin_set task

jcoyne:
Yes, I'm suggesting that this line is wrong: https://github.com/samvera/hyrax/blob/valkyrie/lib/generators/hyrax/templates/config/valkyrie.yml#L2

geekscruff:
ah, bingo ... is that worth a PR?

PR #2175, opened 2017-11-14
[WIP] linked_data_resource_input

geekscruff:
[soon to be rebased on the new valkyrie]

The refactoring of controlled_vocabulary to linked_data_resources means that the form no longer receives a object as expected. Although the tests for controlled_vocabulary_input pass, in the application itself this would fail as the form is trying to call node? on a String. This PR fixes it by creating an object via the linked_data_resource_factory which will response to node? and rdf_subject and rdf_label. I've renamed it in line with the renaming to linked_data_resource in other bits of the stack. For this to work GeonamesResource needs to respond to node? and rdf_label.

There is an outstanding issue that the rdf_label (ie. the place name) is not retrieved, but I have a TODO elsewhere to fix that with a solr query. This is the same as current behaviour AFAIK.

### DIFF
--- a/app/inputs/linked_data_resource_input.rb
+++ b/app/inputs/linked_data_resource_input.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ControlledVocabularyInput < MultiValueInput
+class LinkedDataResourceInput < MultiValueInput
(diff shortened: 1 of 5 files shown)


geekscruff:
Replaced by a PR on valkyrie-2.

PR #2180, opened 2017-11-14
Use the provided helpers to avoid having to keep track of html_safe

jcoyne:


spec/views/hyrax/users/_user_info.html.erb_spec.rb
```diff
@@ -1,27 +1,25 @@
 RSpec.describe 'hyrax/users/_user_info.html.erb', type: :view do
-  let(:user) { stub_model(User, user_key: 'jdoe42') }
+  let(:user) { stub_model(User, user_key: 'jdoe42', orcid: '000-000', zotero_userid: 'jdoe42zotero') }
+  let(:arkvo_api) { true }
```
mjgiarlo: Not a big deal, but I suspect this should be `arkivo_api` throughout the spec.

spec/helpers/hyrax_helper_spec.rb
```diff
@@ -82,7 +82,6 @@ def new_state
           car_link   = search_catalog_path(f: { 'vehicle_type_sim' => ['car'] })
           truck_link = search_catalog_path(f: { 'vehicle_type_sim' => ['truck'] })
           expect(subject).to eq "<a href=\"#{car_link}\">car</a>, <a href=\"#{truck_link}\">truck</a>"
-          expect(subject).to be_a ActiveSupport::SafeBuffer
```
mjgiarlo: If we no longer expect this to return an `ActiveSupport::SafeBuffer`, should we also reflect this in the YARD doc for the `#link_to_facet_list` method above?

jcoyne: Yes, I wasn't sure what it should return. Ideas?

mjgiarlo: @jcoyne Is it just returning a string now?

jcoyne: Sometimes a string and sometimes a SafeBuffer.

mjgiarlo: Should we make the method return a SafeBuffer consistently?

jcoyne: No. It's perfectly reasonable to return a string that isn't safe... it will just be escaped...which is good because that prevents XHR attacks.

mjgiarlo: Works for me.

PR #2166, opened 2017-11-13
Generic_work_change_set_spec

geekscruff:
Fixes one failure where result of eq is not in same order. Order doesn't matter here so use include.

PR #2181, opened 2017-11-14
fix generic_work_changeset_spec failure by switching test from eq to …

jcoyne:
…(#2166)

include ... order doesn't matter; added 'import_url'

foo

This is a forward port of #2166


jcoyne:
Looks like we don't need this because this branch already has match_array.

PR #2182, opened 2017-11-14
Fixity check job

cjcolvar:
This PR assumes that the checksum and size fields are being populated by the characterization service in FileNodeBuilder but that isn't happening currently because there aren't any services configured or available. Future work here is writing a FileCharacterizationService for hydra-file_characterization or pulling in the tika implementation from valkyrie MVP. Additionally this change might want to be pulled out into its own service analogous to ActiveFedora::FixityService.

Replaces #2097.

### DIFF
--- a/app/jobs/fixity_check_job.rb
+++ b/app/jobs/fixity_check_job.rb
@@ -21,14 +21,14 @@ class FixityCheckJob < Hyrax::ApplicationJob
   # record recording the check.
   #
   # @param uri [String] uri - of the specific file/version to fixity check
-  # @param file_set_id [FileSet] the id for FileSet parent object of URI being checked.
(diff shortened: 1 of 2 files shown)


cjcolvar:
@jcoyne any ideas on how we should merge this and #2198?  It looks like there a few small things in this PR that should end up in yours.  I'm not sure how much time I'll have to work on this after this week.

jcoyne:
@cjcolvar I consider #2198 blocked until https://github.com/samvera-labs/valkyrie/pull/312 is resolved... so we need to start by getting valkyrie to support versions.  I'm fine merging this PR as soon as the `[WIP]` is removed.

cjcolvar:
@jcoyne Removed.

jcoyne:
@cjcolvar the commit message needs an update too.

cjcolvar:
@jcoyne Done.

PR #2183, opened 2017-11-14
Valkyrize some views

cjcolvar:
Use change sets, create_for_repository, and deal with Valkyrie::IDs

### DIFF
--- a/app/views/hyrax/base/file_manager.html.erb
+++ b/app/views/hyrax/base/file_manager.html.erb
@@ -1,7 +1,7 @@
 <h1><%= t("hyrax.file_manager.link_text") %></h1>
 <ul class="breadcrumb">
   <li>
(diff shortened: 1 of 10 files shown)


app/views/hyrax/base/file_manager.html.erb
```diff
@@ -1,7 +1,7 @@
 <h1><%= t("hyrax.file_manager.link_text") %></h1>
 <ul class="breadcrumb">
   <li>
-    Back to <%= link_to @change_set.to_s, [main_app, @change_set] %>
+    Back to <%= link_to @change_set.resource.to_s, [main_app, @change_set] %>
```
jcoyne: I wonder if this shouldn't be delegated in the change_set?

cjcolvar:
@jcoyne good idea, done.

PR #2185, opened 2017-11-14
[DON'T MERGE] Updated selector to match generated html

cjcolvar:
The generated html changed so I updated the test to reflect what was being generated.  I'm not sure if this is the right thing to do here or if we should coerce the html to match what was being generated before.


jcoyne:
This seems like the change made earlier may have affected the ability for the javascript to work. @geekscruff Can you review this?

geekscruff:
@jcoyne @cjcolvar will do, in meetings today so might be tomorrow

geekscruff:
@cjcolvar @jcoyne ... actually the original test is correct and was failing because of changes I made. PR  #2202 fixes that. Sorry. Once 2202 is merged, this can be closed.

PR #2191, opened 2017-11-15
String#coerce always returns a string

cjcolvar:
This fixes `spec/views/hyrax/admin/admin_sets/index.html.erb_spec.rb:30` without breaking the tests in `spec/models/solr_document_spec.rb`


jcoyne:
@cjcolvar can this be forward ported to master?

cjcolvar:
Sure I can do that tomorrow.

cjcolvar:
@jcoyne Can you rereview since I added a test?  Should I add any more tests?

PR #2192, opened 2017-11-15
String#coerce always returns a string

cjcolvar:
Also add Boolean#coerce to return a boolean since String was just passing it through before.  I believe this might break any customizations of SolrDocument that used `Solr::String` for boolean values unless they were changed to `Solr::Boolean`.

I made this change after finding that a `Time` object passed to `Date#coerce` threw an exception.  Now `Date#coerce` should be able to handle that and any other object that the `#to_s` of can be passed to `Date#parse`.

### DIFF
--- a/app/models/concerns/hyrax/solr_document/metadata.rb
+++ b/app/models/concerns/hyrax/solr_document/metadata.rb
@@ -25,7 +25,7 @@ module Hyrax
         class String
           # @return [String]
           def self.coerce(input)
-            ::Array.wrap(input).first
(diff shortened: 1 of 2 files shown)


cjcolvar:
@jcoyne Do you think I should also add tests showing that passing a `String` to `Boolean#coerce` now works and that all non-Strings passed to `String#coerce` are in fact coerced into Strings?

cjcolvar:
Closing due to risk and will include in valkyrie work (#2191).

PR #2197, opened 2017-11-15
Deprecate human_readable_type=

jcoyne:
This isn't translatable. Instead, set the type by updating the i18n like
this:

```yaml
en:
  activefedora:
    models:
      generic_work: Plain Ol' Worktype

```

### DIFF
--- a/app/models/concerns/hyrax/human_readable_type.rb
+++ b/app/models/concerns/hyrax/human_readable_type.rb
@@ -11,6 +11,9 @@ module Hyrax
       def human_readable_type=(val)
         @_human_readable_type = val
       end
+      deprecation_deprecate :human_readable_type= => 'human_readable_type is deprecated. ' \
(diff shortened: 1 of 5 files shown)


app/models/concerns/hyrax/human_readable_type.rb
```diff
@@ -11,6 +11,7 @@ def human_readable_type
       def human_readable_type=(val)
         @_human_readable_type = val
       end
+      deprecation_deprecate :human_readable_type= => 'human_readable_type is deprecated.  Set the i18n key for activefedora.models.#{model_name.i18n_key} instead. This will be removed in Hyrax 3'
```
no-reply: Could we wrap the message to shorten the line length?

jcoyne:
@no-reply fixed the line length, but that wiped out your approval.  Can you look again?

PR #2207, opened 2017-11-16
Use create_for_repository and traits when using collection factories

cjcolvar:


jcoyne:
@cjcolvar want to backport this to master?

cjcolvar:
I'm not sure if it is worth it until collections-sprint is merged.

PR #2218, opened 2017-11-17
redirect GET requests on work#new when the user is not logged in

geekscruff:
WorksControllerBehavior #build_change_set errors if there is no current_user (ie. if user isn't logged in). Reordering in ResourceController #new fixes it.

There are two unrelated failing tests in `spec/controllers/hyrax/generic_works_controller_spec.rb` which I had a look at but the fix isn't obvious (not sure if it's a test issue or a code issue)

### DIFF
--- a/spec/controllers/hyrax/generic_works_controller_spec.rb
+++ b/spec/controllers/hyrax/generic_works_controller_spec.rb
@@ -6,474 +6,484 @@ RSpec.describe Hyrax::GenericWorksController do
   let(:hyrax) { Hyrax::Engine.routes.url_helpers }
   let(:user) { create(:user) }
 
-  before { sign_in user }
+  describe 'with a logged in user' do
+    before { sign_in user }
 
-  describe 'integration test for suppressed documents' do
-    let(:work) do
-      create_for_repository(:work, :public, state: Vocab::FedoraResourceStatus.inactive)
-    end
-
-    before do
-      create(:sipity_entity, proxy_for_global_id: work.to_global_id.to_s)
-    end
-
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -86,6 +86,7 @@ def inspect_work
 
       # Overridden to provide depositor (used on app/views/hyrax/base/_form_share.html.erb)
       def build_change_set(resource)
+        raise Hydra::AccessDenied unless current_user
```
jcoyne: This method is about building a change set, not about checking the users access, so is there a different place we could put this?

jcoyne: Would flipping lines 15 & 16 do the trick? https://github.com/samvera/hyrax/blob/204e6b6d529daaff9d6eb25c596dfade91d87a5d/app/controllers/concerns/hyrax/resource_controller.rb#L16

geekscruff: Yes, that does it. Thanks!

spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -476,4 +476,14 @@
       expect(assigns(:change_set)).not_to be_blank
     end
   end
+
+  describe 'logged out user' do
+    before { sign_out user }
```
jcoyne: Can you remove this line?  By default it shouldn't be signed in, right?

PR #2223, opened 2017-11-17
Add Lease as a Valkyrie::Resource

jcoyne:


app/actors/hyrax/actors/interpret_visibility_actor.rb
```diff
@@ -60,14 +60,14 @@ def sanitize_unrestricted_params
                                :visibility_after_embargo)
           end
 
-          def sanitize_embargo_params
+          def sanitize_lease_params
```
mjgiarlo: Buh. Why'd you swap the method names?

jcoyne: Look at the implementation.  It was doing the lease stuff.

PR #2225, opened 2017-11-20
Merge latest master into collections-sprint

elrayle:
Merge latest master into collections-sprint


jcoyne:
There appear to be some legitimate test failures here.

elrayle:
@jcoyne Not that I can see.  Which one is of concern to you?

For 5.1.4 build...
* Fails with a typical feature test failure -- rspec ./spec/features/batch_edit_spec.rb:15 # batch editing changes the value of each field for all selected works
```
Failure/Error: find('.status', text: 'Changes Saved')
     Capybara::ElementNotFound:
       Unable to find visible css ".status" with text "Changes Saved" within #<Capybara::Node::Element tag="form" path="/html/body/div[2]/div/div[2]/div/div[1]/div[9]/form">
```
* all others are feature tests failing on Save/other action with `undefined method 'size' for nil:NilClass`

For 5.0.6 build...
* has a single failure which is a known randomly failing test

mjgiarlo:
I'm guessing a dependency was upgraded.

PR #2229, opened 2017-11-20
Remove position: fixed from workflow-affix as it breaks scrolling in …

aaron-collier:
A small update to remove `position: fixed` from the workflow-affix class as it breaks the scrolling on the reformatted work-show page. Also add a missing space in the html.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_positioning.scss
+++ b/app/assets/stylesheets/hyrax/_positioning.scss
@@ -17,7 +17,6 @@
 }
 
 .workflow-affix {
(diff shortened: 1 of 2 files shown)


jcoyne:
I think we should remove the whole workflow-affix class, because the workflow section is no longer affixed to the page in anyway.

aaron-collier:
@jcoyne, I will be. It’ll be a separate PR today in line with the proposed page format from Gary.

PR #2246, opened 2017-11-21
[WIP] Lazy load default collection type gid if gid is missing or blank

### LINKED ISSUE #2241: BUG Collections: unable to read in old collections when gid is missing
### Descriptive summary

Collections created in earlier versions do not have a gid and should default to the User Collection type's gid.

### Rationale

Allow older collections to be accessible without requiring migration to add gid.  It should lazy load the gid for older collections.

### Expected behavior

When collection has gid missing or blank in the solr doc, the collection should be assigned the User Collection type's gid.

### Actual behavior

Get's an error on Dashboard -> Collections and any other access of older collections complaining that the gid is missing or blank.

1. Create a collection without a gid in the metadata or solrdoc
1. Dashboard -> Collections

Should display the older collections without gid
Confirm in solr that the collection does NOT have a gid assigned

1. Dashboard -> Collections -> in action menu for older collection -> Edit collection

Should go to the edit form for the older collection
Without saving, confirm in solr that the collection now does have a gid assigned

### Related work

Lazy migration was removed in PR #1618

elrayle:
Fixes #2241

Hyrax::CollectionType.find_by_gid! and find_by_gid now default to return User Collection gid if the gid is missing or blank.  This can be prevented from happening by passing the second parameter as false.

The collection model has an after_find callback that will update the gid for a collection that is missing the gid and save it.  This is effectively the lazy migration of the collection to include a gid and only happens once per older collection.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -25,6 +25,8 @@ module Hyrax
         index.as(*index_collection_type_gid_as)
       end
 
+      after_find { |col| load_collection_type_instance(col) } # lazy migrate older collections to have collection type gid
+
       # validates that collection_type_gid is present
       validates :collection_type_gid, presence: true
       after_initialize :ensure_collection_type_gid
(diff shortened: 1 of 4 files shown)


app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -25,6 +25,8 @@ module CollectionBehavior
         index.as(*index_collection_type_gid_as)
       end
 
+      after_find { |col| load_collection_type_instance(col) } # lazy migrate older collections to have collection type gid
```
jcoyne: Can we avoid callbacks?  Callbacks like this break the single responsibility principal and  aren't going to work with Valkyrie (so we'd just remove it?).

elrayle:
Closed with preference for PR #2251 which doesn't use callbacks.

PR #2251, opened 2017-11-21
Default the collection type when collection_type_gid_ssim is missing from the Solr document

cjcolvar:
…issing from the collection solr document


elrayle:
There is one failure that is a known random failing test.

@mjgiarlo I reviewed Chris' changes and approved.  Needs another review since I contributed a test that would have caught the failure to view the collection's index page when there are legacy collections.

PR #2252, opened 2017-11-21
Reformat workflow review panel

aaron-collier:
Related to: https://github.com/samvera-labs/hyku/issues/1033

This:
- replaces the workflow-affixed class with a panel-workflow class that includes the proposed styling layout
- cleans up bootstrap related formatting
- Makes the review and approve panel more prominent and obvious

@samvera/hyrax-code-reviewers


app/assets/stylesheets/hyrax/_positioning.scss
```diff
@@ -16,12 +16,13 @@
...
-  z-index: 2;
+  border-color: #e0e0e0;
+  bottom: 0;
   left: 0;
-  right: 0;
-  max-width: 1140px;
   margin: 0 auto;
+  max-width: 1140px;
+  right: 0;
+  z-index: 2;
```
jcoyne: Do we still need `bottom`, `z-index` and `right` since it's no longer fixed?

app/assets/stylesheets/hyrax/_work-show.scss
```diff
@@ -56,3 +56,24 @@ ul.tabular {
   margin-bottom: 18px;
   margin-top: -8px;
 }
+
+.panel-workflow {
```
jcoyne: why are we defining this class in two places?

app/assets/stylesheets/hyrax/_positioning.scss
aaron-collier: Probably not. I will remove and test.

app/assets/stylesheets/hyrax/_work-show.scss
aaron-collier: I was following the perceived convention that part of this was including position (hence in the _position file). But I'm happy to do it only in one place.

jcoyne: I think it's all a mismash, but please let's not define something twice.

aaron-collier: For sure. Fixed.

app/assets/stylesheets/hyrax/_work-show.scss
```diff
@@ -56,3 +56,28 @@ ul.tabular {
...
+  background: #FFF;
+  border-color: #e0e0e0;
+  left: 0;
```
jcoyne: Do we need this?

app/assets/stylesheets/hyrax/_work-show.scss
```diff
@@ -56,3 +56,28 @@ ul.tabular {
...
+  margin: 0 auto;
+  max-width: 1140px;
```
jcoyne: Why are we setting all these directives and not just using `class="panel-default"` or `@extend .panel-default`?

aaron-collier: I don't want to extend `.panel-default` as it would throw of the use of it for the show page itself. And I would want to avoid overriding any of the .panel-* configs to match this mockup in case they are used later with unexpected results.

app/assets/stylesheets/hyrax/_work-show.scss
aaron-collier: I think so, but can test again. I believe I had some issues without it.

app/assets/stylesheets/hyrax/_work-show.scss
```diff
@@ -56,3 +56,28 @@ ul.tabular {
   margin-bottom: 18px;
   margin-top: -8px;
 }
+
+.panel-workflow {
+  background: #fff;
+  border-color: #e0e0e0;
+  left: 0;
```
jcoyne: Why is this needed?  Is there some reason we aren't using or inheriting from the bootstrap `panel-default` class?

aaron-collier: `panel-default` is used by the main panel, so I think overriding it would cause a problem. I think overriding any of the bootstrap panels provided is likely to cause confusion down the line if they are used on that page (or the style is picked up accidentally) and this makes the panel styling explicit for our use case. I'm open to a suggestion on a best practice that shows this approach is problematic though :-)

jcoyne: Okay, I'm fine with the class name. I didn't notice that this is still a `.panel`. Is the `left: 0`, `margin: 0 auto`, `max-width: 1140px` still required? Those seem like properties you'd set on fixed position elements, but not on inline elements.

 How about the background-color? Aren't the panel-heading and panel-body setting a background color that will override this one?

app/views/hyrax/base/_workflow_actions.html.erb
```diff
@@ -1,11 +1,11 @@
-<div id="workflow_controls" class="panel panel-default workflow-affix">
```
jcoyne: workflow-affix is bound to code in `app/assets/javascripts/hyrax/workflow_actions_affix.js`.  can that script be removed now?

aaron-collier: Removed.

Issue #1719, opened 2017-09-25
Collections: Populate edit Relationships tab with list of parent collections and sub-collections


elrayle:
### Descriptive summary

Get the list of parent collections and sub-collections for a collection from the nesting API.  Populate these lists in the `Relationships` tab of the collection edit form.

### Implementation

Update `app/views/hyrax/dashboard/collections/_form_relationships.html.erb` partial to... 
- [ ] use the nesting api to get the list of sub-collections
- [ ] use the nesting api to get the list of parent collections. 
- [ ] use the nesting api to add a sub-collection.
- [ ] use the nesting api to add a parent collection. 

### Steps to test the behavior

1. Create a collection type that allows nesting via Dashboard -> Settings -> Collection Types
1. Create 3 collections of that type via Dashboard -> Collections -> New Collection -> select that new type -> add title and save.  Assign titles 'A', 'B', 'C', 'D', 'E'
1. From Dashboard -> Collections, click the action menu for collection 'A'
  1. select `Add collection to another collection` and
  1. select collection 'B'
1. From Dashboard -> Collections, click the action menu for collection 'B'
  1. select `Add collection to another collection` and
  1. select collection 'C'
1. From Dashboard -> Collections, click the action menu for collection 'B' and 
  1. select 'Edit collection'
  1. click 'Relationships' tab

- [ ] Verify that the parent list includes collection 'A'.
- [ ] Verify that the child list includes collection 'C'.
- [ ] Verify that the select list for parent collections is populated with all possible parent collections = all collections of this type - current subcollections - current parents... ['D', 'E'].
- [ ] Verify that the select list for subcollections is populated with all possible subcollections = all collections of this type - current subcollections - current parents... ['D', 'E'].

1. Select 'D' as parent collection and click Add

- [ ] Verify the parent is added to the list of parents... ['A', 'D']

1. Select 'E' as subcollection and click Add

- [ ] Verify the subcollection is added to the list of subcollections... ['C', 'E']

### Related work

#1377 Add relationships tab to collection edit form
PR #1702 Update UI for relationship tab

laritakr:
@elrayle I have some concerns about this design. Something just feels wrong about it. We are mixing the editing of an object with the creation and removal of its relationships to other objects. 

[...]

Issue #2221, opened 2017-11-17
Hyrax::Collections::NestedCollectionQueryService logic not adequate


laritakr:
### Descriptive summary

( Collections Sprint)

The list of available child collections and available parent collections should not include any collections currently in ANY relationship (either parent or child) with the collection being modified.

The nesting query service needs to take into account the full ancestral path. It should match the logic used in the [nesting indexer](https://github.com/samvera-labs/samvera-nesting_indexer).

### Rationale

Adding these as a parent or child should cause a PCDM error, and does cause circular indexing errors. We need to keep them from the list of options to select.

Nesting decisions should be made in `/app/services/hyrax/collections/nested_collection_query_service.rb`. Additional validations (if specific to one particular case), could occur in the form `/app/forms/hyrax/forms/dashboard/nest_collection_form.rb`, but in general, we want the query service provide a fully valid list of collections for nesting.

### Details

Given a collection, available parent collections should include only:
- collections of the same collection type
- collections which the user has deposit permissions to
- collections which are not already anywhere in the parent or child path

Given a collection, available subcollections should include only:
- collections of the same collection type
- collections which the user has read permission to
- collections which are not already anywhere in the parent or child path

Solr fields for nesting collections are:
- "member_of_collection_ids_ssim" - immediate parents
- "nesting_collection__ancestors_ssim" - all partial and full paths to current collection excluding the current collection
- "nesting_collection__parent_ids_ssim" - immediate parents
- "nesting_collection__pathnames_ssim" - all ancestor paths with current collection the end point of the path

### Related Issues
Pull Request: Populate edit Relationships tab #1719
Manage Collection Relationships Tab #2257
Refactor the nested collection modal queries to use the form #2294
Wire up nested collection delete buttons #2295

elrayle:
A slight adjustment to the available children:
* include collections that are not already anywhere in the parent path
* include collections that are not immediate children

This means that assuming existing paths A-B-D and A-C, it is ok to create A-D and A-C-D.
```
    A
   /  \
  B    C
  |
  D
```
can become...
```
      A
   /  |  \
  B   C   D
  |   |
  D   D
```

elrayle:
* Ancestors: parse all collection ids in current collection's `nesting_collection__pathnames_ssim` removing dups and removing current collection's id

* All Descendants of qv33rw64w:  matches for solr query `nesting_collection__pathnames_ssim:*qv33rw64w/*`

[...]

PR #2257, opened 2017-11-22
Manage Collection Relationships Tab

laritakr:
Completes issue #1719, except for todo items listed below which are moving to new issues. 

Make edit#relationships tab functional.
  - add list of parent collections
  - add list of child collections
  - chg button to add an existing collection as a parent via modal
  - remove list of parents for selection (done via modal)
  - chg button to add existing collection as a subcollection via modal
  - remove list of child collections for selection (done via modal)
  - add button functionality to create new collection as subcollection

Also add above nesting buttons to dashboard/show view.

Make all nesting ability tests consistent: _The user should have deposit access to the parent and read access to the child._

Reverse the nesting direction of the collection persistence service.

Remove the list of works in the collection, as we do not want to manage work relationships within the edit collection form.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/nest_collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/nest_collections_controller.rb
@@ -2,39 +2,42 @@ module Hyrax
   module Dashboard
     class NestCollectionsController < ApplicationController
       include Blacklight::Base
-      class_attribute :form_class
+      class_attribute :form_class, :new_collection_form_class
       self.form_class = Hyrax::Forms::Dashboard::NestCollectionForm
+      self.new_collection_form_class = Hyrax::Forms::CollectionForm
 
-      def new_within
-        @form = build_within_form
-      end
-
-      def create_within
+      # Add this collection as a subcollection within another existing collection
+      def create_relationship_within
         @form = build_within_form
         if @form.save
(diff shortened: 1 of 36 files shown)


app/controllers/hyrax/dashboard/nest_collections_controller.rb
```diff
@@ -2,39 +2,42 @@ module Hyrax
...
-        unless params[:parent_id] == "none"
-          parent_coll = Collection.find params[:parent_id]
-          child_coll = Collection.find params[:child_id]
-
-          child_coll.member_of_collections << parent_coll
-          child_coll.save!
-
-          notice = "Collections were mapped!"
+      # create and link new subcollection under this collection, (this collection as parent)
+      def new_under
```
jeremyf: I'm confused what is happening here. My expectation with a method prefix "new_" is that the action would not be redirecting but instead rendering HTML or JSON. The "create_" action would handle redirecting on success or rendering the "new_" action's template on failure. The request cycle would look as follows:

http request mermaid

laritakr:
Remaining todo list:
- pass in appropriate “hasAccess” values rather than always “true”
- verify that the appropriate ability tests are being done in all locations for nesting
- clean up the modal queries to rely on form
- delete buttons options - run through nest_collections_controller using routes
- fix the service to take full ancestry into account in the list of collections it presents for possible nesting (https://github.com/samvera/hyrax/issues/2221)

elrayle:
Nice work!

PR #2266, opened 2017-11-27
seed generated objects use same process UI does to create objects

elrayle:
* updated all objects used for QA testing
* updated all objects used with nested collection ad hoc testing
* DID NOT update objects used for work embargo ad hoc testing

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/templates/db/seeds.rb
+++ b/lib/generators/hyrax/templates/db/seeds.rb
@@ -5,93 +5,144 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
-
+#
 # To use this file, run the following command in the .internal_test_app:
 #   rails generate hyrax:sample_data
+#
+# To re-use this file, you will likely want to clean out the test app content
+#   rails console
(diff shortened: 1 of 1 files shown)


lib/generators/hyrax/templates/db/seeds.rb
```diff
@@ -9,89 +9,118 @@
...
+  attrs = options
+  attrs[:title] = [title]
+  attrs[:description] = [description]
+  col.attributes = attrs
+  col.apply_depositor_metadata(user.user_key)
+  col.collection_type_gid = type_gid
+  col.visibility = visibility
+  col.save
+  Hyrax::Collections::PermissionsCreateService.create_default(collection: col, creating_user: user)
+  col.update_index
```
jcoyne: Is there a reason this step isn’t a part of PermissionsCreateService? You always need to run it afterwards, right?

elrayle:
Fixed rubocop errors.  Removed the index update.  It is not needed.

jcoyne:
In lieu of this change, can we entertain the idea of removing seeds.rb entirely?   I believe the purpose of seeds.rb is to seed initial data that is required for an application (e.g. a list of statuses)... it should not be used for test data, which is how it is used here.

PR #2283, opened 2017-11-29
Update Child works UI

### LINKED ISSUE #2247: Work form, relationships tab, child works section
1. Update the child work section to enable adding work to a work via a select menu and button. Text label to the left of the select menu is “Add to work”. Button is to the right of the select menu and has label “Add”.
1. Add a table below the child work section showing what works the work is currently part of (so the user doesn't try to add the work to a collection it is already part of). If this is a new work or an existing work that hasn't been added to any works, omit the "This work is currently in these works" heading and the table beneath it.

This was intended to mirror #2245 

relationships-tab-changes

Ref #2034

### LINKED ISSUE #2248: Text change on Work form, relationships tab, add new child work
Change "Attach new work" to "Deposit new work as child of this work"

relationships-tab-changes

Ref #2034

jcoyne:
This makes the child works UI behave like the collection members

Fixes #2247 
Fixes #2248


app/views/hyrax/base/_form_child_work_relationships.html.erb
```diff
@@ -10,14 +10,29 @@ HTML Properties:
...
+                        autocomplete: 'off',
+                        data: {
+                          autocomplete: 'work',
+                          'autocomplete-url' => Rails.application.routes.url_helpers.qa_path + '/search/find_works',
+                          'exclude-work': f.object.model.id # exclude this item from the result set.
+                        } %>
+      <a class="btn btn-primary" data-behavior="add-relationship">Add</a>
+  </div>
+  <div class="message has-warning hidden"></div>
+  <div style="margin-top: 10px">
```
aaron-collier: Are we ok with inline styling here, or should this be classed?

aaron-collier:
This only includes translations in the `hyrax.en.yml`, will those be pulled by default when another language is selected?

mjgiarlo:
@aaron-collier We update non-English translations as part of the release process, so yep.

app/views/hyrax/base/_form_child_work_relationships.html.erb
jcoyne: I don't know that we want another class just to bump a button down 10px.  Do you have a preference?

PR #2292, opened 2017-12-01
Generate a test that actually creates a work

jcoyne:
Previously it was just testing that the "new work" form would draw


lib/generators/hyrax/work/templates/feature_spec.rb.erb
```diff
@@ -36,6 +40,30 @@ RSpec.feature 'Create a <%= class_name %>', js: false do
...
+      fill_in('Title', with: 'My Test Work')
+      fill_in('Creator', with: 'Doe, Jane')
+      fill_in('Keyword', with: 'testing')
+      select('In Copyright', from: 'Rights statement')
+
+      # With selenium and the chrome driver, focus remains on the
+      # select box. Click outside the box so the next line can't find
+      # its element
+      find('body').click
+      choose('<%= file_name %>_visibility_open')
```
jeremyf: Are you sure you want ERB here?

jcoyne: Yes.. This way the template is generated as `choose('generic_work_visibilty_open')` (just as in the `create_work_spec.rb` that is in Hyrax.

PR #2297, opened 2017-12-03
Ensure that empty strings are not stored for single-valued attributes

### LINKED ISSUE #2296: Single-valued attributes are stored as empty strings after form submission
### Descriptive summary

When a work has single-valued attributes and the form is submitted, empty strings are stored. 

### Rationale

Having these values change after form submission means that views and other code that exports them has to check for `nil` and an empty string. It would cleaner for the values to remain nil unless they aren't an empty string. 

### Expected behavior

The value of the single-valued attribute is `nil` until it becomes a non-empty string. 

### Actual behavior

The value of blank single-valued attribute become "" after submitted the work form. 

1. Create a work that has a single-valued attribute
2. Submit a form for the work
3. Check the attribute: 

```ruby
work.single_value
""
```
or view the show page: 

screen shot 2017-12-02 at 2 13 44 pm

### Related work

https://github.com/curationexperts/epigaea/issues/331
https://github.com/curationexperts/epigaea/issues/618

little9:
Fixes #2296 

The commit adds methods to `Hyrax::Actors::BaseActor` to ensure that empty strings are not stored for single-valued attributes.

This fixes an issue that occurs if you create a work with single
valued attributes. When the form is submitted without values the
values in the record become and empty string: `""`. These values should
remain `nil`.

* add methods to `BaseActor` to normalize single-valued attributes

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -98,6 +98,24 @@ module Hyrax
         def multivalued_form_attributes(attributes)
           attributes.select { |_, v| v.respond_to?(:select) && !v.respond_to?(:read) }
         end
+
+        # If any single-valued attributes are empty strings make them nil
+        # e.g.:
+        #   self.attributes = { 'single_valued' => '' }
+        #   remove_empty_string_attributes!
+        #   self.attributes
+        # => { 'single_valued' => nil }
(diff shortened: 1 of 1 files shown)


little9:
I think was happening for me because I was using a custom module to normalize params in the controller. I fixed it over at: https://github.com/curationexperts/epigaea/pull/639

PR #2310, opened 2017-12-05
Update null source_type on permission templates

### LINKED ISSUE #1580: Collections: Migrate existing permission templates to use new source_type column

jgondron:
Fixes #1580

This fixes the missed data migration in [20170821152307_add_source_type_to_permission_templates](https://github.com/samvera/hyrax/blob/collections-sprint/db/migrate/20170821152307_add_source_type_to_permission_templates.rb). At this point, time has passed and users have likely run migrations on their dbs, so had to add this as a new migration and assume that any nils are due to this missed data migration.

### DIFF
--- a/db/migrate/20171205205934_update_null_source_type_on_permission_templates.rb
+++ b/db/migrate/20171205205934_update_null_source_type_on_permission_templates.rb
@@ -1,4 +1,4 @@
-class UpdateNullSourceTypeOnPermissionTemplates < ActiveRecord::Migration[5.1]
+class UpdateNullSourceTypeOnPermissionTemplates < ActiveRecord::Migration[5.0]
   def up
     Hyrax::PermissionTemplate.find_each do |permission_template|
       permission_template.source_type = 'admin_set' if permission_template.source_type.nil?
--- a/db/migrate/20171205205934_update_null_source_type_on_permission_templates.rb
+++ b/db/migrate/20171205205934_update_null_source_type_on_permission_templates.rb
@@ -1,0 +1,8 @@
+class UpdateNullSourceTypeOnPermissionTemplates < ActiveRecord::Migration[5.1]
+  def up
+    Hyrax::PermissionTemplate.find_each do |permission_template|
+      permission_template.source_type = 'admin_set' if permission_template.source_type.nil?
+      permission_template.save!
+    end
+  end
+end


db/migrate/20171205205934_update_null_source_type_on_permission_templates.rb
```diff
@@ -0,0 +1,8 @@
+class UpdateNullSourceTypeOnPermissionTemplates < ActiveRecord::Migration[5.1]
+  def up
+    Hyrax::PermissionTemplate.find_each do |permission_template|
```
jcoyne: Since all collections are in the repository, is there some reason we need to record their type? Can't we just look up the PermissionTemplate by ID?

jgondron: I'll have to defer to the original motivation for splitting this into two fields: #1576. This PR is just to fix a data migration gap that arose from that separation, which is captured in issue #1580.

jcoyne: Yeah, I think we ought to back that work out. I don’t want to make the migration harder than it needs to be. Should we wait for Lynette to return so we can all discuss this?

mjgiarlo: @jcoyne Yes, we ought to wait on @elrayle.

elrayle: @hortongn We discussed this earlier in the process. I asked the question about whether we really needed the source type and as I recall, you indicated that it was needed. Can you comment on this question?

hortongn: @jcoyne @elrayle One reason we need `source_type` is because we're using a single PermissionTemplateAccessesController to handle templates for both collections and admin sets.  When that controller processes a destroy, it needs to know if the template was used for a collection or an admin set so it can redirect to the proper place.

See https://github.com/samvera/hyrax/blob/collections-sprint/app/controllers/hyrax/admin/permission_template_accesses_controller.rb#L42-L50

elrayle: When we discussed this, the use of a single approach to permissions for admin sets and collections was to set the stage for admin sets to become just another type of collection.  The hope being that there will be one set of code to maintain for groups of things instead of two similar concepts with different implementations.  It may make some things a bit awkward in the short run, but only in the sense that there are some if-else type statements in the permission template controllers.  This was discussed at the in-person development week.

db/migrate/20171205205934_update_null_source_type_on_permission_templates.rb
```diff
@@ -0,0 +1,8 @@
+class UpdateNullSourceTypeOnPermissionTemplates < ActiveRecord::Migration[5.0]
+  def up
+    Hyrax::PermissionTemplate.find_each do |permission_template|
+      permission_template.source_type = 'admin_set' if permission_template.source_type.nil?
```
mjgiarlo: Will `#find_each` iterate over every `Hyrax::PermissionTemplate` row in the database (in batches, sure)? If so, can this be simplified to:

```ruby
Hyrax::PermissionTemplate.where(source_type: nil) do |permission_template|
  permission_template.update(source_type: 'admin_set')
end
```

:question:

jgondron: This would be an improvement, and I can do so if required. My preference would be to just merge as is for the sake of time, but if you think this is important enough I can find time to change later this week.

jcoyne:
source_type was removed, so can this be closed?

PR #2320, opened 2017-12-06
Inline collection helpers that are used only once

jcoyne:
This helps to reduce indirection in the code making it easier to read.

### DIFF
--- a/app/helpers/hyrax/collections_helper.rb
+++ b/app/helpers/hyrax/collections_helper.rb
@@ -22,24 +22,8 @@ module Hyrax
       params[:cq].present?
     end
 
-    def button_for_remove_from_collection(collection, document, label = 'Remove From Collection')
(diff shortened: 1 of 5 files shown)


elrayle:
@mjgiarlo @jcoyne This is an example of the type of change that makes merging of master and collections-sprint time consuming.  Although the change is simple, I have to do a search of the code to determine if the change conflicts with other changes made in the collections extensions work.  There are currently 11 file conflicts that I have to review individually.  Then there will be more to be resolved after running tests and seeing how similar changes are required in new files.

elrayle:
@jcoyne The button_for_remove_from_collection is used in two places in the collections-sprint branch.  I am putting this code back into the collections-sprint branch.

jcoyne:
@elrayle I'd prefer we just inline the code.  The button_for_remove_from_collection has always been a lot of inderection that makes it harder to follow the control flow.

PR #2188, opened 2017-11-14
Hyrax::TikaFileCharacterizationService for file characterization

bmquinn:
- Adds `Hyrax::TikaFileCharacterizationService`
- Adds `Hyrax::MultiChecksum` model, which stores md5, sha1 and sha256 checksums for uploaded files
- Adds `file_nodes` attribute to `Hyrax::FileSetBehavior`
- Update template generator to add `Hyrax::TikaFileCharacterizationService` to `Valkyrie::FileCharacterizationService.services` in `valkyrie.rb`
- Adds `Hyrax::FileNode` and `Hyrax::Multichecksum` specs

PR #2337, opened 2017-12-08
Hyrax::TikaFileCharacterizationService for file characterization

kdid:
( Previous PR here: https://github.com/samvera/hyrax/pull/2188 )

- Adds `Hyrax::TikaFileCharacterizationService`
- Adds `Hyrax::MultiChecksum` model, which stores md5, sha1 and sha256 checksums for uploaded files
- Adds `file_nodes` attribute to `Hyrax::FileSetBehavior`
- Update template generator to add `Hyrax::TikaFileCharacterizationService` to `Valkyrie::FileCharacterizationService.services` in `valkyrie.rb`
- Adds `Hyrax::FileNode` and `Hyrax::Multichecksum` specs
- Calls `Hyrax::TikaFileCharacterizationService` from the `CharacterizeJob`
- Update specs for `CharacterizeJob`

@samvera/hyrax-code-reviewers


app/services/hyrax/tika_file_characterization_service.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    # characterizes the file_node passed into this service
+    # Default options are:
+    #   save: true
+    # @param save [Boolean] should the persister save the file_node after Characterization
+    # @return [FileNode]
+    # @example characterize a file and persist the changes by default
+    #   Valkyrie::FileCharacterizationService.for(file_node, persister).characterize
+    # @example characterize a file and do not persist the changes
+    #   Valkyrie::FileCharacterizationService.for(file_node, persister).characterize(save: false)
```
jcoyne: I don't see where you accepting this `save` argument. Should we remove this documentation?

app/services/hyrax/tika_file_characterization_service.rb
```diff
@@ -0,0 +1,71 @@
...
+    def characterize
+      result = JSON.parse(json_output).last
+      @file_characterization_attributes = {
+        width: result['tiff:ImageWidth'],
```
jcoyne: Does this work on all types of images or just TIFFs?

app/services/hyrax/tika_file_characterization_service.rb
```diff
@@ -0,0 +1,71 @@
...
+      }
+      new_file_node = original_file.new(@file_characterization_attributes.to_h)
+      @file_set.member_ids = @file_set.member_ids.reject { |x| x == new_file_node.id } + [new_file_node.id]
+      @persister.save(resource: new_file_node)
+      @persister.save(resource: @file_set)
+      @file_set
+    end
+
+    def json_output
+      "[#{RubyTikaApp.new(filename.to_s).to_json.gsub('}{', '},{')}]"
```
jcoyne: Can you leave a comment explaining why you're doing this transformation?

spec/jobs/characterize_job_spec.rb
```diff
@@ -1,66 +1,33 @@
...
-  # let(:io)          { JobIoWrapper.new(file_set_id: file_set.id, user: create(:user), path: filename) }
-  let(:file) do
-    Hydra::PCDM::File.new.tap do |f|
-      f.content = 'foo'
-      f.original_name = 'picture.png'
-      f.save!
-      allow(f).to receive(:save!)
-    end
-  end
+require 'rails_helper'
```
jcoyne: I'm pretty sure we don't need this.

spec/models/hyrax/file_node_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
jcoyne: This isn't necessary.

spec/services/hyrax/tika_file_characterization_service_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
jcoyne: You can omit this.

spec/services/hyrax/tika_file_characterization_service_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+require 'valkyrie/specs/shared_specs'
```
jcoyne: Why are we pulling in shared specs here?

spec/support/stub_tika.rb
```diff
@@ -0,0 +1,9 @@
...
...s Per Strip":"13 rows/strip","Samples Per Pixel":"3 samples/pixel","Software":"Adobe Photoshop CS5.1 Macintosh","Strip Byte Counts":"7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 7800 600 bytes","Strip Offsets":"[23 longs]","Unknown tag (0x0129)":"0 1","Unknown tag (0x02bc)":"[14622 shorts]","Unknown tag (0x8649)":"[8822 shorts]","White Point":"1343036288/4294967295 1413044224/4294967295","X Resolution":"1120 dots per inch","X-Parsed-By":["org.apache.tika.parser.DefaultParser","org.apache.tika.parser.image.TiffParser"],"Y Resolution":"1120 dots per inch","date":"2014-12-03T12:40:50","dcterms:created":"2014-12-03T12:40:50","dcterms:modified":"2014-12-03T12:40:50","meta:creation-date":"2014-12-03T12:40:50","meta:save-date":"2014-12-03T12:40:50","modified":"2014-12-03T12:40:50","resourceName":"example.tif20170713-88893-1cirjkm","tiff:BitsPerSample":"8","tiff:ImageLength":"287","tiff:ImageWidth":"200","tiff:Make":"Phase One","tiff:Model":"P65+","tiff:Orientation":"1","tiff:ResolutionUnit":"Inch","tiff:SamplesPerPixel":"3","tiff:Software":"Adobe Photoshop CS5.1 Macintosh","tiff:XResolution":"1120.0","tiff:YResolution":"1120.0"}'
```
jcoyne: If you were to use:
```
output =<<~EOF
              Json stuff here
              and here.
EOF
```
then you could avoid the rubocop exception.

.rubocop.yml
```diff
@@ -29,6 +29,8 @@ Lint/AmbiguousBlockAssociation:
 
 Metrics/LineLength:
   Max: 200
+  Exclude:
+    - 'spec/support/stub_tika.rb'
```
jcoyne: 😢

spec/jobs/characterize_job_spec.rb
```diff
@@ -1,66 +1,33 @@
...
-    let(:filename) { File.join(fixture_path, 'world.png') }
+RSpec.describe CharacterizeJob do
+  include ActionDispatch::TestProcess
 
-    it 'skips Hyrax::WorkingDirectory' do
-      expect(Hyrax::WorkingDirectory).not_to receive(:find_or_retrieve)
-      expect(Hydra::Works::CharacterizationService).to receive(:run).with(file, filename)
-      described_class.perform_now(file_set, file.id, filename)
+  describe ".perform" do
+    let(:char) { instance_double("Valkyrie::FileCharacterizationService") }
```
jcoyne: No quotes needed here.

spec/support/stub_tika.rb
kdid: Then I'm seeing this: 
```
spec/support/stub_tika.rb:5:13: E: unexpected token tLSHFT
(Using Ruby 2.1 parser; configure using TargetRubyVersion parameter, under AllCops)
    output =<<~EOF
            ^^
```

app/services/hyrax/tika_file_characterization_service.rb
kdid: Yes, works for other image types

spec/support/stub_tika.rb
kdid: Nevermind! Got it.

app/models/concerns/hyrax/file_set_behavior.rb
```diff
@@ -21,16 +21,17 @@ module FileSetBehavior
...
 
-      attribute :file_identifiers, Valkyrie::Types::Set
       attribute :member_ids, Valkyrie::Types::Array
       attribute :embargo_id, Valkyrie::Types::ID.optional
       attribute :lease_id, Valkyrie::Types::ID.optional
+
+      delegate :width, :height, :mime_type, :size, to: :original_file, allow_nil: true
     end
 
     def original_file
```
jcoyne: Can you add documentation about what you expect this method to return?

app/services/hyrax/tika_file_characterization_service.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    # characterizes the file_node (FileSet) passed into this service
+    # @return [FileSet]
+    # @example characterize a file and persist the changes
+    #   Valkyrie::FileCharacterizationService.for(file_node, persister).characterize
+    def characterize
+      result = JSON.parse(json_output)
+      @file_characterization_attributes = {
+        width: result['tiff:ImageWidth'],
```
jcoyne: is this only going to work for TIFF sources?  What about JP2 or Jpeg?

app/services/hyrax/tika_file_characterization_service.rb
```diff
@@ -0,0 +1,66 @@
...
+    # @return Valkyrie::StorageAdapter::File
+    def file_object
+      @file_object ||= Valkyrie::StorageAdapter.find_by(id: original_file.file_identifiers[0])
+    end
+
+    def original_file
+      @file_set.original_file
+    end
+
+    def valid?
```
jcoyne: What is the purpose of this method?

spec/jobs/characterize_job_spec.rb
```diff
@@ -1,66 +1,31 @@
...
-    let(:collection) { build(:collection) }
-
-    before do
-      allow(file_set).to receive(:parent).and_return(work)
-      allow(work).to receive(:in_collections).and_return([collection])
-    end
-    it "reindexes the collection" do
-      expect(collection).to receive(:update_index)
-      described_class.perform_now(file_set, file.id)
+    xit "invokes a CreateDerivativesJob" do
```
jcoyne: Why are we skipping this test?  Should we just remove it?

app/services/hyrax/tika_file_characterization_service.rb
kdid: Yes, works for other image types as well.

spec/jobs/characterize_job_spec.rb
kdid: I have the call to CreateDerivativesJob commented out in the CharacterizeJob because it does not seem to be working at this point, and it causes the tests not to run at all since it gets called from the factory when the fileSet is created. I can certainly comment this out, just was thinking maybe it was easier to remember to uncomment it when the derivatives job is updated and working. It seems like something we should be testing.

PR #2342, opened 2017-12-08
Move methods from the work form to the WorkChangeSet

jcoyne:


app/change_sets/hyrax/work_change_set.rb
```diff
@@ -123,8 +129,27 @@ def work_members_json
...
 
+    def permissions_changed?
+      changed.include?('edit_users') ||
+        changed.include?('edit_groups') ||
+        changed.include?('read_users') ||
+        changed.include?('read_groups')
+    end
+
+    def visibility_changed?
+      changed['visibility']
```
mjgiarlo: Are we cool with truthiness here, or should we make it explicit with `changed.key?('visibility')`?

app/change_sets/hyrax/work_change_set.rb
```diff
@@ -123,8 +129,27 @@ def work_members_json
       end.to_json
     end
 
+    def permissions_changed?
+      changed.include?('edit_users') ||
```
mjgiarlo: I don't feel strongly about this, but what would you feel about using `#any?` here rather than checking one-by-one? 

```ruby
permissions_keys.any? { |key| changed.key?(key) }

private

  def permissions_keys
    ['edit_users', 'edit_groups', 'read_users', 'read_groups']
  end
```

OTOH, the way it's currently written is easy to understand, and I'm not sure there's any performance benefit to using the iterator method here since the method as currently written should return true as soon as the first expression returns true, right?

jcoyne: Yeah, I think it's easier to read as written.

PR #2353, opened 2017-12-12
Test for presence of logger

jcoyne:


jcoyne:
@mjgiarlo because we have no way of preventing the user from doing `Rails.logger = nil`.  Rails also checks for the presence of the logger on each log line. See: https://github.com/rails/rails/blob/b5db73076914e7103466bd76bec785cbcfe88875/actionpack/lib/action_controller/metal/implicit_render.rb#L57

PR #2355, opened 2017-12-12
Add sequence rendering and metadata into the IIIF manifest

geekscruff:
This PR pulls in Hyku PRs [1376 ](https://github.com/samvera-labs/hyku/pull/1376) and [1436 ](https://github.com/samvera-labs/hyku/pull/1436).

Description of Main Changes

**Metadata:**

- Adds a metadata block to the IIIF manifest, which in UV will display in the more info panel.
- Adds a configuration for the fields to include in this section.
- Uses the Hyrax::Form::WorkForm.required_fields as the default

**Rendering**

- Adds a sequences#rendering block to the IIIF manifest - items in this block are displayed as a download in UV and are used for a download which represents the whole item, rather than the single page being displayed (eg. a PDF of a book)
- Adds rendering_ids into the form so that one or more 'renderings' can be selected

See [hydrox 88](https://github.com/sul-dlss/hydrox/issues/88)

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -144,15 +144,21 @@ module Hyrax
           !((file_set_params || {}).keys.map(&:to_s) & %w[visibility embargo_release_date lease_expiration_date]).empty?
         end
 
+        # Must clear the fileset from the thumbnail_id, representative_id and rendering_ids fields on the work
+        #   and force it to be re-solrized.
+        # Although ActiveFedora clears the children nodes it leaves those fields in Solr populated.
+        # rubocop:disable Metrics/AbcSize
+        # rubocop:disable Metrics/CyclomaticComplexity
         def unlink_from_work
           work = file_set.parent
-          return unless work && (work.thumbnail_id == file_set.id || work.representative_id == file_set.id)
-          # Must clear the thumbnail_id and representative_id fields on the work and force it to be re-solrized.
-          # Although ActiveFedora clears the children nodes it leaves those fields in Solr populated.
+          return unless work && (work.thumbnail_id == file_set.id || work.representative_id == file_set.id || work.rendering_ids.include?(file_set.id))
           work.thumbnail = nil if work.thumbnail_id == file_set.id
           work.representative = nil if work.representative_id == file_set.id
+          work.rendering_ids -= [file_set.id]
           work.save!
         end
+      # rubocop:enable Metrics/AbcSize
+      # rubocop:enable Metrics/CyclomaticComplexity
     end
   end
 end
--- a/app/models/concerns/hyrax/file_set/belongs_to_works.rb
+++ b/app/models/concerns/hyrax/file_set/belongs_to_works.rb
@@ -3,10 +3,6 @@ module Hyrax
     module BelongsToWorks
       extend ActiveSupport::Concern
 
-      included do
-        before_destroy :remove_representative_relationship
-      end
-
       def parents
         in_works
       end
@@ -33,17 +29,6 @@ module Hyrax
           end
         end
       end
-
-      # If any parent objects are pointing at this object as their
-      # representative, remove that pointer.
-      def remove_representative_relationship
-        parent_objects = parents
-        return if parent_objects.empty?
-        parent_objects.each do |work|
-          work.update(representative_id: nil) if work.representative_id == id
-        end
-      end
     end
   end
 end
-\
--- a/app/models/concerns/hyrax/has_rendering.rb
+++ b/app/models/concerns/hyrax/has_rendering.rb
@@ -1,0 +1,11 @@
+module Hyrax::HasRendering
+  extend ActiveSupport::Concern
+
+  included do
+    # rubocop:disable Rails/HasAndBelongsToMany
+    has_and_belongs_to_many :renderings,
+                            predicate: Hyrax.config.rendering_predicate,
+                            class_name: 'ActiveFedora::Base'
+    # rubocop:enable Rails/HasAndBelongsToMany
+  end
+end
--- a/app/models/concerns/hyrax/solr_document/metadata.rb
+++ b/app/models/concerns/hyrax/solr_document/metadata.rb
@@ -71,6 +71,8 @@ module Hyrax
         attribute :workflow_state, Solr::String, solr_name('workflow_state_name', :symbol)
         attribute :human_readable_type, Solr::String, solr_name('human_readable_type', :stored_searchable)
         attribute :representative_id, Solr::String, solr_name('hasRelatedMediaFragment', :symbol)
+        # extract the term name from the rendering_predicate (it might be after the final / or #)
+        attribute :rendering_ids, Solr::Array, solr_name(Hyrax.config.rendering_predicate.value.split(/#|\/|,/).last, :symbol)
         attribute :thumbnail_id, Solr::String, solr_name('hasRelatedImage', :symbol)
         attribute :thumbnail_path, Solr::String, CatalogController.blacklight_config.index.thumbnail_field
         attribute :label, Solr::String, solr_name('label')
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -2,6 +2,7 @@ module Hyrax
   class WorkShowPresenter
     include ModelProxy
     include PresentsAttributes
+
     attr_accessor :solr_document, :current_ability, :request
 
     class_attribute :collection_presenter_class
@@ -38,7 +39,7 @@ module Hyrax
     delegate :title, :date_created, :description,
              :creator, :contributor, :subject, :publisher, :language, :embargo_release_date,
              :lease_expiration_date, :license, :source, :rights_statement, :thumbnail_id, :representative_id,
-             :member_of_collection_ids, to: :solr_document
+             :rendering_ids, :member_of_collection_ids, to: :solr_document
 
     def workflow
       @workflow ||= WorkflowPresenter.new(solr_document, current_ability)
@@ -52,10 +53,6 @@ module Hyrax
     def download_url
       return '' if representative_presenter.nil?
       Hyrax::Engine.routes.url_helpers.download_url(representative_presenter, host: request.host)
-    end
-
-    def manifest_url
-      manifest_helper.polymorphic_url([:manifest, self])
     end
 
     # @return [Boolean] render the UniversalViewer
@@ -156,17 +153,50 @@ module Hyrax
 
     delegate :member_presenters, :file_set_presenters, :work_presenters, to: :member_presenter_factory
 
+    def manifest_url
+      manifest_helper.polymorphic_url([:manifest, self])
+    end
+
+    # IIIF rendering linking property for inclusion in the manifest
+    #  Called by the `iiif_manifest` gem to add a 'rendering' (eg. a link a download for the resource)
+    #
+    # @return [Array] array of rendering hashes
+    def sequence_rendering
+      renderings = []
+      if solr_document.rendering_ids.present?
+        solr_document.rendering_ids.each do |file_set_id|
+          renderings << manifest_helper.build_rendering(file_set_id)
+        end
+      end
+      renderings.flatten
+    end
+
+    # IIIF metadata for inclusion in the manifest
+    #  Called by the `iiif_manifest` gem to add metadata
+    #
+    # @return [Array] array of metadata hashes
+    def manifest_metadata
+      metadata = []
+      Hyrax.config.iiif_metadata_fields.each do |field|
+        metadata << {
+          'label' => I18n.t("simple_form.labels.defaults.#{field}"),
+          'value' => Array.wrap(send(field))
+        }
+      end
+      metadata
+    end
+
     private
+
+      def manifest_helper
+        @manifest_helper ||= ManifestHelper.new(request.base_url)
+      end
 
       def featured?
         if @featured.nil?
           @featured = FeaturedWork.where(work_id: solr_document.id).exists?
         end
         @featured
-      end
-
-      def manifest_helper
-        @manifest_helper ||= ManifestHelper.new(request.base_url)
       end
 
       def user_can_feature_works?
--- a/app/views/hyrax/base/_form_media.html.erb
+++ b/app/views/hyrax/base/_form_media.html.erb
@@ -1,4 +1,5 @@
 <% if f.object.persisted? && f.object.member_ids.present? %>
   <%= render 'form_representative', f: f %>
   <%= render 'form_thumbnail', f: f %>
+  <%= render 'form_rendering', f: f %>
 <% end %>
--- a/app/views/hyrax/base/_form_rendering.html.erb
+++ b/app/views/hyrax/base/_form_rendering.html.erb
@@ -1,0 +1,14 @@
+<div class="row">
+  <div class="col-md-12">
+    <fieldset id="representative-media">
+      <legend>
+        <%= t("hyrax.base.form_rendering.legend_html") %>
+      </legend>
+      <p><%= t("hyrax.base.form_rendering.help_html") %></p>
+      <%= f.select :rendering_ids,
+                   f.object.select_files,
+                   { include_blank: true },
+                   { class: 'form-control', multiple: true } %>
+    </fieldset>
+  </div>
+</div>
--- a/lib/generators/hyrax/templates/config/initializers/hyrax.rb
+++ b/lib/generators/hyrax/templates/config/initializers/hyrax.rb
@@ -16,6 +16,11 @@ Hyrax.config do |config|
   # avoid clashes if you plan to use the default (dct:isPartOf) for other relations.
   # config.admin_set_predicate = ::RDF::DC.isPartOf
 
+  # Which RDF term should be used to relate objects to a rendering?
+  # If this is a new repository, you may want to set a custom predicate term here to
+  # avoid clashes if you plan to use the default (dct:hasFormat) for other relations.
+  # config.rendering_predicate = ::RDF::DC.hasFormat
+
   # Email recipient of messages sent via the contact form
   # config.contact_email = "repo-admin@example.org"
 
@@ -142,6 +147,9 @@ Hyrax.config do |config|
 
   # Returns a IIIF image size default
   # config.iiif_image_size_default = '600,'
+
+  # Fields to display in the IIIF metadata section; default is the required fields
+  # config.iiif_metadata_fields = Hyrax::Forms::WorkForm.required_fields
 
   # Should a button with "Share my work" show on the front page to all users (even those not logged in)?
   # config.display_share_button_when_not_logged_in = true
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -321,6 +321,13 @@ module Hyrax
       @admin_set_predicate ||= ::RDF::Vocab::DC.isPartOf
     end
 
+    # Set predicate for rendering to dc:hasFormat as defined in
+    #   IIIF Presentation API context:  http://iiif.io/api/presentation/2/context.json
+    attr_writer :rendering_predicate
+    def rendering_predicate
+      @rendering_predicate ||= ::RDF::Vocab::DC.hasFormat
+    end
+
     attr_writer :work_requires_files
     def work_requires_files?
       return true if @work_requires_files.nil?
@@ -401,6 +408,14 @@ module Hyrax
       @iiif_image_size_default ||= '600,'
     end
     attr_writer :iiif_image_size_default
+
+    # IIIF metadata - fields to display in the metadata section
+    #
+    # @return [#Array] fields
+    def iiif_metadata_fields
+      @iiif_metadata_fields ||= Hyrax::Forms::WorkForm.required_fields
+    end
+    attr_writer :iiif_metadata_fields
 
     # Should a button with "Share my work" show on the front page to users who are not logged in?
     attr_writer :display_share_button_when_not_logged_in
--- a/spec/presenters/hyrax/work_show_presenter_spec.rb
+++ b/spec/presenters/hyrax/work_show_presenter_spec.rb
@@ -379,4 +379,40 @@ RSpec.describe Hyrax::WorkShowPresenter do
       end
     end
   end
+
+  describe "#manifest" do
+    let(:work) { create(:work_with_one_file) }
+    let(:solr_document) { SolrDocument.new(work.to_solr) }
+
+    describe "#sequence_rendering" do
+      subject do
+        presenter.sequence_rendering
+      end
+
+      before do
+        Hydra::Works::AddFileToFileSet.call(work.file_sets.first,
+                                            File.open(fixture_path + '/world.png'), :original_file)
+      end
+
+      it "returns a hash containing the rendering information" do
+        work.rendering_ids = [work.file_sets.first.id]
+        expect(subject).to be_an Array
+      end
+    end
+
+    describe "#manifest_metadata" do
+      subject do
+        presenter.manifest_metadata
+      end
+
+      before do
+        work.title = ['Test title', 'Another test title']
+      end
+
+      it "returns an array of metadata values" do
+        expect(subject[0]['label']).to eq('Title')
+        expect(subject[0]['value']).to include('Test title', 'Another test title')
+      end
+    end
+  end
 end
--- a/app/builders/hyrax/manifest_helper.rb
+++ b/app/builders/hyrax/manifest_helper.rb
@@ -11,5 +11,26 @@ module Hyrax
       opts[:host] ||= @hostname
       super(record, opts)
     end
+
+    # Build a rendering hash
+    #
+    # @return [Hash] rendering
+    def build_rendering(file_set_id)
+      file_set_document = query_for_rendering(file_set_id)
+      label = file_set_document.label.present? ? ": #{file_set_document.label}" : ''
+      mime = file_set_document.mime_type.present? ? file_set_document.mime_type : I18n.t("hyrax.manifest.unknown_mime_text")
+      {
+        '@id' => Hyrax::Engine.routes.url_helpers.download_url(file_set_document.id, host: @hostname),
+        'format' => mime,
+        'label' => I18n.t("hyrax.manifest.download_text") + label
+      }
+    end
+
+    # Query for the properties to create a rendering
+    #
+    # @return [SolrDocument] query result
+    def query_for_rendering(file_set_id)
+      ::SolrDocument.find(file_set_id)
+    end
   end
 end
--- a/app/forms/hyrax/forms/work_form.rb
+++ b/app/forms/hyrax/forms/work_form.rb
@@ -24,7 +24,7 @@ module Hyrax
       self.terms = [:title, :creator, :contributor, :description,
                     :keyword, :license, :rights_statement, :publisher, :date_created,
                     :subject, :language, :identifier, :based_near, :related_url,
-                    :representative_id, :thumbnail_id, :files,
+                    :representative_id, :thumbnail_id, :rendering_ids, :files,
                     :visibility_during_embargo, :embargo_release_date, :visibility_after_embargo,
                     :visibility_during_lease, :lease_expiration_date, :visibility_after_lease,
                     :visibility, :ordered_member_ids, :source, :in_works_ids,
@@ -117,7 +117,7 @@ module Hyrax
           [:files, :visibility_during_embargo, :embargo_release_date,
            :visibility_after_embargo, :visibility_during_lease,
            :lease_expiration_date, :visibility_after_lease, :visibility,
-           :thumbnail_id, :representative_id, :ordered_member_ids,
+           :thumbnail_id, :representative_id, :rendering_ids, :ordered_member_ids,
            :member_of_collection_ids, :in_works_ids, :admin_set_id]
       end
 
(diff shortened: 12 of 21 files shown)


app/models/concerns/hyrax/file_set/belongs_to_works.rb
```diff
@@ -4,7 +4,7 @@ module BelongsToWorks
       extend ActiveSupport::Concern
 
       included do
-        before_destroy :remove_representative_relationship
+        before_destroy :remove_representative_relationship, :remove_rendering_relationship
```
mjgiarlo: I wonder if the three changes you made to this module ought to move to the new `Hyrax::HasRendering` module so that all the rendering stuff is together.

app/models/concerns/hyrax/has_rendering.rb
```diff
@@ -0,0 +1,12 @@
+module Hyrax::HasRendering
+  extend ActiveSupport::Concern
+
+  included do
+    # rubocop:disable Rails/HasAndBelongsToMany
+    # dc:hasFormat defined in Presentation API context:  http://iiif.io/api/presentation/2/context.json
+    has_and_belongs_to_many :renderings,
+                            predicate: ::RDF::Vocab::DC.hasFormat,
```
mjgiarlo: Will using this predicate be problematic? @no-reply (Tagging you since you started the work to make the Admin Set relationship predicate configurable.)

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -38,7 +39,7 @@ def page_title
     delegate :title, :date_created, :description,
              :creator, :contributor, :subject, :publisher, :language, :embargo_release_date,
              :lease_expiration_date, :license, :source, :rights_statement, :thumbnail_id, :representative_id,
-             :member_of_collection_ids, to: :solr_document
+             :rendering_ids, :member_of_collection_ids, to: :solr_document
```
mjgiarlo: Should this move to `Hyrax::ManifestComponents` where it is used?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    # IIIF metadata for inclusion in the manifest
+    #
+    # @return [Array] array of metadata hashes
+    def manifest_metadata
+      metadata = []
+      metadata_fields.each do |field|
+        metadata << {
+          'label' => field.to_s.humanize.capitalize,
```
mjgiarlo: Do we need string keys here or can we use Ruby 1.9-style hash syntax?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+          'value' => get_metadata_value(field)
```
mjgiarlo: Can we rename this and the method below to `wrapped_metadata_value(s)`?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      def manifest_helper
+        @manifest_helper ||= ManifestHelper.new(request.base_url)
+      end
+
+      # Build a rendering hash
+      #
+      # @return [Hash] rendering
+      def build_rendering(file_set_id)
+        query_for_rendering(file_set_id).map do |x|
```
mjgiarlo: Would you consider renaming the block variable to be more descriptive? Something like `file_set_document`?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+          label = x['label_ssi'] ? ": #{x.fetch('label_ssi')}" : ''
+          {
+            '@id' => Hyrax::Engine.routes.url_helpers.download_url(x.fetch(ActiveFedora.id_field), host: request.host),
+            'format' => x.fetch('mime_type_ssi') ? x.fetch('mime_type_ssi') : 'unknown mime type',
```
mjgiarlo: Do we need string keys for format and label in this hash or can we use 1.9-style hash syntax?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      # Build a rendering hash
+      #
+      # @return [Hash] rendering
+      def build_rendering(file_set_id)
+        query_for_rendering(file_set_id).map do |x|
+          label = x['label_ssi'] ? ": #{x.fetch('label_ssi')}" : ''
+          {
+            '@id' => Hyrax::Engine.routes.url_helpers.download_url(x.fetch(ActiveFedora.id_field), host: request.host),
```
mjgiarlo: Are you using `#fetch` with a single arg here so that the line raises an exception if the fetch fails? Just checking!

app/presenters/hyrax/manifest_components.rb
mjgiarlo: Use `#fetch` here instead of the ternary:

```ruby
x.fetch('mime_type_ssi', 'unknown mime type')
```

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+      def manifest_helper
+        @manifest_helper ||= ManifestHelper.new(request.base_url)
+      end
+
+      # Build a rendering hash
+      #
+      # @return [Hash] rendering
+      def build_rendering(file_set_id)
+        query_for_rendering(file_set_id).map do |x|
+          label = x['label_ssi'] ? ": #{x.fetch('label_ssi')}" : ''
```
mjgiarlo: I don't know that `#fetch` is useful here since you already known the key exists.

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+        end
+      end
+
+      # Query for the properties to create a rendering
+      #
+      # @return [SolrResult] query result
+      def query_for_rendering(file_set_id)
+        ActiveFedora::SolrService.query("id:#{file_set_id}",
+                                        fl: [ActiveFedora.id_field, 'label_ssi', 'mime_type_ssi'],
+                                        rows: 1)
```
mjgiarlo: If this is always returns 1 row, why are we `map`ing across the results above?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,77 @@
...
+      #
+      # @return [Array] fields
+      def metadata_fields
+        Hyrax.config.iiif_metadata_default
+      end
+
+      # Get the metadata value(s).
+      #
+      # @return [Array] field value(s)
+      def get_metadata_value(field)
```
mjgiarlo: Pls rename to `wrapped_metadata_values` to better reflect whats happening in the method.

app/views/hyrax/base/_form_media.html.erb
```diff
@@ -1,4 +1,8 @@
 <% if f.object.persisted? && f.object.member_ids.present? %>
   <%= render 'form_representative', f: f %>
   <%= render 'form_thumbnail', f: f %>
+  <%# display the rendering dropdown only if the representitive responds to image? (ie. Universal Viewer is used) %>
```
mjgiarlo: `s/representitive/representative/`

app/views/hyrax/base/_form_media.html.erb
```diff
@@ -1,4 +1,8 @@
...
+  <% if !curation_concern.representative.blank? && curation_concern.representative.image? %>
```
mjgiarlo: Prefer `#present?` over the negation of `#blank?`

app/views/hyrax/base/_form_rendering.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="row">
+  <div class="col-md-12">
+    <fieldset id="representative-media">
+      <legend>
+        Rendering
+      </legend>
+      <p>Select file(s) to be offered as as downloads for every image in Universal Viewer, for example a PDF of the whole <%= f.object.human_readable_type %>.</p>
```
mjgiarlo: `s/as as/as/`

lib/generators/hyrax/templates/config/initializers/hyrax.rb
```diff
@@ -143,6 +143,9 @@
   # Returns a IIIF image size default
   # config.iiif_image_size_default = '600,'
 
+  # Fields to display in the IIIF metadata section; default is the required fields
+  # config.iiif_metadata_default = Hyrax::Forms::WorkForm.required_fields
```
mjgiarlo: Would you mind renaming this to `iiif_metadata_fields`?

lib/hyrax/configuration.rb
```diff
@@ -402,6 +402,14 @@ def iiif_image_size_default
     end
     attr_writer :iiif_image_size_default
 
+    # IIIF metadata - fields to display in the metadata section
+    #
+    # @return [#Array] fields
+    def iiif_metadata_default
```
mjgiarlo: Would you mind renaming this to iiif_metadata_fields?

app/presenters/hyrax/manifest_components.rb
jcoyne: I don't think there's any reason to prefer `x['label_ssi']` over `x.fetch('label_ssi')` is there?

app/models/concerns/hyrax/file_set/belongs_to_works.rb
jcoyne: Can we please not add callbacks?  Can you put this in the actor stack instead?  This will make it much easer to integrate with Valkyrie.

geekscruff: I've moved this into the vile_set_actor, and also removed remove_representative_relationship as this is also in the (unlink_works) file_set_actor

app/models/concerns/hyrax/has_rendering.rb
geekscruff: I guess it could be, but this is the *proper* predicate according to IIIF. We could make it configurable ... or leave as-is until Hyrax 3 when changing predicates will get easier.

app/presenters/hyrax/manifest_components.rb
geekscruff: I *think* iiif_manifest wants them this way ... that could be changed, but the @id syntax would be more problematic as you can't give that as a symbol AFAIK ... I'm minded to leave them like this, unless you feel strongly about this and I can do more digging

app/presenters/hyrax/manifest_components.rb
geekscruff: done :)

app/presenters/hyrax/manifest_components.rb
geekscruff: I'm not quite understanding?

app/presenters/hyrax/manifest_components.rb
geekscruff: Done :)

app/presenters/hyrax/manifest_components.rb
geekscruff: I've removed that fetch ... I don't think that should fail?

app/presenters/hyrax/manifest_components.rb
geekscruff: Done the fetch change; see above comment re syntax

app/presenters/hyrax/manifest_components.rb
geekscruff: Done :)

app/presenters/hyrax/work_show_presenter.rb
geekscruff: Yes, done.

app/views/hyrax/base/_form_media.html.erb
geekscruff: Done

app/views/hyrax/base/_form_media.html.erb
geekscruff: Done :)

app/views/hyrax/base/_form_rendering.html.erb
geekscruff: Done

lib/generators/hyrax/templates/config/initializers/hyrax.rb
geekscruff: Done

lib/hyrax/configuration.rb
geekscruff: Done

app/models/concerns/hyrax/has_rendering.rb
mjgiarlo: :ok: I'd like to hear from @no-reply on this too.

app/presenters/hyrax/manifest_components.rb
mjgiarlo: K, sounds fine to me.

app/presenters/hyrax/manifest_components.rb
mjgiarlo: LGTM

spec/presenters/hyrax/work_show_presenter_spec.rb
```diff
@@ -327,4 +327,40 @@
...
+                                            File.open(fixture_path + '/world.png'), :original_file)
+      end
+
+      it "returns a hash containing the rendering information" do
+        work.rendering_ids = [work.file_sets.first.id]
+        expect(subject).to be_an Array
+      end
+    end
+
+    describe "#manifest metadata" do
```
mjgiarlo: change `#manifest metadata` to `#manifest_metadata`, pls!

app/models/concerns/hyrax/has_rendering.rb
no-reply: Configurable would be great. It doesn't look like there's any particular reason a schema for a Hyrax Work would need to align with the IIIF schema, so I'm not sure we care what they say is the proper predicate.

It does suggest the kinds of conflicts--with a specialized use of a common DC predicate--that we saw with the partitive predicates. I would vote for:

  - Follow IIIF by default and use `hasFormat`, but make it configurable; or
  - Use something internal (`something:iiifRenderings`) and serialize to IIIF manifiests as appropriate (and maybe make it configurable either now later?).

I don't have strong opinions about which.

mjgiarlo: @geekscruff I'm fine with either of these approaches. Clear?

app/presenters/hyrax/manifest_components.rb
geekscruff: Yep, sounds good to me. I'll try to get this wrapped up tomorrow.

app/models/concerns/hyrax/has_rendering.rb
geekscruff: Yep, sounds good to me. I'll try to get this wrapped up tomorrow.

app/models/concerns/hyrax/file_set/belongs_to_works.rb
```diff
@@ -34,16 +30,10 @@ def related_files
...
-      # If any parent objects are pointing at this object as their
-      # representative, remove that pointer.
-      def remove_representative_relationship
-        parent_objects = parents
-        return if parent_objects.empty?
-        parent_objects.each do |work|
-          work.update(representative_id: nil) if work.representative_id == id
-        end
+      # Required by FileSetIndexer
+      def rendering_ids
```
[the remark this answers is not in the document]
geekscruff: The method is used by the file_set_indexer to assign itself to the value of hasFormat_ssim. I was unsure about this when I originally did the Hyku PR, but followed what was happening for representative and thumbnail. Looking back at it now, this seems unnecessary and I can't think of any reason for the FileSet itself to have a 'rendering'. So I've removed it and run the tests, and tested in the test app and everything is working as expected. We just don't now have the FileSet being assigned itself as hasFormat_ssim.

app/views/hyrax/base/_form_rendering.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="row">
+  <div class="col-md-12">
+    <fieldset id="representative-media">
+      <legend>
+        Rendering
```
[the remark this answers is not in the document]
geekscruff: Done this, and the help text. Also done the same for representative and thumbnail.

app/views/hyrax/base/_form_rendering.html.erb
```diff
@@ -0,0 +1,14 @@
...
+      </legend>
+      <p>Select file(s) to be offered as a download for every image in Universal Viewer, for example a PDF of the whole <%= f.object.human_readable_type %>.</p>
```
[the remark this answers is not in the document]
geekscruff: Done

spec/presenters/hyrax/work_show_presenter_spec.rb
```diff
@@ -327,4 +327,40 @@
       end
     end
   end
+
+  describe "#manifest" do
+    let(:work) { create(:work_with_one_file) }
+    let(:solr_document) { SolrDocument.new(work.to_solr) }
+
+    describe "#sequence rendering" do
```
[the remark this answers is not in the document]
geekscruff: done

spec/presenters/hyrax/work_show_presenter_spec.rb
geekscruff: done

app/models/concerns/hyrax/has_rendering.rb
geekscruff: I've gone with option 1.

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
...
+    def manifest_url
+      manifest_helper.polymorphic_url([:manifest, self])
+    end
+
+    # IIIF rendering linking property for inclusion in the manifest
+    #
+    # @return [Array] array of rendering hashes
+    def sequence_rendering
+      renderings = []
+      if solr_document.rendering_ids.present?
```
jcoyne: rendering_ids is not ordered, is this method expecting that it is?

app/presenters/hyrax/manifest_components.rb
jcoyne: Can we use i18n for the field labels?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+
+      # Build a rendering hash
+      #
+      # @return [Hash] rendering
+      def build_rendering(file_set_id)
+        file_set_document = query_for_rendering(file_set_id).first # only one result is returned
+        label = file_set_document['label_ssi'] ? ": #{file_set_document['label_ssi']}" : ''
+        {
+          '@id' => Hyrax::Engine.routes.url_helpers.download_url(file_set_document[ActiveFedora.id_field], host: request.host),
```
jcoyne: can we just use `file_set_id` here?

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
...
+          'format' => file_set_document.fetch('mime_type_ssi', 'unknown mime type'),
+          'label' => 'Download whole resource' + label
```
jcoyne: i18n please.

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
+module Hyrax
+  module ManifestComponents
```
jcoyne: Rather than add these methods to the presenter via a module, could we extract a Manifest class that the presenter has an instance of?

app/presenters/hyrax/manifest_components.rb
geekscruff: No, whilst rendering goes inside the 'sequence' block in the manifest, the rendering has no expectation about order.

app/views/hyrax/base/_form_media.html.erb
```diff
@@ -1,4 +1,8 @@
 <% if f.object.persisted? && f.object.member_ids.present? %>
   <%= render 'form_representative', f: f %>
   <%= render 'form_thumbnail', f: f %>
+  <%# display the rendering dropdown only if the rendering responds to image? (ie. Universal Viewer is used) %>
+  <% if curation_concern.representative.present? && curation_concern.representative.image? %>
```
jcoyne: Why do we have this conditional?  Shouldn't we be able to select the representative and the rendering all in one step?  Also I'm not sure that responding to `image?` implies that UV is being used (which is how I read this comment)

app/presenters/hyrax/manifest_components.rb
geekscruff: `iiif_manifest` calls various methods directory on the object it receives, which is the presenter hence the methods being added to presenter?

app/models/concerns/hyrax/file_set/belongs_to_works.rb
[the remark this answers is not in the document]
mjgiarlo: OK! Thanks, @geekscruff

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      # Build a rendering hash
+      #
+      # @return [Hash] rendering
+      def build_rendering(file_set_id)
+        file_set_document = query_for_rendering(file_set_id).first # only one result is returned
+        label = file_set_document['label_ssi'] ? ": #{file_set_document['label_ssi']}" : ''
+        {
+          '@id' => Hyrax::Engine.routes.url_helpers.download_url(file_set_document[ActiveFedora.id_field], host: request.host),
+          'format' => file_set_document.fetch('mime_type_ssi', 'unknown mime type'),
```
mjgiarlo: `'unknown mime type'` should probably be i18nized as well.

app/presenters/hyrax/manifest_components.rb
geekscruff: I'm not following?

app/views/hyrax/base/_form_media.html.erb
geekscruff: Assigning rendering if there isn't an image assigned to representative will have no effect. It won't even appear in the manifest as there's no sequences block to add it to. That's why I originally did it conditionally, to avoid confusion of having a pointless option. But the downside is that it can take two edits to 1) assign representative, 2) assign rendering. I've removed the conditional.

mjgiarlo:
@jcoyne care to re-review?

app/presenters/hyrax/manifest_components.rb
jcoyne: That's fine. Can we pass a reference to the presenter?

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -100,7 +100,8 @@ def update_metadata(attributes)
       end
 
       def destroy
-        unlink_from_work
+        unlink_from_work(file_set.parent)
+        unlink_rendering(file_set.parent)
```
jcoyne: It looks like this combination of calls is going to save `file_set.parent` twice.  Can we make it only save once?

app/models/concerns/hyrax/file_set/belongs_to_works.rb
```diff
@@ -3,10 +3,6 @@ module FileSet
     module BelongsToWorks
       extend ActiveSupport::Concern
 
-      included do
```
jcoyne: 👏

app/models/concerns/hyrax/solr_document/metadata.rb
```diff
@@ -71,6 +71,7 @@ def self.coerce(input)
         attribute :workflow_state, Solr::String, solr_name('workflow_state_name', :symbol)
         attribute :human_readable_type, Solr::String, solr_name('human_readable_type', :stored_searchable)
         attribute :representative_id, Solr::String, solr_name('hasRelatedMediaFragment', :symbol)
+        attribute :rendering_ids, Solr::Array, solr_name('hasFormat', :symbol)
```
jcoyne: Is this dependent on `Hyrax.config.rendering_predicate`?

app/presenters/hyrax/manifest_components.rb
jcoyne: use `file_set_document.id` rather than `file_set_document[ActiveFedora.id_field]` because `ActiveFedora` is going away as a dependency when we add Valkyrie. `::SolrDocument.primary_key` may be even more correct.

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
...
+          'label' => I18n.t("hyrax.manifest.download_text") + label
+        }
+      end
+
+      # Query for the properties to create a rendering
+      #
+      # @return [SolrResult] query result
+      def query_for_rendering(file_set_id)
+        ActiveFedora::SolrService.query("id:#{file_set_id}",
+                                        fl: [ActiveFedora.id_field, 'label_ssi', 'mime_type_ssi'],
```
jcoyne: Can we do `::SolrDocument.find(file_set_id)` here? This reduces coupling to ActiveFedora which is going away soon.

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
...
+          renderings << build_rendering(file_set_id)
+        end
+      end
+      renderings.flatten
+    end
+
+    # IIIF metadata for inclusion in the manifest
+    #
+    # @return [Array] array of metadata hashes
+    def manifest_metadata
```
jcoyne: What calls this method?  I only see tests calling it.

app/presenters/hyrax/manifest_components.rb
```diff
@@ -0,0 +1,78 @@
...
+    delegate :rendering_ids, to: :solr_document
+
+    def manifest_url
+      manifest_helper.polymorphic_url([:manifest, self])
+    end
+
+    # IIIF rendering linking property for inclusion in the manifest
+    #
+    # @return [Array] array of rendering hashes
+    def sequence_rendering
```
jcoyne: Nothing appears to be using this method. Is that right?

app/presenters/hyrax/manifest_components.rb
jcoyne: Alternatively, can most of these methods go into the ManifestHelper class?

app/presenters/hyrax/manifest_components.rb
geekscruff: The `iiif_manifest` gem calls it on the object it's given (in this case the presenter).

app/presenters/hyrax/manifest_components.rb
geekscruff: The `iiif_manifest` gem calls it on the object it's given (in this case the presenter).

app/presenters/hyrax/manifest_components.rb
geekscruff: Yes. ManifestComponents is gone, a couple of methods have gone back into WorkShowPresenter and the rest into ManifestHelper ... I think this is better.

app/presenters/hyrax/manifest_components.rb
geekscruff: Done

app/models/concerns/hyrax/solr_document/metadata.rb
geekscruff: Changed so that it retrieves the name from the config.

app/presenters/hyrax/manifest_components.rb
geekscruff: Done

app/actors/hyrax/actors/file_set_actor.rb
geekscruff: I've gone back to the original single method, and disabled the two rubocop errors

app/presenters/hyrax/manifest_components.rb
geekscruff: It's used by `iiif_manifest` to build the manifest, ditto the other one.

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -156,19 +153,50 @@ def model
...
     delegate :member_presenters, :file_set_presenters, :work_presenters, to: :member_presenter_factory
 
+    def manifest_url
+      manifest_helper.polymorphic_url([:manifest, self])
+    end
+
+    # IIIF rendering linking property for inclusion in the manifest
+    #
+    # @return [Array] array of rendering hashes
+    def sequence_rendering
```
jcoyne: I'm not seeing this used anywhere.  What's it for?

jcoyne: Can you add a comment here saying `This is called by the "iiif_manifest"` gem?

jenlindner:
Have all of your requested changes been made, @jcoyne? Looks like they might have been?

geekscruff:
@jenlindner @jcoyne ... I think they have. Would be nice to get this merged.

jenlindner:
Awesome, @geekscruff. I'll check back on this in a bit.

mjgiarlo:
@jcoyne do you have time to review this again soon?

geekscruff:
@jcoyne @mjgiarlo @jenlindner I think the final questions/requests from Justin were addressed. The last thing was a question about the two Rubocop disable statements in the Actor change ... they are both needed and I felt disabling Rubocop in this instance was better than refactoring as the reported complexity is really very slight and the code (to me) makes sense. Would be nice to get this into 2.1 if possible.

PR #2383, opened 2017-12-15
Don't create a Hydra::AccessControl for an unsaved resource

jcoyne:
Hopefully this fixes the bug we've seen in `Dashboard::CollectionsController#new`


jcoyne:
This is strange. While I prevented the problem happening in `Dashboard::CollectionsController#new`, it's still happening in `#create`.  This still seems like a good change to make though.

PR #2384, opened 2017-12-15
Sets ingest_queue_name config param for all jobs

eefahy:
`Hyrax.config.ingest_queue_name` (as set in the hyrax initializer) is not currently declared by all jobs. Therefore, if using a queue name other than `default`, only the jobs that explicitly bring in the configuration are able to find the queue. To fix this, this change sets `queue_as` in `application_job.rb` which therefore sets the queue name for each job to the name configured in the initializer.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/attach_files_to_work_job.rb
+++ b/app/jobs/attach_files_to_work_job.rb
@@ -1,7 +1,5 @@
 # Converts UploadedFiles into FileSets and attaches them to works.
 class AttachFilesToWorkJob < Hyrax::ApplicationJob
-  queue_as Hyrax.config.ingest_queue_name
-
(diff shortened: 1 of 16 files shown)


mjgiarlo:
I'm not sure we want all jobs on the same queue, though. That would be a change in behavior.

tap87:
Is there any information on which jobs where affected by this change. From what I can tell, not all jobs where.
I am using queue prefixes in my app and some of the jobs won't run (e.g. ResolrizeJob)

mjgiarlo:
@tap87 Yeah, that one is broken. I believe there's a ticket for that.

PR #2385, opened 2017-12-15
Get Dashboard::CollectionController tests working with Valkyrie

jcoyne:


spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -163,31 +163,35 @@
...
-      it "renders the form again" do
+      it 'renders the form again' do
         put :update, params: {
           id: collection,
           collection: collection_attrs
         }
         expect(response).to be_successful
         expect(response).to render_template(:edit)
-        expect(assigns[:member_docs]).to be_kind_of Array
+        expect(controller).to have_received(:query_collection_members)
```
cjcolvar: I kind of like the test of `assigns[:member_docs]` because it describes the post condition of the controller action.  Despite the ugly mocking to support this I feel like it might still be better than only checking that a private method was called.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -120,16 +120,6 @@ def after_destroy_success(change_set)
         end
       end
 
-      def after_destroy_error(id)
```
jgondron: Is this no longer called? If not, how are we notifying the user that something went wrong?

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
jcoyne: Thanks for the feedback. I'll make that change.

app/controllers/hyrax/dashboard/collections_controller.rb
jcoyne: I don't believe that there was ever a way to get into this bit of code.  There are no destroy validations on the model. We only had test coverage of this method because we were mocking the response here: https://github.com/samvera/hyrax/blob/master/spec/controllers/hyrax/dashboard/collections_controller_spec.rb#L315

PR #2393, opened 2017-12-18
Create collection migration service

### LINKED ISSUE #2351: Lazy migration of collections not migrating permissions
### Descriptive summary

Lazy migration of collections is setting the gid to User Collections, but is not creating permission templates with proper access based on read/edit access in the collection.

### Rationale

Lazy migration should perform all steps to migrate a legacy user collection to collection-type based collections.

### Expected behavior

A collection that is migrated should...
* have collection_type_gid set to User Collection
* should create a permission template for the collection
* should assign users with edit_access as Managers
* should assign users with read_access as Viewers

### Actual behavior

Only assigns the collection_type_gid

### Related Work

PR #2251 Default the collection type when collection_type_gid_ssim is missing from the Solr document
PR #1618 Raise error when attempting to set collection_type when collection has already been persisted with a collection_type (removed original lazy load code)

elrayle:
Fixes #2351 

Move away from lazy migration which complicates the code and potentially could have performance impacts.  The Hyrax::Collections::MigrationService.migrate_all_collections process in this PR adds the default collection type gid to collections and creating a permission template with permission access assigned for each collection in the repository as part of the migration from Hyrax 2.0.x to Hyrax 2.1.x.

Access is assigned to collections as...
* users/groups are made managers of the collection if they currently have edit access to the collection
* users/groups are made viewers of the collection if they currently have view access to the collection

Also, uses collection type create service to create User Collection and Admin Set collection types instead of FactoryBot.  There were conflicts when FactoryBot was used to create these more than once in a test which isn't allowed.

NOTE:  Lazy migration was removed.  In dashboard/collection_controller.rb, there remains code that assigns the collection_type_gid to be the default user collection type in #new and #create.  These are provided to support backward compatibility of API access.  Any calls coming from the UI should have collection_type_gid as a parameter.  In the long run, the API should probably also require the collection_type_gid to be passed in.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/collection_presenter.rb
+++ b/app/presenters/hyrax/collection_presenter.rb
@@ -25,15 +25,14 @@ module Hyrax
     delegate(*Hyrax::CollectionType.collection_type_settings_methods, to: :collection_type, prefix: :collection_type_is)
 
     def collection_type
-      gid = Array.wrap(solr_document.fetch('collection_type_gid_ssim', [Hyrax::CollectionType.find_or_create_default_collection_type.gid])).first
-      @collection_type ||= CollectionType.find_by_gid!(gid)
+      @collection_type ||= Hyrax::CollectionType.find_by_gid!(collection_type_gid)
     end
 
     # Metadata Methods
     delegate :title, :description, :creator, :contributor, :subject, :publisher, :keyword, :language,
              :embargo_release_date, :lease_expiration_date, :license, :date_created,
              :resource_type, :based_near, :related_url, :identifier, :thumbnail_path,
-             :title_or_label, :collection_type_gid_ssim, :create_date, :modified_date, :visibility, :edit_groups,
+             :title_or_label, :collection_type_gid, :create_date, :modified_date, :visibility, :edit_groups,
              :edit_people,
              to: :solr_document
 
--- a/lib/tasks/migrate.rake
+++ b/lib/tasks/migrate.rake
@@ -26,5 +26,13 @@ namespace :hyrax do
         end
       end
     end
+
+    task add_collection_type_and_permissions_to_collections: :environment do
+      # Run collection migration which sets the collection_type of legacy collections to User Collection and adds
+      # a permission template assigning users/groups with edit_access to the Manager role and users/groups with
+      # read_access to the Viewer role.  Legacy collections are those created prior to collections extensions
+      # added in Hyrax 2.1.0
+      Hyrax::Collections::MigrationService.migrate_all_collections
+    end
   end
 end
--- a/spec/factories/collection_types.rb
+++ b/spec/factories/collection_types.rb
@@ -95,55 +95,10 @@ FactoryBot.define do
   end
 
   factory :user_collection_type, class: Hyrax::CollectionType do
-    title 'User Collection'
-    description 'A user oriented collection type'
-
-    nestable true
-    discoverable true
-    sharable true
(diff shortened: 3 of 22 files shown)


app/presenters/hyrax/collection_presenter.rb
```diff
@@ -25,7 +25,7 @@ def initialize(solr_document, current_ability, request = nil)
     delegate(*Hyrax::CollectionType.collection_type_settings_methods, to: :collection_type, prefix: :collection_type_is)
 
     def collection_type
-      gid = Array.wrap(solr_document.fetch('collection_type_gid_ssim', [Hyrax::CollectionType.find_or_create_default_collection_type.gid])).first
+      gid = Array.wrap(solr_document['collection_type_gid_ssim']).first
```
jcoyne: Should this be defined as a property on the `SolrDocument` class so we can do `gid = solr_document.collection_type_gid`?

lib/tasks/migrate.rake
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/BlockLength
```
jcoyne: This would probably be a good cop to ignore globally for any `*.rake` file.

spec/factories/collection_types.rb
```diff
@@ -95,55 +95,10 @@
...
-                     agent_id: ::Ability.admin_group_name,
-                     agent_type: Hyrax::CollectionTypeParticipant::GROUP_TYPE }
-      create(:collection_type_participant, attributes)
-      attributes = { hyrax_collection_type_id: collection_type.id,
-                     access: Hyrax::CollectionTypeParticipant::MANAGE_ACCESS,
-                     agent_id: ::Ability.admin_group_name,
-                     agent_type: Hyrax::CollectionTypeParticipant::GROUP_TYPE }
-      create(:collection_type_participant, attributes)
-    end
+    initialize_with { Hyrax::CollectionType.find_or_create_admin_set_type }
```
jcoyne: 🆒 ❗️

spec/factories/permission_templates.rb
```diff
@@ -45,13 +45,21 @@
...
+               agent_id: evaluator.depositor.user_key)
+      end
     end
 
     transient do
       with_admin_set false
       with_collection false
       with_workflows false
       with_active_workflow false
+      depositor nil
```
jcoyne: I believe we call this attribute `user` on all the other factories.  Can we change the name for consistency?

spec/presenters/hyrax/collection_presenter_spec.rb
```diff
@@ -50,19 +50,10 @@
 
       it 'finds the collection type based on the solr_document#collection_type_gid if one exists' do
         expect(solr_doc).to be_key('collection_type_gid_ssim')
-        expect(solr_doc).to receive(:fetch).with('collection_type_gid_ssim', Array).and_return(collection_type.gid)
+        expect(solr_doc['collection_type_gid_ssim']).to eq [collection_type.gid]
```
jcoyne: I would just remove this assertion (and the previous line). It's not testing the code in `CollectionPresenter`, which is what this test should be about

spec/factories/collection_types.rb
elrayle: Not sure I understand this comment.

elrayle:
@jcoyne Your requested changes are ready for review.

PR #2394, opened 2017-12-18
Remove import export job as it's Fedora dependent

jcoyne:


cjcolvar:
And these as well: https://github.com/samvera/hyrax/blob/valkyrie-5/lib/hyrax/configuration.rb#L164-L169

jcoyne:
@cjcolvar Done.

mjgiarlo:
@cjcolvar That would be good, as we've been saying that Hyrax 3 is Hyrax 2 with more back-end integration options, i.e., that no features would go away. cc: @vantuyls

vantuyls:
hopefully the impending (?) bulk import, export, edit work will align well with the valkyrie work. in fact, we ought to touch base about that in the new year.

PR #2398, opened 2017-12-18
get by_resource_type_spec.rb passing

kdid:
use `resource_type_ssim` rather than `resource_type_sim`

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/statistics/works/by_resource_type.rb
+++ b/app/services/hyrax/statistics/works/by_resource_type.rb
@@ -5,7 +5,7 @@ module Hyrax
         private
 
           def index_key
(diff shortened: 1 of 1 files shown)


app/services/hyrax/statistics/works/by_resource_type.rb
```diff
@@ -5,7 +5,7 @@ class ByResourceType < Statistics::TermQuery
         private
 
           def index_key
-            Solrizer.solr_name("resource_type", :facetable)
+            Solrizer.solr_name("resource_type", :symbol)
```
jcoyne: How about hard coding `resource_type_ssim` here?  Solrizer is deprecated and is going away, so this might just make it easier.

PR #2399, opened 2017-12-18
Move helper method to helper module

jcoyne:
It's weird to have helper methods in controllers.

### DIFF
--- a/app/helpers/hyrax/hyrax_helper_behavior.rb
+++ b/app/helpers/hyrax/hyrax_helper_behavior.rb
@@ -211,6 +211,18 @@ module Hyrax
       options[:value].map { |right| link_to service.label(right), right }.to_sentence.html_safe
     end
 
+    # A Blacklight facet field helper_method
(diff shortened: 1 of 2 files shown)


app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -211,6 +211,18 @@ def rights_statement_links(options)
       options[:value].map { |right| link_to service.label(right), right }.to_sentence.html_safe
     end
 
+    # A Blacklight facet field helper_method
+    # @param [String] value from blacklight helper_method invocation.
+    # @return [String] the decoded meaning of the boolean field
+    def suppressed_to_status(value)
+      case value
+      when "false"
+        "Published"
```
mjgiarlo: I don't suppose you could be convinced to i18nize these strings?

jcoyne: You are one tough reviewer.

PR #2401, opened 2017-12-18
Fix file_set_change_set_spec.rb

kdid:
initialize empty array for title

@samvera/hyrax-code-reviewers


jcoyne:
If we don't initialize the title to `['']` will it draw the form correctly?

PR #2412, opened 2017-12-19
remove uneccessary to_param override test

kdid:
Removing test as I don't see where the file set is overriding `to_param`. Let me know if I'm missing something.

@samvera/hyrax-code-reviewers


jcoyne:
This one could be ported to master too.

Issue #1024, opened 2017-05-23
"Not Authorized" page is shown for any inability to save workflow action form


laritakr:
### Descriptive summary

UI design on this issue is needed.

The  workflow action form has several validations:

https://github.com/projecthydra-labs/hyrax/blob/master/app/forms/hyrax/forms/workflow_action_form.rb#L37-L48

The controller routes any inability to update the form to an "unauthorized" page, regardless of the reason. This is not intuitive behavior for an error where the action name is not present.

https://github.com/projecthydra-labs/hyrax/blob/master/app/controllers/hyrax/workflow_actions_controller.rb#L16

### Rationale

It is functional but not intuitive to show a "not authorized" error page for situations that are not due to lack of permission.

### Expected behavior

The desired behavior in the UI needs to be determined.

### Actual behavior

The "unauthorized" page is displayed.

### Steps to reproduce the behavior

This requires at least two separate users:

1. Admin user create admin set, change the workflow of the default admin set to be one-step mediated deposit, and add a depositor role to another user to the admin set. 
2. Add that same user as depositor to this admin set/workflow on workflows menu.
3. Sign on as depositor user and deposit a work. Make sure the default admin set is selected in the relationships tab of the 'add new work' view. Sign off.
4. Sign on as admin (or other reviewer) user. Reviewer should get a notification that a work is waiting for review
5. Click the link in the notification or from the review menu. This will take you to the work show view.
6. Expand the "review and approval" widget at the bottom of the window, and click the submit button without selecting a radio button. This takes you to the unauthorized page:

![screen shot 2017-05-22 at 3 54 33 pm](https://cloud.githubusercontent.com/assets/17851674/26365137/ae011450-3fb4-11e7-8606-1b43521e173e.png)

### Related work

https://github.com/projecthydra-labs/hyrax/issues/671

mjgiarlo:
See above, @projecthydra-labs/hyrax-ui-ux-advisors 

Advice is welcome!

ggeisler:
n my opinion, the first step is simply to make one of the radio buttons selected by default (in general there should always be one radio button selected in a related group of buttons). I think "Comment" makes the most sense, in that it is likely to be the most innocuous choice if the user submits without noticing the radio buttons. This would prevent routing to a new error page in the first place.

(I think the actual actions might vary by workflow? Even so, maybe "Comment" is always -- or if not, should be -- an action of a workflow, and thus could reliably be the default action?)

[...]

PR #2415, opened 2017-12-19
Move static form html to radio_button_tag

aaron-collier:
Partially Addresses #1024 

This moves the existing raw html (`input type="radio"`) into `radio_button_tag` code in order to set a default checked value on the `Comment Only` option to avoid the `Not Authorized` message.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_workflow_actions.html.erb
+++ b/app/views/hyrax/base/_workflow_actions.html.erb
@@ -12,7 +12,7 @@
         <% presenter.workflow.actions.each do |key, label| %>
           <div class="radio">
             <label>
-              <%= radio_button_tag 'workflow_action[name]', key, :checked => 'checked' %>
+              <%= radio_button_tag 'workflow_action[name]', key, key == 'comment_only' %>
               <%= label %>
             </label>
           </div>
--- a/app/views/hyrax/base/_workflow_actions.html.erb
+++ b/app/views/hyrax/base/_workflow_actions.html.erb
@@ -12,7 +12,7 @@
         <% presenter.workflow.actions.each do |key, label| %>
           <div class="radio">
             <label>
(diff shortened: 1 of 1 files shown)


app/views/hyrax/base/_workflow_actions.html.erb
```diff
@@ -12,7 +12,7 @@
         <% presenter.workflow.actions.each do |key, label| %>
           <div class="radio">
             <label>
-              <input type="radio" name="workflow_action[name]" id="workflow_action_name_<%= key %>" value="<%= key %>">
+              <%= radio_button_tag 'workflow_action[name]', key, :checked => !!key == 'comment_only' %>
```
jcoyne: `!!key` is going to cast whatever key is to a boolean, right?  then compare that with the string `'comment_only'` which will always be false? Am I reading that right?

aaron-collier: Yup, you're right - DOH - I'll rework that logic.

app/views/hyrax/base/_workflow_actions.html.erb
```diff
@@ -12,7 +12,7 @@
         <% presenter.workflow.actions.each do |key, label| %>
           <div class="radio">
             <label>
-              <input type="radio" name="workflow_action[name]" id="workflow_action_name_<%= key %>" value="<%= key %>">
+              <%= radio_button_tag 'workflow_action[name]', key, :checked => 'checked' %>
```
jcoyne: This will make every radio button default to checked.  Shouldn't it only be one of them?

aaron-collier: Yeah - i was having trouble with any conditional logic not producing the same result. But I think I've fixed that. The 3rd parameter here needs to be just true/false, not an option for `checked`. It should now only be checked on `comment_only`

PR #2416, opened 2017-12-20
Merge latest changes from master into collections-sprint branch

elrayle:


jcoyne:
Looks like there's a problem generating the migrations:
```
SQLite3::SQLException: no such table: permission_templates: ALTER TABLE "permission_templates" ADD "source_type" varchar
```

elrayle:
@jcoyne Yes, I'm not sure what's up with that.  It thinks the permission_templates table doesn't exist.  I assume that was not removed from master.  Any thoughts on the changes in master that might have caused this problem?

PR #2287, opened 2017-11-30
Add more QA objects

elrayle:
Fixes non-scalar values for title and description
Adds QA empty and in-use collection types
Adds QA in-use collection and works
Returns existing users if they were created on a previous run of the seed file

PR #2424, opened 2017-12-20
Add more QA objects to seeds

elrayle:
* Fixes non-scalar values for title and description
* Adds QA empty and in-use collection types
* Adds QA in-use collection and works
* Returns existing users if they were created on a previous run of the seed file
* Update to use collection_attributess instead of just ids


elrayle:
Replaces #2287 which was too out of date to bring current.

PR #2425, opened 2017-12-20
Workflow action - validate that a comment is present

### LINKED ISSUE #1024: "Not Authorized" page is shown for any inability to save workflow action form

aaron-collier:
Fixes #1024 

Adds validation for the presence of a comment in a workflow action. If not present, adds an alert message and reloads the page instead of redirecting to a confusing "Unauthorized page"

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/workflow_actions_controller.rb
+++ b/app/controllers/hyrax/workflow_actions_controller.rb
@@ -7,7 +7,7 @@ module Hyrax
         after_update_response
       else
         respond_to do |wants|
-          wants.html { render 'hyrax/base/unauthorized', status: :unauthorized }
+          wants.html { respond_to_html }
           wants.json { render_json_response(response_type: :unprocessable_entity, options: { errors: curation_concern.errors }) }
         end
       end
@@ -35,6 +35,14 @@ module Hyrax
         respond_to do |wants|
           wants.html { redirect_to [main_app, curation_concern], notice: "The #{curation_concern.human_readable_type} has been updated." }
           wants.json { render 'hyrax/base/show', status: :ok, location: polymorphic_path([main_app, curation_concern]) }
+        end
+      end
+
+      def respond_to_html
+        if workflow_action_form.authorized_for_processing
+          redirect_to [main_app, curation_concern], alert: "Invalid submission: comment required."
+        else
+          render 'hyrax/base/unauthorized', status: :unauthorized
         end
       end
   end
--- a/app/forms/hyrax/forms/workflow_action_form.rb
+++ b/app/forms/hyrax/forms/workflow_action_form.rb
@@ -35,6 +35,8 @@ module Hyrax
       end
 
       validates :name, presence: true
+      validate :valid_comment
+
       validate :authorized_for_processing
 
       def authorized_for_processing
@@ -53,6 +55,10 @@ module Hyrax
         def convert_to_sipity_objects!
           @subject = WorkflowActionInfo.new(work, current_user)
           @sipity_workflow_action = PowerConverter.convert_to_sipity_action(name, scope: subject.entity.workflow) { nil }
+        end
+
+        def valid_comment
+          errors.add(:base, :missing_comment) if comment.blank? && name != 'approve'
         end
 
         attr_reader :subject, :sipity_workflow_action
--- a/app/views/hyrax/base/_workflow_actions.html.erb
+++ b/app/views/hyrax/base/_workflow_actions.html.erb
@@ -1,32 +1,40 @@
 <div id="workflow_controls" class="panel panel-workflow">
   <div class="panel-heading">
     <a data-toggle="collapse" href="#workflow_controls_collapse">
-      <h2 class="panel-title">Review and Approval <i class="fa fa-chevron-right pull-right"></i></h2>
+      <h2 class="panel-title"><%= t('hyrax.base.items.workflow.review') %> <i class="fa fa-chevron-right pull-right"></i></h2>
     </a>
   </div>
   <%= form_tag main_app.hyrax_workflow_action_path(presenter), method: :put do %>
     <div id="workflow_controls_collapse" class="panel-body panel-collapse collapse">
       <div class="col-sm-3 workflow-actions">
-        <h3>Actions</h3>
+        <h3><%= t('hyrax.base.items.workflow.action') %></h3>
 
         <% presenter.workflow.actions.each do |key, label| %>
           <div class="radio">
             <label>
               <%= radio_button_tag 'workflow_action[name]', key, key == 'comment_only' %>
-              <%= label %>
+              <%= t('hyrax.base.items.workflow.actions.' + key) %>
             </label>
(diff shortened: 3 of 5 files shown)


app/controllers/hyrax/workflow_actions_controller.rb
```diff
@@ -37,5 +37,13 @@ def after_update_response
           wants.json { render 'hyrax/base/show', status: :ok, location: polymorphic_path([main_app, curation_concern]) }
         end
       end
+
+      def respond_to_html
+        if workflow_action_form.authorized_for_processing
```
jcoyne: We shouldn't have authorization logic in this method that is only used for HTML users.   We'd expect the same logic when making a JSON request too, right?

app/forms/hyrax/forms/workflow_action_form.rb
```diff
@@ -34,7 +34,7 @@ def save
         true
       end
 
-      validates :name, presence: true
+      validates :name, :comment, presence: true
```
jcoyne: Is a comment always required? Even when it's approved?

app/views/hyrax/base/_workflow_actions.html.erb
```diff
@@ -1,32 +1,40 @@
...
             </label>
           </div>
         <% end %>
       </div>
       <div class="col-sm-9 workflow-comments">
-        <div class="form-group">
-          <label for="workflow_action_comment">Review comment:</label>
-          <textarea class="form-control" name="workflow_action[comment]" id="workflow_action_comment"></textarea>
-        </div>
+          <% if flash['alert'].present? %>
```
jcoyne: Might there be any other possible way to get to this page and have an `flash['alert']` that has nothing to do with this?

app/views/hyrax/base/_workflow_actions.html.erb
```diff
@@ -1,32 +1,40 @@
...
+            <div class="form-group">
+              <label for="workflow_action_comment"><%= t('hyrax.base.items.workflow.comment_label') %>:</label>
+              <textarea class="form-control" name="workflow_action[comment]" id="workflow_action_comment"></textarea>
+            </div>
+          <% end %>
 
         <input class="btn btn-primary" type="submit" value="Submit">
 
-        <h4>Previous Comments</h4>
+        <h4><%= t('hyrax.base.items.workflow.previous') %></h4>
```
jcoyne: 👏

spec/controllers/hyrax/workflow_actions_controller_spec.rb
```diff
@@ -47,7 +57,6 @@
         sign_in(user)
 
         put :update, params: { id: generic_work.to_param, workflow_action: { name: 'advance', comment: '' } }, format: :json
-        expect(response.status).to eq 422
```
jcoyne: Why are you removing this assertion?

app/controllers/hyrax/workflow_actions_controller.rb
jcoyne: Should you call `workflow_action_form.valid?` instead. It looks like `authorized_for_processing` is just one of the validators of the form.

spec/controllers/hyrax/workflow_actions_controller_spec.rb
aaron-collier: This was a mistake I think. Good catch.

app/controllers/hyrax/workflow_actions_controller.rb
aaron-collier: I could be wrong, but I believe we drop into this branch when `workflow_action_form.valid?` is already false as a result of `workflow_action_form.save`, so in this check, we're determining if the validation error is authorization based or due to the missing comment. So checking `workflow_action_form.valid?` would be redundant and not provide enough information. Is there another check available that would make more sense in this context?

app/forms/hyrax/forms/workflow_action_form.rb
aaron-collier: Perhaps my misunderstanding of the mockup, but I assumed so. Will work thought this.

app/views/hyrax/base/_workflow_actions.html.erb
aaron-collier: I'm not sure. It doesn't seem that the `errors` are available on that page in the presenter... Open to suggestion. I could check for the inclusion of "comment" in the alert or something.

jcoyne: Yeah, I don't have a good suggestions.  Since it's a redirect, there's no errors available.

app/controllers/hyrax/workflow_actions_controller.rb
jcoyne: I'm concerned that if  you call this a second time it will add an additional error here: https://github.com/samvera/hyrax/blob/e005a915b3ff56b39bff985885def458f350eb63/app/forms/hyrax/forms/workflow_action_form.rb#L49

aaron-collier: Is that a significant enough issue to warrant concern? Maybe we can walk through this together at some point soon? I'm a bit stuck in that we need to know if the error is based on authorization or just a missing comment, and I'm not seeing another way to check this without duplication of code.

hannahfrost:
@jcoyne Can this be merged? It's related to a ticket that I would like to close.

aaron-collier:
@cjcolvar Didn't realize this was still hanging around. i should be able to get to it this week.

PR #2426, opened 2017-12-20
find correct parent and child works

kdid:
- sets up parent and child works correctly in tests
- remove the "id-" before searching

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/authorities/qa/authorities/find_works_spec.rb
+++ b/spec/authorities/qa/authorities/find_works_spec.rb
@@ -36,7 +36,7 @@ RSpec.describe Qa::Authorities::FindWorks do
       end
 
       it 'displays a list of other works deposited by current user, exluding the child work' do
-        expect(subject.map { |result| result[:id] }).to match_array [work2.id.to_s, work4.id.to_s]
+        expect(subject.map { |result| result[:id] }).to match_array [work2.id.to_s]
       end
     end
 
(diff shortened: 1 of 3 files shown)


spec/authorities/qa/authorities/find_works_spec.rb
```diff
@@ -31,18 +31,18 @@
...
       it 'displays a list of other works deposited by current user, exluding the child work' do
-        expect(subject.map { |result| result[:id] }).to match_array [work2.id.to_s, work4.id.to_s]
+        expect(subject.map { |result| result[:id] }).to match_array [work4.id.to_s]
       end
     end
 
     context "when work has parent works" do
       before do
-        work1.member_ids += [work4.id]
+        work4.member_ids += [work1.id]
```
jcoyne: I think the correct fix here is to back out this change and to change line 51 so that it is only finding `work2`. This would be consistent to how it works in master:

https://github.com/samvera/hyrax/blob/master/spec/authorities/qa/authorities/find_works_spec.rb#L48

spec/authorities/qa/authorities/find_works_spec.rb
```diff
@@ -31,18 +31,18 @@
 
     context "when work has child works" do
       before do
-        work4.member_ids += [work1.id]
-        persister.save(resource: work4)
+        work1.member_ids += [work2.id]
```
jcoyne: I think we should revert this change. and make line 39 just look for work2 similar to https://github.com/samvera/hyrax/blob/master/spec/authorities/qa/authorities/find_works_spec.rb#L37

Issue #2280, opened 2017-11-29
make read access works available to add to collections


vantuyls:
Adding works a user can read/see to a collection is an important use case.

# Current Behavior
User can only add works to a collection that they "own" (ie that show up in My Works)

# Expected Behavior
User should be able to add works to a collection that they can see/read

elrayle:
The behavior is...
* can add works to collections when current_user can edit or deposit into the collection
* cannot add works to collections when current_user only has read access to the collection

elrayle:
This was addressed by the addition of checkboxes to Managed Works such that batch Add to Collection can be used to add readonly works to a collection.  PR #2434

elrayle:
@vantuyls I would like to close this issue.  The title of the issue and the description don't seem to match.  I believe the title is the correct description of the issue.  Such that, it's not that you want to add any work to a readonly collection.  It is that you want to add a readonly work to a collection which you have deposit/edit access.  Please confirm so I can close this issue.

vantuyls:
Yes, i believe that i accurate. Closing now.

Issue #1191, opened 2017-06-12
Adding existing works to existing collection fails


awead:
### Descriptive summary

Given I have a work owned by me
And I have a collection owned by me
When I visit my dashboard
And select my work
And click "Add to Collection"
And select my collection
And click "Update Collection"
Then I should not see `ActionController::InvalidAuthenticityToken`

### Expected behavior

My work should be added to my collection

### Actual behavior

``` bash
Started PUT "/collections/g158bh28p?locale=en" for 127.0.0.1 at 2017-06-12 10:21:54 -0400
Processing by Hyrax::CollectionsController#update as HTML
  Parameters: {"authenticity_token"=>"MdqIljoqgvEyFCRiXnKmGPbxMu8y7ykUlMWUamXLLqMkoLky/vAbpSiLprweauguBwQVCIeAYKTCEYrFzmPVLg==", "batch_document_ids"=>["5m60qr88b"], "collection"=>{"members"=>"add"}, "locale"=>"en", "id"=>"g158bh28p"}
Can't verify CSRF token authenticity.
Completed 422 Unprocessable Entity in 1ms (ActiveRecord: 0.0ms)



ActionController::InvalidAuthenticityToken - ActionController::InvalidAuthenticityToken:
  actionpack (5.1.1) lib/action_controller/metal/request_forgery_protection.rb:195:in `handle_unverified_request'
  actionpack (5.1.1) lib/action_controller/metal/request_forgery_protection.rb:227:in `handle_unverified_request'
  devise (4.3.0) lib/devise/controllers/helpers.rb:253:in `handle_unverified_request'
  actionpack (5.1.1) lib/action_controller/metal/request_forgery_protection.rb:222:in `verify_authenticity_token'
  activesupport (5.1.1) lib/active_support/callbacks.rb:413:in `block in make_lambda'
  activesupport (5.1.1) lib/active_support/callbacks.rb:197:in `block (2 levels) in halting'
  actionpack (5.1.1) lib/abstract_controller/callbacks.rb:12:in `block (2 levels) in <module:Callbacks>'
  activesupport (5.1.1) lib/active_support/callbacks.rb:198:in `block in halting'
  activesupport (5.1.1) lib/active_support/callbacks.rb:507:in `block in invoke_before'
  activesupport (5.1.1) lib/active_support/callbacks.rb:507:in `invoke_before'
  activesupport (5.1.1) lib/active_support/callbacks.rb:130:in `run_callbacks'
[...]

PR #2434, opened 2017-12-21
Add batch actions to all/managed works page.

jonathandixon:
Adds ability to batch select works on the All/Managed page (dashboard/works) and perform batch operations: delete, add to a collection, and edit.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/dashboard/works/_list_works.html.erb
+++ b/app/views/hyrax/dashboard/works/_list_works.html.erb
@@ -1,5 +1,9 @@
 <tr id="document_<%= document.id %>">
   <td>
+    <label for="batch_document_<%= document.id %>" class="sr-only"><%=t("hyrax.dashboard.my.sr.batch_checkbox")%></label>
+    <%= render 'hyrax/batch_select/add_button', document: document %>&nbsp;
+  </td>
+  <td>
     <div class='media'>
       <%= link_to [main_app, document], class: 'media-left', 'aria-hidden' => true do %>
         <%= render_thumbnail_tag document, { class: 'hidden-xs file_listing_thumbnail' }, { suppress_link: true } %>
--- a/app/views/hyrax/my/works/_batch_actions.html.erb
+++ b/app/views/hyrax/my/works/_batch_actions.html.erb
@@ -1,0 +1,14 @@
+<div class="batch-info">
+  <%= render 'hyrax/dashboard/collections/form_for_select_collection', user_collections: @user_collections %>
+
+  <div class="batch-toggle">
+    <% session[:batch_edit_state] = "on" %>
+    <div class="button_to-inline">
+      <%= button_to t('hyrax.dashboard.my.action.edit_selected'), hyrax.edit_batch_edits_path, method: :get, class: "btn btn-update btn-primary hidden submits-batches", data: { behavior: 'batch-edit' }, id: 'batch-edit' %>
+    </div>
+    <%= batch_delete %>
+    <%= button_tag t('hyrax.dashboard.my.action.add_to_collection'),
+        class: 'btn btn-primary submits-batches submits-batches-add',
+        data: { toggle: "modal", target: "#collection-list-container" } %>
+  </div>
+</div>
--- a/app/views/hyrax/my/works/_tabs.html.erb
+++ b/app/views/hyrax/my/works/_tabs.html.erb
@@ -1,7 +1,7 @@
 <ul class="nav nav-tabs" id="my_nav" role="navigation">
   <span class="sr-only">You are currently listing your works.  You have <%= @response.docs.count %> <%= 'work'.pluralize(@response.docs.count)%> </span>
   <li<%= ' class="active"'.html_safe if current_page?(hyrax.dashboard_works_path(locale: nil)) %>>
-    <%= link_to t("hyrax.dashboard.#{current_ability.admin? ? 'all' : 'managed'}.works"), hyrax.dashboard_works_path %>
+    <%= link_to t("hyrax.dashboard.#{current_ability.admin? ? 'all' : 'managed'}.works"), hyrax.dashboard_works_path, data: { turbolinks: false } %>
   </li>
   <li<%= ' class="active"'.html_safe if current_page?(hyrax.my_works_path(locale: nil)) %>>
     <%= link_to t('hyrax.dashboard.my.works'), hyrax.my_works_path, data: { turbolinks: false } %>
--- a/spec/controllers/hyrax/batch_edits_controller_spec.rb
+++ b/spec/controllers/hyrax/batch_edits_controller_spec.rb
@@ -10,11 +10,13 @@ RSpec.describe Hyrax::BatchEditsController, type: :controller do
   describe "#edit" do
     let(:one) { create(:work, creator: ["Fred"], title: ["abc"], language: ['en']) }
     let(:two) { create(:work, creator: ["Wilma"], title: ["abc2"], publisher: ['Rand McNally'], language: ['en'], resource_type: ['bar']) }
+    let(:three) { create(:work, creator: ["Dino"], title: ["abc3"]) }
 
     before do
-      controller.batch = [one.id, two.id]
+      controller.batch = [one.id, two.id, three.id]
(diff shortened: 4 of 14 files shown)


elrayle:
This addresses issue #2280 - Make read access works available to add to collections

jonathandixon:
Added suggested tests: batch edit/delete that includes read-only works updates/deletes the edit access works and not the read access works.

spec/views/hyrax/my/_search_header.html.erb_spec.rb
```diff
@@ -12,6 +12,7 @@
 
   context "on my works page" do
     before do
+      view.lookup_context.prefixes.push "hyrax/my/works"
       allow(view).to receive(:on_my_works?).and_return(true)
```
jcoyne: Remove

spec/views/hyrax/my/_search_header.html.erb_spec.rb
```diff
@@ -22,15 +23,14 @@
     end
   end
 
-  context "not on my works page (i.e. one of the collections index pages)" do
+  context "on my collections page" do
     before do
+      view.lookup_context.prefixes.push "hyrax/my/collections"
       allow(view).to receive(:on_my_works?).and_return(false)
```
jcoyne: remove?

jonathandixon:
@jcoyne I'm fine with removing `DashboardHelperBehavior#on_my_works?`. Should this be done in a separate issue? I worry that removing it here will get lost and that downstream applications that use it in their templates will have problems on upgrade.

jcoyne:
@jonathandixon I think you should just remove it here because we have no other uses of it in hyrax.  If you don't want to  remove it here out of an abundance of caution, you should at least deprecate it.

app/views/hyrax/my/works/_tabs.html.erb
```diff
@@ -1,7 +1,7 @@
 <ul class="nav nav-tabs" id="my_nav" role="navigation">
   <span class="sr-only">You are currently listing your works.  You have <%= @response.docs.count %> <%= 'work'.pluralize(@response.docs.count)%> </span>
   <li<%= ' class="active"'.html_safe if current_page?(hyrax.dashboard_works_path(locale: nil)) %>>
-    <%= link_to t("hyrax.dashboard.#{current_ability.admin? ? 'all' : 'managed'}.works"), hyrax.dashboard_works_path %>
+    <%= link_to t("hyrax.dashboard.#{current_ability.admin? ? 'all' : 'managed'}.works"), hyrax.dashboard_works_path, data: { turbolinks: false } %>
```
jcoyne: Why are you doing `turbolinks: false` here?

app/views/hyrax/my/works/_batch_actions.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="batch-info">
+  <%= render 'hyrax/dashboard/collections/form_for_select_collection', user_collections: @user_collections %>
+
+  <div class="batch-toggle">
+    <% session[:batch_edit_state] = "on" %>
+    <div class="button_to-inline">
+      <%= button_to "Edit Selected", hyrax.edit_batch_edits_path, method: :get, class: "btn btn-update btn-primary hidden submits-batches", data: { behavior: 'batch-edit' }, id: 'batch-edit' %>
```
jcoyne: Can you add this string to i18n? and on line 10?

spec/controllers/hyrax/batch_edits_controller_spec.rb
```diff
@@ -35,20 +37,27 @@
...
       expect(controller).to receive(:can?).with(:edit, two.id).and_return(true)
+      expect(controller).to receive(:can?).with(:edit, three.id).and_return(false)
     end
 
     it "is successful" do
       put :update, params: { update_type: "delete_all" }
       expect(response).to redirect_to(dashboard_path(locale: 'en'))
       expect { GenericWork.find(one.id) }.to raise_error(Ldp::Gone)
       expect { GenericWork.find(two.id) }.to raise_error(Ldp::Gone)
+      expect(GenericWork.find(three.id).id).to eq(three.id)
```
jcoyne: This should be `expect(GenericWork).to exist(three.id)` (or alternately `expect(GenericWork.exists?(three.id).to be true`)

spec/controllers/hyrax/batch_edits_controller_spec.rb
```diff
@@ -61,13 +70,15 @@
       expect(response).to be_redirect
       expect(GenericWork.find(one.id).subject).to eq ["zzz"]
       expect(GenericWork.find(two.id).subject).to eq ["zzz"]
+      expect(GenericWork.find(three.id).subject).not_to eq ["zzz"]
```
jcoyne: Negative assertions are not nearly as specific as positive assertions.  Is there a reason you can't do `expect(GenericWork.find(three.id).subject).to eq ["whavever value it initially had"]`?

spec/controllers/hyrax/batch_edits_controller_spec.rb
```diff
@@ -61,13 +70,15 @@
...
     end
 
     it "updates permissions" do
       put :update, params: { update_type: "update", visibility: "authenticated" }
       expect(response).to be_redirect
       expect(GenericWork.find(one.id).visibility).to eq "authenticated"
       expect(GenericWork.find(two.id).visibility).to eq "authenticated"
+      expect(GenericWork.find(three.id).visibility).not_to eq "authenticated"
```
jcoyne: Can we assert what we expect it to be? `restricted`?

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -74,7 +74,7 @@
 
       it "adds docs to the collection and adds the collection id to the documents in the collection" do
         post :create, params: {
-          batch_document_ids: [asset1.id],
+          batch_document_ids: [asset1.id, unowned_asset.id],
```
jcoyne: Why are you doing this change?  It seems odd that no assertions are changing.

spec/controllers/hyrax/batch_edits_controller_spec.rb
```diff
@@ -35,20 +37,27 @@
     let!(:two) do
       create(:generic_work, creator: ["Fred"], title: ["abc"], language: ['en'], user: user)
     end
+
+    let!(:three) do
+      create(:generic_work, creator: ["Fred"], title: ["abc"], language: ['en'])
```
jcoyne: On line 13 above you are using `create(:work,...)`  can we stick with that for consistency?

app/views/hyrax/dashboard/works/_list_works.html.erb
```diff
@@ -1,4 +1,8 @@
 <tr id="document_<%= document.id %>">
+  <td>
+    <label for="batch_document_<%= document.id %>" class="sr-only"><%=t("hyrax.dashboard.my.sr.batch_checkbox")%></label>
+    <%= render 'hyrax/batch_select/add_button', document: document %>&nbsp;
```
jcoyne: What's the point of the `&nbsp` here?

app/views/hyrax/my/works/_tabs.html.erb
jonathandixon: This relates to issue #1191. Without it, you end up with an `ActionController::InvalidAuthenticityToken` error.

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
jonathandixon: This is to verify that only works the user has access to are added to the collection. This passes an ID of an unowned work with the request but that shouldn't change the original expectation that only `asset1` is added to the collection.

app/views/hyrax/dashboard/works/_list_works.html.erb
jonathandixon: This was a copy and paste. It appears the `&nbsp;` is used to manipulate the vertical position of the checkbox inside the table cell (without it the checkbox renders below verticle center).

PR #2451, opened 2018-01-04
force tests to run

elrayle:
There appears to be a dependency change causing test failures. Using this branch to try and track down the problem.

PR #2450 - Tests passed (except for known random failure) at master.  

Performing a similar forced test at collections-sprint.


elrayle:
This did indeed reveal the same set of errors seen in pending PRs.  In Slack, @geekscruff recognized the pattern of failures as being caused by a solr configuration that was fixed at master.

PR #2452, opened 2018-01-04
Merge latest changes from master into collections-sprint branch

elrayle:
Includes a configuration change required for latest solr that was causing tests to fail.

### DIFF
--- a/app/services/hyrax/collections/nested_collection_query_service.rb
+++ b/app/services/hyrax/collections/nested_collection_query_service.rb
@@ -39,7 +39,10 @@ module Hyrax
         query_builder = Hyrax::Dashboard::NestedCollectionsSearchBuilder.new(access: access, collection: collection, scope: scope)
         # No sense returning everything, just limit to a single entry
         query_builder.where(id: limit_to_id) if limit_to_id
-        scope.repository.search(query_builder.query).documents
+        # TODO: Need to investigate further to understand why this particular query fails if prepended with `{!lucene}` when all others work with the prepend
+        query = query_builder.query.to_hash
+        query['q'].gsub!('{!lucene}', '') if query.key? 'q'
+        scope.repository.search(query).documents
       end
       private_class_method :query_solr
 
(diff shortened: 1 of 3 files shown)


elrayle:
The failing test appears to be caused by the prepending of `{!lucene}` to the solr query that checks if parent is available for nesting.  If I include that and execute the query in the solr dashboard, nothing is returned.  If I remove `{!lucene}` from the query, it correctly finds the parent.

@geekscruff Did you see any other queries failing with the addition of lucene in the config?  Do you know if I need to change the NestedCollectionsSearchBuilder?  Or why all other search builders might work with `{!lucene}`, but not this one?

app/services/hyrax/collections/nested_collection_query_service.rb
```diff
@@ -39,7 +39,10 @@ def self.query_solr(collection:, access:, scope:, limit_to_id:)
         query_builder = Hyrax::Dashboard::NestedCollectionsSearchBuilder.new(access: access, collection: collection, scope: scope)
         # No sense returning everything, just limit to a single entry
         query_builder.where(id: limit_to_id) if limit_to_id
-        scope.repository.search(query_builder.query).documents
+        # TODO: Need to investigate further to understand why this particular query fails if prepended with `{!lucene}` when all others work with the prepend
+        query = query_builder.query.to_hash
+        query['q'].gsub!('{!lucene}', '') if query.has_key? 'q'
```
elrayle: I'm not sure why this particular query fails when others were fine with the prepending of `{!lucene}` to the query.

There is likely a better way to prevent `{!lucene}` from being prepended instead of removing it after the fact.

elrayle: Failing test: `rspec ./spec/services/hyrax/collections/nested_collection_query_service_spec.rb:95 # Hyrax::Collections::NestedCollectionQueryService.parent_and_child_can_nest? given parent and child are nestable and are of the same collection type should eq true`

To reproduce through the UI:
* Create two User Collection collections
* Navigate to the show page of one of the collections
* Click `Add a subcollection` button
* Select the other collection as the child
* Click `Add a subcollection` button

You will get an error complaining that the child cannot be added to the parent and the parent cannot have the child as a subcollection.  (Wording not exact.)

NOTE: You will not see either error with commit `remove {!lucene} from query_solr used with parent_and_child_can_nest?`

elrayle:
@jcoyne @geekscruff I plan to merge the code changes as is and address the question of whether this is the best approach in issue #2453.  This will unblock 3 PRs that are currently blocked by the lucene issue.  Please make your comments in that issue.

PR #2460, opened 2018-01-08
Update translations

mjgiarlo:
Per the updated documentation at https://github.com/samvera/hyrax/wiki/Translations-(internationalization)#update-translations

Fixes #2444 

@samvera/hyrax-code-reviewers

Issue #2444, opened 2017-12-24
locales   hyrax.??.yml  files out of sync 


jmfrenzel:
### Descriptive summary

hyrax.??.yml files should have identical entries >  went through the locales in HEAD an found some missing/additional/not-removed entries. Backtracking this (see below section Related Work) I see the problem when entries in hyrax.en.yml are deleted but not in all the other  hyrax.??.yml files. 

### Rationale
Must make sure having the same  l18 variable entries in __all__ the hyrax.??.yml files when PR get accepted.  If possible a test would be ideal.  

### Expected behavior
Block merging PR's if a l18 variable entry or block in hyrax.en.yml got deleted but at the same time the corresponding ones in the hyrax.??.yml files are not.

### Actual behavior
There is no (blocking) check when a l18 variable entry or block in hyrax.en.yml got deleted but not the corresponding ones in the hyrax.??.yml files.

### Steps to reproduce the behavior

1. '''diff hyrax.??.yml''' in the current version (commit 28f8773a63100cc3169bcfda52f777914123c911)
1. hyrax.en.yml bisecting commits starting from the current version (commit 28f8773a63100cc3169bcfda52f777914123c911) tracing back to th commits (see below) 

### Related work 
I created some comments to the PR's questioning the purpose.
* https://github.com/samvera/hyrax/commit/f7ef416772b4e7df4e34ce031f71f93aaf197fdc#r26461437  
* https://github.com/samvera/hyrax/commit/27c405d924b046cbe755ff79864bfe60f1b21fc3#r26461527 
* https://github.com/samvera/hyrax/commit/1743c7b569b37cc3cb5d8aff441dedd87c35d744#r26475591
* https://github.com/samvera/hyrax/commit/18557e5ee4ca602d7a320b6fb91f3596326487f7#r26475664
* https://github.com/samvera/hyrax/commit/36f44803dd1fdbb06f7022d6ed5b4156823fcaf8#r26476187
* thats all so far, except missing translation two entries

mjgiarlo:
@jmfrenzel See the following PR: https://github.com/samvera/hyrax/pull/2460

And the updated documentation: https://github.com/samvera/hyrax/wiki/Translations-(internationalization)#update-translations (step 3, which I added, in particular)

Per current practice, I would like to address this at point of release rather than at point of PR to reduce the burden on PR authors. (This is addressed by step two here: https://github.com/samvera/hyrax/wiki/Release-management-process)

[...]

PR #2462, opened 2018-01-09
Use Blacklight 6.14+ to pick up Chinese translations

mjgiarlo:
Follows #2460, in which the steps documented in https://github.com/samvera/hyrax/wiki/Translations-(internationalization)#update-translations were run, one of which removed Chinese translations for Blacklight (because they do not exist in English, the base language for Hyrax). We should avoid managing Blacklight translations in Hyrax, and now that someone has contributed a Chinese translation to Blacklight, we no longer need this in Hyrax.

Refs #2444 

Also, get the build working by adding a TravisCI workaround.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: ruby
-sudo: false
+# Workaround for https://github.com/travis-ci/travis-ci/issues/8836
+# sudo: false
+sudo: required
 dist: trusty
 
 addons:
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -32,7 +32,7 @@ SUMMARY
   spec.add_dependency 'active-fedora', '~> 11.5', '>= 11.5.2'
   spec.add_dependency 'almond-rails', '~> 0.1'
   spec.add_dependency 'awesome_nested_set', '~> 3.1'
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: ruby
 sudo: false
-dist: trusty
+dist: xenial
```
jcoyne: Is this valid?  I'm under the impression that the only valid values here are `precise` or `trusty`

jcoyne: well, it looks like it works, even if it isn't in the documentation.

mjgiarlo: It was a shot in the dark after reading https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-356362524. Unfortunately, that didn't fix the build.

.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: ruby
-sudo: false
+# Workaround for https://github.com/travis-ci/travis-ci/issues/8836
```
mjgiarlo: I have subscribed to the above GitHub issue so that I am notified of updates. Once this has been resolved, I will put in a follow-up PR to revert this change.

PR #2470, opened 2018-01-10
Fix work show presenter spec

jgondron:
- The solr doc no longer includes hasRelatedMediaFragment, is now representative_id. This id is a Valkyrie::ID, which has a prefix of "id-". Had to create a custom type that would remove the prefix since the rest of Hyrax doesn't expect this atm. Updated the readme to note this change in the schema will require reindexing once these changes are released.
- Several of the tests were broken because the CompositePresenterFactory was not creating a FileSetPresenter correctly. Was due to a type mismatch in the check for inclusion in file_set_ids.
- Not entirely sure if the "displays them in order" test is still valid. ordered_member_ids is no longer a method, now uses member_ids. I'm assuming the rest of hyrax still expects that to be ordered correctly, so updated to check the member_ids order.
- Updated the solr query tests to spy on the Solr::Client instance that the metadata adapter is using since it no longer uses ActiveFedora::SolrService.

@samvera/hyrax-code-reviewers

### DIFF
--- a/Valkyrie_README.md
+++ b/Valkyrie_README.md
@@ -23,7 +23,7 @@
 * Update config/initializers/riiif.rb
   * set `Riiif::Image.file_resolver = Hyrax::Riiif::ValkyrieFileResolver.new`
   * Remove the `Riiif::Image.file_resolver.id_to_uri = lambda do |id| ... end `
-
+* There have been changes to the solr document schema, so all documents will need to be reindexed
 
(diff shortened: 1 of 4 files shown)


jcoyne:
@jgondron looks good can you rebase this (and squash it while you're at it)?

jgondron:
@jcoyne rebased/squashed

PR #2483, opened 2018-01-12
[WIP] create solr test data directly instead of using save to fedora to create it

elrayle:
IMPROVE TEST SPEED

Two tests were identified as needing data in Solr but not Fedora.  Previously, saving to Fedora was used to create the Solr documents.  They were reworked to directly create Solr documents and make 0 writes to Fedora.

Improvements on local machine...
* spec/authorities/qa/authorities/collections_spec.rb - went from 11s to < 1s
* spec/services/hyrax/collections_service_spec.rb - went from 6s to < 1s

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/authorities/qa/authorities/collections_spec.rb
+++ b/spec/authorities/qa/authorities/collections_spec.rb
@@ -1,41 +1,76 @@
 RSpec.describe Qa::Authorities::Collections, :clean_repo do
   let(:controller) { Qa::TermsController.new }
-  let(:user1) { create(:user) }
-  let(:user2) { create(:user) }
+  let(:user1) { build(:user) }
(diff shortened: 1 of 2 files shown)


jenlindner:
@elrayle Travis is timing out, which is pretty ironic given that these tests speed up the suite. I restarted the build once but got the same problem. But since Travis might be a little iffy right now, I don't know that this is our problem yet. I'll try restarting it again in a few hours. But I ++ this PR and the lighter tests.

jenlindner:
Hey @elrayle, this is still timing out, I restarted a few times. I don't know what's going with it :-(

elrayle:
Closing this with a preference for the full collection factory refactor in PR #2667

Issue #2489, opened 2018-01-14
User cannot add subcollections or works to a collection for which they have deposit access.


elrayle:
### Descriptive summary

A user with deposit access to a collection cannot add subcollections or works to the deposit access collection in a number of areas.

* cannot using subcollection buttons on show page
* cannot using batch add to collection on works index page

NOTE: 
* Manager users with edit access can perform these actions.
* Deposit users can add a work to a deposit collection from the work edit form.

### Rationale

A deposit user must have access to the subcollection buttons on the collection admin show page to be able to manage sub-collections.  

To add a subcollection, the user must have...
* read access to the child collection
* deposit access to the parent collection

### Expected behavior

If the user is on the collection admin show page, the user should be able to add an existing valid collection as a subcollection or create a new collection to serve as the subcollection.

### Actual behavior

If the user is on the collection admin show page
1.  and the user chooses to add an existing collection as a subcollection, the modal is displayed, but there are 0 collections listed.
2.  and the user chooses to add a new collection as a subcollection, the user is shown message stating "Unauthorized. The page you have tried to access is private."

If the user is on the works page using batch add to collection process
3. the user can select the deposit access collection, but the works are not added

For 3, the cause is at least in part  that the process for adding uses controller/hyrax/dashboard/collection_controller which limits update action to editors only.  So depositors are rejected.

#### Prep
1. Create 2 users
1. User 1 creates a collection of type User Collection named 'col-with-user2-depositor' and shares Deposit access to it

#### Add a subcollection from collection admin show page
1. Login as user 2
1. Create a second collection of type User Collection named 'sub-collection'
1. Dashboard -> Collections -> Managed Collections -> click on title 'col-with-user2-depositor'
1. Click `Add a subcolleciton` button

Verify the subcollection modal is displayed and the select list includes 'sub-collection'.

1. Select 'sub-collection'
1. Click `Add a subcollection` button

Verify that 'sub-collection' is listed as a Subcollection of 'col-with-user2-depositor'.

#### Add a new collection as subcollection from collection admin show page
1. Login as user 2
1. Dashboard -> Collections -> Managed Collections -> click on title 'col-with-user2-depositor'
1. Click `Create new collection as subcollection` button
1. Set title of new collection to 'sub-collection-2'
1. Save
1. Dashboard -> Collections -> Managed Collections -> click on title 'col-with-user2-depositor'

Verify that 'sub-collection-2' is listed as a Subcollection of 'col-with-user2-depositor'.

[...]

PR #2491, opened 2018-01-16
allow deposit user to create a new collection as a subcollection

elrayle:
Partial Fix for #2489

The access checks for whether a collection can be a parent or subcollection were too restrictive.  The expected access requirements are...
* require deposit access for parent collection
* require read access for sub-collection

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/nest_collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/nest_collections_controller.rb
@@ -45,21 +45,21 @@ module Hyrax
 
         def build_within_form
           child = Collection.find(params.fetch(:child_id))
-          authorize! :edit, child
+          authorize! :read, child
           parent = params.key?(:parent_id) ? Collection.find(params[:parent_id]) : nil
           form_class.new(child: child, parent: parent, context: self)
         end
 
         def build_under_form
           parent = Collection.find(params.fetch(:parent_id))
-          authorize! :edit, parent
+          authorize! :deposit, parent
           child = params.key?(:child_id) ? Collection.find(params[:child_id]) : nil
           form_class.new(child: child, parent: parent, context: self)
         end
 
         def build_create_collection_form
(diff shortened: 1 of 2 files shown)


app/controllers/hyrax/dashboard/nest_collections_controller.rb
```diff
@@ -45,21 +45,21 @@ def create_relationship_under
 
         def build_within_form
           child = Collection.find(params.fetch(:child_id))
-          authorize! :edit, child
+          authorize! :read, child
```
jeremyf: My understanding of collections is, since we are editing the child's relationships, we should have :edit permissions. I also believe there is a tension in the direction of the relationship. 

In the case where the Library has deposited a "Featured Work of Utmost Awesomeness" and a user creates a collection "Silly Jokey Stuff" and includes the "Featured Work of Utmost Awesomeness" in that collection, then on the work there would be an asserted relationship to "Silly Jokey Stuff". I don't believe that would be an acceptable relationship in the eyes of the Library.

I'm not certain how we should proceed with this.

jeremyf: Based on Slack conversations, I'm fine proceeding with this implementation. I'm not certain how we could do otherwise given the current implementation constraints.

elrayle:
Summary from Slack conversation:
> (elr37) We switched the relationship for arbitrary performance reasons, not for access reasons.  From an intuitive perspective, disregarding the direction of the relationship, it seems like I should be able to collect together in a collection anything I can see.  And I wouldn't expect to be able to add a sub-collection to a parent collection simply because I can edit the sub-collection and its is_member_of relationship.

> (elr37) Additionally, an institution could decide they have more parents than children and decide to switch the relationship back the other way.  If we codify in the code the access levels, we will mess up one direction or the other.  It makes more sense to me to decide what is meaningful and intuitive.

PR #2500, opened 2018-01-16
Fix member_presenter_factory_spec.rb

jgondron:
- Added attributes that are now expected in the solr document.

@samvera/hyrax-code-reviewers


spec/presenters/hyrax/member_presenter_factory_spec.rb
```diff
@@ -2,7 +2,7 @@
   describe "#file_set_presenters" do
     describe "getting presenters from factory" do
       let(:solr_document) { SolrDocument.new(attributes) }
-      let(:attributes) { {} }
+      let(:attributes) { { id: "id-fileset", :created_at_dtsi=>"2018-01-16T18:56:36Z", internal_resource_ssim: ["FileSet"] } }
```
jcoyne: Please use ruby 1.9 hash syntax here.

jgondron: Doh. Surprised hound didn't catch this...

PR #2504, opened 2018-01-16
 show parent collections on collection show pages 

### LINKED ISSUE #1699: Collections: Parent and subcollections in collection show page
### Descriptive summary

Need to include nested collections in the collection show page.  This includes showing the parent collection(s) and the sub-collections.

#### Parent Collections

We discussed having the parent collections display titles only near the top of the page, likely just below the title.

#### Sub-collections

There are two approaches on the table for sub-collections:

* OPTION 1: sub-collections are mixed in with works in the Items list
* OPTION 2: Works only are in the items list.  Just above or below the list of works is the list of sub-collections

Nuances for option 2:
Should the list of subcollections include...
* titles only with a link to the subcollection
* titles + items count (works + subcollections)
* titles + works count + subcollections count (similar to search results)
* full details like on collections index page

1. Add nested collections at least 3 deep
1. Go to show page for each of the collections

Verify that parents are shown as appropriate.
Verify that sub-collections are show as appropriate.

### Related work

TBD

elrayle:
Partially Fixes #1699

Shows list of parent collections on the admin and public collection show pages.
* by default, show the first X parents based on the value of parents_to_show
* adds parents_to_show in presenter with a default value of 3
* uses javascript to show/hide parents beyond the first 3

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_collections.scss
+++ b/app/assets/stylesheets/hyrax/_collections.scss
@@ -624,3 +624,14 @@ button.branding-banner-remove:hover {
   margin-left: 0.5em;
 
 }
+
+.more-parent-collections-block {
+  display: none;
+}
+
+.show-more-parent-collections-btn,
+.show-less-parent-collections-btn {
+  color: #8598b7;
+  font-size: 0.8em;
+  margin-left: 10px;
+}
--- a/app/views/hyrax/collections/_show_parent_collections.html.erb
+++ b/app/views/hyrax/collections/_show_parent_collections.html.erb
@@ -1,14 +1,31 @@
-<% if @parent_collections.nil? || @parent_collections.empty? %>
+<% if presenter.parent_collection_count <= 0 %>
 		<div class="alert alert-warning" role="alert"><%= t('hyrax.collections.show.no_visible_parent_collections') %></div>
 <% else %>
 		<div class="media-body">
 			<h4 class="media-heading">
-				<% @parent_collections.each do |document| %>
+				<div class="less-parent-collections-block" id="less-parent-collections">
+					<% presenter.visible_parent_collections.each do |document| %>
 						<ul>
 							<li>
 								<%= link_to document, [hyrax, document], id: "src_copy_link_#{document.id}" %>
 							</li>
 						</ul>
+					<% end %>
+					<% if presenter.more_parent_collections? %>
+							<button id="show-more-parent-collections" class="btn show-more-parent-collections-btn"><%= t("hyrax.collections.show.show_more_parent_collections") %></button>
+					<% end %>
+				</div>
+				<% if presenter.more_parent_collections? %>
+					<div class="more-parent-collections-block" id="more-parent-collections">
+						<% presenter.more_parent_collections.each do |document| %>
+							<ul>
+								<li>
+									<%= link_to document, [hyrax, document], id: "src_copy_link_#{document.id}" %>
+								</li>
+							</ul>
+							<button id="show-less-parent-collections" class="btn show-less-parent-collections-btn"><%= t("hyrax.collections.show.show_less_parent_collections") %></button>
+						<% end %>
+					</div>
 				<% end %>
 			</h4>
 		</div>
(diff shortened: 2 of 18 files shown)


app/assets/stylesheets/hyrax/_collections.scss
```diff
@@ -624,3 +624,13 @@ button.branding-banner-remove:hover {
   margin-left: 0.5em;
 
 }
+
+.more-parent-collections-block {
+  display: none;
+}
+
+.show-more-parent-collections-btn, .show-less-parent-collections-btn {
```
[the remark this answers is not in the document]
elrayle: Fixed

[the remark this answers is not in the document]
elrayle: Fixed

app/views/hyrax/collections/_show_parent_collections.html.erb
```diff
@@ -0,0 +1,30 @@
+<% if @parent_collections.nil? || @parent_collections.empty? %>
+		<div class="alert alert-warning" role="alert"><%= t('hyrax.collections.show.no_visible_parent_collections') %></div>
+<% else %>
+		<div class="media-body">
+			<h4 class="media-heading">
```
adamjarling: Typically you would't want to keep content within a `<h4>` or any headline `<hX>` tag just text, and not embed markup (like the nested `<ul><li>...`).  If it works that's fine, but will raise some accessibility flags.

app/views/hyrax/collections/_show_parent_collections.html.erb
```diff
@@ -0,0 +1,30 @@
...
+							<ul>
+								<li>
+									<%= link_to document, [hyrax, document], id: "src_copy_link_#{document.id}" %>
+								</li>
+							</ul>
+							<% i += 1 %>
+							<% include_more_tag = true if(i == @presenter.parents_to_show && i < @parent_collections.size) %>
+							<% if include_more_tag %>
+									<button id="show-more-parent-collections" class="btn show-more-parent-collections-btn"><%= t("hyrax.collections.show.show_more_parent_collections") %></button>
+									</div>
```
adamjarling: I'd verify that the markup properly closes these `<div>` tags.  Generally you'd probably want to keep paired opening and closing `<div>` tags within the same `if/else` block.

app/views/hyrax/collections/show.html.erb
```diff
@@ -55,6 +55,22 @@
 	<div class="row hyc-body">
 		<div class="col-md-8 hyc-description">
 			<%= render 'collection_description', presenter: @presenter %>
+
+			<% if @presenter.collection_type_is_nestable? && @parent_collection_count > 0 %>
+					<div class="row hyc-blacklight hyc-bl-title">
```
adamjarling: You might be able to remove the `row` class here, and remove the `<div class="col-md-12"></div>` below, to keep the markup cleaner.

app/views/hyrax/collections/show.html.erb
```diff
@@ -55,6 +55,22 @@
...
+						<div class="col-md-12">
+							<h2>
+								<%= t('.parent_collection_header') %> (<%= @parent_collection_count %>)
+							</h2>
+						</div>
+					</div>
+					<div class="row hyc-blacklight hyc-bl-results">
```
adamjarling: Same as above here.  Could remove the `row` class and the `<div class="col-md-12"></div>` tags.

app/views/hyrax/collections/show.html.erb
elrayle: done

app/views/hyrax/collections/show.html.erb
elrayle: done

app/views/hyrax/collections/_show_parent_collections.html.erb
elrayle: Yea, I wouldn't normally do that.  I copied it from _subcollection_list.html.erb and didn't update it.  I think that was being used to make the collection names bold and larger.  That pattern is used in a number of places.

I think I will leave it for now and we should probably revisit with a refactor of all the places it is used.

app/views/hyrax/collections/_show_parent_collections.html.erb
elrayle: Improved presenter to include a variable for the visible_parent_collections and more_parent_collections.

app/views/hyrax/collections/_show_parent_collections.html.erb
elrayle: Created issue #2505 to request this pattern to be refactored.

PR #2506, opened 2018-01-17
Extract the nested permission processing into a change_set mixin

jcoyne:


app/actors/hyrax/actors/grant_edit_to_depositor_actor.rb
```diff
@@ -0,0 +1,13 @@
+module Hyrax
+  module Actors
+    # Gives the depositor edit access to their work
+    class GrantEditToDepositorActor < Hyrax::Actors::AbstractActor
+      # @param [Hyrax::Actors::Environment] env
+      # @return [Valkyrie::Resource,FalseClass] the saved resource if creates was successful
+      def create(env)
+        env.change_set.read_users += [env.user.user_key]
```
cjcolvar: It looks like this should be `edit_users +=`.

app/actors/hyrax/actors/interpret_visibility_actor.rb
```diff
@@ -111,6 +111,9 @@ def apply_visibility(env, intention)
               env.curation_concern.visibility = env.attributes[:visibility]
             end
           end
+
+          # Copy any visibility changes to the ChangeSet
+          env.change_set.read_groups = env.curation_concern.read_groups
```
cjcolvar: Is this supposed to be `visibility` instead of `read_groups` as the comment states?

app/change_sets/hyrax/collection_change_set.rb
```diff
@@ -77,10 +75,14 @@ def secondary_terms
       # rubocop:enable Metrics/MethodLength
     end
 
-    # Give the depositor edit access
     def sync
-      self.edit_users = [user.user_key]
-      super
+      # Give the depositor edit access
+      self.edit_users += [user.user_key]
```
cjcolvar: Should this guard against duplicate entries of `user.user_key` in `edit_users`?

app/actors/hyrax/actors/interpret_visibility_actor.rb
jcoyne: Not really.  The visibility is like a macro for read_groups.

app/change_sets/hyrax/collection_change_set.rb
jcoyne: I believe Valkyrie::Set is already doing that for us.

PR #2510, opened 2018-01-17
Use change_set for errors

kdid:
@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/import_url_failure_service_spec.rb
+++ b/spec/services/hyrax/import_url_failure_service_spec.rb
@@ -5,14 +5,15 @@ RSpec.describe Hyrax::ImportUrlFailureService do
   let(:file) do
     create_for_repository(:file_set, user: depositor)
   end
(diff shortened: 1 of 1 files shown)


spec/services/hyrax/import_url_failure_service_spec.rb
```diff
@@ -5,14 +5,15 @@
...
 
   before do
-    allow(file.errors).to receive(:full_messages).and_return(['huge mistake'])
+    allow(change_set.errors).to receive(:full_messages).and_return(['huge mistake'])
   end
 
   describe "#call" do
     before do
-      described_class.new(file, depositor).call
+      described_class.new(change_set, depositor).call
```
jcoyne: You are calling `ImportUrlFailureService` with a change set, but in `app/` it's called with a `FileSet`

kdid: @jcoyne - do the full error messages need to be in the email? If they are no longer in the file set, I guess I'm not sure where this service has access to them.

PR #2512, opened 2018-01-17
Change set multivalued fields initialize to empty array

cjcolvar:


spec/change_sets/hyrax/file_set_change_set_spec.rb
```diff
@@ -24,7 +24,7 @@
       change_set.prepopulate!
     end
     it 'initializes multivalued fields' do
-      expect(subject.title).to eq ['']
+      expect(subject.title).to eq []
```
jcoyne: I'm not sure that this is a good test anymore.  The reason we used to require one element is that the UI did an `change_set.title.each { input_tag ... }` so we either need to make sure we prepopulate with empty strings, or verify the UI works without that and then remove this test altogether.

PR #2515, opened 2018-01-17
Avoid create_content in import_url_job

jcoyne:


app/jobs/import_url_job.rb
```diff
@@ -41,6 +39,33 @@ def perform(file_set, operation)
...
+    # @param file_set [FileSet]
+    # @param file [File]
+    # @param user [User]
+    def create_content(file_set:, file:, user:)
+      # If the file set doesn't have a title or label assigned, set a default.
+
+      file_set.label ||= File.basename(Addressable::URI.parse(file_set.import_url).path)
+      file_set.title = [file_set.label] if file_set.title.blank?
+
+      # Need to save to get an id
```
cjcolvar: This comment seems to imply that the file set hasn't been persisted yet, but the file set passed in is from a query on line 27 so it must have already been persisted and thus have an id.  Is the comment incorrect?  If so can the return on line 53 be skipped?

spec/jobs/import_url_job_spec.rb
```diff
@@ -13,10 +17,10 @@
   end
 
   let(:operation) { create(:operation) }
-  let(:actor) { instance_double(Hyrax::Actors::FileSetActor, create_content: true) }
+  let(:actor) { instance_double(JobIoWrapper, ingest_file: true) }
```
cjcolvar: Is `actor` a misleading name for this variable now that it is `JobIoWrapper` and not an "actor"?

jcoyne: Yeah, good call.

app/jobs/import_url_job.rb
jcoyne: Uh.. yeah, I can fix that.

PR #2520, opened 2018-01-18
Fix spec/services/hyrax/statistics/file_sets/by_format_spec.rb

jgondron:
- Changed the rendering of file_format from index_mime_type to treat mime_type and format_label as arrays, since they will now be arrays in the indexed document. This change broke services/hyrax/file_set_csv_service_spec, so updated mock_file.mime_type to return an array.
- Added a delegate for format_label in FileSetBehavior since this attribute is handled by the associated FileNode, not the FileSet itself
- Added format_label when constructing a FileNode via the spec factory for FileSets
- Expanded the file_node_spec to test the handling of format_label
- Updated by_format_spec.rb to the new way of creating file node/set for specs.
- Expanded the Tika stubs to return output based on the file type. Unfortunately, all we have at this point is the file name, so have to base it on file extension. The fallback case still has to return the tiff output since many other specs still expect this, even for things like pdfs, wavs, etc
- Added 1x1 pixel files in each format just to have speedy files to test with.
- Updated the expected labels in by_format_spec to match what is given by the tika stubs.

@samvera/hyrax-code-reviewers


cjcolvar:
@jgondron Looks like there are rubocop issues.

jgondron:
@cjcolvar cop issues should be fixed

PR #2522, opened 2018-01-19
Controller that adds members to a collection

### LINKED ISSUE #2489: User cannot add subcollections or works to a collection for which they have deposit access.

elrayle:
Partially Fixes #2489

This allows members to be added to a collection from Dashboard -> Works -> batch Add Collection -> select a collection with deposit access.  You cannot use the regular controller for Dashboard -> Collections -> edit collection because it enforces edit access on the collection, as it should.  To be able to add a member, the user only needs deposit access.  The update member controller enforces read access on the members and deposit access on the collection.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/collections_membership_actor.rb
+++ b/app/actors/hyrax/actors/collections_membership_actor.rb
@@ -51,15 +51,15 @@ module Hyrax
         # Adds the item to the ordered members so that it displays in the items
         # along side the FileSets on the show page
         def add(env, id)
-          member = Collection.find(id)
-          return unless env.current_ability.can?(:deposit, member)
-          env.curation_concern.member_of_collections << member
+          collection = Collection.find(id)
+          return unless env.current_ability.can?(:deposit, collection)
+          env.curation_concern.member_of_collections << collection
         end
 
         # Remove the object from the members set and the ordered members list
         def remove(curation_concern, id)
-          member = Collection.find(id)
-          curation_concern.member_of_collections.delete(member)
+          collection = Collection.find(id)
+          curation_concern.member_of_collections.delete(collection)
         end
 
         # Determines if a hash contains a truthy _destroy key.
--- a/app/controllers/hyrax/dashboard/collection_members_controller.rb
+++ b/app/controllers/hyrax/dashboard/collection_members_controller.rb
@@ -1,0 +1,65 @@
+module Hyrax
+  module Dashboard
+    ## Shows a list of all collections to the admins
+    class CollectionMembersController < Hyrax::My::CollectionsController
+      before_action :filter_docs_with_read_access!
+
+      include Hyrax::Collections::AcceptsBatches
+
+      def after_update
+        respond_to do |format|
+          format.html { redirect_to success_return_path, notice: t('hyrax.dashboard.my.action.collection_update_success') }
+          format.json { render json: @collection, status: :updated, location: dashboard_collection_path(@collection) }
+        end
+      end
+
+      def after_update_error(err_msg)
+        respond_to do |format|
+          format.html { redirect_to err_return_path, alert: err_msg }
+          format.json { render json: @collection.errors, status: :unprocessable_entity }
+        end
+      end
+
+      def update_members
+        err_msg = validate
+        after_update_error(err_msg) if err_msg.present?
+        return if err_msg.present?
+
+        collection.add_member_objects batch_ids
+        after_update
+      end
+
+      private
+
+        def validate
+          return t('hyrax.dashboard.my.action.members_no_access') if batch_ids.blank?
+          return t('hyrax.dashboard.my.action.collection_deny_add_members') unless current_ability.can?(:deposit, collection)
+          return t('hyrax.dashboard.my.action.add_to_collection_only') unless member_action == "add" # should never happen
+        end
+
+        def success_return_path
+          dashboard_collection_path(collection_id)
+        end
+
+        def err_return_path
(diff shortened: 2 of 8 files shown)


app/actors/hyrax/actors/collections_membership_actor.rb
```diff
@@ -51,15 +51,15 @@ def assign_nested_attributes_for_collection(env, attributes_collection)
...
+          return unless env.current_ability.can?(:deposit, collection)
+          env.curation_concern.member_of_collections << collection
         end
 
         # Remove the object from the members set and the ordered members list
         def remove(curation_concern, id)
-          member = Collection.find(id)
-          curation_concern.member_of_collections.delete(member)
+          collection = Collection.find(id)
+          curation_concern.member_of_collections.delete(collection)
```
elrayle: Changed variable name because the variable holds the parent collection, not the member.

jeremyf: My understanding is this is an incidental (and welcome) renaming, but not related to the rest of this PR.

app/controllers/hyrax/dashboard/collection_members_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  module Dashboard
+    ## Shows a list of all collections to the admins
+    class CollectionMembersController < Hyrax::My::CollectionsController
+      before_action :filter_docs_with_read_access!
+
+      include Hyrax::Collections::AcceptsBatches
+
+      def after_update
+        respond_to do |format|
+          format.html { redirect_to success_return_path, notice: 'Collection was successfully updated.' }
```
jeremyf: I believe there should be an internationalization of the notice.

app/controllers/hyrax/dashboard/collection_members_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+        after_update_error(err_msg) if err_msg.present?
+        return if err_msg.present?
+
+        collection.add_member_objects batch_ids
+        after_update
+      end
+
+      private
+
+        def validate
```
jeremyf: My non-blocking expectation is that this behavior would be captured in a form object. However, I accept that that is not the case and am okay with proceeding as written.

app/actors/hyrax/actors/collections_membership_actor.rb
elrayle: Correct.

app/controllers/hyrax/dashboard/collection_members_controller.rb
elrayle: That was a copy and paste error.  So I fixed it in the new controller and in the collections controller that was the source of the error.

PR #2524, opened 2018-01-19
Get work action menu spec passing

kdid:
@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/views/hyrax/dashboard/collections/_work_action_menu.html.erb_spec.rb
+++ b/spec/views/hyrax/dashboard/collections/_work_action_menu.html.erb_spec.rb
@@ -2,11 +2,12 @@ require 'spec_helper'
 
 RSpec.describe 'hyrax/dashboard/collections/_work_action_menu.html.erb', type: :view do
   let(:item) { stub_model(GenericWork) }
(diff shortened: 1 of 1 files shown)


spec/views/hyrax/dashboard/collections/_work_action_menu.html.erb_spec.rb
```diff
@@ -2,11 +2,13 @@
 
 RSpec.describe 'hyrax/dashboard/collections/_work_action_menu.html.erb', type: :view do
   let(:item) { stub_model(GenericWork) }
-  let(:collection) { stub_model(Collection) }
+  let(:collection) { stub_model(Collection, id: 'xyz123z4', title: ["Make Collections Great Again"]) }
+  let(:change_set) { Hyrax::CollectionChangeSet.new(collection) }
 
   before do
     allow(view).to receive(:display_trophy_link)
     assign(:collection, collection)
```
jcoyne: I think you can remove this?

cjcolvar: Following on to Justin's comment do you need to assign the change set here instead of collection?

PR #2529, opened 2018-01-19
FileSet#mime_type is multivalued

cjcolvar:


jcoyne:
@cjcolvar , it seems like this change caused a number of tests to start failing that wern't failing before. Can you take a look?

cjcolvar:
@jcoyne I think I fixed them now.

Issue #2326, opened 2017-12-07
Draft 2.1.0 release notes


mjgiarlo:
# IIIF

Make clear in the release notes the requirement that Apache and Passenger require configuration to get RIIIF integration working. See https://github.com/curationexperts/riiif/blob/753ed1d73bc58d18124b247dc5a006fc60cb54a6/README.md#special-note-for-passenger-and-apache-users.

Also, write up the steps needed to use RIIIF when upgrading:

1. Copy [new config values](https://github.com/samvera/hyrax/blob/e2b2e9fd718f72c283b7165c93ccbc98a335a1ed/lib/generators/hyrax/templates/config/initializers/hyrax.rb#L114-L144) to `config/initializers/hyrax.rb`
2. Leave the new config values commented out if you won't be enabling the IIIF image server. You're done! Else, uncomment at least `config.iiif_image_server` and set it to true.
3. If using RIIIF, run `rails g hyrax:riiif`.
4. [Turn on public file server](https://github.com/pulibrary/pul_uv_rails/issues/8)
5. See more in-depth instructions here: https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#image-server

# Actor stack

If you've customized the actor stack or integrated with it in your code, note that the structure of collection members has changed. `member_of_collection_ids` was deprecated, to be replaced by `member_of_collections_attributes`:

```diff
- { 
-   member_of_collection_ids: ['123']
- }
+ {
+   member_of_collections_attributes: { '0' => { id: '123' } },
+ }
```

# Show View Attribute CSS

The CSS classes for individual attributes in the show view have been changed to use a prefix. This is to avoid collisions between attribute names and classes elsewhere in the codebase (e.g. "pagination"). 

```diff
- <li class='attribute title'>
+ <li class='attribute attribute-title'>
```

no-reply:
@mjgiarlo Is it okay to edit the second section of this, since we again support `member_of_collection_ids` env params?

mjgiarlo:
@no-reply Yes, please! Feel free to build out the release notes as you/we see fit in the description above. Thank you for your work on this.

mkorcy:
@mjgiarlo The IIIF notes you added for Apache + Passenger users are also true for the default AdminSet, loading `/admin/admin_sets/admin_set%2Fdefault` so those notes may be more generally applicable and worth noting outside the IIIF scope.

mjgiarlo:
[...]

PR #2533, opened 2018-01-20
Reinstate `:member_of_collection_id` argument option to Actor Stack

no-reply:
780c1489251c2d removed existing actor stack behavior. This reinstates it and deprecates it for removal in 3.0.0.

@samvera/hyrax-code-reviewers

The **Actor Stack** section of https://github.com/samvera/hyrax/issues/2326 can be edited to reflect a deprecation, rather than needed action after merge.

### DIFF
--- a/app/actors/hyrax/actors/collections_membership_actor.rb
+++ b/app/actors/hyrax/actors/collections_membership_actor.rb
@@ -12,28 +12,29 @@ module Hyrax
       # @param [Hyrax::Actors::Environment] env
       # @return [Boolean] true if create was successful
       def create(env)
-        attributes_collection = env.attributes.delete(:member_of_collections_attributes)
-        assign_nested_attributes_for_collection(env, attributes_collection) &&
-          next_actor.create(env)
+        assign_nested_attributes_for_collection(env) && next_actor.create(env)
       end
 
       # @param [Hyrax::Actors::Environment] env
       # @return [Boolean] true if update was successful
       def update(env)
-        attributes_collection = env.attributes.delete(:member_of_collections_attributes)
-        assign_nested_attributes_for_collection(env, attributes_collection) &&
-          next_actor.update(env)
+        assign_nested_attributes_for_collection(env) && next_actor.update(env)
       end
 
       private
 
+        ##
         # Attaches any unattached members.  Deletes those that are marked _delete
         # @param [Hash<Hash>] a collection of members
-        def assign_nested_attributes_for_collection(env, attributes_collection)
-          return true unless attributes_collection
+        # rubocop:disable Metrics/MethodLength
+        def assign_nested_attributes_for_collection(env)
+          attributes_collection ||= env.attributes.delete(:member_of_collections_attributes)
+          return assign_for_collection_ids(env) unless attributes_collection
+
+          emit_deprecation if env.attributes.delete(:member_of_collection_ids)
+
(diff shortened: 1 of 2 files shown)


app/actors/hyrax/actors/collections_membership_actor.rb
```diff
@@ -12,28 +12,29 @@ class CollectionsMembershipActor < AbstractActor
...
       end
 
       private
 
+        ##
         # Attaches any unattached members.  Deletes those that are marked _delete
         # @param [Hash<Hash>] a collection of members
-        def assign_nested_attributes_for_collection(env, attributes_collection)
-          return true unless attributes_collection
+        # rubocop:disable Metrics/MethodLength
```
no-reply: The rubocop disables explicitly support the call to `emit_deprecation`. When the dual behavior is removed this will handily pass all Rubocop's ABC and MethodLength checks.

app/actors/hyrax/actors/collections_membership_actor.rb
```diff
@@ -12,28 +12,29 @@ class CollectionsMembershipActor < AbstractActor
       # @param [Hyrax::Actors::Environment] env
       # @return [Boolean] true if create was successful
       def create(env)
-        attributes_collection = env.attributes.delete(:member_of_collections_attributes)
-        assign_nested_attributes_for_collection(env, attributes_collection) &&
-          next_actor.create(env)
+        assign_nested_attributes_for_collection(env) && next_actor.create(env)
```
elrayle: Moving the assignment of attributes_collection to be inside `assign_nested_attributes_for_collection(env) will impact collections-sprint which has this code...
```
def create(env)
        extract_collection_id(env, attributes_collection)
        assign_nested_attributes_for_collection(env, attributes_collection) &&
          next_actor.create(env)
end
```

elrayle: Note that `extract_collection_id(env, attributes_collection)` is not part of #update.

app/actors/hyrax/actors/collections_membership_actor.rb
```diff
@@ -12,28 +12,29 @@ class CollectionsMembershipActor < AbstractActor
...
+        ##
         # Attaches any unattached members.  Deletes those that are marked _delete
         # @param [Hash<Hash>] a collection of members
-        def assign_nested_attributes_for_collection(env, attributes_collection)
-          return true unless attributes_collection
+        # rubocop:disable Metrics/MethodLength
+        def assign_nested_attributes_for_collection(env)
+          attributes_collection ||= env.attributes.delete(:member_of_collections_attributes)
+          return assign_for_collection_ids(env) unless attributes_collection
+
```
elrayle: For collections-sprint, the collection id is needed in methods other than just `assign_nested_attributes_for_collection`.

no-reply:
It's not clear to me how the refactor presents a problem for the linked code. It would still have access to the passed in attributes until *later* in execution than in current master.

@elrayle, I'm blown away that you admit to blocking this PR without having read it and can't help but feel you are looking for reasons to block it now.

elrayle:
I see what you are suggesting.  The env parameters will still exist.  I was attempting to avoid having to process the parameters twice.  My intention of pointing you to the collections-sprint code was to provide context for my concerns.  It seemed to me easier to write the code here in a way that would be compatible with collections-sprint to avoid the need to re-work it on the next merge of master into collections-sprint.

no-reply:
@jcoyne has indicated that he won't have time to review this.

PR #2573, opened 2018-01-29
Collections #2487 consolidate modals

### LINKED ISSUE #2487: Collection index page - consolidate individual modals per row into dynamically populated modals
### Description
Current in the Collections index page, each collection row generates 3 modals in the markup, per row, which are identified by unique #id.  This could easily get out of hand with say 100 rows (300 modals).  The goals is to create 3 skeleton modals in terms of markup, and dynamically populate the necessary row data into the modal when the modal window is initiated.

Part of the collection extensions sprint.

### Rationale

This is an optimization.  It will drastically reduce the amount of markup rendered to the collections index page.

### Expected behavior

Fewer modals will be rendered in the markup, and dynamically populated by the content from the appropriate row interacted with.

### Actual behavior

Multiple modals are being generated per row.  Code duplication.

adamjarling:
Partially Fixes #2487 

This code changes modals from individually rendering with a unique `id` attribute (tied to the record `id`) to dynamically rendering.  So, instead of 3 modals rendering for each row in a collections listing table (ie. Add, Delete, Delete Deny, etc.), only 3 (an example) total modal windows would render to the entire page, and be dynamically updated with data applied through a click event a collection row's button.

This addresses the Add to Collection, and Delete buttons initially.


adamjarling:
@elrayle The issue in the comment above is that previously the translation string was handled on a case by case basis in each individual modal.  Now that the modal is generic, I haven't found a way to call the `t()` function from with a javascript file. 

There would need to be another way of plugging `type_name` into the description, that is Javascript based, I think.  

For now, would this modal only reference "Admin Set" instead of a dynamic `type_name`?  If so, I suggest eliminating the `type_name` param for the translation text, and create a new ticket to figure out how to incorporate more clear messaging in the modals, which now will probably be driven by Javascript.

That sound ok?

adamjarling:
The idea of adding `type_name` dynamically through Javascript, is putting `collection_presenter.collection_type.title` on the collections table row (similar to the other collection info) as a data attribute.  Then putting `<span>`s in the locale text, or some way to signify what should be replaced.

elrayle:
It is keyed off the `requires_membership` collection_type setting, which is currently only available to Admin Sets.  So this is an ok temporary solution.  Is there a good place to make a TODO note stating that it `requires_membership` is opened up to additional collection types, then it will need to be able to dynamically set the type.

Would it be possible to pass the collection type title as part of the data going to the modal when clicked?

adamjarling:
@elrayle Yes, correct.  The js can grab the data attribute of the collection type title from the clicked collection row, then put that value into the modal text.

PR #2574, opened 2018-01-29
Prevent circular nesting

### LINKED ISSUE #2221: Hyrax::Collections::NestedCollectionQueryService logic not adequate

laritakr:
Fixes #2221

1. The nested collection query service needs to remove collections from the lists for nesting if they would cause nested indexing errors. This updates the query service to exclude collections which would cause circular indexing errors. 
2. It also prevents nesting of collections which would exceed the total nesting depth. However, this logic is too expensive to justify being done on each potential collection, so the nesting depth is only checked for the selected collection, prior to making the requested connection.
3. Additionally this implements a modification to the collection factory which allows for creating a solr_document for a nestable collection for querying purposes, using the build(:collection) option, without requiring the full collection creation and nested indexing process. 

### The rules for allowed nesting between collections are as follows:

**if adding as a parent collection:**
1) cannot have any pathnames containing the collection's ID
2) cannot already be Collection G's direct parent
3) cannot cause total nesting depth to exceed limit

**if adding as a child (subcollection):**
1) Cannot have any pathnames containing any of the child's pathname or ancestors
2) cannot already be the collection's direct child
3) cannot cause total nesting depth to exceed limit

_Note:_ due to the wildcard search, etc, the queries ARE expensive (the inclusion of the ability to add a subcollection on the index page concerns me for this reason... ideally this should only happen when someone wants to add a subcollection, not for each collection as an index page is loaded). However, in general, the use of nesting is a minor portion of the code, and as this is a necessity, any changes should focus on refactoring its use on the index page. It's my understanding that https://github.com/samvera/hyrax/pull/2573 accomplishes this.
 
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/nest_collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/nest_collections_controller.rb
@@ -13,8 +13,7 @@ module Hyrax
           notice = I18n.t('create_within', scope: 'hyrax.dashboard.nest_collections_form', child_title: @form.child.title.first, parent_title: @form.parent.title.first)
           redirect_to redirect_path(item: @form.child), notice: notice
         else
-          flash[:error] = @form.errors.full_messages
-          redirect_to redirect_path(item: @form.child)
+          redirect_to redirect_path(item: @form.child), error: @form.errors.full_messages
         end
       end
 
@@ -24,8 +23,7 @@ module Hyrax
         if @form.validate_add
           redirect_to new_dashboard_collection_path(collection_type_id: @form.parent.collection_type.id, parent_id: @form.parent)
         else
-          flash[:error] = @form.errors.full_messages
-          redirect_to redirect_path(item: @form.parent)
+          redirect_to redirect_path(item: @form.parent), error: @form.errors.full_messages
         end
       end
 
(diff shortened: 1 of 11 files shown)


laritakr:
There is an issue remaining where the error messages resulting from the nesting are swallowed and not shown after the redirect. (I'm not sure yet if it happens on all views, but I did encounter it on the relationships tab). I will be writing this up as a separate issue.

PR #2580, opened 2018-01-30
Initialize AuthoritySelect in editor

### LINKED ISSUE #2579: AuthoritySelect fields not initialized
### Descriptive summary

In Hyrax 2.x AuthoritySelect fields are not being initialized.

### Rationale

Hyrax documentation suggests that one should be able to set up an authority select dropdown in their Hyrax based app.
https://github.com/samvera/hyrax/wiki/Autocomplete-in-Hyrax#activating-a-dropdown-with-authorities

Subsequent Hyrax commits look like AuthoritySelect should be automatically initialized on select fields with `data-authority-select` attributes, but this does not appear to be happening as AuthoritySelect is never initialized.

https://github.com/samvera/hyrax/blob/master/app/assets/javascripts/hyrax/editor.es6#L31 

### Expected behavior

Override a default Hyrax view partial - for example creator -  `/app/views/records/edit_fields/_creator.html.erb`

with 

```
<% subject_authorities = Hyrax::NamesAuthorities.new %>

<%=
  f.input key,
  as: :multi_value,
  input_html: {
    class: 'form-control',
    data: { 'autocomplete-url' => "/authorities/search/loc/names",
      'autocomplete' => key }
  } ,
  required: f.object.required?(key) %>

  <%= f.input key,
    input_html: {
      class: 'form-control image_subject',
      data: { 'authority-select' => "generic_work_creator" }
    },
    collection: names_authorities.select_active_options,
    label: false %>
```

And you would expect a dropdown which allows you to select authorities will switch the authority on the associated autocomplete field.

### Actual behavior

Autocomplete field works but AuthoritySelect dropdown does not change the authority.

kdid:
Fixes #2579 

AuthoritySelect should be automatically initialized on select fields with `data-authority-select` attributes, but this does not appear to be happening as AuthoritySelect is never initialized.

 - Initializes authoritySelect in editor
 - Removes editor argument that was being passed to AuthoritySelect as the editor was not needed and it was actually passing a select box instead anyways.
- called autocomplete setup correctly when select box value is changed or new field is added.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/authority_select.es6
+++ b/app/assets/javascripts/hyrax/authority_select.es6
@@ -1,3 +1,5 @@
+import Autocomplete from 'hyrax/autocomplete'
+
 /** Class for authority selection on an input field */
 export default class AuthoritySelect {
     /**
@@ -6,8 +8,7 @@ export default class AuthoritySelect {
      * @param {string} selectBox - The selector for the select box
      * @param {string} inputField - The selector for the input field
      */
-    constructor(editor, options) {
-      this.editor = editor
+    constructor(options) {
     	this.selectBox = options.selectBox
     	this.inputField = options.inputField
     	this.selectBoxChange();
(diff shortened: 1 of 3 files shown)


tpendragon:
@kdid Could you take a look at the ES6 style errors?

kdid:
@tpendragon - does houndci not run on existing code? I don't think I introduced this with what I changed. 

I tried simply removing `new` but it seems to cause other issues. Does it make sense to break out the work to refactor this class into a separate issue? Right now this functionality is not working.

PR #2585, opened 2018-01-30
Update search bar location and collection type label in collection public show page

### LINKED ISSUE #2163: Collection public show page search bar location implies untrue behavior
### Descriptive summary

On the collections sprint branch, the public collection show page has the search bar between the subcollection list and the works list. It should be located above both lists.

In addition, the header image was supposed to show collection type, but instead was changed to show "Collection". See the dashboard show page for more details.

### Rationale

The search behavior searches both works and subcollections. Unless this behavior is modified, the search bar should be above both. Having it located between the two lists implies that it only searches works. 

Also, collection type was the intended content to be shown on the header image.

Related to https://github.com/samvera/hyrax/pull/2072

### Tasks to complete in this issue:

- [x] Change label in header where it says "Collection" to instead show the collection type in the similar format. 
- [x] Move the search bar above the subcollections list since it searches both lists.

adamjarling:
Fixes #2163 

Move the search bar above subcollections in the UI, and update the label tag overlaying the image to read collection type title, rather than hardcoded "Collection".

### DIFF
--- a/app/views/hyrax/collections/show.html.erb
+++ b/app/views/hyrax/collections/show.html.erb
@@ -11,7 +11,7 @@
 
 			<div class="hyc-title">
 				<h1><%=@presenter.title.first%></h1>
-				<span class="label label-success">Collection</span>
+				<span class="label label-success"><%= @presenter.collection_type_badge %></span>
 				<%= @presenter.permission_badge %>
 			</div>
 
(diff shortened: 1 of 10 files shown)


app/views/hyrax/collections/show.html.erb
```diff
@@ -11,7 +11,7 @@
 
 			<div class="hyc-title">
 				<h1><%=@presenter.title.first%></h1>
-				<span class="label label-success">Collection</span>
+				<span class="label label-success"><%= @presenter.collection_type.title %></span>
```
laritakr: The presenter has a method "collection_type_badge" which does the same thing and is used elsewhere (i.e. the dashboard show page, index page, etc). It doesn't actually make a badge, simply the text for the badge. I'd like to see all of these locations do it consistently.

Also, at least some of the other locations do not have the "badge" format, simply the text. If not part of this issue, perhaps we need another issue to make consistent formatting among all collection type "badge" locations?

adamjarling: I just updated to use the `collection_type_badge` method in this file and re-pushed.   Maybe a new ticket is created to make these usage instances consistent in the application?

adamjarling:
@elrayle Note I added some other cleanup in the UI to make Lynette's latest header display size request consistent with other Collections show pages.   The scope of this ticket expanded somewhat.

elrayle:
Additional issues added based on feedback on this PR...
* Issue #2590 - Collections admin show page - search form results stay on same page
* Issue #2591 - Collections pages - update text sizing in list results and header to create consistency between views

PR #2595, opened 2018-02-01
Avoid potential infinite loop in `WorkShowPresenter`

no-reply:
The `WorkShowPresenter#representative_presenter` can enter an infinite loop in
the case that it returns another instance of itself as representative. This
behavior has been spotted in production shortly after upload of a new item--
presumably while background jobs are running--when visiting the work show page
immediately after deposit. This clears itself up shortly after, and doesn't
present a major issue. Still, it seemed worth addressing the edge case.

Note that an infinite loop is still technically possible within the method; we
believe this fixes the real case we have encountered.

* Don't recursively call `#representative_presenter` from itself.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -72,6 +72,7 @@ module Hyrax
       @representative_presenter ||=
         begin
           result = member_presenters([representative_id]).first
+          return nil if result.try(:id) == id
           if result.respond_to?(:representative_presenter)
             result.representative_presenter
           else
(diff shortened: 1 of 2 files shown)


no-reply:
Notably, no prior unit tests for this module fail if we just make the entire method body.

```ruby
@representative_presenter ||= member_presenters([representative_id]).first
```

I have a branch up at https://github.com/samvera/hyrax/tree/presenter-loop-alt to investigate if that is a better approach. I'm letting travis run the full suite to see if anything else fails.

no-reply:
@tpendragon do you have any thoughts about my comment at https://github.com/samvera/hyrax/pull/2595#issuecomment-362123680?

PR #2611, opened 2018-02-03
Use add users search to select a user.

### LINKED ISSUE #2578: Collections - edit form -> sharing tab allows any text as user
### Descriptive summary

The Add user search box allows any text as the user.  It should show 'No matches found.' when the text entered does not match a valid user.

### Rationale

Provide feedback to the user when they type in an invalid value.

### Expected behavior

NOTE: Check out how Edit Administrative Set -> Participants handles this.  The code should be the same for the most part.

#### Error 1:
When searching for a user on the Sharing tab that does not exist, a message is displayed stating 'No matches found.'  When the user clicks outside the search box, the text should go back to 'Search for a user...', such that the invalid user cannot be assigned a role.

#### Error 2:
Add button should not process if either the user or role is not selected.  A dialog should tell the user their error...  'Select a user and a role before adding access.'  There should be a similar check for group add button.

#### Error 1:
When searching for a user on the Sharing tab that does not exist, no message is displayed.  And you can add the invalid user to one of the roles.

#### Error 2:
If role is selected and user is blank, adds an entry for the role with a blank user.

#### Error 1:
1. Dashboard -> Collections -> action menu  (beside any collection you own) -> Edit collection -> Sharing tab
1. Type random text in box Search for a user...

Verify a message is displayed below the search box stating 'No matches found.'

#### Error 2: 
1. Dashboard -> Collections -> action menu  (beside any collection you own) -> Edit collection -> Sharing tab
1. Leave user and role blank and click Add

Verify a modal is displayed with message 'Select a user and a role before adding access.'

1. Dashboard -> Collections -> action menu  (beside any collection you own) -> Edit collection -> Sharing tab
1. Leave user blank, select a role, and click Add

Verify a modal is displayed with message 'Select a user and a role before adding access.'

1. Dashboard -> Collections -> action menu  (beside any collection you own) -> Edit collection -> Sharing tab
1. Select an existing user, leave role blank, and click Add

Verify a modal is displayed with message 'Select a user and a role before adding access.'

1. Dashboard -> Collections -> action menu  (beside any collection you own) -> Edit collection -> Sharing tab
1. Leave group and role blank and click Add

Verify a modal is displayed with message 'Select a group and a role before adding access.'

1. Dashboard -> Collections -> action menu  (beside any collection you own) -> Edit collection -> Sharing tab
1. Leave group blank, select a role, and click Add

Verify a modal is displayed with message 'Select a group and a role before adding access.'

### Related work

See widgets for adding participants on Edit Administrative Sets -> Participants tab.

elrayle:
Fixes #2578

Use the same javascript used by admin sets to select a user.  This prevents garbage from being entered as a user.  Non-existent users will cause 'No matches found' to be displayed to the user.

At this point, it still needs some UI adjustments to make the Add group and Add user section align and flow better.

### DIFF
--- a/app/views/hyrax/dashboard/collections/_form_share.html.erb
+++ b/app/views/hyrax/dashboard/collections/_form_share.html.erb
@@ -9,54 +9,58 @@
         <% access_options = options_for_select([['Manager', 'manage'], ['Depositor', 'deposit'], ['Viewer', 'view']]) %>
 
         <%= simple_form_for @form.permission_template,
-          url: [hyrax, :dashboard, @form, :permission_template],
-          html: { id: 'group-participants-form' } do |f| %>
+                            url: [hyrax, :dashboard, @form, :permission_template],
+                            html: { id: 'group-participants-form' } do |f| %>
 
           <%= f.fields_for 'access_grants_attributes',
-            f.object.access_grants.build(agent_type: 'group'),
-            index: 0 do |builder| %>
+                           f.object.access_grants.build(agent_type: 'group'),
+                           index: 0 do |builder| %>
 
             <div class="form-inline add-sharing-form">
-              <div class="form-group">
-                <label><%= t('.add_group') %>:</label>
+              <div class="form-group col-md-10 col-xs-8">
+                <label class="col-md-2 col-xs-4 control-label"><%= t('.add_group') %>:</label>
                 <%= builder.hidden_field :agent_type %>
-                <%= builder.text_field :agent_id, placeholder: "Search for a group...", class: 'form-control search-input' %>
+                <%= builder.text_field :agent_id,
+                                       placeholder: "Search for a group...",
+                                       class: 'form-control search-input' %>
                 as
-                <%= builder.select :access, access_options,
-                { prompt: "Select a role..." },
-                class: 'form-control' %>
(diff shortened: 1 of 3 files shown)


app/views/hyrax/dashboard/collections/_form_share.html.erb
```diff
@@ -9,54 +9,61 @@
...
+            <% end %>
+          </div>
         <% end %>
 
         <%= simple_form_for @form.permission_template,
-          url: [hyrax, :dashboard, @form, :permission_template],
-          html: { id: 'user-participants-form' } do |f| %>
+                            url: [hyrax, :dashboard, @form, :permission_template],
+                            html: { id: 'user-participants-form' } do |f| %>
+          <div class="clearfix spacer">
```
adamjarling: Are these `<div class="clearfix spacer">`s necessary?

adamjarling: Ah, @jgondron just saw your comment about adding some space between rows.  I get it.  If the Add group `<form>` was rendering properly, then could probably target that form class with some CSS to add some bottom margin/padding instead of adding extra markup to get the spacing, but can always revisit that later.

Issue #2545, opened 2018-01-22
Reconcile reinstating `:member_of_collection_id` argument in CE branch


elrayle:
### Descriptive summary

PR #2533 adds back `:member_of_collection_id` in addition to `member_of_collections_attributes`.  This allows the CollectionsMembershipActor to process either env parameter.  The same changes need to be completed in collections-sprint branch.
 
### Rationale

Support for both parameters, which carry the same information, allows for backward compatibility.

### Related work

PR #2533 Reinstate `:member_of_collection_id` argument option to Actor Stack
PR #2326 Draft 2.1.0 release notes

no-reply:
I have this in progress.

PR #2612, opened 2018-02-03
Reinstate `:member_of_collection_id` argument option to Actor Stack 

no-reply:
Refs #2545, #2533.

This merge includes some minor refactors of `#create`; including some changes to
the actor's private methods.

`#extract_collection_ids` and `#assign_nested_attributes_for_collection` now
accept only the `env` as an argument. `#assign_nested_attributes_for_collection`
ensures that the `env` is left in the correct state. This reduces the impact of
mutable state across methods.

Assigning the singleton `#collection_id` is still a separate mutation of
`env`. This may be necessary without a much more extensive refactor.

`#extract_collection_id` is updated to handle the old syntax, which
ensures that `#create` populates the `:collection_id` environment attribute in
the same way as it does for the new syntax. Likewise, `#valid_membership?` is
updated to accept the collection ids it should check; existing calls are changed
to pass the ids directly.

* Merge commit '089bff9e8f8100db6119f5' into collections-sprint
* Support new collections features for old collections_id actor syntax

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/collections_membership_actor.rb
+++ b/app/actors/hyrax/actors/collections_membership_actor.rb
@@ -39,7 +39,9 @@ module Hyrax
           return assign_for_collection_ids(env) unless attributes_collection
 
           emit_deprecation if env.attributes.delete(:member_of_collection_ids)
-          return false unless valid_membership?(env, attributes_collection)
+
+          return false unless
+            valid_membership?(env, collection_ids: attributes_collection.map { |_, attributes| attributes['id'] })
 
           attributes_collection = attributes_collection.sort_by { |i, _| i.to_i }.map { |_, attributes| attributes }
           # checking for existing works to avoid rewriting/loading works that are already attached
@@ -70,6 +72,8 @@ module Hyrax
         # @deprecated supports old :member_of_collection_ids arguments
         def assign_for_collection_ids(env)
           collection_ids = env.attributes.delete(:member_of_collection_ids)
+
+          return false unless valid_membership?(env, collection_ids: collection_ids)
 
           if collection_ids
             Deprecation.warn(self, ':member_of_collection_ids has been deprecated for removal in Hyrax 3.0. ' \
@@ -133,11 +137,17 @@ module Hyrax
           attributes_collection =
             env.attributes.fetch(:member_of_collections_attributes) { nil }
 
-          # Determine if the work is being created in one and only one collection.
-          return unless attributes_collection && attributes_collection.size == 1
+          if attributes_collection
+            # Determine if the work is being created in one and only one collection.
+            return unless attributes_collection && attributes_collection.size == 1
 
-          # Extract the collection id from attributes_collection,
-          collection_id = attributes_collection.first.second['id']
+            # Extract the collection id from attributes_collection,
+            collection_id = attributes_collection.first.second['id']
+          else
+            collection_ids = env.attributes.fetch(:member_of_collection_ids) { [] }
+            return unless collection_ids.size == 1
+            collection_id = collection_ids.first
+          end
 
           # Do not apply permissions to work if collection type is configured not to
           collection = ::Collection.find(collection_id)
@@ -147,8 +157,7 @@ module Hyrax
           env.attributes[:collection_id] = collection_id
         end
 
-        def valid_membership?(env, attributes_collection)
-          collection_ids = attributes_collection.map { |_, attributes| attributes['id'] }
+        def valid_membership?(env, collection_ids:)
           multiple_memberships = Hyrax::MultipleMembershipChecker.new(item: env.curation_concern).check(collection_ids: collection_ids)
           if multiple_memberships
             env.curation_concern.errors.add(:collections, multiple_memberships)
--- a/spec/actors/hyrax/actors/collections_membership_actor_spec.rb
+++ b/spec/actors/hyrax/actors/collections_membership_actor_spec.rb
@@ -229,6 +229,105 @@ RSpec.describe Hyrax::Actors::CollectionsMembershipActor do
     end
   end
 
+  context 'when old style env parameter' do
+    describe 'create' do
+      let(:collection) { create(:collection) }
+      let(:attributes) do
+        { member_of_collection_ids: [collection.id], title: ['test'] }
+      end
+
+      it 'adds it to the collection' do
+        expect(subject.create(env)).to be true
+        expect(collection.reload.member_objects).to eq [curation_concern]
+      end
+
+      context 'when multiple membership checker returns a non-nil value' do
+        before do
+          allow(Hyrax::MultipleMembershipChecker).to receive(:new).and_return(checker)
+          allow(checker).to receive(:check).and_return(error_message)
+        end
+
+        let(:checker) { double('checker') }
+        let(:error_message) { 'Error: foo bar' }
+
+        it 'adds an error and returns false' do
+          expect(env.curation_concern.errors).to receive(:add).with(:collections, error_message)
+          expect(subject.create(env)).to be false
+          expect(curation_concern.member_of_collections).to be_empty
+        end
+      end
+
+      context "when work is in user's own collection" do
+        let(:collection) { create(:collection, user: user, title: ['A good title']) }
+        let(:attributes) { { member_of_collection_ids: [] } }
+
+        before do
+          subject.create(Hyrax::Actors::Environment.new(curation_concern, ability,
+                                                        member_of_collection_ids: [collection.id], title: ['test']))
+        end
+
+        it "removes the work from that collection" do
+          expect(subject.create(env)).to be true
+          expect(curation_concern.member_of_collections).to eq []
+        end
+      end
+
+      context "when work is in another user's collection" do
+        let(:other_user) { create(:user) }
+        let(:collection) { create(:collection, user: other_user, title: ['A good title']) }
+
+        it "doesn't remove the work from that collection" do
+          subject.create(env)
+          expect(subject.create(env)).to be true
+          expect(curation_concern.member_of_collections).to eq [collection]
+        end
+      end
+
+      context "updates env" do
+        let(:collection) do
+          create(:collection, collection_type_settings: [:discoverable, :sharable])
+        end
+
+        subject(:middleware) do
+          stack = ActionDispatch::MiddlewareStack.new.tap do |middleware|
+            middleware.use described_class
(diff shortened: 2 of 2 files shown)


app/actors/hyrax/actors/collections_membership_actor.rb
```diff
@@ -75,19 +121,18 @@ def has_destroy_flag?(hash)
...
         # Given an array of collection_attributes when it is size:
         # * 0 do not set `env.attributes[:collection_id]`
         # * 1 set `env.attributes[:collection_id]` to the one and only one collection
         # * 2+ do not set `env.attributes[:collection_id]`
         #
         # NOTE: Only called from create.  All collections are being added as parents of a work.  None are being removed.
-        def extract_collection_id(env, attributes_collection)
+        def extract_collection_id(env)
+          attributes_collection =
+            env.attributes.fetch(:member_of_collections_attributes) { nil }
```
elrayle: This needs to be able to extract the id from the new :member_of_collections_attributes OR from the old style :member_of_collection_ids.  

See diffs for method extract_collection_id (line 84) in the PR that switched from the old style :member_of_collection_ids to the new :member_of_collections_attributes...  https://github.com/samvera/hyrax/commit/85591cafa9e9935d18fd3b07cf86fb35cb9a73d4#diff-dcaa34b138208d26b5cc4a005542ad71

spec/actors/hyrax/actors/collections_membership_actor_spec.rb
```diff
@@ -100,6 +100,52 @@
...
+            Hyrax::VERSION.split('.').first.to_i >= 3
+
+          expect(subject.create(env)).to be true
+          expect(curation_concern.member_of_collections)
+            .to contain_exactly(collection, my_collection)
+        end
+      end
+    end
+
     context "updates env" do
```
elrayle: Need to also update the tests under `context "updated env"` to test that the update to env happens for both :member_of_collections_attributes and :member_of_collection_ids parameters.

Effected tests:  https://github.com/samvera/hyrax/blob/9c5fa41dd2e161ac70421aae573e0dea3c7a8e77/spec/actors/hyrax/actors/collections_membership_actor_spec.rb#L149-L203

elrayle: You can copy in the tests that use the old style parameter from here...
https://github.com/samvera/hyrax/blob/d67f598e992bcf5897c0731db1551ca8698cb04e/spec/actors/hyrax/actors/collections_membership_actor_spec.rb#L89-L129

no-reply:
@elrayle See the questions in the PR description.

It's not clear to me what the best way to handle the new behavior for `member_of_collection_ids`. In particular, adding the `#valid_membership?` check would be breaking for DCE's current use cases, which rather defeats the purpose.

Some additional context on the motivation for the `#valid_membership?` behavior, and how `collection_id` is intended to be used would help me determine the best implementation for the old syntax.

no-reply:
It might help as well if the `collection_id` attribute were named more explicitly. What behavior would I expect as a caller to the actor stack if I pass such an id directly, or extract it in another actor, either before or after this?

elrayle:
Questions - answered

> Do we need to handle extraction of the #collection_id attribute in the #member_of_collection_ids case?

Yes.  It was handled before the change was made to member_of_collection_ids.  You can see the code that processed it prior to the change at https://github.com/samvera/hyrax/blob/d67f598e992bcf5897c0731db1551ca8698cb04e/app/actors/hyrax/actors/collections_membership_actor.rb#L40-L51

This older code does not include additional changes that are seen in the current code that need to be handled now for both parameters.

> Does multiple membership need to be treated as invalid under #member_of_collection_ids, as well as in the new attributes syntax?

Yes.  See description of multiple membership below next question.

> Should we be concerned that the current head on collection-sprint prevents adding multiple collections at the actor level? What is the reasoning for this restriction? Am I misunderstanding #valid_membership?

There is `multiple_membership` configuration for collection types.  #valid_membership will only deny multiple collections if that configuration is false.  The variable could probably be better named `multiple_membership_denied`.  `Hyrax::MultipleMembershipChecker` will only return a value if multiple collections are being added AND `multiple_membership=false`.
https://github.com/samvera/hyrax/blob/9c5fa41dd2e161ac70421aae573e0dea3c7a8e77/app/services/hyrax/multiple_membership_checker.rb#L33

If you look at that service, you will see how the test is applied.  The multiple membership test only applies to collections of the same type.

> It might help as well if the collection_id attribute were named more explicitly. What behavior would I expect as a caller to the actor stack if I pass such an id directly, or extract it in another actor, either before or after this?

Because this actor deletes the env parameter, the collection that effects permissions has to be added into the env.  To avoid confusion, a different name was given.  The `collection_id` parameter is processed by a later actor in the stack to apply permissions... 
https://github.com/samvera/hyrax/blob/collections-sprint/app/actors/hyrax/actors/apply_permission_template_actor.rb#L26-L31

no-reply:
I think the info you've given is enough for me to finish this work up.

I'm still of the opinion that the current PR supports the status quo on both branches, and can be safely merged. Though perhaps the ticket needs to be left open to track the remaining compatibility work. If you disagree, I can wrap this up first thing in the morning and we can move forward.

----

> Because this actor deletes the env parameter, the collection that effects permissions has to be added into the env. To avoid confusion, a different name was given. The collection_id parameter is processed by a later actor in the stack to apply permissions...

Would it be more accurate to rename it `permission_template_source` or similar? I'm concerned a bit about how tightly coupled these two actors are---I know this kind of thing is endemic in the actor stack, and I'm not too interested in wringing my hands about it, but some explict naming would help a lot, I think.

In either case, I think this is a separate issue. If it seems worth further discussion to you, I can open it after merge.

elrayle:
@no-reply Would you consider a simpler change that would ensure that all functionality is properly applied regardless of which parameter format is used.

You can replace line 15 and 24...
```
```
with a method call that extracts the values from env...
```
attributes_collection = process_env_attribute(env)
```
and add method...
```
# Extract env attribute that specifies collection membership processing.  There are two supported formats:
# @example a collections attribute hash - New style that allows for adding and deleting collection membership.
#   'member_of_collections_attributes' => {
#     '0' => { 'id' = '12312412'},
#     '1' => { 'id' = '234234234'},
#     '2' => { 'id' = '99981228', '_destroy' => 'true' }
#   }
#
# @example a collection ids hash - Old style that allows for adding collection membership.  This will be supported through Hyrax 3.0 for backward compatibility.
#   'member_of_collection_ids' => { '12312412', '234234234' }
def process_env_attribute(env)
  # return new env attribute approach if it exists
  return env.attributes.delete(:member_of_collections_attributes) if env.attributes.key? :member_of_collections_attributes

  # convert old style env attribute into new style supporting backward compatibility
  return nil unless env.attributes.key? :member_of_collection_ids
  collection_ids = env.attributes.delete(:member_of_collection_ids)
  return nil if collection_ids.blank?
  attributes = {}
  0.upto(collection_ids.size-1) { |i| attributes[i.to_s] = { 'id' => collection_ids[i] }
  attributes
end
```

elrayle:
I make this suggestion, because on first review, I missed that you did not include all the new functionality that was put in place in the [#assign_collections method](https://github.com/samvera/hyrax/blob/d67f598e992bcf5897c0731db1551ca8698cb04e/app/actors/hyrax/actors/collections_membership_actor.rb#L22-L34) as written before the switch to the new style env parameter.  I am hesitant to see the processing of each style parameter happen completely separate because it opens up the possibility of something getting missed and causes duplicate code.

app/actors/hyrax/actors/collections_membership_actor.rb
```diff
@@ -46,6 +52,46 @@ def assign_nested_attributes_for_collection(env, attributes_collection)
...
+        def emit_deprecation
+          Deprecation.warn(self, ':member_of_collections_attributes and :member_of_collection_ids were both ' \
+                                 ' passed. :member_of_collection_ids is ignored when both are passed and is ' \
+                                 'deprecated for removal in Hyrax 3.0.')
+        end
+
+        ##
+        # @deprecated supports old :member_of_collection_ids arguments
+        def assign_for_collection_ids(env)
+          collection_ids = env.attributes.delete(:member_of_collection_ids)
```
elrayle: The assign_for_collection_ids before the switch to the new style env parameter was...
```
        def assign_collections(env, collection_ids)
          return true unless collection_ids
          multiple_memberships = Hyrax::MultipleMembershipChecker.new(item: env.curation_concern).check(collection_ids: collection_ids)
          if multiple_memberships
            env.curation_concern.errors.add(:collections, multiple_memberships)
            return false
          end
          # grab/save collections this user has no edit access to
          other_collections = collections_without_edit_access(env)
          env.curation_concern.member_of_collections = ::Collection.find(collection_ids)
          env.curation_concern.member_of_collections.concat other_collections
end
```

I don't have a problem with the addition of the exception, but this method needs to include the check for multiple members.

elrayle:
Seems like there are minimally two pieces of work for this PR to be merged:
* add in the multi_membership check for the old style env parameter and 
* copy in most of the tests that were in place before the env parameter change from https://github.com/samvera/hyrax/blob/d67f598e992bcf5897c0731db1551ca8698cb04e/spec/actors/hyrax/actors/collections_membership_actor_spec.rb#L36-L129

no-reply:
@elrayle I've force pushed with the requested changes.

I think you will probably want some more refactoring here (I certainly do), but I'm inclined (again) to push forward with this and take on the remainder after the collections merge. I've also run into various potential test suite issues, where stubs are being set incorrectly or ineffectually, which has me questioning how effective the unit tests on this class are. I would like to address any refactors and test suite fixes in follow-up work.

no-reply:
> Need to open a new issue to revisit difference in processing for the two style env parameters. We should be able to reformat the old style into the new style and use a single approach for processing. If this isn't sufficient, then we need to understand why.

I honestly think I prefer the branched logic to increased mutations of `env`. But I do agree that there are more refactors to be had here. My goals would be to:

no-reply:
The biggest issue I see in the current code is that both `#extract_collection_id` and `#assign_nested_attributes_for_collection` need to read the same information from `env`. I have a notion to  add a flag parameter to `#assign_nested_attributes_for_collection` like `extract_permission_source_id` (is this an informative name for what `extract_collection_id` does?). This way we can extract the data once, then pass it along to `#extract_collection_id`. Something like:

PR #2622, opened 2018-02-06
Don't render required form fields that are not terms in the form

### LINKED ISSUE #2569: Hyrax form renders required field that is not a term
I can add a field to the list of `required_fields` in a Hyrax form without adding it to the list of `terms`.  Hyrax will render the field but not persist the data.

I would expect Hyrax to raise an exception if a field is defined as required but not in the list of terms.

no-reply:
Adding a `required_field` to a form class used to render an errant primary term
above the fold in the on-screen form. The data from that field was silently
ignored. We now log an warn level error and decline to render the non-term form
field.

Fixes #2569.

* decline to render fields that aren't terms in a form.
* log warnings when implementers obviously misconfigure form terms & required fields

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/forms/hyrax/forms/work_form_spec.rb
+++ b/spec/forms/hyrax/forms/work_form_spec.rb
@@ -185,6 +185,45 @@ RSpec.describe Hyrax::Forms::WorkForm do
     it { is_expected.to eq 'restricted' }
   end
 
+  describe '#primary_terms' do
+    it 'contains the required fields' do
+      expect(form.primary_terms).to include(*form.required_fields)
+    end
+
+    context 'with a field that is not in terms' do
+      let(:bad_term) { :BadWorkFormSpecTerm }
+
+      before { form.class.required_fields += [bad_term] }
+      after  { form.class.required_fields -= [bad_term] }
+
+      it 'logs a warning' do
+        expect(Rails.logger).to receive(:warn).with(/#{bad_term}/)
+        form.primary_terms
+      end
+
+      it 'does not include the errant term' do
+        expect(form.primary_terms).not_to include bad_term
+      end
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


spec/forms/hyrax/forms/work_form_spec.rb
```diff
@@ -185,6 +185,43 @@
...
+      end
+    end
+  end
+
+  describe '#secondary_terms' do
+    it 'does not contain the primary terms' do
+      expect(form.secondary_terms).not_to include(*form.primary_terms)
+    end
+
+    context 'with an new non-primary term' do
```
cjcolvar: Small typo.  This should probably be `with a new`.

cjcolvar:
The travis failures looks like you need to clean up after the tests by removing the term in an after block.

no-reply:
Thanks @cjcolvar.

I fixed the typo and did the spec cleanup. I think CI should be happy now.

spec/forms/hyrax/forms/work_form_spec.rb
```diff
@@ -185,6 +185,45 @@
...
+  describe '#secondary_terms' do
+    it 'does not contain the primary terms' do
+      expect(form.secondary_terms).not_to include(*form.primary_terms)
+    end
+
+    context 'with a new non-primary term' do
+      let(:new_term) { :WorkFormSpecTerm }
+
+      before { form.class.terms += [new_term] }
+      after  { form.class.terms += [new_term] }
```
cjcolvar: Interesting that travis didn't catch this typo: the after hook should be `-=`.

no-reply: :joy: I don't think the suite ever checks what's in `terms`. Thanks for your good eye. :eyes:

PR #2627, opened 2018-02-06
Prefix attribute classes with `attribute-`

### LINKED ISSUE #2567: Namespace field classes
### Descriptive summary

In the 'show' page HTML, fields are laid out like so:

```
<tr>
  <th>Pages</th>
  <td>
    <ul class="tabular">
      <li class="attribute pagination">135</li>
    </ul>
    </td>
  </tr>
```

The `<li>` has a class of the field/attribute name.

In the example above, there is already a `pagination` class (in bootstrap probably) so the display of this field is different to other fields.

It strikes me that this could easily occur with other fields, leading to the need to either change property names or put in unnecessary bespoke css.

Namespacing the field classes, simply by prefacing the class with something (like attribute_name_pagination) would prevent this from happening.

mbklein:
Fixes #2567

* Namespace attribute-specific CSS classes with `attribute-`
* Update test expectations accordingly

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/renderers/hyrax/renderers/attribute_renderer.rb
+++ b/app/renderers/hyrax/renderers/attribute_renderer.rb
@@ -25,7 +25,7 @@ module Hyrax
 
         return markup if values.blank? && !options[:include_empty]
         markup << %(<tr><th>#{label}</th>\n<td><ul class='tabular'>)
-        attributes = microdata_object_attributes(field).merge(class: "attribute #{field}")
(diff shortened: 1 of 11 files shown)


no-reply:
@mbklein Does this need a note in the [Release Notes](https://github.com/samvera/hyrax/issues/2326)?

I took a first pass, but feel free to fix up.

Issue #2372, opened 2017-12-14
ActiveFedora::Base.reindex_everything loses all nested indexing fields in solr


laritakr:
### Descriptive summary

In Collections Sprint branch, nested indexer adds **"nesting_collection__ancestors_ssim"**,
**"nesting_collection__parent_ids_ssim"**, and **"nesting_collection__pathnames_ssim"**. Running 
`ActiveFedora::Base.reindex_everything` from the rails console removed all three of those fields from solr. 

### Rationale

The nested collection indexer is triggered as an after update_index action in Hyrax. Running the Active Fedora reindexer is indexing separately from the nested collections indexer and doesn't trigger the necessary action.

### Expected behavior

Running a full reindex needs to reindex all nested collections as well.

### Actual behavior

The nested collection indexing is not being included.

### Steps to reproduce the behavior

1. Create nested collections.
1. Solr indexes should include the three nesting collection fields.
1. Open rails console
1. ActiveFedora::Base.reindex_everything
1. Solr indexes are missing all three nesting collection fields.

laritakr:
The issue here is that ActiveFedora uses to_solr for its reindexing. For nested collections, to_solr does not include the additional nested collections indexed fields. Those are added in Hyrax as an after_update_index callback which reindexes the collection again.

Methods `each_preservation_document` and `each_perservation_document_id_and_parent_ids` are not implemented in Hyrax::Adapters::NestingIndexAdapter so Samvera::NestingIndexer::RepositoryReindexer will not currently function.

In the meantime, if each TOP level nested collection is manually reindexed, it will reindex all of the child collections.

elrayle:
@laritakr Can this be closed with the merge of PR #2636?

laritakr:
Resolved via PR #2636

PR #2636, opened 2018-02-09
Reindex everything

laritakr:
fixes https://github.com/samvera/hyrax/issues/2372

Implements #each_perservation_document_id_and_parent_ids in the
nesting index adapter to allow for Samvera::NestingIndexer.reindex_all!
Objects which include collection nesting as part of the model will be
reindexed via Samvera::NestingIndexer, while still using to_solr for
objects where collection nesting does not apply.

Adds an override of ActiveFedora's reindex_everything method, which will
now use Samvera Nesting Indexer's reindex_all!

Fixes a bug in calculating the nesting depth that was encountered when
testing reindexing triggered a situation which exceeded the maximum.

@samvera/hyrax-code-reviewers


jeremyf:
👍 - holding off on Approval as I'd like a non-Notre Dame developer to have a chance at review.

Issue #1567, opened 2017-08-24
Warn the user when works could not be added to a single-membership collection


mjgiarlo:
Currently there are two code paths that add works to collections: one goes through the `WorksControllerBehavior` and the other goes through the `CollectionBehavior`. The latter does all of its membership checking down in the model, with no shared context w/ the controller, so currently if the membership checker balks, there's no way to surface errors or warnings. We should warn users if a work is not added to a collection they have specified.

See:

https://github.com/samvera/hyrax/blob/2306ca16c6ad8bedf628310a097eddd8cedbed82/app/models/concerns/hyrax/collection_behavior.rb#L48-L52

mjgiarlo:
@elrayle :point_up:

blancoj:
I have done some work on this, but am not sure my approach is correct. @mjgiarlo suggested I use  `collection.errors.add(:collections, error_messages) if error_messages.any?` instead of `flash[:notice] = error_messages.join(" ") if error_messages.any?`, but I  I did not see the message show up.  This is where the message gets sent out:  https://github.com/samvera/hyrax/blob/coll-1567/app/controllers/hyrax/dashboard/collections_controller.rb#L138

If I use `flash[:notice] = error_messages.join(" ") if error_messages.any?` I see the message.

I have a branch of the work I've done: coll-1567.

elrayle:
Note: This QA test fails to prevent a work from being added to two collections of the same Single Membership collection type.

Prereq: Create 2 collections of type Single-Membership C
* From Dashboard -> Works, check box beside any work not already in the two prereq collections
* Click Add to collection button
* Choose one of the two prereq collections
* Verify the work was added to the collection
* From Dashboard -> Works, check box beside several works including the same work just added to the first collection
* Click Add to collection button
* Choose the other prereq collection
* Verify the work added to the first was NOT added to second collection and a flash message tells the user why it was not added
* Verify the other works that are not members of the first collection were added to the second collection

elrayle:
Fixed by PR #2640 and PR #2687

mcwhitaker:
@elrayle -- In the current Hyrax QA testing scenarios, (see https://docs.google.com/spreadsheets/d/1XxG00RCkypCqLjfzx2RnoKLbZdaMxXMbW8A4hNHRrc0/edit#gid=1192600172 test CM_1.6) there is a question:

> Question: Should error message show when Add button is clicked OR when Save changes button is clicked.

[...]

PR #2640, opened 2018-02-12
Pass an array of collection type gid strings to where query

cjcolvar:
The current joined string was causing solr to always return an empty set leading to a positive check instead of an error.  This PR reworks the tests to check the params of Collection#where.

Refs #1567.  Rest of work for #1567 will be in a future PR.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/multiple_membership_checker_spec.rb
+++ b/spec/services/hyrax/multiple_membership_checker_spec.rb
@@ -101,6 +101,50 @@ RSpec.describe Hyrax::MultipleMembershipChecker, :clean_repo do
         expect(Collection).to receive(:where).with(id: [collection2.id], collection_type_gid_ssim: [collection_type.gid]).once.and_return([collection2])
         expect(subject).to eq 'Error: You have specified more than one of the same single-membership collection types: Greedy (Foo and Bar)'
       end
+
+      context 'with multiple single membership collection types' do
+        let!(:collection_type_2) { create(:collection_type, title: 'Doc', allow_multiple_membership: false) }
+        let(:collection_types) { [collection_type.gid, collection_type_2.gid] }
+
+        it 'returns an error' do
+          expect(item).to receive(:member_of_collection_ids)
+          expect(Collection).to receive(:where).with(id: collection_ids, collection_type_gid_ssim: collection_types).once.and_call_original
+          expect(Collection).to receive(:where).with(id: [collection2.id], collection_type_gid_ssim: collection_types).once.and_call_original
+          expect(subject).to eq 'Error: You have specified more than one of the same single-membership collection types: Greedy (Foo and Bar)'
+        end
+      end
+    end
+
+    context 'when multiple single-membership collection instances are in the list, but are different collection types' do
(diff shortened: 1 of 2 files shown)


spec/services/hyrax/multiple_membership_checker_spec.rb
```diff
@@ -40,64 +40,65 @@
...
+
+      context 'with multiple single membership collection types' do
+        let!(:collection_type_2) { create(:collection_type, title: 'Doc', allow_multiple_membership: false) }
+        let(:collection_types) { [collection_type.gid, collection_type_2.gid] }
+
+        it 'returns an error' do
+          expect(item).not_to receive(:member_of_collection_ids)
+          expect(checker).to receive(:single_membership_collections).with(collection_ids).once.and_call_original
+          expect(Collection).to receive(:where).with(id: collection_ids, collection_type_gid_ssim: collection_types).once.and_call_original
+          expect(subject).to eq 'Error: You have specified more than one of the same single-membership collection types: Greedy (Foo and Bar)'
```
elrayle: Would you mind rewording this to...
```
Error: You have specified more than one collection of the same single-membership collection type: Greedy
```
I wonder about including the collection names.  The tested collections have very short names Foo and Bar.  But typical collections have longer names.

spec/services/hyrax/multiple_membership_checker_spec.rb
```diff
@@ -40,64 +40,65 @@
...
-      it 'returns nil' do
+      it 'returns an error' do
         expect(item).to receive(:member_of_collection_ids)
-        expect(Collection).to receive(:where).with(id: collection_ids, collection_type_gid_ssim: collection_type.gid).once.and_return(collections)
-        expect(Collection).to receive(:where).with(id: [collection2.id], collection_type_gid_ssim: collection_type.gid).once.and_return([collection2])
+        expect(Collection).to receive(:where).with(id: collection_ids, collection_type_gid_ssim: [collection_type.gid]).once.and_return(collections)
+        expect(Collection).to receive(:where).with(id: [collection2.id], collection_type_gid_ssim: [collection_type.gid]).once.and_return([collection2])
         expect(subject).to eq 'Error: You have specified more than one of the same single-membership collection types: Greedy (Foo and Bar)'
       end
     end
```
elrayle: I would like to see a test case:
- col_type_1 is single_membership
- col_type_2 is single_membership
- col_1 is of type col_type_1
- col_2 is of type col_type_2
- checker allows both to be added since they are not of the same collection type

spec/services/hyrax/multiple_membership_checker_spec.rb
cjcolvar: I think this should be a separate issue since it's outside of the scope of this PR.  The changes to the tests were just refactoring and fixing and not any functional change.

spec/services/hyrax/multiple_membership_checker_spec.rb
cjcolvar: :+1: I can add this.

PR #2646, opened 2018-02-13
Don't try to clean up temp directories and files that may still be in use

### LINKED ISSUE #2645: Race Condition in ImportUrlJob deletes temp files in use
### Descriptive summary

This fix prevents two potential file deletion race conditions at the expense of leaving temp files for out-of-process cleanup.

### Rationale

`ImportUrlJob` contains the following:

```ruby
      Dir.mktmpdir do |dir|
        File.open(File.join(dir, filename), 'wb') do |f|
          retriever = BrowseEverything::Retriever.new
          retriever.retrieve('url' => uri) do |chunk|
            f.write(chunk)
          end
          f.rewind
          yield f
          # Other jobs (e.g., file characterization and derivative creation)
          # get kicked off asynchronously here
        end
      end
```

Because `dir` gets deleted when it goes out of scope, the file characterization and derivative jobs may or may not get a shot at the temp file before it disappears. Even worse, if the working directory is on an NFS mount, the directory removal will fail if another job manages to open the file before `ImportUrlJob` deletes it. This throws an exception, causing the job to fail and get re-queued, even though the rest of the import/ingest/attach process may have completed successfully.

There are two possible solutions I can think of:

1. Just use the standalone, non-block form of `dir = Dir.mktmpdir`, and accept that temporary files and directories will be left behind to clean up with a cron job. **Pro:** Quick & easy to implement. **Con:** Requires app developers and/or local devops to come up with their own cleanup strategies.
2. Create a `TempFileCleanupActor` that cleans up after a given set of actors and jobs. **Pro:** Cleans up after itself. **Con:** Requires a bunch of retooling to get every FileSet and job to keep track of the temp files they create along the way.

mbklein:
Fixes #2645. See issue description for full explanation.

This PR implements the first (easy) suggested fix.

* Change the `Dir.mktmpdir` call in `ImportUrlJob` from block form to bare form
* Add a little debug-level logging code to make it easier to spot remote retrieval

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -49,15 +49,16 @@ class ImportUrlJob < Hyrax::ApplicationJob
     # @yield [IO] the stream to write to
     def copy_remote_file(uri)
       filename = File.basename(uri.path)
-      Dir.mktmpdir do |dir|
-        File.open(File.join(dir, filename), 'wb') do |f|
-          retriever = BrowseEverything::Retriever.new
-          retriever.retrieve('url' => uri) do |chunk|
-            f.write(chunk)
-          end
-          f.rewind
-          yield f
+      dir = Dir.mktmpdir
+      Rails.logger.debug("ImportUrlJob: Copying <#{uri}> to #{dir}")
+      File.open(File.join(dir, filename), 'wb') do |f|
+        retriever = BrowseEverything::Retriever.new
+        retriever.retrieve('url' => uri) do |chunk|
+          f.write(chunk)
         end
+        f.rewind
+        yield f
       end
+      Rails.logger.debug("ImportUrlJob: Closing #{File.join(dir, filename)}")
     end
 end
--- a/spec/jobs/import_url_job_spec.rb
+++ b/spec/jobs/import_url_job_spec.rb
@@ -29,15 +29,29 @@ RSpec.describe ImportUrlJob do
   end
 
   context 'after running the job' do
+    let!(:tmpdir) { Rails.root.join("tmp/spec/#{Process.pid}") }
+
     before do
       file_set.id = 'abc123'
       allow(file_set).to receive(:reload)
+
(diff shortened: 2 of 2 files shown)


mbklein:
@samvera/hyrax-code-reviewers Anyone have time to put eyes on this? It's a minor blocker for NU work – we're using a monkeypatched fork but I don't want to get too far behind.

RudyOnRails:
@mbklein would it be good to add a spec in https://github.com/samvera/hyrax/blob/master/spec/jobs/import_url_job_spec.rb that would expect the tmp directory not to be removed in the job via `FileUtils.remove_entry` since a block is no longer passed to `Dir.mktmpdir`?

cjcolvar:
@mbklein I think this looks good.  The only thing missing would be some documentation explaining the need to clean out these temp directories and how to do it that would be added to the [Hyrax Management Guide](https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#production-concerns).  If you add that as a comment here or to a commit message then I think this would be good to go!

mbklein:
How's this? 👇 

### Cleaning up Temporary Directories and Files

Due to the asynchronous nature of the ingest pipeline, Hyrax doesn't always clean up its own temp directories and files. It is, therefore, a good idea to do some out-of-band cleanup via a periodic task (such as a cron job). For example, the following command will clear out any files in `/tmp` over 1 megabyte in size that haven't been changed in the past 4 days:

```sh
$ find /tmp -ctime +3 -and -size +1M -delete
```

In a production environment, specifics such as the location of the temp directory, how often to clear it, and the age/size of the files to remove are system-dependent, so a proper cleanup strategy should be implemented in consultation with sysadmins and/or devops.

cjcolvar:
I added the note to the Hyrax Management Guide:
https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#temp-directories

Issue #2601, opened 2018-02-01
Add Admin Set title on edit view?


elrayle:
Copied from https://github.com/curationexperts/nurax/issues/145

### Descriptive summary

Collections now have a title showing on the edit page which is really helpful. Can we do that for Admin Sets too?

Collection:

![screen shot 2017-12-14 at 11 15 58 am](https://user-images.githubusercontent.com/4356335/34003045-414d272c-e0c2-11e7-8243-25f92a19ed89.png)

Admin Set:

![screen shot 2017-12-14 at 11 30 40 am](https://user-images.githubusercontent.com/4356335/34003069-52e212c2-e0c2-11e7-967a-9cb2cda7265f.png)

PR #2652, opened 2018-02-15
Add title to admin set edit form

laritakr:
fixes https://github.com/samvera/hyrax/issues/2601

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/dashboard/collection_spec.rb
+++ b/spec/features/dashboard/collection_spec.rb
@@ -760,48 +760,70 @@ RSpec.describe 'collection', type: :feature, clean_repo: true do
     end
 
     context 'from dashboard -> collections action menu' do
-      before do
-        sign_in user
-        visit '/dashboard/my/collections'
+      context 'for a collection' do
+        before do
+          sign_in user
+          visit '/dashboard/my/collections'
+        end
+
+        it "edits and update collection metadata" do
+          # URL: /dashboard/my/collections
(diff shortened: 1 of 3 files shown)


laritakr:
I'm not sure... I copied from the collection edit view which doesn't have one. There is only one spec which covers the <% provide :page_header do %> code in the entire app.

elrayle:
@laritakr Can you test for this using a feature test?  You can add it to features/dashboard/collection_spec.rb and make the code similar to the following updated to one of the admin sets.

https://github.com/samvera/hyrax/blob/master/spec/features/dashboard/collection_spec.rb#L768-L779

elrayle:
@cjcolvar Can you do a final review?  I added a feature test to cover the change in this PR.

There was one random Travis failure.  I've kicked it off again.  I expect it to pass since it did everywhere else.

spec/features/dashboard/collection_spec.rb
```diff
@@ -760,48 +760,70 @@ def get_url_fragment(type)
...
+        end
       end
 
-      it "edits and update collection metadata" do
-        # URL: /dashboard/my/collections
-        expect(page).to have_content(collection.title.first)
-        within("#document_#{collection.id}") do
-          find('button.dropdown-toggle').click
-          click_link('Edit collection')
+      context 'for a collection' do
```
cjcolvar: I'm guessing you mean this to be `admin set`.

Issue #2558, opened 2018-01-24
Work Analytics Attributes Page


nestorw:
### Descriptive summary

This represents the page at Reports > Works
**Attributes Tab**

### Mockups
<img width="1247" alt="c3-analytics-works-sttributes-summary" src="https://user-images.githubusercontent.com/1766876/35351819-1f81760c-0110-11e8-8f29-8d1c6c2b76da.png">
<img width="1260" alt="c4-analytics-works-attributes-monthly" src="https://user-images.githubusercontent.com/1766876/35351820-1f8a8f08-0110-11e8-937e-00716f36da9b.png">

### Related work

Link to related tickets or prior related work here.

vantuyls:
Needs:
- [ ] Customize Report form including:
- [ ] date range
- [ ] filters drop downs (work type, status, visibility)
- [ ] sort column (date published)
- [ ] radio buttons for summary tables and month-by-month totals
- [ ] build reporting text (text immediately below report title - summary of filters and totals)
- [ ] table download button with options for csv download 
- [ ] for Month-by-month - monthly table by visibility
- [ ] for Summary table - tables by visibility, work type, resource type, status

Issue #2557, opened 2018-01-24
Work Analytics Activity Page


nestorw:
#### View the number of unique downloads for files in any given collection.
This represents the page at Reports > Works
**Activity Tab**

### Mockups
<img width="1265" alt="c6-analytics-works-activity-monthly" src="https://user-images.githubusercontent.com/1766876/35351769-ed0675a6-010f-11e8-9f3f-9574ee4fd9bb.png">
<img width="1285" alt="c5-analytics-works-activity-summary" src="https://user-images.githubusercontent.com/1766876/35351770-ed130f50-010f-11e8-9254-80ef39d905ee.png">

### Related work

https://wiki.duraspace.org/pages/viewpage.action?pageId=87467336

nestorw:
Status, Attributes, and Activity buttons across the top of this view may be replaced by a layer of nesting in the left side menu under Reports-Works

See: #2565

nestorw:
Needs:
- [ ] Customize Report form including:
- [ ] time period drop down
- [ ] filters drop downs (work type, status, visibility)
- [ ] sort column (status)
- [ ] radio buttons for summary tables and month-by-month totals
- [ ] build reporting text (text immediately below report title - summary of filters and totals)
- [ ] table download button with options for csv download 
- [ ] for Month-by-month - table for Summary of user activity
- [ ] for Month-by-month - table for monthly user activity
- [ ] for Summary table - paginated table of works by user activity

Issue #2555, opened 2018-01-24
Work Analytics Status Page


nestorw:
### Descriptive summary

This represents the page at Reports > Works
**Status Tab**

Backend tasks that need to be completed in order to get data to the front-end

- [ ] #2719 Collect Work Status History
- [ ] #2720 Collect Work Visibility History

For the form:

 - [ ] time period drop down
 - [ ] filters drop downs (work type, status, visibility)
 - [ ] sort column (status)
 - [ ] radio buttons for summary tables and month-by-month totals
 - [ ] build reporting text (text immediately below report title - summary of filters and totals)
 - [ ] table download button with options for csv download
 - [ ] for Month-by-month - table for Visibility status
 - [ ] for Month-by-month - table for Publication status
 - [ ] for Summary table - paginated table of works

### Mockups
<img width="1318" alt="c2-analytics-works-status-monthly" src="https://user-images.githubusercontent.com/1766876/35351693-b17edc26-010f-11e8-8b7b-6d7edc6601f3.png">
<img width="1266" alt="c1-analytics-works-status-summary" src="https://user-images.githubusercontent.com/1766876/35351694-b18b81ba-010f-11e8-850f-aad7adcca296.png">

### Related work

https://wiki.duraspace.org/pages/viewpage.action?pageId=87467336

vantuyls:
Needs:
- [ ] Customize Report form including:
- [ ] time period drop down
- [ ] filters drop downs (work type, status, visibility)
- [ ] sort column (status)
- [ ] radio buttons for summary tables and month-by-month totals
- [ ] build reporting text (text immediately below report title - summary of filters and totals)
- [ ] table download button with options for csv download 
- [ ] for Month-by-month - table for Visibility status
- [ ] for Month-by-month - table for Publication status
- [ ] for Summary table - paginated table of works

vantuyls:
Status, Attributes, and Activity buttons across the top of this view may be replaced by a layer of nesting in the left side menu under Reports-Works

sethaj:
From what I'm gathering from the mockups, "status" seems to be a mix of:

1) Hydra::AccessControls::AccessRight (Embargo for example) 
2) The "status" field that relates to sipity workflows, "suppressed_bsi" in solr where "false" means Published, "true" means Unpublished and "nil" means Unknown
[...]

PR #2660, opened 2018-02-16
Resolves #2565 Nesting report categories for analytics

### LINKED ISSUE #2565: Nesting report categories
### Descriptive summary

Need another layer of nesting under each report group for Status, Attributes and Activity.

sethaj:
Fixes #2565 

This is work towards an analytics redesign enabled by the Filpflop.analytics_resdesign? switch.

When `Flipflop.analytics_redesign? == true`:

* In the dashboard sidebar, Reports is moved from the Activity heading to the Tasks heading
* Reports is now a collapsable menu item containing (currently) Works
* Works is also a collapsable menu item containing Status, Attributes, Activity
* Status, Attributes and Activity while currently blank will be populated through #2555, #2558, #2557 

@samvera/hyrax-code-reviewers


config/routes.rb
```diff
@@ -229,6 +229,14 @@
     resources :users, only: [:index]
     resources :permission_template_accesses, only: :destroy
     resource 'stats', only: [:show]
+    # TODO: from the analytics sprint, admin 'reports' will eventually replace 'stats' above
+    namespace :reports do
+      namespace :works do
+        get 'status', to: 'work_status_report#show'
```
nestorw: @sethaj Are you expecting to have more routes to each of these controllers? Instead of a controller for each action, would it make sense to have one controller called Report(or similar) with actions for status, attributes, and activity. It could reduce the number of nested files and keep similar actions grouped together. If you expect these controller to get complicated then yeah they should probably be separate.

app/controllers/hyrax/admin/reports/works/work_activity_report_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Hyrax
+  class Admin::Reports::Works::WorkActivityReportController < ApplicationController
```
nestorw: Could these be combined?

app/controllers/hyrax/admin/reports/works/work_attribute_report_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Hyrax
+  class Admin::Reports::Works::WorkAttributeReportController < ApplicationController
```
nestorw: this

app/controllers/hyrax/admin/reports/works/work_status_report_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Hyrax
+  class Admin::Reports::Works::WorkStatusReportController < ApplicationController
```
nestorw: and this

app/controllers/hyrax/admin/reports/works/work_activity_report_controller.rb
sethaj: Right. So, the way the Menu Presenter currently works is based off of the controller_name. Which means the menu knows when to be collapsed/un-collapsed based on the active controller, https://github.com/samvera/hyrax/pull/2660/files#diff-6e8bbf07f89e869d65937ee64202ee7bR48

That's why currently each controller is individualized, and the routes are the way they are. But it probably makes sense to not allow the menu_presenter to dictate what routes and controllers we use, but to instead change the menu_presenter?

app/presenters/hyrax/menu_presenter.rb
```diff
@@ -42,6 +42,18 @@ def user_activity_section?
...
+    # @return [Boolean] true if the current controller happens to be one of the controllers that deals
+    # with admin reports. This is used to keep the parent section on the sidebar open
+    def admin_reports_section?
+      %w[works].include?(controller_name) && %w[status attributes activity].include?(action_name)
+    end
+
+    # @return [Boolean] true if the current controller happens to be one of the controllers that deals
+    # with admin work reports. This is used to keep the parent section on the sidebar open
+    def admin_reports_works_section?
+      controller.is_a? Hyrax::Admin::Reports::WorksController
```
nestorw: I think this is a great solution

PR #2666, opened 2018-02-17
Stop swallowing errors

jcoyne:
If there is a problem writing to the database, we need to see that
exception raised so we can fix the problem. Loging the problem and
proceeding is not enough visibility. Furthermore the user has no
indication that what they thought they just succeeded at doing actually
failed.


no-reply:
@jcoyne Logging and reraising seems like an option, yes? Removing logging that's currently present seems too heavy handed for the intended fix.

I agree that this should not swallow the error. If the the user needs to see a flash message instead of a generic error, it should reraise a specific error type that the controller can handle appropriately.

no-reply:
@jcoyne thoughts about https://github.com/samvera/hyrax/pull/2666/commits/3dcb2ae4b0f4c82b00312ec68b7b6e35aaa5d08d ?

jcoyne:
@no-reply That works. I'd be even happier if we knew what errors we are expecting. Is it a KeyError?

no-reply:
Closing as #2855 fixed this issue more completely.

PR #2667, opened 2018-02-17
refactor collection factory to avoid fedora save unless specifically requested

elrayle:
Temporarily, there are two factories to build/create collections.

* collections_factory.rb is the current factory unchanged.  Use `create(:collection)` or `build(:collection)` to use.  NOTE: All new tests should use **NOT** use this factory.  The new light weight factory should be used.
* collections.rb is the new factory that only creates parts related to the collection when they are specifically requested.  Use `build(:collection_lw)`.  You can use `create(:collection_lw)`, but it is discouraged unless you really need the object saved in Fedora.  See the factory for documentation of options that can be passed to `build`.  NOTE: All new tests should use this factory.

Several tests were updated to use the new light-weight factory with the following results.

spec/authorities/qa/authorities/collections_spec
```
  BEFORE:  12.8s
  AFTER:    6.0s
```

spec/services/hyrax/collections_service_spec
```
  BEFORE:  8.7ss
  AFTER:   3.9s
```

spec/services/hyrax/multiple_membership_checker_spec
```
   BEFORE: 13.5s
   AFTER:   3.1s
```

spec/services/hyrax/collections/permissions_service_spec
```
   BEFORE: 1m 8.2s
   AFTER:   37.4s
```

**Next Steps:**  Rewrite other tests that build/create collections.  Once all are converted to the light weight factory, the old factory can be deleted.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/factories/collections.rb
+++ b/spec/factories/collections.rb
@@ -1,0 +1,296 @@
+FactoryBot.define do
+  # Tests that create a Fedora Object are very slow.  This factory lets you control which parts of the object ecosystem
+  # get built.
+  #
+  # PREFERRED: Use build whenever possible.  You can control the creation of the permission template, collection type, and
+  #            solr document by passing parameters to the build(:collection_lw) method.  That way you can build only the parts
+  #            needed for a specific test.
+  #
+  # AVOID: Do not use create unless absolutely necessary.  It will create everything including the Fedora object.
+  #
+  # @example Simple build of a collection with no additional parts created.  Lightest weight.
+  #          NOTE: A user is automatically created as the owner of the collection.
+  #   let(:collection) { build(:collection_lw)
+  #
+  # @example Simple build of a collection with no additional parts created.  User is the owner of the collection.  Lightest weight.
+  #   let(:collection) { build(:collection_lw, user:)
+  #
+  # @example Simple build of a collection with only solr-document.  Owner is given edit-access in solr-document. Light weight.
+  #   let(:collection) { build(:collection_lw, with_solr_document: true)
+  #
+  # @example Simple build of a collection with only a permission template created.  Owner is set as a manager.  Light weight.
+  #   let(:collection) { build(:collection_lw, with_permission_template: true)
+  #
+  # @example Build a collection with only a permission template created.  Permissions are set based on
+  #          attributes set for with_permission_template.  Middle weight.
+  #   # permissions passed thru with_permission_template can be any of the following in any combination
+  #   let(:permissions) { { manage_users: [user.user_key],  # multiple users can be listed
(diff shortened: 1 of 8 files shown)


spec/factories/collections_factory.rb
```diff
@@ -1,5 +1,8 @@
 FactoryBot.define do
   factory :collection do
+    # DEPRECATION: This factory is being replaced by collection_lw defined in collections.rb.  New tests should use the
+    # light weight collection factory.  DO NOT ADD tests using this factory.
+    #
```
RudyOnRails: I'm not sure of how deprecations are handled b/c i've never dealt with it, but should we do something like this so the testmaker sees the warning?

`warn "[DEPRECATION] :collection is deprecated, please use :collection_lw instead, located in spec/factories/collections.rb"`

elrayle: I stuck that in there for myself and failed to update it to an official deprecation warning.  Thanks for catching that.

elrayle: I updated to have a deprecation warning, but there are still many many tests using the original factory.  It really clogs up the test output.  I wonder if we could keep this as a comment until more of the tests are converted over to the new factory.  I'll bring this up at tech call to get a feel for how folks prefer to handle this.

spec/factories/collections.rb
```diff
@@ -0,0 +1,305 @@
...
+                               group_depositors(evaluator.with_permission_template)
+      ActiveFedora::SolrService.add(collection.to_solr, commit: true)
+    end
+
+    # Process the with_nesting_attributes transient property such that...
+    # * adds nesting related solr-document fields for ancestors, parent_ids, pathnames, and depth
+    # @param [Collection] collection object being built/created by the factory
+    # @param [Class] evaluator holding the transient properties for the current build/creation process
+    def self.process_with_nesting_attributes(collection, evaluator)
+      return unless evaluator.with_nesting_attributes.present? && collection.nestable?
```
laritakr: I don't see any other references to `with_nesting_attributes` in the factory or in the specs. Are you not implementing this portion? If not, why are you including this method?

spec/factories/collections_factory.rb
elrayle: What do you think about a notice at the top saying something like...

```
# WILL BE DEPRECATED: This factory is being replaced by collection_lw defined in collections.rb.  New tests should use the
# light weight collection factory.  DO NOT ADD tests using this factory.
```
I definitely didn't like having deprecate warning on.  There are too many places where it is still in use within our own code.

laritakr: I think we definitely want a "do not use" statement. But no noisy deprecation warning.

PR #2671, opened 2018-02-19
Add embargo and lease links to admin dashboard in tasks section

### LINKED ISSUE #1127: Add embargo/lease to dashboard
### Descriptive summary

@mjgiarlo notes - after PR #1042 is merged, we should link the pages into our current dashboard. This will require one or more new tickets. Design-related questions:

- Where do we want embargoes and leases in the dashboard menu?
- Who should see the embargo and leave menu items? (It's currently limited to repo admins, and they can see all embargoed and leased works. Maybe that's OK for an initial implementation, and we need another ticket for letting admin set managers see them, including making sure the lease and embargo controllers and views only show admin set managers works in their admin sets?)

### Related work

PE #1042 

TODO items in #1123 might be better resolved after this design is done.

smithjp:
Resolves #1127 

Adds links to embargo and lease management pages in the admin dashboard task section for all users who can access the admin dashboard.  

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/dashboard/sidebar/_tasks.html.erb
+++ b/app/views/hyrax/dashboard/sidebar/_tasks.html.erb
@@ -15,9 +15,7 @@
     <%= menu.nav_link(hyrax.embargoes_path) do %>
       <span class="fa fa-flag"></span> <span class="sidebar-action-text"><%= t('hyrax.embargoes.index.manage_embargoes') %></span>
     <% end %>
-  <% end %>
 
-  <% if can? :read, :admin_dashboard %>
     <%= menu.nav_link(hyrax.leases_path) do %>
       <span class="fa fa-flag"></span> <span class="sidebar-action-text"><%= t('hyrax.leases.index.manage_leases') %></span>
     <% end %>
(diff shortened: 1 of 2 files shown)


app/views/hyrax/dashboard/sidebar/_tasks.html.erb
```diff
@@ -10,3 +10,15 @@
...
     <% end %>
   <% end %>
+
+  <% if can? :read, :admin_dashboard %>
+    <%= menu.nav_link(hyrax.embargoes_path) do %>
+      <span class="fa fa-flag"></span> <span class="sidebar-action-text"><%= t('hyrax.embargoes.index.manage_embargoes') %></span>
+    <% end %>
+  <% end %>
+
+  <% if can? :read, :admin_dashboard %>
```
cjcolvar: Since this check is the same as in line 14 could this check be skipped or was there a reason to keep them both?

smithjp: I don't think there is any reason to keep them both.  Combining these into one conditional statement makes a lot more sense.  I'll update this now.  Thank you for your feedback!

julesies:
hoping we can get this in beta @elrayle @cjcolvar

Issue #2651, opened 2018-02-15
Feature tests needed for multiple-membership


elrayle:
### Descriptive summary

Write feature tests for the various scenarios of attempting to add a work to collections of the same type that don't allow multiple membership.

### Rationale

User interactions in the UI should be covered by feature tests.

### Expected behavior

Feature tests need to be created that pass/fail when the following functionality is working or not working in the UI.

work A which is in collection_1 should not be added to collection_2 when the two collections are of the same single_membership collection type.  Places that add a work to a collection are...
* Dashboard -> Works -> batch processing using checkboxes
* Dashboard -> Works -> edit work A -> Relationships tab
* Dashboard -> Collections -> click single_membership_collection2 title -> Add existing works to this collection

### Actual behavior

If you test the expected scenarios through the UI, all of them fail to prevent the work from being added to the second collection.

### Related work

#1555 Add service object to enforce single-membership collections
#1550 Collection types should support restricting multiple membership
#2640 Pass an array of collection type gid strings to where query

elrayle:
Fixed by PR #2687

PR #2687, opened 2018-02-21
fix single membership and add single/multi-membership spec tests

elrayle:
Right now, all tests allow the work to be added to both collections even when it shouldn’t.  These tests can be used to determine when the problem is fixed.

There is one tests for single membership that is incomplete, set to skip, and marked with a TODO because I wasn’t sure how to get capybara to do the collection selection process on the work’s edit form Relationships tab.

Addresses #2651 which sets up testing for a fix of multiple membership validation

### DIFF
--- a/spec/features/collection_multi_membership_spec.rb
+++ b/spec/features/collection_multi_membership_spec.rb
@@ -1,0 +1,185 @@
+RSpec.describe 'Adding a work to multiple collections', type: :feature, clean_repo: true, js: true, with_nested_reindexing: true do
+  let!(:admin_user) { create(:admin, email: 'admin@example.com') }
+  let!(:single_membership_type_1) { create(:collection_type, :not_allow_multiple_membership, title: 'Single-membership 1') }
+  let!(:single_membership_type_2) { create(:collection_type, :not_allow_multiple_membership, title: 'Single-membership 2') }
+  let!(:multi_membership_type_1) { create(:collection_type, :allow_multiple_membership, title: 'Multi-membership 1') }
+  let!(:multi_membership_type_2) { create(:collection_type, :allow_multiple_membership, title: 'Multi-membership 2') }
+
+  let!(:col1_sm_1) { create(:collection_lw, id: 'col1_sm_1', title: ['Collection 1 for SM1'], user: admin_user, collection_type_gid: single_membership_type_1.gid) }
+  let!(:col2_sm_1) { create(:collection_lw, id: 'col2_sm_1', title: ['Collection 2 for SM1'], user: admin_user, collection_type_gid: single_membership_type_1.gid) }
+  let!(:col3_sm_2) { create(:collection_lw, id: 'col3_sm_2', title: ['Collection 3 for SM2'], user: admin_user, collection_type_gid: single_membership_type_2.gid) }
+  let!(:col4_mm_1) { create(:collection_lw, id: 'col4_mm_1', title: ['Collection 4 for MM1'], user: admin_user, collection_type_gid: multi_membership_type_1.gid) }
+  let!(:col5_mm_1) { create(:collection_lw, id: 'col5_mm_1', title: ['Collection 5 for MM1'], user: admin_user, collection_type_gid: multi_membership_type_1.gid) }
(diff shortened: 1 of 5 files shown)


elrayle:
@cjcolvar The two failing tests show how multi-membership validation is not working through the UI.

spec/features/collection_multi_membership_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+  let!(:col2_sm_1) { create(:collection_lw, id: 'col2_sm_1', title: ['Collection 2 for SM1'], user: admin_user, collection_type_gid: single_membership_type_1.gid) }
+  let!(:col3_sm_2) { create(:collection_lw, id: 'col3_sm_2', title: ['Collection 3 for SM2'], user: admin_user, collection_type_gid: single_membership_type_2.gid) }
+  let!(:col4_mm_1) { create(:collection_lw, id: 'col4_mm_1', title: ['Collection 4 for MM1'], user: admin_user, collection_type_gid: multi_membership_type_1.gid) }
+  let!(:col5_mm_1) { create(:collection_lw, id: 'col5_mm_1', title: ['Collection 5 for MM1'], user: admin_user, collection_type_gid: multi_membership_type_1.gid) }
+  let!(:col6_mm_2) { create(:collection_lw, id: 'col6_mm_2', title: ['Collection 6 for MM2'], user: admin_user, collection_type_gid: multi_membership_type_2.gid) }
+
+  let!(:work) { create(:generic_work, user: admin_user, title: ['The highly valued work that everyone wants in their collection']) }
+
+  before do
+    admin_user
```
twalpole: No need to call this - you're using `let!` so the object is already created (not lazy) - and even if you were using the lazy `let` it would be created in the next line when you call it again.

elrayle:
Forwarding for success/failure of adding works to collections isn't quite what would be expected.  This is described in Issue #2690 Expected forwarding when batch adding works to a collection.

cjcolvar:
I feel like there's too many feature tests here and that most of them should be converted into controller tests and unit tests.  I like the idea that feature tests should be for critical paths and be positive tests. 
 So we probably only need one for "Adding works to collections via works dashboard".  I think they've been helpful in identifying the issues to fix them and I can work on converting these if you agree.

jenlindner:
@cjcolvar that's an intriguing strategy, and I'd like to learn more about what you mean and how to implement it. Just wanted to mention it.

elrayle:
@cjcolvar Would you be open to merging this as is and address the test refactor in another issue?  I would like to get this in today so it will be part of beta on Monday.

@mjgiarlo Can you give a review of this since you are familiar with the multi-membership code?

mjgiarlo:
@elrayle I'll have a look

mjgiarlo:
@cjcolvar @jenlindner We've talked before about the idea of:

1. Making feature tests not run as part of the CI build (to speed it up);
2. Expanding the number of feature tests; and
3. Creating a [periodic (nightly? weekly? etc.) CI build](https://docs.travis-ci.com/user/cron-jobs/) that runs all the feature specs.

This would take some effort to do, and we'd need folks to commit to keeping an eye on those scheduled builds and to troubleshooting/fixing them when they break, of course.

elrayle:
@cjcolvar @jenlindner  I would totally support the proposal that @mjgiarlo describes.  The tests as written for this PR mirror the ones that QA testing will follow.  If the feature tests are not part of the PR build, then I would be in favor of adding more tests that mirror the QA testing process.  It doesn't eliminate the need for QA, as that does significant cross-browser testing.  But it should make the first few rounds of QA testing quicker and easier.

It still feels to me like this question is bigger than this issue.

elrayle:
I am going to merge this to have the fixed behavior in the QA beta testing.  @cjcolvar If you feel strongly about refactoring the tests, we can revisit next week while the QA beta testing is happening.

cjcolvar:
@elrayle That sounds fine.  My main concern was that these tests might be "the straw that breaks the camels back" in terms of causing travis builds to timeout more often since they were taking ~1min on my machine.

PR #2710, opened 2018-02-27
connect to GA using Google's API gem

conorom:


app/services/hyrax/analytics/google_analytics.rb
```diff
@@ -1,76 +1,114 @@
...
+      # Return an authorized Google Analytics Reporting Service
+      def self.setup_and_authorize
+        unless File.exist?(config['privkey_path'])
+          raise "Private key file for Google Analytics was expected at '#{config['privkey_path']}', but no file was found."
         end
+        analytics = Google::Apis::AnalyticsreportingV4::AnalyticsReportingService.new
+        credentials = Google::Auth::ServiceAccountCredentials.make_creds(json_key_io: File.open(config['privkey_path']))
+        credentials.scope = 'https://www.googleapis.com/auth/analytics.readonly'
+        analytics.authorization = credentials.fetch_access_token!({})["access_token"]
+        analytics
```
mcritchlow: Saw your Slack message about timeout concerns re: memoization. Which does make perfect sense. But I wonder if the auth library has any features that can help mitigate that problem? I see it supports token stores (file and redis) but I'm not sure, after a cursory glance, where there's an easy way to ask if the current credentials/token are still valid and then refresh if necessary.

conorom: Will spin a new ticket to figure this out. It's not easy to see if the gem takes care of refresh tokens for you or not.

PR #2722, opened 2018-02-28
Hide the search option when "All" is the only option in the search box dropdown.

### LINKED ISSUE #1202: Search box dropdown should disappear if only "All" option is available
### Descriptive summary

The search box has a drop down to choose "All", "my collections", etc. For a non-logged-in user this is simply a drop down with one option: "All". It's confusing and unnecessary; in the case where there's only the "All" option there should be no dropdown.

Observed in Sufia 7.3 but I believe still the case in hyrax master

### Rationale

For a digital collections-type site users who never log in greatly outnumber users who do log in, and this dropdown makes no sense for them.

### Expected behavior

Dropdown should disappear, and 'go' button should be styled accordingly (rounded exterior corners)

### Actual behavior

Dropdown with only one option on every search bar for non-logged-in users.

lsitu:
Fixes #1202

Search box dropdown should disappear if only "All" option is available.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/catalog/_search_form.html.erb
+++ b/app/views/catalog/_search_form.html.erb
@@ -10,18 +10,17 @@
       <button type="submit" class="btn btn-primary" id="search-submit-header">
         <%= t('hyrax.search.button.html') %>
       </button>
-      <button class="btn btn-default dropdown-toggle" type="button" data-toggle="dropdown">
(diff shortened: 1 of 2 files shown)


cjcolvar:
@vantuyls This sound good to you?

vantuyls:
Fine by me.

Issue #2726, opened 2018-02-28
Consider removing more/less feature from parent collection lists on collection show views


laritakr:
### Descriptive summary

The list of a nested collection's parents includes both a more/less option and pagination. The code for this feature needs to be revisited. We at Notre Dame feel that this is not good UI design. 

We believe that pagination alone is adequate, and we propose that the more/less feature should be removed. 

### Rationale

1. The feature adds unnecessary code complication to the list of parents collections, both for the feature itself and for the specs to test it. It is located on both the public and dashboard show pages, so the code is duplicated in both places. This includes
- collection_presenter
- dashboard/collections/show
- collections/show
- specs for all of the above
- CSS for hidden attribute
- JS for more/less list
- specs for all of the above

2. Testing for this feature requires building or creating multiple collections so we can look for the visibility attribute, adding unnecessary spec time.
3. Pagination provides a more straightforward way to navigate the list, and more/less is duplicate functionality to maintain.

Please discuss below

laritakr:
Replaced by https://github.com/samvera/hyrax/issues/2743

PR #2692, opened 2018-02-22
Wire up nesting relationship delete buttons

laritakr:
Adds functionality described in https://github.com/samvera/hyrax/issues/2295

Add remove buttons to the parent collections and subcollections list on the collection show page. 
Make the buttons functional to remove a parent collection or a subcollection.

With the move of the remove options to the show page, the relationships tab is being removed. It is now unnecessary as all nesting options are available on the show page. (The relationships tab being available only to those with edit authority was problematic, as it didn't give appropriate access to the
relationship management options.)

**==>> Still needs feature specs revised for remove functionality**

@samvera/hyrax-code-reviewers

Issue #2686, opened 2018-02-21
Revert search builder's use of nested indexing to find works back to member_of_collection_ids


elrayle:
### Descriptive summary

As part of the nested indexing work, Hyrax::CollectionMemberSearchBuilder was updated to search with `self.collection_membership_field = 'nesting_collection__parent_ids_ssim'` instead of `self.collection_membership_field = 'member_of_collection_ids_ssim'`.  That change requires everything to be reindexed.  Reverting should remove the need to reindex everything.

### Rationale

We want to avoid requiring reindexing everything as part of the migration process.

### Expected behavior

Viewing the show page of a non-nested collection with works without having nesting indexing solr fields.

### Actual behavior

Viewing the show page of a non-nested collection with works without having nesting indexing solr fields throws an exception.

### Steps to reproduce the behavior

1. In rspec, create a collection (i.e. :collection_lw) setting the collection type to :non-nestable
1. add at least one work
1. navigate to the collection's show page

Page should successfully load and include the work.

Should be able to modify the show page test in feature/dashboard/collection_spec.rb to not include :with_nested_reindexing.

Issue #2295, opened 2017-12-01
Wire up nested collection delete buttons on collection's admin show page


laritakr:
### Descriptive summary

Collections Sprint: Nested collections relationships can be added on the index page actions, show page buttons, or edit view relationships tab. The relationships tab contains buttons to remove the relationships, but the modals are not yet wired up.

- /app/views/hyrax/dashboard/collections/_modal_remove_from_collection.html.erb
- /app/views/hyrax/dashboard/collections/_modal_remove_sub_collection.html.erb

### Details

All functionality regarding nesting collections should be routed through the nest collection controller, form, and persistence service. This originates on the edit collections view, so this task will require:

- [ ] adding routes for the remove actions, similar to the routes already present for the actions which create the relationships
- [ ] new remove methods in the nest collection controller corresponding to the new actions
- [ ] a new method in the nested collection persistence service to persist the changes
- [ ] wire up the modals to use the new routing to remove the relationships

### Design Modification

In order to simplify the UI and make it more consistent, the management of collection relationships is going to only be available on the collection's admin show page.  This is already the location where work relationships with a collection are managed.

The following changes should be made as part of this issue's resolution.

* remove Relationships tab from edit form
* move Remove modal to admin show page
* add Remove buttons to admin show page for parents and subcollections
* hook up Remove buttons so they launch the Remove confirmation modal and remove the relationship if confirmed

Make sure the ability to Create new collection as subcollection from the parent collection's admin show page still works with the Relationships tab removed from the edit form.

### Related issues
#1719 
#2294 
#2221

laritakr:
Referencing Issue #2573 because PR #2487 may change the design for this issue mentioned above for this issue.

elrayle:
@laritakr Is this done with PR #2692 and #2727?

PR #2727, opened 2018-02-28
Clean up from wiring up remove collection relationships

laritakr:
Related to comments on https://github.com/samvera/hyrax/pull/2692
- removes stray divs
- adds a test for when the remove button is disabled due to lack of edit authority to parent.

Completes issue #2295  

Issue #2726 would remove the need for additional feature specs to test whether the link is visible on the more/less segment of the parent list, so I am not adding any specs at this point for that feature. 

I will be revisiting the collection feature specs as part of issue #2686 which requires a fair amount of rework. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/views/hyrax/dashboard/collections/_show_parent_collection_row.html.erb_spec.rb
+++ b/spec/views/hyrax/dashboard/collections/_show_parent_collection_row.html.erb_spec.rb
@@ -9,15 +9,31 @@ RSpec.describe 'hyrax/dashboard/collections/_show_parent_collection_row.html.erb
   let(:document) { SolrDocument.new(parent_collection_doc) }
   let(:subject) { render('show_parent_collection_row.html.erb', id: child_collection.id, document: document) }
 
-  before do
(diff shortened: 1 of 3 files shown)


spec/views/hyrax/dashboard/collections/_show_parent_collection_row.html.erb_spec.rb
```diff
@@ -9,15 +9,31 @@
...
-    expect(rendered).to have_link(document.title.first)
-    expect(rendered).to have_button("Remove")
-    expect(subject).to render_template("_modal_remove_from_collection")
+  context 'disable button if no edit permission' do
+    before do
+      stub_template "_modal_remove_from_collection.html.erb" => 'modal'
+      allow(view).to receive(:can?).with(:edit, document.id).and_return(false)
+    end
+
+    it 'shows link to collection title and active remove button' do
```
elrayle: shouldn't this be... 'and disabled remove button'?

PR #2745, opened 2018-03-02
GA download event from Work Items drop-down

### LINKED ISSUE #2737: FileSet downloads from Work-view action drop-down not being sent to Google Analytics
### Quick Fix:
Do here...
https://github.com/samvera/hyrax/blob/7c302721f25d56a550c1712f6251e26023cc9d0c/app/views/hyrax/file_sets/_actions.html.erb#L32
... what was done for the media partials download links here:
https://github.com/samvera/hyrax/commit/107f02540c8308c867181cf983679f5779679765#diff-5f70ca21b3d83736a866846775e66e0e

### Background (mostly for Analytics Sprint devs, I guess, and posterity):
Ever since this file was added, folks have presumably expected file download events to be sent to GA:
https://github.com/samvera-deprecated/sufia-core/blob/master/app/assets/javascripts/sufia/ga_events.js
This click on a link with the `file_download` id is still the download-tracking method in Hyrax as of `2.0.1` and soon-to-be `2.1.0`.
https://github.com/samvera/hyrax/blob/7c302721f25d56a550c1712f6251e26023cc9d0c/app/assets/javascripts/hyrax/ga_events.js

However, any time a developer (either directly in Hyrax/Sufia/CC or within a custom view in an app built on same) has added a link to download a file without this id, download events on that link are not sent (unlike the problem in #2718, which is a pulling-stats-down issue). This caveat should be documented for devs, but in an analytics rewrite we should make it harder to break sending events up (which actually is discussed in #968 below, which I found while writing this essay). It's also covered (using gems we may not want to use) in this issue, which we've discussed before:
https://github.com/samvera/hyrax/issues/2400

Unfortunately, the first download link a user sees in Hyrax (when navigating from a parent work) does not have this id etc (as linked up top). That problem goes way back to former apps as well, added to Sufia here and Curation Concerns shortly after:
https://github.com/samvera/curation_concerns/blob/3f843a67faaa27bccda3173d3b4e4acc0fb2f910/app/views/curation_concerns/file_sets/_actions.html.erb#L14

Others, like the media partials do have it, but that's thanks to the resolved issue below.

### Related work
While looking at these I noticed that there has been an almost-identical issue before, which was resolved with changes in the second link above:
https://github.com/samvera/hyrax/issues/968

Hyrax 1.x, Sufia and CC should have this fix on the Work show page FileSet action drop-down too.

### LINKED ISSUE #2767: 2.1.0.beta1:  Search "My shares" gives error messages
also reported to 
* hyku:  https://github.com/samvera-labs/hyku/issues/1518
* nurax: https://github.com/curationexperts/nurax/issues/213

### Descriptive summary

/dashboard/shares page is not present in hyrax/nurax/hyku. 

Nurax: QA Testing for Hyrax
Hyrax Version: 2.1.0.beta1
Hyrax Branch: v2.1.0.beta1_qa
Hyrax Revision: 8fe35f54e067fd499a4467b9ccbefa48233d43a2 

Hyku v1.0.0.beta2 Docker 

Firefox 58.0.2 (64-bit) canonical 1.0

### Rationale

This is a bug.

### Expected behavior
Search within the with the option "My shares" selected presents results on my shared work. 
The search fields which allow this selection of "My shares" are located at    
* /             hyrax starting page (eg. see localhost:3000 or https://nurax.curationexperts.com/) and 
* /catalog
* that very field on the top of the Edit dialog of a single work.

### Actual behavior
Search with the  option "My shares" selected directs to /dashboard/shares which is not preset.

1. Login
1. Go to  
* /            hyrax starting page (eg. see localhost:3000 or https://nurax.curationexperts.com/)  
* /catalog
* that very field on the top of the Edit dialog of a single work.
1. Then select "My Shares" from the pull down right next to the search "Go" button
1. (optional) enter search term 
1. Press the "Go" button 

### Related work

also reported to 
* hyku:  https://github.com/samvera-labs/hyku/issues/1518
* nurax: https://github.com/curationexperts/nurax/issues/213

conorom:
Fixes #2737 

Enables the tracking of download events from the "actions" drop-down on a Work's show page.

detail:
All file download links need to have `id=file_download` and a `data-label` attribute equal to the files NOID in order for the code here to send the event to Google Analytics:
https://github.com/samvera/hyrax/blob/df1e15894f41f4d31aeaddf39845c1f0a214a781/app/assets/javascripts/hyrax/ga_events.js#L6

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/file_sets/_actions.html.erb
+++ b/app/views/hyrax/file_sets/_actions.html.erb
@@ -29,8 +29,12 @@
 
   <% if can?(:download, file_set.id) %>
     <li role="menuitem" tabindex="-1">
-      <%= link_to 'Download', hyrax.download_path(file_set),
-        title: "Download #{file_set.to_s.inspect}", target: "_blank" %>
+      <%= link_to 'Download',
+                  hyrax.download_path(file_set),
+                  title: "Download #{file_set.to_s.inspect}",
+                  target: "_blank",
+                  id: "file_download",
+                  data: { label: file_set.id } %>
     </li>
   <% end %>
 
--- a/spec/views/hyrax/file_sets/_actions.html.erb_spec.rb
+++ b/spec/views/hyrax/file_sets/_actions.html.erb_spec.rb
@@ -1,0 +1,15 @@
+RSpec.describe 'hyrax/file_sets/_actions.html.erb', type: :view do
+  let(:file_set) { stub_model(FileSet) }
+
+  before do
+    allow(view).to receive(:can?).with(:edit, file_set.id).and_return(false)
+    allow(view).to receive(:can?).with(:destroy, file_set.id).and_return(false)
+    allow(view).to receive(:can?).with(:download, file_set.id).and_return(true)
+    render 'hyrax/file_sets/actions', file_set: file_set
+  end
+
+  it "includes google analytics data in the download link" do
+    expect(rendered).to have_css('a#file_download')
+    expect(rendered).to have_selector("a[data-label=\"#{file_set.id}\"]")
+  end
+end


cjcolvar:
@conorom Thanks for working on this!  I think you meant to reference issue #2737.  Can you explain a bit more in either your commit message or this PR description how this change fixes #2737?

conorom:
Dunno how I typo'ed the issue number!
Let me know if the added detail is insufficient.

cjcolvar:
Much clearer.  Thanks!

Issue #1773, opened 2017-10-05
Feature request: Always make default admin set the first item in the select menu of the relationships tab


bess:
### Descriptive summary

Currently, when a new AdminSets is created, it becomes the first (and thus, the auto-selected) option on the "Add as member of administrative set" select menu. This results in unexpected behavior for users. 

### Rationale

An institution using Hyrax starts by only using the default admin set. Staff members are depositing items, and they get used to not having to select anything in the "Relationships" tab because it has automatically selected the default admin set for them. Then, someone adds a second admin set. Now that new admin set is the one being automatically selected, and to many users it appears that their deposited items are disappearing.

### Expected behavior

I would expect that the behavior of the deposit would not change just because a new AdminSet had been created.

### Related work

This has come up at more than one institution. In slack, @bmquinn says he fixed it like this:

Brendan Quinn [1:41 PM] 
We added `selected: AdminSet::DEFAULT_ID` after https://github.com/samvera/hyrax/blob/master/app/views/hyrax/base/_form_relationships.html.erb#L6 in an override of `_form_relationships.html.erb`

VivianChu:
@bess - Do you know if anyone working on this issue?  Thanks

julesies:
i tested this on nurax and it works! Thank you!

PR #2748, opened 2018-03-02
Select admin set already assigned when editing and default when new

### LINKED ISSUE #2747: Editing a work that is not in Default Admin Set - moves the work to that Admin Set
nurax: https://github.com/curationexperts/nurax/issues/206

## Description
If I edit a work that is not in "Default Admin Set", and the Admin Set is then changed to Default Admin Set.

## Expected behavior. 
Original Admin set should be retained unless it is manually moved. 

## Reproduce
1. navigate to a work that is in an admin set that is not the default one.
2. hit edit
3. change some MD or add a file
4. Save (or look at relationships tab and see the Admin set is changed)

elrayle:
Fixes a bug introduced by PR #2639

Fixes #2747

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -46,12 +46,7 @@ module Hyrax
     def new
       # TODO: move these lines to the work form builder in Hyrax
       curation_concern.depositor = current_user.user_key
-
-      # admin_set_id is required on the client, otherwise simple_form renders a blank option.
-      # however it isn't a required field for someone to submit via json.
-      # Set the first admin_set they have access to.
-      admin_set = Hyrax::AdminSetService.new(self).search_results(:deposit).first
-      curation_concern.admin_set_id = admin_set && admin_set.id
+      curation_concern.admin_set_id = admin_set_id_for_new
       build_form
     end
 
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -49,9 +49,9 @@ def new
 
       # admin_set_id is required on the client, otherwise simple_form renders a blank option.
       # however it isn't a required field for someone to submit via json.
-      # Set the first admin_set they have access to.
+      # Set the default admin set if it exists; otherwise, set to first admin_set they have access to.
       admin_set = Hyrax::AdminSetService.new(self).search_results(:deposit).first
-      curation_concern.admin_set_id = admin_set && admin_set.id
+      curation_concern.admin_set_id = AdminSet::DEFAULT_ID || (admin_set && admin_set.id)
```
cjcolvar: Does a user always have access to deposit in the default admin set?  If not, then I think this might cause problems.

elrayle: Good catch.  I'll modify this to be sure the user can access.

smithjp:
I know that this issue has already been resolved and merged, but it looks like setting the default admin set to be the built-in default admin set has been causing some problems with editing works (https://github.com/samvera/hyrax/issues/2747, https://github.com/samvera/hyrax/issues/2773).  It may also cause conflict with local implementations.  I know that OSU (https://github.com/osulp/Scholars-Archive/blob/master/app/services/scholars_archive/admin_set_select_service.rb) and UNC (https://github.com/UNC-Libraries/hy-c/blob/master/app/services/admin_set_select_service.rb) (and probably other institutions) have local implementations for setting default admin sets which may not be compatible with this change.

Looking back at the original issue (https://github.com/samvera/hyrax/issues/1773), it seems to me that some people were just wanting the admin sets to be displayed oldest to newest.  That could be resolved by just reversing the order of the array returned by the AdminSetService and AdminSetOptionsPresenter.  Another option may be to have a local environment variable specifying the default admin set instead of using the pre-defined one in the gem since some institutions may not want to use that as their default.

Issue #1653, opened 2017-09-11
Date sorting on the Notifications page is not date sorting


vantuyls:
### Descriptive summary

The notifications page allows the notification table to be sortable by date (among other things) but the sorting on this field isn't actually sorting on date, it is sorting alphanumerically on the text in the date field. 

### Expected behavior
Sorting the Date field in this (or any) table should sort on date/time

### Actual behavior
Sorting the Date field in this table sorts alphanumerically

### Steps to reproduce the behavior

1. look at notifications tab
2. sort by date
3. alphanumeric!

jcoyne:
@vantuyls what version? I thought this was fixed by https://github.com/samvera/hyrax/pull/1416

jcoyne:
Ah, maybe `:long_ordinal` was a poor choice if the notifications span a month boundary.  Maybe `:db` would have been better.

vantuyls:
this was observed on master.

mjgiarlo:
@vantuyls I'd like to tag this with the 2.0.0 milestone, but that milestone is currently polluted with many many issues. Might you have time to work w/ me on clearing out that milestone so we can make it useful for us?

vantuyls:
Can do.

vantuyls:
see also: https://github.com/curationexperts/nurax/issues/33

blackrat:
This seems to be a basic misunderstanding of how dataTables sorting works.
If you set a 'data-sort' option on the <td>, dataTables will use that for sorting rather than the generated content.

i.e. in _notifications.html.erb
`<td data-sort="<%= msg.last_message.created_at.getutc.iso8601 %>">`

will be sort by the iso8601 representation of the date as opposed to the underlying relative-time string. The (:long_ordinal) content will just be a default form displayed if the relative-time fails to correctly parse that datetime element.

PR #2754, opened 2018-03-02
Add `data-sort` attributes to dataTables rows to support date sort

no-reply:
Adding a `data-sort` with ISO 8601 datetimes allows dataTables to sort
independent of the visible row values. We continue to use javascript to generate
the human-readable time for the reasons given in  #1416 (support for page
caching).

We sort to `#iso8601(5)` (i.e. to 5 decimal places). This still allows a
mismatch between displayed times and correctly sorted times (off by o.ne second),
apparently due to rounding behavior in the javascript sort dates. The actual
dates sort correctly, so display dates sort themselves out (ha!) as a
notification ages past one minute.

Related to #1653; backports #2741.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/notifications/_notifications.html.erb
+++ b/app/views/hyrax/notifications/_notifications.html.erb
@@ -12,7 +12,7 @@
       <tbody>
         <% messages.each do |msg| %>
           <tr>
-            <td>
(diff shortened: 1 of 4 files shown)


no-reply:
Blocked pending #2765.

PR #2758, opened 2018-03-02
Lock `simple_form` less strictly

no-reply:
We previously supported `~> 3.2`. The regression introduced in v3.5.1 is
described at https://github.com/plataformatec/simple_form/issues/1549. We'll
need to fix our presenters to be `ActiveModel` compliant to make `simple_form`
work correctly past that version.

In the meanwhile, we shouldn't require any simple_form upgrade as part of an
upcoming Hyrax release.

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -105,8 +105,9 @@ SUMMARY
   ########################################################
   # Temporarily pinned dependencies. INCLUDE EXPLANATIONS.
   #
-  # simple_form 3.5.1 broke local and Travis builds in bad ways for unknown reasons
-  spec.add_dependency 'simple_form', '=3.5.0'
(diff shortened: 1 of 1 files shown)


cjcolvar:
Would it be better to replace this PR with an upgrade of hydra-editor once https://github.com/samvera/hydra-editor/pull/138 gets into a formal release?  If not, I'm happy to merge this one now with the hydra-editor upgrade later so we can remove this restriction.

mbklein:
I'm not married to any particular solution. Whatever works.

no-reply:
@cjcolvar I'm not sure a forced upgrade of `hyrda-editor` is ideal, if the only reason for it is to update the downstream dependency. This is slightly more overhead on our end, but might save adopters some grief in managing the `hydra-editor` dependency.

Issue #2541, opened 2018-01-22
Admin Dashboard Redesign


vantuyls:
### Descriptive summary
The Admin Dashboard should show a table of the number of works by collection

- [x] Paginated Collection table starting with 10 (?) collections listed
- [x] Paginated Works table starting with 10 (?) works listed
- [x] Toggle on table to show Most Recently Updated vs. Pinned collections or works
- [x] Add concept of pinning a collection or work to this table
- [x] Wire up Total Visitors, Returning Visitors and New Visitors Boxes
- [ ] Repository Objects donut chart toggles between Visibility, Work Types, and Resource Types
- [x] Dropdown to select number of records on each summary table.

![a5-analytics-dashboard-collections-works](https://user-images.githubusercontent.com/9534576/35238732-4291554e-ff63-11e7-8d4f-4c0ad2190ee0.png)

![a2-analytics-dashboard-collections-pinned](https://user-images.githubusercontent.com/9534576/35357158-8c3ab758-0107-11e8-9702-e1baaf13662d.png)

Refer To:
https://wiki.duraspace.org/pages/viewpage.action?pageId=87467336

lfarrell:
Are admin sets moving to a different view? They're on the current default admin view.

lfarrell:
There doesn't currently seem to be a concept of "pinned" anywhere in hyrax. Would we want to set this on the admin dashboard or on the individual collection's page? Also, how would we want to store this info, solr, db, local browser storage, even? Seems pinned collections could even vary by user even within the same institution.

lfarrell:
As best as I can tell there's also no concept of "last updated" for a collection. Would we want the most recently updated record in the collection? That seems wrong, though.

vantuyls:
I wonder if the "featured" functionality is similar enough to "pinned" to be useful. 

As for "last updated", i think we do want to use the most recently updated object in the collection. Can you say more about why that might not be right?

lfarrell:
I was thinking it would be the last time the collection was changed, e.g. a work was added or removed.

vantuyls:
For the statistics boxes across the top of the page:
I think we're settled on the definitions of "visitor" and "session" over in https://github.com/samvera/hyrax/issues/2850

The clarification of these definitions, and need to report compatible analytics from Google Analytics and Matomo has resulted in us changing the design of this page such that there are only **three** analytics boxes shown:
- Registered Users
- Total Visitors
- Total Sessions

PR #2771, opened 2018-03-05
Add initial finders for ResourceStat

mcritchlow:
Changes proposed:
  * `resource_daily_stats` - daily stats query for a user/resource
combination
  * `site_daily_stats` - daily stats query for the entire site
  * `resource_range_stats` - date range query for a user/resource
combination
  * `site_range_stats` - date range query for the entire site

@nestorw @lfarrell -- Please let me know if these initial queries are a good starting point, if I'm missing any that should be added, etc. I tried to base these on what I'm seeing in the mockups, assuming there might be some front-end manipulation/sorting of the data. So, feedback welcome :smile: 

Fixes #2736 

@samvera/hyrax-code-reviewers

Issue #430, opened 2017-02-16
Create user activity graph


jcoyne:
Lower left bar chart on this mockup: 
https://ggeisler.github.io/lerna-dashboard/admin.html

PR #2772, opened 2018-03-06
Charts updated

### LINKED ISSUE #2714: Toggle Repository Object Charts
Repository Objects donut chart toggles between Visibility, Work Types, and Resource Types

### Related work

See #2541

lfarrell:
Fixes #2714 ; refs #2541

Adds toggling for admin charts, save the repository objects chart. 

Switches out the un-maintained morris.js lib for chartkick gem with chart.js.
End users get the new charting libraries, but not the toggling ability, even without the analytics flipper, as the new library needs data in a different format. Charts, should basically look the same

screen shot 2018-03-06 at 10 39 12 am

@samvera/hyrax-code-reviewers


mcritchlow:
Overall this LGTM :+1: 

Related to my current PR in #2771 , I'd be interested to see whether the finders added will do the trick for what you need in https://github.com/samvera/hyrax/blob/827abed05ebaa86ea60d37d912f3e6403e685fb2/app/presenters/hyrax/admin/user_activity_presenter.rb#L36 or if we need to customize things a bit more.

mcritchlow:
@vantuyls - I'm curious about the graph reference above for the user graph. There's a method in there (see above) that references `returning_users` which @lfarrell has included in the graph.

That reference appears to have been put in when the class was created for https://github.com/samvera/hyrax/issues/430

but i'm curious if we're conflating stats here. The notion of new registered users and "returning users" from a remote service like Google or Matomo is a pretty different concept.. I guess what I'm wondering is what story we want that graph to tell.

vantuyls:
so, the graph in question is trying to capture total users (though we probably mean sessions) and returning users (sessions) for the time period. maybe i'm not understanding the question?

mcritchlow:
@vantuyls - my concern is that the currently implemented ["New Users" query](https://github.com/samvera/hyrax/blob/827abed05ebaa86ea60d37d912f3e6403e685fb2/app/presenters/hyrax/admin/user_activity_presenter.rb#L30-L33) that is already in use has nothing to do with the analytics concept of users/sessions, but is actually executing a query for the number of [registered users](https://github.com/samvera/hyrax/blob/827abed05ebaa86ea60d37d912f3e6403e685fb2/app/services/hyrax/statistics/users/over_time.rb#L20), with the site for the given date range.  

So perhaps the current query needs to be changed so both queries come from the new (remote) analytics?

vantuyls:
ah, i see what you mean. Yeah, it make sense to me to pull those two datasets from the same data stream (GA or MATOMO). So that would be "Total Users" and "Total New Users" for whatever the time period is?

mcritchlow:
@vantuyls I think so yes, though we might need to use Analytics language like "Unique Visitors" (or "New Visitors" and "Returning Visitors" or something like that. My concern with using the work User it might make people think of registered users (the current behavior). That said, I think this issue isn't directly related to @lfarrell's PR, so I've created a separate ticket https://github.com/samvera/hyrax/issues/2776 that I'm happy to work on if we're on the same page for desired behavior.

Issue #2677, opened 2018-02-21
Private objects shown to Public Users


julesies:
Moved from nurax: https://github.com/curationexperts/nurax/issues/172

Descriptive summary
A non-authenticated user, I can see objects that are private. This includes Files, Child Works, Collections. I can not get to the actual files, but I can see representations of files and works (with generic File title) and I can see full titles of Private Collections on the Work View.

In addition the Admin Set Title appears on the work view.

Here is a good example including all issues listed:
https://nurax.curationexperts.com/concern/images/r494vk17h?locale=en

Here is a place Private Collections are NOT shown to unauthenticated users in nesting context. And so it's working as expected.

https://nurax.curationexperts.com/collections/9019s2486?locale=en

Desired behavior is:
Private collections are never shown to unauthenticated user/or users with read
Admin sets are never shown on work page
Private files are never shown to unauthenticated user/or users without read

cc @vantuyls for any clarification

laritakr:
The issue of displaying private files on a work is a larger issue, and preferred behavior is still to be determined. This will be moved to another issue. Reference discussion here: https://groups.google.com/forum/#!topic/samvera-tech/eCQdh9qy1io

julesies:
sorry just catching up. I think @vantuyls  made the call to no show private files on a work, so pinging him for clarification.

laritakr:
ongoing discussion is here: https://groups.google.com/forum/#!topic/samvera-tech/eCQdh9qy1io

vantuyls:
Okay, i've made a call over on that google groups thread. 

Based on this and previous conversations, it sounds (to me) like the most common need is for hiding the files if the user does not have permission to access them (ie. they will not appear in the file list).

Issue #2220, opened 2017-11-17
Collections Sprint apparently broke "render_collection_links" method


laritakr:
### Descriptive summary

In Collections Sprint, method "render_collection_links" doesn't seem to return anything, probably because of the parent/child nesting direction switch. I'm not sure exactly where all it is used or what it is intended to return.

It appears to be in the expand link on a collection list, to perhaps it is intended show some sort of relationship information there. I haven't done analysis past debugging the method definition to insure that a solr_doc is coming in, but there are no collection links found by Hyrax::CollectionMemberService.

laritakr:
A bit more detail:

solr field `child_object_ids_ssim` is now obsolete, but is still referenced in two places in Hyrax.

Hyrax::CollectionMemberService is is used by the CollectionsHelper in `def render_collection_links`. This method is used in `list_works`, `list_row`, and `show_document_list_row` partials. Apparently it was intended to show a linked list of the collections that a work belongs to. 

Hyrax::CollectionSizeService is not used anywhere.

PR #2789, opened 2018-03-08
Work show view modifications for collections

laritakr:
Resolves https://github.com/samvera/hyrax/issues/2220
- fixes Hyrax::CollectionMemberService, and adds permission test to
the search.

Resolves https://github.com/samvera/hyrax/issues/2677 (remaining work moved to issue https://github.com/samvera/hyrax/issues/2796)
- shows admin sets only if user is logged in
- shows only authorized collections

Renames titles in collection and collection type factories to avoid
duplicate titles. This fixes a random spec failure.

Suggest removing Hyrax::CollectionSizeService which doesn't work and isn't used.  Opened issue #2801 for this work.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/helpers/hyrax/collections_helper_spec.rb
+++ b/spec/helpers/hyrax/collections_helper_spec.rb
@@ -1,4 +1,7 @@
 RSpec.describe Hyrax::CollectionsHelper do
+  let(:user) { create(:user, groups: ['admin']) }
+  let(:ability) { Ability.new(user) }
+
   before do
     # Stub route because helper specs don't handle engine routes
     # https://github.com/rspec/rspec-rails/issues/1250
@@ -9,9 +12,13 @@ RSpec.describe Hyrax::CollectionsHelper do
   end
 
   describe '#render_collection_links' do
-    let!(:work_doc) { SolrDocument.new(id: '123', title_tesim: ['My GenericWork']) }
+    before do
+      allow(controller).to receive(:current_ability).and_return(ability)
+    end
 
     context 'when a GenericWork does not belongs to any collections', :clean_repo do
+      let!(:work_doc) { SolrDocument.new(id: '123', title_tesim: ['My GenericWork']) }
+
       it 'renders nothing' do
         expect(helper.render_collection_links(work_doc)).to be_nil
       end
@@ -20,8 +27,9 @@ RSpec.describe Hyrax::CollectionsHelper do
     context 'when a GenericWork belongs to collections' do
       let(:coll_ids) { ['111', '222'] }
       let(:coll_titles) { ['Collection 111', 'Collection 222'] }
-      let(:coll1_attrs) { { id: coll_ids[0], title_tesim: [coll_titles[0]], child_object_ids_ssim: [work_doc.id] } }
-      let(:coll2_attrs) { { id: coll_ids[1], title_tesim: [coll_titles[1]], child_object_ids_ssim: [work_doc.id, 'abc123'] } }
+      let(:coll1_attrs) { { id: coll_ids[0], title_tesim: [coll_titles[0]] } }
+      let(:coll2_attrs) { { id: coll_ids[1], title_tesim: [coll_titles[1]] } }
+      let!(:work_doc) { SolrDocument.new(id: '123', title_tesim: ['My GenericWork'], member_of_collection_ids_ssim: coll_ids) }
 
       before do
         ActiveFedora::SolrService.add(coll1_attrs)
@@ -80,7 +88,7 @@ RSpec.describe Hyrax::CollectionsHelper do
       let(:test_collection_type) { FactoryBot.create(:collection_type) }
 
       it "returns the CollectionType title" do
-        expect(collection_type_label(test_collection_type.gid)).to match %r{^Title \d+$}
+        expect(collection_type_label(test_collection_type.gid)).to eq test_collection_type.title
       end
     end
 
--- a/spec/services/hyrax/collection_member_service_spec.rb
+++ b/spec/services/hyrax/collection_member_service_spec.rb
@@ -1,28 +1,19 @@
 RSpec.describe Hyrax::CollectionMemberService, :clean_repo do
-  let(:work_attrs) { { id: '123', title_tesim: ['A generic work'] } }
-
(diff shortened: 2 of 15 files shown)


app/services/hyrax/collection_size_service.rb
```diff
@@ -1,40 +0,0 @@
-# returns a list of solr documents for collections the item is a part of
-module Hyrax
-  class CollectionSizeService
```
elrayle: I like that you have an eye out for technical debt that is not in use.  In this case, the service should be deprecated instead of removed.  It has been around for 3 years and it is possible that sites are making use of it.  I would deprecate it with a targeted removal of 3.0.  That should give people plenty of time to comment if they want to keep it around.  

Don't we have a method somewhere else that does this count that we use in search results and other places?  The deprecation notice can point them to the other method.

spec/helpers/hyrax/collections_helper_spec.rb
```diff
@@ -80,7 +88,7 @@
       let(:test_collection_type) { FactoryBot.create(:collection_type) }
 
       it "returns the CollectionType title" do
-        expect(collection_type_label(test_collection_type.gid)).to match %r{^Title \d+$}
+        expect(collection_type_label(test_collection_type.gid)).to match %r{^Collection Type \d+$}
       end
```
elrayle: I would suggest that you check for an exact match of the title using...
```
expect(collection_type_label(test_collection_type.gid)).to eq test_collection_type.title
```

The reasoning behind this is that the title generator produces the match pattern for every collection type.  So this general match pattern could match any number of collection types and not test_collection_type specifically.  Looks like that got missed when this test was originally created.

spec/services/hyrax/collection_member_service_spec.rb
```diff
@@ -1,28 +1,19 @@
...
-  let(:work_attrs) { { id: '123', title_tesim: ['A generic work'] } }
-
-  let(:coll1_attrs) { { id: 'col1', title_tesim: ['A Collection 1'], child_object_ids_ssim: [work.id] } }
-  let(:coll2_attrs) { { id: 'col2', title_tesim: ['A Collection 2'], child_object_ids_ssim: [work.id, 'abc123'] } }
-  let(:coll3_attrs) { { id: 'col3', title_tesim: ['A Collection 3'], child_object_ids_ssim: ['abc123'] } }
-
+  let(:user) { create(:user, groups: 'potato') }
+  let!(:ability) { ::Ability.new(user) }
+  let!(:col1) { build(:collection_lw, id: 'col1', user: user, with_solr_document: true) }
+  let!(:col2) { build(:collection_lw, id: 'col2', with_solr_document: true) }
```
elrayle: Excellent use of new light weight factory.

spec/services/hyrax/collection_size_service_spec.rb
```diff
@@ -1,74 +0,0 @@
-RSpec.describe Hyrax::CollectionSizeService do
-  let(:coll1_attrs) do
```
elrayle: This will need to be reinstated with the service and deprecated.

spec/views/hyrax/base/_attributes.html.erb_spec.rb
```diff
@@ -14,7 +14,7 @@
       description_tesim: description
     }
   end
-  let(:ability) { nil }
+  let(:ability) { double(admin?: true) }
```
elrayle: This seems like a significant change in the test.  Can you say more about the need for this change?

spec/views/hyrax/base/_relationships.html.erb_spec.rb
```diff
@@ -29,6 +30,7 @@
     let(:member_of_collection_presenters) { [collection] }
 
     before do
+      allow(controller).to receive(:current_user).and_return true
```
elrayle: I would have expected this to return `user` instead of true. (again at line 49)

elrayle:
One other question... do you know what kicks off the call to ParentCollectionSearchBuilder.  This is called once per work on the Dashboard -> Works index page.  I understand the call from Work show page.  But it seems like it shouldn't do the search for every work on the index page.

This was not introduced by your PR.  It was doing this before.  I thought you might have some insight into the calling sequence.

app/services/hyrax/collection_size_service.rb
laritakr: I haven't found anything regarding collection size. The size attribute in Solr is always  empty. If this worked before Collections Sprint, it wasn't actually part of Hyrax. But it doesn't work now. If we want to keep it, it should probably be fixed.

spec/helpers/hyrax/collections_helper_spec.rb
laritakr: Ok, I can change that. I didn't look at what this was doing... I only changed it because of changing the collection type title.

spec/services/hyrax/collection_member_service_spec.rb
laritakr: :) So convenient too!

spec/views/hyrax/base/_attributes.html.erb_spec.rb
laritakr: because I added the user ability testing to the method that displays the collection links, it now accesses a method that checks for admin? Since the actual search was tested elsewhere, This view doesn't need to test the ability logic.

spec/views/hyrax/base/_relationships.html.erb_spec.rb
laritakr: hmmm that looks like a copy/edit error... I had to make several adjustments to the search service and places using it to get it to always have the ability, so I had to change the specs several times. I'm not sure what was intended here.

laritakr:
Yes, if you look at the list of works, each work now includes the parent collection titles on the list. It was broken before, so fixing the search makes them now appear.

spec/views/hyrax/base/_relationships.html.erb_spec.rb
laritakr: It was passing because it was only looking to see if we had a current user, not what the user was. But I changed it to return user so it's a more appropriate response.

PR #2790, opened 2018-03-09
Set video width to force it to stay inside column

### LINKED ISSUE #2744: Video and audio files do not download properly on view pages. All player buttons not working on Work View
Nurax Issue: https://github.com/curationexperts/nurax/issues/153

### Descriptive summary
For both the work view and file set pages, video files and audio files do not download properly. This was tested on Chrome and Safari on Mac.

On the work view page, the full screen and download buttons do not work at all.
here is an example: https://nurax.curationexperts.com/concern/images/r494vk17h?locale=en

On the fileset page, fullscreen and download buttons work, but downloads the wrong file. See audio download example. 

screen shot 2017-12-19 at 12 52 40 pm

On the work view page, under Items: if one chooses "Select an action: download" on a video file, it downloads the proper file. 

screen shot 2017-12-19 at 12 50 40 pm

1. Navigate to a video or audio work
2. try to use the player buttons
3. click the download button
4. navigate to fileset page
5. try to use the player buttons
6. click the download button

cjcolvar:
Fixes #2744.

## Before
before

The native browser video player was sizing itself and overflowing the allotted column width of the parent `col-sm-3` div.  For the nurax item linked in #2744 the right-side metadata column was overlaying the buttons making them unactionable.

## After
after

The player is constrained to the column by setting its width to 100%.  This does make the player seem really tiny but it is a larger problem to resize the parent `col-sm-3` div.

## Incorrect file downloads
Issue #2744 mentions the download button downloading the incorrect file but doesn't go into too much detail.  I believe the issue is that the download button on the player is a download of the source loaded in the video player thus a derivative file and not the original file.  This can be confusing since the download action on the file in the items section will give you the original file.  I believe this is out of scope to fix since this is the default behavior of the native browser player.  Future work could change this player to a JS player like mediaelement.js or a IIIF presentation 3.0 player like UV 3.0 (both of which are still in development).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/file_sets/media_display/_audio.html.erb
+++ b/app/views/hyrax/file_sets/media_display/_audio.html.erb
@@ -1,6 +1,5 @@
-  <audio controls="controls" class="audiojs" preload="auto">
+  <audio controls="controls" class="audiojs" style="width:100%" preload="auto">
     <source src="<%= hyrax.download_path(file_set, file: 'ogg') %>" type="audio/ogg" />
     <source src="<%= hyrax.download_path(file_set, file: 'mp3') %>" type="audio/mpeg" />
     Your browser does not support the audio tag.
   </audio>
-
--- a/app/views/hyrax/file_sets/media_display/_video.html.erb
+++ b/app/views/hyrax/file_sets/media_display/_video.html.erb
@@ -1,4 +1,4 @@
-  <video controls="controls" class="video-js vjs-default-skin" data-setup="{}" preload="auto">
+  <video controls="controls" class="video-js vjs-default-skin" style="width:100%" data-setup="{}" preload="auto">
     <source src="<%= hyrax.download_path(file_set, file: 'webm') %>" type="video/webm" />
     <source src="<%= hyrax.download_path(file_set, file: 'mp4') %>" type="video/mp4" />
     Your browser does not support the video tag.


julesies:
re: download. On file types where the audio or video player loads, there is no download file link as there are with other file types. And the download button on these two players downloads a "j-ogg.ogg" file. I feel like this behavior changed since 2.0 but I don't have a 2.0 instance to compare to. 

The download issue you can see:

A PDF example (shows download file under thumbnail): https://nurax.curationexperts.com/concern/parent/r494vk17h/file_sets/fj236208t

these examples are where there is no download file option except in the player. when you click that button it downloads a file named  "j-ogg.ogg".

a video example: https://nurax.curationexperts.com/concern/images/r494vk17h?locale=en

an audio example: https://nurax.curationexperts.com/concern/parent/r494vk17h/file_sets/wh246s13j

I think something could be addressed here...maybe minimally we add download links to these pages under the play to be consistent with other filetypes? happy to copy to a separate issue... cc @vantuyls @cjcolvar @chrisdaaz

cjcolvar:
@julesies I agree it would probably be best to hide the download button on the player and add a download link below it for audio and video files.  This would also have the added benefit of analytics tracking similar to the other media types.

julesies:
@cjcolvar so I can write a new ticket?

cjcolvar:
@julesies Does this PR look good to you?  I wanted to check before merging.

julesies:
I assume it makes the buttons all usable on the work view for both audio and video files? if so, yes. And i'll write a new hyrax ticket for the download issue.

cjcolvar:
I forgot to make this change on the audio player as well.  Adding that now.

jenlindner:
👍 @cjcolvar

Issue #2672, opened 2018-02-20
Collection type creation loses settings if you add a participant before a save.


elrayle:
### Descriptive summary
When creating a Collection Type, if you select/deselect settings, then don't hit save, then click to the participants tab, then add a participant, when you return to the "settings" tab, your settings have returned to default. I can see this leading to some mistakes and having to delete collection types since they can't be edited after created. So I think this is important to address somehow. 

### Expected behavior
Selections are retained as you move through tabs or there is a warning given. 

### Steps to reproduce the behavior

1. Create a collection type
2. Add title, hit save
3. Go to settings: Change the Settings, don't click save
4. Go to participants: add a partcipant
5. Go to Settings: See defaults back. 

### Related Work

Reported in Nurax issue https://github.com/curationexperts/nurax/issues/178

vantuyls:
could we use the approach here?: https://github.com/samvera/hyrax/pull/2799

dlpierce:
I'm near finished implementing the approach mentioned above, while also addressing #2907, which depends on fixing #2273.

PR #2799, opened 2018-03-09
Adds confirmation alert before leaving unsaved content blocks

### LINKED ISSUE #1088: Warn or auto-save before leaving unsaved content blocks
### Descriptive summary

When editing content blocks in the administrative dashboard, entered text stays in place if you move between tabs without saving but is lost for all but the focussed tab when saving. Either save changes across all tabs, or warn when leaving a tab that unsaved content will be lost.

[ observed in Hyku v1.0.0.beta1 ]

### Rationale

It is an understandable assumption that all changes will be saved across different tabs. Some of these blocks could be quite length, so losing text will be a frustration!

### Expected behavior

Either pop-up a warning when changing tabs, or save changes in all tabs on 'Save'.

### Actual behavior

Only the tab in focus is saved.

1. Go to the admin dashboard
2. Enter text in more than one tab
3. Save

VivianChu:
Fixes #1088  ; refs #1088 

Adds confirmation alert when changing tabs to warn user that the unsaved content will be lost.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/content_blocks/_modal_content_block.html.erb
+++ b/app/views/hyrax/content_blocks/_modal_content_block.html.erb
@@ -1,0 +1,15 @@
+<div class="modal fade" id="change-tab-content-block" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="delete-collection-form">
+        <div class="modal-body">
+          <%= t(:'hyrax.content_blocks.change_tab_message') %>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-default" data-dismiss="modal"><%= t('helpers.action.cancel') %></button>
+          <%= link_to 'OK', hyrax.edit_content_blocks_path, class: 'btn btn-default', id: "change-tab-btn"%>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
--- a/app/assets/javascripts/hyrax/content_blocks.js
+++ b/app/assets/javascripts/hyrax/content_blocks.js
@@ -9,4 +9,16 @@ Blacklight.onLoad(function() {
     $this.parent().hide();
     $($this.data('target')).show();
   });
+
+  $('.show-confirm').on('click', function(evt) {
+    evt.preventDefault();
+    evt.stopPropagation();
+    $('#change-tab-content-block').modal('show');
+  });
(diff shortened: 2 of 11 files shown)


cjcolvar:
@VivianChu I'm wondering why you closed this PR.  It seems to be relevant to https://github.com/samvera/hyrax/issues/2672.

VivianChu:
@cjcolvar - I closed it because the build 8023.1 passed and 8023.2 failed https://travis-ci.org/samvera/hyrax/builds/352448875?utm_source=email&utm_medium=notification
Do you know how to trigger the build job again?  Thanks

cjcolvar:
@VivianChu If you're logged in and have permissions there should be a button on the far right to restart an individual build job.  There is also a button to restart all of the build jobs labelled Restart build.  Do you see those buttons on that travis page?

There are some hyrax tests which randomly fail but which usually pass and it looks like that is what is happening for your PR so it probably just needs a restart like you're thinking.

It is totally fine and probably good to have the PR open while you're dealing with travis errors.  If the code is still evolving it is best to change the title of the PR to start with "[WIP]" to let everyone know.

VivianChu:
@cjcolvar - I see those buttons after I logged in.  The build is running now.  Thanks

vantuyls:
functionality-wise, i think this looks good. this is a much complained-about part of hyrax, and this fix will help clarify the UX.

VivianChu:
@adamjarling - The submit_tag has been replaced with link_to.  The javascript and the test have also been updated.  Thanks

app/views/hyrax/content_blocks/_modal_content_block.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="modal fade" id="change-tab-content-block" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="delete-collection-form">
+        <div class="modal-body">
+          <p>Are you sure you want to leave this tab?  Any unsaved data will be lost.</p>
```
cjcolvar: Please pull this string out to a local file for internationalization.

VivianChu: @cjcolvar - This string has been pull out.  Thanks.

Issue #2777, opened 2018-03-06
Changing visibility on a work doesn't reindex correctly


laritakr:
### Descriptive summary

When visibility is changed on a work, the access in Fedora gets changed, but it doesn't reindex correctly. This is similar to a prior issue in collections which required an extra "update_index". See http://samvera.github.io/nested-indexing.html#potential-pitfalls for more information. 

The nesting reindexer is causing the data to not be indexed on changes, because indexing happens before the actual data is persisted to Fedora, and the nesting reindexer uses persisted data, not cached data. 

There may be other errors... UI testing is needed to find everywhere where data is not being indexed appropriately for works.

Identified situations with bugs
- [ ]  Changing visibility setting on a work

cjcolvar:
:+1: I noticed this specific case where changing visibility of a work required two saves in order to change the visibility badge but thought I was just doing something odd.

laritakr:
Caused by this issue: https://github.com/samvera/hyrax/issues/1826

Issue #1826, opened 2017-10-10
Indexing following save/update -- sequence of events


laritakr:
### Descriptive summary

When we index an object, depending on the transaction which is underway, we may be updating the Solr index before the object is actually saved in Fedora. If we are saying that Fedora is to be the canonical source of information, this is not correct.

### Detail

As we [update permissions on a collection](https://github.com/samvera/hyrax/blob/collections-sprint/app/models/concerns/hyrax/collection_behavior.rb#L126-L133), ActiveFedora goes through the following steps:
- Update Collection
- Index Collection
- Update Hydra::AccessControl
- Index Hydra::AccessControl
- Update Hydra::AccessControls::Permission
- Index Hydra::AccessControls::Permission

What this means is that when we index the collection, we are using the in-memory permissions but they have not yet been saved to Fedora because that happens when we make the change to Hydra::AccessControls::Permission.

With the addition of collection nesting, we first added an after_save callback to index all of the nested collections. However, there are cases where we do an update_index without a save, when indexing values which are not part of Fedora. This misses some needed reindexing, so we changed it to an after_update_index callback. 

This changes the above sequence to: 
- Update Collection
- Index Collection 
--> using values in memory but not yet saved, because actual values are stored in Hydra::AccessControls::Permission and have not yet been written
- Index nested collections 
--> using values from Fedora only. It no longer has the in-memory values, thus removing prior indexing changes.
- Update Hydra::AccessControl
- Index Hydra::AccessControl
- Update Hydra::AccessControls::Permission
- Index Hydra::AccessControls::Permission

To solve this situation, we needed to call update_index on the Collection at the end to insure that the values from each of the objects are included in the Collection's Solr document. There is still the potential that the after_update_index callback to reindex could overwrite other indexing changes as well.

jeremyf:
In working through the particular problem, we found that calling `update_index` after the `update!` was an adequate work around. This does add several more processes related to indexing onto that particular method.

laritakr:
I would say the concern is if we could be missing other places where nesting indexing could similarly override prior indexing because the indexing happened before the Fedora update.

PR #2803, opened 2018-03-12
Reindex work after visibility change

laritakr:
Resolves https://github.com/samvera/hyrax/issues/2777

Adding an update_index after a visibility update to the work.
Nested reindexing causes the visibility to not be indexed on the work because the work indexing happens before the visibility is updated in fedora. (see https://github.com/samvera/hyrax/issues/1826 for more info regarding situation)

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/interpret_visibility_actor.rb
+++ b/app/actors/hyrax/actors/interpret_visibility_actor.rb
@@ -96,7 +96,7 @@ module Hyrax
         attributes = intention.sanitize_params
         new_env = Environment.new(env.curation_concern, env.current_ability, attributes)
         validate(env, intention, attributes) && apply_visibility(new_env, intention) &&
-          next_actor.update(new_env)
+          next_actor.update(new_env) && reindex_visibility_change(new_env)
       end
 
       private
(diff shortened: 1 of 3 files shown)


laritakr:
@cjcolvar Test is now added, and I confirmed that it failed before change and passes after.

cjcolvar:
@elrayle I moved the update_index call into the actor stack where I think it is more natural and useful in more cases beyond the works controller.  If this change doesn't make sense then we can revert it.

app/actors/hyrax/actors/interpret_visibility_actor.rb
```diff
@@ -227,6 +227,13 @@ def apply_embargo(env, intention)
           return unless env.curation_concern.embargo
           env.curation_concern.embargo.save # see https://github.com/samvera/hydra-head/issues/226
         end
+
+        def reindex_visibility_change(env)
+          return true unless env.curation_concern.visibility_changed?
+          env.curation_concern.update_index
+          # Don't try to handle the indexing response and always return true
+          true
```
elrayle: Should this be a TODO where we handle the indexing response in the future?  If not, can your comment include why it is find to ignore the indexing response?

cjcolvar: I'm concerned that reaching into the response hash (`response["responseHeader"]["status"]`) will be too brittle but I could put that in.  For what it's worth, it looks like even ActiveFedora ignores the rsolr response: https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/indexing.rb#L74-L79

elrayle: OK

PR #2810, opened 2018-03-15
Add admin_set_id field to edit embargo and lease forms

### LINKED ISSUE #2773: Updating embargoes from Embargo Management Page reverts object to Default Admin Set
Nurax Issue: https://github.com/curationexperts/nurax/issues/219

### Descriptive summary

Browser: Safari 11.0.3 (13604.5.6)
OS: MacOS 10.13.3
Issue: When depositing a work to a templated admin set (i.e., Admin Set Test 12), any updates to the embargo from the Embargo Management Page revert the object to the Default Admin Set.

### Rationale

Object should stay in the specified admin set, and user should only be allowed to apply the embargo specified by that admin set.

1. Create a new work 
2. Under Relationships tab, select an admin set that places restrictions on the length of an embargo (like Admin Set Test 12)
3. Specify the embargo length within the appropriate window and save
4. From the show page for the work, select Edit

embargoes1

5. Navigate to the Embargo Management Page and update the embargo to something outside the range specified by the admin set and save

embargoes2

6. The work will now be in the Default Admin Set with the updated embargo.

embargoes3

smithjp:
Fixes #2773 

Adding admin_set_id fields to the embargo edit form fixes the problem of the admin set reverting to the default admin set when a user attempts to change the embargo release date to one that is invalid based on its current admin set.  Now when this happens, the page redirects to the work edit page with a message which says that the release date is invalid (default behavior).

I added the admin_set_id field to the lease form as well since it is required in both the embargo and lease validation methods in the app/actors/hyrax/actors/interpret_visibility_actor.rb file.

@samvera/hyrax-code-reviewers


elrayle:
Replaced by the fix in PR #2935.  Closing.

PR #2816, opened 2018-03-19
Handle byte range requests for local files

### LINKED ISSUE #2815: Video playback not working in Safari
### Descriptive summary

For the 2.1.0 beta1 - Video is not playing back in Safari desktop or iOS, it is working in Chrome as expected. 

Audio is also playing back in both Safari Desktop and iOS

This link has both Audio and Video Files: https://nurax.curationexperts.com/concern/images/r494vk17h?locale=en

From: https://github.com/curationexperts/nurax/issues/204

cjcolvar:
Fixes #2815.

Byte-range requests in hyrax for derivatives is broken keeping Safari and iOS devices from showing audio and video.  This PR attempts to fix that.

TODO
Avoid overriding so much of Hydra::DownloadController
Avoid conditional logic in most methods
Add new tests for byte-range requests

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/downloads_controller.rb
+++ b/app/controllers/hyrax/downloads_controller.rb
@@ -15,26 +15,98 @@ module Hyrax
         super
       when String
         # For derivatives stored on the local file system
-        response.headers['Accept-Ranges'] = 'bytes'
-        response.headers['Content-Length'] = File.size(file).to_s
-        send_file file, derivative_download_options
+        send_content
       else
         raise ActiveFedora::ObjectNotFoundError
       end
     end
+
+    protected
+
+      # Override
+      # render an HTTP Range response
+      # rubocop:disable  Metrics/AbcSize
+      def send_range
+        _, range = request.headers['Range'].split('bytes=')
+        from, to = range.split('-').map(&:to_i)
+        to = file_size - 1 unless to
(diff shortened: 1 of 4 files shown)


jcoyne:
@cjcolvar Does this https://github.com/samvera/hydra-head/pull/427 fix the problem you are seeing?

cjcolvar:
@jcoyne I don't think so, but that might be good to try for further improvement.  The issue here is that the byte-range request handling just wasn't happening for local files with `send_file` being used instead.
FWIW the documentation for `send_data` says that it is similar to `render plain: data` and does a slightly different render itself:
`render options.slice(:status, :content_type).merge(body: data)`

cjcolvar:
I separated the overrides into a mixin module.  I think I prefer this approach over the prior commit but both pass the added tests.  I can squash the commits depending on the code review and the preferred approach.

app/controllers/hyrax/downloads_controller.rb
```diff
@@ -1,6 +1,7 @@
 module Hyrax
   class DownloadsController < ApplicationController
     include Hydra::Controller::DownloadBehavior
+    include Hyrax::LocalFileDownloadsControllerBehavior
```
jenlindner: 👍

PR #2821, opened 2018-03-20
Check http status and notify user that the download failed for Import…

### LINKED ISSUE #1235: ImportUrlJob should check http status and handle non-200 codes differently
### Descriptive summary

In Hyrax master the import_url job does not check the status of the url from the remote server.  This could cause an error page to be copied as the content into hyrax.

### Rationale

This causes the system to believe that the content was captured successfully, when really no content was preserved.  

### Expected behavior

When a url times out we should notify the user that the download failed, not just copy the error page as content.

### Actual behavior

The html error page is copied to the content of the work and preserved instead of the content.

1. In general if you try to gather many large files from Box, or Dropbox the last few URLS will tend to timeout.

### Related work

We addressed this issue in ScholarSphere here: https://github.com/psu-stewardship/scholarsphere/commit/22fe08c1957dda228bf4d6a5e239f6a48bee25f1s

VivianChu:
Fixes #1235  ; refs #1235 

Adds http status checking, renames fileset, and notifies user if there's something wrong when uploading files from remote server.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/jobs/import_url_job_spec.rb
+++ b/spec/jobs/import_url_job_spec.rb
@@ -14,6 +14,9 @@ RSpec.describe ImportUrlJob do
   let(:operation) { create(:operation) }
   let(:actor) { instance_double(Hyrax::Actors::FileSetActor, create_content: true) }
 
+  let(:mock_retriever) { double }
+  let(:inbox) { user.mailbox.inbox }
+
   before do
     allow(Hyrax::Actors::FileSetActor).to receive(:new).with(file_set, user).and_return(actor)
 
@@ -26,6 +29,9 @@ RSpec.describe ImportUrlJob do
     stub_request(:get, "http://example.org#{file_hash}").to_return(
       body: File.open(File.expand_path(file_path, __FILE__)).read, status: 200, headers: response_headers
     )
+
+    allow(BrowseEverything::Retriever).to receive(:new).and_return(mock_retriever)
+    allow(mock_retriever).to receive(:retrieve)
   end
 
   context 'before enqueueing the job' do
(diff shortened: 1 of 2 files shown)


elrayle:
@cam156 Does this PR meet the requirements of the issue you created?

afred:
@VivianChu, during our post-tech-call PR review session, we were wondering why the name of the `FileSet` should be changed to include an error message in case of failure.

VivianChu:
@afred - I was looking at how ScholarSphere fix the issue by renaming the Fileset and checking http status.  If it's not what you wanted, I can remove it.

VivianChu:
@afred - The file renaming is removed and the tests are updated.  Thanks

spec/jobs/import_url_job_spec.rb
```diff
@@ -26,6 +31,10 @@
     stub_request(:get, "http://example.org#{file_hash}").to_return(
       body: File.open(File.expand_path(file_path, __FILE__)).read, status: 200, headers: response_headers
     )
+
+    allow(BrowseEverything::Retriever).to receive(:new).and_return(mock_retriever)
+    allow(mock_retriever).to receive(:retrieve)
+    allow(HTTParty).to receive(:head).and_return(mock_http_result)
```
no-reply: Should we use WebMock instead of mocking `HTTParty` directly? We are [requiring it in the test suite](https://github.com/samvera/hyrax/blob/51b45cc2dc35eb8101f3eec5bde28922b63a0872/spec/spec_helper.rb#L60) (but maybe not actually using it), and it seems like it will be more robust than coupling the test to the specific client software.

no-reply:
@bess I know you've worked on closely related code in https://github.com/curationexperts/laevigata/pull/925; do you have design feedback on this? I'm particularly interested in whether there's a better pattern than introducing network latency inline in `before_enqueue`.

spec/jobs/import_url_job_spec.rb
VivianChu: @no-reply - The test has been updated to use WebMock instead of mocking HTTParty.  Could you review again.  Thanks

no-reply:
Hi @VivianChu thanks for the changes. This looks great.

I'm going to see if I can resolve the conflicts and push this forward.

no-reply:
Hi @VivianChu thanks for the changes. This looks great.

I'm going to see if I can resolve the conflicts and push this forward.

cjcolvar:
@jrgriffiniii Do you mind reviewing this PR?  Particularly the changes dealing with `#operation` since I removed that method in the merge from master.  It was causing tests to fail and I believe was causing https://github.com/samvera/hyrax/issues/3046.  I'm wanting to make sure that it wasn't necessary for any of your changes from https://github.com/samvera/hyrax/pull/2128.

PR #2828, opened 2018-03-21
add facet for collections to dashboard -> works index

### LINKED ISSUE #2602: Collections filter on Dashboard -> Works only shows admin sets
### Descriptive summary

The Collections filter on Dashboard -> Works should show both admin sets and collections, but only shows admin sets.

### Rationale

The concept of 'Collections' includes both admin sets and collections.

### Expected behavior

The Collections filter on the works index page should list admin sets and collections.

### Actual behavior

The Collections filter on the works index page lists only admin sets.

Prereq: Create several works and several collections.  Put the works in one of the collections

1. Dashboard -> Works
1. Click Collections filter

Verify that admin sets and collections are listed in the filter selection.

### Related work

PR #2564 Clean up collection type filter
PR #2582 All/Managed Collections should not show Type filter

elrayle:
Fixes #2602

This is not ideal.  It would be better if Admin Sets where listed in the same filter as the Collections.  But creating that filter is very complex and would likely require a kludge because Admin Sets and Collections are not stored in the same solr field.

Also, it would be even better if the facet was a pivot on collection type and collection/admin set name.  But that is even more complex.

For Hyrax 2.1 release, I recommend going with this implementation.  For a later release, we can explore putting in a more complex filter.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/my/works_controller.rb
+++ b/app/controllers/hyrax/my/works_controller.rb
@@ -1,6 +1,15 @@
 module Hyrax
   module My
     class WorksController < MyController
+      # Define collection specific filter facets.
+      def self.configure_facets
+        configure_blacklight do |config|
+          config.add_facet_field solr_name("admin_set", :facetable), limit: 5
+          config.add_facet_field solr_name('member_of_collections', :symbol), limit: 5
+        end
+      end
+      configure_facets
+
       class_attribute :create_work_presenter_class
       self.create_work_presenter_class = Hyrax::SelectTypeListPresenter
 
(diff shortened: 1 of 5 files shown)


elrayle:
@vantuyls @julesies Can you confirm that this is sufficient for 2.1?  The filter can be improved later, but seems functional enough for the 2.1 release.

vantuyls:
i'm okay with this as long as the issue is logged and we can plan to implement the fix at a later date. Possibly around the 3.0 release, where breaks and migrations will be necessary anyway?

julesies:
agree it's fine!

Issue #2792, opened 2018-03-09
Deactivating embargo makes Works private


chrisdaaz:
### Reporting for

- Testing for Hyrax 2.1.0 beta 1
- Browser: Firefox 58.0.2 (64-bit)
- Repository Administration (RA 1.18 and RA 1.21)

### Descriptive Summary
I have logged in as a Repository Administrator and navigated to the "Manage Embargoes" page. I would like to deactivate an embargo on a Generic Work, which the "Visibility will change to" column says will change to "Institution" after the embargo is deactivated

![image](https://user-images.githubusercontent.com/24395592/37212413-b043f24c-2374-11e8-95b4-0fa5151d928e.png)

After clicking the "Deactivate Embargo" button and the "Yes please" button, the visibility of the file is changed to "Private", rather than "Institution".

### Expected behavior

Visibility of embargoed work should change to the visibility setting indicated in the "Visibility will change to" column after the embargo is deactivated.

### Related work

moved from https://github.com/curationexperts/nurax/issues/205

laritakr:
@chrisdaaz can you go to that work and edit/save it (no changes required) and see if the visibility is as expected after a save?

laritakr:
The expected behavior of deactivating a lease or embargo is to remove any future changes to the visibility. 

Embargo and Lease deactivation always leaves the work in the current visibility state.
- if deactivated prior to expiration date, stays at visibility during embargo or lease
- if deactivated after expiration date, stays at visibility after embargo or lease

This is not a behavioral change from the current release, and the release candidate works as expected based on my testing. This will not be fixed unless you can show behavior different than I just documented.

elrayle:
@julesies @chrisdaaz Can you confirm based on the analysis that this isn't really a bug?

julesies:
We continued this conversation on the nurax version of the ticket: see https://github.com/curationexperts/nurax/issues/205#issuecomment-373831221

PR #2830, opened 2018-03-22
Solve errors with visibility & reindexing

laritakr:
Fixes https://github.com/samvera/hyrax/issues/2792

- expired tab options all checked by default (per request in slack channel)
- rename column heading "Work" to "Item", since list also includes files (request in slack channel)
- add conditions to collection_nesting to prevent nested reindexing until changes are ready
- remove additional reindexing that was added previously to handle visibility changes
- add ability to pass options to factory for building leases and embargoes
- add lease spec which failed prior to changes to verify changes

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/collection_nesting.rb
+++ b/app/models/concerns/hyrax/collection_nesting.rb
@@ -13,8 +13,20 @@ module Hyrax
       define_model_callbacks :update_index, only: :after
       after_update_index :update_nested_collection_relationship_indices
       after_destroy :update_child_nested_collection_relationship_indices
+      before_save :before_update_nested_collection_relationship_indices
+      after_save :after_update_nested_collection_relationship_indices
+
+      def before_update_nested_collection_relationship_indices
+        @during_save = true
+      end
+
+      def after_update_nested_collection_relationship_indices
+        @during_save = false
+        Hyrax.config.nested_relationship_reindexer.call(id: id)
+      end
 
       def update_nested_collection_relationship_indices
+        return if @during_save
         Hyrax.config.nested_relationship_reindexer.call(id: id)
       end
 
--- a/spec/actors/hyrax/actors/lease_actor_spec.rb
+++ b/spec/actors/hyrax/actors/lease_actor_spec.rb
@@ -35,4 +35,20 @@ RSpec.describe Hyrax::Actors::LeaseActor do
       end
     end
   end
+
+  context 'deactivating an expired lease', clean_repo: true do
+    let(:lease_attributes) do
+      { lease_date: Date.tomorrow.to_s,
+        current_state: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC,
+        future_state: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED }
+    end
+    let(:leased_work) { create(:leased_work, with_lease_attributes: lease_attributes) }
+    let(:subject) { described_class.new(leased_work) }
+
+    it 'destroys and reindexes the new permission appropriately in solr', with_nested_reindexing: true do
+      allow(leased_work.lease).to receive(:active?).and_return false
+      subject.destroy
+      expect(::SolrDocument.find(leased_work.id)[:visibility_ssi]).to eq(Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED)
+    end
+  end
 end
--- a/spec/factories/generic_works.rb
+++ b/spec/factories/generic_works.rb
@@ -88,31 +88,95 @@ FactoryBot.define do
     end
 
     factory :with_embargo_date do
+      # build with defaults:
+      # let(:work) { create(:embargoed_work) }
+
+      # build with specific values:
+      # let(:embargo_attributes) do
+      #   { embargo_date: Date.tomorrow.to_s,
+      #     current_state: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE,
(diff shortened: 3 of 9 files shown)


spec/actors/hyrax/actors/lease_actor_spec.rb
```diff
@@ -35,4 +35,24 @@
...
+    let(:lease_attributes) do
+      { lease_date: Date.tomorrow.to_s,
+        current_state: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC,
+        future_state: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_AUTHENTICATED }
+    end
+    let(:leased_work) { create(:leased_work, with_lease_attributes: lease_attributes) }
+    let(:subject) { described_class.new(leased_work) }
+
+    before do
+      leased_work
```
cjcolvar: I don't think you need this since `leased_work` will be called first thing on line 53.  If that wasn't the case, you could make line 45 be `let!` which would force the loading of `leased_work`.

spec/models/concerns/hyrax/collection_nesting_spec.rb
```diff
@@ -15,6 +17,13 @@ def update_index
           true
         end
 
+        def before_save
+          false
```
cjcolvar: Should this be true like the return values for the other callbacks in this class?

spec/factories/generic_works.rb
```diff
@@ -88,31 +88,95 @@
...
+          else
+            work.apply_embargo(evaluator.embargo_date,
+                               evaluator.current_state,
+                               evaluator.future_state)
+          end
+        end
       end
       factory :embargoed_work_with_files do
-        after(:build) { |work, evaluator| work.apply_embargo(evaluator.embargo_date, evaluator.current_state, evaluator.future_state) }
+        after(:build) do |work, evaluator|
```
cjcolvar: If this is doing the same thing as the factory above, how about pulling it up into the parent factory `:with_embargo_date` to DRY it up?  It looks like the same could be done for the `:with_lease_date` factory below.

app/models/concerns/hyrax/collection_nesting.rb
```diff
@@ -13,8 +13,20 @@ module CollectionNesting
...
       after_destroy :update_child_nested_collection_relationship_indices
+      before_save :before_update_nested_collection_relationship_indices
+      after_save :after_update_nested_collection_relationship_indices
+
+      def before_update_nested_collection_relationship_indices
+        @during_save = true
+      end
+
+      def after_update_nested_collection_relationship_indices
+        @during_save = false
```
cjcolvar: Even though I suggested it, I wonder if this is entirely safe to do.  Would it be possible for the save callback chain to terminate before running the after_save callback and leave the object in a bad state such that subsequent `.update_index` calls on it wouldn't fire the nested reindexing?  I'd feel more comfortable if someone with a lot of rails experience could review this.  @jeremyf @jcoyne @no-reply

spec/models/concerns/hyrax/collection_nesting_spec.rb
laritakr: no, making this true would cause the other specs to not run. But I am adding a new test to verify that this being true causes the nesting reindexer to not get called.

app/models/concerns/hyrax/collection_nesting.rb
jeremyf: In reviewing the documentation (http://guides.rubyonrails.org/active_record_callbacks.html#available-callbacks) and thinking through this, I'd imagine we could enter a state in which the object fails to save (e.g. Fedora went away). In this case, the instance variable wouldn't matter, as we failed to write to Fedora and there are far greater problems. I see it key that the validations have already run before we hit the `before_save` callback.

### Creating an Object
* before_validation
* after_validation
* before_save
* around_save
* before_create
* around_create
* after_create
* after_save
* after_commit/after_rollback

### Updating an Object
* before_validation
* after_validation
* before_save
* around_save
* before_update
* around_update
* after_update
* after_save
* after_commit/after_rollback

jeremyf: To directly answer @cjcolvar, I think the solution is an improvement on the current state. I appreciate keeping an eye open for possible data integrity issues.

spec/factories/generic_works.rb
cjcolvar: @laritakr and I chatted and agreed that my suggestion is a refactor that is outside the scope of this PR.

app/models/concerns/hyrax/collection_nesting.rb
laritakr: I did quite a bit of debugging through this as part of UI testing. In addition to testing the embargoes and leases, I also I tried the persistence service, reindex_all!, and the visibility changes in all of the areas where we had previously added update_index. In each case, this worked quite well.

Creating a new collection does update_index on the collection three times. Reindexer is now called once rather than three times.

Creating a new work did update_index on the work at least six times (it was hard to keep track!) Previously this meant 6 nested reindexer calls. It now does two as part of the after_save callbacks, and two others which were apparently separate update_index callbacks not part of a save.

The persistence service is still a bit problematic. `child.save` after the member_of_collections change is needed only to force a final update_index. The save itself is unnecessary. The process that happens is:
- `child.member_of_collections << parent` adds the ldp contains member_of_collections to the child. 
- parent is then indexed, and the nesting reindexer triggers from it. (This is unnecessary, as there are no changes to parent, but that's beside the point)
- after this parent reindex, the ldp contains relationship is added to the member_of_collections object
At this point, everything is complete in Fedora. However child never reindexes at all, so the subsequent `child.save` is what forces that reindex. `hyrax/app/services/hyrax/collections/nested_collection_persistence_service.rb` could substitute update_index on the child instead of `child.save`

This still seems like an error to me, and something that needs to be discussed further.

PR #2831, opened 2018-03-22
Provide more leniency for a missing database connection during initialization

mbklein:
### Summary
Rails should be allowed to start even when Hyrax can't persist its roles during initialization.

### Detailed Description
When developing using a non-sqlite DB backend (e.g., Postgres), the `Hyrax::Engine` initializer will throw a connection-related exception if the server isn't running or the database hasn't been created. The same "warn but keep going" strategy should apply that Hyrax has been using to catch `ActiveRecord::StatementInvalid` exceptions during startup.

The reason for rescuing `StandardError` instead of an ActiveRecord-specific exception is that the exact exception class raised may vary by driver (e.g., `PG::ConnectionBad`).

* Attempt to establish a database connection before attempting to persist roles
* Fail gracefully with a warning if a connection cannot be esablished
* Continue on with old behavior

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/engine.rb
+++ b/lib/hyrax/engine.rb
@@ -32,10 +32,28 @@ module Hyrax
     end
 
     config.after_initialize do
-      begin
+      # Attempt to establish a connection before trying to do anything with it. This has to rescue
+      # StandardError instead of, e.g., ActiveRecord::ConnectionNotEstablished or ActiveRecord::NoDatabaseError
+      # because we can't be absolutely sure what the specific database adapter will raise. pg, for example,
+      # raises PG::ConnectionBad. There's no good common ancestor to assume. That's why this test
+      # is in its own tiny chunk of code – so we know that whatever the StandardError is, it's coming
+      # from the attempt to connect.
+      can_connect = begin
+        ActiveRecord::Base.connection
+        true
+      rescue StandardError
+        false
+      end
+
+      can_persist = can_connect && begin
         Hyrax.config.persist_registered_roles!
         Rails.logger.info("Hyrax::Engine.after_initialize - persisting registered roles!")
+        true
       rescue ActiveRecord::StatementInvalid
+        false
+      end
+
+      unless can_persist
         message = "Hyrax::Engine.after_initialize - unable to persist registered roles.\n"
         message += "It is expected during the application installation - during integration tests, rails install.\n"
         message += "It is UNEXPECTED if you are booting up a Hyrax powered application via `rails server'"


lib/hyrax/engine.rb
```diff
@@ -32,10 +32,22 @@ class Engine < ::Rails::Engine
     end
 
     config.after_initialize do
-      begin
+      can_connect = begin
+        ActiveRecord::Base.connection
+        true
+      rescue StandardError
```
jcoyne: Can we use `ActiveRecord::ConnectionNotEstablished`, `ActiveRecord::NoDatabaseError` here?

lib/hyrax/engine.rb
```diff
@@ -32,10 +32,22 @@ class Engine < ::Rails::Engine
...
+        false
+      end
+
+      can_persist = can_connect && begin
         Hyrax.config.persist_registered_roles!
         Rails.logger.info("Hyrax::Engine.after_initialize - persisting registered roles!")
+        true
       rescue ActiveRecord::StatementInvalid
```
jcoyne: What makes this invalid?

lib/hyrax/engine.rb
mbklein: Nope. If using Postgres, the error raised (and not trapped by ActiveRecord) is `PG::ConnectionBad`. I tried but couldn't identify a good error or common ancestor to trap aside from `StandardError`.

lib/hyrax/engine.rb
mbklein: Nothing makes this invalid. The old behavior was to call `Hyrax.config.persist_registered_roles!` and catch `ActiveRecord::StatementInvalid`. I'm just using `true` and `false` as an indicator of whether persistence went off without a hitch (under the pre-PR rubric).

The warning used to be logged in the rescue clause itself. Now I'm using a two-stage success/failure boolean to determine if the warning should be logged.

mbklein: Sorry, I misread the question (and the line it was directed at). I believe this catch is what allows `db:migrate` to happen. If the table doesn't exist yet, this would raise `ActiveRecord::StatementInvalid` on load. It's part of the preexisting code, though.

jcoyne: Okay, that makes sense.

lib/hyrax/engine.rb
mbklein: Done.

Issue #2592, opened 2018-01-31
Errors and notices missing when nesting collections


laritakr:
### Descriptive summary

When nesting collections, the notices and errors do not display when adding or removing a parent collection, or adding or removing a subcollection. The notices are filled correctly and passed with the redirect, but are apparently eaten by Turbolinks. (Notices work appropriately for adding a subcollection as a new collection.)

The [response after submitModalAjax](https://github.com/samvera/hyrax/blob/master/app/assets/javascripts/hyrax/collections.js#L16) is:
```
Turbolinks.clearCache()
Turbolinks.visit(“http://localhost:3000/dashboard/collections/parent_nested?locale=en“, {“action”:“replace”})
```

Turbolinks needs to be deactivated somehow, to allow the responses to use the [controller redirect_to](https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/dashboard/nest_collections_controller.rb).

### Rationale

It is important to have a confirmation of what is happening, particularly in the case where a collection cannot be nested because it would exceed the specified nesting depth. In this situation, nothing happens but there is no message to explain why. At least in the case of a successful nesting, you can see the nested collection on the new view.

### Expected behavior

- In cases of valid nesting, a confirmation notice states that the colllections were nested.
- In cases of errors, one or more errors state why the collections cannot be nested.

### Actual behavior

- The nesting completes normally through the UI and all specs pass.
- in `/app/controllers/hyrax/dashboard/nest_collections_controller.rb`, the appropriate notices and errors are loaded as expected.
- Confirmation notices do not appear after the redirect.
- A notice regarding the nesting depth does not appear after the redirect.
(since the query service was expanded to do appropriate validation, it would be very unlikely that you would be able to produce the other flash errors via the UI without forcing an error in method `valid_combined_nesting_depth?` in `/app/services/hyrax/collections/nested_collection_query_service.rb`)

### Steps to reproduce the behavior

- Create several collections of the same nestable collection type.
- Go to the show page for a collection.
- Select `add to collection`, choose a collection, and submit.
- The page redisplays and shows the newly linked parent.
- No confirmation notice appears.
- Select `add a subcollection`, choose a collection, and submit.
- The page redisplays and shows the newly linked child.
- No confirmation notice appears.
- Chain collections 5 nesting levels deep (`"nesting_collection__deepest_nested_depth_isi":5`)
- Select `Create new collection as subcollection`.
- Page redisplays, with no changes and no error notice.

[...]

PR #2832, opened 2018-03-22
Change nested forms in collections > edit > sharing tab to ajax to fix broken submission

### LINKED ISSUE #2619: Collections > Edit > Sharing: Add group form element not displaying
### Descriptive summary

See screenshot for more detail.  On the Collections > Edit > Sharing tab, the row for "Add group" is a `<form>` (similar to "Add user"), but the Add group `<form>` element is not rendering.  It could be due to using an embedded form, perhaps?   Or some type of misconfiguration.  The scope of this issue could be to confirm Add group is not working as expected, and discover the cause while fixing as-is.  

If it's a nesting form causing the functionality not to work, then a new issue should be created to remove the nested forms for Add group and Add user.

### Rationale

Add group does not currently work?  Nested forms are not legal HTML5.

edit-collection-sharing-nested-forms

@elrayle Issue added.  Feel free to add/label as you see fit.

adamjarling:
Fixes #2619 

Fixes broken form submission resulting from nested "simple_forms", through javascript.  Blocks the default form submit event, serializes the form (or form-wrapper element) data, then makes an AJAX post.

Also did a little tweaking to the UI:
- Each share table columns is now equal width.
- The Remove button in each share table row is now a `class="btn-sm"`, to be consistent with other buttons in display table rows.

I made the following comment next to the javascript form button click listener, so other devs can hopefully understand what's going on:

**Notes:**
  This is a workaround for a scoping issue with 'simple_form' and nested forms in the  'Edit Collections' partials.  All tabs were wrapped in a 'simple_form'. Nested forms, for example inside a tab partial, have behaved erratically, so the pattern has been to remove nested form instances for relatively non-complex forms and replace with AJAX requests.  For this instance of Add Sharing > Add user and Add group, seem more complex in how the form is built from '@form.permission_template', so since it's not working, but the form is already built, this code listens for a click event on the nested form submit button, prevents Default submit behavior, and manually makes the form post.

#### Existing issues resulting from the workaround
The alert messaging which displays when a user or group is removed from a row doesn't display when adding.  Perhaps turbolinks is eating these as well?  I recall @laritakr was looking into this a few weeks back, and if there was any solution this would be a good place to see if it can be applied.

#### Future work
Wrapping the entire Collections > Edit tabs in a simple_form might not be the best approach, and scope collisions have been occurring frequently between nested forms.  In the future, a better pattern might be to place a click listener via javascript on the overall wrapper "form"... which shouldn't be a `<form>` element, and then each tab could have as many forms as it wants.

### DIFF
--- a/app/assets/javascripts/hyrax/collections.js
+++ b/app/assets/javascripts/hyrax/collections.js
@@ -311,8 +311,46 @@ Blacklight.onLoad(function () {
     submitModalAjax(url, 'POST', data, $(this));
   });
 
+
   // Handle add a subcollection button click on the collections show page
   $('.sub-collections-wrapper button.add-subcollection').on('click', function (e) {
     $('#add-subcollection-modal-' + $(this).data('presenterId')).modal('show');
+  });
+
+
+  // Edit Collection: Sharing tab: Add Sharing section click handlers
+  /*
+  Notes:
+  This is a workaround for a scoping issue with 'simple_form' and nested forms in the
+  'Edit Collections' partials.  All tabs were wrapped in a 'simple_form'. Nested forms, for example inside a tab partial,
+  have behaved erratically, so the pattern has been to remove nested form instances for relatively non-complex forms
+  and replace with AJAX requests.  For this instance of Add Sharing > Add user and Add group, seem more complex in how
+  the form is built from '@form.permission_template', so since it's not working, but the form is already built, this
+  code listens for a click event on the nested form submit button, prevents Default submit behavior, and manually makes
+  the form post.
+   */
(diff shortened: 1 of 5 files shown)


elrayle:
@laritakr Can you review and comment with regards to messages being swallowed?

laritakr:
If it's using ajax requests, then I'm almost certain the swallowed messages are the same as https://github.com/samvera/hyrax/issues/2592. I haven't been able to find any way to fix these. Hopefully someone who knows Turbolinks can figure out something.

elrayle:
Remaining Work:
* work described in Future Work to refactor the forms is covered by Issue #2838 
* problem with messages being swallowed has been added to Issue #2592

PR #2834, opened 2018-03-23
show bibliographic_citation in show and edit views

kdid:
There is a `bibliographic_citation` property defined in `Hyrax::BasicMetadata` but it is not exposed in the edit form or display of work. 

https://github.com/samvera/hyrax/blob/1-0-stable/app/models/concerns/hyrax/basic_metadata.rb#L64

* Display `bibliographic_citation` in edit form and on work show page.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/work_form.rb
+++ b/app/forms/hyrax/forms/work_form.rb
@@ -23,7 +23,7 @@ module Hyrax
 
       self.terms = [:title, :creator, :contributor, :description,
                     :keyword, :license, :rights_statement, :publisher, :date_created,
-                    :subject, :language, :identifier, :based_near, :related_url,
+                    :subject, :language, :identifier, :based_near, :related_url, :bibliographic_citation,
                     :representative_id, :thumbnail_id, :rendering_ids, :files,
                     :visibility_during_embargo, :embargo_release_date, :visibility_after_embargo,
                     :visibility_during_lease, :lease_expiration_date, :visibility_after_lease,
--- a/app/models/concerns/hyrax/solr_document/metadata.rb
+++ b/app/models/concerns/hyrax/solr_document/metadata.rb
@@ -66,6 +66,7 @@ module Hyrax
         attribute :source, Solr::Array, solr_name('source')
         attribute :date_created, Solr::Array, solr_name('date_created')
         attribute :rights_statement, Solr::Array, solr_name('rights_statement')
+        attribute :bibliographic_citation, Solr::Array, solr_name('bibliographic_citation')
 
         attribute :mime_type, Solr::String, solr_name('mime_type', :stored_sortable)
         attribute :workflow_state, Solr::String, solr_name('workflow_state_name', :symbol)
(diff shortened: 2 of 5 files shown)


ruthtillman:
Primarily ++ing @vantuyls ... this was an issue we encountered at ND, where we had a very nice citation generator set up on test sites but we encountered primarily problem 1, at times problem 2, and also sometimes an additional problem that a Work Type broad... so two things falling under the same Work Type might need slightly different citations for them to be good enough to just copy-paste into a bibliography.

julesies:
is this the same MD that is already exposed here: https://nurax.curationexperts.com/works/707957706/citation?locale=en

vantuyls:
@no-reply @kdid thoughts on @julesies question?

i suspect it is the same metadata

shlake:
For #1, if you have the appropriate fields for a citation & make those required, then you have a valid citation.
But then for #2, local mods could not (should not) remove the required "citation" fields (or the fact that they are "required").

no-reply:
@julesies: 
> is this the same citation that is already exposed here https://nurax.curationexperts.com/works/707957706/citation?locale=en

No. This citation is an unrelated metadata term. On the object you've linked above, it is unpopulated: https://nurax.curationexperts.com/concern/generic_works/707957706.ttl

vantuyls:
@no-reply @julesies but is it the same metadata used to construct the citation? Thats how i interpreted that question. Julie?

julesies:
I was wondering both because if the concern is bad MD, bad citation then the citation buttons are also problem. I didn't understand if this as a auto-generated field or a user generated value. in the case of user generated there may be different citations for the same work. So is the use case "I don't want to use the citation button" or "I want to create my own citations"?

no-reply:
@vantuyls, @julesies: As far as I can tell, the citations that display throughout the app are unrelated to this field entirely. In a typical Hyrax app (and in Nurax) this field is normally empty for objects submitted through the UI forms.

no-reply:
@vantuyls Do we need to put this on an upcoming Samvera Tech and/or Metadata WG agenda?

kdid:
@no-reply, @vantuyls & @julesies 

Thanks for your comments. When I submitted this PR I made two assumptions: 1) that this `bibliographic_citation` field was not related to the other citations that appear in the Hyrax app and that 2) it was just a mistake or oversight that it wasn't exposed in the form. 

From your discussion it sounds like I was wrong on the second assumption? It sounds like it doesn't make sense that there would be a user entered bibliographic citation, that would be something that would be generated? 

I probably need to get more info from our metadata specialist on how she plans to use this field, but if she wants to use it differently than core Hyrax, we can just go ahead and override it locally - so maybe best if I close this PR?

I do think some documentation might help here on the intended use of this field in Hyrax.

jenyoung:
I'm @kdid's metadata specialist. I had the same assumption that she did - that this was more of a citation for the work a digitized object **came from** - like an image from a book.  `dcterms:source` is a better fit actually, but without a clear definition of the difference between the two and how Hyrax uses them I wasn't sure which way to go.

no-reply:
Closing this for now, pending some recommendation from @samvera/hyrax-metadataists and/or @vantuyls

PR #2809, opened 2018-03-15
Works table server

lfarrell:
Fixes #2779 ; refs #2541

Creates dataTable with server side processing

There are too many works to hold in browser memory, so process all sorts and filters for works dataTable on the server.

@samvera/hyrax-code-reviewers

PR #2839, opened 2018-03-23
Updates based on dev comments

lfarrell:
Fixes for analytics-sprint PR

Fixes for analytics-sprint PR, based on dev comments

@samvera/hyrax-code-reviewers


config/routes.rb
```diff
@@ -104,14 +104,17 @@
...
+    get 'repository_growth', action: :repository_growth
+    get 'repository_object_counts', action: :repository_object_counts
+  end
+
   # Dashboard page
   resource :dashboard, controller: 'dashboard', only: [:show]
   resources :dashboard, only: [] do
     collection do
       get 'activity', action: :activity, as: :dashboard_activity
-      get 'update_collections', action: :update_collections, as: :update_collections
```
jcoyne: Does this just disappear?

app/assets/javascripts/hyrax/admin/analytics_stats.js
```diff
@@ -0,0 +1,25 @@
+Blacklight.onLoad(function() {
```
jcoyne: Rather than doing `Blacklight.onLoad`, would you mind adding something in https://github.com/samvera/hyrax/blob/master/app/assets/javascripts/hyrax/app.js.erb this is the newer way we're trying to initialize all of the javascript in the Hyrax application.

config/routes.rb
lfarrell: It was changed to use a dataTable and having everything run in the browser. Before you could toggle between several states, that required hitting solr, and hence the update_collections route, which

app/services/hyrax/works_count_service.rb
```diff
@@ -10,6 +11,7 @@ def search_results_with_work_count(access)
       works = search_results(access)
 
       works.map do |work|
+        next if work['system_create_dtsi'].nil?
```
jcoyne: I don't understand what the purpose of this is. In what situation would a record not have this? Can you document that? Is there a test for this?

spec/features/create_work_admin_spec.rb
```diff
@@ -28,17 +28,19 @@
...
-      click_link "Add new work"
-      choose "payload_concern", option: "GenericWork"
-      click_button 'Create work'
-      click_link "Relationship" # switch tab
-      expect(page).to have_content('Administrative Set')
-      expect(page).to have_content('Second Admin Set')
-      expect(page).to have_content('Default Admin Set')
-      expect(page).to have_selector('select#generic_work_admin_set_id')
-      expect(page).to have_select('generic_work_admin_set_id', selected: 'Default Admin Set')
+      pending("Admin sets no longer appear on the admin dashboard in the analytics rewrite. Not sure where admin set page will move to")
```
jcoyne: Why does this test change in this Pull request. I'm not seeing the logic for this change here.  Am I missing it?

lfarrell: At least in our analytics mockups. Admin Sets don't appear on the dashboard page for admins anymore. The test seemed like it was testing for the presence of something that's not going to be there.

app/services/hyrax/works_count_service.rb
lfarrell: Honestly, I'm not sure it does need it. I was getting periodic build failures rais 5.0.6 for this:
```
 1) The homepage as an admin shows featured works that I can sort
     Failure/Error: created_date = DateTime.parse(work['system_create_dtsi']).in_time_zone.strftime("%Y-%m-%d")
     
     TypeError:
       no implicit conversion of nil into String
     # ./app/services/hyrax/works_count_service.rb:14:in `parse'
     # ./app/services/hyrax/works_count_service.rb:14:in `block in search_results_with_work_count'
     # ./app/services/hyrax/works_count_service.rb:13:in `map'
     # ./app/services/hyrax/works_count_service.rb:13:in `search_results_with_work_count'
```
As far as I know the homepage should never hit that code.

jcoyne: This seems like a good change, but it also seems like it should be it's own pull request. I'd also like to see this raise an error where it shows the whole document we're getting so we can do a better job of preventing this situation, rather than silently filtering unexpected values.

spec/features/create_work_admin_spec.rb
jcoyne: Did this behavior change in this PR? We shouldn't make test changes in a PR that isn't also changing the code that provides this behavior, because it's confusing if there isn't some correlation between where the tests change and where the behavior changes.

lfarrell:
I'm closing this request. All its changes have been integrated into pull request #2809. Can reopen if it's deemed necessary.

PR #2846, opened 2018-03-26
Only show viewable images with UniversalViewer

### LINKED ISSUE #2802: Image Viewer shows on public image work with private image file
Moved from Nurax: https://github.com/curationexperts/nurax/issues/232

Given that a repo is using the UV, and the work is public, but the primary files is an image and is set to PRIVATE, the image viewer is still showing with a 404 or unauthorized message.

Create a private work
attach 1 file, an image
Set the file to private, leaving the work public
Log out and view the work
See ex: https://nurax.curationexperts.com/concern/images/8p58pc99n?locale=en#?c=0&m=0&s=0&cv=0&xywh=-652%2C-3%2C1755%2C604

image

UPDATE: See comment from Linda that shows private thumbnail is being exposed on a multi-file work.

hortongn:
Fixes #2802 

Only shows the UV on a work when there is at least one attached image that the user has permission to view.

If UV is displayed, the only images loaded into the viewer are files that the user has permission to view.

### DIFF
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -57,10 +57,8 @@ module Hyrax
 
     # @return [Boolean] render the UniversalViewer
     def universal_viewer?
-      representative_id.present? &&
-        representative_presenter.present? &&
-        representative_presenter.image? &&
-        Hyrax.config.iiif_image_server?
+      Hyrax.config.iiif_image_server? &&
+        members_include_viewable_image?
     end
 
     # @return FileSetPresenter presenter for the representative FileSets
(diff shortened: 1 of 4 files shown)


elrayle:
With a quick look at this, it looks good.  My only hesitation is on the final condition you lay out where the "Download image" link is not shown if the user can't view the representative image.  I'm concerned about UX where sometimes the user can download and sometimes they can't without there being some indicator of why.

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -218,5 +216,12 @@ def member_of_authorized_parent_collections
         # member_of_collection_ids with current_ability access
         @member_of ||= Hyrax::CollectionMemberService.run(solr_document, current_ability).map(&:id)
       end
+
+      def members_include_viewable_image?
+        file_set_presenters.each do |presenter|
```
no-reply: What do you think about `file_set_presenters.any? { |presenter| presenter.image? && current_ability.can?(:read, presenter.id) }`? It took me a beat to figure out that the current code would detect a presenter meeting the conditions.

hortongn:
@elrayle merging is blocked again after I pushed the refactor @no-reply suggested.

PR #2855, opened 2018-03-27
Fail eagerly when adding invalid participants to collection types

no-reply:
`Hyrax::CollectionTypes::CreateService#add_participants` previously added participants except those incomplete data. Incomplete participants were logged, but otherwise silently ignored.

This reworks the method to fail eagerly when bad data is encountered. The logging is kept with slightly different error messages. A typed error (`Hyrax::CollectionTypes::CreateService::InvalidParticipantError`) is raised.

This replaces #2666.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/collection_types/create_service.rb
+++ b/app/services/hyrax/collection_types/create_service.rb
@@ -123,23 +123,46 @@ module Hyrax
         add_participants(collection_type_id, default_participants)
       end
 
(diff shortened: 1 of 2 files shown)


jcoyne:
Looks like it needs a rebase.

no-reply:
@jcoyne rebased.

PR #2859, opened 2018-03-28
Don't hit external services for view tests

jcoyne:
This makes the view tests faster and ensures they are focused on testing
just the code in the views.


elrayle:
@jcoyne There are currently two collection factories:
* collections_factory.rb - The original factory without optimization to avoid writing to Fedora which uses `create(:collection)` and `build(:collection)`
* collections.rb - An optimized factory that allows you to create only the parts you need (e.g. solr doc, permissions, fedora).  This uses `create(:collection_lw)` and `build(:colletion_lw)` with optional parameters that controls what gets built.  See the factory for examples.

Some of the tests were converted to collection_lw, but not all because I didn't have the time and because I wanted to avoid creating a huge PR.  The idea is that tests can be converted over time and then the extra factory will be deleted once all tests are converted.

I think we could do something similar for Admin Sets and Works to get even more efficiency in the tests.

cjcolvar:
@jcoyne There are 3 failing tests which look like they are all the same issue.  Do you think you'll have time to cycle back to this or would you like help finishing it off?

jcoyne:
I'd like to point out that the tests pass on travis, even though the status is indicating as didn't pass, scroll to the end of the logs and search for 

```
inished in 24 minutes 31 seconds (files took 21.22 seconds to load)
3496 examples, 0 failures, 19 pending
```

no-reply:
Travis is so weird. I restarted the build.

PR #2861, opened 2018-03-29
ensures only one set of controls when a work has multiple titles

### LINKED ISSUE #2847: Works with double titles show two edit menus
Copied from:  https://github.com/curationexperts/nurax/issues/234

### Descriptive summary
If you have a work with two titles the edit menu appears twice. 

### Expected behavior
We don't want the edit menu two show twice as it does the same thing.

1. Create a work with two titles
2. go to work to dashboard view and see double menu

screen shot 2018-03-12 at 1 27 56 pm

revgum:
fixes #2847

### DIFF
--- a/app/views/hyrax/base/_work_title.erb
+++ b/app/views/hyrax/base/_work_title.erb
@@ -10,7 +10,9 @@
       <% end %>
     </div>
     <div class="col-sm-4 text-right">
-      <%= render "show_actions", presenter: presenter %>
+      <% if index == 0 %>
+        <%= render "show_actions", presenter: presenter %>
+      <% end %>
     </div>
   </div>
 <% end %>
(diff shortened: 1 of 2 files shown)


app/views/hyrax/base/_work_title.erb
```diff
@@ -10,7 +10,9 @@
       <% end %>
     </div>
     <div class="col-sm-4 text-right">
-      <%= render "show_actions", presenter: presenter %>
+      <% if index == 0 %>
```
no-reply: Do we need the empty `<div>` for additional items, or should the conditional wrap lines 12 and 16 as well?

revgum: I don't think we need the additional divs since the `col-sm-4` renders (in my test) the proper block on the page.

revgum: deleted my previous comment.. I misread your question. I do think we should have the empty div so that they maintain that column/spacing and alignment on the page. If that doesn't really matter, I'm happy to adjust the statement and wrap the whole thing.. LMK @no-reply

PR #2870, opened 2018-03-29
Shows modal when user cannot delete collection

### LINKED ISSUE #1788: Collections: Batch delete should show message if fails
### Descriptive summary

If any selected collection fails to delete, there should be a message stating that some of the collections failed to delete due to lack of permissions.

Users can select any collection they can see in Dashboard -> Collections.  They may not have edit access to all those collections.  If they select collections with read access and click `Delete collections`, only collections with edit access will be deleted.  The read only access collections will silently continue to exist. 

### Rationale

Users should receive feedback when an action they take is not fully completed.

### Expected behavior

When attempting to delete a read only collection or a non-empty admin set, a message should be displayed stating the attempt failed.

> "One or more of the selected collections could not be deleted. You cannot delete a collection for which you have read-only access. In addition, you cannot delete a collection of the type "Administrative set" unless that collection is empty."

Note: Sites can customize the name of admin sets, so get the exact name from the collection type with `machine_id='admin_set'`.

Required changes:
- [ ] show checkbox for all collection types, including admin sets
- [ ] during delete processing, ensure user can delete
  - [ ] for admin-set, check that user has edit_access and the admin set is empty
  - [ ] for all other types, check that user has edit_access
  - [ ] delete all that can be deleted
- [ ] if there were any that could not be deleted, show a dialog with the message text as shown above

### Actual behavior

When attempting to delete a read only collection, no message is displaying stating the attempt failed.

1. Create two users: user 1 (an admin user), user 2 (a regular user)
1. Login as user 1, the admin
1. Create a collection with title 'Managed Collection' and give user 2 'Manage' access to the collection
1. Create a collection with title 'Viewable Collection' and give user 2 'View' access to the collection
1. Login as user 2
1. Dashboard -> Collections -> Managed Collections

- [ ] Verify that 'Managed Collection' and 'Viewable Collection' are both in the list of collections.

1. select all -> Delete collections

- [ ] Verify that 'Managed Collection' is no longer in the list.
- [ ] Verify that 'Viewable Collection' is in the list.
- [ ] Verify that a message stating that not all selected collections could be deleted is displayed as a flash message.  (Exact text in section 'Expected behavior' above.)

1. Login as user 1, the admin
1. Create an empty admin set with title 'Empty Admin Set'
1. Create another admin set with title 'Admin Set with Works'
1. Create a work and put it in 'Admin Set with Works'
1. Dashboard -> Collections -> Your Collections

- [ ] Verify that 'Empty Admin Set' and 'Admin Set with Works' are both in the list of collections.

1. select all -> Delete collections

- [ ] Verify that 'Empty Admin Set' is no longer in the list.
- [ ] Verify that 'Admin Set with Works' is in the list.
- [ ] Verify that a message stating that not all selected collections could be deleted is displayed as a flash message.  (Exact text in section 'Expected behavior' above.)

### Related work

#1510 - PR #1716 - batch delete collections
#1708 UI/UX question regarding checkboxes for AdminSets on Collections index page

afred:
- Changed the render of the inputs in the table to also include data on the
  checkbox that describes whether or not the current user can delete the
  corresponding collection
- Added a new modal for this specific case. This can probably be DRYed up
- Changed the js handler for batch delete. Now checks the selected inputs to see
  if any of them will deny the user, and if so, shows the modal and bails
- Also removed an additional branching in the js that was unreachable. If no
  collections are selected, then the user cannot click the delete collections
  button, so this was unreachable.

Fixes (partial) #1788.

### DIFF
--- a/app/assets/javascripts/hyrax/collections.js
+++ b/app/assets/javascripts/hyrax/collections.js
@@ -240,27 +240,29 @@ Blacklight.onLoad(function () {
     };
     var $deleteWordingTarget = $('#selected-collections-delete-modal .pluralized');
 
-    tableRows.each(function(i, row) {
-      checkbox = $(row).find('td:first input[type=checkbox]');
-      if (typeof checkbox[0] !== "undefined") {
-        if (checkbox[0].checked) {
-          numRowsSelected++;
-        }
-      }
-    });
+    var canDeleteAll = true;
+    var selectedInputs = $('#documents table.collections-list-table tbody tr')
+      // Get all inputs in the table
+      .find('td:first input[type=checkbox]')
+      // Filter to those that are checked
+      .filter((i, checkbox) => checkbox.checked)
 
-    if (numRowsSelected > 0) {
+    var cannotDeleteInputs = selectedInputs.filter((i, checkbox) => (checkbox.dataset.hasaccess === "false"))
+    if(cannotDeleteInputs.length > 0) {
+      // TODO: Can we pass data to this modal to be more specific about which ones they cannot delete?
+      $('#collections-to-delete-deny-modal').modal('show');
+      return;
+    }
+
+    if (selectedInputs.length > 0) {
       // Collections are selected
       // Update singular / plural text in delete modal
-      if (numRowsSelected > 1) {
+      if (selectedInputs.length > 1) {
         $deleteWordingTarget.text(deleteWording.plural);
       } else {
         $deleteWordingTarget.text(deleteWording.singular);
       }
       $('#selected-collections-delete-modal').modal('show');
-    } else {
-      // No collections are selected
-      $('#collections-to-delete-deny-modal').modal('show');
     }
   });
 
--- a/app/views/hyrax/my/collections/_modal_delete_collections_deny.html.erb
+++ b/app/views/hyrax/my/collections/_modal_delete_collections_deny.html.erb
@@ -1,0 +1,14 @@
+<div class="modal fade" id="collection-to-delete-deny-modal" tabindex="-1" role="dialog" aria-labelledby="deny-delete-collection-label">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <form class="deny-delete-collection-form">
+        <div class="modal-body">
+          <p><%= t('hyrax.dashboard.my.action.delete_collections_deny') %></p>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-primary" data-dismiss="modal"><%= t('helpers.action.close') %></button>
+        </div>
+      </form>
+    </div>
+  </div>
+</div>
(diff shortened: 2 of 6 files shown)


cjcolvar:
@afred Can you please add a before/after for these changes?

app/assets/javascripts/hyrax/collections.js
```diff
@@ -240,27 +240,29 @@ Blacklight.onLoad(function () {
...
+    var canDeleteAll = true;
+    var selectedInputs = $('#documents table.collections-list-table tbody tr')
+      // Get all inputs in the table
+      .find('td:first input[type=checkbox]')
+      // Filter to those that are checked
+      .filter((i, checkbox) => checkbox.checked)
+
+    var cannotDeleteInputs = selectedInputs.filter((i, checkbox) => (checkbox.dataset.hasaccess === "false"))
+    if(cannotDeleteInputs.length > 0) {
+      // TODO: Can we pass data to this modal to be more specific about which ones they cannot delete?
```
adamjarling: Regarding this "TODO", since we're re-using one modal partial and the content is being built dynamically, probably the best way to specify data you want to access in the modal is to add it to the table row, or table, or wherever makes sense, `data-` attributes.  Maybe as a stringify'd array or something?

afred:
**After:**

app/views/hyrax/my/collections/_modal_delete_collections_deny.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="modal fade" id="collection-to-delete-deny-modal" tabindex="-1" role="dialog" aria-labelledby="deny-delete-collection-label">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <form class="deny-delete-collection-form">
```
adamjarling: It might not matter and not sure if this was existing code, but I'd check that this `<form>` is not embedded within another form for best practice.  Does the dismiss need to be wrapped in a `<form>`, or is it just styling?

elrayle: Looks like there are 5 modals for collections in the same directory that also use <form> in the same way.  @adamjarling If you have a suggestion for an alternative approach, can you open a refactor issue for all the modals to follow a recommended pattern?

app/assets/javascripts/hyrax/collections.js
```diff
@@ -240,27 +240,29 @@ Blacklight.onLoad(function () {
...
-          numRowsSelected++;
-        }
-      }
-    });
+    var canDeleteAll = true;
+    var selectedInputs = $('#documents table.collections-list-table tbody tr')
+      // Get all inputs in the table
+      .find('td:first input[type=checkbox]')
+      // Filter to those that are checked
+      .filter((i, checkbox) => checkbox.checked)
```
lfarrell: .filter(function(i, checkbox) { return checkbox.checked; })

app/assets/javascripts/hyrax/collections.js
```diff
@@ -240,27 +240,29 @@ Blacklight.onLoad(function () {
...
+
+    var cannotDeleteInputs = selectedInputs.filter((i, checkbox) => (checkbox.dataset.hasaccess === "false"))
```
lfarrell: Change `.filter((i, checkbox) => (checkbox.dataset.hasaccess === "false"))` to `.filter(function(i, checkbox) { return checkbox.dataset.hasaccess === "false"; })`

elrayle:
js comments addressed in PR #2932

PR #2879, opened 2018-03-30
set read_access on admin sets excluding registered and public groups

### LINKED ISSUE #2760: Viewer of an admin set cannot see private works
Moved from: https://github.com/curationexperts/nurax/issues/191

Reporting for
Testing for Hyrax 2.1.0 beta 1
Browser: Chrome Version 64.0.3282.167 (64-bit)
Admin Sets and Workflow testing (AS 1.14)
Descriptive summary
When logged in as a Viewer of an admin set, when going to the URL of the work this error is displayed:

image

In addition, as a viewer, when I click on the admin set in Managed Collections, I get that same error.

elrayle:
Fixes #2760

Sets user read access to all users with Deposit and View access.
Sets group read access to all groups with Deposit and View access EXCEPT registered group (i.e. all users with logins) and public group (i.e. everyone)

### DIFF
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -20,8 +20,9 @@ module Hyrax
     self.admin_set_create_service = AdminSetCreateService
 
     def show
+      add_breadcrumb I18n.t('hyrax.controls.home'), hyrax.root_path
       add_breadcrumb t(:'hyrax.dashboard.title'), hyrax.dashboard_path
-      add_breadcrumb t(:'hyrax.dashboard.my.collections'), hyrax.admin_admin_sets_path
+      add_breadcrumb t(:'hyrax.dashboard.my.collections'), hyrax.my_collections_path
       add_breadcrumb @admin_set.title.first
       super
     end
@@ -111,14 +112,6 @@ module Hyrax
       # initialize the form object
       def form
         @form ||= form_class.new(@admin_set, current_ability, repository)
-      end
-
-      # Overrides the parent implementation so that the returned search builder
-      #  searches for edit access
-      # Instantiates the search builder that builds a query for a single item
-      # this is useful in the show view.
-      def single_item_search_builder
-        single_item_search_builder_class.new(self, :edit).with(params.except(:q, :page))
       end
 
       def action_breadcrumb
--- a/app/models/admin_set.rb
+++ b/app/models/admin_set.rb
@@ -83,12 +83,34 @@ class AdminSet < ActiveFedora::Base
   # Calculate and update who should have edit access based on who
   # has "manage" access in the PermissionTemplateAccess
   def reset_access_controls!
-    # NOTE: This is different from Collections in that it doesn't update read access.  See the notes in services/hyrax/collections/permissions_service.rb
-    update!(edit_users: permission_template.agent_ids_for(access: 'manage', agent_type: 'user'),
-            edit_groups: permission_template.agent_ids_for(access: 'manage', agent_type: 'group'))
+    # NOTE: This is different from Collections in that it doesn't update read access based on visibility.
+    #       See also app/models/concerns/hyrax/collection_behavior.rb#reset_access_controls!
+    update!(edit_users: permission_template_edit_users,
+            edit_groups: permission_template_edit_groups,
+            read_users: permission_template_read_users,
+            read_groups: permission_template_read_groups)
   end
 
   private
+
+    def permission_template_edit_users
+      permission_template.agent_ids_for(access: 'manage', agent_type: 'user')
+    end
+
+    def permission_template_edit_groups
+      permission_template.agent_ids_for(access: 'manage', agent_type: 'group')
+    end
+
+    def permission_template_read_users
+      (permission_template.agent_ids_for(access: 'view', agent_type: 'user') +
+          permission_template.agent_ids_for(access: 'deposit', agent_type: 'user')).uniq
+    end
+
+    def permission_template_read_groups
+      (permission_template.agent_ids_for(access: 'view', agent_type: 'group') +
+        permission_template.agent_ids_for(access: 'deposit', agent_type: 'group')).uniq -
+        [::Ability.registered_group_name, ::Ability.public_group_name]
+    end
 
     def destroy_permission_template
       permission_template.destroy
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -125,20 +125,32 @@ module Hyrax
     # Calculate and update who should have read/edit access to the collections based on who
     # has access in PermissionTemplateAccess
     def reset_access_controls!
-      edit_users = permission_template.agent_ids_for(access: 'manage', agent_type: 'user')
-      edit_groups = permission_template.agent_ids_for(access: 'manage', agent_type: 'group')
-      read_users = permission_template.agent_ids_for(access: 'view', agent_type: 'user') +
-                   permission_template.agent_ids_for(access: 'deposit', agent_type: 'user')
-      read_groups = (permission_template.agent_ids_for(access: 'view', agent_type: 'group') +
-                     permission_template.agent_ids_for(access: 'deposit', agent_type: 'group') +
-                     visibility_group).uniq
-      update!(edit_users: edit_users,
-              edit_groups: edit_groups,
-              read_users: read_users,
-              read_groups: read_groups)
+      update!(edit_users: permission_template_edit_users,
+              edit_groups: permission_template_edit_groups,
+              read_users: permission_template_read_users,
+              read_groups: (permission_template_read_groups + visibility_group).uniq)
     end
 
     private
+
+      def permission_template_edit_users
+        permission_template.agent_ids_for(access: 'manage', agent_type: 'user')
+      end
+
+      def permission_template_edit_groups
+        permission_template.agent_ids_for(access: 'manage', agent_type: 'group')
+      end
+
+      def permission_template_read_users
+        (permission_template.agent_ids_for(access: 'view', agent_type: 'user') +
+          permission_template.agent_ids_for(access: 'deposit', agent_type: 'user')).uniq
+      end
+
+      def permission_template_read_groups
+        (permission_template.agent_ids_for(access: 'view', agent_type: 'group') +
+          permission_template.agent_ids_for(access: 'deposit', agent_type: 'group')).uniq -
+          [::Ability.registered_group_name, ::Ability.public_group_name]
+      end
 
       def visibility_group
         return [Hydra::AccessControls::AccessRight::PERMISSION_TEXT_VALUE_PUBLIC] if visibility == Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
--- a/app/services/hyrax/admin_set_create_service.rb
+++ b/app/services/hyrax/admin_set_create_service.rb
@@ -120,7 +120,7 @@ module Hyrax
 
       # Gives deposit access to registered users to default AdminSet
       def create_default_access_for(permission_template:, workflow:)
-        permission_template.access_grants.create(agent_type: 'group', agent_id: 'registered', access: Hyrax::PermissionTemplateAccess::DEPOSIT)
+        permission_template.access_grants.create(agent_type: 'group', agent_id: ::Ability.registered_group_name, access: Hyrax::PermissionTemplateAccess::DEPOSIT)
         deposit = Sipity::Role[Hyrax::RoleRegistry::DEPOSITING]
         workflow.update_responsibilities(role: deposit, agents: Hyrax::Group.new('registered'))
       end
--- a/app/services/hyrax/collections/migration_service.rb
+++ b/app/services/hyrax/collections/migration_service.rb
@@ -13,6 +13,11 @@ module Hyrax
           migrate_collection(col)
           Rails.logger.info "  migrating collection - id: #{col.id}, title: #{col.title}"
         end
+
+        AdminSet.all.each do |adminset|
+          migrate_adminset(adminset)
+          Rails.logger.info "  migrating adminset - id: #{adminset.id}, title: #{adminset.title}"
+        end
         Rails.logger.info "--- Migration Complete"
       end
 
@@ -30,6 +35,20 @@ module Hyrax
       end
       private_class_method :migrate_collection
 
+      # @api private
+      #
+      # Migrate a single adminset to grant depositors and viewers read access to the admin set unless the grant is for
+      # registered (authenticated users) or public (anyone) groups.  The adjustment is being made to adminsets created
+      # before Hyrax 2.1.0.  Migrating twice will not adversely impact the adminset.
+      #
+      # @param adminset [AdminSet] adminset object to be migrated
+      def self.migrate_adminset(adminset)
+        Hyrax::PermissionTemplateAccess.find_or_create_by(permission_template_id: adminset.permission_template.id,
+                                                          agent_type: "group", agent_id: "admin", access: "manage")
+        adminset.reset_access_controls!
+      end
+      private_class_method :migrate_adminset
+
       # @api public
       #
       # Validate that migrated collections have both the collection type gid assigned and the permission template with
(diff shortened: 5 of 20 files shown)


app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -20,8 +20,9 @@ class Admin::AdminSetsController < ApplicationController
     self.admin_set_create_service = AdminSetCreateService
 
     def show
+      add_breadcrumb I18n.t('hyrax.controls.home'), hyrax.root_path
       add_breadcrumb t(:'hyrax.dashboard.title'), hyrax.dashboard_path
-      add_breadcrumb t(:'hyrax.dashboard.my.collections'), hyrax.admin_admin_sets_path
+      add_breadcrumb t(:'hyrax.dashboard.my.collections'), hyrax.my_collections_path
```
elrayle: Make breadcrumbs match those being setup in the breadcrumbs refactor.

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -113,14 +114,6 @@ def form
...
       end
 
-      # Overrides the parent implementation so that the returned search builder
-      #  searches for edit access
-      # Instantiates the search builder that builds a query for a single item
-      # this is useful in the show view.
-      def single_item_search_builder
-        single_item_search_builder_class.new(self, :edit).with(params.except(:q, :page))
-      end
-
```
elrayle: Remove the restriction that the show page be limited to those with `:edit` access.

app/models/admin_set.rb
```diff
@@ -83,13 +83,35 @@ def active_workflow
...
   def reset_access_controls!
-    # NOTE: This is different from Collections in that it doesn't update read access.  See the notes in services/hyrax/collections/permissions_service.rb
-    update!(edit_users: permission_template.agent_ids_for(access: 'manage', agent_type: 'user'),
-            edit_groups: permission_template.agent_ids_for(access: 'manage', agent_type: 'group'))
+    # NOTE: This is different from Collections in that it doesn't update read access based on visibility.
+    #       See also app/models/concerns/hyrax/collection_behavior.rb#reset_access_controls!
+    update!(edit_users: permission_template_edit_users,
+            edit_groups: permission_template_edit_groups,
+            read_users: permission_template_read_users,
+            read_groups: permission_template_read_groups)
```
elrayle: Add read access for depositors and viewers.  Excludes groups `['registered', 'public']`

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -125,21 +125,33 @@ def permission_template
...
-                     permission_template.agent_ids_for(access: 'deposit', agent_type: 'group') +
-                     visibility_group).uniq
-      update!(edit_users: edit_users,
-              edit_groups: edit_groups,
-              read_users: read_users,
-              read_groups: read_groups)
+      update!(edit_users: permission_template_edit_users,
+              edit_groups: permission_template_edit_groups,
+              read_users: permission_template_read_users,
+              read_groups: (permission_template_read_groups + visibility_group).uniq)
```
elrayle: Make process for collections and admin sets match.

app/models/hyrax/permission_template.rb
```diff
@@ -22,6 +22,19 @@ def agent_ids_for(agent_type:, access:)
...
+    # @param agent_id [String] user id or group name who has access
+    # @param access [String] type of access (e.g. Hyrax::PermissionTemplateAccess::MANAGE, Hyrax::PermissionTemplateAccess::DEPOSIT, Hyrax::PermissionTemplateAccess::VIEW)
+    # @return [Hyrax::PermissionTemplateAccess] the created or found access
+    def find_or_create_access(agent_type:, agent_id:, access:)
+      records = Hyrax::PermissionTemplateAccess.where(permission_template_id: id, agent_type: agent_type, agent_id: agent_id, access: access)
+      return records.first if records.present? # should only be one match
+      access_grants.create(agent_type: agent_type, agent_id: agent_id, access: access)
+    end
+
     # The list of workflows that could be activated; It includes the active workflow
```
elrayle: Allows migration service to try and grant admin's edit access without having to check if it already exists.  If you just try to add it, a not-uniq exception is raised.

laritakr:
@elrayle It looks good to me, but I’d still prefer to let @cjcolvar take a look as well. I especially appreciate the additional use of the lw factories. :) If you need me to officially approve, lmk but otherwise I’m holding off a bit to get other eyes on it since a number of files are involved.

spec/services/hyrax/collections/migration_service_spec.rb
```diff
@@ -36,24 +44,107 @@
...
+
+        Hyrax::Collections::MigrationService.migrate_all_collections
+
+        AdminSet.all.each do |adminset|
+          pt_id = Hyrax::PermissionTemplate.find_by!(source_id: adminset.id)
+          expect(pt_id).not_to be_nil
+          expect_solr_access(pt_id, 'user', :manage, adminset.edit_users)
+          expect_solr_access(pt_id, 'group', :manage, adminset.edit_groups)
+          expect_solr_access(pt_id, 'user', :view, adminset.read_users)
+          expect_solr_access(pt_id, 'group', :view, adminset.read_groups)
```
cjcolvar: Are lines 99-100 supposed to be for depositor user and group?  Currently they are a repeat of lines 101-102.  It looks like this the case in the other new tests below as well for both pre and post migration tests.

spec/services/hyrax/collections/migration_service_spec.rb
```diff
@@ -277,4 +446,26 @@ def expect_no_access(permission_template_id, agent_type, access, agent_ids)
...
+  def expect_solr_access(permission_template_id, pt_agent_type, pt_access, solrdoc_access)
+    ptas = Hyrax::PermissionTemplateAccess.where(permission_template_id: permission_template_id, agent_type: pt_agent_type, access: pt_access)
+    expect_solr_group_access(ptas, solrdoc_access) if pt_agent_type == 'group'
+    expect_solr_user_access(ptas, solrdoc_access) if pt_agent_type == 'user'
+  end
+
+  def expect_solr_group_access(permission_templates, solrdoc_access)
+    permission_templates.each do |pta|
+      if ['registered', 'public'].include? pta.agent_id
+        expect(solrdoc_access).not_to include pta.agent_id
```
cjcolvar: I wasn't expecting to find the negative expectations here.  I would have thought they would be separate and distinct from the positive expectations.  I think they might be more obvious if kept outside of `expect_solr_access` and maybe in a similarly named `expect_no_solr_access`?

spec/views/hyrax/admin/admin_sets/_show_actions.html.erb_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      allow(ability).to receive(:can?).with(:edit, solr_document).and_return(true)
+    end
+
+    context "when presenter has delete disabled" do
+      before do
+        allow(presenter).to receive(:disable_delete?).and_return(true)
+        render
+      end
+      it "displays a disabled delete button" do
+        expect(rendered).to have_selector(:css, "a.btn-danger.disabled")
```
cjcolvar: Do you think this selector will be a problem in the future if more buttons are added to this partial?  Do other tests scope their selectors to the element in question instead of the whole partial?

spec/views/hyrax/admin/admin_sets/_show_actions.html.erb_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    context "with default admin set" do
+      before do
+        allow(presenter).to receive(:disable_delete?).and_return(true)
+        render
+      end
+      it "displays a disabled delete button" do
+        expect(rendered).to have_selector(:css, "a.btn-danger.disabled")
+      end
+    end
+  end
```
cjcolvar: What do you think about adding expectations to the tests in the above context for the presence of the edit button?

spec/factories/admin_sets_lw.rb
```diff
@@ -0,0 +1,215 @@
...
+  #
+  # PREFERRED: Use build whenever possible.  You can control the creation of the permission template and solr document
+  #            by passing parameters to the build(:admin_set_lw) method.  That way you can build only the parts needed
+  #            for a specific test.
+  #
+  # AVOID: Do not use create unless absolutely necessary.  It will create everything including the Fedora object.
+  #
+  # @example Simple build of an admin set with no additional parts created.  Lightest weight.
+  #          NOTE: A user is automatically created as the owner of the admin set.
+  #   let(:adminset) { build(:adminset_lw)
```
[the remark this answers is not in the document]
elrayle: Glad you caught that!

spec/views/hyrax/admin/admin_sets/_show_actions.html.erb_spec.rb
elrayle: I'm not sure I completely understand your question.  I will try to answer what I think you are asking, but please clarify if I am answering the wrong question.

 `have_selector` is in common usage in the tests, although most do not use the `:css` as the first parameter.  Many are used to check for the presence of expected buttons., links, fields, etc.

If other buttons/links are added, it is expected they will not have the same `btn-danger` class.

What I'm not sure how to answer is...  
> Do other tests scope their selectors to the element in question instead of the whole partial?

Can you give an example of scoping to the whole partial as opposed to the element?

cjcolvar: `rendered` here is the partial that is being tested (`hyrax/admin/admin_sets/_show_actions.html.erb`) so I was wondering if it would be good to first find the button element and then make the expectation on that instead of `rendered`.   Maybe something like this https://github.com/samvera/hyrax/blob/master/spec/views/hyrax/citations/work.html.erb_spec.rb#L44.

spec/factory_tests/adminsets_factory_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        expect { build(:adminset_lw, with_permission_template: { manage_users: [user_mgr] }) }.to change { Hyrax::PermissionTemplateAccess.count }.by(2)
+        expect { build(:adminset_lw, with_permission_template: { manage_users: [user_mgr], deposit_users: [user_dep], view_users: [user_vw] }) }
+          .to change { Hyrax::PermissionTemplate.count }.by(1)
+        expect { build(:adminset_lw, with_permission_template: { manage_users: [user_mgr], deposit_users: [user_dep], view_users: [user_vw] }) }
+          .to change { Hyrax::PermissionTemplateAccess.count }.by(4)
+      end
+    end
+
+    context 'with_solr_document' do
+      it 'will not created a solr document by default' do
```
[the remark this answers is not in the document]
elrayle: And also a copy and paste typo, so I fixed it here and in the collections_factory_spec.

app/models/hyrax/permission_template.rb
elrayle: Nice.  I switched to use the standard method.

app/services/hyrax/collections/migration_service.rb
```diff
@@ -30,6 +35,23 @@ def self.migrate_collection(collection)
...
+      # before Hyrax 2.1.0.  Migrating twice will not adversely impact the adminset.
+      #
+      # @param adminset [AdminSet] adminset object to be migrated
+      def self.migrate_adminset(adminset)
+        adminset.permission_template.find_or_create_access(
+          agent_type: Hyrax::PermissionTemplateAccess::GROUP,
+          agent_id: ::Ability.admin_group_name,
+          access: Hyrax::PermissionTemplateAccess::MANAGE
+        )
+        adminset.reset_access_controls!
```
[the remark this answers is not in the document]
elrayle: In 2.0, access to admin sets is granted as follows...
* read access is never granted
* edit access is given to the creating user
* edit access is granted to any managers assigned through Participants

There is no other way through the UI to effect the access to the admin set.

The reset_access_controls! (previously named update_access_controls!) is always called after a save of an admin set, such that the access is overwritten every time the admin set is modified.  This means that ultimately, the access to an admin set is controlled by the PermissionTemplate and PermissionTemplateAccess tables.  This was the case in 2.0 and continues to be the case in 2.1.

The abilities, which looks at the PermissionTemplate/Access and solr read/edit access, answers the questions of can?. 

What made me comment on the necessity of this change is that in 2.0, admin group is granted edit access to everything.  The abilities still should be granting edit access to everything in 2.1.  So I'm not sure what change happened that makes it necessary for the admin group to be specifically granted MANAGE access of the admin set to be able to see it.

I'm going to take one more look at this to see if I can figure out 1) if this is necessary and expected to be necessary, or 2) should not be necessary and represents a bug.

spec/services/hyrax/collections/migration_service_spec.rb
elrayle: Good catch.

spec/services/hyrax/collections/migration_service_spec.rb
elrayle: These are not really testing anything specific to the migration_service.  The conditions expressed here are tested in...
* spec/services/hyrax/collections/permissions_service_spec.rb.source_ids_for_deposit returns admin set ids where user has deposit access except excluded groups

spec/views/hyrax/admin/admin_sets/_show_actions.html.erb_spec.rb
elrayle: 👍

app/services/hyrax/admin_set_create_service.rb
```diff
@@ -120,7 +120,8 @@ def workflow_agents
 
       # Gives deposit access to registered users to default AdminSet
       def create_default_access_for(permission_template:, workflow:)
-        permission_template.access_grants.create(agent_type: 'group', agent_id: 'registered', access: Hyrax::PermissionTemplateAccess::DEPOSIT)
+        # permission_template.access_grants.create(agent_type: 'group', agent_id: ::Ability.admin_group_name, access: Hyrax::PermissionTemplateAccess::MANAGE)
```
cjcolvar: Is there a reason you're adding this commented out line?  Could it be removed?

spec/services/hyrax/collections/migration_service_spec.rb
cjcolvar: It looks like these still need to be changed to :deposit.

spec/services/hyrax/collections/migration_service_spec.rb
```diff
@@ -250,6 +341,84 @@
...
+
+        AdminSet.all.each do |adminset|
+          pt_id = Hyrax::PermissionTemplate.find_by!(source_id: adminset.id)
+          expect(pt_id).not_to be_nil
+          expect_solr_access(pt_id, 'user', :manage, adminset.edit_users)
+          expect_solr_access(pt_id, 'group', :manage, adminset.edit_groups)
+          expect_solr_access(pt_id, 'user', :view, adminset.read_users)
+          expect_solr_access(pt_id, 'group', :view, adminset.read_groups)
+          expect_solr_access(pt_id, 'user', :view, adminset.read_users)
+          expect_solr_access(pt_id, 'group', :view, adminset.read_groups)
```
cjcolvar: Same here.  These still need to be changed to :deposit.

spec/services/hyrax/collections/migration_service_spec.rb
```diff
@@ -250,6 +341,84 @@
...
+
+      it "doesn't change the adminset" do # rubocop:disable RSpec/ExampleLength
+        pt_id = Hyrax::PermissionTemplate.find_by!(source_id: adminset.id).id
+        expect(pt_id).to eq permission_template.id
+        expect_solr_access(pt_id, 'user', :manage, adminset.edit_users)
+        expect_solr_access(pt_id, 'group', :manage, adminset.edit_groups)
+        expect_solr_access(pt_id, 'user', :view, adminset.read_users)
+        expect_solr_access(pt_id, 'group', :view, adminset.read_groups)
+        expect_solr_access(pt_id, 'user', :view, adminset.read_users)
+        expect_solr_access(pt_id, 'group', :view, adminset.read_groups)
```
cjcolvar: Here and right and lines 418-419 as well.

elrayle: Ugh, can't believe I missed these.  Thanks.

PR #2883, opened 2018-04-02
[WIP] add work to collection from work show page.

### LINKED ISSUE #2279: Add an "add to collection" button on work showpage
### Descriptive summary
to facilitate adding content to collections, add an 'add to collection' button to the showpage

blancoj:
Fixes #2279 to add a button to the work show page that adds the work to a collection.


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -68,6 +68,8 @@ def create
     # Finds a solr document matching the id and sets @presenter
     # @raise CanCan::AccessDenied if the document is not found or the user doesn't have access to it.
     def show
+      @user_collections = collections_service.search_results(:deposit)
```
cjcolvar: How about guarding this line so it doesn't run for every work show request since it looks like it only gets used `if presenter.editor?`

app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -13,6 +13,12 @@
           <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
             Attach Child <span class="caret"></span>
           </button>
+
+          <input type="checkbox"  style="display:none" name="batch_document_ids[]" id="batch_document_<%= presenter.id %>" value="<%= presenter.id %>" class="batch_document_selector" checked="checked" />
+          <%= button_tag t('hyrax.dashboard.my.action.add_to_collection'),
+              class: 'btn btn-default submits-batches submits-batches-add',
+              data: { toggle: "modal", target: "#collection-list-container", turbolinks: false } %>
```
elrayle: Try taking off `turbolinks: false`.  There was a previous PR that should make that unnecessary.  So we don't want to add another use of it unless it is required.

app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -35,4 +41,6 @@
 
 <!-- COinS hook for Zotero -->
   <span class="Z3988" title="<%= export_as_openurl_ctx_kev(presenter) %>"></span>
+  <%= render 'hyrax/dashboard/collections/form_for_select_collection', user_collections: @user_collections %>
```
elrayle: Can you add a blank line and a comment before this render with a comment stating `  <!-- Render Modals -->`?

elrayle: I notice that the modal renders with all the content right justified.  The same modal renders left justified when accessed from Dashboard -> Works -> click checkboxes -> Add to Collection.  Can you explore why they render differently?

app/views/hyrax/base/_show_actions.html.erb
blancoj: will try!

app/views/hyrax/base/_show_actions.html.erb
blancoj: got it!

PR #2884, opened 2018-04-02
No rep obj toggle

### LINKED ISSUE #2714: Toggle Repository Object Charts

lfarrell:
Fixes #2714 ; refs #2541

Adds toggling between work type and work visibility

Adds toggling between work type and work visibility for repository object chart on admin dashboard

@samvera/hyrax-code-reviewers


app/presenters/hyrax/admin/repository_object_presenter.rb
```diff
@@ -18,13 +18,17 @@ def as_json(*)
...
         def counts
-          translation = { 'false' => :published, 'true' => :unpublished, nil => :unknown }
+          translation = translation_keys
           raw_count = Hash[*results.to_a.flatten]
           @counts ||= raw_count.each_with_object({}) { |(k, v), o| o[translation[k]] = v }
         end
 
         def search_builder
-          Stats::WorkStatusSearchBuilder.new(self)
+          if @object_type == 'works'
```
mcritchlow: I'm seeing this check for object_type == `works` 3 or so times in this class. Which makes me wonder if it should be a constant in case the name of this default gets changed in the future.

lfarrell:
Good call. Switched it over to a constant

Issue #2798, opened 2018-03-09
NestedCollectionPersistenceService#persist_nested_collection_for is causing unexpected save and reindex of parent.


jgondron:
### Descriptive summary

Nested collections are triggering a save on the parent and should not be. This is causing additional after save/index events on the parent object that are unnecessary. The result is a cascade to reindex to all of that parent's descendants. The effect is that adding a sub-collection or work to a large collection is slow.

### Rationale

Since we are only storing the relationship on the child, modifying the parent should not be necessary. 

### Expected behavior
When https://github.com/samvera/hyrax/blob/master/app/services/hyrax/collections/nested_collection_persistence_service.rb#L15 is called, I would expect it to update the child only.

### Actual behavior
When https://github.com/samvera/hyrax/blob/master/app/services/hyrax/collections/nested_collection_persistence_service.rb#L15 is called, it updates the child, then saves both the child and the parent. This can be traced down to the indirect association insert_record method https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/associations/indirectly_contains_association.rb#L15. Within this context, the "container" is the child collection/work and the "record" is the parent collection.

Here are the relevant bits of a stack trace related with this operation:

```
    #22 ActiveFedora::Associations::IndirectlyContainsAssociation.insert_record(record#Collection, force#TrueClass, validate#TrueClass) at /Users/jgondron/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/active-fedora-11.5.2/lib/active_fedora/associations/indirectly_contains_association.rb:15
    #23 block (2 levels) in ActiveFedora::Associations::CollectionAssociation.block (2 levels) in concat_records(*records#Array) at /Users/jgondron/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/active-fedora-11.5.2/lib/active_fedora/associations/collection_association.rb:162
    #24 ActiveFedora::Associations::CollectionAssociation.add_to_target(record#Collection, skip_callbacks#FalseClass) at /Users/jgondron/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/active-fedora-11.5.2/lib/active_fedora/associations/collection_association.rb:260
[...]

PR #2893, opened 2018-04-04
Replace save with reindex on child collection

laritakr:
References https://github.com/samvera/hyrax/issues/2798

This is a PR to reduce some unnecessary overhead, but it doesn't actually solve the referenced issue.

In the case of adding a `member_of_collections` relationship (either work and collection), Active Fedora saves and reindexes the parent, which triggers the nested reindexer to reindex the parent and ALL of its child works AND collections. This is a lot of unnecessarily overhead because nothing is changing on the parent, and only the one child work or collection actually needs to change.

The nested child persistence service updates the `member_of_collections` relationship between a child and parent collection. Active Fedora makes all of the expected changes to the data in Fedora with no need for a subsequent save. However, it doesn't cause a subsequent reindex on the child collection to record the relationship in Solr. 

This reindexing had previously been accommodated by a subsequent `child.save`. Since all that we really need is to reindex the child, this PR is changing the save to simply force reindexing. It calls the nested reindexer directly, rather than using `update_index` which would trigger the additional call to the reindexer, thereby reindexing the child collection twice.

**Note:** I've been told that the indirect container relationship requires that the parent be saved, though I can't claim to understand why, especially since it only does this when adding the `member_of_collections` relationship, and  doesn't do  it while removing the same relationship. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/collection_nesting.rb
+++ b/app/models/concerns/hyrax/collection_nesting.rb
@@ -22,18 +22,18 @@ module Hyrax
 
       def after_update_nested_collection_relationship_indices
         @during_save = false
-        Hyrax.config.nested_relationship_reindexer.call(id: id)
+        reindex_nested_relationships_for(id: id)
       end
 
       def update_nested_collection_relationship_indices
         return if @during_save
-        Hyrax.config.nested_relationship_reindexer.call(id: id)
+        reindex_nested_relationships_for(id: id)
       end
 
       def update_child_nested_collection_relationship_indices
         children = find_children_of(destroyed_id: id)
         children.each do |child|
-          Hyrax.config.nested_relationship_reindexer.call(id: child.id)
+          reindex_nested_relationships_for(id: child.id)
         end
       end
     end
@@ -49,5 +49,11 @@ module Hyrax
     def use_nested_reindexing?
       true
     end
+
+    private
+
+      def reindex_nested_relationships_for(id:)
+        Hyrax.config.nested_relationship_reindexer.call(id: id)
+      end
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/concerns/hyrax/collection_nesting.rb
```diff
@@ -49,5 +49,9 @@ def find_children_of(destroyed_id:)
     def use_nested_reindexing?
       true
     end
+
+    def reindex_id(id:)
```
cjcolvar: Is this method needed as a public method or is it just to DRY up other methods?  If it needs to be public how about renaming it to something more specific like `reindex_nested_relationships_for(id:)` since it will be visible on all models that include this module?

laritakr: I just did this to DRY other methods.

laritakr:
@cjcolvar The corresponding tests are those for the [nested collection persistence service](https://github.com/samvera/hyrax/blob/5445d49f887bdf347bbe2982d4f3f0ecacd93a89/spec/services/hyrax/collections/nested_collection_persistence_service_spec.rb). Indexing is based on what has been saved in AF without caching. So if the save doesn't happen appropriately, the indexing will also not find the information needed to index both directions.

cjcolvar:
@laritakr I tried out [the tests](https://github.com/samvera/hyrax/blob/5445d49f887bdf347bbe2982d4f3f0ecacd93a89/spec/services/hyrax/collections/nested_collection_persistence_service_spec.rb) locally and they pass even after switching the expectations to `parent.reload` and `child.reload` so I'm convinced.

Issue #2841, opened 2018-03-24
Combine Admin Set and Collections facet and pivot with Collection Types


elrayle:
### Descriptive summary

In Hyrax 2.1.0, Admin Sets are being treated as just another collection type.  Under the covers they are treated differently.  In the solr doc for a work, the admin set is identified in field `admin_set_tesim` and collections are identified in field `member_of_collections_ssim`.

Currently, on Dashboard -> Works, there are two filters: 1) Admin Sets, 2) Collections

To maintain the illusion that Admin Set is just another collection type, these should be displayed in the same filter with the label Collections.  Additionally, if possible, it would be even better to pivot on collection type.  See images in Expected behavior.

### Rationale

Maintain the illusion of admin sets being a type of collection.

### Expected behavior

Minimal goal of admin sets and collections in one filter...

![image](https://user-images.githubusercontent.com/6855473/37863866-ff7aebe0-2f3b-11e8-93c0-9b337a3d1607.png)

Extended goal of pivoting on collection type...

![image](https://user-images.githubusercontent.com/6855473/37864070-cea06cb2-2f3f-11e8-809a-15ebd14d0c43.png)

### Actual behavior

Admin Set filter...

![image](https://user-images.githubusercontent.com/6855473/37863888-6bb20dde-2f3c-11e8-88bb-caaa6e6c3ca3.png)

Collection filter...

![image](https://user-images.githubusercontent.com/6855473/37863881-5ed8a8f2-2f3c-11e8-9489-0aaabe93b58a.png)

### Steps to reproduce the behavior

1. Create multiple collection types
1. Create collections of each type and at least 1 admin set
1. Create works and put a few in each collection and admin set
1. Navigate to Dashboard -> Works

Verify Collection filter includes both collections and admin sets.

### Related work

PR #2828 Add facet for collections on Dashboard -> Works index

PR #2904, opened 2018-04-07
[WIP] Combine collection and admin set facets on works dashboard

hortongn:
Partial resolution for #2841

This satisfies the "minimal goal" of #2841 by combining the `Collection` and `Admin Set` facets on the works dashboard into a single facet.  

Clicking on the `Collection` facet will now list any collections that have works along with any admin sets that have works.  Items in the dropdown are sorted by count.

@elrayle

### DIFF
--- a/app/controllers/hyrax/my/works_controller.rb
+++ b/app/controllers/hyrax/my/works_controller.rb
@@ -4,8 +4,15 @@ module Hyrax
       # Define collection specific filter facets.
       def self.configure_facets
         configure_blacklight do |config|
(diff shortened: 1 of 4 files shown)


elrayle:
Solution does not work.  To see the issue...
* create 2 admin sets
* create 2 collections
* create 8 works with 5 in 1 admin set and 3 in the other
* put at least 1 work from each admin set in the same collection
You will see the collection name repeated, once under each admin set for the work

elrayle:
Closing as this approach will not be used.  I confirmed with @hortongn that he will open up a new PR when the alternate approach he is working on it ready.

PR #2875, opened 2018-03-29
Fix for #2873 - CSRF Exception after adding to Collections/Sharing

jgondron:
- Found the root issue is related to turbolinks + jquery-ujs + a button that uses ajax to post. This may also fix #1191. Adding a call to refresh the CSRF tokens in the rendered DOM after turbolinks loads fixes it.

Related to https://github.com/rails/jquery-ujs/issues/456

@samvera/hyrax-code-reviewers

PR #2908, opened 2018-04-10
Force loading of UV

### LINKED ISSUE #2906: UV does not always show without a browser refresh
moved from https://github.com/curationexperts/nurax/issues/238

### Descriptive summary

It really hard to get this consistently and repeat a pattern. About 7 of 10 times, I have to reload the page to get the UV to show up. When it does not show, it's just blank white. A refresh solves it. When discussed on slack, folks suspected TURBOLINKS.

Here are some links to test:

https://nurax.curationexperts.com/concern/images/qf85nb38d?locale=en#?c=0&m=0&s=0&cv=0&xywh=-85%2C-11%2C444%2C203

https://nurax.curationexperts.com/concern/images/47429921g?locale=en

@vantuyls comment i'm seeing this consistently on Firefox. Seems like the first time i navigate to a UV item (images) it loads fine, but then each subsequent UV item i navigate to requires refresh before UV loads.

cjcolvar:
Fixes #2906.

Set flag to false before loading UV embed javascript (https://github.com/pulibrary/pul_uv_rails/blob/master/public/universalviewer/dist/uv-2.0.1/lib/embed.js#L77-L80) to force loading of UV.  Without this the UV loading script will sometimes think UV is already loaded and skip the loading.  This is probably due to turbolinks keeping the parent window context.

I took this approach of inlining the javascript to keep it close to what it is related to and avoid having to make changes to `pul_uv_rails` or `universalviewer`.  If there is a more elegant solution or if disabling turbolinks is preferred I'd happily close this PR.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/turbolinks_events.js
+++ b/app/assets/javascripts/hyrax/turbolinks_events.js
@@ -2,4 +2,7 @@
 // See https://github.com/rails/jquery-ujs/issues/456
 $(document).on('turbolinks:load', function() {
   $.rails.refreshCSRFTokens();
+  // Explicitly set flag to false to force loading of UV
+  // See https://github.com/samvera/hyrax/issues/2906
+  window.embedScriptIncluded = false;
 });


elrayle:
Does this turbolinks related PR impact this work?  https://github.com/samvera/hyrax/pull/2875

cjcolvar:
I tested again locally with the latest from master and was able to reproduce the issue without the code change using these steps:
1. Perform empty search from homepage
1. Find an image work and click into it
1. Click `Back to Search Results`
1. Find another image work and click into it
1. UV doesn't load and leaves a big white space on the page

Applying this PR made this scenario work.  So I think this change is still needed and not addressed by the previous turbolinks PR.

jgondron:
I wouldn't expect #2875 to directly fix, but it might give you an avenue for better integration of this UV flag with turbolinks globally instead of just on this page. I'm not sure of the order of events here, but you could potentially move your fix into the turbolinks:load event https://github.com/samvera/hyrax/pull/2875/files#diff-da58ad8d67bcab5238c1389809939992R3.

cjcolvar:
@jgondron Great suggestion!  I tested and that worked so I've updated the PR to use that approach.

PR #2911, opened 2018-04-10
Use :test `QueueAdapter` in specs

### LINKED ISSUE #2888: Use `:test` adapter for ActiveJob in specs
### Descriptive summary

The test suite uses `:inline` as the `ActiveJob` adapter throughout. This is expensive and limiting for testing purposes. Instead, we should use the `:test` adapter. See: http://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/TestAdapter.html

This adapter makes running jobs configurable; it also makes it easy to check which jobs have been enqueued (and with what arguments) *without* running them.

See also: 
  - https://relishapp.com/rspec/rspec-rails/docs/matchers/have-enqueued-job-matcher
  -
 https://github.com/curationexperts/mahonia/blob/724e1f49691b1006e6f6cc583a44a44eb595cb0c/spec/rails_helper.rb#L194-L210

no-reply:
Fixes #2888

The test suite used :inline as the ActiveJob adapter, which performs jobs
inline. This is expensive and unnecessary in most cases.

The `TestAdapter` supports an in-memory queue which makes it easy to test which
jobs have been enqueued without processing them, or to process jobs selectively. See:
http://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/TestAdapter.html

We also introduce RSpec example metadata for `:perform_enqueued`, making it easy
to turn on job processing for particular example groups.

* Use `:test` as the queue adapter when running specs
* Add `before` and `after` behavior for `:perform_enqueued`
* Setup `:perform_enqueued` metadata for jobs that need to run enqueued jobs.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/actors/hyrax/actors/file_set_actor_spec.rb
+++ b/spec/actors/hyrax/actors/file_set_actor_spec.rb
@@ -223,7 +223,7 @@ RSpec.describe Hyrax::Actors::FileSetActor do
       expect(IngestJob).to receive(:perform_later).with(any_args).and_return(IngestJob.new)
       expect(actor.update_content(local_file)).to be_a(IngestJob)
     end
-    it 'runs callbacks' do
+    it 'runs callbacks', :perform_enqueued do
       # Do not bother ingesting the file -- test only the result of callback
       allow(file_actor).to receive(:ingest_file).with(any_args).and_return(double)
       expect(ContentNewVersionEventJob).to receive(:perform_later).with(file_set, user)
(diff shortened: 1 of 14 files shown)


no-reply:
This seems to significantly speed up the test suite. Anecdotally:

> Finished in 20 minutes 45 seconds (files took 11.93 seconds to load)

compared to a recent: 

> Finished in 32 minutes 3 seconds (files took 16.44 seconds to load)

It seems to be failing on Rails 5.0.6, though. I'm looking into that.

elrayle:
This is looking good.  It is an excellent candidate for best practices documentation, especially since it is a change from how background jobs have been handled previously.  The documentation could include a how to run a background job in a test, how to avoid running a background job in a test, and recommendations of when to and not to run a background job.

no-reply:
I like the idea of documenting in an appropriate place. Are there any suggestions about the best way to scope that?

> One question, could you allow the :perform_enqueud flag to go ahead and accept the list of jobs it will perform. It looks like you are always following it with a subsequent ActiveJob::Base.queue_adapter.filter = [...]. Seems like that list could be specified at the same time you are telling the test to :perform_enqueued, yeah?

Is the proposed syntax `describe '#my_method, perform_enqueued: [MyJob, OtherJob] do`?

I think that could work. I'll give it a go.

elrayle:
RE: Documentation - There is a stubbed out page for testing best practices.  This would be a good place to add best practices around testing and background jobs.

https://github.com/samvera/samvera.github.io/blob/master/pages/samvera/developer_resources/best_practices/testing.md

elrayle:
Add an issue to documentation project...
https://github.com/samvera/samvera.github.io/issues/232

PR #2913, opened 2018-04-11
remove extraneous field inputs

### LINKED ISSUE #2909: Extra box (that doesn't work) when adding a collection to a collection
### Descriptive summary

Moved from https://github.com/curationexperts/nurax/issues/259
When adding an existing collection to a collection, a Select Collection dialog box appears above Select drop down menu. This box seems like it's not supposed to be there or do anything, so we should remove it?

image

Reporting for
Browser: Firefox Quantum 59.0.1 (64 bit)
OS: Windows 7
Collections Nesting (CN_1.4)

elrayle:
Fixes #2909

<input /> was inadvertantly added after several real <input> fields.  The extra <input /> are being removed.

@samvera/hyrax-code-reviewers


jcoyne:
Sorry, I think this was due to the linter I ran.

PR #2924, opened 2018-04-13
limit the exposure of the permissions_attributes deprecation warning for collections

### LINKED ISSUE #2388: Deprecation warning on the primary code path.
There doesn't seem to be any way to avoid this deprecation.

```
DEPRECATION WARNING: passing in permissions with a new collection is deprecated and will be removed from Hyrax 3.0 (). (called from collection_params at /Users/jcoyne/workspace/samvera/hyrax-collections/app/controllers/hyrax/dashboard/collections_controller.rb:342)
```

elrayle:
Fixes #2388

Moved the deprecation down so users will only see it if they actually pass in read/edit permissions in the permissions_attributes.  This deprecation will never be triggered through the UI.

It is triggered through tests that confirm the older style permission parameters (e.g. read, edit) are correctly converted to the new style (e.g. manage, deposit, view).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -337,9 +337,10 @@ module Hyrax
 
         def extract_old_style_permission_attributes(attributes)
           # TODO: REMOVE in 3.0 - part of deprecation of permission attributes
-          Deprecation.warn(self, "passing in permissions with a new collection is deprecated and will be removed from Hyrax 3.0 ()") # TODO: elr - add alternative in ()
           permissions = attributes.delete("permissions_attributes")
           return [] unless permissions
+          Deprecation.warn(self, "Passing in permissions_attributes parameter with a new collection is deprecated and support will be removed from Hyrax 3.0. " \
+                                 "Use Hyrax::PermissionTemplate instead to grant Manage, Deposit, or View access.")
           participants = []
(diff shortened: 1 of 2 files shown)


no-reply:
Are there breadcrumbs for where the use of permission attributes was removed?

I'm not finding a relevant commit and #2388 seems to assert that this is coming up in normal use of the codebase. I see `permissions_attributes` passed explicitly (perhaps not in ways that reach these controller actions), and I want to make sure something gets into the commit message that makes it possible to understand this change when someone encounters this warning.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -337,9 +337,9 @@ def collection_params
 
         def extract_old_style_permission_attributes(attributes)
           # TODO: REMOVE in 3.0 - part of deprecation of permission attributes
-          Deprecation.warn(self, "passing in permissions with a new collection is deprecated and will be removed from Hyrax 3.0 ()") # TODO: elr - add alternative in ()
           permissions = attributes.delete("permissions_attributes")
           return [] unless permissions
+          Deprecation.warn(self, "passing in permissions_attributes parameter with a new collection is deprecated and support will be removed from Hyrax 3.0")
```
no-reply: What should I do instead?

PR #2926, opened 2018-04-13
Pin simple-form to fix build

bess:
Fixes 1-0-stable travis build

@samvera/hyrax-code-reviewers


no-reply:
Did you find we need to pin `parser`? I think the issue there might have been fixed in some 2.5.0.x release.

PR #2930, opened 2018-04-16
Add partial nested indexing option

laritakr:
Fixes https://github.com/samvera/hyrax/issues/2798

While this doesn't solve the issue of the unnecessary save of the parent collection when adding a subcollection or work to the collection, it does solve the problematic results of that save (the nested reindexing of the parent's entire graph).

Saving the parent collection loses its nested indexing fields as it goes through AF indexing, so the nested reindexing is necessary for the parent collection. But by adding a parameter to the nesting indexer, this parameter can be referenced by the adapter to allow the children of this parent to be reindexed only if necessary (only when they are missing THEIR nested indexing fields).

By removing the reindexing of all child collections and child works, this change will remove most of the issues with in-line indexing of nested collections. There is still a necessary delay for reindexing when adding or removing a large subcollection, but the implications when nesting works and subcollections based on the parent collection's size is removed.

Limited reindexing will be used for:
- indexing of collection when adding a work to a collection
- indexing of parent collection when adding a subcollection
- indexing a collection update where no nesting changes take place

depends on newly released samvera-nesting_indexer v2.0.0 gem

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/collections_membership_actor.rb
+++ b/app/actors/hyrax/actors/collections_membership_actor.rb
@@ -102,6 +102,8 @@ module Hyrax
         # along side the FileSets on the show page
         def add(env, id)
           collection = Collection.find(id)
+          collection.reindex_extent = Hyrax::Adapters::NestingIndexAdapter::LIMITED_REINDEX
+
           return unless env.current_ability.can?(:deposit, collection)
           env.curation_concern.member_of_collections << collection
         end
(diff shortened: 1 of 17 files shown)


laritakr:
Blocked pending release of samvera-nesting_indexer gem v 2.0 samvera-labs/samvera-nesting_indexer#14

laritakr:
samvera-nesting_indexer v2.0.0 is now released and this is unblocked and ready for review.

PR #2932, opened 2018-04-17
Convert ES6 syntax to ES5 syntax so Uglifier can deal with it.

lfarrell:
Convert ES6 syntax to ES5 syntax so that uglifier can minifiy it.

@elrayle

### DIFF
--- a/app/assets/javascripts/hyrax/collections.js
+++ b/app/assets/javascripts/hyrax/collections.js
@@ -245,9 +245,9 @@ Blacklight.onLoad(function () {
       // Get all inputs in the table
       .find('td:first input[type=checkbox]')
       // Filter to those that are checked
-      .filter((i, checkbox) => checkbox.checked)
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/hyrax/collections.js
```diff
@@ -245,9 +245,9 @@ Blacklight.onLoad(function () {
       // Get all inputs in the table
       .find('td:first input[type=checkbox]')
       // Filter to those that are checked
-      .filter((i, checkbox) => checkbox.checked)
+      .filter(function(i, checkbox) { return checkbox.checked });
```
adamjarling: To be safe, put an ending semicolon on the `return` statement?  `.filter(function(i, checkbox) { return checkbox.checked; });`

lfarrell: Ah, good catch. :)

PR #2938, opened 2018-04-18
improve single-membership collection error message

elrayle:
Includes some general cleanup of new/edit work errors.  When multi-membership feature was added, it tried to use flash messages to display errors.  Works uses curation_concern.errors to display messages in the new/edit forms.  Unused flash[:error] was removed.  Note that controller tests for create and update failure were reverted to what they were before multi-membeship feature was added.  The existence of the error messages is tested in the feature test instead.

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

*
*
*

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -56,7 +56,6 @@ module Hyrax
       else
         respond_to do |wants|
           wants.html do
-            flash[:error] = curation_concern.errors.messages[:collections].to_sentence
             build_form
             render 'new', status: :unprocessable_entity
           end
@@ -101,7 +100,6 @@ module Hyrax
       else
         respond_to do |wants|
           wants.html do
-            curation_concern.errors[:single_collection] << curation_concern.errors.messages[:collections].to_sentence
             build_form
             render 'edit', status: :unprocessable_entity
           end
--- a/app/views/hyrax/base/_form_visibility_error.html.erb
+++ b/app/views/hyrax/base/_form_visibility_error.html.erb
@@ -1,4 +1,3 @@
 <%= f.full_error(:visibility) %>
 <%= f.full_error(:embargo_release_date) %>
 <%= f.full_error(:visibility_after_embargo) %>
-<%= f.full_error(:single_collection) %>
--- a/spec/controllers/hyrax/generic_works_controller_spec.rb
+++ b/spec/controllers/hyrax/generic_works_controller_spec.rb
@@ -262,20 +262,9 @@ RSpec.describe Hyrax::GenericWorksController do
     context 'when create fails' do
       let(:work) { create(:work) }
       let(:create_status) { false }
-      let(:errors) { double }
-      let(:messages) { double }
-      let(:error_messages) { ['Error: foo bar'] }
-
-      before do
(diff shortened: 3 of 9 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -56,7 +56,6 @@ def create
       else
         respond_to do |wants|
           wants.html do
-            flash[:error] = curation_concern.errors.messages[:collections].to_sentence
             build_form
```
elrayle: Flash messages are not used for new work form.  Uses curation_concern.errors instead.  The addition of _form_collections_error.html.erb partial is what actually displays this message.

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -101,7 +100,6 @@ def update
       else
         respond_to do |wants|
           wants.html do
-            curation_concern.errors[:single_collection] << curation_concern.errors.messages[:collections].to_sentence
```
elrayle: This is only doing a rename of the error message.  It is not needed. _form_collections_error.html.erb partial is what actually displays this message.

app/views/hyrax/base/_form_visibility_error.html.erb
```diff
@@ -1,4 +1,3 @@
 <%= f.full_error(:visibility) %>
 <%= f.full_error(:embargo_release_date) %>
 <%= f.full_error(:visibility_after_embargo) %>
-<%= f.full_error(:single_collection) %>
```
elrayle: This error should not be mixed in with visibility related errors.  Moved to _form_collections_error.html.erb which will hold all collection related errors.

spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -262,20 +262,9 @@
...
-      before do
-        allow(controller).to receive(:curation_concern).and_return(work)
-        allow(work).to receive(:errors).and_return(errors)
-        allow(errors).to receive(:messages).and_return(messages)
-        allow(messages).to receive(:[]).with(:collections).and_return(error_messages)
-      end
 
       it 'draws the form again' do
         post :create, params: { generic_work: { title: ['a title'] } }
-        expect(flash[:error]).to eq error_messages.to_sentence
```
elrayle: Reverted to pre-multiple-membership feature since `curation_concern.errors` are used instead of `flash[:error]` which is tested in the features spec.

spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -495,22 +484,9 @@
...
-          allow(controller).to receive(:curation_concern).and_return(work)
-          allow(work).to receive(:errors).and_return(errors)
-          allow(errors).to receive(:messages).and_return(messages)
-          allow(messages).to receive(:[]).with(:collections).and_return(error_messages)
-          allow(errors).to receive(:[]).with(:single_collection).and_return(error_messages)
-        end
 
         it 'renders the form' do
           patch :update, params: { id: work, generic_work: {} }
-          expect(work.errors[:single_collection].to_sentence).to eq error_messages.to_sentence
```
elrayle: Errors are tested in the features spec.

PR #2947, opened 2018-04-19
Refactor UpdateCollectionTypeColumnOptions migration template to make…

### LINKED ISSUE #2946: Make UpdateCollectionTypeColumnOptions migration reversible
### Descriptive summary

Hyrax 2.1.0.beta2

The migration generated by `hyrax/lib/generators/hyrax/templates/db/migrate/20170810190549_update_collection_type_column_options.rb.erb` is not reversible as currently written.  Attempting to rollback the migration generated by that template results in the following error:
```
[ubuntu@samvera rdr (upgrade-to-hyrax-210)]$ rake db:rollback
DEPRECATION WARNING: human_readable_type= is deprecated and will be removed from a future release (human_readable_type is deprecated. Set the i18n key for activefedora.models.#{model_name.i18n_key} instead. This will be removed in Hyrax 3). (called from <class:Dataset> at /vagrant/rdr/app/models/dataset.rb:11)
== 20180418185222 UpdateCollectionTypeColumnOptions: reverting ================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:
...
Caused by:
ActiveRecord::IrreversibleMigration: 
This migration uses change_column, which is not automatically reversible.
To make the migration reversible you can either:
1. Define #up and #down methods in place of the #change method.
2. Use the #reversible method to define reversible behavior.
```

### Rationale

I discovered this when I tried to roll back all the database changes made by a trial upgrade of our application from Hyrax v2.0.0 to v2.1.0.beta2.  I think it would be useful if it were possible to roll back all the migrations added by this upgrade and the one cited is the only irreversible migration in that group.

### Expected behavior

The migration generated from the `20170810190549_update_collection_type_column_options.rb.erb` template is reversible.

### Actual behavior

The migration generated from the `20170810190549_update_collection_type_column_options.rb.erb` template is not reversible.

1. Perform the migration generated by the `20170810190549_update_collection_type_column_options.rb.erb` template.
2. Attempt to roll back the migration performed in step 1 above.

coblej:
… it reversible.

Fixes #2946 

Refactors `lib/generators/hyrax/templates/db/migrate/20170810190549_update_collection_type_column_options.rb.erb` to make the migration reversible.

As currently written, the migration generated by this template is not reversible.  The proposed change makes it reversible, which makes it possible to roll back all the database migrations introduced by the collection extensions work in v2.1.0.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/templates/db/migrate/20170810190549_update_collection_type_column_options.rb.erb
+++ b/lib/generators/hyrax/templates/db/migrate/20170810190549_update_collection_type_column_options.rb.erb
@@ -1,9 +1,17 @@
 class UpdateCollectionTypeColumnOptions < ActiveRecord::Migration<%= migration_version %>
-  def change
+  def up
     change_column :hyrax_collection_types, :title, :string, unique: true
     change_column :hyrax_collection_types, :machine_id, :string, unique: true
 
     remove_index :hyrax_collection_types, :machine_id
     add_index :hyrax_collection_types, :machine_id, unique: true
+  end
+
+  def down
+    change_column :hyrax_collection_types, :title, :string
+    change_column :hyrax_collection_types, :machine_id, :string
+
+    remove_index :hyrax_collection_types, :machine_id
+    add_index :hyrax_collection_types, :machine_id
   end
 end


coblej:
@no-reply @cjcolvar I don't know if this addresses the concern we discussed yesterday about insuring that revising the migration as proposed in the PR does not have bad effects for persons who have already run the migration as originally written but I did this test this morning:
1. Upgraded our current 2.0.0 Hyrax-based project to Hyrax 2.1.0.rc1.
2. Installed the migrations: `rails hyrax:install:migrations`.
3. Copied `db/schema.rb` to a safe location.  This is the "pre-210-migration" version of `schema.rb`.
4. Ran the migrations: `rails db:migrate`.  This performed 10 migrations.
5. Copied `db/schema.rb` to a safe location.  This is the "post-210-migration" version of `schema.rb`.
6. Replaced the `#change` method of the 2.1.0 `UpdateCollectionTypeColumnOptions` migration with the `#up`/`#down` methods of this PR.
7. Rolled back the 10 migrations that had been run in step 3 above: `rails db:rollback STEP=10`.  Note that, for the `UpdateCollectionTypeColumnOptions` migration, this rollback used the `#down` method from this PR.
8. Copied `db/schema.rb` to a safe location.  This is the "post-210-rollback" version of `schema.rb`.
9. Diff'ed the "pre-210-migration" version of `schema.rb` with the "post-210-rollback" version.  There were no differences.  So, rolling back the 2.1.0 migrations using the altered version of the `UpdateCollectionTypeColumnOptions` migration produced a `schema.rb` that was exactly the same as the one before the 2.1.0.rc1 migrations were run.  I take this to indicate that the revised migration in this PR was successfully able to roll back the original 2.1.0.rc1 migration.
10. Re-ran the migrations: `rake db:migrate:status`.  Noted that this run used the `#up` method from the `UpdateCollectionTypeColumnOptions` migration in this PR.
11. Copied `db/schema.rb` to a safe location.  This is the "post-210-re-migration" version of `schema.rb`.
12. Diff'ed the "post-210-migration" version of `schema.rb` (which used the "original" v2.1.0.rc1 migrations) with the "post-210-re-migration" version (which used the `UpdateCollectionTypeColumnOptions` migration in this PR).  There were no differences.

I take all this to mean ...
- that the migration change proposed in this PR can be used to roll back the migrations in the current v2.1.0.rc1 code without a problem
- that the migration change proposed in this PR produces the same result as the one in the current v2.1.0.rc1 code.

As I said, I'm not sure that this fully addresses the concerns expressed yesterday and I can understand that it would probably be best to have this confirmed by something other than the PR submitter ... but I thought I would document the testing that I did this morning.

cjcolvar:
I updated my Hyrax 2.1.0.beta2 app to 2.1.0.rc1 and ran `rake hyrax:install:migrations db:migrate` then switched to this branch and ran `rake hyrax:install:migrations`.  This found all of the migrations to be identical except the modified one which it skipped:
`skip  db/migrate/20180503193534_update_collection_type_column_options.hyrax.rb`
This looks like the default behavior for Hyrax: https://github.com/samvera/hyrax/blob/master/app/services/hyrax/database_migrator.rb#L35-L45.  So it looks like this PR will not affect Hyrax apps which have already run `hyrax:install:migrations` and will not have any affect on future runs of the rake task.  But when upgrading from a hyrax that didn't have the migration in question then the new version would get installed and everything should be happy.

I'm :+1: on merging this after this testing.

PR #2951, opened 2018-04-20
Add managed access to list of collections under Managed Collections tab

elrayle:
Adds a column to the collections list that shows the level of managed access a user has to each collection.  This applies to the Managed Collections tab only.  On Your Collections and All Collections the access level is always Manage and adds no additional information for the user.

I would also like to see a filter by managed access.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/admin_set_presenter.rb
+++ b/app/presenters/hyrax/admin_set_presenter.rb
@@ -31,6 +31,9 @@ module Hyrax
       []
     end
 
+    # For the Managed Collections tab, determine the label to use for the level of access the user has for this admin set.
+    # Checks from most permissive to most restrictive.
+    # @return String the access label (e.g. Manage, Deposit, View)
     def managed_access
       return I18n.t('hyrax.dashboard.my.collection_list.managed_access.manage') if current_ability.can?(:edit, solr_document)
       return I18n.t('hyrax.dashboard.my.collection_list.managed_access.deposit') if current_ability.can?(:deposit, solr_document)
@@ -38,6 +41,12 @@ module Hyrax
       ''
     end
 
+    # Determine if the user can perform batch operations on this admin set.  Currently, the only
+    # batch operation allowed is deleting, so this is equivalent to checking if the user can delete
+    # the admin set determined by criteria...
+    # * user must be able to edit the admin set to be able to delete it
+    # * the admin set itself must be able to be deleted (i.e., there cannot be any works in the admin set)
+    # @return Boolean true if the user can perform batch actions; otherwise, false
     def allow_batch?
       return false unless current_ability.can?(:edit, solr_document)
       !disable_delete?
--- a/app/presenters/hyrax/admin_set_presenter.rb
+++ b/app/presenters/hyrax/admin_set_presenter.rb
@@ -30,5 +30,17 @@ module Hyrax
     def available_parent_collections(*)
       []
     end
+
+    def managed_access
+      return I18n.t('hyrax.dashboard.my.collection_list.managed_access.manage') if current_ability.can?(:edit, solr_document)
+      return I18n.t('hyrax.dashboard.my.collection_list.managed_access.deposit') if current_ability.can?(:deposit, solr_document)
+      return I18n.t('hyrax.dashboard.my.collection_list.managed_access.view') if current_ability.can?(:read, solr_document)
+      ''
+    end
+
+    def allow_batch?
+      return false unless current_ability.can?(:edit, solr_document)
+      !disable_delete?
+    end
   end
 end
--- a/app/views/hyrax/dashboard/collections/_default_group.html.erb
+++ b/app/views/hyrax/dashboard/collections/_default_group.html.erb
@@ -1,0 +1,30 @@
+<table id="collections-list-table" class="table table-striped collections-list-table">
+  <caption class="sr-only"><%= t("hyrax.dashboard.my.sr.listing") %> <%= application_name %></caption>
+  <thead>
+  <tr>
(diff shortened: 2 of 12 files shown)


app/presenters/hyrax/admin_set_presenter.rb
```diff
@@ -30,5 +30,17 @@ def show_path
...
     end
+
+    def managed_access
+      return I18n.t('hyrax.dashboard.my.collection_list.managed_access.manage') if current_ability.can?(:edit, solr_document)
+      return I18n.t('hyrax.dashboard.my.collection_list.managed_access.deposit') if current_ability.can?(:deposit, solr_document)
+      return I18n.t('hyrax.dashboard.my.collection_list.managed_access.view') if current_ability.can?(:read, solr_document)
+      ''
+    end
+
+    def allow_batch?
```
cjcolvar: Could you please add a comment or yard doc here describing what the cases and return values are when one has edit access?  It wasn't obvious to me from the code here but made sense when I was reading the corresponding tests.

app/views/hyrax/dashboard/collections/_list_collections.html.erb
```diff
@@ -52,6 +54,9 @@
       </div>
     </div>
   </td>
+  <% if !current_ability.admin? %>
```
cjcolvar: Same question as above.

app/views/hyrax/dashboard/collections/_default_group.html.erb
```diff
@@ -0,0 +1,30 @@
+<table id="collections-list-table" class="table table-striped collections-list-table">
+  <caption class="sr-only"><%= t("hyrax.dashboard.my.sr.listing") %> <%= application_name %></caption>
+  <thead>
+  <tr>
+    <th class="check-all">
+      <label for="check_all" class="sr-only"><%= t("hyrax.dashboard.my.sr.check_all_label") %></label>
+      <input type="checkbox" name="check_all" id="check_all" value="yes" />
+    </th>
+    <th><%= t("hyrax.dashboard.my.heading.title") %></th>
+    <% if !current_ability.admin? %>
```
cjcolvar: Why not show it when admin?? Would it be better to be consistent and label as Manage even though it doesn't give any meaningful information to filter on?

app/views/hyrax/dashboard/collections/_default_group.html.erb
```diff
@@ -0,0 +1,30 @@
+<table id="collections-list-table" class="table table-striped collections-list-table">
```
cjcolvar: This file doesn't have any tests.  Could you please add some or is it effectively covered elsewhere?

spec/views/hyrax/dashboard/collections/_list_collections.html.erb_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    before do
+      allow(ability).to receive(:admin?).and_return(false)
+    end
+    context 'show access' do
+      context 'for manager' do
+        before do
+          allow(presenter).to receive(:managed_access).and_return('Manage')
+        end
+        it "shows Manage access" do
+          allow(view).to receive(:current_user).and_return(true)
```
cjcolvar: I don't understand why `current_user` is returning true and not an actual `User` or double.  Could this line also be moved up into the before block since it is done in all of the tests in this file?

spec/views/hyrax/dashboard/collections/_list_collections.html.erb_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+          render('list_collections.html.erb', collection_presenter: presenter)
```
cjcolvar: This line is repeated in every test and is duplicative of `subject`.  Maybe it could be removed from each test if `subject` is added to the end of the before block.

spec/presenters/hyrax/admin_set_presenter_spec.rb
```diff
@@ -70,4 +70,74 @@
...
+      before do
+        allow(ability).to receive(:can?).with(:edit, solr_document).and_return(false)
+        allow(ability).to receive(:can?).with(:deposit, solr_document).and_return(true)
+      end
+      it 'returns Deposit label' do
+        expect(presenter.managed_access).to eq 'Deposit'
+      end
+    end
+
+    context 'when manager' do
```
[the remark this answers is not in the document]
elrayle: good catch

app/presenters/hyrax/admin_set_presenter.rb
```diff
@@ -30,5 +30,26 @@ def show_path
...
+      return I18n.t('hyrax.dashboard.my.collection_list.managed_access.view') if current_ability.can?(:read, solr_document)
+      ''
+    end
+
+    # Determine if the user can perform batch operations on this admin set.  Currently, the only
+    # batch operation allowed is deleting, so this is equivalent to checking if the user can delete
+    # the admin set determined by criteria...
+    # * user must be able to edit the admin set to be able to delete it
+    # * the admin set itself must be able to be deleted (i.e., there cannot be any works in the admin set)
+    # @return Boolean true if the user can perform batch actions; otherwise, false
```
cjcolvar: 💯

app/views/hyrax/dashboard/collections/_default_group.html.erb
cjcolvar: @elrayle explained to me in slack the scenarios when the All Collections and Managed Collections appear for different users and how there isn't any inconsistencies.

PR #2952, opened 2018-04-20
Create a Collections JavaScript Utilities Class 

### LINKED ISSUE #2949: Collection and Collection Types JavaScript Utils module to DRY up shared functions
### Descriptive summary

In order to not duplicate JavaScript code between Collections and Collection Types, we should create a shared "Utils" module to handle shared, Collections-related JavaScript.

Doing this work will fix the conditional disabling behavior of the Add button on Collection Types > Edit > Add Participants.  It will use the same code which fixed: https://github.com/samvera/hyrax/pull/2942  by pulling the code into the Utils module so both Collections and Collection Types can use it.

### Rationale

DRY.  Create flexible, re-usable functions/methods instead of code duplication.  This is a shot at a pattern for scalable, organized JavaScript related to Collections.

### Related work

#2942 #2921 #2915

adamjarling:
Fixes #2949 

This PR creates a Collections javascript utilities class.  The initial motivation is to DRY up some code duplication between Collections and Collections Types' Edit views, specifically the "Add Sharing" or "Add Participants" tab for adding groups and users.

The code here does a few things, and might look bigger than it actually is.  There's a good amount of reorganizing along with new code.

1. Creates a `CollectionUtilities` javaScript class, which can be the source for shared modules/classes which encapsulate functionality which Collections and Collection Types need.  The `CollectionUtilities` class can be imported where it's needed, assuming the importer is an `.es6` file.

2. Moves the previous code for handling adding users and groups via AJAX when editing a Collection or Collection Type, Sharing or Participants tab, to the new utility class.

3. Moves the previous code for handling the enabled/disabled state of 2.'s "Add" buttons to the new utility class. 

4. Creates a new ES6 file `app/assets/javascripts/hyrax/collections_v2.es6`, which begins a work in progress of eventually replacing the ES5 file `app/assets/javascripts/hyrax/collections.js`, so we can take advantage of modularity, code-reuse, this utility class, etc.

5. Begins thinning out the `app/assets/javascripts/hyrax/collections.js` file by moving one handler function (for starters) to the new utility class.

6. The Collection Types js file, `app/assets/javascripts/hyrax/collection_types.es6` might look like a decent amount was modified outside of the additional code, but this is just running it through Prettier (https://prettier.io/), to get some consistency. 

@samvera/hyrax-code-reviewers


app/assets/javascripts/hyrax/collection_types.es6
```diff
@@ -1,94 +1,124 @@
...
+    let $collapseHeader = $('a.collapse-header');
+    let $collapseHeaderSpan = $('a.collapse-header').find('span');
 
     // Toggle show/hide of collapsible content on bootstrap toggle events
     $('#collapseAbout').on('show.bs.collapse', () => {
-      $collapseHeader.addClass('open')
-      $collapseHeaderSpan.html('Less')
-    })
+      $collapseHeader.addClass('open');
+      $collapseHeaderSpan.html('Less');
```
lfarrell: Should the text 'Less/More' be in the translation file?

app/assets/javascripts/hyrax/collections_v2.es6
```diff
@@ -0,0 +1,57 @@
+import CollectionUtilities from 'hyrax/collections_utils';
+
+export default class CollectionsV2 {
+  constructor() {
+    this.collectionUtilities = new CollectionUtilities();
+    this.sharingAddButtonDisabler();
```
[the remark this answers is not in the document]
adamjarling: If you don't mind, I'd prefer to leave new class instantiations at the top of the constructor, as a pattern.  Once you start bringing in multiples, it helps (from a visual perspective) to stack at the top, and kind of see quickly what you have to work with.

app/assets/javascripts/hyrax/collection_types.es6
adamjarling: Yea, good call.  Updating now.

adamjarling: Actually @lfarrell , I'm not sure the best way to access that rails method `t()` in a javascript file.  Any ideas?

adamjarling:
@lfarrell Requested changes are pushed.  For the translation replacement I commented about earlier, I just put the translation values as `data-attributes` on the element, then just reference those attributes from within `handleCollapseToggle()` context.

app/assets/javascripts/hyrax/collection_types.es6
lfarrell: I you'll have to have the file name like so `collection_types.es6.erb`. This should work for ES5 files e.g. `collection_types.js.erb`. Not sure what it'll do to es6 files.

PR #2954, opened 2018-04-20
[WIP] Change participant/sharing groups to a select list

### LINKED ISSUE #2937: Verification of groups not happening for adding participants/sharing
### Descriptive summary

A user can add any group name whether or not it exists for Collection Types, Collections, and Admin Sets.  This error existed for Hyrax 2.0 as well.  New/Edit works operates different using a selection list of all groups instead of having the user type in the group name.

### Rationale

Users should not be able to add non-existent groups.

### Expected behavior

Adding a non-existent group name should not be allowed and at least show a flash message stating the name doesn't exist.

### Actual behavior

Adding a non-existent group name is allowed and the non-existent group is added to the roles table.

1. Dashboard -> Collections -> edit a collection -> Sharing tab
1. Type in group name 'Junk', select a role, and click Add

Verify that you get a flash message stating that "Junk group does not exist."
Verify for Collections, Collection Types, and Admin Sets.

As an alternative, the change could be made to match that used for works sharing tab, which uses a selection list of existing groups.

### Related work

Issue #2921 - empty user field allowed when adding sharing

elrayle:
Fixes #2937

For collection types, collections, and admin sets, the process for granting a group access involved typing in the name.  This has been switched to use the approach used by works new/edit form.  Each controller now passes a list of groups for the current user.  If the user is an admin, it appends the registered group to allow admins to grant access to all registered users.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -38,6 +38,7 @@ module Hyrax
     end
 
     def edit
+      @groups = groups
       setup_form
     end
 
@@ -146,6 +147,12 @@ module Hyrax
         else
           redirect_to hyrax.my_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
         end
+      end
+
+      def groups
+        groups = current_user.groups || []
+        groups << ::Ability.registered_group_name if current_ability.admin?
+        groups.uniq
       end
   end
 end
--- a/app/controllers/hyrax/admin/collection_types_controller.rb
+++ b/app/controllers/hyrax/admin/collection_types_controller.rb
@@ -38,6 +38,7 @@ module Hyrax
     end
 
     def edit
+      @groups = groups
       setup_form
       setup_participants_form
       add_common_breadcrumbs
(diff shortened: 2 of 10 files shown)


elrayle:
Needs tests and merge with PR #2952 when it is complete.

elrayle:
@vantuyls @no-reply @julesies UX judgement call...

I am looking into what groups should be included in the group select list for collection types, admin sets, and collections participants/sharing.

### Current behavior

* select list includes only groups the user is in
* for admin sets, there is a button below the depositor table allowing the user to select All Registered Users ('registered' group) as a depositor

### Proposed behavior

#### regular users

* select list includes only groups the user is in
* no option for adding 'registered' group to any role

#### admin users

* select list includes all groups in the system
* the list also includes the 'registered' group
* no longer have a separate button for assigning the 'registered' group

vantuyls:
i'm okay with the proposed change.

julesies:
@elrayle does the drop-down mean only one group can be assigned? i'm think about the "registered" group issue the most and trying to understand if there is a need to have multiple groups....

elrayle:
@julesies You can assign as many groups as you want.  You just have to do it one at a time, as in, select a group, select a role, click add... repeat.

julesies:
ok, this sounds 👍

elrayle:
Per discussions in Slack, it was decided not to fix issue #2937.  Hyrax is designed to work with any of a number of group systems including but not limited to LDAP, Shibboleth, Groupify, RoleMapper, etc.  Some of these include predefined groups and can have large numbers of groups defined making getting all groups return an unwieldy number of groups.

There is an effort to rework groups and authentication.  It is anticipated this UI will be improved through better API access to groups when that work is complete.

@vantuyls If you concur, can you close this PR?

elrayle:
See comments in issue #2954

vantuyls:
Yep, concur. Thanks @elrayle

Issue #2151, opened 2017-11-10
Sipity::Workflow.activate! should be idempotent


jeremyf:
The current state of documentation says the following:

> Calling `#activate!` on a Sipity::Workflow a second time will
> *deactivate* that workflow. If you want your workflow setup script to
> be idempotent, you should add a check to see if it is already
> activated.
>
> https://github.com/samvera/samvera.github.io/commit/30a31995769199d6ce5b06107cbda7d09d68e447#diff-c813ad58bd3e03065da783330bfbe7c8R77

I believe the following is the expected behavior:

```gherkin
Given a `Sipity::Workflow` object that is active == true
When I call `Sipity::Workflow.activate!` with the given workflow
Then the given Sipity::Workflow will still be active == true
```

The reported behavior is:

```gherkin
Given a `Sipity::Workflow` object that is active == true
When I call `Sipity::Workflow.activate!` with the given workflow
Then the given Sipity::Workflow is active == false
```

bess:
👍 I wrote this documentation and can confirm this is the behavior I see when setting up a Hyrax project

mjgiarlo:
As part of this work, pls also remove this TODO: https://github.com/samvera/hyrax/pull/2150/files#diff-2fdd2e6a4cf83fbbca6fdd8ba6b17fafR43

laritakr:
@mjgiarlo Before my change, if you went to the tab to update the workflow and saved it with the same one marked, it would be deactivated. It now keeps whichever one active that you mark and save.

julesies:
I tested this on nurax and don't experience this bug.

PR #2956, opened 2018-04-23
Fix workflow.activate!

laritakr:
https://github.com/samvera/hyrax/issues/2151

Calling #activate! on an active workflow will keep it in active status.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/sipity/workflow.rb
+++ b/app/models/sipity/workflow.rb
@@ -39,20 +39,19 @@ module Sipity
     # Within the given permission_template scope:
     #   * Activate the specified workflow_id or workflow_name
     #   * Deactivate the other workflows
-    #
-    # TODO: Resolve https://github.com/samvera/hyrax/issues/2151 as the documentation is aspirational and not reality
-    #
     # @param permission_template [Hyrax::PermissionTemplate] The scope for activation of the workflow id
     # @param workflow_id [Integer] The workflow_id within the given permission_template that should be activated
     # @param workflow_name [String] The name of the workflow within the given permission template that should be activated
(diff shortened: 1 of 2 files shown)


app/models/sipity/workflow.rb
```diff
@@ -39,20 +39,19 @@ def self.find_active_workflow_for(admin_set_id:)
...
     def self.activate!(permission_template:, workflow_id: nil, workflow_name: nil)
       raise "You must specify a workflow_id or workflow_name to activate!" if workflow_id.blank? && workflow_name.blank?
-      finder_attributes = { permission_template: permission_template, id: workflow_id, name: workflow_name }.compact
-      Sipity::Workflow.find_by!(finder_attributes).tap do |workflow|
-        Sipity::Workflow.where(permission_template: permission_template, active: true).update(active: nil)
+      workflow_to_activate = Sipity::Workflow.find_by!({ permission_template: permission_template, id: workflow_id, name: workflow_name }.compact)
+      active_workflow = Sipity::Workflow.where(permission_template: permission_template, active: true)
+      return workflow_to_activate if workflow_to_activate == active_workflow.first
+      workflow_to_activate.tap do |workflow|
+        active_workflow.update(active: nil)
```
afred: Won't `active_workflow` be multiple `Sipity::Workflow` objects here?

laritakr: No, there is always only one active workflow, though I'm not positive right now where that gets enforced.

no-reply:
I'm working on the backport of this to the release branch.

PR #2958, opened 2018-04-25
Update broken GitHub documentation link

mark-dce:
The pull-request documentation link was broken.  Update it to point to a newer version of the documentation.

@no-reply for in-class demo + fixes broken link


cjcolvar:
Ugh, looks like we may need to bump up the wait time for the batch edit feature test similar to https://github.com/samvera/hyrax/pull/2948/commits/f7b484af68e6153ce5c26749d33f349194cddc6a .

PR #2961, opened 2018-04-25
Update translations for 2.1.0

no-reply:
Ran the google translate API on untranslated keys.

* Add `i18n-tasks` gem as development dependency
* Update translations for 2.1.0
* Remove unused translations

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/controllers/hyrax/admin/strategies_controller_spec.rb
+++ b/spec/controllers/hyrax/admin/strategies_controller_spec.rb
@@ -1,14 +1,20 @@
 RSpec.describe Hyrax::Admin::StrategiesController do
   describe "#update" do
     before do
+      # Added when Flipflop bumped to 2.3.2. See also https://github.com/voormedia/flipflop/issues/26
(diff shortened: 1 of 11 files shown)


spec/controllers/hyrax/admin/strategies_controller_spec.rb
```diff
@@ -1,14 +1,20 @@
 RSpec.describe Hyrax::Admin::StrategiesController do
   describe "#update" do
     before do
+      # Added when Flipflop bumped to 2.3.2. See also https://github.com/voormedia/flipflop/issues/26
+      features_hash = Flipflop::FeatureSet.current.instance_variable_get(:@features)
+      Flipflop::FeatureSet.current.instance_variable_set(:@features, features_hash.merge(feature_id => feature))
```
cjcolvar: Looking at the failures in the Push builds maybe the FeatureSet needs to be cleaned out in an after hook.

PR #2968, opened 2018-04-26
Fix the Admin::StrategiesController spec

no-reply:
* Backports parts of #2961 for 2.0-stable

@samvera/hyrax-code-reviewers


no-reply:
An additional backport is needed for 1.0

PR #2970, opened 2018-04-26
Improve on German language (native Speaker)

jmfrenzel:
* Improve Gtranslators guess on German locale (native speaker). Mainly fails on single words. 

@samvera/hyrax-code-reviewers


no-reply:
Excellent! Thanks so much!

PR #2981, opened 2018-04-30
display number of collections in repository in collection list pages

### LINKED ISSUE #2974: Collection count on /dashboard/collections page is wrong
### Descriptive summary

As a logged in user, on the collections page, there is text that says how many collections are "in the repository". This text actually only shows the up to the number of collections on the page and only the number of the collections the user has access to. 

### Rationale

It seems that the purpose of this text is to show users how many collections are in the repository. 

### Expected behavior

The number should reflect the number of collections in the repository.

### Actual behavior

(Admin) Displays number of collections on the page: 
image

(Admin) Changes when you change the number:
image

(User) Only the collection the user has edit access to:
image

1. Dashboard- > Collections
2. Try to determine how many collections are in the repository

### Related work

moved from https://github.com/curationexperts/nurax/issues/275

blancoj:
Fixes #2974

### DIFF
--- a/spec/features/dashboard/collection_spec.rb
+++ b/spec/features/dashboard/collection_spec.rb
@@ -36,7 +36,7 @@ RSpec.describe 'collection', type: :feature, clean_repo: true do
         visit '/dashboard/my/collections'
       end
 
-      it "has page title, does not have tabs, and lists only user's collections" do
+      it "has page title, does not have tabs, lists only user's collections, and displays number of collections in the respository" do
         expect(page).to have_content 'Collections'
         expect(page).not_to have_link 'All Collections'
         within('section.tabs-row') do
@@ -47,6 +47,7 @@ RSpec.describe 'collection', type: :feature, clean_repo: true do
         expect(page).to have_link(admin_set_b.title.first)
         expect(page).not_to have_link(collection3.title.first)
         expect(page).not_to have_link(admin_set_a.title.first)
+        expect(page).to have_content("3 collections in the repository")
       end
 
       it "has collection type and visibility filters" do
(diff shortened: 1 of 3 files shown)


spec/features/dashboard/collection_spec.rb
```diff
@@ -47,6 +47,7 @@
         expect(page).to have_link(admin_set_b.title.first)
         expect(page).not_to have_link(collection3.title.first)
         expect(page).not_to have_link(admin_set_a.title.first)
+        expect(page).to have_content("3 collections in the repository")
```
no-reply: @elrayle and I discussed this test and were brainstorming about ways to get a regression test in for this.

The current test checks that a count appears, but not that it's the correct "full" count rather than the paginated one. We have at least 5 options:

  - Create > 10 collections for the test;
    - This would probably be prohibitively slow.
  - Add a view spec for the index view with a mock response object and test the behavior there;
  - Stub the response object
    - Something like `allow_any_instance_of(@response.class).to receive(:total_count).and_return(1000)`;
  - Set `page_size` to 1 (how?) and leave this feature test otherwise the same;
  - Do nothing.

I'm open to any of these. @blancoj could you and @elrayle agree on a way forward?

spec/views/hyrax/my/collections/index.html.erb_spec.rb
```diff
@@ -0,0 +1,34 @@
+RSpec.describe "hyrax/my/collections/index.html.erb", type: :view do
+  before do
+    assign(:managed_collection_count, 0)
+    resp = []
```
elrayle: I believe the preferred way to do this kind of stub is...
```
let(:resp) { double(docs: "", total_count: 11) }
```

This would replace...
```
resp = []
allow(resp).to receive(:docs).and_return("")
allow(resp).to receive(:total_count).and_return(11)
```

Issue #2973, opened 2018-04-27
Modal to add works to a collection doesn't include all collections


laritakr:
### Descriptive summary

RC1 - Add work to a collection modal limits collection list to 100. If more than 100, the collection you came from is not pre-selected, and hitting enter gives a rails error.

### Expected behavior

List should include all eligible collections, with the selection you came from pre-selected.

### Actual behavior

List includes only 100 collections, and pre-selection only happens if the collection you came from is in that list of 100.

### Steps to reproduce the behavior

1. View collection
2. Select "add existing works to this collection"
3. Verify that nothing is pre-selected.
4. Hit enter to see rails error.

This will only happen when you have over 100 collections that you can add works to, and your chosen collection is not part of the query results that fills the modal list.

laritakr:
The task now is to modify the add to collection modal to work like the add to work option in the work's edit relationships tab. This involves:

- [ ] If params[:add_works_to_collection] exists, use it directly rather than loading the dropdown autocomplete option for selection
- [ ] Add a form to the collections-list section of `app/views/hyrax/dashboard/collections/_form_for_select_collection.html.erb`, grabbing the data from the same auto-complete link as used in `app/views/hyrax/base/_form_member_of_collections.html.erb`
- [ ] Create new javascript for new autocomplete functionality in `pp/assets/javascripts/hyrax/collections` based on only the relevant portions of the javascript in `app/assets/javascripts/hyrax/relationships`. This needs to handle the data-autocomplete url, use it to fill the simple2 dropdown selection list, and allow user to select from the list
- [ ] Modify the existing javascript functionality in `app/assets/javascripts/hyrax/collections.js` so it will still update `collection_replace_id` with the selected id for the add button in `hyrax/dashboard/collections/button_for_update_collection`

Add work to collection modal is called in three places, but one is obsolete:
1. `app/views/hyrax/base/_show_actions.html.erb`
2. `app/views/hyrax/dashboard/collections/_sort_and_per_page.html.erb`
 - [ ] verify this is unused and delete the edit code at the beginning of this partial. (This was previously used to add works to a collection from the edit view via `app/controllers/hyrax/dashboard/collections_controller.rb`. I believe it is stray code from removing that functionality. Relevant code has already been removed from controller)
3. `app/views/hyrax/my/works/_batch_actions.html.erb`

Test:
- [ ] Work Index view
- [ ] Collection Show view (select "add existing works to this collection") to get to work Index view with params set

[...]

PR #2984, opened 2018-04-30
Fix add works to a collection modal

laritakr:
Resolves https://github.com/samvera/hyrax/issues/2973

1. Added config variable which defaults to total number of collections plus 1000.
2. Use new variable as max number of collections to return in modal list when adding works to a collection.
3. If adding works to a collection from collection show view, include only the originating collection in modal list.
4. Other cleanup of unused or unnecessary code encountered during testing: 
* remove stray more/less parent collection javascript
* remove loading of works on collection edit view. The display of the works was previously removed, but the code loading the data for these views remained.

@samvera/hyrax-code-reviewers


app/assets/javascripts/hyrax/collections.js
```diff
@@ -266,18 +266,6 @@ Blacklight.onLoad(function () {
...
-    $("#show-less-parent-collections").show();
-  });
-
-  $('#show-less-parent-collections').on('click', function () {
-    $(this).hide();
-    $("#more-parent-collections").hide();
-    $("#show-more-parent-collections").show();
-  });
-
   // Add to collection modal form post
```
elrayle: Nice catch

lib/generators/hyrax/templates/config/initializers/hyrax.rb
```diff
@@ -219,6 +219,9 @@
   # How long to hold the lock in milliseconds
   # config.lock_time_to_live = 60_000
 
+  # Maximum number of collections possible in solr queries where we expect to return all collections
+  # config.collection_query_limit = Collection.count + 1_000
+
```
elrayle: I'm not sure I follow why Collection.count + 1_000.  Why not just 1_000?  Or the greater of Collection.count or 1_000?

lib/hyrax/configuration.rb
```diff
@@ -506,6 +506,13 @@ def default_nested_relationship_reindexer
       ->(id:, extent:) { Samvera::NestingIndexer.reindex_relationships(id: id, extent: extent) }
     end
 
+    # @return [integer] max # of collections to return in query
+    # Max number of collections, to be used when we need to return all in a solr query
+    attr_writer :collection_query_limit
+    def collection_query_limit
+      @collection_query_limit ||= Collection.count + 1_000
+    end
+
```
elrayle: I would have expected this to use the value from the initializer config/initializers/hyrax.rb `config.collection_query_limit`.

spec/services/hyrax/collections_service_spec.rb
```diff
@@ -38,9 +39,17 @@
     context "with read access" do
       let(:access) { :read }
 
-      it "returns three collections" do
+      it "returns four collections" do
         expect(subject.map(&:id)).to match_array [collection1.id, collection2.id, collection3.id, collection4.id]
```
elrayle: Good catch.

lib/generators/hyrax/templates/config/initializers/hyrax.rb
laritakr: @julesies and @cam156 said to use the count of collections plus 1000 so that's what i did.

elrayle: 👍

lib/hyrax/configuration.rb
elrayle: I confirmed that the configuration set in `config/initializers/hyrax.rb` is honored.  So this is fine.

elrayle:
@vantuyls @julesies @no-reply This PR does has a UI/UX change that I would like to run by you.  

### Behavior in this PR

* When getting a list of collections to select collections for nesting or to bulk add works to collection, the length of the list is limited based on a configuration.  **(good change)**  This is for all uses of the collection modal except the next scenario.
* When you start from a collection and select Add existing works to the collection, you are forwarded to Dashboard -> Works to select the works **(no change here)**.  When you select Add to Collection button, the select collection modal is displayed and there is only one collection listed, the one from which you started the process.  -- **This is the change in question.**  It used to list all collections with the collection pre-selected to the one from which the process was started.

### Options for handling long list of collections in collection modal

This means that the options for the collection modal are...
* OPTION 1:  **This is how it worked in the past**
  * when starting from anywhere else - Show all collections (up to limit) and allow user to select.
  * when starting from a collection - Show all collections (up to limit) and pre-select the collection where the process started.  
* OPTION 2: **This is how this PR works**
  * when starting from anywhere else - same as OPTION 1
  * when starting from a collection - Show just the collection where the process started.  (User cannot choose to put the works in a different collection.)
* OPTION 3:  **This would be new work with code potentially copied from work edit form**
  * when starting from anywhere else - Do not show a list of collections at all.  Use a search box similar to that used on Works edit form -> Relationships tab to search for a collection.  
  * when starting from a collection - Same as when coming from anywhere else, exepct that in the case where the user came from a collection, that collection is pre-selected.

NOTE: OPTION 1 is complicated by the limit if  `limit < count of all collections` because it is possible that the collection that should be pre-selected didn't come back in the set of collections returned under the limit.

This is all complicated by the fact that we are re-using the Dashboard -> Works batch select feature to select the works.  This could use UI/UX design.  The default limit that was put in place is `> 1000` which means the modal can list over 1000 collections, which is undesirable from a UX perspective.  

The question at hand is how we want to handle it for 2.1.0.

no-reply:
In slack, @elrayle, @laritakr, @julesies and I talked about exploring the search box option.

The status quo can be summarized as:

> when I start the app, i will have a working modal (i.e. hundreds of collections are listed) until 1,000 collections are added or the app restarts

Issue #2997, opened 2018-05-01
When deleting a collection display title instead of collection ID


elrayle:
https://github.com/curationexperts/nurax/issues/271

### Descriptive summary
Currently, when deleting a collection a user sees the following flash notice:

* _Collection 41687h57c was successfully deleted_

### Recommended Improvement

Show the collection title instead since a collection ID isn't very helpful

* _How the Cat Tamed the Dog was successfully deleted_

### Steps to reproduce the behavior
Delete via Collection's show page:
* Select Collection to delete
* Click on Delete collection
* Delete this collection?
* Click on OK
* Collection 41687h57c was successfully deleted
------------------
Delete via listing of all collections using Action option:
* Go to list of Collections
* Identify collection for deletion
* Select Delete Collection from Actions
* Are you sure you want to delete this collection? This action cannot be undone
* Click Delete
* Collection dz010q15f was successfully deleted

no-reply:
Closed by #3001; backported in #3021

PR #3001, opened 2018-05-01
Use collection title in delete confirmation

laritakr:
Replace collection id with collection title in confirmation and error flash notices after collection deletion.

Completes https://github.com/samvera/hyrax/issues/2997

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -149,11 +149,12 @@ module Hyrax
         end
       end
 
-      def after_destroy(id)
+      def after_destroy(_id)
+        # leaving id to avoid changing the method's parameters prior to release
         respond_to do |format|
           format.html do
             redirect_to my_collections_path,
-                        notice: t('hyrax.dashboard.my.action.collection_delete_success', id: id)
+                        notice: t('hyrax.dashboard.my.action.collection_delete_success')
           end
           format.json { head :no_content, location: my_collections_path }
         end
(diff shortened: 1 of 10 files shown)


elrayle:
@vantuyls @samvera/hyrax-repo-managers  Can I get confirmation that it is desired to show title instead of id?  The other option is to show both title and id.  The review group was leaning toward going with title for 2.1.0.

no-reply:
This PR changes the method signature on the callbacks. These were introduced in 18557e5ee4, so I think we are at liberty to change them prior to release.

However, it seems pretty late in the release process for method signature changes and I'd like to make sure we discuss this as @samvera/hyrax-code-reviewers.

Apart from the late change, is `#after_destroy(title:)` the interface we want to move toward? Do we envision use cases for `#after_destroy` that use data other than the title?

vantuyls:
im in favor of seeing title. collection IDs are, i think for most use cases, opaque enough to be useless in an alert message.

mjgiarlo:
If collection title is a multi-valued field, would the first title suffice here?

vantuyls:
ugh. with ordering not enforced, this can create confusion. I'd rather show all values of title to help with clarity.

laritakr:
You DO realize how long titles can be, and how large this flash message could become, right?

julesies:
@laritakr can we truncate the title? I think that would be more meaningful to a deleter than the id.

vantuyls:
100 characters?

laritakr:
screen shot 2018-05-02 at 2 51 44 pm

julesies:
i was testing this earlier, and have a few more thoughts. I think it already warns you before deleting like this which is very helpful. I think the flash message is also helpful because it lets the user know the operation was successfully completed. Even if you can't do anything about it, it's nice to know if it succeeds or fails. The collection ID is sort of confusing to the user, if title is problematic wouldn't it be ok to just say "the collection was successfully deleted"?

laritakr:
Removing the title is definitely the easiest thing if you're ok with that.  

Yes it does warn you before, and I could add the title there instead. But I don't know if there is a size limit, and it's obviously easier not to add variables into that message.

So the decision is just to remove the id AND title info and use a simple confirmation message that the collection was deleted?

PR #3007, opened 2018-05-02
Add popstate event handler to cover the case where the browser histor…

### LINKED ISSUE #2964: Back button does not function properly when navigating from work to file and Back to work
Moved from https://github.com/curationexperts/nurax/issues/272

**Descriptive summary**
There are problems with browser navigation related to navigating from a work showpage, to a file, then back to the work showpage. The Back button navigation doesn't properly navigate back.

**Rationale**
Users have expectations that the Back button will take you back to the previous webpage.

**Expected behavior**
Back button, when engaged from a file, takes one back the previous page

**Actual behavior**
Back button, when engaged from a file, takes one to the current page

go to: https://nurax.curationexperts.com/concern/images/r494vk17h
select any of the files in that work
once on the file showpage, use the Back button in the browser
Hyrax tries, but fails, to get to the previous page. In this case, failure looks like Hyrax staying on the file showpage.
###Additional notes
maybe related, but maybe not:
When loading this page:
https://nurax.curationexperts.com/concern/images/r494vk17h

As soon as Universal Viewer loads, the URL changes to this:
https://nurax.curationexperts.com/concern/images/r494vk17h#?c=0&m=0&s=0&cv=0&xywh=-1654%2C-169%2C7338%2C3360

This URL is the one the browser is trying to go Back to, not the work URL (above).

cjcolvar:
…y state is missing a turbolinks object due to Universal Viewer's manipulation of the browser history outside of turbolinks.  This fix is based on https://github.com/turbolinks/turbolinks/issues/219#issuecomment-275838923.  This handler gets registered after the turbolinks handler which does nothing if the turbolinks object isn't in the event state (https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/history.coffee#L26-L31).

This is ugly but it was the only working solution that I could find without modifying Universal Viewer or turbolinks.  I did some manual testing locally but more testing is needed to ensure that this fix works and is safe.

Fixes #2964.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/monkey_patch_turbolinks.js.coffee
+++ b/app/assets/javascripts/hyrax/monkey_patch_turbolinks.js.coffee
@@ -9,3 +9,15 @@ if Turbolinks?
       else
         @failed = true
         @delegate.requestFailedWithStatusCode(@xhr.status, @xhr.responseText)
+
+  # Fixes a back/forward navigation problem with UV and turbolinks
+  # See https://github.com/samvera/hyrax/issues/2964
+  # This is based on https://github.com/turbolinks/turbolinks/issues/219#issuecomment-275838923
+  $(window).on 'popstate', (event) =>
+    @turbolinks_location = Turbolinks.Location.wrap(window.location)
+    return if Turbolinks.controller.location.requestURL == @turbolinks_location.requestURL
+    return if event.state?.turbolinks
+    if @window_turbolinks = window.history.state?.turbolinks
+      Turbolinks.controller.historyPoppedToLocationWithRestorationIdentifier(@turbolinks_location, @window_turbolinks.restorationIdentifier)
+    else
+      Turbolinks.controller.historyPoppedToLocationWithRestorationIdentifier(@turbolinks_location, Turbolinks.uuid())


app/assets/javascripts/hyrax/turbolinks_events.js
```diff
@@ -6,3 +6,20 @@ $(document).on('turbolinks:load', function() {
   // See https://github.com/samvera/hyrax/issues/2906
   window.embedScriptIncluded = false;
 });
+
+// Fixes a back/forward navigation problem with UV and turbolinks
+// See https://github.com/samvera/hyrax/issues/2964
+// This is based on https://github.com/turbolinks/turbolinks/issues/219#issuecomment-275838923
+$(window).on('popstate', function(event) {
+  var turbolinks_location = Turbolinks.Location.wrap(window.location);
+  if (Turbolinks.controller.location.requestURL === turbolinks_location.requestURL) {
```
adamjarling: Does `turbolinks_location` always return a value?   If it's `undefined`, it'll throw an error.  If it might not always be an object, you could test that it is an object and/or has the `requestURL` property.  

Other than that, the JS looks clean to me.

cjcolvar: I think it will always be an object since `window.location` should always be an object and `wrap` creates a new object around it.

cjcolvar:
I wonder if it would be better to move this into https://github.com/samvera/hyrax/blob/master/app/assets/javascripts/hyrax/monkey_patch_turbolinks.js.coffee and handle the case when Turbolinks isn't enabled.  I'll do that and repush if it works out.

elrayle:
Confirmed with @cjcolvar that hound issues are not to be fixed because it is not desired to split the URLs.

PR #3010, opened 2018-05-04
Update the message about why polyfill is necessary

jcoyne:
It was added here
https://github.com/samvera/hyrax/commit/eb95157b52dc72d099cc336328439ad45e34432a
It's probably not used now that we're off of phantomjs


no-reply:
@jcoyne this looks like it's breaking Jasmine. Thoughts?

no-reply:
I'm merging this around the build, since it's only a comment change.

PR #3025, opened 2018-05-04
Add new automated translation in preparation for 2.4.0-rc2

no-reply:
Ran using the process documented at
https://github.com/samvera/hyrax/wiki/Translations-(internationalization).

@samvera/hyrax-code-reviewers


no-reply:
Closing this since more translations will be added before release.

PR #3027, opened 2018-05-04
Change radio buttons into autocomplete field for add works to collection modal.

### LINKED ISSUE #2973: Modal to add works to a collection doesn't include all collections

cjcolvar:
It was easiest to remove the ability for the user to select a different collection when adding existing works to a collection from the collection view page.  This makes the modal act like a confirmation dialog since the user has already selected a collection when they initiated the process.  The collection title is added as a query param instead of doing a lookup when rendering the page or via AJAX.

Before:
add-to-collection-modal-before

After:
From the Dashboard->Works page:
add-to-collection-modal
From a Collection view page:
add-to-collection-modal-2

Fixes #2973.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/collections.js
+++ b/app/assets/javascripts/hyrax/collections.js
@@ -1,3 +1,5 @@
+var autocompleteModule = require('hyrax/autocomplete');
+
 Blacklight.onLoad(function () {
 
   /**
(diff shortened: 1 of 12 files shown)


elrayle:
Thanks for the screen shots.  The changes you made look good.

elrayle:
@cjcolvar The tests that were selecting collections using radio buttons can be changed to use [Selectors::Dashboard#select_collection](https://github.com/samvera/hyrax/blob/master/spec/support/selectors.rb#L24-L38).  

See example usage in spec/features/collection_multi_membership_spec.rb...
* [include Selectors::Dashboard](https://github.com/samvera/hyrax/blob/master/spec/features/collection_multi_membership_spec.rb#L2)
* [select_collection(new_collection)](https://github.com/samvera/hyrax/blob/master/spec/features/collection_multi_membership_spec.rb#L120)

elrayle:
Randomly failing tests have been marked pending.  Issue #3038 was opened to address the need to fix the failing tests.

julesies:
@cjcolvar For: From collection page... are you supposed to be able to type into the box and essentially designate a different collection than you came from? currently you are not able to, just checking it's working as expected. (FWIW, i think this is fine, just checking it's right)

cjcolvar:
@julesies I thought you shouldn't be able to do that since it would be really weird so I made the input field be disabled and effectively turning it into a confirmation instead of a choice selection.   This is a change from prior behavior where I believe you could select a different collection from the radio buttons.

julesies:
cool then it's working!

PR #2256, opened 2017-11-22
[WIP] Update ActiveFedora to 12.0

jcoyne:
Waiting on releases of:
- [x] Hydra-head (https://github.com/samvera/hydra-head/pull/420)
- [x] hydra-derivatives
- [x] active_fedora-noid
- [x] hydra-pcdm.
- [x] ActiveFedora (https://github.com/samvera/active_fedora/pull/1298)

PR #3030, opened 2018-05-04
Add Rails 5.2 to the build matrix

no-reply:
Rails 5.0.x is no longer supported. Drop it from the build matrix and replace it with 5.2.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ env:
   # It should be sufficient to test only the latest of the patch versions for a minor version, they
   # should be compatible across patch versions (only bug fixes are released in patch versions).
   matrix:
-    - "RAILS_VERSION=5.1.4"
-    - "RAILS_VERSION=5.0.6"
+    - "RAILS_VERSION=5.2.0"
+    - "RAILS_VERSION=5.1.6"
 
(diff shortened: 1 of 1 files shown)


cjcolvar:
I'm pretty certain rails 5.2 isn't going to work until active-fedora gets a release that includes https://github.com/samvera/active_fedora/pull/1312.  Other dependencies may need patches as well. 
 I think `hydra-pcdm` is okay but I believe `hydra-works` has a couple small issues to resolve.  Even then Hyrax will need to upgrade to that version of ActiveFedora: https://github.com/samvera/hyrax/pull/2256 and the other dependencies.

I guess I'm wondering if we should wait on this until we know that we're closer to a PR adding support for rails 5.2.  That way we wouldn't be spending extra travis concurrent builds for something we know is going to fail.

no-reply:
@cjcolvar yes, it's definitely not my intent that we will change the build matrix until this is passing. This is here mainly to provide a target and a place for discussion. 

This PR represents the intent to drop support for Rails 5.0 in a future release. We need to talk about how to approach that drop, what it means for our versions, etc...

PR #3031, opened 2018-05-04
Change radio buttons into autocomplete field for add works to collect…

no-reply:
…ion modal.

It was easiest to remove the ability for the user to select a different collection when adding existing works to a collection from the collection view page.  This makes the modal act like a confirmation dialog since the user has already selected a collection when they initiated the process.  The collection title is added as a query param instead of doing a lookup when rendering the page or via AJAX.

Co-authored-by: Brian Keese <bkeese@indiana.edu>

Backports #3027 

@samvera/hyrax-code-reviewers


elrayle:
Too many changes.  Going to redo the backport.

Issue #2004, opened 2017-10-27
Viewer Access to Admin Set


vantuyls:
Descriptive summary

Viewers currently have no access to admin sets for which they have been designated a Viewer Role. Here's what seems to be missing:

Notification that they have been added as a Viewer to an Admin Set
Ability to see the Admin Set in the Administrative Sets (/admin/admin_sets) panel
Ability to see pending Works in the Admin Set from the Review Submissions (/admin/workflows) panel
From what I can tell, when a user has been given the Viewer role in an Admin Set, the only thing they are able to do is view a work that is in that Admin Set, pending approval, if they have been given a direct URL. There is no way to navigate to that work, or know that it's there.

Expected behavior

Viewers should be notified when they've been given View access to an Admin Set and should be able to navigate to that Admin Set and take actions on Works in that Admin Set.

Actual behavior

Here is an Admin Set with a Work that is pending approval:
image

Here are the participants of the Admin Set:
image

Here is what the Viewer can see from their Admin Sets panel (note: they can't see the Admin Set they're supposed to View):
image

Here is what they see when the try to view their submissions (note: these are works from admin sets for which the user is a Manager; none of these are from the admin set for which the user is a Viewer):
image

No notifications:
image

The purpose of the Viewer role is not clear to me.

Questions

What am I missing?
Should we add capabilities to the Viewer role?
Should we remove the Viewer role?

**More comments at:**
https://github.com/curationexperts/nurax/issues/121

elrayle:
I saw this as well in collections-extensions and was unclear if the inability to view the admin-set was intentional.  What does happen with the viewer role is that when a work is created, the admin-set viewers are given `read-access` to the new work.

@jcoyne @mjgiarlo What is the intended behavior for Viewers of admin-sets?  Can you check which of these you expect to apply.
- [ ] Viewer is assigned `read access` to the admin set
- [ ] Viewer is assigned `read access` to new works
- [ ] Viewer is assigned `reviewer role` to new works that have mediated deposit workflow

FYI... for collections in collection extensions
- [x] Viewer is assigned `read access` to the collection (meaning they see the collection in Dashboard -> Collections -> Managed Collections and can view the collection's admin show page)
- [x] Viewer is assigned `read access` to new works (when created directly in the collection)
- [ ] Viewer is assigned `reviewer role` to new works that have mediated deposit workflow (Collections do not impact workflows, so this is NOT assigned based on a collection)

elrayle:
[...]

PR #3056, opened 2018-05-09
allow viewers and depositors to view admin set’s admin show page

elrayle:
This excludes viewers and depositors who have access only by being in the public or registered groups.

Partial Fix of #2004 and https://github.com/curationexperts/nurax/issues/121

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/collections/permissions_service_spec.rb
+++ b/spec/services/hyrax/collections/permissions_service_spec.rb
@@ -26,10 +26,10 @@ RSpec.describe Hyrax::Collections::PermissionsService do
 
       subject { described_class }
 
-      it ".can_deposit_in_collection? returns true" do
+      it '.can_deposit_in_collection? returns true' do
         expect(subject.can_deposit_in_collection?(collection_id: collection.id, ability: ability)).to be true
       end
-      it ".can_view_admin_show_for_collection? returns true" do
+      it '.can_view_admin_show_for_collection? returns true' do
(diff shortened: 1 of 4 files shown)


spec/services/hyrax/collections/permissions_service_spec.rb
```diff
@@ -66,21 +66,105 @@
...
+          expect(subject.can_view_admin_show_for_collection?(collection_id: collection.id, ability: ability)).to be false
+        end
+      end
+    end
+
     context 'when user without access' do
       subject { described_class }
 
-      it ".can_deposit_in_collection? returns true" do
+      it '.can_deposit_in_collection? returns false' do
```
cjcolvar: Nice catch!

julesies:
to test viewer access to unapproved works in mediated deposit workflow: 
1. Create an admin set with 1 -step mediated deposit. 
2. using 3 different test accounts, add 1 account to viewer, 2nd to manager, add registered users as depositor. use the 3rd account as registered user. 
3. Log in as 3rd account, deposit a work to that admin set
4. Log in as 2nd account, verify the work is there, but do not approve. Grab the url to the work
5. Log in as 1st user, go to URL, verify if you can or can’t see the work.

Not sure if you then need to test viewer role after the work has been approved? if so
1. Log back in as manager, approve work, grab URL
2. Log in as viewer, go to URL

Issue #3063, opened 2018-05-09
Private child works show on work regardless of flipper status 


laritakr:
### Descriptive summary

Relates to https://github.com/curationexperts/nurax/issues/172

The decision was to use a flipper configuration to allow private files to be listed but hidden if the user doesn't have rights to view them. This is not working as expected. The following issues remain:
1. When flipper does not allow private objects to be shown, child works and nested works both continue to be shown in the list.
2. When flipper allows private objects to be shown, logged in user without access sees thumbnail for nested works.
3. Empty action dropdown shows for a private file-set on the work

Changes needed:
1. hide thumbnail when logged in for **all** private items without view authority
2. hide action dropdown for fileset if no valid options
3. hide **all** private items on a work if flipper is off (filesets, child works, and nested works)

### Rationale

Private work items should only be shown to authorized users, and only when flipper setting allows it.

### Steps to reproduce the behavior

Create public work
Change fileset to private
Add a private child work to it
Add a private nested work to it
Log off
View work as a different, unauthorized user
View work as public user
Change flipper setting
View work as a different, unauthorized user
View work as public user

laritakr:
![screen shot 2018-05-09 at 2 44 32 pm](https://user-images.githubusercontent.com/17851674/39840282-d70ee36a-53ac-11e8-8222-a51379b71d68.png)

laritakr:
View if "hide files" flipper is on, public user. Thumbnail shows for both child work and nested work. Neither object should be shown.
![screen shot 2018-05-09 at 5 24 22 pm](https://user-images.githubusercontent.com/17851674/39840605-f4781506-53ad-11e8-9b03-647e7068e967.png)

laritakr:
Note: thumbnails were displayed due to caching. No change needed for that item.

PR #3064, opened 2018-05-10
Use flipper status for all work's children

laritakr:
Fixes https://github.com/samvera/hyrax/issues/3063

Additionally:
- Adds internalization to work's actions.
- Checks for download ability before displaying download link on work's show view.
- Tightens up specs for `hide_private_files?` flipper to avoid building objects unnecessarily.

@samvera/hyrax-code-reviewers


app/views/hyrax/base/_items.html.erb
```diff
@@ -1,5 +1,5 @@
 <h2><%= t('.header') %></h2>
-<%  members = Flipflop.hide_private_files? ? presenter.exclude_unauthorized_file_sets : presenter.member_presenters %>
+<%  members = Flipflop.hide_private_files? ? presenter.exclude_unauthorized_members : presenter.member_presenters %>
```
elrayle: Should the flipflop be `hide_private_items?`

app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -1,42 +1,44 @@
-<div class="btn-group">
+<% if can?(:edit, file_set.id) || can?(:destroy, file_set.id)  || can?(:download, file_set.id) %>
```
elrayle: Is there a presenter available here that would allow this to be encapsulated in the presenter?  Something like...
```
presenter.can_perform_any_action?(file_set.id)
```

config/locales/hyrax.en.yml
```diff
@@ -830,6 +831,17 @@ en:
...
+        delete_confirm: Deleting %{file_set} from %{application_name} is permanent. Click OK to delete this from %{application_name}, or Cancel to cancel this operation
+        delete_title: Delete %{file_set}
+        download: Download
+        download_title: Download %{file_set}
+        edit: Edit
+        edit_title: Edit %{file_set}
+        header: Select an action
+        versions: Versions
+        versions_title: Display previous versions
       groups_description:
```
elrayle: Nice that you got these in!

PR #3066, opened 2018-05-11
authorize adminset resource before extra validation and provide better redirect

elrayle:
Related to Issue #2004, PR #3056, PR #3059

* moves `load_and_authorize_resource` before extra validation
* rescues access denied exceptions and forwards to home page if user is already logged in or to login page if they are not.  This is consistent with how other controllers handle this type of exception.
* tighten up ensure_manager! to check for manage access to the current admin set instead of any admin set

### DIFF
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -2,9 +2,13 @@ module Hyrax
   class Admin::AdminSetsController < ApplicationController
     include Hyrax::CollectionsControllerBehavior
 
+    before_action :authenticate_user!
     before_action :ensure_manager!, except: [:show]
-    before_action :ensure_viewer!, only: [:show]
     load_and_authorize_resource
+    before_action :ensure_viewer!, only: [:show]
+
+    # Catch permission errors
+    rescue_from Hydra::AccessDenied, CanCan::AccessDenied, with: :deny_adminset_access
 
     with_themed_layout 'dashboard'
     self.presenter_class = Hyrax::AdminSetPresenter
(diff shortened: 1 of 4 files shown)


app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -2,8 +2,6 @@ module Hyrax
   class Admin::AdminSetsController < ApplicationController
     include Hyrax::CollectionsControllerBehavior
 
-    before_action :ensure_manager!, except: [:show]
-    before_action :ensure_viewer!, only: [:show]
     load_and_authorize_resource
```
elrayle: See the explanation for why these are being removed in the main description of the PR.

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -2,9 +2,13 @@ module Hyrax
   class Admin::AdminSetsController < ApplicationController
     include Hyrax::CollectionsControllerBehavior
 
+    before_action :authenticate_user!
+    load_and_authorize_resource
```
elrayle: `load_and_authorize_resource` has to happen before additional validation or `@admin_set` is `nil`

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -2,9 +2,13 @@ module Hyrax
...
     before_action :ensure_manager!, except: [:show]
     before_action :ensure_viewer!, only: [:show]
-    load_and_authorize_resource
+
+    # Catch permission errors
+    rescue_from Hydra::AccessDenied, CanCan::AccessDenied, with: :deny_adminset_access
```
elrayle: This plus the `deny_adminset_access` method provide a forwarding behavior more consistent with other controllers.

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -95,7 +108,7 @@ def update_referer
       def ensure_manager!
         # Even though the user can view this admin set, they may not be able to view
         # it on the admin page.
-        authorize! :manage_any, AdminSet
+        authorize! :manage, @admin_set
```
elrayle: Checking for `:manage_any` is too permissive now that users other than admins can edit an admin set.

elrayle:
@cjcolvar I added the requested tests that validate a non-admin user CAN access an admin set they manage, and CANNOT access other admin sets.

PR #3077, opened 2018-05-16
Ensure locale is preserved when following links with urls generated outside of controller context.

### LINKED ISSUE #3075: Stats filter resets locale to default (QA)
### Descriptive summary
Locale falls back to default upon setting stats filter (QA)

### Expected behavior
Locale are kept unchanged   

### Actual behavior
Locale is set to the default one (en)

1.go to either 
  - https://nurax.curationexperts.com/dashboard/my/collections?locale=en
  - https://nurax.curationexperts.com/dashboard/collections?locale=de
1. set locale to German 
1. choose one collection an click on its title 
1. land on the the collections page with default locale, because no locale set 

### Related work
https://github.com/samvera/hyrax/pull/2864
https://github.com/samvera/hyrax/issues/2679

cjcolvar:
Fixes #3075.

This PR fixes all places that are the same scenario:
- Links to collections and admin sets on the collections dashboard
- Metadata values on the show page which link to searches
- ~Links to user profiles (like on featured works on the homepage)~
- ~Link to contact page on default terms of use page (might not actually work but shouldn't hurt)~

Not fixed in this PR are links embedded within notification messages.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/admin_set_presenter.rb
+++ b/app/presenters/hyrax/admin_set_presenter.rb
@@ -24,7 +24,7 @@ module Hyrax
     end
 
     def show_path
-      Hyrax::Engine.routes.url_helpers.admin_admin_set_path(id)
+      Hyrax::Engine.routes.url_helpers.admin_admin_set_path(id, locale: I18n.locale)
     end
 
     def available_parent_collections(*)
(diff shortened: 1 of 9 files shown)


cjcolvar:
I fixed the failing tests but noticed that the fix for the user profile link is also showing up in the notification messages meaning that messages created by one user may have a different locale specified then the user viewing the notification leading to the wrong locale being applied should that user follow the link.  This can be reproduced by changing the locale, adding a new version of a file, then examining the links in the User Activity for that file.

PR #3079, opened 2018-05-17
Ordered_members actor to add file_sets to work in one action

geekscruff:
See discussion here: https://groups.google.com/forum/#!topic/samvera-tech/vYpl7F2r5Dk

This PR adds a new actor (ordered_members_actor) to add file_sets to a work in a single action, rather than incrementally in the file_set_actor.

This is a proposed solution to the problem that works with many file_sets take a really long time to process (AttachFilesToWorkJob). The cause of this is that for each file_set, the work is re-saved to commit the new member to the ordered_members list for each file_set. As the number of members grows, the save takes longer each time because the ordered_members implementation in AF and Fedora is quite slow. So we see an exponential rise in the time taken. A work with a few hundred file_sets can be hours running AttachFilesToWorkJob.

UPDATE: the second commit refactors this to be a swap-able feature using two new Actors rather than a change to existing code.

### DIFF
--- a/app/actors/hyrax/actors/create_with_files_ordered_members_actor.rb
+++ b/app/actors/hyrax/actors/create_with_files_ordered_members_actor.rb
@@ -1,0 +1,13 @@
+module Hyrax
+  module Actors
+    # Creates a work and attaches files to the work
+    class CreateWithFilesOrderedMembersActor < CreateWithFilesActor
+      # @return [TrueClass]
+      def attach_files(files, env)
+        return true if files.blank?
+        AttachFilesToWorkWithOrderedMembersJob.perform_later(env.curation_concern, files, env.attributes.to_h.symbolize_keys)
+        true
+      end
+    end
+  end
+end
--- a/app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb
+++ b/app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb
@@ -1,0 +1,63 @@
+module Hyrax
+  module Actors
+    # If there is a key `:remote_files' in the attributes, it attaches the files at the specified URIs
+    # to the work. e.g.:
+    #     attributes[:remote_files] = filenames.map do |name|
+    #       { url: "https://example.com/file/#{name}", file_name: name }
+    #     end
+    #
+    # Browse everything may also return a local file. And although it's in the
+    # url property, it may have spaces, and not be a valid URI.
+    class CreateWithRemoteFilesOrderedMembersActor < CreateWithRemoteFilesActor
+      attr_reader :ordered_members
+
+      # @param [HashWithIndifferentAccess] remote_files
+      # @return [TrueClass]
+      def attach_files(env, remote_files)
+        return true unless remote_files
+        @ordered_members = env.curation_concern.ordered_members.to_a
+        remote_files.each do |file_info|
+          next if file_info.blank? || file_info[:url].blank?
+          # Escape any space characters, so that this is a legal URI
+          uri = URI.parse(Addressable::URI.escape(file_info[:url]))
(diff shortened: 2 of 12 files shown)


no-reply:
This looks like an amazing step forward!

There are a few places where the PR makes pretty dramatic changes to existing classes/methods; there I think I would be more comfortable if we explored solutions using polymorphism, leaving the current implementations in place. I'm going to think on the trade-offs on this a bit more.

@geekscruff if you'd like to pair on a extracting your new implementation into extending classes/modules we could setup a time for next week?

geekscruff:
@no-reply I'd really like to do that, I'll ping you on slack to setup a time.

geekscruff:
@no-reply I've refactored this to PR leave current implementation unchanged. I also reviewed other uses of FileSetActor's `attach_to_work` method (where the slowness happens).

There are three places where the attach_to_work method of FileSetActor is called:

- CreateWithRemoteFilesActor (calls FileSetActor.attach_to_work)
- AttachFilesToWorkJob (calls FileSetActor.attach_to_work, called by CreateWithFilesActor)
- Hyrax::Arkivo::Actor (calls FileSetActor.attach_to_work)

I've made the assumption that the Hyrax::Arkivo::Actor would be unlikely to be dealing with large numbers of filesets, so I've left this alone.

For the other two, I have two swap-able actors, that call different jobs/actors. Essentially all this is doing is  overriding the `FileSetActor.attach_to_work` method to *not* do the adding of ordered_members, instead building an array of ordered_member file_sets in the calling class and passing this to a new `OrderedMembersActo`.

There are the following new Actors/Jobs:

- `CreateWithFilesOrderedMembersActor` swaps out `CreateWithFilesActor` - this just calls a different Job (`AttachFilesToWorkWithOrderedMembersJob`)
- `AttachFilesToWorkWithOrderedMembersJob` - new Job called by `CreateWithFilesOrderedMembersActor` - this will now call the new `FileSetOrderedMembersActor` and `OrderedMembersActor`
- `CreateWithRemoteFilesOrderedMembersActor` swaps out `CreateWithRemoteFilesActor` - this will now call our new `FileSetOrderedMembersActor` and `OrderedMembersActor`
- `FileSetOrderedMembersActor` - extends `FileSetActor` to override the attach_to_work method
- `OrderedMembersActor` - handles adding the ordered_members and making the after_create_fileset callback

To switch to this ordered_members approach just needs the following configuration:

```
Hyrax::CurationConcern.actor_factory.swap(Hyrax::Actors::CreateWithFilesActor, Hyrax::Actors::CreateWithFilesOrderedMembersActor)
Hyrax::CurationConcern.actor_factory.swap(Hyrax::Actors::CreateWithRemoteFilesActor, Hyrax::Actors::CreateWithRemoteFilesOrderedMembersActor)
```

I haven't added tests for anything except OrderedMembersActor yet, as I wanted to get your thoughts on the approach first.

cjcolvar:
@no-reply Could you please take a look at this one again now that @geekscruff has refactored it?

geekscruff:
@no_reply could you have a look at this one? If it looks like the right direction I'll update to latest Hyrax and add tests. Thanks!

no-reply:
Thanks. Putting my eyes on this again now.

geekscruff:
@no-reply I merged with master and added the tests. I've written up this gist: https://gist.github.com/geekscruff/a9ee3cbddef3e38cf51f94582f6425c6 - not sure where to put that?

I have a bunch of failures relating to BlacklightAccessControls which must have been introduced today further down the stack:
```
You may not use Blacklight::AccessControls::SearchBuilder and include Blacklight::AccessControls::Enforcement on SearchBuilder at the same time
```

geekscruff:
The failure here is circleci, and it's a stale failure.

PR #3100, opened 2018-05-25
Make work item rows shown configurable

laritakr:
Add a config variable for show_work_item_rows. Default value of 10.

@samvera/hyrax-code-reviewers


no-reply:
Merged; @samvera/admins is there useful documentation we can add for this new configuration option?

PR #3114, opened 2018-06-08
Remove aria-hidden from Previous, Page Count and Next pagination

### LINKED ISSUE #3083: Remove aria-hidden from Previous, Page Count and Next pagination elements
### Descriptive summary

From the Accessibility360 Audit Report:

> Remove aria-hidden from these elements.

> Aria-hidden is being used to hide content that is useful to screen reader users.

pagination-hidden-attribute

#### Severity Level
Serious

### Rationale

Improved accessibility.  Aria-hidden is being used to hide content that is useful to screen reader users.

### Expected behavior

`aria-hidden="true"` attribute removed.

### Actual behavior
`aria-hidden="true"` attribute currently exists.

1. Go to: https://nurax.curationexperts.com
2. Search for a term with lots of results.  For example, "collection".
3. Inspect the html for referenced items above.

no-reply:
These `aria-hidden` flags were added explicitly from the generated paginator. Per our accessibility report, the elements should be re-added. Since this file was included in Hyrax only for this override, it can be simply removed, in line with the comment at the top.

Fixes #3083.

@samvera/hyrax-code-reviewers


no-reply:
cc: @samvera/hyrax-ui-ux-advisors

chrisdaaz:
tested on `nurax-dev` and confirmed.

PR #3118, opened 2018-06-11
namespace call to route helper path collection_path

### LINKED ISSUE #3113: collection_path route path not found in collections helper
### Descriptive summary

Access to the route path collection_path returns method not found when used to list a single type of work in a local app.  This can be fixed by prefixing the collection_path method call with the hyrax namespace.

### Rationale

Local apps should be able to use helper methods, so they should use proper namespacing to allow for this.

### Expected behavior

`render_collection_links(document)` defined in [app/helpers/hyrax/collections_helper.rb](https://github.com/samvera/hyrax/blob/master/app/helpers/hyrax/collections_helper.rb#L14) should include namespacing on its call to `collection_path`...

```
hyrax.collection_path(collection.id)
```

### Actual behavior

`render_collection_links(document)` defined in [app/helpers/hyrax/collections_helper.rb](https://github.com/samvera/hyrax/blob/master/app/helpers/hyrax/collections_helper.rb#L14) does not include namespacing on its call to `collection_path`...

```
collection_path(collection.id)
```

1. Create a work and put it in a collection
1. From a view in the test app, call `<%= render_collection_links(document) %>` which calls `collection_path(collection_id)` where document is the solr doc for the work
1. Should render a link with a path to the collection for the work

elrayle:
Fixes #3113

Add `hyrax` namespace to call to `collection_path`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/collections_helper.rb
+++ b/app/helpers/hyrax/collections_helper.rb
@@ -11,7 +11,7 @@ module Hyrax
     def render_collection_links(solr_doc)
       collection_list = Hyrax::CollectionMemberService.run(solr_doc, controller.current_ability)
       return if collection_list.empty?
-      links = collection_list.map { |collection| link_to collection.title_or_label, collection_path(collection.id) }
+      links = collection_list.map { |collection| link_to collection.title_or_label, hyrax.collection_path(collection.id) }
       collection_links = []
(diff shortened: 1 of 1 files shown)


no-reply:
:+1:

PR #3121, opened 2018-06-12
Change SingleUseLink itemId and downloadKey from camel to snake case

mbklein:
PostgreSQL, by default, doesn't handle camelCase column names very well, so let's not use them.

This PR renames two database columns, both in `single_use_links`: `itemId` -> `item_id` and `downloadKey` -> `download_key`. This is because PostgreSQL is case insensitive by default, but not _totally_ case insensitive. Creating `itemId` turns it into `itemid` but in such a way that it does not recognize `itemId` in queries.

Ideally, this would be fixed in the `pg` gem, but they have enough compatibility and versioning issues going on right now that I think avoiding capital letters in column names is the safest bet.

* Rename `itemId` to `item_id`
* Rename `downloadKey` to `download_key`
* Change all references to them throughout the codebase
* Add `alias_attribute` statements to the `SingleUseLink` model to allow the camelCase attribute names to continue to work

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/single_use_link.rb
+++ b/app/models/single_use_link.rb
@@ -2,10 +2,13 @@ class SingleUseLink < ActiveRecord::Base
   validate :expiration_date_cannot_be_in_the_past
   validate :cannot_be_destroyed
 
+  alias_attribute :downloadKey, :download_key
+  alias_attribute :itemId, :item_id
+
   after_initialize :set_defaults
 
   def create_for_path(path)
-    self.class.create(itemId: itemId, path: path)
+    self.class.create(item_id: item_id, path: path)
   end
 
   def expired?
@@ -13,7 +16,7 @@ class SingleUseLink < ActiveRecord::Base
   end
 
   def to_param
-    downloadKey
+    download_key
   end
 
   private
@@ -29,6 +32,6 @@ class SingleUseLink < ActiveRecord::Base
     def set_defaults
       return unless new_record?
       self.expires ||= DateTime.current.advance(hours: 24)
-      self.downloadKey ||= (Digest::SHA2.new << rand(1_000_000_000).to_s).to_s
+      self.download_key ||= (Digest::SHA2.new << rand(1_000_000_000).to_s).to_s
     end
 end
--- a/app/models/single_use_link.rb
+++ b/app/models/single_use_link.rb
@@ -7,6 +7,28 @@ class SingleUseLink < ActiveRecord::Base
 
   after_initialize :set_defaults
 
+  # rubocop:disable Naming/MethodName
+  def downloadKey
+    Deprecation.warn(self, 'The #downloadKey attribute is deprecated. Use #download_key instead.')
+    download_key
+  end
+
+  def downloadKey=(val)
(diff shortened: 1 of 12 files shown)


mbklein:
I would love to add deprecations, but I can only figure out how to do them for traditional getters and setters, not where they might (e.g.) appear as keys in a `find_by` or a `create`, and I feel that might complicate the issue more than it helps.

no-reply:
Is there a ticket for this in the `pg` gem?

I'm thinking about how best to make this backwards compatible. Does `.alias_attribute` present a way forward, perhaps? http://api.rubyonrails.org/classes/Module.html#method-i-alias_attribute

mbklein:
@no-reply Yes, `.alias_attribute` provides 100% backward compatibility for the old attribute names. It's included in this PR.

no-reply:
Ah, sorry, I misread your comment about the deprecations. :rocket:

app/models/single_use_link.rb
```diff
@@ -2,18 +2,21 @@ class SingleUseLink < ActiveRecord::Base
   validate :expiration_date_cannot_be_in_the_past
   validate :cannot_be_destroyed
 
+  alias_attribute :downloadKey, :download_key
```
jcoyne: Should these alias methods be deprecated so as to discourage their use?

mbklein:
FYI: This PR is currently blocked by the versioning discussion in the [branching model proposal](https://docs.google.com/document/d/18WRJNxLaOnBU0li4e2nr7zgMC5ppdSeEJ9Ql4oZJUsg/edit?disco=AAAAB_cxzT4).

app/models/single_use_link.rb
mbklein: I wanted to deprecate them, but I couldn't figure out how to deprecate them comprehensively. They're not just methods; they're attributes that can show up anywhere. I don't know how to warn that, for example, `SingleUseLink.find_by(downloadKey: 'foo')` is deprecated in favor of `SingleUseLink.find_by(download_key: 'foo')`.

jcoyne: I think at least deprecating the attributes would be better than not deprecating any usages at all.  I think doing a `SingleUseLink.find_by(downloadKey: 'foo')` is broken after this PR.  But I don't know how you would avoid that.

mbklein: The `alias_attribute` you highlighted specifically supports the continued use of the old attributes everywhere, even as hash keys in `find_by` and `create`. I can add deprecations to the straight-up attribute getter/setter methods, but not the other usages.

mbklein:
@jcoyne I added deprecations where I could

no-reply:
@mbklein I'm going to spare a cycle this week to see if I can make the migration optional--in part an an experiment informing the discussion linked in https://github.com/samvera/hyrax/pull/3121#issuecomment-397671851. I'll report out and we can move this forward from there.

no-reply:
I'm thinking about rebasing and merging this and just assuming that this implies Hyrax 3.0. 

If that happened in the coming weeks, would there be objection? @samvera/hyrax-code-reviewers @vantuyls @samvera/hyrax-repo-managers.

PR #3130, opened 2018-06-25
Use a one-byte GET request instead of a HEAD request to check the validity of a remote URL

mbklein:
The current strategy of using a `HEAD` request to check remote URLs fails for Amazon S3 pre-signed URLs. S3 URLs are generated for a single verb only (in this case, `GET`), and will return a `403 Forbidden` response if any other verb is used.

The lack of `HEAD` support for pre-signed `GET` URLs is an old and unresolved topic on the AWS support forums. I don't see them fixing this any time soon, if at all.

This is not an optimal solution, as it depends on the remote server being configured to satisfy byte range requests. I think it's safe to assume that the cloud providers we support all do so.

* Change `ImportUrlJob` to issue a `GET` request for bytes 0-0 instead of a `HEAD`.
* Alter the `ImportUrlJob` specs to stub that range request instead of the head request.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -24,7 +24,7 @@ class ImportUrlJob < Hyrax::ApplicationJob
     @file_set = file_set
     @operation = operation
 
-    unless HTTParty.get(file_set.import_url, headers: { Range: 'bytes=0-0' }).success?
+    unless can_retrieve?(uri)
       send_error('Expired URL')
       return false
     end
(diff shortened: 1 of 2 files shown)


app/jobs/import_url_job.rb
```diff
@@ -24,7 +24,7 @@ def perform(file_set, operation, headers = {})
     @file_set = file_set
     @operation = operation
 
-    unless HTTParty.head(file_set.import_url).success?
+    unless HTTParty.get(file_set.import_url, headers: { Range: 'bytes=0-0' }).success?
```
jcoyne: I think you should add documentation here. Your commit message would do nicely.

mbklein:
@jcoyne Done

mbklein:
I can change it to `HEAD || GET`, which is how I almost implemented it in the first place. Or special-case it only to do the byte range if it's S3. Or maybe an `OPTIONS` request to see if byte ranges are supported before deciding how to proceed.

mbklein:
I'll experiment and update.

PR #3134, opened 2018-06-28
[refactor] Invert dependencies for `AdminStatsPresenter`

no-reply:
This is a straight refactor to invert the previously hard coded dependencies in
`AdminStatsPresenter`. This allows callers to inject stats services with custom
behavior.

This is particularly motivated by the eager failure behavior of
`Hyrax::Statistics::Depositors::Summary`, which raises an error in the case
that a work has been saved with a `depositor` that can't be found (e.g. the
depositor's user_key has changed, or a depositor string has been saved that does
not have a matching record in the application). Since we don't enforce integrity
of reference for depostor strings, an option for a more fault tolerant stats
service seems desirable. This opens the presenter to that change.

We disable the method parameter size rubocop rule for the `#initialize` method,
since having many optional parameters in an intializer seems like a step forward
from having many hard coded dependencies. Further changes to reduce the
parameter list may be worth exploring.

Lastly, we add class variables to the stats behavior mixin allowing custom presenters
and service options in client code.

* Invert dependencies in `AdminStatsPresenter`
* Add `admin_stats_presenter` and `admin_stats_services` class variables to the `Admin::StatsBehavior` controller mixin.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/admin/stats_behavior.rb
+++ b/app/controllers/concerns/hyrax/admin/stats_behavior.rb
@@ -1,20 +1,44 @@
 module Hyrax
   module Admin
+    ##
+    # @example using a custom presenter and stats services
+    #   class MyStatsController < ApplicationController
+    #     include Hyrax::Admin::StatsBehavior
+    #
+    #     self.admin_stats_presenter = MyCustomStatsPresenter
+    #     self.admin_stats_services = { by_depositor:      CustomDepositorService,
+    #                                   depositor_summary: CustomSummaryService }
+    #     # see AdminStatsPresenter#initialize for supported services
+    #   end
+    #
     module StatsBehavior
       extend ActiveSupport::Concern
       included do
         with_themed_layout 'dashboard'
+
+        class_attribute :admin_stats_presenter, :admin_stats_services
+        self.admin_stats_presenter = AdminStatsPresenter
+        self.admin_stats_services  = {}
       end
 
       def show
         authorize! :read, Hyrax::Statistics
         stats_filters = params.fetch(:stats_filters, {})
         limit = params.fetch(:limit, "5").to_i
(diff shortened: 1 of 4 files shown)


no-reply:
@jcoyne this is rebased on the fixed build with an added commit and updated PR text. If you have time to finish your previous review, that would be :rocket:. Otherwise, thanks for the initial review anyway. :)

app/controllers/concerns/hyrax/admin/stats_behavior.rb
```diff
@@ -1,20 +1,44 @@
 module Hyrax
   module Admin
+    ##
+    # @example using a custom presenter and stats services
+    #   class MyStatsController < ApplicationController
+    #     include Hyrax::Admin::StatsBehavior
+    #
+    #     self.admin_stats_presenter = MyCustomStatsPresenter
+    #     self.admin_stats_services = { by_depositor:      CustomDepositorService,
+    #                                   depositor_summary  CustomSummaryService }
```
jcoyne: missing a colon here?

no-reply: thanks!

no-reply: Updated!

jcoyne:
This seems to have introduced the following deprecation warning:
```
DEPRECATION WARNING: Comparing equality between `ActionController::Parameters` and a `Hash` is deprecated and will be removed in Rails 5.1. Please only do comparisons between instances of `ActionController::Parameters`. If you need to compare to a hash, first convert it using `ActionController::Parameters#new`. (called from block (5 levels) in <top (required)> at /home/travis/build/samvera/hyrax/spec/controllers/hyrax/admin/stats_controller_spec.rb:41)
```

PR #3140, opened 2018-07-05
Add Display Name to Workflow Roles management page

bess:
Adding the display name to the workflow roles page makes it
more user friendly for repository admins.

Also, add the ability to sort by Display Name or User Key. 

Fixes https://github.com/curationexperts/laevigata/issues/1173 (we could just fix this locally, but we thought others might want this improvement too).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/admin/workflow_roles/index.html.erb
+++ b/app/views/hyrax/admin/workflow_roles/index.html.erb
@@ -26,15 +26,17 @@
         <h2 class="panel-title h2"><%= t('.current_roles') %></h3>
       </div>
       <div class="panel-body">
-        <table class="table table-striped">
+        <table class="table table-striped datatable">
           <thead>
+            <th><%= t('.header.name') %></th>
             <th><%= t('.header.user') %></th>
             <th><%= t('.header.roles') %></th>
           </thead>
           <tbody>
           <% @presenter.users.each do |user| %>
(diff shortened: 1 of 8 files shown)


no-reply:
@bess this looks like it's failing with missing i18n keys.

no-reply:
@samvera/hyrax-repo-managers any thoughts about this change? It looks like a clear improvement to me.

app/views/hyrax/admin/workflow_roles/index.html.erb
```diff
@@ -26,15 +26,17 @@
...
           <thead>
+            <th><%= t('.header.display_name') %></th>
             <th><%= t('.header.user') %></th>
             <th><%= t('.header.roles') %></th>
           </thead>
           <tbody>
           <% @presenter.users.each do |user| %>
             <tr>
-              <td><%= user.user_key %></td>
+              <td data-sort="<%= user.display_name %>"><%= user.display_name %></td>
```
no-reply: Should we use `User#name` instead? https://github.com/samvera/hyrax/blob/b498d8ba181f4e76da89cae60e92f9f54675fad5/app/models/concerns/hyrax/user.rb#L115-L119

chrisdaaz:
i agree that this will make the page more user friendly

PR #3142, opened 2018-07-10
Consolidate dependency on ActiveFedora::Base.logger

no-reply:
This dependency was hard coded in a number of places. This moves it to a single place where it can more easily be overridden. 

This is a (very small) part of a larger program to clarify and isolate the extent of our `ActiveFedora` dependencies.

* Add a `Hyrax.logger`, setting it to `ActiveFedora::Base.logger`.
* Depend on `Hyrax.logger` instead of `ActiveFedora::Base.logger` throughout.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/fixity_check_job.rb
+++ b/app/jobs/fixity_check_job.rb
@@ -58,6 +58,6 @@ class FixityCheckJob < Hyrax::ApplicationJob
     end
 
     def logger
-      ActiveFedora::Base.logger
+      Hyrax.logger
     end
 end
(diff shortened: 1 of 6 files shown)


no-reply:
@cjcolvar good call. I added a test checking for [`#log`](https://ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html#method-i-log) which seemed like the easiest "are you a duck/logger" check.

PR #3143, opened 2018-07-10
[refactor] Don't depend on `Hyrax::WorkRelation` from the top level

no-reply:
The `Hyrax` module had a hard dependency on `Hyrax::WorkRelation::DummyModel`, which relied in turn on `Hyrax.config`. This reaching back and forth across layers is confusing and unnecessary. Instead, `Hyrax` can get its `.primary_work_type` from closer to home.

* Change the `Hyrax.primary_work_type` implementation to retrieve directly from `.config` instead of an obscure downstream module.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -47,6 +47,6 @@ module Hyrax
   end
 
   def self.primary_work_type
-    Hyrax::WorkRelation::DummyModel.primary_concern
+    config.curation_concerns.first
   end
 end


jcoyne:
I support this. But I wonder if we may want to move the `primary_concern` method to the config class itself.

no-reply:
> But I wonder if we may want to move the primary_work_type method to the config class itself.

Seems reasonable, though not all the details are immediately clear to me.

We do need to keep this method, at least for the time being, in either case. It's clearly public API.

Issue #1199, opened 2017-06-12
Change button to "Save" in file set edit form


awead:
### Descriptive summary

The language of the file set edit screen is misleading. "Update Attached File" doesn't seem to apply. Use "Save" instead.

### Rationale

Consistency, really. Also the term "attached file" doesn't seem correct.

![fileset](https://user-images.githubusercontent.com/312085/27044878-1b8ba2ac-4f6d-11e7-9a5e-b406a4aec2ee.png)

mjgiarlo:
May I ask the @samvera/hyrax-ui-ux-advisors to chime in on this?

That is super old code (and needs i18n to :boot:): https://github.com/samvera/hyrax/blame/master/app/views/hyrax/file_sets/_form.html.erb

newmanld:
+1 to change button to 'Save'

PR #3145, opened 2018-07-10
Change "Update Attached File" button to "Save" on file_set edit view (#1199)

laritakr:
fixes https://github.com/samvera/hyrax/issues/1199

Add internalization to file set edit screen.
Change "Update Attached File" button to "Save".

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_share_with.html.erb
+++ b/app/views/hyrax/base/_share_with.html.erb
@@ -1,14 +1,6 @@
-<p>You may grant &quot;View/Download&quot; or &quot;Edit&quot; access for specific users and/or groups to files.
-  Enter a valid <%= t('hyrax.account_name') %>, one at a time, select the access level for that user and click
-  <button class='btn btn-xs btn-inverse' onclick='return false;'><i class='glyphicon glyphicon-plus'></i>Add</button>.
-</p>
-
+<p><%= t('.share_with_html', account_name: t('hyrax.account_name')) %></p>
 <%= render 'hyrax/file_sets/groups_description' %>
-
-<h5><b>Permission Definitions</b></h5>
+<h5><b><%= t('.definition_heading') %></b></h5>
 <p>
-  <strong>View/Download:</strong> this file (both contents and metadata) is accessible from within
-  <%= application_name %>.<br />
-  <strong>Edit:</strong> this file (both contents and metadata) can be edited. You may only grant
-  this permission to <%= institution_name %> users and/or groups.
+  <%= t('.definitions_html', application_name: application_name, institution_name: institution_name) %>
 </p>
--- a/app/views/hyrax/file_sets/_permission_form.html.erb
+++ b/app/views/hyrax/file_sets/_permission_form.html.erb
@@ -3,11 +3,11 @@
 <% public_perm = true if params[:controller] == 'batch' %>
 <% registered_perm = f.object.permissions.map { |perm| perm.access if perm.agent_name == "registered"}.compact.first %>
 
-<h2 id="permissions_display"><% if params[:controller] == 'batch' %>Bulk <% end %>Permissions <% if params[:controller] == 'batch' %>
-      <small>(applied to all files just uploaded)</small><% end %>
+<h2 id="permissions_display"><% if params[:controller] == 'batch' %>Bulk <% end %><%= t('.header') %> <% if params[:controller] == 'batch' %>
+      <small><%= t('.applied_to') %></small><% end %>
 </h2>
 
-<div class="alert alert-info hidden" id="save_perm_note">Permissions are <strong>not</strong> saved until the &quot;Save&quot; button is pressed at the bottom of the page.</div>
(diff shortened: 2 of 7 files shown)


app/views/hyrax/file_sets/_permission_form.html.erb
```diff
@@ -26,11 +26,11 @@
 
 <div class="form-group row">
   <div id="new-user">
-    <p class="col-sm-12">Enter <%= t('hyrax.account_label') %> (one at a time)</p>
+    <p class="col-sm-12"><%= t('.enter', account_label: t('hyrax.account_label')) %></p>
     <p class="sr-only">Use the add button to give access to one <%= t('hyrax.account_label') %> at a time (it will be added to the list below).  Select the user, by name or <%= t('hyrax.account_label') %>. Then select the access level you wish to grant and click on Add this <%= t('hyrax.account_label') %> to complete adding the permission.</p>
     <div class="col-sm-5">
       <label for="new_user_name_skel" class="sr-only"><%= t('hyrax.account_label') %> (without the <%= t('hyrax.directory.suffix') %> part)</label>
-      <%= text_field_tag 'new_user_name_skel', nil %>
+      <%= text_field_tag 'new_user_name_skel', t('.user_search') %>
```
jcoyne: why are you adding text into the skeleton template?

laritakr: What's a "skeleton template"? 

This is used for the permission form, and I wanted it to not use the default text. This allows for internationallization to work for the text.

no-reply: @jcoyne can you clarify your question? I don't think anyone understands the significance of "skeleton template", and searching for related documentation hasn't been fruitful?

jcoyne: the skeleton template is what the javascript uses to make a new row.  Since it's a new row, it should always be `nil`, right?  Why stick a translation on this particular line?

jcoyne: This is the value of the form field, so it's not appropriate to put "Search for a user" here. That should either be in a label or in a placeholder attribute.

app/views/hyrax/base/_share_with.html.erb
```diff
@@ -1,14 +1,6 @@
-<p>You may grant &quot;View/Download&quot; or &quot;Edit&quot; access for specific users and/or groups to files.
-  Enter a valid <%= t('hyrax.account_name') %>, one at a time, select the access level for that user and click
-  <button class='btn btn-xs btn-inverse' onclick='return false;'><i class='glyphicon glyphicon-plus'></i>Add</button>.
-</p>
-
+<p /><p><%= t('.share_with_html', account_name: t('hyrax.account_name')) %></p>
```
jcoyne: What's going on with the empty `<p />` tag here? Can that be removed?

app/views/hyrax/file_sets/_versioning.html.erb
```diff
@@ -1,27 +1,29 @@
...
-          <%= version.current? ? 'Current version:' : 'Restore from' %>
+          <%= version.current? ? t('.current') : t('.restore_from') %>
           <%= version.created %> [by <%= version.committer %>]
         </label>
       </div>
     <% end %>
     <div id="save_version_note" class="alert hide">You must click &quot;Save Revision&quot; to revert a previous version of this file</div>
-    <%= f.button '<i class="glyphicon glyphicon-save"></i> Save Revision'.html_safe, type: 'submit', class: 'btn btn-primary', onclick: "confirmation_needed = false;", name: "revert_submit", id: "revert_submit" %>
+    <%= f.button :button, name: "revert_submit", id: "revert_submit", onclick: "confirmation_needed = false;", class: "btn btn-primary", type: 'submit' do %>
+      <i class="glyphicon glyphicon-save"></i> <%= t('.save') %>
```
jcoyne: Why did you keep the icon on this button but remove it on the `app/views/hyrax/file_sets/_permission.html.erb` ?

app/views/hyrax/base/_share_with.html.erb
laritakr: I did it to improve the spacing under the Share With heading below. 

These views need more extensive CSS work to get them consistent with other views, and there wasn't a class here to modify so I did it the easy way. I'll just take it out and leave that fix for someone else to do the right way.

screen shot 2018-07-12 at 10 05 08 am

app/views/hyrax/file_sets/_versioning.html.erb
laritakr: An oversight. I didn't notice the icon when I changed this view. I assume we want the icons gone to be more consistent with other views, so I'll remove it here as well.

app/views/hyrax/base/_share_with.html.erb
jcoyne: Thanks. that makes sense to me.

PR #3148, opened 2018-07-10
honor ability on workflow unavailable page

### LINKED ISSUE #3104: Private works minimally visible to all users by url if pending review
### Descriptive summary

When a work is created with the one-step mediated workflow a non-authorized user is able to access a minimal view of the work by url.  The work is not searchable.  Once the work has been approved then the user is prompted with a login.  This is a small security hole which leaks the work's existence, title, type, and status.

### Expected behavior
Unauthorized users should never be able to see any information about items that are private.

### Actual behavior
Unauthorized users can see minimal information about the item while it is pending review.

1. Create a private work with the one-step mediated deposit workflow, but don’t approve it.
1. In an incognito window, navigate to the url of the new un-approved work.  You can see some limited information including title.
1. Approve the work
1. In the incognito window, refresh and you’ll be prompted to login.

kdid:
Fixes #3104

* For suppressed works unavailable due to workflow, don't show even the minimal unavailable page to the user unless they have read access. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/controllers/hyrax/generic_works_controller_spec.rb
+++ b/spec/controllers/hyrax/generic_works_controller_spec.rb
@@ -199,9 +199,10 @@ RSpec.describe Hyrax::GenericWorksController do
       end
       let(:work) { instance_double(GenericWork, id: '99999', to_global_id: '99999') }
 
-      context 'with a user lacking workflow permission' do
+      context 'with a user lacking both workflow permission and read access' do
         before do
           allow(SolrDocument).to receive(:find).and_return(document)
+          allow(controller.current_ability).to receive(:can?).with(:read, document).and_return(false)
         end
         let(:document_list) { [] }
         let(:document) { instance_double(SolrDocument, suppressed?: true) }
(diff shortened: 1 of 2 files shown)


spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -199,18 +199,34 @@
       end
       let(:work) { instance_double(GenericWork, id: '99999', to_global_id: '99999') }
 
-      context 'with a user lacking workflow permission' do
+      context 'with a user lacking both workflow permission and read access' do
         before do
           allow(SolrDocument).to receive(:find).and_return(document)
+          allow(controller.current_ability).to receive(:can?).with(any_args).and_return(false)
```
cjcolvar: How about making this `.with(:read, SolrDocument)`?  I worry that `any_args` is too broad a scope and might bite us later since the allow is for the whole get request and not just a call to `document_not_found!`

no-reply: :+1:

kdid:
Thanks @cjcolvar -  I updated it.

PR #3151, opened 2018-07-16
Add a method to indicate enabled status of save button based on valid…

### LINKED ISSUE #380: Improve interaction between save button and file uploads in work form
Part of the work done to close out the Sufia 7.3.0 release included the following JavaScript changes, which fixed couple error conditions involving how in-progress uploads interact with the save button in the work form:

https://github.com/projecthydra/sufia/pull/3069

The downside of this implementation is that the save button does *appear* to be clickable before all uploads have finished, but it's not actually clickable. (IOW, this isn't a bug but rather a UX enhancement.)

The save button should be disabled (not merely unclickable via `preventDefault()`) while any uploads are running.

@jcoyne made the following suggestions about how we might implement this:
> I think we need a `isSaveButtonActive` method that checks both form validity and `this.uploads.inProgress`. It should be triggered when a upload starts or finishes (the `fileuploadstarted` and `fileuploadstopped` events), or the checkbox is checked, or the metadata is changed.

adamjarling:
… form state and no file uploads currently in progress.

Fixes #380 

This PR introduces a new method, `isSaveButtonEnabled` on the `SaveWorkControl` class, which expands the check for enabled status on the form Save button.   It checks against any current file uploads in progress.

I went the route of hooking into the existing pattern this form uses for checking file upload status based on the `inProgress` method in `UploadedFiles.es6`, instead of the the jQuery File Upload events which probably could be also consumed to arrive at a similar state:
https://github.com/blueimp/jQuery-File-Upload/wiki/Options#callback-options

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/save_work/save_work_control.es6
+++ b/app/assets/javascripts/hyrax/save_work/save_work_control.es6
@@ -108,13 +108,18 @@ export default class SaveWorkControl {
 
   // Called when a file has been uploaded, the deposit agreement is clicked or a form field has had text entered.
   formStateChanged() {
-    this.saveButton.prop("disabled", !this.isValid());
+    this.saveButton.prop("disabled", !this.isSaveButtonEnabled);
   }
 
   // called when a new field has been added to the form.
   formChanged() {
     this.requiredFields.reload();
     this.formStateChanged();
+  }
+
+  // Indicates whether the "Save" button should be enabled: a valid form and no uploads in progress
+  get isSaveButtonEnabled() {
+    return this.isValid() && !this.uploads.inProgress;
   }
 
   isValid() {
--- a/spec/javascripts/save_work_spec.js
+++ b/spec/javascripts/save_work_spec.js
@@ -173,6 +173,56 @@ describe("SaveWorkControl", function() {
 
   });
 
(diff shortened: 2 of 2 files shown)


chrisdaaz:
during testing on `nurax-dev`, i was able to save the work during the file upload process, so long as i had satisfied the metadata and deposit agreement requirements. it used to be the case that the `Save` button wouldn't be clickable unless there were files attached to the work.

adamjarling:
@chrisdaaz This sounds like a regression I may have caused when updating this item.  I'll look into it this afternoon...

adamjarling:
Actually with this PR the Save button should not be enabled until after the file upload process is completed.   Wonder if this change is live on `nurax-dev` yet?

adamjarling:
@chrisdaaz Just checked on https://nurax.curationexperts.com/, and looks like updates have made their way in.  For some reason, the file upload Save button behavior is different than my local environment.  I'll keep digging into it and try to see why.

chrisdaaz:
@adamjarling i learned that it's a configuration in `hyrax.rb` on `nurax` and `nurax-dev` that allows works to be created without files, which is not default behavior, thus the confusion.

PR #3153, opened 2018-07-17
Move license attribute row from _metadata partial to _attribute_rows …

### LINKED ISSUE #394: Move rights rows from _metadata to _attribute_rows
I'm not sure why these are here; it seems reasonable that people may want to order them differently with the rest of the attribute rows. Does anyone have a memory of why they are separated out this way? As it is they are somewhat hidden when needed.

https://github.com/projecthydra-labs/hyrax/blob/master/app/views/hyrax/base/_metadata.html.erb#L8-L10

adamjarling:
…partial

Fixes #394 

This PR attempts to fix "Move rights rows from _metadata to _attribute_rows", but see issue https://github.com/samvera/hyrax/issues/394 for needed clarification of what's the acceptance criteria for this issue.   It's not clear to me.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_attribute_rows.html.erb
+++ b/app/views/hyrax/base/_attribute_rows.html.erb
@@ -12,4 +12,4 @@
 <%= presenter.attribute_to_html(:resource_type, render_as: :faceted, html_dl: true) %>
 <%= presenter.attribute_to_html(:source, html_dl: true) %>
 <%= presenter.attribute_to_html(:rights_statement, render_as: :rights_statement, html_dl: true) %>
-
+<%= presenter.attribute_to_html(:license, render_as: :license, html_dl: true) %>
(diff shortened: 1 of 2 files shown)


cjcolvar:
We actually made this same change in our hyrax app so I think this is a good change.

no-reply:
I'd like to get this in in the near future, but I have a concern:

Many apps override `_attribute_rows`, but not `_metadata`, which means this change will remove `license` from their runtime copies of `_metadata`, but *not* add it to `_attribute_rows`.

I'd like to hold this for the time being. I've put it on the agenda for Samvera Tech next week.

RudyOnRails:
Looks like this agenda item got lost in the setup of the next week's agenda: https://wiki.duraspace.org/display/samvera/Copy+of+Samvera+Tech+Call+2018-07-25

I have added it to the new agenda for next week..

no-reply:
There is consensus on [Samvera Tech](https://wiki.duraspace.org/display/samvera/Samvera+Tech+Call+2018-08-15) is that this would constitute a breaking change and require bumping the version to 3.0.

Since it's a small change, I'm inclined to hold it until we have other API changes in the pipeline.

no-reply:
Considering this along with #3121 as motivating 3.0.0 soon.

Requesting feedback, especially from @vantuyls @samvera/hyrax-code-reviewers @samvera/hyrax-repo-managers

adamjarling:
@no-reply What do you suggest for this?  Ok to merge now?

PR #3154, opened 2018-07-17
Fixes file count in dashboard AdminSets

### LINKED ISSUE #446: File count in collections panel of admin dashboard tops out at 10
screen shot 2017-02-17 at 5 09 29 pm

cc @andjsmit This was what I was worried about ☝️

kdid:
Fixes #446 

* Show correct count of files in each AdminSet in Admin Dashboard

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/admin_set_service.rb
+++ b/app/services/hyrax/admin_set_service.rb
@@ -33,10 +33,11 @@ module Hyrax
       results = ActiveFedora::SolrService.instance.conn.get(
         ActiveFedora::SolrService.select_path,
         params: { fq: query,
+                  rows: 0,
                   'facet.field' => join_field }
       )
       counts = results['facet_counts']['facet_fields'][join_field].each_slice(2).to_h
-      file_counts = count_files(results)
+      file_counts = count_files(admin_sets)
       admin_sets.map do |admin_set|
         SearchResultForWorkCount.new(admin_set, counts[admin_set.id].to_i, file_counts[admin_set.id].to_i)
       end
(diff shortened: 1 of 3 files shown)


cjcolvar:
@kdid The issue being fixed here mentions a limit of 10 files being reported.  Could you please add a test that shows a count > 10?

laritakr:
@kdid I'm having difficulty seeing how the spec actually tests the cause of the bug you are fixing to verify the fix. It appears to me that this is mocking out the search to provide results, but the search itself was the problem. I want to make sure that the specs are actually doing a meaningful test. Can you clarify?

kdid:
@laritakr - What do you recommend?

The initial problem was that the code did not do an actual search for files in each AdminSet, but instead just counted whatever files happened to be contained in the first 10 rows of the search result that was designed to get the work count. (which is why it generally was 10 files but could have been more or less than 10 files depending on how many happened to be attached to those 10 works). 

So now the `.count_files` method is doing a separate query to actually find the number of files in admin sets. This is a private method - I think we generally consider private methods implementation details and don't test them directly. The method under test: `search_results_with_work_count` is able to use the results from what's returned by the `count_files` method - whether it's below or above 10. I'm a little hesitant with our concerns with test performance to create 10+ actual objects.

laritakr:
@kdid 

My approach would be to try to create solr docs only for the searches, and not the actual active fedora objects. I assume that only solr is needed for the searches. To be useful, the admin set service specs should verify that it can actually perform a search. From what I see, it is not testing returning an actual count of works or files, but it DOES query for admin sets.

If the admin set service is adequately tested, anywhere else in the code should be able to safely mock out the search results or create the bare minimum objects, and not have to test the search service itself. (You definitely don't want to test 10+ files in the feature spec where you are actually creating the objects).

laritakr:
@kdid you might want to look at how the collection_size_service_spec is setting up the solr docs for the specs and mimic that.

no-reply:
@laritakr @cjcolvar does this latest commit address your test concerns?

PR #3158, opened 2018-07-19
Update _head_tag_content.html.erb <meta> tag to not disable zooming f…

### LINKED ISSUE #203: Do not disable zooming.
`<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">`  should be `<meta name="viewport" content="width=device-width, initial-scale=1.0">`, via http://a11yproject.com/posts/never-use-maximum-scale/

adamjarling:
…or accessibility improvement

Fixes #203 

This updates the `<meta .../>` tag in the header to improve accessibility by not disabling zooming.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/layouts/_head_tag_content.html.erb
+++ b/app/views/layouts/_head_tag_content.html.erb
@@ -7,7 +7,7 @@ signed in %>
     <%= tag :meta, name: 'current-user', data: { user_key: current_user.user_key } %>
 <% end %>
 <!-- added for use on small devices like phones -->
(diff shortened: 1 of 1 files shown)


jcoyne:
I think the `1.0` could just be `1` but NBD.  Also wonder if we want to add `
shrink-to-fit=no`?

adamjarling:
@jcoyne Think it's worth adding that in?   If so, I will... these builds go green about 1 in 3 for me, so could be another few hours to get that in if so.

PR #3159, opened 2018-07-20
Add checkbox label to table header for checkbox column in Work and Co…

### LINKED ISSUE #126: Checkbox on Dashboard Works/Collections pages needs a label
<a href="https://github.com/jhallida"></img></a> **Issue by [jhallida](https://github.com/jhallida)**
_Thursday May 05, 2016 at 14:27 GMT_
_Originally opened as https://github.com/projecthydra/sufia/issues/1982_

----

A checkbox hides/shows the sort by and batch edit options in many pages, but this checkbox has no label. Also it is not lined up with the corresponding gear icon.

adamjarling:
…llection listings pages

Fixes #126 

Add a label to the checkbox table header for checkbox column in Works and Collections listing pages.  There was no text specified for the label in the issue, so until someone has a better option, for now "Select" is the checkbox label.

Also cleaned up the layout of the Collections listing table headers a bit, reducing the header horizontal padding to get all header text on one line (at least for English language).

**Screenshot after:**

works-list-page-checkbox-label-after

Screenshot before:

works-list-page-checkbox-label-before

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_collections.scss
+++ b/app/assets/stylesheets/hyrax/_collections.scss
@@ -154,10 +154,16 @@ button.branding-banner-remove:hover {
     td {
       text-align: center;
       vertical-align: top;
+      padding: 8px 2px;
 
+      &:nth-child(1),
       &:nth-child(2) {
         text-align: left;
       }
+    }
+
+    .check-all-checkbox {
+      margin-right: 5px !important;
     }
 
     .modal {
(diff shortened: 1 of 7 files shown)


cjcolvar:
@adamjarling Is it possible to separate this into two PRs: the "select" text and the CSS layout cleanup?

adamjarling:
@cjcolvar I'd say they go together, and the general cleanup allows for the "select" label to be readable.

PR #3160, opened 2018-07-20
Attach visibilityComponent to the DOM as other JS objects to allow be…

j-dornbusch:
Attach visibilityComponent to the DOM as other JS objects to allow behavior customization

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/save_work/visibility_component.es6
+++ b/app/assets/javascripts/hyrax/save_work/visibility_component.es6
@@ -14,6 +14,7 @@ export default class VisibilityComponent {
     this.form.on('submit', () => { this.enableAllOptions() })
     this.showForm()
     this.limitByAdminSet()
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/hyrax/save_work/visibility_component.es6
```diff
@@ -14,6 +14,7 @@ export default class VisibilityComponent {
     this.form.on('submit', () => { this.enableAllOptions() })
     this.showForm()
     this.limitByAdminSet()
+    element.data('visibility_component', this)
```
adamjarling: @j-dornbusch This is setting `data-visibility_component="... this "` on the passed in element?  Do you need all the data from `this`?   Does the value need to be serialized?   I'm not sure how you're using here, just wondering if the value could be lighter.

PR #3194, opened 2018-08-10
Add label to dashboard manage proxies proxy user dropdown for accessi…

### LINKED ISSUE #3039: Dashboard Authorize Proxies form field dropdown add label
### Descriptive summary

From the Accessibility 360 Report:

> This form field is not labeled.
> 
> All form fields should have labels (typically via a label tag) which are properly associated with their fields (via the `for` attribute). The best approach is to use this method and delete the "sr-only" class, but it is also acceptable to move the "sr-only" class to the link tag.

authorize-proxies

### Rationale

Accessibility improvement.

### Tasks
- [x] Form field label added, and/or move the `sr-only` class to the link tag.

adamjarling:
…bility

Fixes #3039 

Adds a form `<label />` to Dashboard > Activity > Manage Proxies authorize Proxies dropdown for improved accessibility.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_proxy-rights.scss
+++ b/app/assets/stylesheets/hyrax/_proxy-rights.scss
@@ -8,3 +8,9 @@ table#authorizedProxies {
     }
   }
 }
+
+.proxy-rights {
+  label {
+    display: block;
+  }
+}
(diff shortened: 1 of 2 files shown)


laritakr:
I looked at the spec that failed here, and debugged it. 
```  1) Hyrax::HomepageController#index with a document not created this second sets recent documents in the right order
     Failure/Error: expect(create_times).to eq create_times.sort.reverse
     
     ArgumentError:
       comparison of String with nil failed
     # ./spec/controllers/hyrax/homepage_controller_spec.rb:74:in `sort'```

The Generic Works being tested here do not include `date_uploaded`, which it is attempting to sort by, so the line `create_times = assigns(:recent_documents).map { |d| d['date_uploaded_dtsi'] }` returns [nil, nil, nil, nil] for the value of `create_times`.

PR #3211, opened 2018-08-17
Fixes #3210 Restores devise trackable module.

### LINKED ISSUE #3210: user.last_sign_in_at is not a default devise feature now
### Descriptive summary

Devise 4.5.0 no longer includes the trackable module by default to be more GDPR-friendly. user.last_sign_in_at= is expected by `spec/views/hyrax/admin/users/index.html.erb_spec.rb:11`

Should we reenable the trackable module, or remove our dependency on it?

dlpierce:
Fixes #3210 

Restores devise trackable module to fix the test suite.

Devise removed the trackable module from their generator, breaking our test app and newly generated hyrax applications. This restores the trackable module in the User model and db migration. 

We may want to consider if removing our dependence on the trackable module is desirable.

* Manually insert the trackable modules into the devise setup in the User model when installing hyrax.
* Add a migration duplicating the trackable db columns previously added by the default devise migrations.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/install_generator.rb
+++ b/lib/generators/hyrax/install_generator.rb
@@ -107,6 +107,12 @@ module Hyrax
       end
     end
 
+    def inject_devise_trackable
+      inject_into_file 'app/models/user.rb', after: /\s*devise :database_authenticatable, :registerable,\s*\n/ do
+        "         :trackable,\n"
+      end
+    end
+
     # Add behaviors to the application controller
     def inject_hyrax_application_controller_behavior
       file_path = "app/controllers/application_controller.rb"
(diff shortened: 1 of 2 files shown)


tap87:
Devise removed the trackable module for supporting GDPR, here is the issue: https://github.com/plataformatec/devise/issues/4849
We actually removed it from our app as well.

jcoyne:
I agree with @tap87, maybe we should do the default thing and update `hyrax/admin/users/index.html.erb` so that it checks for the trackable module, and only then shows the last sign in at column? I don't think we want to force users to install trackable.

jcoyne:
I'm closing this since https://github.com/samvera/hyrax/pull/3218 was merged.

PR #3212, opened 2018-08-21
Major review of the Portuguese translation

mcwhitaker:
This commit is a major review of the Brazilian Portuguese version of the hyrax locale text. I consider this a first phase. I have to actually use the application more heavily to see all the terms in their context and validate some choices.
Besides fixing some really funny errors due to the ambiguity of the English term (is _parent work_ a work by a parent? Is a _job_ you submit an employment? Is _date_ a romantic meeting? Is _collection_ related to financial debt?), some major decisions had to be made:

1.	**Work** - there are two possible words in Portuguese, **trabalho** and **obra**. I consulted a Librarian at the _Escola de Comunicações e Artes - USP_, Sarah Lorenzon Ferreira. She gave me several examples where the use of **obra** has been established in Brazil, so I followed the established practice. This implied a very large number of changes because **trabalho** is masculine and **obra** is feminine, so all the adjectives, conjunctions, etc, had to be changed to conform to the gender of the word chosen.
2.	**Release** - the original machine translation had used 4 different terms to translate this word: **lançamento, divulgação, liberação,** and **libertação,** this last being totally wrong in this context, and the first one being inadequate. I chose **divulgação** most of the time, with **liberação** being used in a couple of places due to context.
3.	**Lease** – several different terms were in use: **locação, concessão, arrendamento, empréstimo**, **aluguel**. These are all possible translations depending on whether you are leasing a DVD, a car, a farm, a book, a house. In the Library context, the term in Portuguese is **empréstimo**.
4.	**Nested collections** - the literal translation makes no sense in Portuguese. It was hard to find a good translation for it, but I settled on **coleções embutidas**. I am not married to this, so if any native speaker of Brazilian Portuguese objects to it, I am open to changes as long as it is not the literal translation (in Portuguese **nests** are for birds and other creatures).
5.	**Branding** - there is no Portuguese translation for this word in the marketing context. The English word is used.
6.	**Viewer** - was machine translated into **visualizador**, but this word in Portuguese cannot refer to a person. It can be used for devices or tools, such as the Universal Viewer. The literal translation would be **vedor**, but this word basically only exists in the dictionary and Brazilians prefer to use a sentence to refer to the person who has viewing access to something. This is what I did.
7.	**Add** - had been translated into **adicionar**. Not the best one (we are not doing math), so I picked **acrescentar** instead.
8.	**Manage** - most of the time, I accepted the translation into **gerenciar**. On a few instances, I used **curatelar**, which is a more technical term related to _curators_ and _curation_ and _curate_. I foresee objections from other native speakers on this one. Again, I am open to changes.

### DIFF
--- a/config/locales/hyrax.pt-BR.yml
+++ b/config/locales/hyrax.pt-BR.yml
@@ -7,17 +7,17 @@ pt-BR:
   activemodel:
     errors:
       messages:
-        cannot_have_child_nested: não pode ter um filho aninhado nele
-        cannot_nest_in_parent: não pode aninhar no pai
-        cannot_remove_relationship: a permissão é inadequada para a remoção do relacionamento de nidificação
-        conflict: Suas alterações não podem ser salvas porque outro usuário (ou trabalho em segundo plano) atualizou este %{model} depois de começar a editar. Certifique-se de que todos os anexos de arquivos tenham sido concluídos com sucesso e tente novamente. Este formulário foi atualizado com a cópia salva mais recente do %{model}.
-        exceeds_maximum_nesting_depth: o nidificação excede a profundidade máxima de nidificação permitida.
-        is_not_nestable: não é nestable
+        cannot_have_child_nested: não pode ter obra embutida
+        cannot_nest_in_parent: não pode ser embutida nesta obra
+        cannot_remove_relationship: a permissão é inadequada para a remoção da relação de embutidura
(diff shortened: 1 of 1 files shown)


no-reply:
Thank you so much!

It looks like there might be a minor syntax issue causing the build failure. I'll see if we can get someone to fix it up and we can get this into a release ASAP!

mcwhitaker:
Thanks, Tom. I looked at the error message and I don't understand what the syntax error is. I can look at it a little more, but if anyone sees it right away, that would be great.

mcwhitaker:
I fixed an indentation and that might be the issue.

PR #3218, opened 2018-08-27
Don't show trackable fields if Devise trackable isn't installed

### LINKED ISSUE #3210: user.last_sign_in_at is not a default devise feature now

jcoyne:
Fixes #3210

### DIFF
--- a/spec/views/hyrax/admin/users/index.html.erb_spec.rb
+++ b/spec/views/hyrax/admin/users/index.html.erb_spec.rb
@@ -3,19 +3,43 @@ RSpec.describe 'hyrax/admin/users/index.html.erb', type: :view do
   let(:users) { [] }
 
   before do
-    (1..4).each { |i| users << build(:user, display_name: "user#{i}", email: "email#{i}@example.com", last_sign_in_at: Time.zone.now - 15.minutes, created_at: Time.zone.now - 3.days) }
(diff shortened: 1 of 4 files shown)


spec/views/hyrax/admin/users/index.html.erb_spec.rb
```diff
@@ -3,19 +3,38 @@
...
-    expect(page).to have_content("Roles")
-    expect(page).to have_content("Last access")
-    (1..4).each do |i|
-      expect(page).to have_content("email#{i}@example.com")
+  let(:page) { Capybara::Node::Simple.new(rendered) }
+
+  context 'when show_last_access? is true' do
+    let(:show_last_access) { true }
+
+    it 'draws user list with all users' do
```
cjcolvar: This test fails because the internal test app doesn't get generated with the trackable module.  Maybe the solution here is to add more stubbing even though that does leave some code untested?

jcoyne: Yep. I was using my old generated code.

jcoyne: Should be fixed now.

PR #3223, opened 2018-08-29
Lock Devise to pre-4.5.0

no-reply:
Backported fix for #3210

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -38,6 +38,7 @@ EOF
   spec.add_dependency 'browse-everything', '>= 0.10.5'
   spec.add_dependency 'blacklight', '~> 6.9'
   spec.add_dependency 'blacklight-gallery', '~> 0.7'
+  spec.add_dependency 'devise', '<= 4.4.99'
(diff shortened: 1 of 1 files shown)


hyrax.gemspec
```diff
@@ -38,6 +38,7 @@ EOF
   spec.add_dependency 'browse-everything', '>= 0.10.5'
   spec.add_dependency 'blacklight', '~> 6.9'
   spec.add_dependency 'blacklight-gallery', '~> 0.7'
+  spec.add_dependency 'devise', '<= 4.4.99'
```
jcoyne: Maybe drop in a comment here about why this is locked?

no-reply: is the commit message good enough?

jcoyne: Not really, because it would be several steps:
1. Look at the gemspec and say, "Hmm, why is this here"
2. then `git blame` and figure out it was this commit. Which says "See #3210"
3 Open up github and find issue 3210 

It would be kinder to future code readers to just drop a comment here.

Issue #3187, opened 2018-08-07
Cross-site scripting vulnerabilities


hortongn:
### Descriptive summary

For a school project, a student at the University of Cincinnati performed a manual penetration test on UC's Scholar@UC application (based on Hyrax 1.x).  He found four potential cross-site scripting vulnerabilities where the work or file title fields were not being properly sanitized.  

We then replicated the attacks on nurax-dev.curationexperts.com and found the same behavior.  So the vulnerabilities appear to be in the current version of Hyrax.

### Vulnerability 1

1. Go to the Add New Generic Work form
1. Create and save a work (no file is needed) with the title `<ScriPt>prompt("Confirm Password")</ScRIPT>`
1. On the work's show page click the `Citations` button
1. You will see a popup asking the user to enter their password.

(the vulnerability seems to be in the Chicago citation format when it tries to list the work title)

### Vulnerability 2 

1. Go to the Add New Generic Work form
1. Create and save a work (no file is needed) with the title `<ScriPt>prompt("Confirm Password")</ScRIPT>`
1. Visit /dashboard/my/works
1. Under `Actions` click the `Select` dropdown for the work you just created
1. Click on `Delete Work` and confirm the delete
1. The page you land on after delete will display a popup asking the user to enter their password.

(The vulnerability appears to be in the flash message that tries to list the deleted work's title)

### Vulnerability 3

1. Go to the Add New Generic Work form
1. Create and save a work using the title `<ScriPt>prompt("Confirm Password")</ScRIPT>` and attach at least 1 file
1. After the work is saved, click `Edit` to go to the work's edit form.
1. Change the work to a different visibility setting and save the work
1. You will see a popup asking the user to enter their password.

(the vulnerability seems to be on the "Apply changes to contents?" page when it tries to list the work title)

### Vulnerability 4

1. Make sure the UniversalViewer is configured
1. Go to the Add New Generic Work form
1. Create and save a work using the title `<img src=xx:x onerror=eval('\x61ler\x74(1)') /> and attach 1 image file
1. When you are redirected to the work's show page you will see a popup with the number 1

(The vulnerability appears to be in the UniversalViewer.  The popup only shows when the UV is present on the page)

NOTE: The same behavior exists if `<img src=xx:x onerror=eval('\x61ler\x74(1)') />` is used for the **file** title instead of **work** title.

no-reply:
@cjcolvar could I get your input on **Vulnerability 4** above?

[...]

PR #3225, opened 2018-08-29
Don't allow XSS on title in Chicago Formatter

no-reply:
There's probably a larger fix related to forms here, but the chicago formatter
needn't be so permissive.

Related to #3187.

Acceptance: Check XSS referenced in #3187, **Vulnerability 1**

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/citations_behaviors/formatters/chicago_formatter.rb
+++ b/app/helpers/hyrax/citations_behaviors/formatters/chicago_formatter.rb
@@ -13,11 +13,11 @@ module Hyrax
           text = "<span class=\"citation-author\">#{text}</span>" if text.present?
           # Get Pub Date
           pub_date = setup_pub_date(work)
-          text << " #{pub_date}." unless pub_date.nil?
+          text << " #{whitewash(pub_date)}." unless pub_date.nil?
 
           text << format_title(work.to_s)
           pub_info = setup_pub_info(work, false)
-          text << " #{pub_info}." if pub_info.present?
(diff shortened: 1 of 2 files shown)


no-reply:
> If Chicago Formatter is returning a string of HTML, shouldn’t something downstream notice it’s not Html_safe? and do the escape?

no-reply:
@jcoyne:

> If Chicago Formatter is returning a string of HTML, shouldn’t something downstream notice it’s not Html_safe? and do the escape?

Yes I think so. This is clearly not the case, however, and the issue seems not to be isolated to the citation formatters. Pending a more complete fix, this patches a known XSS vulnerability.

no-reply:
@jcoyne after a closer look, this formatter marks its output `#html_safe`, which was the real source of the issue.

I've added whitewashing for the remainder of the fragments compiled into the formatted output.

app/helpers/hyrax/citations_behaviors/formatters/chicago_formatter.rb
```diff
@@ -13,11 +13,11 @@ def format(work)
           text = "<span class=\"citation-author\">#{text}</span>" if text.present?
           # Get Pub Date
           pub_date = setup_pub_date(work)
-          text << " #{pub_date}." unless pub_date.nil?
+          text << " #{whitewash(pub_date)}." unless pub_date.nil?
```
jcoyne: Can we use `sanitize` in the helpers?  That seems to be the built in fix.

PR #3229, opened 2018-08-30
Don't blindly mark flash strings HTML safe

### LINKED ISSUE #3228: Refactor HTML Safety of flash messages
### Descriptive summary

`app/views/_flash_msg.html.erb` blindly marks any string given for a flash message as `#html_safe` (see: 930c9ad9f3a). In fact, these strings are often generated substantially from user input, rendering us vulnerable to XSS and injection attacks (see **Vulnerability 2** in #3187).

The html safety behavior should be removed from this view, and all downstream providers of these strings should be responsible for marking their output html safe **only** when it has been properly scrubbed.

### Rationale

This line creates a whole class of vulnerabilities related to user input appearing in flash messages.

### Expected behavior

When inputting a title with html tags, then deleting an object, the flash message should escape the user input html. 

### Actual behavior

User input is not scrubbed in any flash messages. Several cleanup efforts that existed previously were removed in 930c9ad9f3a.

### Related work

#3187, #3225

no-reply:
This view partial had originally marked all of the strings it received as `#html_safe`, effectively trusting all content. Using sanitize essentially does the opposite: distrusting all flash message content. Any tags or attributes not in the whitelist (https://github.com/flavorjones/loofah/blob/master/lib/loofah/html5/whitelist.rb) will be removed, regardless of the source.

This seems like the best solution, since flash messages are serialized as strings, so the providers cannot give us carefully constructed `#html_safe` strings. In general, flash messages shouldn't require markup more complex than basic formatting and links.

Fixes #3228
Connected to #3187

* Remove `map(&:html_safe)`!!!
* Call `sanitize` on flash messages

* Verify that this fixes *Vulnerability 2* in #3187.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/_flash_msg.html.erb
+++ b/app/views/_flash_msg.html.erb
@@ -2,7 +2,7 @@
   <% if flash[type].present? %>
     <div class="alert <%= flash_dom_class %> alert-dismissable" role="alert">
       <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-      <%= safe_join(Array.wrap(flash[type]), tag(:br)) %>
+      <%= sanitize Array.wrap(flash[type]).join(tag(:br)) %>
     </div>
     <% flash.delete(type) %>
   <% end %>
--- a/app/views/_flash_msg.html.erb
+++ b/app/views/_flash_msg.html.erb
@@ -2,7 +2,7 @@
   <% if flash[type].present? %>
     <div class="alert <%= flash_dom_class %> alert-dismissable" role="alert">
       <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-      <%= safe_join(Array.wrap(flash[type]).map(&:html_safe), tag(:br)) %>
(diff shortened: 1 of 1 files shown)


jcoyne:
If I recall correctly, there is no way for flash messages, which get serialized into the session to be marked as HTML safe.  I believe there are one or two places where we've inserted html link tags into the flash, so this change may break those links.

no-reply:
@jcoyne I think I follow. Since the flash messages are serialized into the session using the JSON serializer, they come out as strings, even if there were originally safe buffers?

Yeah, it seems like `sanitize` and a conservative whitelisted scrubber are the way to go here.

Do we have any example flash messages with HTML formatting?

jcoyne:
@no-reply that's what I'm thinking. Sorry, I don't know where we have html in the flash though. Maybe these?
```
$ git grep flash app/controllers/|grep html
app/controllers/concerns/hyrax/works_controller_behavior.rb:            flash[:notice] = view_context.t('hyrax.works.create.after_create_html', application_name: view_context.application_name)
app/controllers/hyrax/batch_uploads_controller.rb:        flash[:notice] = view_context.t('hyrax.works.create.after_create_html', application_name: view_context.application_name)
```

PR #3249, opened 2018-09-10
Wrap filter options in a fieldset element, using a legend for screen …

### LINKED ISSUE #3131: Wrap Filter Options in Fieldset with Legend
### Descriptive summary

In the dashboard for Collections and Works in Hyrax 2.1.0 RC3, there are filter options that are dropdown button for options like "Visibility" or "Status." Each individual dropdown should be contained within a `<fieldset>` with a `<legend>` .

### Rationale

Using `<fieldset>` and `<legend>` ensures that the text description is read to screen reader users when the grouping is navigated to when exploring the interface. The legend text is generally read for each control in the fieldset, so the legend text should be brief and descriptive.

### Expected behavior

Groupings of form controls should be wrapped in a `<fieldset>` and `<legend>`.

### Actual behavior

Bootstrap `btn-group` is being used to organize the controls, but the CSS class is not usable for screen readers.

adamjarling:
…readers to aid accessibility

Fixes #3131 
Wraps the filter options in Works and Collections list pages with a `<fieldset>` element, and use `<legend>` for screen readers to better aid in accessibility.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/my/_facet_layout.html.erb
+++ b/app/views/hyrax/my/_facet_layout.html.erb
@@ -1,7 +1,8 @@
-<div class="btn-group">
+<fieldset class="btn-group">
+  <legend class="sr-only"><%= facet_field_label(facet_field.field) %></legend>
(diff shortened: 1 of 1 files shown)


no-reply:
These builds are successful. Travis and GitHub don't seem to be playing nice today.

Issue #3166, opened 2018-07-25
User can remove (repo admin or any user) permission in Sharing tab on work


julesies:
### Descriptive summary

Hyrax 2.1. We are testing permissions for our go live of hyrax and came across this bug which I confirmed on nurax today.

A user/depositor can remove any person/role via the Sharing tab in the UI including the repo admin. We tested adding a work and having the depositor remove access to the repo admin, which was allowed. The repo admin could still access the work and edit it, but on visiting the sharing tab, the role was not visible. 

### Expected behavior
depositor can only see, remove roles/permissions that they added.  permissions that are removed, are actually removed. 

### Steps to reproduce the behavior

1. Create a work as a regular user
2. Go to sharing tab
3. Remove admin access, hit save
4. log in as admin, find and edit work, hit save

### Done looks like
- [x] Admins cannot have their edit access revoked by any user for any work
- [x] Users cannot override edit access permissions created by the work's admin set or collection
- [x] User view: the "currently shared with" table **hides** rows for `admin` and any groups that exist as participants in the work's admin set or collection
- [x] Admin/ Manager/ Editor view: the "currently shared with" table **shows** `admin` and all users and groups that have edit access to the work (for example, if User A is listed as a Manager of a collection, User A can go to any work in that collect, click `Edit`, view the `Sharing` tab, and see a table that lists all users and groups that have edit access to that work and act according to the settings of the admin set or collection containing that work).

julesies:
@vantuyls @chrisdaaz @no-reply ping!

julesies:
Also, I just tested adding an editor to the work, then revoking the editor and that worked as expected.

chrisdaaz:
@julesies would removing this row from the "Currently Shared With" table satisfy the issue?

![image](https://user-images.githubusercontent.com/24395592/43212286-30b80256-8ff9-11e8-975d-26aef3e34045.png)

i believe admins should always have edit access to all works in the repository. it doesn't seem useful to show the user that admins have edit access to the work they deposited, let alone showing them options to edit or revoke that access.

julesies:
yeah agree it seems odd to display admin access on the work view but it does go beyond admins. We have 3 groups from using role mapper, for example and the user can "remove" any user or group who has access granted to them another way. Since currently the only way to remove an individual users listed on the admin set is by visiting each work, I'm thinking you would have to show that to managers, admins, (or what ever group has the right to edit permissions on works).  

here is an example: (ignore the weird tab bug we have)

[...]

PR #3272, opened 2018-09-13
[WIP] Lock unauthorized permissions from update

laritakr:
Fixes https://github.com/samvera/hyrax/issues/3166

## Details:

- user is permitted to update any work permissions coming from collections they manage
- user is permitted to update work permissions which are non-manager permissions from any collections work belongs to
- user is permitted to update work permissions which are not related to its collections

Process used:
- find all of the other managers of collections a user cannot manage who are not also managers in collections that the user CAN manage.

This gives us the permissions the user is not authorized to update. 

Permission logic was moved into a new shared partial `currently_shared.html.erb` which embeds all of the above logic and uses it to display all of the work or file set's permissions in either a fixed or editable format.

The new shared partial results in a slight reformat of the display of these permissions, as there were previously minor differences between the two displays.

### Work Permissions View
screen shot 2018-09-13 at 2 45 02 pm

### File Set Permissions View
screen shot 2018-09-13 at 2 44 53 pm

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_currently_shared.html.erb
+++ b/app/views/hyrax/base/_currently_shared.html.erb
@@ -1,0 +1,97 @@
+<%# ================================================================= %>
+<%# Lock unauthorized permissions from update:
+    - user is permitted to update any work permissions coming from collections they manage
+    - user is permitted to update non-manager permissions from any Collections
+    - user is permitted to update non-collection permissions
+    Process used:
+    - find all of the work's collections a user can manage
+    - find all of the work's collections a user cannot manage
+    - find all of the other managers of collections a user can manage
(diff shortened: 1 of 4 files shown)


laritakr:
Note: This PR includes a change in the path to two title translations.

laritakr:
Note: the spec failures are (at least mostly) due to needing access_grants to exist for the feature specs to pass. No spec changes were made yet... I was waiting to clarify the desired changes were in place before trying to write specs for them. 

My intention was to move the code out of the view to isolate the testing, once the desired behavior is achieved. I'm not sure yet if we are at this point, but I believe so.

PR #3280, opened 2018-09-14
Prepare 2.3.0 release

no-reply:
Draft release notes are: https://github.com/samvera/hyrax/releases/tag/untagged-e5d52ca19728329f7b59

@samvera/hyrax-code-reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ The Samvera community is here to help. Please see our [support guide](./.github/
 # Getting started
 
 This document contains instructions specific to setting up an app with __Hyrax
-v2.2.4__. If you are looking for instructions on installing a different
+v2.3.0__. If you are looking for instructions on installing a different
 version, be sure to select the appropriate branch or tag from the drop-down
 menu above.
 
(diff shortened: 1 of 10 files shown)


no-reply:
Testing carried out, and release notes written by @chrisdaaz.

We'll probably announce this release as a significant accessibility milestone, since it closes the remaining `a11y` audit issues that are rooted in Hyrax. Hopefully tickets that remain open can be fixed upstream, so adopters can upgrade without further Hyrax releases.

Gemfile
```diff
@@ -5,7 +5,7 @@ gemspec
 
 group :development, :test do
   gem 'coveralls', require: false
-  gem 'i18n-tasks'
+  gem 'i18n-tasks', '0.9.21'
```
jcoyne: Why such pessimism in versions?

no-reply: The new version of `i18n-tasks` breaks our current translation process. This is the version we are using now. Locking to it buys us time to update the process.

Note this only impacts development users trying to run translations through the google translate API.

In other words, is it really pessimism if the next version is actually broken? :)

jcoyne: So, `.22`, `.23`, & `.24` are all broken?  Can you add a comment to that effect here?

no-reply:
@jcoyne I ended up updating the documentation to work with the latest releases. https://github.com/samvera/hyrax/wiki/Translations-(internationalization)

Issue #2575, opened 2018-01-29
admin_set without description cannot display


bess:
### Descriptive summary
In Hyrax 1.0.5 (and 1.x stable) an admin_set without a description cannot be displayed. The error is `ActionView::Template::Error: undefined method `each' for nil:NilClass`

### Expected behavior

I expect to see the admin set show view. 

### Actual behavior

I see an application error. 

### Steps to reproduce the behavior

1. Create an admin_set without a description
1. Go to http://YOUR_HYRAX_APP/admin/admin_sets/YOUR_ADMIN_SET_ID

### Related work

This can be fixed by editing `app/views/hyrax/collections/_collection_description.erb` like this:
```ruby
<% presenter.description.try(:each) do |description| %>
   <p class="collection_description"><%= iconify_auto_link(description) %></p>
<% end %>
```

See https://github.com/curationexperts/laevigata/pull/867/files for example fix.

no-reply:
I think this should be reproducible in 2.x, which suggests we need patches in both places.

no-reply:
I can't reproduce in 2.x

PR #3283, opened 2018-09-17
Display collections without descriptions

no-reply:
`CollectionPresenter#description` is delegated to `#solr_document`, which can return `nil`. When we display the collection, we try to call `#each`, which gives `NoMethodError` in the `nil` case.

Using the safe access operator `&.` avoids this bug and allows display of Collections and AdminSets without any descriptions.

Connected to #2575.

* Create a collection with no description;
* Visit the collection show page and ensure there is no error.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/collections/_collection_description.erb
+++ b/app/views/hyrax/collections/_collection_description.erb
@@ -1,3 +1,3 @@
-<% presenter.description.each do |description| %>
+<% presenter&.description&.each do |description| %>
   <%= simple_format(auto_link(description)) %>
 <% end %>


no-reply:
@jcoyne:

> Have you considered having the description return [] if it’s not present?

This would be a nice change, but isn't sufficient here for two reasons:

  - In practice, applications provide their own implementations of `SolrDocument` via the generator; these normally return `[Array<Object>, nil]`. To fix this bug for existing applications, we need to be tolerant.
  - Changing from returning `nil` in the empty case to `[]` would change the method signature for either this one accessor, or (probably better?) all field accessors on `SolrDocument`. It seems reasonable to expect that some users do `.nil?` checks which would be broken on the change.

In short, this feels like a larger `SolrDocument` reworking, and we need to continue to support current behavior until at least 3.0.0.

no-reply:
> The second &. makes sense given your description, but why is the first needed?

I've not seen any cases where `presenter` is `nil`, but the implementation proposed here seems to give a reasonable default for this partial in that case; i.e. don't display any descriptions if no presenter is given.

I could readily be talked into removing it (I also prefer to avoid proliferation of conditionals). I have the sneaking suspicion that I'm going to be back here in a month re-adding it in response to some honeybadger ticket, but maybe that will prompt a better rethink(?). If you want to ask for the change, I'll put it in.

PR #3286, opened 2018-09-18
Downgrade chromedriver-helper to fix failing capybara tests

no-reply:
Backports #3284 to fix the build.

@samvera/hyrax-code-reviewers


no-reply:
@cjcolvar thanks, I repushed.

PR #3291, opened 2018-09-21
Fixes #3164 - Collection Thumbnails don't show in search results

### LINKED ISSUE #3164: Collection Thumbnails don't show in search results
### Descriptive summary

Hyrax 2.1 - We are seeing that collection thumbnails aren't showing in search results. You can add a thumbnail and it shows in the dashboard collection index, but not the search results. I tested this on nurax and it behaves the same. I thought we confirmed this work in the testing process, but I'm not sure. 

### Rationale

Seems more useful for a thumbnail to be in search results than dashboard. 

1. Create a collection
2. add a thumbnail (must select from image in collection)
3. View collection in search results
4. generic icon is visible

hweng:
Fixes #3164 ; refs #3164

Please see screenshots.
Before changes:
screen shot 2018-09-11 at 12 17 17 pm

And after changes:
screen shot 2018-09-11 at 3 25 41 pm

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/catalog/_thumbnail_list_collection.html.erb
+++ b/app/views/catalog/_thumbnail_list_collection.html.erb
@@ -1,3 +1,4 @@
 <div class="col-md-2">
-  <span class="<%= Hyrax::ModelIcon.css_class_for(Collection) %> collection-icon-search" aria-hidden="true"></span>
-</div>
+  <%= render_thumbnail_tag document, {}, suppress_link: true %>
+</div> 
+
--- a/spec/features/search_spec.rb
+++ b/spec/features/search_spec.rb
@@ -53,7 +53,7 @@ RSpec.describe 'searching' do
       expect(page).to have_content('Search Results')
       expect(page).to have_content "Toothbrush"
       expect(page).to have_content('collection title abc')
-      expect(page).to have_css("span.collection-icon-search")
+      expect(page).to have_selector("//img")
 
       expect(page.body).to include "<span itemprop=\"keywords\"><a href=\"/catalog?f%5Bkeyword_sim%5D%5B%5D=taco&amp;locale=en\">taco</a></span>"
       expect(page.body).to include "<span itemprop=\"keywords\"><a href=\"/catalog?f%5Bkeyword_sim%5D%5B%5D=mustache&amp;locale=en\">mustache</a></span>"
--- a/spec/views/catalog/_thumbnail_list_collection.html.erb_spec.rb
+++ b/spec/views/catalog/_thumbnail_list_collection.html.erb_spec.rb
@@ -1,10 +1,43 @@
 RSpec.describe 'catalog/_thumbnail_list_collection.html.erb', type: :view do
-  before do
-    stub_template 'catalog/_thumbnail_list_collection.html.erb' => '<div class="col-sm-3"><span class="fa fa-cubes collection-icon-search"></span></div>'
-    render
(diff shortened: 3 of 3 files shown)


app/views/catalog/_thumbnail_list_collection.html.erb
```diff
@@ -1,3 +1,9 @@
 <div class="col-md-2">
-  <span class="<%= Hyrax::ModelIcon.css_class_for(Collection) %> collection-icon-search" aria-hidden="true"></span>
+  <% collection_presenter = Hyrax::CollectionPresenter.new(document, current_ability) %>
+  <% if (collection_presenter.thumbnail_path == nil) %>
+    <span class="<%= Hyrax::ModelIcon.css_class_for(Collection) %> collection-icon-search" aria-hidden="true"></span>
+  <% else %>
+    <%= image_tag(collection_presenter.thumbnail_path) %>
```
cjcolvar: Do we need any alt text here?  It seems to me like we do.

app/views/catalog/_thumbnail_list_collection.html.erb
```diff
@@ -1,3 +1,9 @@
 <div class="col-md-2">
-  <span class="<%= Hyrax::ModelIcon.css_class_for(Collection) %> collection-icon-search" aria-hidden="true"></span>
+  <% collection_presenter = Hyrax::CollectionPresenter.new(document, current_ability) %>
+  <% if (collection_presenter.thumbnail_path == nil) %>
+    <span class="<%= Hyrax::ModelIcon.css_class_for(Collection) %> collection-icon-search" aria-hidden="true"></span>
```
cjcolvar: Is this line still tested or is it worth testing this line (maybe in a view test instead of a feature test)?

geekscruff:
@hweng are you able to have a look at the questions from @cjcolvar ? tx

app/views/catalog/_thumbnail_list_collection.html.erb
hweng: @cjcolvar Yes, the image_tag would generate alt text.

app/views/catalog/_thumbnail_list_collection.html.erb
hweng: Thanks for the suggestions. I've added tests for it. Please review the updated commits.

hweng:
@cjcolvar @geekscruff Thanks for reviewing. I've updated tests for this PR. Let me know if you have any other question.

app/views/catalog/_thumbnail_list_collection.html.erb
cjcolvar: I'm seeing the test for the new case displaying the thumbnail (line 18 of the view spec) but nothing for the original case that displays the icon.  Does the feature test cover this case or is the feature test finding the thumbnail as well?  I think it would be nice to retain the original test in the view spec making for two tests that cover both cases.

hweng:
@cjcolvar Yes, it is neat way. Good finding 👍
 Also I tested that the original case that displays the icon is not necessary because it would display the default_image if there is not any collection thumbnail attached. 
Please review the updated code and let me know what do you think about it?

hweng:
@cjcolvar  The extra commit in this PR is removed. Thank you for your reviewing!

cjcolvar:
I'm going to go ahead and merge this since the travis PR build appears red but none of the rspec tests failed.

PR #3303, opened 2018-09-25
[Backport] Remove ActiveFedora 11.5.2 guard

cjcolvar:
This guard artificially breaks Hyrax for explictly supported `ActiveFedora`
versions. When `ActiveFedora` 11.5.3 was released recently, this started hitting
us when `bundle update` runs.

@samvera/hyrax-code-reviewers


no-reply:
are we okay to close this one, since a 2.3.x release now has this fix?

cjcolvar:
Yup, sounds good to me.

PR #3310, opened 2018-09-25
Allow support for `active-fedora` 12

no-reply:
Expand support for `ActiveFedora` to allow use of 12.x releases. This brings us
up to date on current releases.

We explictly add a `Solrizer` dependency, which has been removed from
`ActiveFedora` and inlined to `ActiveFedora::Indexing`. Adding this explictly
retains existing behavior for downstream users. We should remove internal
references to `Solrizer` soon, and remove the dependency altogether on the next
major release.

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -31,7 +31,8 @@ SUMMARY
   # http://guides.rubyonrails.org/maintenance_policy.html
   spec.add_dependency 'rails', '~> 5.0'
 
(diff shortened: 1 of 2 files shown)


jcoyne:
@no-reply do you need to `require 'solrizer'` too?

no-reply:
> @no-reply do you need to require 'solrizer' too?

Good question. If we weren't doing that before, maybe it's not necessary?

PR #3317, opened 2018-10-02
Remove datatables class from All Works display table, which was causi…

### LINKED ISSUE #3306: Non-working pagination link on dashboard: works
### Descriptive summary

on nurax-dev (not sure what version of hyrax it's running!), the dashboard: works view has an extra pagination section that does not work. See bottom right (that is the part that doesn't work). the bottom left nav does work. 

screen shot 2018-09-25 at 11 46 36 am

adamjarling:
…ng double search and pagination

Fixes #3306 

This removes the double search and pagination which was appearing in Works "All Items" page.

**After:**

works-regular-search-pagination

**Before:**

works-extra-search-pagination

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/dashboard/works/_default_group.html.erb
+++ b/app/views/hyrax/dashboard/works/_default_group.html.erb
@@ -1,4 +1,4 @@
-<table class="table table-striped works-list datatable">
+<table class="table table-striped works-list">
   <caption class="sr-only"><%= t("hyrax.dashboard.my.sr.listing") %> <%= application_name %></caption>
   <thead>
     <tr>


app/views/hyrax/dashboard/works/_default_group.html.erb
```diff
@@ -1,4 +1,4 @@
-<table class="table table-striped works-list datatable">
+<table class="table table-striped works-list">
```
no-reply: This looks like it was added in https://github.com/samvera/hyrax/pull/3173. I suspect we need a larger rework of this to allow both sorting and pagination.

@samvera/hyrax-ui-ux-advisors, @vantuyls, @bess

adamjarling: Ah, so `datatables` was added for sorting, but it (from #3173 PR screenshot), looks like it doubled the UI.  Is the solution to rip out the existing pagination and search, and let `datatables` handle pagination, search, and sorting?

no-reply: maybe, but i think that is also problematic, because the datatables only work on data in the current page. i think we would also need to change the relevant search builder to make the datatables work over all results. this problem exists elsewhere (e.g. notifications), as well.

the other alternative is to remove datatables and implement our own sorting.

since the original contribution wasn't part of a particular ticket, i'm tempted to suggest we revert it as proposed here and start from scratch. but i'd also like to get @bess (or maybe @mark-dce's) input before moving forward. can either of you comment on the original motivation for this?

PR #3325, opened 2018-10-10
Setup stale github integration to automatically label and close stale pull requests

cjcolvar:
This PR sets up https://github.com/probot/stale.  It will label pull requests as `stale` that sit without activity for 30 days.  After 7 days of inactivity these are closed.

I've already enabled the github integration for hyrax but it will only run when this PR is merged to master.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,19 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,19 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
```
jcoyne: How about 14 instead? I can see someone being on vacation for a week and not responding in time.

cjcolvar: :+1: I debated about the same thing.

PR #3334, opened 2018-10-16
[Request review] Fixes #3245 - Add flash notice for Batch Deactivation of Expired Emba…

### LINKED ISSUE #3245: Add flash notice for Batch Deactivation of Expired Embargo
### Descriptive summary

Batch Deactivation of Expired Embargo does not show flash notice, but just refreshes to "All embargoes tab"

Propose to add it in: 
Message could be something like "Deactivation succeeded" 

### Rationale

Provide the rationale or user story that describes "why" this issue should be addressed. Especially if this is a new feature or significant change to the existing implementation.

1. Log in as admin
2. go to Manage Embargoes
3. Click Expired tab
4. select 2 works
5. Click Deactivate Embargoes for selected
6. Redirects to All embargo tab (the embargo is actually deactivating fine, this ticket just addressed the lack of flash notice)

Comment from @laritakr Yes, there is no flash specified for batch leases or embargoes. What should the message say?

[Need to add a flash notice here and in corresponding leases controller.](https://github.com/samvera/hyrax/blob/master/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb#L38-L39)

hweng:
…rgo.

Fixes #3245 ; refs #3245

The "Deactivation succeeded" flash notice is added for Batch Deactivation:
screen shot 2018-10-15 at 12 42 58 pm

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
@@ -36,7 +36,7 @@ module Hyrax
           curation_concern.copy_visibility_to_files
         end
       end
-      redirect_to embargoes_path
+      redirect_to embargoes_path, notice: t('.embargo_deactivated')
     end
 
     # This allows us to use the unauthorized template in curation_concerns/base
--- a/config/locales/hyrax.de.yml
+++ b/config/locales/hyrax.de.yml
@@ -826,6 +826,8 @@ de:
         type: Art von Arbeit
         viz_after: Sichtbarkeit wird geändert
         viz_current: Aktuelle Sichtbarkeit
+      update:
+        embargo_deactivated: Deaktivierung erfolgreich!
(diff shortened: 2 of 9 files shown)


app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
```diff
@@ -36,7 +36,7 @@ def update
           curation_concern.copy_visibility_to_files
         end
       end
-      redirect_to embargoes_path
+      redirect_to embargoes_path, notice: 'Deactivation succeeded!'
```
no-reply: Could we make this text `i18n`ized?

hweng:
@no-reply The i18nized flash message was added. Please see updated code. Thank you for your reviewing!

hweng:
@samvera/hyrax-code-reviewers @no-reply The requested change was applied. Can anyone review it? Thanks!

PR #3335, opened 2018-10-18
A whole sentence ("To create a separate work for each of the files, g…

jdpro:
…o to...") was not translated.

It was impossible for us  to translate in french the sentence "To create a separate work for each of the files, go to batch upload" : it is written in plain english in an erb file, and Rails i18n mechanism is not called.
Translations other than french and english were automatically  generated through Google translate. They're probably not suitable.

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.de.yml
+++ b/config/locales/hyrax.de.yml
@@ -348,6 +348,9 @@ de:
     base:
       citations:
         header: 'Zitate:'
+      form:
+        batch_upload: batch upload
+        create_separate_work_for_each_file: Um eine separate Arbeit für jede Datei zu erstellen, gehen Sie zu
       form_child_work_relationships:
         actions:
           remove: Diese Arbeit entfernen
--- a/app/views/hyrax/base/_form.html.erb
+++ b/app/views/hyrax/base/_form.html.erb
@@ -16,7 +16,7 @@
   <% end %>
   <% if Flipflop.batch_upload? && f.object.new_record? %>
     <% provide :metadata_tab do %>
-      <p class="switch-upload-type">To create a separate work for each of the files, go to <%= link_to "Batch upload", hyrax.new_batch_upload_path %></p>
+      <p class="switch-upload-type"><%= t(".create_separate_work_for_each_file") %> <%= link_to t(".batch_upload"), hyrax.new_batch_upload_path %></p>
     <% end %>
   <% end %>
   <%= render 'hyrax/base/guts4form', f: f %>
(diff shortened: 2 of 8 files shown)


config/locales/hyrax.de.yml
```diff
@@ -348,6 +348,9 @@ de:
     base:
       citations:
         header: 'Zitate:'
+      form:
+        batch_upload: batch upload
+        create_separate_work_for_each_file: Um eine separate Arbeit für jede Datei zu erstellen, gehen Sie zu
```
jmfrenzel: For German I would like to make it even more specific by adding: 
```suggestion
        create_separate_work_for_each_file: Um eine separate Arbeit für jede einzelne Datei zu erstellen, gehen Sie zu
```

jdpro: Our german translation was automatically generated

jmfrenzel: This is the common workflow in hyrax https://github.com/samvera/hyrax/wiki/Translations-(internationalization).
Doing a rather good job, Google and Co. fail miserably for very specific context or short phrases/ single words. At least the German part. How is id performing with French?

jmfrenzel: I would appreciate if you include my update in your PR instead of me adding a follow-up PR correcting the German.

jdpro: French translations are very often unusable or even comic. For exemple, "date" is translated by "rendez-vous amoureux" (it means, an appointment with a girl or a boy for whom you have feelings...). It goes beyond a translation problem, the expressions used are not necessarily adapted (for example, we can't us "work" at all). We sometimes need to rewrite the English version! And that prevents us from integrating our French translation propositions into Hyrax.

no-reply:
Thanks! I really appreciate you taking the time to run the automatic translations (most people haven't been able to get the API keys needed to do this).

> We sometimes need to rewrite the English version! And that prevents us from integrating our French translation propositions into Hyrax.

I wonder if we can come up with a resolution for this. Maybe adopting a consistent style and vocabulary for French--even if it's a major departure from the english text--is the way to go? 

@jmfrenzel's consistent attention has really benefited the German translations, and we'd love to coordinate similar efforts for other translations.

vantuyls:
@jdpro can you let me know what your CLA status is (both cCLA and iCLA)? we can't merge this until we have that resolved. Happy to chat over on slack if thats easier.

jdpro:
Thanks @vantuyls  I've just sent an ICLA to the requested adress in order to cover pull requests from this Github account.

PR #3337, opened 2018-10-18
Add dutch translation

geekscruff:
Add i18n translation for dutch (nl) using Google translate API.

### DIFF
--- a/config/locales/hyrax.nl.yml
+++ b/config/locales/hyrax.nl.yml
@@ -1,0 +1,1240 @@
+---
+nl:
+  activefedora:
+    models:
+      batch_upload_item: Werkt door Batch
+      file_set: het dossier
+  activemodel:
+    errors:
+      messages:
+        cannot_have_child_nested: kan er geen kind in laten nestelen
+        cannot_nest_in_parent: kan niet nestelen binnen de ouder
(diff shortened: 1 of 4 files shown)


geekscruff:
@vantuyls that would be nice, I was asked for it in our newest repo, so I'll ask if they can find a volunteer

no-reply:
@geekscruff is looking into human-supported translation in the coming days.

no-reply:
@vantuyls @geekscruff do we want to move forward with this?

vantuyls:
i'd defer to @geekscruff finding someone. are we firm on the policy that we should have human eyes on all translations? if yes, i don't want to merge this until we have that.

no-reply:
>  are we firm on the policy that we should have human eyes on all translations?

I don't think this is a standing policy, and it isn't true of existing translations.

On the other hand, it would be much harder to justify removing a translation once its in place than to simply hold off on support pending some language-capable champion.

On the other-other hand, the cost of shipping and maintaining additional translations seems very low.

no-reply:
@vantuyls do we want to include this in the 3.0.0 beta, or should we adopt an official policy about new translations?

geekscruff:
Held until @geekscruff gets feedback. No need to review every week :)

PR #3338, opened 2018-10-18
Allow readonly works to be added to collections

### LINKED ISSUE #3323: Cannot add a read only work to a collection
### Descriptive summary

Users should be able to include works they can read in a collection to which they have deposit access.

### Rationale

Primary Use Case:  Building an exhibit collection where the exhibit curator is not the content creator.

### Expected behavior

While viewing a read only work, if the user has deposit access to at least one collection, the Add to Collection button should be visible and should provide a means for a read only work to be added to a collection.
 
### Actual behavior

While viewing a read only work, the Add to Collection button is never displayed.  Edit access to the work is required to see this button.

As user A...
1. create a public work named work-1

As user B...
2. create a collection named col-1
3. navigate to work-1

Should see Add to Collection button.

4. click Add to Collection
5. search for and select col-1
6. click Save Changes

elrayle:
Fixes #3323

Allow logged in users to add read-only works to collections.  This supports the exhibit building use case as described in issue #3323.

### Public API
* 0 public methods changed
* 1 public method added to work_show_presenter

### UI Behavior
impacts work show page only
* behavior for editors of the work is unchanged
* behavior for viewers of the work - now will show the Add to Collection button if the user can deposit to at least one collection or can create a collection

## Guidance for testing, such as acceptance criteria or new user interface behaviors:
* As editor of the work, see Add to Collection button.
* As a logged in viewer of the work, see Add to Collection button IFF the viewer can deposit in at least one collection OR can create a new collection
* As a non-logged in user viewing the work, do NOT see the Add to Collection button

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -133,7 +133,7 @@ module Hyrax
     private
 
       def user_collections
-        collections_service.search_results(:deposit) if presenter.editor?
+        collections_service.search_results(:deposit)
       end
 
       def collections_service
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -204,6 +204,13 @@ module Hyrax
       metadata
     end
 
+    # determine if the user can add this work to a collection
+    # @param collections <Collections> list of collections to which this user can deposit
+    # @return true if the user can deposit to at least one collection OR if the user can create a collection; otherwise, false
+    def show_deposit_for?(collections:)
+      collections.present? || current_ability.can?(:create_any, Collection)
+    end
+
     private
 
       # list of item ids to display is based on ordered_ids
--- a/spec/features/work_show_spec.rb
+++ b/spec/features/work_show_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe "display a work as its owner" do
+RSpec.describe "work show view" do
   include Selectors::Dashboard
 
   let(:work_path) { "/concern/generic_works/#{work.id}" }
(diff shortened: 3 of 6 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -133,7 +133,7 @@ def manifest
     private
 
       def user_collections
-        collections_service.search_results(:deposit) if presenter.editor?
+        collections_service.search_results(:deposit)
```
elrayle: The restriction to `if presenter.editor?` prevents viewers of the work to have access to their collections.

spec/features/work_show_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe "display a work as its owner" do
+RSpec.describe "work show view" do
   include Selectors::Dashboard
```
elrayle: Made a few modifications to test labels in this file because the output of the combined labels did not make sense.

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -204,6 +204,13 @@ def manifest_metadata
       metadata
     end
 
+    # determine if the user can add this work to a collection
+    # @return true if the user can deposit to at least one collection OR if the user can create a collection; otherwise, false
+    def can_add_to_collection?(depositable_collections)
```
no-reply: I'm slightly uncomfortable with the `Presenter` leaking this permission information into the view, and the view conditional on https://github.com/samvera/hyrax/pull/3338/files#diff-600b68f4bb5cd3833e5a8be24f8b356eR25 that results.

Could we explore refactors that avoid this--hopefully by making the new presenter method return some presentable data?

One (maybe too awkward?) possibility that comes to mind is to change the new method to change the new method to:

```ruby
  def depositable_work_ids(collections)
    return [] unless collections.present? || current_ability.can?(:create_any, Collection)
    [id]
  end
```

Another option might be to just rename the existing method to something presentation oriented:

```ruby
  def show_deposit_for(collections:)
    collections.present? || current_ability.can?(:create_any, Collection)
  end
```

The second option doesn't squash the need for the view conditional, but does at least reframe so the presenter is focused on responding to messages in its own domain.

Thoughts?

elrayle:
[WIP] The Travis failure (and Circle CI) occurred after the end of the testing process where all tests passed.  It is unclear why there is a failure at that point.

Debugging:
* 1st attempt - remove all changes so this branch matches master to see if it passes

elrayle:
@no-reply Thanks for the quick review.  Requested changes have been applied.  All tests passing.

elrayle:
@vantuyls @julesies Do either of you want to comment on this PR to indicate acceptance of the proposed UI changes?  I'm hoping for this to get the final code review tomorrow after Tech call.

vantuyls:
this looks like the changes we'd be hoping for. thanks.

PR #3341, opened 2018-10-21
[WIP] Fix Universal Viewer embed code

geekscruff:
UV data-uri needs full URI otherwise the embed code won't work on other sites

Also, external sites will make a standard get request to the manifest, so we need to return the json manifest for both a json and html request


geekscruff:
@jcoyne good spot, there is a separate PR for the NL translation, which I've backed out of here.

geekscruff:
The travis builds got all in a tangle when there were problems the other day, and now I can't seem to get them to run. Circle CI passed, is that enough?!

cjcolvar:
I restarted the failed PR builds and now they don't have any test failures but are still exiting with an error.  I'm going to go ahead and approve and merge this PR.

PR #3342, opened 2018-10-22
Remove unnecessary case clause from NestCollectionsController#redirec…

jcoyne:
…t_path


afred:
@no-reply confirms that Circle Ci build is broken for unrelated reasons. During PR review session, we approved and agreed it's ready to merge.

PR #3346, opened 2018-10-22
Use polymorphic_url instead of _path for UV data-uri

geekscruff:
UV data-uri needs full URI otherwise the embed code won't work on other sites

replaces #3341 (problems with travis build)


geekscruff:
Looks like #3341 was caused by gh problems.

PR #3352, opened 2018-10-29
Use `i18n` keys for hard coded string values in Collection Type

no-reply:
These hard coded string values already have corresponding `i18n` keys. Using them improves internationalization of the UI.

@samvera/hyrax-code-reviewers


chrisdaaz:
@no-reply i don't think this is working as expected. from nurax this morning:

image

PR #3360, opened 2018-10-30
IIIF viewer choice

cjcolvar:
Pulls in work from https://github.com/samvera-labs/hyrax-iiif_av to generalize the IIIF viewer support to any IIIF viewer and not assuming Universal Viewer.  In the future other viewers could be added to core Hyrax but this PR at least allows a downstream app the ability to customize the IIIF view partial without overriding.  This is helpful for more experimental viewers like the Avalon react player for IIIF Presentation 3.0 AV manifests.  Possible documentation for how to use a custom IIIF viewer is below.

This PR has some view partial changes so I'd hope that it would be considered for Hyrax 3.0.

@samvera/hyrax-code-reviewers

# Adding a custom IIIF viewer

Create a new file in your Hyrax app at `app/views/hyrax/base/iiif_viewers/_my_iiif_viewer.html.erb`:
```
<h3>My IIIF Viewer!</h3>
<a href=<%= main_app.polymorphic_path([main_app, :manifest, presenter], { locale: nil }) %>>Manifest</a>
```

Override `#iiif_viewer` in your work presenter `app/presenters/hyrax/generic_work_presenter.rb`:
```
def iiif_viewer
  :my_iiif_viewer
end
```

### DIFF
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -64,6 +64,12 @@ module Hyrax
         members_include_viewable_image?
     end
 
+    # Override this method to declare a different iiif viewer for your work type
+    # @return [Symbol] the name of the IIIF viewer partial to render
+    def iiif_viewer
+      :universal_viewer
+    end
+
     # @return FileSetPresenter presenter for the representative FileSets
     def representative_presenter
       return nil if representative_id.blank?
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -67,8 +67,19 @@ module Hyrax
     alias universal_viewer? iiif_viewer?
     deprecation_deprecate universal_viewer?: "use iiif_viewer? instead"
 
-    # Override this method to declare a different iiif viewer for your work type
     # @return [Symbol] the name of the IIIF viewer partial to render
+    # @example A work presenter with a custom iiif viewer
+    #   module Hyrax
+    #     class GenericWorkPresenter < Hyrax::WorkShowPresenter
+    #       def iiif_viewer
+    #         :my_iiif_viewer
+    #       end
+    #     end
+    #   end
+    #
+    #   Custom iiif viewer partial at app/views/hyrax/base/iiif_viewers/_my_iiif_viewer.html.erb
+    #   <h3>My IIIF Viewer!</h3>
+    #   <a href=<%= main_app.polymorphic_url([main_app, :manifest, presenter], { locale: nil }) %>>Manifest</a>
     def iiif_viewer
       :universal_viewer
     end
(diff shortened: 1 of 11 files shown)


app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -55,15 +55,24 @@ def download_url
...
+    # @return [Boolean] render a IIIF viewer
+    def iiif_viewer?
       representative_id.present? &&
         representative_presenter.present? &&
         representative_presenter.image? &&
         Hyrax.config.iiif_image_server? &&
         members_include_viewable_image?
     end
 
+    alias universal_viewer? iiif_viewer?
```
no-reply: :+1: nice work!

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -55,15 +55,24 @@ def download_url
...
+    deprecation_deprecate universal_viewer?: "use iiif_viewer? instead"
+
+    # Override this method to declare a different iiif viewer for your work type
```
no-reply: Could we just document that this method returns the viewer type as a symbol? I fear that saying "override" encourages folks to reopen this class, while inheriting it or implementing its interface would be better.

no-reply:
This seems nice and clean, and I'm not sure it requires a 3.0.0. What do you think about releasing this in the 2.x series as well?

cjcolvar:
The failing test is coming from my use of the same changes as https://github.com/samvera/hyrax/pull/3341 without changing the tests.  I'm inclined to merge that PR then rebase this one to pick up the fix.

app/presenters/hyrax/work_show_presenter.rb
cjcolvar: WorkShowPresenter is inherited by the work presenter generated as part of a work type.  I was expecting overrides to happen at that level in the downstream app's work presenter.  What do you think would be clearer language to suggest that pattern?  Do you think the `@example` you suggest would serve this purpose?

cjcolvar:
@no-reply I was hesitant to ask for this in 2.0 since there are changes to views even if the behavior stays the same.  But maybe it is backwards-compatible in that if you have overridden one of the partials I changed then nothing should happen as a result of upgrading to this code.

kdid:
👍 This looks good!

app/presenters/hyrax/work_show_presenter.rb
no-reply: I know lots of folks reach straight for monkey-patching, and I'd like to move docs to language that doesn't invite that interpretation.

no-reply:
> But maybe it is backwards-compatible in that if you have overridden one of the partials I changed then nothing should happen as a result of upgrading to this code.

That's my interpretation. These changes seem carefully made, and can't see a way they could reasonably break expectations of a downstream app.

PR #3362, opened 2018-10-31
Update login and signup forms UI, including Hyrax logo in top nav bar

### LINKED ISSUE #3361: Homepage cleanup
### Descriptive summary

Clean up the Login form on the homepage so it's using Bootstrap form groups properly, and possibly centered.  Also the "Search Hyrax" label is white, and should be darker so it shows up in the top nav bar.

### Rationale

Better UI/UX.

homepage-cleanup

adamjarling:
Fixes #3361 
This updates the following:
- Log in and Sign up forms now have bootstrap styling and are improved UI.
- Use the Hyrax logo in the masthead, per Bootstrap convention.
- Fix the color of Search label in main search bar for different backgrounds depending where user is at in the app.
- Consolidate the footer Copyright text to remove redundant "Copyright" text.  Less text the better...

homepage-updates-after

homepage-updates-login-after

homepage-updates-before

@samvera/hyrax-code-reviewers


jcoyne:
I'm against adding a logo to the repository, as it's something that most institutions are going to want to provide for themselves and it increases the size of the repository.  Can you host it elsewhere and just reference it?

Similarly, I'd prefer we don't include `stylesheets/hyrax/_login_signup.scss` as institutions that don't use devise (e.g. some sort of SSO) won't want that.

adamjarling:
@jcoyne Ok, so for the logo, where else could or should it be hosted?   Just thinking if that server goes down, or people are working on the app offline, it's a broken image vs the 15KB weight.  I don't even know if it looks that great, just trying to do something to break this out of generic Bootstrap app look & feel.... but maybe that's not a concern?

Also, for the Sign In / Log In cleanup, I was thinking non-Devise users would not to be affected.  For the life of me I can't figure out which HTML partial in Hyrax contains the Log In / Sign Up include code.  Would a compromise be if someone could help me locate that file, we wrap the Devise code in a specific class selector, something like `.external-auth-login-wrapper` or whatever and target that way?  It's the generic selector usage you're concerned with?

jcoyne:
The devise code is in the devise gem. As for the image weight, if it’s this image, I wouldn’t worry about it being broken, if you care about controlling its availability you can move it into the hyrax app you generate.

adamjarling:
> The devise code is in the devise gem. 

Would a better approach be to bring over the Devise partials?   I tried this first but it didn't work...the Devise partial's path is `app/views/devise/sessions/new.html.erb`, which I created in Hyrax, re-started the Rails server, but it didn't serve.   Looks like from the Devise docs you add a `config/initializers/devise.rb` file and include a line which seems to allow overrides (`config.scoped_views = true`), which I did, re-started the server, and the entire app chokes when that file is present.  

Any idea what I'm missing?   

After trying this route, that's when I created the `stylesheets/hyrax/_login_signup.scss` file to just extend standard Bootstrap form classes onto those elements.

Back to the point of this ticket, I'm thinking from the perspective of say a potential adopter of Hyrax, they go to Nurax to get an impression of the UI, and the Log In / Sign Up default views, the first thing they see, looks like it's neglected or unfinished.   If it's no big deal to anyone else, I can just close this ticket...that's fine too.

jcoyne:
@adamjarling I think the problem here is that you are trying to fix this in Hyrax, but the fix belongs in an application that uses Hyrax, such as Hyku or Nurax.   Hyrax should just be a template, not a full solution, especially for something like authentication, which many institutions handle differently.  This is why we don't require devise, we just use it for the default install.  Adopters are encouraged to find another solution for auth, because devise is not suitable for all cases.

mbklein:
IMO, if Hyrax is going to provide any kind of default login form out of the box, it should look good. So if this PR isn't worth merging, is the better course of action to remove the login form altogether?

jcoyne:
@mbklein Hyrax doesn't provide a login form, devise does.

app/views/_logo.html.erb
```diff
@@ -1,4 +1,3 @@
 <a id="logo" class="navbar-brand" href="<%= hyrax.root_path %>" data-no-turbolink="true">
-  <span class="glyphicon glyphicon-globe" role="img" aria-label="<%= application_name %>" aria-hidden="true"></span>
-  <span class="institution_name"><%= application_name %></span>
+  <%= image_tag("hyrax_logo-bw.png", alt: application_name) %>
```
jcoyne: perhaps this could be done as something configurable such as https://github.com/samvera/hyrax/blob/5a9d1be16ee1a9150646384471992b03aab527a5/app/views/layouts/homepage.html.erb#L3

adamjarling: I'm going to pull the logo out of this PR.

vantuyls:
on the logo: i think it is a useful placeholder. is the primary issue here one of the size of the image?

vantuyls:
i'm a little confused as tot he status of this work and conversation. @adamjarling thoughts?

adamjarling:
@vantuyls The way I understand it, Devise is a gem which provides the markup for the Log In / Sign Up forms.   This markup is not part of the Hyrax codebase.  My changes are CSS additions (in the Hyrax codebase), which target Devise's Log In / Sign Up elements.  My thought was if this is going to be visible on Nurax, then it should be cleaned up.

The suggestion was this styling code should be added elsewhere, `.internal_test_app`, instead of Hyrax? 

So maybe the path forward is:
1. We do nothing and just let it look like it looks.
2. We add the styles to `.internal_test_app`?   Then say someone or a new institution is evaluating this application on Nurax, one of the first thing they see (Sign Up), looks less neglected.

jcoyne:
Some other alternatives:

1. Make this change in Nurax
1. Make a "bootstrap-devise" gem that provides a boostrap styled login for for devise and have hyrax generate that as a dependency into the host app (so it can be removed if necessary)

no-reply:
I don't see any value in making this change in `nurax`, since it's not intended for actual use.

@jcoyne said:

> @mbklein Hyrax doesn't provide a login form, devise does.

Since the `hyrax` install generators install `devise`, and the user code is deeply intertwined with it, I'm not sure what value this distinction has.

and:

> Make a "bootstrap-devise" gem that provides a boostrap styled login for for devise and have hyrax generate that as a dependency into the host app (so it can be removed if necessary)

This seems like a lot of overhead to avoid putting some `css`/`scss` defaults in `hyrax`. I gather that you're arguing the changes here would make customization harder, but I'm not seeing it. Can you say more about what challenges you expect applications to face if we merge this?

no-reply:
> I'm just concerned that adding these styles to the gem make them "permanent", that is, hard to remove from your local install. 

Does it make them "permanent", or just `hyrax` managed/native? They can be overridden locally (right?), and we provide default styles in other places, even for engine provided views (e.g. blacklight catalog).

> How about moving these styles to the generator so that one could just rm the files if they aren't needed?

The trade-off seems to be that any generated code needs to be manually updated to keep up with changes in the engine. The fragmentation of default styles between the engine and application seems like a risk to me.

We don't seem to generate `.css` or `.scss` files into applications in other cases. Is the thinking that this should be an exception because of the proximity to `devise`?

jcoyne:
@no-reply We do generate https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/templates/hyrax.scss and this is intended to make it possible to keep things we don't want out of the compiled css. And, yes, i'm largely for keeping this out of the default hyrax bundle because it's so tightly coupled to `Devise#database_authenticatable`

no-reply:
@jcoyne what if we moved the devise-specific styles into their own sheets, e.g.: `@import`

no-reply:
This needs to be rebased to pass CI, in any case, so I'm happy to make any changes needed if we can decide on a path forward.

jcoyne:
I would feel better if `hyrax/login_signup` was imported by `hyrax.scss` (in the application) rather than by `hyrax/hyrax.scss` (in the engine)

no-reply:
@jcoyne i pushed a new commit making that change, and rebased the rest. if it looks good, could you approve? I'll hold merge for your consideration.

Thanks for sticking this one out; i think the approach we landed on is a big improvement.

PR #3367, opened 2018-11-01
[WIP] Update to work edit buttons UI layout

### LINKED ISSUE #3364: Edit Work buttons layout re-factor
### Descriptive summary

The layout of main buttons for Edit Work page is confusing, and could be better organized.  Investigate alternative layout options.

work-buttons-edit-page

adamjarling:
Fixes #3364 

This PR cleans up the Edit Work buttons layout.  See before/after screenshots.

edit-work-buttons-mobile-new
edit-work-buttons-mobile-old
work-buttons-edit-page-option1

work-buttons-edit-page

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_work-show.scss
+++ b/app/assets/stylesheets/hyrax/_work-show.scss
@@ -37,8 +37,17 @@ ul.tabular {
 }
 
 .show-actions {
-  padding: $panel-body-padding;
-  // text-align: center;
+  padding-right: $panel-body-padding;
+
+  .row {
+    &:first-child {
+      margin-bottom: 8px;
+    }
+  }
+
+  @media screen and (max-width: 640px) {
+    padding-top: $panel-body-padding;
+  }
 }
 
 .work-show {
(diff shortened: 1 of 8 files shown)


jcoyne:
I'm a bit confused about why we'd want 5 "primary" buttons.  Can all of them be the primary action for the form?

adamjarling:
@jcoyne Yea, that makes sense.  Guess I was mis-using `primary` for display instead of function.  I'll update it.  Given that, what's your thoughts on:

- What's the primary button in these five?
- Keeping a consistent pattern in the application, if there's only one button in a form, should it be primary or default?

jcoyne:
I don't think there is a primary, that's why they were all default before.  I suspect on a one button form, that button would be a primary button.

jcoyne:
After taking a look at this, I'd probably put them all under a drop-down menu, but I'm not sure of the accessibility concerns with that.

adamjarling:
This ticket has moved more into a page layout re-factor in addition to just the buttons, so am going to close this ticket to avoid confusion.

PR #3368, opened 2018-11-02
Fixed punctuation error in helptext

simholt:
Removed excessive periods and quotation marks.

@samvera/hyrax-repo-managers 
@samvera/hyrax-code-reviewers


jcoyne:
Oh, I see they are required due to the colin

PR #3370, opened 2018-11-02
`i18n`-ize "Collection Details" string

no-reply:
This hard coded string should be translatable.

@samvera/hyrax-code-reviewers


cjcolvar:
Merged even though circleci build failing due to #3373.

PR #3374, opened 2018-11-05
Re-factor the layout (and buttons, original motivation) for the work show page

### LINKED ISSUE #3364: Edit Work buttons layout re-factor

adamjarling:
Fixes #3364 

See #3364 for more details, but essentially this PR updates the layout of a Work Show page.

#### Before
generic-work-show-page-before

#### After
generic-work-show-page-idea-v2

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_work-show.scss
+++ b/app/assets/stylesheets/hyrax/_work-show.scss
@@ -37,8 +37,17 @@ ul.tabular {
 }
 
 .show-actions {
-  padding: $panel-body-padding;
-  // text-align: center;
+  padding-right: $panel-body-padding;
+
+  .row {
+    &:first-child {
+      margin-bottom: 8px;
+    }
+  }
+
+  @media screen and (max-width: 640px) {
+    padding-top: $panel-body-padding;
+  }
 }
 
 .work-show {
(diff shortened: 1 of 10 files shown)


no-reply:
How badly do you think this orphans content if existing adopters have copied any of these views in wholesale? It looks like it might not be too bad at a glance(?).

I'm trying to gauge whether we should roll this into a 2.5.0 (this rearrangement seems like more than a bugfix), or hold for merge along with the existing 3.0.0 items later this week.

chrisdaaz:
i recommend rolling out adam's UI enhancements along with 3.0.0

adamjarling:
@no-reply Hmm, I think it depends on how much each customization has strayed from the original view.  There are some changes that could break a custom layout, but I don't think it'd be that hard to fix?  It's hard to say in general.

I'm working on the Collections View page at the moment (on request from @julesies in the Repo managers Slack channel), and that will follow the same patterns as this page.

So if someone has customized their Hyrax views, this change is starting to consolidate styles, layouts and hopefully will eventually help making customizing easier.

PR #2367, opened 2017-12-13
Update featured works and recently uploaded display

aaron-collier:
refs https://github.com/samvera-labs/hyku/issues/457

This is the initial update to the styling of the featured works list and recently uploaded list based on the homepage mockups in the above hyku issue.

The outcomes of this PR are:
- [x] Consistent styling between the featured and recent list
- [x] Updated layout to better present the information
- [x] Changes the tab panels to pill panels (in preparation for additional updates) for a more consistent look with the formatted tables of information

*DISPLAY BEFORE*
screen shot 2017-12-13 at 10 01 48 am

screen shot 2017-12-13 at 10 01 55 am

*DISPLAY AFTER*

screen shot 2017-12-13 at 9 48 25 am

screen shot 2017-12-13 at 9 48 39 am

@samvera/hyrax-code-reviewers

PR #3375, opened 2018-11-05
Add a little space to landing page layout

### LINKED ISSUE #3369: Try a new idea for the homepage layout
### Descriptive summary

Try something and do a quick user-test via repo managers, of an alternative layout the homepage, using standard Bootstrap components.

### Rationale

It's the first thing users see, and should be somewhat engaging.

adamjarling:
…omponent

Fixes #3369 

Provide an updated layout to the homepage, which opens it up a bit and gives a little more breathing room to the intro page, and also hopefully sends the message that this homepage can and should be customized by a Hyrax user.

#### Before
hyrax-homepage-current-layout

#### After
hyrax-homepage-alternate-layout-idea

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_home-page.scss
+++ b/app/assets/stylesheets/hyrax/_home-page.scss
@@ -69,7 +69,14 @@
 
 .home-content {
   .nav {
-    margin-bottom: 20px;
+    margin-bottom: $padding-large-vertical * 2;
+  }
+
+  // Give some separation between the Works and Collections columns, which stack up vertically on mobile
+  @media screen and (max-width: 640px) {
+    > div {
+      margin-bottom: $padding-large-vertical * 3;
+    }
   }
 }
 
@@ -84,12 +91,12 @@
   }
 }
 
-.home_share_work {
-  margin-bottom: $padding-large-vertical * 2.5;
+.home-share-work {
+  margin-bottom: $padding-large-vertical * 4;
+  margin-top: $padding-large-vertical * 2;
 
-  p {
-    margin: 0.25em auto;
-    width: 40%;
+  .terms-of-use {
+    margin-top: $padding-large-vertical;
   }
 }
 
--- a/app/views/hyrax/homepage/_home_content.html.erb
+++ b/app/views/hyrax/homepage/_home_content.html.erb
@@ -1,5 +1,5 @@
-<div class="col-xs-12 col-sm-6">
-  <ul id="homeTabs" class="nav nav-pills">
+<div class="col-sm-6">
+  <ul id="homeTabs" class="nav nav-tabs">
     <li class="active"><a href="#featured_container" data-toggle="tab" role="tab" id="featureTab"><%= t('hyrax.homepage.featured_works.tab_label') %></a></li>
     <li><a href="#recently_uploaded" data-toggle="tab" role="tab" id="recentTab"><%= t('hyrax.homepage.recently_uploaded.tab_label') %></a></li>
   </ul>
@@ -13,9 +13,9 @@
   </div>
 </div><!-- /.col-xs-6 -->
 
-<div class="col-xs-12 col-sm-6">
+<div class="col-sm-6">
 
-  <ul class="nav nav-pills" role="tablist">
+  <ul class="nav nav-tabs" role="tablist">
     <li class="active"><a aria-expanded="true" href="#tab-col2-first" role="tab" data-toggle="tab"><%= t('hyrax.homepage.admin_sets.title') %></a></li>
     <li class=""><a aria-expanded="false" href="#tab-col2-second" role="tab" data-toggle="tab"><%= t('hyrax.homepage.featured_researcher.tab_label') %></a></li>
   </ul>
--- a/app/views/hyrax/homepage/index.html.erb
+++ b/app/views/hyrax/homepage/index.html.erb
@@ -1,30 +1,28 @@
 <% provide :page_title, application_name %>
 
 <% if @presenter.display_share_button? %>
-  <div class="home_share_work row">
-    <div class="col-sm-12 text-center">
-      <% if signed_in? %>
-        <% if @presenter.create_many_work_types? %>
-          <%= link_to '#',
-            class: "btn btn-primary btn-lg",
-            data: { behavior: 'select-work', target: '#worktypes-to-create', 'create-type' => 'single' } do %>
-            <i class="glyphicon glyphicon-upload" aria-hidden="true"></i> <%= t('hyrax.share_button') %>
-          <% end %>
-        <% else # simple link to the first work type %>
-          <%= link_to new_polymorphic_path([main_app, @presenter.first_work_type]),
-                class: 'btn btn-primary' do %>
-            <i class="glyphicon glyphicon-upload" aria-hidden="true"></i> <%= t('hyrax.share_button') %>
-          <% end %>
-        <% end %>
-      <% else %>
-        <%= link_to hyrax.my_works_path,
-          class: "btn btn-primary btn-lg" do %>
-          <i class="glyphicon glyphicon-upload" aria-hidden="true"></i> <%= t('hyrax.share_button') %>
-        <% end %>
+<div class="home-share-work text-center">
+  <% if signed_in? %>
+    <% if @presenter.create_many_work_types? %>
+      <%= link_to '#',
+        class: "btn btn-primary btn-lg",
+        data: { behavior: 'select-work', target: '#worktypes-to-create', 'create-type' => 'single' } do %>
+        <i class="glyphicon glyphicon-upload" aria-hidden="true"></i> <%= t('hyrax.share_button') %>
       <% end %>
-      <p><%= link_to t(:'hyrax.pages.tabs.terms_page'), hyrax.terms_path %></p>
-    </div>
-  </div>
+    <% else # simple link to the first work type %>
+      <%= link_to new_polymorphic_path([main_app, @presenter.first_work_type]),
+            class: 'btn btn-primary btn-lg' do %>
+        <i class="glyphicon glyphicon-upload" aria-hidden="true"></i> <%= t('hyrax.share_button') %>
+      <% end %>
+    <% end %>
+  <% else %>
+    <%= link_to hyrax.my_works_path,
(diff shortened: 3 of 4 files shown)


jcoyne:
I think having the banner and the jumbotron on the front page makes it look too crowded.  @ggeisler I'd love to hear your opinion.

ggeisler:
I'll qualify my response to @jcoyne's by pointing out that I haven't really been following the development of Hyrax and so am not sure I understand all the current goals and use cases, but yes, I'd agree with @jcoyne about the jumbotron. I understand @adamjarling's goal but am not sure having a large jumbotron on the homepage is good for users on an on-going basis. It might be nice for the first-time or very occasional user to see "some type of welcome message" as the text suggests, but for regular users it takes up a lot of space for something they've seen and don't care about any more. That space would be better used by actual content that might help them do what they came to the site to do.

So I'd worry that the jumbotron is actually encouraging implementors to do something that might not benefit most of their users.

Also, if there are good reasons for the jumbotron as a place to put "welcome" (or other, such as some sort of notice or general news) content, I'd think you'd want to keep the `Share your work` and `Terms of use` elements outside of it, since they are (I think) generally intended to be permanent content and don't really seem like they are part of the "welcome" content? It feels odd to me to mix those elements into the jumbotron.

ggeisler:
I'll add that I don't think the "gives a little more breathing room" part of @adamjarling's goal is a bad idea, but adding some space before and after the Share your work and Terms of use elements could address that.

I'd also say that in my mind this page is not "an intro page" but the homepage of an application that many users will use on an on-going basis. Jumbotron-like content makes more sense on a true intro page for a site where users see it once or occasionally, but not so much for a regularly used application like Hyrax or a library catalog, etc.

jcoyne:
I'd also be concerned on how it fits on a site that has lots of content such as Nurax:

screen shot 2018-11-06 at 12 11 25 pm

adamjarling:
@jcoyne @ggeisler Got it, and thanks for the feedback.  Guessing I'm just kicking the tires a bit and still figuring out what Hyrax is.  Here's an example of how we customize the Hyrax 'home page', 'landing page', whatever that view is, for one of our Hyrax applications.   I think this biased my idea of trying to show something more on first impression, for others who do in fact customize this view, but I get the fact it's a screen people will see over and over again and don't need to be relentlessly welcomed.  I'm not a big fan of the "jumbotron" either, but seemed like the only component Bootstrap along these lines to communicate the idea.

Probably an unnecessary ticket to begin with, but I do appreciate the eyes on it and feedback.  I'll close it up.

arch-hyrax-homepage

chrisdaaz:
i think the changes are an improvement over the current default homepage. there's also [some support on slack](https://samvera.slack.com/archives/C5MPUVBB5/p1541185078027000) in favor of this direction.

ggeisler:
@adamjarling That's a nice example so I totally get your motivation. I think the current homepage design is intended to provide implementors easy access to features they _might_ want on the home page, rather than trying to provide a particular visual style or layout.

The Northwestern example is certainly much more visually inviting than the default Hyrax homepage. But many institutions have created much more visually inviting Blacklight-based library catalogs than the default Blacklight application, without explicit encouragement in the default BL application for customization.

So maybe the question is really how much we actively try to encourage implementors to customize and provide a suggested direction (e.g., the jumbotron) versus just letting them customize if they want (as in the Northwestern example).

vantuyls:
ok. sorry to be late to this conversation. i'm not sure the jumbotron is the best choice, but, like @ggeisler indicated, giving some breathing room is not a bad idea. @adamjarling any chance you could mock up a non-jumbotron more breathing room version?

as for general considerations. Yes, hyrax is meant as a starting point that will be customized. I think, though, that offering some idea of the types of easy customizations one could make is helpful. For instance, we already provide a place where people can upload a banner image, change colors, etc., so it's not outside the bounds of the general direction hyrax has gone recently to provide some more simple customization options.

jcoyne:
@adamjarling the tabs look *way* better 👏 .  They changed from tabs to pills here: https://github.com/samvera/hyrax/commit/3151117b1b235175972aef4d4d7d663c250fc03b

But it's unclear why.  @aaron-collier do you have any recollection about why you made that change?

jcoyne:
Ah, there's some context provided on https://github.com/samvera/hyrax/pull/2367 and @ggeisler 's mockups here: https://github.com/samvera-labs/hyku/issues/457

ggeisler:
@adamjarling and all: I think the tabs look great (I think the pills looked okay in the original mockups because they were more distant from the Share your work button; in the implementation they are closer together (and maybe larger and bolder?) so the page feels visually heavier and more busy than the mockup version). 

So to be clear: just because the original mockups used pills doesn't mean the page needs to be implemented exactly that way. The mockups (and this includes any mockups I did for Samvera) are not intended to be pixel-perfect mockups that need to be followed exactly. As I hope is clear by the variety of mockups @jcoyne referenced in [samvera-labs/hyku#457](https://github.com/samvera-labs/hyku/issues/457), my goal was rather to present ideas for what sort of content might be useful on the homepage and how it might be generally arranged than specifying certain UI elements.

adamjarling:
@ggeisler Sounds like a good approach.  It may be beneficial to whoever shepherds Hyrax UI in the future, to know about and have access to the wireframes and mockups.  We could host them in the Samvera UX Interest Group space, or better yet they might serve as an asset to Hyrax's core documentation.  It'd keep people like myself from going rogue and trying to kick things which have already been thought out, but perhaps not quite implemented true to their source. 

I just tried looking through https://github.com/samvera-labs/hyku for any....is this where they would mostly live, or attached to individual, closed issues in the Hyrax repo?

ggeisler:
@adamjarling Yeah, I know it seems like there should be some central repository for all mockups and ideally there would be. I can definitely see how that would be helpful for developers coming along later and wanting some guidance on UI design.

In practice, it's a little less straightforward. Most of the mockups I did were in the context of the Hyku project, which mostly involved a small team at Stanford that was actively working on the project at the same time as I was producing mockups. So I'd do some mockups, run them by the project owner and maybe some of the developers, and then add to Github issues when we thought they were good to go. Sometimes when working on issues, developers would have questions or ideas that would prompt me to make adjustments, which sometimes I'd just do in code or via Web Inspector rather than going back to the mockup because it was quicker. So to a large extent there isn't a definitive set of mockups I can point to.

Also, most of the formal mockups I did were for the administration side of things, because that was what Hyku was adding to the existing Hyrax product (way oversimplified, but from a design standpoint mostly true). The homepage mockups were sort of the exception to that.

I also provided mockup ideas for the Collections extensions and Analytics working groups. For those I put mockups in single PDFs, which I assume are on a wiki page somewhere. I don't know offhand where they are but @elrayle led both those efforts and was very organized so I'm sure she knows where they are if those mockups are of interest.

I do agree that a more formal process for managing design documents would be great. There are newer tools and services like InVision that make that easier (I was just working on my own using OmniGraffle, which doesn't lend itself well to managing revisions and posting to the cloud, etc.). Maybe the Samvera community now has more UX-interested people that could help evaluate options and figure out a plan to manage design work better in the future, I'm not sure.

app/views/hyrax/homepage/_home_content.html.erb
```diff
@@ -1,5 +1,5 @@
-<div class="col-xs-12 col-sm-6">
-  <ul id="homeTabs" class="nav nav-pills">
+<div class="col-sm-6">
```
jcoyne: Was the change to not make it full width on mobile intentional?

app/views/hyrax/homepage/index.html.erb
```diff
@@ -1,30 +1,28 @@
...
-                class: 'btn btn-primary' do %>
-            <i class="glyphicon glyphicon-upload" aria-hidden="true"></i> <%= t('hyrax.share_button') %>
-          <% end %>
-        <% end %>
-      <% else %>
-        <%= link_to hyrax.my_works_path,
-          class: "btn btn-primary btn-lg" do %>
-          <i class="glyphicon glyphicon-upload" aria-hidden="true"></i> <%= t('hyrax.share_button') %>
-        <% end %>
+<div class="home-share-work text-center">
```
jcoyne: could a CSS class name change be considered a breaking change? @no-reply

app/assets/stylesheets/hyrax/_home-page.scss
```diff
@@ -84,12 +91,12 @@
...
-.home_share_work {
-  margin-bottom: $padding-large-vertical * 2.5;
+.home-share-work {
+  margin-bottom: $padding-large-vertical * 4;
+  margin-top: $padding-large-vertical * 2;
 
-  p {
-    margin: 0.25em auto;
-    width: 40%;
+  .terms-of-use {
```
jcoyne: 👍

app/views/hyrax/homepage/_home_content.html.erb
adamjarling: I think when using any `col-sm-*`, full-width is implied for mobile sizing.

app/views/hyrax/homepage/index.html.erb
adamjarling: I did this for 1.) CSS convention, and 2.) Hound complains about underscored classes too.

app/views/hyrax/homepage/_home_content.html.erb
jcoyne: 👌

app/views/hyrax/homepage/index.html.erb
no-reply: :+1: I think CSS class name changes are breaking changes in general, unless a compelling case can be made that they aren't.

If we can get this in today, I think it makes `3.0.0-beta1`

no-reply:
Thanks for all the great design conversation here. Did we land on a version of thins changeset that's ready to merge?

adamjarling:
@no-reply The latest version of this branch I think has been agreed upon?   It's current visual state and changes are:

1. Jumbotron idea is gone.
2. Some breathing room margin has been added to the primary Submit Your Work button area.
3. Nav pills have been changed to tabs.
4. Class name change (which is the breaking change... and if it's a problem I can change it back?)

PR #3383, opened 2018-11-06
Move email text to locale files

### LINKED ISSUE #3321: Move email text to locale files
### Descriptive summary
When an email is sent by hyrax, it appears that hard-coded text is used.  This means that the emails are always sent in English and are not easy to modify.  We would like to see the email text moved to locale files for translation and to allow local modifications.

We are interested in submitting a PR for this change and want to get some feedback before moving forward.

### Expected behavior
Email text should be moved to locale files as interpolated strings.

### Actual behavior
Email text is hard-coded.

### Related work
List of files with hard-coded message text:
* app/controllers/hyrax/depositors_controller.rb
* app/models/proxy_deposit_request.rb
* app/services/hyrax/import_url_failure_service.rb
* app/services/hyrax/batch_create_success_service.rb
* app/services/hyrax/batch_create_failure_service.rb
* app/services/hyrax/fixity_check_failure_service.rb
* app/services/hyrax/workflow/changes_required_notification.rb
* app/services/hyrax/workflow/deposited_notification.rb
* app/services/hyrax/workflow/pending_review_notification.rb

### Possible problems
I believe that the i18n language is chosen based on the current user's preferences.  If a user has chosen to view the application in Spanish, then the email messages should be sent in Spanish.  The recipients of the email may not read Spanish which could be a problem.  This may require setting an application-wide default language somewhere to force the emails to be sent in an appropriate language.

smithjp:
Fixes #3321 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/depositors_controller.rb
+++ b/app/controllers/hyrax/depositors_controller.rb
@@ -45,14 +45,14 @@ module Hyrax
       end
 
       def send_proxy_depositor_added_messages(grantor, grantee)
-        I18n.with_locale(:en) {
+        I18n.with_locale(:en) do
           message_to_grantee = I18n.t('hyrax.notifications.proxy_depositor_added.grantee_message', grantor: grantor.name)
           message_to_grantor = I18n.t('hyrax.notifications.proxy_depositor_added.grantor_message', grantee: grantee.name)
           Hyrax::MessengerService.deliver(::User.batch_user, grantor, message_to_grantor,
                                           I18n.t('hyrax.notifications.proxy_depositor_added.subject'))
           Hyrax::MessengerService.deliver(::User.batch_user, grantee, message_to_grantee,
                                           I18n.t('hyrax.notifications.proxy_depositor_added.subject'))
-        }
+        end
       end
 
       def decide_layout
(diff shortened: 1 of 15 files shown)


smithjp:
Would it make sense to set the locale in a hyrax config variable (in config/initializers/hyrax.rb)?

jcoyne:
@smithjp no, the language is already set in `I18n.locale`, please use that.

jcoyne:
I'm unclear why you need the `I18n.with_locale(:en)` call at all.  Is there some reason not to use the currently set locale?

smithjp:
The feedback in #3321 indicated that the first step in having emails sent in a particular user's preferred language would be to move the email text to the locale files and always send in english.  Further work will need to be done to either have emails sent in an application-wide preferred language or in the recipient's preferred language.  Otherwise, emails may be sent based on a depositor's preferred language, which may not be a reviewer's preferred language.

jcoyne:
@smithjp for now, the system default could be set to `:fr`, so it would be unexpected to send the emails in `:en`.  You are doing extra work to make it send in `:en`, but I think just using the default behavior is preferable.

no-reply:
:+1: to using the existing default `I18n.t()` values and considering that fulfillment of the first step in #3321.

Since that original discussion, I've learned that users already have a `preferred_locale`. An example usage is here: https://github.com/samvera/hyrax/blob/5a9d1be16ee1a9150646384471992b03aab527a5/app/models/user_mailbox.rb#L20

PR #3314, opened 2018-10-01
Update chromedriver helper to the latest minor release

jcoyne:
Remove the temporary pinning that was needed due to upstream bugs.

PR #3387, opened 2018-11-08
Upgrade to capybara-maleficent 0.3

cjcolvar:
Capybara-maleficent 0.3 was released yesterday to include a change needed for #3314, but it included a change in its API which is breaking the build.  This PR completes the upgrade and bumps the dependency in the gemspec.

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -82,7 +82,7 @@ SUMMARY
 
   # temporary pin to 2.17 due to failures caused in 2.18.0
   spec.add_development_dependency "capybara", '~> 2.4', '< 2.18.0'
-  spec.add_development_dependency 'capybara-maleficent', '~> 0.3'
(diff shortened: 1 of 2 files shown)


hyrax.gemspec
```diff
@@ -82,7 +82,7 @@ SUMMARY
 
   # temporary pin to 2.17 due to failures caused in 2.18.0
   spec.add_development_dependency "capybara", '~> 2.4', '< 2.18.0'
-  spec.add_development_dependency 'capybara-maleficent', '~> 0.2'
+  spec.add_development_dependency 'capybara-maleficent', '~> 0.3'
```
no-reply: Can we be more strict about this?

It seems like `~> 0.x` is going to cause the build to break on future releases. Perhaps `~> 0.3.0`, or even `0.3`

PR #3392, opened 2018-11-09
Test on the latest Rails and Ruby releases

jcoyne:


jcoyne:
@cjcolvar do you know why we need that legacy image?  We don't use phatomjs AFACT. See https://github.com/samvera/hyrax/commit/9c5322f0f7a7619430cc614dd998411ff8d221be

jcoyne:
one build here reports a failure, but the logs indicate success.

Issue #3232, opened 2018-08-31
Upgrade Universal Viewer


no-reply:
### Descriptive summary

The `pul_uv_rails` gem we use to deliver Universal Viewer is long out of date, and the UV version it ships is vulnerable to **Vulnerability 4** as described in #3187.  We are looking at local (`iiif_manifest`-side) fixes for this particular issue, but should upgrade UV anyway

### Rationale

Since `pul_uv_rails` is un/under-maintained, it's probably time for Hyrax to move to a different solution. At the very least, we should upgrade to a recent version of UV.

### Related work

#3187

escowles:
BTW, we tested a current version of UV 2 and a recent version of UV 3, and found both are vulnerable.  We filed https://github.com/UniversalViewer/universalviewer/issues/612 about this.

escowles:
WIP PR to update pul_uv_rails to UV 2.0.2 is here: https://github.com/pulibrary/pul_uv_rails/pull/11

kdid:
Moving back to ready, while we wait for dependencies to be figured out/updated for a UV update, we're working around vulnerability no. 4 in #3187 by sanitizing output from the `iiif_manifest` gem.

geekscruff:
Another bug lending weight to fixing `pul_uv_rails` https://github.com/pulibrary/pul_uv_rails/issues/12

no-reply:
We now

PR #3395, opened 2018-11-12
Sanitize output from iiif_manifest gem

kdid:
refs #3187 
refs #3232 

Per @no-reply - sanitize the output from the `iiif_manifest` gem before presenting the manifest to address vulnerability no. 4 in #3187 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -127,9 +127,12 @@ module Hyrax
 
     def manifest
       headers['Access-Control-Allow-Origin'] = '*'
+
+      json = sanitize_manifest(JSON.parse(manifest_builder.to_h.to_json))
+
       respond_to do |wants|
-        wants.json { render json: manifest_builder.to_h }
-        wants.html { render json: manifest_builder.to_h }
+        wants.json { render json: json }
+        wants.html { render json: json }
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -1,5 +1,8 @@
+require 'action_view'
+
 module Hyrax
   class WorkShowPresenter
+    include ActionView::Helpers::SanitizeHelper
```
no-reply: Instead of including the entire `SanitizeHelper`, could we use `Loofah` directly?

hyrax.gemspec
```diff
@@ -56,6 +56,7 @@ SUMMARY
   spec.add_dependency 'iiif_manifest', '>= 0.3', '< 0.6'
   spec.add_dependency 'jquery-datatables-rails', '~> 3.4'
   spec.add_dependency 'jquery-ui-rails', '~> 6.0'
+  spec.add_dependency 'jsonpath', '~> 0.5.8'
```
no-reply: I'm hoping that, in the long run, we can do this without adding the `jsonpath` dependency.

spec/presenters/hyrax/work_show_presenter_spec.rb
```diff
@@ -531,6 +531,16 @@
         expect(subject[0]['label']).to eq('Title')
         expect(subject[0]['value']).to include('Test title', 'Another test title')
       end
+
+      context "when there are html tags in the metadata" do
```
no-reply: :+1:

spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -618,5 +618,14 @@
       get :manifest, params: { id: work, format: :html }
       expect(response.body).to eq "{\"test\":\"manifest\"}"
     end
+
+    describe "when there are html tags in the labels or description" do
```
no-reply: :+1:

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -1,8 +1,11 @@
 require 'iiif_manifest'
+require 'action_view'
+require 'jsonpath'
 
 module Hyrax
   module WorksControllerBehavior
     extend ActiveSupport::Concern
+    include ActionView::Helpers::SanitizeHelper
```
no-reply: `Loofah` here, too?

kdid:
@cjcolvar - I think I'm sanitizing the 'metadata' values on the way in:
https://github.com/samvera/hyrax/pull/3395/files#diff-cd3c371abc5c4eafc5a0bbb1c4039539R221

Is that what you're referring to?

cjcolvar:
@kdid Yup, sorry I missed that.

PR #3400, opened 2018-11-13
Deprecate `Solrizer::Common` methods

no-reply:
Solizer is unsupported, and `Common` is removed from `WorkBehavior` in #3399.

@samvera/hyrax-code-reviewers


no-reply:
It looks like this module may have literally done nothing. Testing that before closing this PR.

Issue #801, opened 2017-04-18
jobs should be idempotent


hackartisan:
### Descriptive summary

AttachFilesToWorkJob, e.g. adds files repeatedly if re-run. Jobs need to be reviewed for idempotence and updated where they are not idempotent.

### Rationale

When background jobs fail it should be possible to restart them and get a successful result. Sidekiq, for example, re-queues failed jobs automatically.

atz:
This is absolutely critical.

jcoyne:
"Add a thing to a list" is not an idempotent operation.  I don't know how we would make a job be idempotent if the operation itself is not.

hackmastera:
"Ensure the thing is in the list" can be idempotent; I think that would be acceptable logic.

If it can't be made idempotent maybe it shouldn't be a job?

jcoyne:
@hackmastera but this is a list, not a set, you can have duplicates.

hackmastera:
@jcoyne maybe it's worth investigating whether we have an actual use case for duplicates. I'm willing to sacrifice that flexibility.

jcoyne:
I don't think it matters whether we have a use case as we are using a linked list to model the data.   The operation is "Create a new file set" (create is not idempotent), "create a new bitstream" (not idempotent), attach bitstream to the FileSet (idempotent!) and then "attach it to a linked list" (also not idempotent).

hackmastera:
If it can't be made idempotent it maybe shouldn't be run as a job.

If we want to keep jobs for some parts of this process, maybe there is a way to do the object creation outside of a job and then the links as jobs. "attach to linked list" can be idempotent; if we accept a constraint that an object can only be there once, we can check for it before adding.

@jcoyne do you think non-idempotent jobs are a non-issue?

cjcolvar:
:+1: I was thinking the same thing.  Maybe we can assume the unique constraint as the default but allow overriding similar to AddFileToFileSet in hydra-works: https://github.com/projecthydra/hydra-works/blob/master/lib/hydra/works/services/add_file_to_file_set.rb

mbklein:
Dredging this back up (because we're currently running into something similar).

“Add a thing to a list” is not an idempotent operation, but “apply these changes to this (new or existing) object” _has_ to be an idempotent operation. That might mean changing something other than the specific job referenced in this issue, but the feature as written is (still) broken.

mbklein:
[...]

PR #3408, opened 2018-11-14
Backport #3407

no-reply:
AttachFileToWorkJob shouldn't create a new FileSet for an UploadedFile
that already has one. This avoids creating and attaching multiple FileSets in
the event that the job fails, rendering the job idempotent with respect to
FileSet attachment.

This is a quick fix for AttachFileToWorksJob idempotency; connected to #801.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/attach_files_to_work_job.rb
+++ b/app/jobs/attach_files_to_work_job.rb
@@ -11,12 +11,14 @@ class AttachFilesToWorkJob < Hyrax::ApplicationJob
     work_permissions = work.permissions.map(&:to_hash)
     metadata = visibility_attributes(work_attributes)
     uploaded_files.each do |uploaded_file|
+      next if uploaded_file.file_set_uri.present?
+
       actor = Hyrax::Actors::FileSetActor.new(FileSet.create, user)
+      uploaded_file.update(file_set_uri: actor.file_set.uri)
       actor.file_set.permissions_attributes = work_permissions
       actor.create_metadata(metadata)
       actor.create_content(uploaded_file)
       actor.attach_to_work(work)
-      uploaded_file.update(file_set_uri: actor.file_set.uri)
     end
   end
 
--- a/spec/jobs/attach_files_to_work_job_spec.rb
+++ b/spec/jobs/attach_files_to_work_job_spec.rb
@@ -1,71 +1,69 @@
 RSpec.describe AttachFilesToWorkJob, perform_enqueued: [AttachFilesToWorkJob] do
-  context "happy path" do
-    let(:file1) { File.open(fixture_path + '/world.png') }
-    let(:file2) { File.open(fixture_path + '/image.jp2') }
-    let(:uploaded_file1) { build(:uploaded_file, file: file1) }
-    let(:uploaded_file2) { build(:uploaded_file, file: file2) }
-    let(:generic_work) { create(:public_generic_work) }
-    let(:user) { create(:user) }
+  let(:file1) { File.open(fixture_path + '/world.png') }
+  let(:file2) { File.open(fixture_path + '/image.jp2') }
+  let(:uploaded_file1) { build(:uploaded_file, file: file1) }
+  let(:uploaded_file2) { build(:uploaded_file, file: file2) }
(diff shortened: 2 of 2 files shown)


app/jobs/attach_files_to_work_job.rb
```diff
@@ -11,12 +11,14 @@ def perform(work, uploaded_files, **work_attributes)
     work_permissions = work.permissions.map(&:to_hash)
     metadata = visibility_attributes(work_attributes)
     uploaded_files.each do |uploaded_file|
+      next if uploaded_file.file_set_uri.present?
+
       actor = Hyrax::Actors::FileSetActor.new(FileSet.create, user)
+      uploaded_file.update(file_set_uri: actor.file_set.uri)
```
cjcolvar: I'm not following why this line was moved up.  It looks to me like the `uploaded_file`'s `file_set_uri` is saved eagerly before it is actually added to the `file_set`.  This appears to leave the chance that if something failed during the attaching that it would never be attached during a retry because it would be skipped.  Could you please add a test that ensures this line is at the right place in the execution flow or at least a comment before this line as to why it is important to be where it is?

no-reply: The `FileSet.create` call on line 16 determines the `FileSet`'s id/URI.

Moving this line directly below it reduces the chance that an incidental failure in the rest of the creation/attachment process will result in a duplicate, by pairing the two as early as possible.

This *does* introduce the possibility that the FileSet will be created, only to have some failure occurs during the permission, metadata, and content creation steps preventing attachment. This is why I've called this a "quick fix". Ideally, repeated runs of this job after an uncaught error will not only be idempotent, but guarantee eventual attachment of all the requested files.

I think this is better than the status quo, though. :robot: 

Other ideas and contributions are welcome.

cjcolvar: I'm thinking that we're choosing between two scenarios:
1. Without this change:
This background job fails then retries creating (and potentially attaching) filesets until it hits the max number of retries.
2. With the change:
This background job fails then retries and succeeds with having created (but potentially not attached) the fileset.

Scenario 1 would give us indication that the job failed whereas scenario 2 wouldn't.  Scenario 1 would create multiiple filesets which may be orphaned if they didn't get attached while scenario 2 would create one fileset which may be orphaned.  So I'm not really sure which is better or worse.  I think my biggest fear is that filesets wouldn't get attached and there wouldn't be any indication since the job would have succeeded on the 2nd run, but I'm not sure of any better way to catch or handle this so ¯\_(ツ)_/¯.

no-reply: > Scenario 1 would give us indication that the job failed whereas scenario 2 wouldn't.

is this true? in both cases the jobs will raise an uncaught error and retry until they succeed. `Scenario 1` *may* do this indefinitely (depending on `QueueAdapter` behavior); `Scenario 2` promises to do it maximally once per `UploadedFile`, without side effects.

In either case, the error will be reported as a failed job in whatever way the queuing system does that, logged, and captured by `honeybadger` or similar (if present).

Both approaches seem to present challenges for remediation, but the addition of the relationship between the `UploadedFile` (which is easily identified from the job params) and the `FileSet` seems likely to improve things on this front; especially combined with the removal of side effects on retries.

More importantly, I think this fix will be short-lived in production. My expectation is that we can complete the work here by replacing `next` with appropriate logic for repairing failures.

no-reply: if i understand correctly, the actual behavior people are seeing is that the file attachment *does* eventually succeed, and the outcome is something of a hybrid of the worst properties of both scenarios:
  - multiple `FileSet`s get created (and at least sometimes attached)
  - the job succeeds and falls out of the retry queue, and no one is really equipped to notice

cjcolvar: Ok, I was thinking that it didn't succeed and hit the max retries.  I guess either way one needs to be vigilant about job failures (even if the retry succeeded) which is probably a good thing to do regardless. 😄

PR #3425, opened 2018-11-17
Upgrade to active-fedora 12.0

jcoyne:


jcoyne:
All tests pass here, even though travis is reporting one as a failure.

jcoyne:
@no-reply yes, there are no remaining references to Solrizer.  Where would you like a note made?

PR #3428, opened 2018-11-17
Remove deprecated member_of_collection_ids CollectionsMembershipActor…

jcoyne:
… behavior

For inclusion in Hyrax 3

### DIFF
--- a/app/actors/hyrax/actors/collections_membership_actor.rb
+++ b/app/actors/hyrax/actors/collections_membership_actor.rb
@@ -32,13 +32,9 @@ module Hyrax
         # @return [Boolean]
         #
         # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


app/actors/hyrax/actors/collections_membership_actor.rb
```diff
@@ -31,15 +31,9 @@ def update(env)
         # @param env [Hyrax::Actors::Enviornment]
         # @return [Boolean]
         #
-        # rubocop:disable Metrics/MethodLength
-        # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
         def assign_nested_attributes_for_collection(env)
           attributes_collection = env.attributes.delete(:member_of_collections_attributes)
```
jcoyne: Missing `return true unless attributes_collection`

Issue #3377, opened 2018-11-05
Randomly failing spec - batch edit spec


adamjarling:
### Descriptive summary

This spec fails on 2018-11-05.
/spec/features/batch_edit_spec.rb:21

PR #3446, opened 2018-11-28
Update dashboard collection list item thumbnails to not overlap width…

### LINKED ISSUE #3430: Scale collection thumbnails by largest border
### Descriptive summary
Scale collection thumbnails according to the largest border. 

### Expected behavior
Thumbnails are generated according to largest border to fit images with a >>1 aspect ratio. 
see https://github.com/samvera/hyrax/issues/3430#issuecomment-439817136
### Actual behavior
Wide images' thumbnails get stretched into the title text.  

e.g. collection thumbnail of collection `Hyrax v3.0.0beta testing Chrome/Win` at page https://nurax-dev.curationexperts.com/dashboard/my/collections?locale=de

see https://github.com/samvera/hyrax/issues/3430#issuecomment-439817320

### Related work

adamjarling:
… into the title.  Also updated to have thumbnails be the same size as thumbnails in Works list view

Fixes #3430 

This fix will prevent a collection thumbnail from overlapping into the title to it's right.  Previously `height` was a fixed style on the thumbnail, but now it's adjusted to a `max-width`, which will retain the positioning within the table row, and also keep the thumbnail image at its original scale.

This also updates the Collections list thumbnails to be the same size as the Works list thumbnails.  

Note this does increase the height of the rows some.  If keeping the Collection list items rows compact is a priority, then the alternative would be to place an `overflow` style on the element wrapper for the Collection thumbnail, which would 'cut off' part of the thumbnail.   If any interest in this, let me know and I can adjust.

#### Updated
collections-list-thumbs

#### Original
collections-list-thumbs-original

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/stylesheets/hyrax/_collections.scss
+++ b/app/assets/stylesheets/hyrax/_collections.scss
@@ -209,11 +209,13 @@ button.branding-banner-remove:hover {
     display: flex;
 
     .thumbnail-wrapper {
-      width: 50px;
+      flex: 0 0 64px;
+      margin-right: 10px;
+      overflow-x: hidden;
 
       > img {
-        height: 30px;
-        width: auto;
+        height: auto;
+        max-width: 64px;
       }
     }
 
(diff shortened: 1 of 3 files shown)


adamjarling:
I've re-started this 4 times, and it keeps failing on this randomly failing test.  @cjcolvar @no-reply Maybe we just merge it in...

https://github.com/samvera/hyrax/issues/3377

Issue #3296, opened 2018-09-24
Change default label for field "description" to "Description"


ruthtillman:
### Descriptive summary

Change default label for field "description" to "Description"

### Rationale

Right now, data encoded in dc:description is labeled as “Abstract or Summary” in display. Both Cultural Heritage and IR use cases may require differentiation between an abstract and a description. 

Note, this may require some people to migrate their data, depending on how they are using abstract/description.

### Related work

https://github.com/samvera/hyrax/issues/3297

PR #3449, opened 2018-11-29
Changes description label to Description

kdid:
Fixes https://github.com/samvera/hyrax/issues/3296


no-reply:
@samvera/hyrax-metadataists can someone from this group confirm that this closes the ticket referenced?

PR #3455, opened 2018-12-04
[WIP] Addresses browse-everything 1.0.0rc2 compatibility issues

### LINKED ISSUE #1852: Upload from dropbox is not working
### Descriptive summary

Upload from cloud via Dropbox is not working due to dropbox dropping support for the APIv1.

See ticket from @awead for description https://github.com/samvera/browse-everything/issues/159

jrgriffiniii:
Fixes #1852

This aims to address compatibility issues with the latest release candidate for 1.0.0 of the `browse-everything` Gem which I've discovered affect the behavior of `ImportUrlJob` (as well as related functionality for retrieving resources from cloud service providers).

* Handling HTTP GET parameter encoding issues (e. g. `X-Amz-Credential`) in `CreateWithRemoteFilesActor#create_file_from_url`
* Ensuring that `BrowseEverything::Retriever.can_retrieve?` is used within `ImportUrlJob`
* Handles errors encountered at the file system level when using `IngestLocalFileJob` to read from a file buffer
* Ensures that filenames are preserved using `FileSet#label` within `ImportUrlJob#copy_remote_file`

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -21,17 +21,18 @@ class ImportUrlJob < Hyrax::ApplicationJob
     operation.performing!
     user = User.find_by_user_key(file_set.depositor)
     uri = URI(file_set.import_url)
+    name = file_set.label
     @file_set = file_set
     @operation = operation
 
-    unless can_retrieve?(uri)
+    unless BrowseEverything::Retriever.can_retrieve?(uri, headers)
       send_error('Expired URL')
       return false
     end
 
     # @todo Use Hydra::Works::AddExternalFileToFileSet instead of manually
     #       copying the file here. This will be gnarly.
-    copy_remote_file(uri, headers) do |f|
+    copy_remote_file(uri, name, headers) do |f|
       # reload the FileSet once the data is copied since this is a long running task
       file_set.reload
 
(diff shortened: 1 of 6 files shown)


app/jobs/import_url_job.rb
```diff
@@ -21,17 +21,18 @@ def perform(file_set, operation, headers = {})
     operation.performing!
     user = User.find_by_user_key(file_set.depositor)
     uri = URI(file_set.import_url)
+    name = file_set.label
     @file_set = file_set
     @operation = operation
 
-    unless can_retrieve?(uri)
+    unless BrowseEverything::Retriever.can_retrieve?(uri, headers)
```
mbklein: `BrowseEverything::Retriever.can_retrieve?` only expects [one argument](https://github.com/samvera/browse-everything/blob/master/lib/browse_everything/retriever.rb#L34)

jrgriffiniii: @lfarrell I'm sorry for having fallen behind on this.  This was raising a problem with requests for certain cloud storage providers: https://github.com/samvera/browse-everything/compare/master...jrgriffiniii:issues-hyrax-3.0.0

I suspect that this assisted in resolving the issue, I'm getting ready to test this soon and issue a PR against browse-everything upstream.  Hopefully this won't be blocked by https://github.com/rails/rails/pull/35154 (5.2.3 still has yet to be released on https://rubygems.org/gems/rails/versions/)

jrgriffiniii: The `X-Amz-Credential` parameter when requesting access to S3 files over the HTTP is somehow getting malformed.  This was blocking this PR from upstream.

I've also discovered that the `BrowseEverything` routes were not automatically mounted by the generator, but this may be new.

jrgriffiniii: This is likely some error with S3 URLs originating from here: https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/create_with_remote_files_actor.rb#L53

jrgriffiniii: https://github.com/samvera/hyrax/compare/master...jrgriffiniii:browse-everything-1.0.0rc2-resolves the S3 parameter issue, but I need to test it a bit further to be certain that it fixes the errors.

jrgriffiniii: I'm afraid that my Hyrax and BrowseEverything branches were effective for S3, but I've now discovered that https://github.com/samvera/browse-everything/issues/260 blocks the usage of Dropbox.  I'm looking at this next and will continue the thread on that issue while proceeding to look for a solution.

jrgriffiniii: This is also being blocked on Dropbox given that browse-everything downloads the files from the cloud server and saves them to a location in /tmp (https://github.com/samvera/browse-everything/blob/master/lib/browse_everything/driver/dropbox.rb#L93).

Unfortunately, this is not whitelisted within (https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/templates/config/initializers/hyrax.rb#L268), so while the files are downloaded successfully, Hyrax raises the following error when it tries to ingest the file:

```
User jrgriffiniii@gmail.com attempted to ingest file from url file:///var/folders/06/9nbfv2tx74jc0flpnq_mgz580000gp/T/d20190208-20689-aeo90w/cea.tif, which doesn't pass validation
```

PR #3468, opened 2019-01-04
Set Work permissions and visibility when using `create_work` transaction

no-reply:
Applies visibility, embargo, and lease when creating transactionally; duplicating the core behavior of `InterpretVisibilityActor`, allowing users of transactional creation to pass in appropriate visibility arguments to achieve embargo, lease, or just custom visibility.

Existing behavior, where simple visibility can be set on a work before a transaction begins, remains.

Several details of the `InterpretVisibilityActor` remain unimplemented. For instance, we don't check visibility/embargo/lease against a `PermissionTemplate`. It's not immediately clear whether these features should be re-added in the transactional context. It appears they cause silent/unrecoverable failures in the Actor Stack context; if so, the status quo is that clients responsible for not running afoul of them.

----

Also applies `PermissionTemplate` based on collection. The AdminSet permission template is already applied in a separate step.

The actor stack has a complex way of handling Collection templates, split across the `CollectionsMembershipActor` and the `ApplyPermissionTemplateActor`, see especially: `CollectionsMembershipActor#extract_collection_id`.

Rather than reproduce this logic, we expect callers to the `create_work` transaction to simply tell us which collections we should apply a permission template for. This allows different user input to specify how collection membership should effect permissions, even allowing multiple templates to be
applied (which is not possible in the actor stack's approach).

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/steps/apply_visibility.rb
+++ b/lib/hyrax/transactions/steps/apply_visibility.rb
@@ -27,7 +27,7 @@ module Hyrax
                                                      during:       during,
                                                      after:        after)
 
-          Hyrax::VisibilityIntentionApplicator.apply(intention).to(work)
+          Hyrax::VisibilityIntentionApplicator.apply(intention).to(model: work)
 
           Success(work)
         rescue Hyrax::VisibilityIntentionApplicator::InvalidIntentionError => err
--- a/lib/hyrax/transactions/steps/apply_visibility.rb
+++ b/lib/hyrax/transactions/steps/apply_visibility.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # A `dry-transaction` step that interprets visibility/lease/embargo from
+      # passed arguments.
+      #
+      # @since 3.0.0
+      class ApplyVisibility
+        include Dry::Transaction::Operation
+
+        ##
+        # @param [Hyrax::WorkBehavior] work
+        # @param [String] visibility
+        # @param [String] release_date
+        # @param [String] during
+        # @param [String] after
+        #
+        # @return [Dry::Monads::Result] `Failure` if there is no
+        #   `PermissionTemplate` for the input; `Success(input)`, otherwise.
+        def call(work, visibility: nil, release_date: nil, during: nil, after: nil)
+          return Success(work) unless visibility.present?
+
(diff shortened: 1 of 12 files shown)


spec/hyrax/transactions/create_work_spec.rb
```diff
@@ -83,6 +83,40 @@
...
+    let(:step_args) { { apply_visibility: [visibility: request, release_date: end_date, during: during, after: after] } }
+
+    it 'sets the embargo' do
+      expect { transaction.with_step_args(step_args).call(work) }
+        .to change { work.embargo }
+        .to be_an_embargo_matching(release_date: end_date, during: during, after: after)
+    end
+  end
+
+  context 'when requesting an embargo' do
```
cjcolvar: This context looks like a duplicate block.  Should it be about setting visibility (apart from lease or embargo) or can it just be removed?

lib/hyrax/transactions/steps/apply_visibility.rb
```diff
@@ -0,0 +1,39 @@
...
+        #   `PermissionTemplate` for the input; `Success(input)`, otherwise.
+        def call(work, visibility: nil, release_date: nil, during: nil, after: nil)
+          return Success(work) unless visibility.present?
+
+          intention = Hyrax::VisibilityIntention.new(visibility:   visibility,
+                                                     release_date: release_date,
+                                                     during:       during,
+                                                     after:        after)
+
+          Hyrax::VisibilityIntentionApplicator.new(intention: intention).apply_to(work)
```
cjcolvar: How about adding the convenience methods that are in Hyrax::PermissionTemplateApplicator to Hyrax::VisibilityIntentionApplicator so that these call chains match and this line becomes:
`Hyrax::VisibilityIntentionApplicator.apply(intention).to(model: work)`
The two applicator service classes are so similar that it seems like they should conform to the same interface so there is a clear pattern for future classes and to ease in reading them.

no-reply: I like this. and will add it.

spec/hyrax/transactions/create_work_spec.rb
no-reply: I've removed this block, thanks!

lib/hyrax/transactions/steps/apply_visibility.rb
no-reply: added. I got some more unit tests in for this class in the meanwhile. Thanks for the feedback.

PR #3471, opened 2019-01-14
I18n omissions

j-dornbusch:
Fixes some omissions in internationalization

@samvera/hyrax-code-reviewers


j-dornbusch:
I have a test failure but I do not understand how it relates to the changes I made.
@samvera/hyrax-code-reviewers

j-dornbusch:
I fixed almost all problems reported by Rubocop, Capybara, etc., but I do not understand what connection there is between my changes (solely related to translations) and this test failure:
```
1) AttachFilesToWorkJob with uploaded files at remote URLs behaves like a file attacher attaches files, copies visibility and permissions and updates the uploaded files
     Failure/Error: expect(ImportUrlJob).not_to have_been_enqueued
       expected not to enqueue exactly 1 jobs, but enqueued 1
```
@samvera/hyrax-code-reviewers

app/views/hyrax/base/_show_actions.html.erb
```diff
@@ -16,20 +16,20 @@
           class: presenter.display_feature_link? ? 'btn btn-default collapse' : 'btn btn-default' %>
     <% end %>
     <% if Hyrax.config.analytics? %>
-      <%= link_to t('.analytics'), presenter.stats_path, id: 'stats', class: 'btn btn-default' %>
+      <%= link_to t('hyrax.base.show_actions.analytics'), presenter.stats_path, id: 'stats', class: 'btn btn-default' %>
```
jcoyne: why are these changes necessary?  Doesn't both the before and after resolve to the same key?

j-dornbusch: I agree. But with the implicit scope version (".analytics"), tests fail. Try.

tasks/quick_translation.rake
```diff
@@ -0,0 +1,70 @@
+# Usage : rake add_entry['fr','blacklight.search.my_new_command','Ma nouvelle commande']
```
no-reply: @j-dornbusch can you comment on the intended use of this rake task? Is this a proposed replacement for the process in https://github.com/samvera/hyrax/wiki/Translations-(internationalization)?

I'm not sure I understand.

j-dornbusch: @no-reply Oops! I was not aware of the existence of the page you mention. Indeed, the rake task I added duplicates the existing process.

j-dornbusch: @no-reply After checking, I think the rake task I added is an interesting solution for people who are not able or willing to put a card number on their Google Account. I remove it from the pull request to allow the merge. But we can share and document it if it is desired by the community.

cjcolvar:
I restarted the failed travis build but now it is hitting the chromedriver issue that was resolved a week or so ago.  I think travis and circleci will pass if this PR is rebased on top of master.

cjcolvar:
It looks like you brought in all of the commits from master as new commits instead of a rebase.  From what I can tell, I think this PR will be good once the commit history gets fixed.  It would also be good to squash the commits in this PR into one since the work is now focused on one topic.  The [Contributing guide](https://github.com/samvera/hyrax/blob/master/.github/CONTRIBUTING.md#submitting-changes) provides guidance on squashing and rebasing if you need it.

j-dornbusch:
@cjcolvar Sorry I pushed from my IDE instead of doing it from the command line and something unclear happened.

cjcolvar:
:+1: I figured it was probably something like that.  I think this is close and looking forward to getting it merged!  Thanks for pushing this through.

j-dornbusch:
@cjcolvar I tried to let Rubymine do the work for me but I probably pressed the wrong button.
By rebasing manually I get a long series of conflicts to solve one by one. I will give up this PR, it is too old, and start on a new branch. The most important in the PR was not the few translated strings but the rake task that translate automatically without a paid Google account. I will continue to use it to internationalize strings faster. Thanks for your help.

cjcolvar:
@j-dornbusch I think the i18n translations that you made are still a worthwhile PR to make.  Would you consider opening a PR for just them?

PR #3472, opened 2019-01-14
Require hyrax/search_state

cjcolvar:
When attempting to set the admin_set_create_service for the AdminSetsController in the hyrax initializer, you get an uninitialized constant error for Hyrax::SearchState.  It is required in an initializer within the Hyrax engine but for some reason that hasn't been loaded by this point in the hyrax initializer.  This commit makes the dependency explicit by requiring the load of search state here when Hyrax::Controller is used.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -1,3 +1,5 @@
+require 'hyrax/search_state'
+
 module Hyrax::Controller
   extend ActiveSupport::Concern
 


cjcolvar:
FWIW The two failed travis builds had all of their rspec tests pass.

no-reply:
This seems okay to me, but I'm not sure I'm understanding why the relevant requires need to be in an tantalizer at all. I'm going to try simply moving them out to the main body of the `Hyrax::Engine`

cjcolvar:
That approach looks good to me.  The only reason I can think of to load those later in an initializer is to help speed up the load time or because of some load order issue but those both seem unlikely and if they work outside of the initializer then I'm :+1:.

cjcolvar:
@no-reply Do you have time to rebase that branch and open a PR or would you like me to do that?

cjcolvar:
Closed by #3686.

PR #3473, opened 2019-01-15
Add `DestroyWork` transaction, and make it destroy the a work

no-reply:
This is a first step to replacing the `#destroy` stack in the Actor Stack.

A `DestroyWork` transaction is added. It's sole current operation is to destroy
the work if it is persisted. When the work is not persisted, we return a
`Success`; errors are caught and wrapped in a `Failure`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/container.rb
+++ b/lib/hyrax/transactions/container.rb
@@ -19,7 +19,9 @@ module Hyrax
     # @see https://dry-rb.org/gems/dry-container/
     class Container
       require 'hyrax/transactions/create_work'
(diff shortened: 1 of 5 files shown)


lib/hyrax/transactions/container.rb
```diff
@@ -20,6 +20,7 @@ module Transactions
     class Container
       require 'hyrax/transactions/create_work'
```
cjcolvar: Looks like you're missing the require for the destroy work transaction.

Issue #3421, opened 2018-11-16
encoding bug of `"` for non-default language


jmfrenzel:
### Descriptive summary
encoding bug of &ldquor; &ldquo; for German  language  

### Expected behavior
Übertragung des Eigentums von „Fedora“ 

Guillemets are preferred in German.  I started to correct this earlier in https://github.com/samvera/hyrax/pull/2443/files
 So #3421 requests to display all quotations as for example: https://nurax-dev.curationexperts.com/concern/generic_works/gm80hv50h/edit?locale=de
 ... im Abschnitt „Freigeben“ einen bestimmten ... 

 
### Actual behavior
https://nurax-dev.curationexperts.com/works/hd76s024f/transfers/new?locale=de

### Steps to reproduce the behavior
1. https://nurax-dev.curationexperts.com/works/hd76s024f/transfers/new?locale=de

### Related work

Link to related tickets or prior related work here.

jmfrenzel:
Changed description,  see:
https://github.com/samvera/hyrax/pull/3476#issuecomment-456905528

jmfrenzel:
![Quotes](https://user-images.githubusercontent.com/30596226/56312711-168b8b80-6151-11e9-8ee6-285d1bd6f93e.png)

cjcolvar:
Was this fixed by https://github.com/samvera/hyrax/pull/3767?

jmfrenzel:
No it wasn't. 
The template does not render correctly, see
https://github.com/samvera/hyrax/blob/3af966faf937f10446e4983ca1f384df5ca9c2da/app/views/hyrax/transfers/new.html.erb#L2  
This is the line in the German locale being affected.
https://github.com/samvera/hyrax/blob/3af966faf937f10446e4983ca1f384df5ca9c2da/config/locales/hyrax.de.yml#L1181
To me I seems the is with the variable %{work_title}

The other non working rendering is the help text on the same page  
https://github.com/samvera/hyrax/blob/3af966faf937f10446e4983ca1f384df5ca9c2da/config/locales/hyrax.de.yml#L1180

fritzfreiheit:
I think a general solution to this problem would be to allow for html characters to remain html characters when processed by I18n.translate, rather than replace strings like '&ldquor;' with '&amp;ldquor;'.

fritzfreiheit:
[...]

PR #3476, opened 2019-01-22
Replace &ldquo with " in German translation

laritakr:
Fixes https://github.com/samvera/hyrax/issues/3421

Note: This could be related to the process which automatically builds the
translation files, but since it only happened in German, it may have
some other cause. We should probably watch to make sure this doesn't
recur.

* Verify that https://nurax-dev.curationexperts.com/works/hd76s024f/transfers/new?locale=de resolves to `Übertragung des Eigentums von "Fedora"`

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.de.yml
+++ b/config/locales/hyrax.de.yml
@@ -147,7 +147,7 @@ de:
             title: Freigabe
             varies:
               any: Dem Deponenten erlauben, sich zu entscheiden
-              between: Zwischen &ldquor; jetzt &ldquo; und
+              between: Zwischen "jetzt" und
               description: 'Variabel - Deponenten können den Freigabedatum für eine einzelne Arbeit festlegen:'
               embargo:
                 1yr: 1 Jahr nach Deponierung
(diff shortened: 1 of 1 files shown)


jcoyne:
Aren't guillemets preferred in German?

laritakr:
I guess that's a decision for @jmfrenzel

no-reply:
I think these changes were mostly added manually in https://github.com/samvera/hyrax/pull/2443/files

laritakr:
Ok that's a relief... I was wondering how they got in there and why no other languages had them, but hadn't taken the time to track it down.

jmfrenzel:
Yes, guillemets are preferred in German. And this is what I requested in #3421, However, I provided a false positive example. (Verify that https://nurax-dev.curationexperts.com/works/hd76s024f/transfers/new?locale=de) . I started to correct this earlier in https://github.com/samvera/hyrax/pull/2443/files

So #3421 requests to display all quotations as for example: 
https://nurax-dev.curationexperts.com/concern/generic_works/gm80hv50h/edit?locale=de
... im Abschnitt „Freigeben“ einen bestimmten ...

So #3421requests to decode e.g.:
Übertragung des Eigentums von &ldquor;Fedora&ldquo;    
like this
Übertragung des Eigentums von „Fedora“

laritakr:
@jmfrenzel the work you linked to in nurax-dev is private so I can't see it. Can you either change the visibility or provide a screen cap instead?

jmfrenzel:
Set it  to public. However, the example refers to the "Save Work" | "Visibility" section | "private" on the edit work view.

geekscruff:
@jmfrenzel is this fixed now? I'm working on the Hyrax WG sprint so can pick it up if not?

PR #3479, opened 2019-01-24
Lock unauthorized permissions from update

laritakr:
Fixes https://github.com/samvera/hyrax/issues/3166

### Allowed permission maintenance:
- user is permitted to update any work permissions coming from collections they manage
- user is permitted to update non-manager permissions from any Collections
- user is permitted to update non-collection permissions 

### Process used:
- find all of the other managers of collections a user cannot manage who are not also managers in collections that the user CAN manage.

This gives us the permissions the user is not authorized to update. The shared partial, `currently_shared.html.erb` embeds all of the above logic and uses it to display all of the work or file set's permissions in either a fixed or editable format. The new shared partial results in a slight reformat of the display of these permissions, as there were previously minor differences between the two displays.

Additionally, a new partial was created for batch edit, as the above rules cannot be enforced at a batch level. Dealing with batch editing sharing issues is beyond the scope of this issue.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/edit_permissions_service.rb
+++ b/app/services/hyrax/edit_permissions_service.rb
@@ -1,0 +1,112 @@
+module Hyrax
+  class EditPermissionsService
+    # Encapsulates the logic to determine which object permissions may be edited by a given user
+    #  - user is permitted to update any work permissions coming ONLY from collections they manage
+    #  - user is not permitted to update a work permission if it comes from a collection they do not manage, even if also from a managed collection
+    #  - user is permitted to update only non-manager permissions from any Collections
+    #  - user is permitted to update any non-collection permissions
+    attr_reader :depositor, :unauthorized_collection_managers
+
+    # @param [Object] GenericWorkForm (if called for object) or GenericWork (if called for file set)
+    # @param [Ability] user's current_ability
+    def initialize(object:, ability:)
+      @object = object
+      @ability = ability
+      @depositor = object.depositor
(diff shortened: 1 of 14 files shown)


laritakr:
@no-reply 

As far as I know, all that is left is to fix failures in the batch_edit and batch_create specs, because the file_sets in these specs do not respond to depositor which `base/currently_shared` partial requires. I'm not sure where in the UI to test this to determine if the problem is with the code or with the specs.

cjcolvar:
@laritakr Is this ready for review?

laritakr:
It was complete, but there was some question whether the behavior I put in was actually wanted after all. I assumed the PR was dead. 

Before merging this, be sure to discuss it with @no-reply

cjcolvar:
Thanks for the background.  I know you spent a lot of effort on this and no one had reviewed it before stale-bot closed it so I brought it up during the weekly PR review time and we couldn't remember if it was considered done or not.  I'll try to get a review in soon but I'll talk with @no-reply about timeline if it should be a part of 3.0 or post-3.0.

laritakr:
I think the problem is that there wasn't consensus whether the change in behavior is desired. The assigning of permissions via admin sets and collections does not apply permissions beyond initial creation of the collection, and there isn't consensus on whether this should be changed. 

It is more than just an issue of timing, so I gave up on trying to get it reviewed.

no-reply:
bumping this since it's assigned to me.

Issue #3128, opened 2018-06-22
Ldp::Conflict occurring on all new works once it starts


elrayle:
### Descriptive summary

In Hyrax 2.1.0, I am able to create a bunch of works.  At some point, creating a new work starts getting an `Ldp::Conflict, "Can't call create on an existing resource"` error.  After that, all new works get this error.  I am able to clear the problem by using ActiveFedora::Cleaner.  This is not a viable solution if this occurs in production.

### Steps to reproduce the behavior

Unfortunately, I do not have a sense of what starts triggering the `Ldp::Conflict` error to start.  Once it does, you can see it with every attempt to create a new work.

### Related work

https://github.com/samvera/ldp/issues/91

elrayle:
Discussions from Slack:

@no-reply 
> I doubt that this conflict thing is actor stack related, but instead has to do with minter state.
>
> if the minter state is emitting an id that a work already has (i.e. it's always one behind), i'd expect Conflict on every `#create` call.
>
> could confirm by bumping the minter state counter in the db manually
i.e. take `#seq` and bump it to some arbitrary number `10_000`. https://github.com/samvera/noid-rails/blob/master/db/migrate/20160610010003_create_minter_states.rb#L9
notably, I haven't run into this in `prod` in over a year of hosting a Hyrax 1.0 app running AF::Noid. I wonder whether we have an issue that's new in Hyrax 2.x with `Noid::Rails`, or else an issue that's db specific

Several replies about databases in use which included MySql, Postgres, and sqlite.

elrayle:
I tried bumping the seq number in MySql db.  No effect.
I restarted rails after bumping seq number.  No effect.

We finally cleared the error by...
* when ldp::conflict happened, we dumped the URL
* deleted the resource from Fedora
* deleted the tombstone from Fedora

Now we can create works again.

elrayle:
We are able to reproduce the `Ldp::Conflict` problem by...

* Dashboard -> Works -> New Work
* upload word doc file (without soffice properly installed)
* save

This gets an error about soffice.

* Dashboard -> Works -> New Work
* upload txt file
* save

This gets Ldp::Conflict error.

elrayle:
From @no-reply in Slack...

> oh, that's interesting... there would be an ActiveRecord transaction rollback when the actor stack fails. I wonder if *that* is the issue.
i.e. minter state rollback once, then each time you try to create with the stack, you fail with Conflict and rollback again.

[...]

PR #3481, opened 2019-01-25
Recover rolled back minter state on actor stack failure

no-reply:
When the actor stack fails, a database transaction rolls back. When the DB-based
NOID minter is in use, this rolls back the minter state. However, it does *not*
roll back any objects actually created by ActiveFedora in the process of running
the stack. This leads lagged minter state, and a perpetual `Ldp::Conflict`
blocking creation of further objects.

Rather than keeping this lagged minter state, we can catch any errors raised
through the transaction, and update the minter state. To do this, we mint ids
and send HEAD requests to Fedora until we can find an id that does not already
have a Fedora object. This first valid id is wasted, but future mints are
unblocked.

This process only runs in cases where an error has been raised by the
transactions. This is mostly an effective time to remediate, except two caveats:

  - Explict raises of `ActiveRecord::Rollback` are eaten by the transaction, so
    won't recover. Actors, therefore, should not explictly rollback; this is
    probably an anti-pattern to begin with.
  - When an error is raised, but no ActiveFedora objects have been created, we
    are over-eager and end up running the recovery code and wasting the test
    id.

Related to #3128.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/transactional_request.rb
+++ b/app/actors/hyrax/actors/transactional_request.rb
@@ -10,6 +10,10 @@ module Hyrax
         ActiveRecord::Base.transaction do
           next_actor.create(env)
         end
+      rescue StandardError => err
+        # note that this does not catch explicit `ActiveRecord::Rollback` calls.
+        env.curation_concern.ensure_valid_minter_state
+        raise err
       end
 
       # @param [Hyrax::Actors::Environment] env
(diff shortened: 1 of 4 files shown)


no-reply:
Some [slack discussion](https://samvera.slack.com/archives/C0F9JQJDQ/p1548450183396000) about this:

> @tpendragon: Is 3481 necessary if 3482 is merged?

Short answer: no.

Long answer: it may be an option for backport to 2.x. Depending on our comfort level with #3482 as a backport.

PR #3484, opened 2019-01-28
Shot in the dark for build failures

jeremyf:
In conversation with Chris, RSpec sometimes reports 0 errors, but
the suite returns a non-zero shell response. The CI tool, then
interprets the non-zero response as a failure to build.

My thinking is that the after blocks of rspec may be returning false
for the last run in the suite, and that is interpretted as a non-zero.
This would be sad panda if that were the case, but hey, maybe.

This is very much a stab in the dark.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -210,6 +210,7 @@ RSpec.configure do |config|
 
   config.after do
     DatabaseCleaner.clean
+    true
   end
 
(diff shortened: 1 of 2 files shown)


jeremyf:
I'm going to close this PR as I believe #3493 may resolve this issue more reliably.

jrgriffiniii:
This is still affecting builds, please see https://github.com/samvera/hyrax/pull/3505

PR #3485, opened 2019-01-29
Pin `dry-logic` to 0.4.x

no-reply:
`dry-logic` 0.5.0 breaks `require` for `dry-validation` on current versions. The
pin should resolve this problem and fix the build.

@samvera/hyrax-code-reviewers


no-reply:
This was fixed by the release of `dry-validation` 0.12.3/0.13.0

PR #3486, opened 2019-01-29
Add a factory for `Valkyrie::Resources` generated from AF objects

no-reply:
This adds a `Valkyrie` dependency.

Please discuss.

@samvera/hyrax-code-reviewers


revgum:
I like it.. a good starting point for the Valkyrie<->AF work and doesn't impact any existing code.

PR #3498, opened 2019-01-31
Add an initial Valkyrie `ResourceFactory` to Wings

no-reply:
This incorporates an initial Valkyrie::RescourceFactory to unblock further work. 

@samvera/hyrax-code-reviewers


jeremyf:
CircleCI specs passed, something in script report non-0 exit.

PR #3503, opened 2019-01-31
Refactor `Wings::Valkyrie::ResourceFactory` to use a converter class

no-reply:
Introduces `Wings::ActiveFedoraConverter` and uses it in the
`ResourceFactory`. The converter knows how to get an id and attributes to build
an AF model for a given `ValkyrieResource`.

It currently counts on `#internal_resource` being present, and assumes that the
id is the first entry in `#alternate_ids`. We may need to revisit either of
these assumptions later.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Wings
+  class ActiveFedoraConverter
+    ##
+    # @!attribute [rw] resource
+    #   @return [Valkyrie::Resource]
+    attr_accessor :resource
+
+    ##
+    # @param [Valkyrie::Resource]
+    def initialize(resource:)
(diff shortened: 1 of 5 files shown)


lib/wings/active_fedora_converter.rb
```diff
@@ -0,0 +1,42 @@
...
+    ##
+    # @param [Valkyrie::Resource]
+    def initialize(resource:)
+      @resource = resource
+    end
+
+    ##
+    # @return [Hash<Symbol, Object>]
+    def attributes
+      attrs = resource.attributes.reject { |k, _| k.to_s.end_with? '_ids' }
```
revgum: I feel like it would be worth commenting the intention here and its near term removal.

no-reply: :+1:

PR #3504, opened 2019-01-31
Add wings initializer

kelynch:
Adds the Valkyrie metadata adapter to Hyrax config from wings

Apps now use `require 'wings'` to find the registered adapter.

@samvera/hyrax-code-reviewers


no-reply:
can we squash these commits? :)

kelynch:
Note -- this is blocked until #3506 is able to be merged into master.  After that, tests should (hopefully) pass.

lib/wings.rb
```diff
@@ -8,7 +8,19 @@ module Wings
...
 require 'valkyrie'
 require 'wings/resource_factory'
+require 'wings/metadata_adapter'
 require 'wings/valkyrizable'
 require 'wings/valkyrie_monkey_patch'
 
 ActiveFedora::Base.include Wings::Valkyrizable
+
+Valkyrie::MetadataAdapter.register(
+  Valkyrie::Persistence::ActiveFedora::MetadataAdapter.new,
```
revgum: I think in this context this should be `Wings::Valkyrie::MetadataAdapter.new, :wings_adapter` here, since we're wanting to register the wings specific active fedora metadata adapter. Is that right?

lib/wings.rb
```diff
@@ -8,7 +8,19 @@ module Wings
 
 require 'valkyrie'
 require 'wings/resource_factory'
+require 'wings/metadata_adapter'
```
revgum: And this can be replaced along with the updates are made to `wings.rb` on `master`.

revgum:
@kelynch This is ready for rebasing from `master` with a couple comments I made.. cool deal, we're getting there! I'll be in the air for a bit, but should be back online by about noon PST. 

Again, it was great pairing with you last week!

jcoyne:
@kelynch there appears to be a conflict now.  Can you resolve it?

no-reply:
I pushed a rebased version as #3562

PR #3505, opened 2019-01-31
Do not cache Bundler dependencies

jrgriffiniii:
This fixes CircleCI builds using cached, older releases of chromedriver-helper


jeremyf:
There are some issues related to ChromeDriver that this PR helps restore. CircleCI has 0 failing specs, but exits with a 1.

PR #3506, opened 2019-01-31
Create wings adapter

laritakr:
Move metadata_adapter, query_service, and persister from hyrax-on-wings to hyrax.

@samvera/hyrax-code-reviewers


revgum:
Tests actually passed on CircleCI, although it shows as having failed.

Verified this branch works in https://github.com/samvera-labs/hyrax-on-wings on PR https://github.com/samvera-labs/hyrax-on-wings/pull/10. We're able to create a new GenericWork and save it to Fedora by way of Wings. 

@cjcolvar @no-reply 

It could be that we want to improve test coverage for the metadata adapter and query service, but otherwise I think it's ready for merging. Thoughts?

Issue #2940, opened 2018-04-18
Refactor tests using collections and admin sets to use light weight factories


elrayle:
### Descriptive summary

New factories that avoid fedora access unless necessary were created for collections and admin sets.  Parts can be created as needed, including permission template, solr doc, nested indexing, and the actual fedora object.  All tests that use collections and admin sets should use the lighter weight factories.  

Tests can be updated and committed a few at a time.  When the last test is updated, the old factories can be removed.

There are lots of examples in the light weight factories for building/creating collections and admin sets.  The basic build is...

```
build(:collection_lw)
build(:adminset_lw)
```

See factories for more examples...
* [spec/factories/collections.rb](https://github.com/samvera/hyrax/blob/master/spec/factories/collections.rb)
* [spec/factories/admin_sets_lw.rb](https://github.com/samvera/hyrax/blob/master/spec/factories/admin_sets_lw.rb)

### Rationale

Avoid accessing Fedora in tests as it is very slow.

### Related work

Link to related tickets or prior related work here.

no-reply:
Closing this. I don't think it's worth focused effort to complete this work in light of Wings/Valkyrie.

PR #3517, opened 2019-02-01
New Collection Factory Methods in Spec Tests

mtribone:
refs #2940 

This is an extension of the previous PRs from a sprint week to tackle updating the use of the old collection factory and `create` for the new factories and using `build` where possible. 

Previous PRs
- #3483 
- #3487 
- #3488 
- #3490 

* swaps `create` for `build`
* updates factory methods to use `collection_lw` or `public_collection_lw` etc.
* replaces `create_access: true` with `with_permission_template_true` where it can be done

@samvera/hyrax-code-reviewers


no-reply:
Thanks for making this happen!

PR #3528, opened 2019-02-06
WIP: Download from S3 if Needed During File Processing

### LINKED ISSUE #3475: Collection Branding Fails Upload if using S3
### Descriptive summary

When using S3 backed file uploads, a call to file.file_url will return a http based url.  This gets passed in to https://github.com/samvera/hyrax/blob/master/app/models/collection_branding_info.rb#L22 where FileUtils.cp is called.  FileUtils doesn't know about remote files and thus throws an exception. We're seeing this in Hyku, based on Hyrax 2.3.3, but the code looks unchanged in master as well. 

### Rationale

As a user when I upload a banner or icon for a collection I expect that the banner or icon to upload successfully. 

### Actual behavior

If Carrierwave is set to anything other than local files (S3 as a common example) the code will throw `Errno::ENOENT (No such file or directory @ rb_sysopen` exception.

1. Set up Hyrax with Carrierwave settings for S3
2. Create a collection
3. Edit that collection and set either a banner or a logo

### Possible Fix

By adding code to make sure the file is locally cached to the controller at [here](https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/dashboard/collections_controller.rb#L233) and [here](https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/dashboard/collections_controller.rb#L258) we were able to work around this problem

```
          if file.file_url.match(/^http/)
            file.file.download!(file.file_url)
          end
 ``` 

After the call to download! file.file_url will return the locally cached copy.

### PR DESCRIPTION
**Pull request by orangewolf**
Logo and branding fail if S3 is used for file uploads. Since all of CarrierWave settings get changed in this case, all file uploaders have to handle the idea that the files may be on S3. 

Fixes #3475

* I'm not sure how to write a test spec for this issue and would love to hear any suggestions about the right way to approach this.
* To manually test this code - set carrierwave initializer to 

```
require 'carrierwave'

if Settings.s3.upload_bucket
  CarrierWave.configure do |config|
    # config.fog_provider = 'fog/aws' # we use carrierwave-aws instead of fog now
    # config.fog_credentials = {
    #   provider: 'AWS',
    #   use_iam_profile: true
    # }
    config.storage = :aws
    config.aws_bucket = Settings.s3.upload_bucket
    config.aws_acl = 'bucket-owner-full-control'
  end
end
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -230,6 +230,9 @@ module Hyrax
 
         def add_new_banner(uploaded_file_ids)
           f = uploaded_files(uploaded_file_ids).first
+          if f.file_url.match(/^http/)
+            f.file.download!(f.file_url)
+          end
           banner_info = CollectionBrandingInfo.new(
             collection_id: @collection.id,
             filename: File.split(f.file_url).last,
@@ -255,6 +258,9 @@ module Hyrax
 
         def create_logo_info(uploaded_file_id, alttext, linkurl)
           file = uploaded_files(uploaded_file_id)
+          if file.file_url.match(/^http/)
+            file.file.download!(file.file_url)
+          end
           logo_info = CollectionBrandingInfo.new(
             collection_id: @collection.id,
             filename: File.split(file.file_url).last,


cjcolvar:
This looks great!  Could you please add a couple tests for the logo and banner scenarios modified in this PR?  I'm guessing these would end up in this testing block: https://github.com/samvera/hyrax/blob/master/spec/controllers/hyrax/dashboard/collections_controller_spec.rb#L257

PR #3529, opened 2019-02-06
Remove valk 15 monkey patch

no-reply:
Remove Wings monkey patches for Valkyrie 1.5.0

@samvera/hyrax-code-reviewers


no-reply:
@cjcolvar thanks! removed.

PR #3533, opened 2019-02-07
Cache test app on template and Gemfile.lock in Circle

no-reply:
This is a somewhat aggressive cache key scheme, leading to eager invalidation
when dependencies change.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ jobs:
 
     - restore_cache:
         keys:
-        - v1-test-app-{{ checksum "template.rb" }}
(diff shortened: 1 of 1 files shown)


jrgriffiniii:
I was trying to find a way around https://github.com/samvera/hyrax/blob/master/.circleci/config.yml#L95 with a more complex cache key, but if this can be postponed then I'm 👍

no-reply:
@jrgriffiniii that would be very welcome. I couldn't figure out a nice way of making the key directory-based, which is why i went with the diff/grep approach. There must be a better way, but...

PR #3535, opened 2019-02-12
Main i18n omissions

j-dornbusch:
Fixes some omissions in internationalization
@cjcolvar This PR includes the contents of the previous one without the automatic translation script based on Yandex : however I continue to use it, to avoid the Google developer account.
@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.de.yml
+++ b/config/locales/hyrax.de.yml
@@ -364,15 +364,10 @@ de:
           actions: Aktion
           title: Arbeitstitel
       form_files:
+        add_files: Dateien hinzufügen
+        add_folder: Ordner hinzufügen
         dropzone: Dateien hier hinziehen.
-        local_upload_browse_everything_html: |
-          <p>Sie können eine oder mehre Dateien von Ihrem lokalen System oder einem Cloudanbieter hinzufügen, welche mit dieser Arbeit
-          verknüpft werden sollen.</p>
-          <p>Bitte beachten Sie, dass der Cloud-Anbieter möglicherweise nicht in der Lage ist Ihrer Anfrage nachzukommen, wenn Sie eine große Anzahl von Dateien
-          innerhalb einer kurzen Zeitspanne hochladen. 
-          Wenn Sie
-          irgendwelche Fehler beim Hochladen aus der Cloud erhalten, teilen Sie uns dies 
-          bitte über %{contact_href} mit.</p>
+        local_upload_browse_everything_html: "<p>Sie können eine oder mehre Dateien von Ihrem lokalen System oder einem Cloudanbieter hinzufügen, welche mit dieser Arbeit\nverknüpft werden sollen.</p>\n<p>Bitte beachten Sie, dass der Cloud-Anbieter möglicherweise nicht in der Lage ist Ihrer Anfrage nachzukommen, wenn Sie eine große Anzahl von Dateien\ninnerhalb einer kurzen Zeitspanne hochladen. \nWenn Sie\nirgendwelche Fehler beim Hochladen aus der Cloud erhalten, teilen Sie uns dies \nbitte über %{contact_href} mit.</p>\n"
         local_upload_html: "<p> Sie können eine oder mehrere Dateien hinzufügen, die dieser Arbeit zugeordnet werden sollen. </p>"
       form_member_of_collections:
         actions:
@@ -411,6 +406,8 @@ de:
       form_thumbnail:
         help_html: Wählen Sie die Datei aus, die als Miniaturansicht für diese Arbeit verwendet werden soll.
         legend_html: Miniaturansicht
+      form_visibility_component:
+        visibility: Sichtbarkeit
       items:
         actions: Aktionen
         date_uploaded: Datum Hochgeladen
@@ -430,7 +427,16 @@ de:
         definitions_html: "<strong>Ansicht / Download: Auf</strong> diese Datei (sowohl Inhalte als auch Metadaten) kann innerhalb von %{application_name} zugegriffen werden. <br /> <strong>Bearbeiten:</strong> Diese Datei (sowohl Inhalte als auch Metadaten) kann bearbeitet werden. Sie dürfen diese Berechtigung nur Benutzern und / oder Gruppen von %{institution_name} erteilen."
         share_with_html: Sie können den Zugriff "Ansicht / Download" oder "Bearbeiten" für bestimmte Benutzer und / oder Gruppen auf Dateien gewähren. Geben Sie nacheinander einen gültigen %{account_name} ein, wählen Sie die Zugriffsebene für diesen Benutzer und klicken Sie auf + Hinzufügen.
       show:
+        items: Objekte
         last_modified: Zuletzt geändert
+        relationships: Beziehungen
+      show_actions:
+        analytics: Analytics
+        attach_child: Untergeordnete Arbeit hinzufügen
+        delete: Löschen
+        edit: Bearbeiten
+        feature: Hervorheben
+        unfeature: Hervorhebung entfernen
       social_media:
         facebook: Facebook
         google: Google+
@@ -440,6 +446,9 @@ de:
       help:
         resource_type: Sie können mehrere Typen auswählen, die für alle Dateien gelten sollen
         title: Dateiname ist der Standardtitel. Bitte geben Sie einen aussagekräftigeren Titel, und Dateinamen werden noch vom System bewahrt.
+    batch_edits:
+      delete_selected:
+        button_label: Ausgewählte Löschen
     batch_uploads:
       disabled: Diese Funktion vom Administrator des Repositoriums deaktiviert
       files:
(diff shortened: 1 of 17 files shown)


j-dornbusch:
@jcoyne I dont really undertand why a test is looking for "en.hyrax.base.show_actions**.html.erb**.analytics". Internationalization keys do not include file extensions ? What am I missing ?

jcoyne:
@j-dornbusch I have no idea. That looked really odd to me too.

j-dornbusch:
@jcoyne 
This line in the test : 
https://github.com/samvera/hyrax/blob/37a08e384bb9b7dbc8859dc3d126e533e534814d/spec/views/hyrax/base/_show_actions.html.erb_spec.rb#L18
seems suspicious to me.

j-dornbusch:
This syntax is present only in this spec and in this other one : https://github.com/samvera/hyrax/blob/master/spec/views/hyrax/file_sets/_single_use_links.html.erb_spec.rb

cjcolvar:
FWIW I verified locally on master that those two spec files still pass after removing `.html.erb` from the `render` lines.  So I think you should go ahead and make that change in this PR.

j-dornbusch:
@cjcolvar Searching with regex `render '[^']*.html.erb'` in spec files gives 13 results, I eliminated 9 of them.

j-dornbusch:
Searching with `render template: '[^']*.html.erb'` gives 2 occurences, I eliminated both.

j-dornbusch:
The rake task for alphabetical sorting had not been applied before this PR. It moved some keys and apparently it does not preserve double line breaks.

j-dornbusch:
@cjcolvar I corrected everything I could but the last failure ("Circular dependency detected while autoloading constant Hyrax::Actors::OrderedMembersActor") has really nothing to do with the changes I made...

cjcolvar:
I was able to fix the circular dependency issue locally by using `Rails.application.reloader.wrap` within the thread (thanks to [this issue](https://github.com/rails/rails/issues/30350) for the hint).  Here's a diff of what I did.  I could make a Hyrax PR with this or you could just include it in your PR whichever is easiest for you.

```
diff --git a/spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb b/spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb
index a33f2728a..e0984fdbd 100644
--- a/spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb
+++ b/spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb
@@ -43,8 +43,10 @@ RSpec.describe Hyrax::Actors::CreateWithRemoteFilesOrderedMembersActor do
       expect(ImportUrlJob).to receive(:perform_later).with(FileSet, null_operation, {}).exactly(4).times
       threads = environments.collect do |env|
         Thread.new do
-          expect(actor.create(env)).to be true
-          expect(env.curation_concern.ordered_members.to_a.collect(&:label)).to eq(filenames)
+          Rails.application.reloader.wrap do
+            expect(actor.create(env)).to be true
+            expect(env.curation_concern.ordered_members.to_a.collect(&:label)).to eq(filenames)
+          end
         end
       end
       threads.each(&:join)
```

j-dornbusch:
Thanks a lot @cjcolvar. I added your path to my PR (with a wrong commit message, sorry).
Now I am able to run the tests locally again: we had a problem with [this issue](https://github.com/samvera/hyrax/issues/3536) but we found a workaround.

j-dornbusch:
@cjcolvar My IDE said it: with your fix, the example is now too long.... Rubocop does not agree...

cjcolvar:
Yeah, I'm surprised that hasn't been fixed yet in master so thanks for writing it up! The known good workaround for now is upgrade engine_cart to 2.2.

j-dornbusch:
@cjcolvar I did not manage to use your fix. Maybe you could pass it yourself and I will rebase my branch?

cjcolvar:
I think those two expectations need to be together in the thread.  How about declaring the threads in a `let` then the body of the example becomes:
```
expect(ImportUrlJob).to receive(:perform_later).with(FileSet, null_operation, {}).exactly(4).times
threads.each(&:join)
```
That or you could make a rubocop exception or disable the cop inline around the test (like [this](https://github.com/samvera/hyrax/blob/master/spec/services/hyrax/adapters/nesting_index_adapter_spec.rb#L150-L183))

cjcolvar:
Sure, I could do that.

cjcolvar:
@j-dornbusch Don't worry about the one failing travis build since circleci is now the standard for mergeability.
I think this could be merged once it has been reviewed and doesn't need to be rebased with #3537 since the tests passed this time.
@jcoyne are you able to review this PR again?

config/locales/hyrax.de.yml
```diff
@@ -430,7 +427,16 @@ de:
         definitions_html: "<strong>Ansicht / Download: Auf</strong> diese Datei (sowohl Inhalte als auch Metadaten) kann innerhalb von %{application_name} zugegriffen werden. <br /> <strong>Bearbeiten:</strong> Diese Datei (sowohl Inhalte als auch Metadaten) kann bearbeitet werden. Sie dürfen diese Berechtigung nur Benutzern und / oder Gruppen von %{institution_name} erteilen."
         share_with_html: Sie können den Zugriff "Ansicht / Download" oder "Bearbeiten" für bestimmte Benutzer und / oder Gruppen auf Dateien gewähren. Geben Sie nacheinander einen gültigen %{account_name} ein, wählen Sie die Zugriffsebene für diesen Benutzer und klicken Sie auf + Hinzufügen.
       show:
+        items: Artikel
         last_modified: Zuletzt geändert
+        relationships: Beziehungen
+      show_actions:
+        analytics: Analytics
+        attach_child: Befestigen Kind
```
jmfrenzel: Please change this into :
"Untergeordnete Arbeit hinzufügen"

config/locales/hyrax.de.yml
```diff
@@ -430,7 +427,16 @@ de:
...
+        delete: Löschen
+        edit: Bearbeiten
+        feature: Feature
```
jmfrenzel: Please change this into :
"Hervorheben"

config/locales/hyrax.de.yml
```diff
@@ -430,7 +427,16 @@ de:
...
+        unfeature: Unfeature
```
jmfrenzel: Please change this into :
"Hervorhebung entfernen"

config/locales/hyrax.de.yml
```diff
@@ -430,7 +427,16 @@ de:
         definitions_html: "<strong>Ansicht / Download: Auf</strong> diese Datei (sowohl Inhalte als auch Metadaten) kann innerhalb von %{application_name} zugegriffen werden. <br /> <strong>Bearbeiten:</strong> Diese Datei (sowohl Inhalte als auch Metadaten) kann bearbeitet werden. Sie dürfen diese Berechtigung nur Benutzern und / oder Gruppen von %{institution_name} erteilen."
         share_with_html: Sie können den Zugriff "Ansicht / Download" oder "Bearbeiten" für bestimmte Benutzer und / oder Gruppen auf Dateien gewähren. Geben Sie nacheinander einen gültigen %{account_name} ein, wählen Sie die Zugriffsebene für diesen Benutzer und klicken Sie auf + Hinzufügen.
       show:
+        items: Artikel
```
jmfrenzel: Please change this into :
"Objekte"

config/locales/hyrax.de.yml
```diff
@@ -364,15 +364,10 @@ de:
           actions: Aktion
           title: Arbeitstitel
       form_files:
+        add_files: Hinzufügen von Dateien
```
jmfrenzel: Please change this into:
"Dateien hinzufügen"

j-dornbusch:
Sorry @jcoyne you just validated the PR but I had to integrate the improvements of @jmfrenzel

cjcolvar:
I reran the circleci build but it still failed on the unrelated test that is fixed by #3537 which is already in master.  I wouldn't hold up this PR over a rebase.

j-dornbusch:
@jcoyne @cjcolvar I rebased and cleaned history. I had to remove Rubocop "disable" directives from create_with_remote_files_ordered_members_actor_spec.rb.

j-dornbusch:
@cjcolvar I took exactly your version of `create_with_remote_files_ordered_members_actor_spec.rb` but circleci test failed although the PR doesn't change that file.

cjcolvar:
That's odd.  I'm not able to recreate that locally.  I'm going to try restarting the build.

j-dornbusch:
@cjcolvar  Restarting it solved the problem. There's something unclear, but the important thing is that it works at the end.

cjcolvar:
🎉 Thanks for all your work this PR!  It definitely had more tricky spots than a usual one.

Issue #3469, opened 2019-01-10
Javascript not functional on collection sharing tab 


j-dornbusch:
### Descriptive summary

Hyrax v3.0.0-beta1

In the collection 'sharing' tab, as we click the 'add' button to give a permission to a group or to a user, nothing happens, or the form is posted.
The share tabs work in the other contexts (Work, FileSet, Batch upload)

Looking at the code (`/app/views/hyrax/dashboard/collections/_form_share.html.erb` ), the ids of the form elements do not match those expected in es6 classes `UserControls` & `GroupControls` (`#new_user_name_skel`, etc.). 
And there's no js template at the bottom of the erb template like in app/views/hyrax/base/_form_share.html.erb for example. The 'tmpl-collection-grant'  indicated by [this line](https://github.com/samvera/hyrax/blob/b5c6605917c01c51b4504f7e27999146fc584713/app/assets/javascripts/hyrax/app.js.erb#L121) does not exist.
Compare search results for ['tmpl-work-grant'](https://github.com/samvera/hyrax/search?q=tmpl-work-grant&unscoped_q=tmpl-work-grant) and ['tmpl-collection-grant'](https://github.com/samvera/hyrax/search?q=tmpl-collection-grant&unscoped_q=tmpl-collection-grant)

### Expected behavior
The same behavior as the share tab for Works, FileSets and batch uploads.

### Actual behavior
The "add" button has no effect, or the form is posted, as if the application was working in accessible mode. 

### Steps to reproduce the behavior

1. Create a new Collection
2. Save it with the minimal mandatory fields
3. Go to the "sharing" tab
4 Try to add sharing with groups or users.

vantuyls:
The issue here is that the Add button does not give any indication that anything has happened, and that the table does not populate with the User or Group that was added. 

On Save, the person/group is added. 

Possible resolutions:
1) have the form populate the table with the user/group that was added to allow the user to see that the Add button has the desired effect
2) add a message upon selecting the Add button indicating that the user should Save for changes to go into effect. 

my preference is 1, but i'm not sure how challenging that is.

no-reply:
I can't reproduce this on Firefox 68.0.1.

rjkati:
I can't reproduce this in Chrome 87.0.4280.141 or Firefox 85.0 in Hyrax 3.0.0 rc3. It was present in Firefox 84.0.2.
@jlhardes

jlhardes:
[...]

PR #3543, opened 2019-02-19
present a confirmation when moving between tabs in collection edit.

### LINKED ISSUE #3454: Adding a user to collection removes unsaved changes in other tabs
### Descriptive summary

Version: Hyrax 3.0

When editing a collection, the user makes changes to the Branding and Discovery tabs, but does not click "Save changes".  If the user adds another user to the same collection via the Sharing tab, the page refreshes and the changes to the branding and discovery tabs are lost.  The user does not receive a warning that prompts them to save their changes when leaving the active tab.

### Rationale

The user should be alerted that their changes will not save when they navigate away from the active tab.

### Expected behavior
User should see a warning message before leaving the Branding and Discovery tabs that prompts them to click the "Save changes" button.

1. Navigate to an existing collection and click the "edit collection" button.
2. Upload a new collection banner, or change the discovery settings.  Do NOT click "Save Changes."
3. Click the "Sharing" tab and add a new user to the collection.
4. Navigate back to Step 2 and see if your changes are still present.

blancoj:
Fixes #3454;

If a user is editing a collection and makes some changes on one of the tabs and does not select the "Save changes" button  at the bottom of the pages, the user is at risk for losing his/her changes.  If he/she for example goes to the Sharing tab and starts selecting users and groups to share with, he/she will lose  the changes from previous tabs.  The safest thing to do is to "Save changes" on each tab and then move on to the next tab.

With this change a confirmation question is generated to remind the user to save his/her changes as the user switches from tab to tab.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/admin/collection_type/settings.es6
+++ b/app/assets/javascripts/hyrax/admin/collection_type/settings.es6
@@ -13,6 +13,15 @@ export default class {
     // Watch for changes to "sharable" checkbox
     $("#collection_type_sharable").on('change', () => { this.sharableChanged() })
     this.sharableChanged()
+
+    // Watch for changes to the tab, and alert user that changes need to be saved.
+    $('.panel-default li.coll-panel').on('click', function() {
+      if(!confirm('It is best to "Save changes" before moving to another tab, or your changes may be lost.  Do you want to change tab?')){
+        event.preventDefault();
+        return false;
+      }
+    })
+
   }
 
   // Based on the "sharable" checked/unchecked, enable/disable adjust share_applies_to_new_works checkbox
--- a/app/views/hyrax/dashboard/collections/_form.html.erb
+++ b/app/views/hyrax/dashboard/collections/_form.html.erb
@@ -1,21 +1,21 @@
 <div class="panel panel-default tabs" id="collection-edit-controls">
   <ul class="nav nav-tabs" role="tablist">
-    <li class="active">
+    <li class="coll-panel active">
       <a href="#description" role="tab" data-toggle="tab"><%= t('.tabs.description') %></a>
     </li>
     <% if @form.persisted? %>
       <% if @collection.brandable? %>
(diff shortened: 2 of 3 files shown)


no-reply:
@blancoj can you clarify which issue this targets? #3554 is a future number.

jrgriffiniii:
While being reviewed following the Samvera Tech. Call held earlier today, a question arose regarding test coverage.  Might it be possible to extend or introduce a new feature spec. for this new behavior please?

blancoj:
@no-reply Sorry about that.  I just edited my initial description.  The issue is 3454

j-dornbusch:
@blancoj The loss of information is only a side effect of https://github.com/samvera/hyrax/issues/3469. Users or groups selection in the sharing tab triggers unwanted page reloads because of a client-side bug. You cannot solve the problem by a warning when switching tabs because there are other symptoms: for example, the locale switches to English when you try to use the sharing feature.

j-dornbusch:
@cjcolvar @blancoj In my opinion, the actual bug is that 
> if he/she for example goes to the Sharing tab and starts selecting users and groups to share with, he/she will lose the changes from previous tabs

This problem does not occur on the work form because the javascript code is working and the page does not reload.
The actual cause of the bug is a missing template : The 'tmpl-collection-grant' called by [this line](https://github.com/samvera/hyrax/blob/b5c6605917c01c51b4504f7e27999146fc584713/app/assets/javascripts/hyrax/app.js.erb#L121) does not exist. As a consequence, javascript is not triggered and the page is submitted, resulting in the loss of data in the other tabs. 'tmp-collection-grant' plays a symmetrical role to that of 'tmpl-work-grant' in _form_share erb template, but weirdly it seems to have been forgotten. 
It would be better to fix the bug than to notify the user of its presence through a javascript dialog. 
Indeed, the bug has another effect for non-English-speaking users: the unexpected change of the locale.

PR #3544, opened 2019-02-20
Backport current CircleCI config to 2.x-stable

no-reply:
Since we have switched to CircleCI, we need to backport configuration to
maintained branches to keep the build working on them.

@samvera/hyrax-code-reviewers


no-reply:
@cjcolvar, @jrgriffiniii  i got that backport in, but it seems like things are still stuck on new chromedriver issues.

I'm thinking we should stop using chromedriver directly in a container via `selenium/standalone-chrome-debug` or similar?

Thoughts?

cjcolvar:
Does this need anything else backported (like #3552) or just a rebase?

PR #3547, opened 2019-02-20
Test with non-legacy image

cjcolvar:
Builds for hyrax master as well as 2.x-stable are failing due to a chromedriver issue even though nothing appears to have changed.  `chromedriver-helper`, `selenium-webdriver`, and the circleci docker image all look the same as when the build was passing.
This PR tries switching to the non-legacy circleci docker image to see if it works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,7 +113,7 @@ jobs:
 
   test:
     docker:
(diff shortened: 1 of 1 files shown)


cjcolvar:
Closed in favor of #3548.

PR #3548, opened 2019-02-20
Ensures that Chrome is manually updated on CircleCI build environments

jrgriffiniii:
Preinstalled releases began failing for certain builds, please see https://circleci.com/gh/samvera/hyrax/1135 for reference.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,11 @@ jobs:
     steps:
     - attach_workspace:
         at: ~/
-
+    # Update to the latest release of Chrome
+    - run: wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


cjcolvar:
@jrgriffiniii I saw that one failing test before.  It doesn't seem to be consistent and may or may not be related to the fix I put in for the same test but a different failure: https://github.com/samvera/hyrax/pull/3537
So I'm inclined to merge this since without it the build is consistently failing.
@no-reply what do you  think?

jrgriffiniii:
`rspec ./spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb:43` was the only failure, so I assume that it’s a flapping test, restarting the build.

jrgriffiniii:
The build passed with a second attempt, I've created https://github.com/samvera/hyrax/issues/3549 in response to the test failure I encountered.

no-reply:
Looks good to me.

It might be nice to figure out whether there's a way to avoid running the update with each build--maybe by maintaining a base image?

But for now. :rocket:

PR #3550, opened 2019-02-20
add rights notes metadata to work.

### LINKED ISSUE #3294: Add "Rights note" field
### Descriptive summary

Add free-text field labeled “Rights note.” Description: “Contains information about rights held in and over the resource.” Map to: http://purl.org/dc/elements/1.1/rights

### Rationale

While rightsstatements are a useful way of managing and displaying data about rights at a large scale, one ongoing concern which came up in the discussion of cultural heritage objects was the need to manage and display the _who_ as well as the what. An item might simultaneously have the rights statement that it is in copyright, a rights note about who holds the rights, and a CC-BY license (for which the rights note might be particularly important, as it could contain information about who should be credited).

WGBH and Tufts area already doing this locally, Oregon State in favor of change.

blancoj:
Fixes #3294

Added the "Rights notes" metadata to work.

Added the :rights_notes metadata to a work and placed it in the "Additional fields" section, joust below license.  You will also see this metadata (if it exists) in the work page.

Guidance for testing:

1. Create a work and be sure to enter something for Rights notes.
2.  Save the work.
3. From the home page hit "Go" and go to the work display page. Make sure the Rights notes you entered is in the page.
4. Edit the work.  Make sure you see the Rights notes metadata and then also change the metadata and make sure the change sticks.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/basic_metadata.rb
+++ b/app/models/concerns/hyrax/basic_metadata.rb
@@ -19,6 +19,8 @@ module Hyrax
       # Used for a license
       property :license, predicate: ::RDF::Vocab::DC.rights
 
+      property :rights_notes, predicate: ::RDF::URI.new('http://purl.org/dc/elements/1.1/rights'), multiple: true
+
       # This is for the rights statement
       property :rights_statement, predicate: ::RDF::Vocab::EDM.rights
       property :publisher, predicate: ::RDF::Vocab::DC11.publisher
--- a/app/models/concerns/hyrax/solr_document/metadata.rb
+++ b/app/models/concerns/hyrax/solr_document/metadata.rb
@@ -66,6 +66,7 @@ module Hyrax
         attribute :source, Solr::Array, solr_name('source')
         attribute :date_created, Solr::Array, solr_name('date_created')
         attribute :rights_statement, Solr::Array, solr_name('rights_statement')
+        attribute :rights_notes, Solr::Array, solr_name('rights_notes')
 
         attribute :mime_type, Solr::String, solr_name('mime_type', :stored_sortable)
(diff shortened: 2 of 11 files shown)


app/models/concerns/hyrax/solr_document/metadata.rb
```diff
@@ -66,6 +66,7 @@ def self.coerce(input)
         attribute :source, Solr::Array, solr_name('source')
         attribute :date_created, Solr::Array, solr_name('date_created')
         attribute :rights_statement, Solr::Array, solr_name('rights_statement')
+        attribute :rights_notes, Solr::Array, solr_name('rights_notes')
```
jeremyf: @blancoj I see that we store this as a Solr::Array. How does that relate or not with `multiple: false`?

app/models/concerns/hyrax/basic_metadata.rb
```diff
@@ -19,6 +19,8 @@ module BasicMetadata
       # Used for a license
       property :license, predicate: ::RDF::Vocab::DC.rights
 
+      property :rights_notes, predicate: ::RDF::URI.new('http://purl.org/dc/elements/1.1/rights'), multiple: false
```
jeremyf: @ruthtillman Would your expectation be that this should be `multiple: true`? In reading your comment (https://github.com/samvera/hyrax/issues/3294), it appears that this may need to be multi-valued.

ruthtillman: Yes, in some cases it may need to be multi-valued. It's a matter of best practice (probably single) vs. handling edge cases (almost certainly multi-valued).

PR #3553, opened 2019-02-21
[WIP] Uses the Docker selenium/standalone-chrome:latest image for feature tests on CircleCI

jrgriffiniii:
Following the example provided by @no-reply in https://gitlab.com/surfliner/surfliner/blob/master/starlight/docker/test/docker-compose.yml#L19-28, this attempts to integrate selenium/standalone-chrome:latest as a replacement for `chromedriver`.

Currently, this fails due to fact that the Selenium server in the Docker container needs to have access to the local files used during tests such as: https://github.com/samvera/hyrax/blob/master/spec/features/dashboard/collection_spec.rb#L675

Normally Docker can be started manually, and these fixture files could be copied using `docker cp`.  However, when this approach is taken, Docker over port 4444 cannot be accessed given CircleCI network restrictions: https://circleci.com/docs/2.0/building-docker-images/#accessing-services.  I attempted to use https://docs.docker.com/engine/reference/commandline/inspect/#get-an-instances-ip-address to find these, but the addresses did not appear to work.

When starting Docker images as additional executors (https://circleci.com/docs/2.0/configuration-reference/#docker), I could not find some way to determine the ability to connect to the container using `docker cp` (the docker daemon does not appear to be running unless `docker run` is explicitly called after `setup_remote_docker` (https://circleci.com/docs/2.0/building-docker-images/#overview)).

I suspect that there might be another approach I can attempt using docker-compose for all of this, but that requires that a separate image for running the rspec tests be created.  Given how much of this might have been encountered by others in the past, I was hoping to gain any insight into these obstacles.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,6 +124,7 @@ jobs:
         CATALINA_OPTS: "-Djava.awt.headless=true -Dfile.encoding=UTF-8 -server -Xms512m -Xmx1024m -XX:NewSize=256m -XX:MaxNewSize=256m -XX:PermSize=256m -XX:MaxPermSize=256m -XX:+DisableExplicitGC"
     - image: solr:7-alpine
       command: bin/solr -cloud -noprompt -f -p 8985
+    - image: selenium/standalone-chrome:latest
 
     working_directory: ~/hyrax
     parallelism: 4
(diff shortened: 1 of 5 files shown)


jrgriffiniii:
This should now pass on CircleCI.  It requires a rebase before merging, but I've left the commits in place for the review process (should there be any questions regarding any variations in this approach which I've attempted).

jrgriffiniii:
This is arguably blocked by the inability to set the branch from which the Hyrax code base is downloaded within the Docker container for the Selenium web server.  I've taken multiple attempts to resolve this on a separate branch, but none seem to be effective so far (please see https://github.com/samvera/hyrax/compare/circleci-standalone-chrome-jrgriffiniii...circleci-standalone-chrome-branches-jrgriffiniii and https://circleci.com/gh/samvera/hyrax/1337).

Using `docker-compose` seems likely to require a Hyrax repository container image to be built, as the `setup_remote_docker` only permits Docker containers communicate with each other in an isolated network (hence, Capybara won't be able to call Selenium by default unless it is running in a container by itself).

Falling back to a `machine` executor might enable a simpler approach, but I need to explore how this might impact parallelization.

PR #3554, opened 2019-02-22
input form has title as single entry, but the title def stays multiple

### LINKED ISSUE #3299: Limit dc:title to only occur once
### Descriptive summary

Require dc:title but limit its use to once with dc:alternative being used for additional titles.

### Rationale

Right now, dc:title may be used multiple times in Hyrax. Since there is no modification to add data about the language (which would be a good reason to repeat title) or control ordering, this may lead to confusion about the authoritative title. Limit dc:title to be used {1} times. Prefer dc:alternative for other forms of the title.

Right now this seems to be managed with: https://github.com/samvera/hyrax/blob/2.0-stable/app/models/concerns/hyrax/core_metadata.rb#L16-L18 but we were unsure if there's anyway for the metadata worker to engage with that.

blancoj:
Fixes #3299

Title will now be displayed as a single entry in the work form.  It remains defined as a multiple ( array ), but in the edit form, it is treated as a single.  In cases where a work has more than one title, the titles are sorted alphabetically and the 1st in the list is displayed on the edit form.  User collections and admin sets presently display only a single title, but because there is a lot of overlap in the form processing of works and collections, the work form and admin form processing had to change too.  

In cases of multiple titles, in order to store values not being displayed on the form, a new metadata type called alt_title was created.  This metadata is never displayed but is used during the processing of the form to store the titles not being displayed by the form and then when the form is saved, saving these titles back in the title metadata.

* Create a work, and see that you are only given one entry for title.  
* Enter the title, and remaining required fields and upload one file.
* Save the work.
* Verify that the correct title is  displayed on the work page
* Edit the work and change the title.
* Verify that the correct title is displayed on the work page.
* Repeat this for another type of work, User collection, and Admin Set.

Testing multiple titles for a work:

* Create a work, and see that you are only given one entry for title.  
* Enter the title, and remaining required fields and upload one file.
* Save the work.
* Using Rails console add a few more titles to the work.
* Verify that the correct title is  displayed on the work page
* Edit the work and notice that the 1st title based on the alphabetization of the titles is displayed.
* Change the title.
* Verify that the correct titles are displayed on the work page with the changed title and the other titles.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/admin_set_form.rb
+++ b/app/forms/hyrax/forms/admin_set_form.rb
@@ -7,6 +7,14 @@ module Hyrax
       # Cast any array values on the model to scalars.
       def [](key)
         return super if key == :thumbnail_id
+        if key == :title
+          @attributes["title"].each do |value|
+            @attributes["alt_title"] << value
+          end
+          @attributes["alt_title"].delete(@attributes["alt_title"].sort.first) unless @attributes["alt_title"].empty?
+          return @attributes["title"].sort.first unless @attributes["title"].empty?
+          return ""
+        end
         super.first
       end
 
--- a/app/forms/hyrax/forms/collection_form.rb
+++ b/app/forms/hyrax/forms/collection_form.rb
@@ -7,7 +7,7 @@ module Hyrax
       # Used by the search builder
       attr_reader :scope
 
-      delegate :id, :depositor, :permissions, :human_readable_type, :member_ids, :nestable?, to: :model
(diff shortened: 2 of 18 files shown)


j-dornbusch:
This change breaks our instance because we previously had a multivalued 'title' field (with language support) and another multivalued 'alternative_title' field (with language support) based on dc: alternative. From a documentary point of view, 'title' and 'alternative title' are two separate fields, and each one is likely to receive several values in case of translation of the content. So the mechanism that copies a field into the other will cause problems over time if increased support of the multilingual is introduced.
@samvera/hyrax-code-reviewers @blancoj

PR #3541, opened 2019-02-19
More i18n omissions 

j-dornbusch:
Extracts  strings from template to allow i18n.

@samvera/hyrax-code-reviewers

Issue #3549, opened 2019-02-20
CreateWithRemoteFilesOrderedMembersActor test is unstable


jrgriffiniii:
The following test has failed randomly on CircleCI: https://github.com/samvera/hyrax/blob/master/spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb#L43

After restarting a build, this passed without error.  Please see https://github.com/samvera/hyrax/pull/3548#issuecomment-465611983 for an example of this occurrence.

cjcolvar:
I'm stumped...I haven't been able to reproduce this locally and I don't have any ideas about why this error would be happening.  @mbklein @no-reply any ideas?
Should we just mark this test as [optional](https://github.com/samvera/hyrax/blob/5a9d1be16ee1a9150646384471992b03aab527a5/spec/support/optional_example.rb) so it can pass or fail without affecting the build?

mbklein:
No clue. What kind of failure was it? Since the build was re-run, I can't see the old log any more.

cjcolvar:
@mbklein There was an autoload issue which I addressed here: https://github.com/samvera/hyrax/pull/3537
But the failing test now has to do with a bad RDF namespace:
```
Failure/Error: fs.save!

Ldp::BadRequest:
  35b076c317f1e7mode:namespaces-http://scholarsphere.psu.edu/ns#
./app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb:72:in `block in create_file_from_url'
./app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb:68:in `new'
./app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb:68:in `create_file_from_url'
./app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb:59:in `block in attach_files'
./app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb:50:in `each'
./app/actors/hyrax/actors/create_with_remote_files_ordered_members_actor.rb:50:in `attach_files'
./app/actors/hyrax/actors/create_with_remote_files_actor.rb:16:in `create'
./spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb:48:in `block (6 levels) in <main>'
./spec/actors/hyrax/actors/create_with_remote_files_ordered_members_actor_spec.rb:47:in `block (5 levels) in <main>'
```
[...]

PR #3559, opened 2019-03-04
Use [* TO *] to work with old and new versions of AF and Solr

cjcolvar:
ActiveFedora 12.1.0 brings a new solr schema which switches from the deprecated solr.TrieDateField (to be removed in solr 8) to solr.DatePointField.  While the TrieDateField allowed queries like `date_dtsi:*`, DatePointField only allows single value and range queries so this example query needs to change to `date_dtsi:[* TO *]`.  Luckily the latter also works in TrieDateField.  This PR updates both the  embargo and lease search builders to use the `[* TO *]` query to work with both types of solr fields.

This fixes the tests that were blocking https://github.com/samvera/hyrax/pull/3541

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/embargo_search_builder.rb
+++ b/app/search_builders/hyrax/embargo_search_builder.rb
@@ -14,7 +14,7 @@ module Hyrax
 
     def only_active_embargoes(solr_params)
       solr_params[:fq] ||= []
-      solr_params[:fq] = 'embargo_release_date_dtsi:*'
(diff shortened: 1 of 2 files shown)


cjcolvar:
The circleci build has two failing tests which are a known issue https://github.com/samvera/hyrax/issues/3549

no-reply:
Will this require action for existing installations?

cjcolvar:
No, it shouldn't.

PR #3560, opened 2019-03-05
added reset.rake

jpengst:

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

* Added reset.rake
*
*

*
*
*

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/tasks/reset.rake
+++ b/lib/tasks/reset.rake
@@ -1,0 +1,29 @@
+namespace :hyrax do
+  namespace :reset do
+    desc 'Reset fedora / solr and corrisponding database tables w/o clearing other active record tables like users'
+    task works_and_collections: [:environment] do
+      confirm('You are about to delete all works and collections, this is not reversable!')
+      require 'active_fedora/cleaner'
+      ActiveFedora::Cleaner.clean!
+      Hyrax::PermissionTemplateAccess.delete_all
+      Hyrax::PermissionTemplate.delete_all
+
+      # we need to wait till Fedora is done with its cleanup
+      # otherwise creating the admin set will fail
+      while AdminSet.exists?(AdminSet::DEFAULT_ID)
+        puts 'waiting for delete to finish before reinitializing Fedora'
+        sleep 20
+      end
+      Rake::Task["hyrax:default_collection_types:create"].invoke
+      Rake::Task["hyrax:default_admin_set:create"].invoke
+    end
+
+    def confirm(action)
+      return if ENV['RESET_CONFIRMED'].present?
+      confirm_token = rand(36**6).to_s(36)
+      STDOUT.puts "#{action} Enter '#{confirm_token}' to confirm:"
+      input = STDIN.gets.chomp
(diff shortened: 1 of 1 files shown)


lib/tasks/reset.rake
```diff
@@ -0,0 +1,30 @@
+namespace :hyrax do
+  namespace :reset do
+    desc 'Reset fedora / solr and corrisponding database tables w/o clearing other active record tables like users'
+    task works_and_collections: [:environment] do
+      confirm('You are about to delete all works and collections, this is not reversable!')
+      require 'active_fedora/cleaner'
+      ActiveFedora::Cleaner.clean!
+      Hyrax::PermissionTemplateAccess.delete_all
+      Hyrax::PermissionTemplate.delete_all
+      Delayed::Job.delete_all
```
jcoyne: Delayed::Job is not a dependency of Hyrax.

lib/tasks/reset.rake
```diff
@@ -0,0 +1,30 @@
...
+
+      # we need to wait till Fedora is done with its cleanup
+      # otherwise creating the admin set will fail
+      while AdminSet.exists?(AdminSet::DEFAULT_ID)
```
jcoyne: ActiveFedora::Cleaner is synchronous.  There shouldn't be a need for this.

no-reply:
Hi @jpengst!

Thanks for taking the time to make this contribution. This looks like something contributed back from one of your Hyrax appplications(?), and as @jcoyne has pointed out there are a few generalization issues.

It might be helpful to start by squashing the existing changes into a commit with a message documenting what this task is useful for, and how you arrived at considerations like the `while/sleep` loop. I like to point folks to https://chris.beams.io/posts/git-commit/ for guidance about what kinds of things to consider in writing commit messages.

Thanks again! I expect this would be a handy task for a lot of folks, so I'm eager to get this one ready to merge.

lib/tasks/reset.rake
orangewolf: good point - @jpengst I forgot to take this out of the copy I sent you. Unfortunately there isn't a good generic way to clear all jobs in the queue, so we should remove this and we'll make a separate task that depends on this one in our various projects.

lib/tasks/reset.rake
orangewolf: @jcoyne we haven't experienced it to be so. The prevailing theory was that it was async because we've seen multiple instances of the admin set creation failing after calling the cleaner. If you, for example, import 100 works, then run this task w/o the while, it will typically fail. If you run the task a second time it will succeed. Are there any other reasons it would be like that?

jpengst:
@no-reply @orangewolf and I squashed the commits and updated message. Looks like @jcoyne has not had a chance to get back on the wait question. Do we want to leave the wait commented out in the rake task if people need it so that this can be merged?

cjcolvar:
@jpengst I'm not sure why your build has two failing tests.  Do you mind rebasing your PR on the latest master branch? I think this might clear up the failing tests.

PR #3563, opened 2019-03-10
[wings] Make Valkyrie a production dependency

no-reply:
We'll start introducing Valkyrie calls into production code shortly, so we need
to make it a production dependency.

Unfortunately, our dependency needs to be 2.0.0+, since the 1.x series has a
hard dependency on `pg`. Passing that dependency along to our users isn't
acceptable; it would make Postgresql a system dependency for every Hyrax
installation. For this reason, we pin to 2.0.0-rc3+.

Since we are ourselves in a beta cycle, this should be okay as long as an actual
2.0.0 release is forthcoming on a short timeline.

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -79,6 +79,7 @@ SUMMARY
   spec.add_dependency 'select2-rails', '~> 3.5'
   spec.add_dependency 'signet'
   spec.add_dependency 'tinymce-rails', '~> 4.1'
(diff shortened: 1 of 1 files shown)


no-reply:
@tpendragon is it safe to assume we'll see a valkyrie 2.0.0 release on a time-scale counted in single digit weeks?

tpendragon:
@no-reply I'll...have to think about that. Ping me on Slack so I don't forget.

PR #3566, opened 2019-03-11
Add support for Ruby 2.6; drop 2.3

no-reply:
Adds support for Ruby 2.6, which is in the build matrix as of #3565.

Ruby 2.3 falls out of its security maintenance phase on the 31st of this
month. That will likely predate the 3.0.0 release, meaning support should be
dropped in that release. This removes it in anticipation of that.

This is really just a README update, documenting official support policy.

@samvera/hyrax-code-reviewers


cjcolvar:
@no-reply Github is waiting on a `test-2-6` circleci build but it looks like only `test` and `test-ruby-2-5` are defined in the circleci config.  Should we change the name in the circleci config or switch the github status checks for the master branch?

no-reply:
I don't know why GitHub's status checks never seem to work properly. :|

I've just merged over its objections.

PR #3582, opened 2019-03-11
Run i18n normalization & automatic translations for 3.0.0-beta2 

no-reply:
Completed following the process at
https://github.com/samvera/hyrax/wiki/Translations-(internationalization)#update-translations

@samvera/hyrax-code-reviewers

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ gemspec
 
 group :development, :test do
   gem 'coveralls', require: false
+  gem 'easy_translate'
   gem 'i18n-tasks'
   gem 'pry' unless ENV['CI']
   gem 'pry-byebug' unless ENV['CI']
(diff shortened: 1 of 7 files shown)


Gemfile
```diff
@@ -5,6 +5,7 @@ gemspec
 
 group :development, :test do
   gem 'coveralls', require: false
+  gem 'easy_translate'
```
jcoyne: Can you add documentation on what this is for and how you use it?

no-reply: Does the commit message for 579eb89 satisfy this?

no-reply: i.e. there are no changes, except that this dependency used to be added by `i18n-tools`, and now it is not

jcoyne: Yes, sorry, I missed that because of the folding.

Issue #3569, opened 2019-03-11
BaseActor rework to decouple tight dependency on ActiveFedora


revgum:
### Descriptive summary

BaseActor needs some rework its persistence calls to use Wings instead of direct ActiveFedora calls.

https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/base_actor.rb

### Related work

hyrax-on-wings addressed this, in a direct way on the GenericWorkActor. This might translate/inform the work that needs to be done on the BaseActor. 

https://github.com/samvera-labs/hyrax-on-wings/blob/master/app/actors/hyrax/actors/generic_work_actor.rb

Link to related tickets or prior related work here.

- [ ] #3572 
- [ ] #3575

PR #3584, opened 2019-03-11
Wings base actor

### LINKED ISSUE #3572: BaseActor replace #save with Valkyrie
### Descriptive summary

Use Valkyrie and the related `valkyrie_metadata_adapter` configuration to save `curation_concern` as a valkyrie resource, and return true if the resource was created.

### Rationale

This should replace the behavior of ActiveFedora `#save` call on a curation concern object.

### Related work

See #3569

revgum:
Fixes #3572 ; refs #3569 

Temporarily adding `require 'wings'` to get Wings into the mix for testing and functionality.

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -1,3 +1,5 @@
+require 'wings'
+
 module Hyrax
   module Actors
     ##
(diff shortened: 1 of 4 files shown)


revgum:
replaced by #3678

Issue #3581, opened 2019-03-11
Implement find_all in wings query_service


hackmastera:
child of #3570

PR #3586, opened 2019-03-11
Implement "find_all" for Wings::Valkyrie::QueryService

laritakr:
Fixes https://github.com/samvera/hyrax/issues/3581

In order to use the Valkyrie shared specs, the Active Fedora adapter needs to support all of the required queries.

This PR implements the `find_all` query.

### DIFF
--- a/lib/wings/valkyrie/query_service.rb
+++ b/lib/wings/valkyrie/query_service.rb
@@ -24,6 +24,18 @@ module Wings
         raise ::Valkyrie::Persistence::ObjectNotFoundError
       end
 
+      # Find all work/collection records, and map to Valkyrie Resources
+      # @return [Array<Valkyrie::Resource>]
+      def find_all
+        klasses = Hyrax.config.curation_concerns.append(::Collection)
+        objects = ::ActiveFedora::Base.all.select do |object|
+          klasses.include? object.class
+        end
+        objects.map do |id|
+          resource_factory.to_resource(object: id)
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


lib/wings/valkyrie/query_service.rb
```diff
@@ -24,6 +24,18 @@ def find_by(id:)
         raise ::Valkyrie::Persistence::ObjectNotFoundError
       end
 
+      # Find all work/collection records, and map to Valkyrie Resources
+      # @return [Array<Valkyrie::Resource>]
+      def find_all
+        begin
```
no-reply: is the `begin` block needed?

no-reply:
Could we also update the commit message to say a bit more about the context?

I often have to create release notes. Nicely formed and contextualizing commit titles make a big difference in that process.

lib/wings/valkyrie/query_service.rb
laritakr: This was needed to get it to pass rubocop. Perhaps there is a better way to pacify rubocop?

spec/wings/valkyrie/query_service_spec.rb
```diff
@@ -50,6 +51,20 @@ class Book < ActiveFedora::Base
     end
   end
 
+  describe ".find_all", clean_repo: true do
+    before do
+      allow(Hyrax.config).to receive(:curation_concerns).and_return(Hyrax.config.curation_concerns.append(::Collection).append(af_resource_class))
```
no-reply: I'm not sure i'm following this stub, can you explain?

PR #3617, opened 2019-03-12
Add find_many_by_ids to wings query_service

### LINKED ISSUE #3614: implement find_many_by_ids in wings query service
child of #3570

hackmastera:
This is a naive implementation, which valkyrie uses in its fedora
adapter. Another implementation was explored, using https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/relation/finder_methods.rb#L94, but that returns SolrHit objects which we believe we'd need to call ActiveFedora.find over, anyway.

fixes #3614

### DIFF
--- a/lib/wings/valkyrie/query_service.rb
+++ b/lib/wings/valkyrie/query_service.rb
@@ -53,14 +53,14 @@ module Wings
       # @param [Array<Valkyrie::ID, String>] ids
       # @return [Array<Valkyrie::Resource>]
       def find_many_by_ids(ids:)
-        ids = ids.uniq
-        ids.map do |id|
-          begin
-            find_by(id: id)
-          rescue ::Valkyrie::Persistence::ObjectNotFoundError
-            nil
-          end
-        end.compact
+        ids.each do |id|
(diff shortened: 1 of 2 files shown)


lib/wings/valkyrie/query_service.rb
```diff
@@ -36,6 +36,17 @@ def find_all
...
 
+      def find_many_by_ids(ids:)
+        ids = ids.uniq
+        ids.map do |id|
+          begin
+            find_by(id: id)
+          rescue ::Valkyrie::Persistence::ObjectNotFoundError
+            nil
+          end
+        end.reject(&:nil?)
```
no-reply: `.compact` might be nicer?

hackmastera:
@no-reply I added a new commit that uses `ActiveFedora::Base.where`; because `ActiveFedora::Base.find` with multiple ids throws an error if one of the ids isn't found, which is against the Valkyrie API. If you want to mess around with other implementations, feel free. I added this new option as a new commit so it can easily rolled back if you prefer the naive implementation.

PR #3624, opened 2019-03-13
implement #objects from pcdm_behavior

### LINKED ISSUE #3585: PCDM::Collection/Object - #objects method using AF relationship reflection #members
### Code

lib/hydra/pcdm/models/concerns/pcdm_behavior.rb #objects

### Active Fedora Usage

* calls #members method created when defining the ActiveFedora relationship named members using the AF #ordered_aggregation method

#### May also have to address:

* #object_ids which calls #objects

### Related issues

* #3571 PCDM::Collection/Object - members relationship defined through AF ordered_aggregation
* #3595 PCDM::Collection - #collections method using AF relationship reflection #members
* #3597 PCDM::Collection - #collection_ids method using AF relationship reflection #members
* #3600 PCDM::AncestorChecker - #former_is_ancestor_of_latter? method using AF relationship reflection #members

elrayle:
returns either AF objects or Valkyrie resources

Fixes #3585

Implements #objects under wings such that it will return ActiveFedora objects (default) or Valkyrie resources.  This continues to use ActiveFedora to get the set of objects, so more work is required for this to be fully valkyrized.

### Usage:

```
af_objects = valkyrie_resource.objects
valkyrie_resources = valkyrie_resource.objects(valkyrie: true)
```

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -44,7 +44,10 @@ module Wings
     ##
     # @return [ActiveFedora::Base]
     def convert
-      resource.internal_resource.new(attributes).tap { |obj| obj.id = id unless id.empty? }
+      resource.internal_resource.new(attributes).tap do |obj|
+        obj.id = id unless id.empty?
+        resource.member_ids.each { |valkyrie_id| obj.members << ActiveFedora::Base.find(valkyrie_id.id) } if resource.respond_to? :member_ids
+      end
     end
 
     ##
--- a/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
+++ b/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
@@ -1,0 +1,15 @@
+require 'wings/active_fedora_converter'
+
+module Wings
+  module PcdmValkyrieBehavior
+    ##
+    # Gives the subset of #members that are PCDM objects
+    # @return [Enumerable<PCDM::ObjectBehavior>] an enumerable over the members
+    #   that are PCDM objects
+    def objects(valkyrie: false)
+      af_objects = Wings::ActiveFedoraConverter.new(resource: self).convert.objects
+      return af_objects unless valkyrie
+      af_objects.map(&:valkyrie_resource)
+    end
+  end
+end
--- a/lib/wings/model_transformer.rb
+++ b/lib/wings/model_transformer.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'wings/value_mapper'
+require 'wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior'
 
 module Wings
   #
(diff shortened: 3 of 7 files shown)


lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
```diff
@@ -0,0 +1,23 @@
...
+  module PcdmValkyrieBehavior
+    ##
+    # Gives the subset of #members that are PCDM objects
+    # @return [Enumerable<PCDM::ObjectBehavior>] an enumerable over the members
+    #   that are PCDM objects
+    def objects(valkyrie: false)
+      af_objects = Wings::ActiveFedoraConverter.new(resource: self).convert.objects
+      return af_objects unless valkyrie
+
+      # return Valkyrie Resources
```
revgum: My only question here is, since `af_objects` are ActiveFedora::Base objects, then they should already be "Valkyrizable" and should respond to `#valkyrie_resource`, is that correct? 

Could it be  that the following block is replaced with `af_objects.map(&:valkyrie_resource)`?

elrayle: The GenericWork class did not respond to #valkyrie_resource which is why I added the include.

spec/wings/model_transformer_spec.rb
```diff
@@ -172,7 +172,7 @@
           .to have_attributes title: book.title,
                               contributor: book.contributor,
                               description: book.description
-        expect_ids_to_match(subject.build.page_ids, ['pg1', 'pg2'])
+        expect(subject.build.page_ids).to match_valkyrie_ids_with_active_fedora_ids(['pg1', 'pg2'])
```
elrayle: Moved to be custom matcher in `wings/support/matchers` so it can be used across spec files.

spec/wings_helper.rb
```diff
@@ -0,0 +1,4 @@
+require 'spec_helper'
+require 'wings'
+
+Dir[File.expand_path(File.join(File.dirname(__FILE__), 'wings', 'support', 'matchers', '**', '*.rb'))].each { |f| require f }
```
elrayle: Brings in custom matchers.

lib/wings/model_transformer.rb
```diff
@@ -81,6 +82,8 @@ def self.convert_class_name_to_valkyrie_resource_class(class_name)
     #   results in long methods
     def self.to_valkyrie_resource_class(klass:)
       Class.new(ActiveFedoraResource) do
+        include Wings::PcdmValkyrieBehavior
```
elrayle: This brings in the replacement code for #objects into the newly create valkyrie resource class.

lib/wings/active_fedora_converter.rb
```diff
@@ -44,7 +44,10 @@ def attributes
     ##
     # @return [ActiveFedora::Base]
     def convert
-      resource.internal_resource.new(attributes).tap { |obj| obj.id = id unless id.empty? }
+      resource.internal_resource.new(attributes).tap do |obj|
+        obj.id = id unless id.empty?
+        resource.member_ids.each { |valkyrie_id| obj.members << ActiveFedora::Base.find(valkyrie_id.id) } if resource.respond_to? :member_ids
```
elrayle: Addresses need for round-trip conversion of members relationship.

PR #3630, opened 2019-03-13
implement #object_ids from pcdm_behavior

elrayle:
refs #3585

Implements #object_ids under wings such that it will return ActiveFedora object ids (default) or Valkyrie resource ids.  This continues to use ActiveFedora to get the set of objects, so more work is required for this to be fully valkyrized.

### Usage:

```
af_object_ids = valkyrie_resource.object_ids
valkyrie_resource_ids = valkyrie_resource.object_ids(valkyrie: true)
```

### DIFF
--- a/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
+++ b/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
@@ -4,12 +4,19 @@ module Wings
   module PcdmValkyrieBehavior
     ##
     # Gives the subset of #members that are PCDM objects
-    # @return [Enumerable<PCDM::ObjectBehavior>] an enumerable over the members
+    # @return [Enumerable<ActiveFedora::Base> | Enumerable<Valkyrie::Resource>] an enumerable over the members
     #   that are PCDM objects
     def objects(valkyrie: false)
       af_objects = Wings::ActiveFedoraConverter.new(resource: self).convert.objects
       return af_objects unless valkyrie
       af_objects.map(&:valkyrie_resource)
+    end
+
+    ##
+    # Gives a subset of #member_ids, where all elements are PCDM objects.
+    # @return [Enumerable<String> | Enumerable<Valkyrie::ID] the object ids
+    def object_ids(valkyrie: false)
+      objects(valkyrie: valkyrie).map(&:id)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior_spec.rb
```diff
@@ -5,46 +5,75 @@
 RSpec.describe Wings::PcdmValkyrieBehavior do
   subject(:factory) { Wings::ModelTransformer.new(pcdm_object: pcdm_object) }
 
-  context 'for PCDM & Hydra-Works models', :clean_repo do
```
elrayle: I got rid of the context which would apply to all tests in this file.  It was a carry over from starting these tests in model_transformer_spec.  This cause the indentation to change which makes it look like more changed than it did.  Starting here, this is only an indentation change.

spec/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior_spec.rb
```diff
@@ -5,46 +5,75 @@
...
+    end
+    context 'when return type is not specified' do
+      it 'returns works only as fedora objects through pcdm_valkyrie_behavior' do
+        af_objects = subject.build.objects
+        expect(af_objects.size).to eq 2
+        expect(af_objects.first.pcdm_object?).to be true
+        expect(af_objects.map(&:id)).to match_array [work1.id, work2.id]
       end
+    end
+  end
```
elrayle: Indentation change ends here.  The rest of this file is the new tests added for `#object_ids`.

lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
```diff
@@ -4,12 +4,20 @@ module Wings
...
       af_objects = Wings::ActiveFedoraConverter.new(resource: self).convert.objects
       return af_objects unless valkyrie
       af_objects.map(&:valkyrie_resource)
     end
+
+    ##
+    # Gives a subset of #member_ids, where all elements are PCDM objects.
+    # @return [Enumerable<String> | Enumerable<Valkyrie::ID] the object ids
+    def object_ids(valkyrie: false)
+      return objects.map(&:id) unless valkyrie
```
revgum: I think these could be combined to just `objects(valkyrie).map(&:id)`

Issue #3571, opened 2019-03-11
PCDM::Collection/Object - members relationship defined through AF ordered_aggregation


elrayle:
### Code

lib/hydra/pcdm/models/concerns/pcdm_behavior.rb

### Model Usage

* PCDM::Collection (and by extension Works::Collection)
* PCDM::Object (and by extension Works::Work, Works::FileSet)

### Active Fedora Usage

* calls AF ordered_aggregation method to create the members relationship.  
* passes `'ActiveFedora::Base'` as the class_name

#### May also have to address:

* uses `type_validator` for validation.  This is defined in PCDM, so it may be ok as is.  Don't know if there is equivalent validation code in Valkyrie that this validator will work with.

elrayle:
PR #3636 - adds the members relationship
Issue #3676 - explores adding to members using `<<` and `+=`

PR #3636, opened 2019-03-13
define member_ids as ordered

elrayle:
refs #3571

Creates members_id relationship as an ordered Array of ids.  Uses #objects method for #members and #ordered_members.  All of these will return the members in order because member_ids is ordered.

TODO: The question is how to handle…

```
obj1.members << obj2
obj1.members += obj3
obj1.ordered_members << obj4
obj1.ordered_members += obj5
```


spec/wings/model_transformer_spec.rb
```diff
@@ -189,110 +189,4 @@
       end
     end
   end
-
-  context 'for PCDM & Hydra-Works models', :clean_repo do
```
elrayle: moved to pcdm_valkyrie_behavior_spec.rb where member_id is now created

PR #3645, opened 2019-03-14
Integrating changes into Hyrax 3.0.0 for supporting BrowseEverything 1.0.x releases

jrgriffiniii:
This addresses the following in order to support integrating 1.0.0 and later 1.0.x releases of [browse-everything:](https://github.com/samvera/browse-everything)
- Ensures that HTTP headers are passed in requests to determine if cloud resources are available
- Ensures that HTTP GET request parameters are decoded when requesting cloud resources
- Ensures that FileSet labels are used when ingesting files which are temporarily downloaded from cloud service providers (this approach is currently used by at least one browse-everything driver)
- Ensures that the Rails `tmp` directory is whitelisted for file uploads (this is where the aforementioned downloaded files are placed)
- Ensures that the `browse_everything:config` generator is run for new Hyrax instances (otherwise errors are raised for a missing BE configuration file)

Please note that this is a WIP given that https://github.com/samvera/browse-everything/pull/270/files should be merged (and a 1.0.1 version released) in order to handle errors raised when the directory for the File System driver does not exist.

### DIFF
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -21,17 +21,19 @@ class ImportUrlJob < Hyrax::ApplicationJob
     operation.performing!
     user = User.find_by_user_key(file_set.depositor)
     uri = URI(file_set.import_url)
+    name = file_set.label
+
     @file_set = file_set
     @operation = operation
 
-    unless can_retrieve?(uri)
+    unless BrowseEverything::Retriever.can_retrieve?(uri, headers)
       send_error('Expired URL')
       return false
     end
 
     # @todo Use Hydra::Works::AddExternalFileToFileSet instead of manually
     #       copying the file here. This will be gnarly.
-    copy_remote_file(uri, headers) do |f|
+    copy_remote_file(uri, name, headers) do |f|
       # reload the FileSet once the data is copied since this is a long running task
       file_set.reload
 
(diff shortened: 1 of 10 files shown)


jrgriffiniii:
This is still blocked until https://github.com/samvera/browse-everything/pull/270/files is reviewed and merged upstream.

jrgriffiniii:
Following the release of `browse-everything` version 1.0.1 (https://github.com/samvera/browse-everything/pull/272), this should now be using the proper dependency.

jrgriffiniii:
I can definitely improve the tests introduced for these changes, thank you for the review @cjcolvar

app/jobs/import_url_job.rb
```diff
@@ -92,7 +85,7 @@ def send_error(error_message)
     # @param f [IO] the stream to write to
     def write_file(uri, f, headers)
       retriever = BrowseEverything::Retriever.new
-      uri_spec = { 'url' => uri }.merge(headers)
+      uri_spec = ActiveSupport::HashWithIndifferentAccess.new(url: uri, headers: headers)
```
no-reply: Does the browse everything `Retriever` need this to have indifferent access? or can it do alright with `String` keys?

cjcolvar: It looks like BE only expects string keys as far as I can tell.

jrgriffiniii: `Typhoeus::Request` seemed to prefer symbols for keys (please see https://github.com/samvera/browse-everything/blob/master/lib/browse_everything/retriever.rb#L39, https://github.com/samvera/browse-everything/blob/master/lib/browse_everything/retriever.rb#L138 and https://github.com/typhoeus/typhoeus#introduction), and there were some legacy areas of the code base where strings were still used (https://github.com/samvera/browse-everything/blob/master/lib/browse_everything/retriever.rb#L95).  I can address this with a patch release for `browse-everything` (https://github.com/samvera/browse-everything/issues/290) and issue a separate PR resolving this.

PR #3655, opened 2019-03-15
enforces only to use wings adapter for now

### LINKED ISSUE #3654: Hyrax configuration for wings metadata adapter
### Descriptive summary

Raise exception in the `setter` related to Wings metadata adapter so that the app fails if the value is anything but `:wings_adapter` to notify the user that Hyraxes `valkryie_metadata_adapter` is currently only supporting the wings.

revgum:
Fixes #3654 

Circle CI configuration change notes;
- Originally if any files in the PR changed `hyrax/generators` then the internal test app wouldn't be generated. Changed to be the inverse, so that the internal test app is forced to be rebuilt when any generator file had changes. 
- Because Circle CI is configured to cache the internal test app directory, it should be restored and functional across test suite runs until a PR includes changes to a `hyrax/generators` file.

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -227,9 +227,13 @@ module Hyrax
     # @see lib/wings
     # @see https://github.com/samvera-labs/valkyrie
     def valkyrie_metadata_adapter
-      Valkyrie::MetadataAdapter.find(@valkyrie_metadata_adapter || :wings_adapter)
+      Valkyrie::MetadataAdapter.find(@valkyrie_metadata_adapter)
     end
-    attr_writer :valkyrie_metadata_adapter
+
+    def valkyrie_metadata_adapter=(adapter)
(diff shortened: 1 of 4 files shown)


lib/hyrax/configuration.rb
```diff
@@ -227,9 +227,13 @@ def curation_concerns
     # @see lib/wings
     # @see https://github.com/samvera-labs/valkyrie
     def valkyrie_metadata_adapter
-      Valkyrie::MetadataAdapter.find(@valkyrie_metadata_adapter || :wings_adapter)
+      Valkyrie::MetadataAdapter.find(@valkyrie_metadata_adapter)
```
no-reply: looks like we still need the default, here.

We could (should?) also get a default into the generator for the config. But either way, we'd need this to support apps that generated their config with an older version,

revgum:
@no-reply ping

PR #3659, opened 2019-03-18
Wings persister saves a native valkyrie resource

### LINKED ISSUE #3629: Wings persister needs to be able to save a Valkyrie resource that was never an AF resource
Shared specs in valkyrie persister use straight valkyrie resources. They'll never pass if we rely on using a valkyrie resource that was created / converted from an AF resource.

Chid of #3627

hackmastera:
Adds all current shared specs from valkyrie, marks pending those that do
not pass. Removed ones that were pending upstream (i.e. not supported in
valkyrie adapters). Removed deep nesting and mixed nesting.

Sketches out the beginning of a "Default" active fedora object for use
with an unrecognized Valkyrie Resource

Resolves #3629 but includes lots of pending tests. Not sure if that is a blocker for merge.

### DIFF
--- a/.rubocop_fixme.yml
+++ b/.rubocop_fixme.yml
@@ -129,7 +129,7 @@ RSpec/ExampleLength:
     - 'spec/services/hyrax/workflow/state_machine_generator_spec.rb'
     - 'spec/services/hyrax/workflow/workflow_importer_spec.rb'
     - 'spec/views/**/*'
-
+    - 'spec/wings/valkyrie/persister_spec.rb'
 RSpec/VerifiedDoubles:
   Enabled: false
 
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -17,7 +17,6 @@ require 'hyrax/engine'
 require 'hyrax/version'
 require 'hyrax/inflections'
 require 'kaminari_route_prefix'
-require 'wings'
 
 module Hyrax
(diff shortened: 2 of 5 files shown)


.rubocop_fixme.yml
```diff
@@ -129,7 +129,7 @@ RSpec/ExampleLength:
     - 'spec/services/hyrax/workflow/state_machine_generator_spec.rb'
     - 'spec/services/hyrax/workflow/workflow_importer_spec.rb'
     - 'spec/views/**/*'
-
+    - 'spec/wings/valkyrie/persister_spec.rb'
```
no-reply: do we plan to fix this as this work develops, or does this belong in a different .rubocop config?

hackmastera: Well, eventually we will remove all the triggering tests, because they'll be replaced with the call to the shared spec.

But mostly I put it here because in my experience you can't exclude in both `_todo` and `.rubocop.yml` for the same infraction; they just all go red. Granted I've never tried it in this codebase.

elrayle:
I reviewed the code and found nothing in these code changes that could account for the failing test related to branding of a collection.  I also tested through the UI by running the Hyrax app and following steps through the UI that matched those taken in the failing test.  The branding files were correctly saved as expected and they were shown on the collection public page.  I kicked the tests again and they all pass.

PR #3667, opened 2019-03-18
Prep for a Hyrax 2.5.0 release

bess:
This is to prep for a 2.5.0 release. I have already talked to @no-reply about this.
Other feedback welcome.

Thanks!

@samvera/hyrax-code-reviewers


no-reply:
it seems like we're missing a backport of the ruby 2.5/.6 test config.

Merging this over that objection

PR #3678, opened 2019-03-18
[wings] Make Embargo, Lease, and Visibility work for `BaseActor#create`

no-reply:
This reduces us to just a few blocking issues before we can do a `BaseActor#create` swap out.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -18,7 +18,10 @@ module Hyrax
       def create(env)
         apply_creation_data_to_curation_concern(env)
         apply_save_data_to_curation_concern(env)
-        save(env) && next_actor.create(env) && run_callbacks(:after_create_concern, env)
+
+        save(env, use_valkyrie: false) &&
+          next_actor.create(env) &&
+          run_callbacks(:after_create_concern, env)
       end
 
       # @param [Hyrax::Actors::Environment] env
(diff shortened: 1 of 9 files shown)


no-reply:
@elrayle e0ce5f2 looks like it's missing some unit tests.

What do you think about dropping it from the PR, reviewing this as-is, and including the collections in follow-up work?

revgum:
fixes #3572

elrayle:
I created a separate PR for the member_of_collections so it could go ahead and get merged.  See PR #3681.

no-reply:
@elrayle with that in mind, would it be okay to drop it from this one? I think the added commit is blocking merge off this right now.

elrayle:
I don't mind if you drop it.  It is a full PR with tests in the other PR.

For consideration, I know these tests don't fail as written, but when I set `use_valkyrie: true` for #save in the #create method, I get the same error across many tests.  Were you hoping to address those before merging this?

Example of one for `rspec ./spec/actors/hyrax/actors/initialize_workflow_actor_spec.rb:29 # Hyrax::Actors::InitializeWorkflowActor create creates an entity`

```
  1) Hyrax::Actors::InitializeWorkflowActor create creates an entity
     Failure/Error:
       Sipity::Workflow.where(active: true).where(
         workflows[:permission_template_id].in(
           templates.project(templates[:id]).where(templates[:source_id].eq(admin_set_id))
         )
       ).first!
     
     ActiveRecord::RecordNotFound:
       Couldn't find Sipity::Workflow with [WHERE "sipity_workflows"."active" = ? AND "sipity_workflows"."permission_template_id" IN (SELECT "permission_templates"."id" FROM "permission_templates" WHERE "permission_templates"."source_id" IS NULL)]
```

no-reply:
> Were you hoping to address those before merging this?
I wasn't. The guard is in place to avoid running that code.

The hope is to merge this so we can continue to work on the remainder of the issues preventing this swap-out.

elrayle:
Seems good to yank the last commit for member_of_collections and get this merged.

no-reply:
@revgum can you re-review?

PR #3681, opened 2019-03-19
convert valkyrie member_of_collection_ids to AF member_of_collections objects

elrayle:
Fixes an issue where member_of_collections was converted to the valkyrie resource but not back to the af_object.


elrayle:
@no-reply Per earlier conversation, I am going ahead and merging this with the assumption that a more efficient approach will replace this in the near future.

Issue #3622, opened 2019-03-12
Implement find_members in wings query service


laritakr:
child of #3570

PR #3682, opened 2019-03-19
Implement find_members in query_service

laritakr:
Fixes https://github.com/samvera/hyrax/issues/3622

Implement `find_members` in query service, using PCDM method `members` on the active_fedora object.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/valkyrie/query_service.rb
+++ b/lib/wings/valkyrie/query_service.rb
@@ -77,6 +77,24 @@ module Wings
         raise ::Valkyrie::Persistence::ObjectNotFoundError
       end
 
+      # Find all members of a given resource, and map to Valkyrie Resources
+      # @param [Valkyrie::Resource]
+      # @param [Valkyrie::ResourceClass]
+      # @return [Array<Valkyrie::Resource>]
+      def find_members(resource:, model: nil)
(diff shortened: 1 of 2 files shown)


lib/wings/valkyrie/query_service.rb
```diff
@@ -77,6 +77,24 @@ def find_by_alternate_identifier(alternate_identifier:)
         raise ::Valkyrie::Persistence::ObjectNotFoundError
       end
 
+      # Find all members of a given resource, and map to Valkyrie Resources
+      # @param [Valkyrie::Resource]
+      # @param [Valkyrie::ResourceClass]
+      # @return [Array<Valkyrie::Resource>]
+      def find_members(resource:, model: nil)
+        find_model = model.internal_resource.constantize if model
```
revgum: I think it's minor, but as I was reading through the code I felt like it could benefit with chaining a select to a map which seemed somewhat easier to read.. 

    member_list = resource_factory.from_resource(resource: resource).try(:members)
    return [] unless member_list

    member_list.select { |obj| model ? obj.class == find_model : obj }
               .map { |obj| resource_factory.to_resource(object: obj) }

PR #3688, opened 2019-03-19
Run Circle at 10x concurrency

no-reply:
We just got a trial to run unlimited containers, why not see how fast we can
make this thing?

@samvera/hyrax-code-reviewers


no-reply:
This cuts the typical build to < 15 minutes, with the bulk of that time spent on the build stage.

Issue #3690, opened 2019-03-20
ValkyrieResource objects using Wings adapter


revgum:
### Descriptive summary

I think we should evaluate if the Wings adapter should be made to support handling objects which are natively `Valkyrie::Resource` objects, having no `ActiveFedora::Base` equivalent class.

The Wings adapter has a general expectation that a class exists which inherits from `ActiveFedora::Base` which can be initialized with the data persisted using Valkyrie. In the case where a native Valkyrie resource was persisted, there wouldn't be an ActiveFedora object to initialize when the resource is queried. The code falls back on a `DefaultWork` [1][2] that is created, currently for the sake of the specs.

[1] https://github.com/samvera/hyrax/blob/master/lib/wings/active_fedora_converter.rb#L40
[2] https://github.com/samvera/hyrax/blob/master/lib/wings/active_fedora_converter.rb#L74-L88

I wonder if this is a valid use-case for the Wings adapter. In the future, it seems like an error should be raised when an adequate `ActiveFedora::Base` object isn't found and the developer should be told to use a different adapter (which would have it's own model converter and resource factory). This seems to be the direction Valkyrie is going with having an adapter tied to a type of model and so on. (https://github.com/samvera/valkyrie/tree/master/lib/valkyrie/persistence/solr)

@no-reply @hackmastera

hackmastera:
We discussed this in a meeting. Summary of outcome of that discussion:

tl;dr: yes, we should support native valkyrie resources.

The end goal is for any part of the Hyrax code (and downstream codebases) to be able to use a Valkyrie resource instead of an ActiveFedora object. For example, creating a new work based on user form input. Therefore the wings adapter needs to accept native Valkyrie resources just as any other Valkyrie adapter would. 

It is okay for there to be limitations on or expectations of said Valkyrie resources, as long as they are well-understood and well-defined. Currently we are working toward making the Valkyrie shared specs pass in order to:
- Get as much general valkyrie functionality as possible and
- Help us clarify what limitations and/or expectations Valkyrie resources in hyrax should have.

@no-reply @revgum please add any additional clarifications.

PR #3691, opened 2019-03-20
allows for a native valkyrie resource to be deleted

### LINKED ISSUE #3673: wings persister: optimistic locking
make these pending specs pass

revgum:
Fixes #3673 

See also #3690 for follow-on discussion about longer term question

### DIFF
--- a/lib/wings/valkyrie/persister.rb
+++ b/lib/wings/valkyrie/persister.rb
@@ -37,7 +37,8 @@ module Wings
       # @param [Valkyrie::Resource] resource
       # @return [Valkyrie::Resource] the deleted resource
       def delete(resource:)
-        af_object = resource_factory.from_resource(resource: resource)
+        af_object = ActiveFedora::Base.new
+        af_object.id = resource.alternate_ids.first.to_s
         af_object.delete
       end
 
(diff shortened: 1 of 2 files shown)


lib/wings/valkyrie/persister.rb
```diff
@@ -37,7 +37,7 @@ def save_all(resources:)
       # @param [Valkyrie::Resource] resource
       # @return [Valkyrie::Resource] the deleted resource
       def delete(resource:)
-        af_object = resource_factory.from_resource(resource: resource)
+        af_object = ActiveFedora::Base.find(resource.alternate_ids.first.to_s)
```
no-reply: Could this behavior be baked into `from_resource`?

It seems like it might be useful in other contexts as well, and the switch to a hard dependency on AF, as opposed to continuing to use the factory, feels like a regression.

Is there more to say about this?

hackmastera: like by adding a flag on the factory method to just fetch it instead of converting it?

hackmastera: @no-reply I don't really understand. there are ActiveFedora dependencies all throughout the persister and the query service. I think that is expected. The persister and query service have to interact directly with the data store.

PR #3692, opened 2019-03-20
Reduce the number, but increase the size, of RSpec containers in CI

no-reply:
It turns out that using a larger container for the rspec runs reduces the build time significantly. We drop the parallelism a bit, since 10x didn't seem to be helping much more than 6x.

@samvera/hyrax-code-reviewers


cjcolvar:
@no-reply Did this have the effect you desired?  It's hard for me to tell if this had any effect.  I'm happy to approve and merge since the tests passed if you want.

PR #3693, opened 2019-03-20
wings persister can save native valkyrie resource with a nested resource property

### LINKED ISSUE #3662: wings persister spec: can save nested resources
- [ ] pass the shared Valkyrie specs related to resource nesting; or
  - [ ] document specific resource nesting behavior supported by Wings (e.g. write this up in the Hyrax Valkyrie Usage Guide)

hackmastera:
closes #3662

### DIFF
--- a/lib/wings/transformer_value_mapper.rb
+++ b/lib/wings/transformer_value_mapper.rb
@@ -19,9 +19,7 @@ module Wings
     end
 
     def result
-      id = ActiveFedora::Base.uri_to_id(value.id)
-      default_obj = ActiveFedora::Base.find id
-      attributes = Wings::ModelTransformer::AttributeTransformer.run(default_obj, default_obj.attributes.keys)
+      attributes = value.attributes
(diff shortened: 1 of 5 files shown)


lib/wings/transformer_value_mapper.rb
```diff
@@ -7,7 +7,28 @@ module Wings
...
+    # instead
+    TransformerValueMapper.register(self)
+
+    def self.handles?(value)
+      value.is_a? Wings::ActiveFedoraConverter::NestedResource
+    end
+
+    def result
+      id = ActiveFedora::Base.uri_to_id(value.id)
+      default_obj = ActiveFedora::Base.find id
```
hackmastera: This I am not sure about. It comes back as a `DefaultWork`. It's submitted as a hash, with the property defined as a `NestedResource`. Why doesn't it come back with that same model?

I think if it would come back right we wouldn't need to transform the attributes and create the nested_object.

hackmastera: thanks @revgum for helping me sort this out and fix it up!

PR #3695, opened 2019-03-21
Add wings query for find_inverse_references_by

### LINKED ISSUE #3694: wings query service: find_inverse_references_by
child of #3570

hackmastera:
closes #3694

### DIFF
--- a/spec/wings/valkyrie/query_service_spec.rb
+++ b/spec/wings/valkyrie/query_service_spec.rb
@@ -8,7 +8,9 @@ RSpec.describe Wings::Valkyrie::QueryService do
     class Book < ActiveFedora::Base
       include Hyrax::WorkBehavior
       property :title, predicate: ::RDF::Vocab::DC.title, multiple: true
-      property :a_member_of, predicate: ::RDF::URI.new('http://www.example.com/a_member_of'), multiple: true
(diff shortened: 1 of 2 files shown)


spec/wings/valkyrie/query_service_spec.rb
```diff
@@ -304,6 +310,69 @@ class Image < ActiveFedora::Base
...
+        it "returns an empty array if there are none" do
+          parent = persister.save(resource: resource_class.new(title: ['parent']))
+
+          expect(query_service.find_inverse_references_by(resource: parent, property: :a_member_of).to_a).to eq []
+        end
+      end
+
+      # not yet supported by the wings persister
+      context "when the property is ordered" do
+        xit "returns everything which references the given resource by the given property" do
```
jeremyf: @hackmastera Did you intend for this to have `xit`

hackmastera: @jeremyf yes it's still pending because ordered properties aren't passing persister specs. It may be that this will never be supported for the hyrax / wings adapter, but the initial phase is to get as many of valkyrie's shared specs passing as we can, and keeping this in as pending will help us track how we did with that.

PR #3697, opened 2019-03-21
Orders members when converting to AF objects.

rotated8:
Adding members via `#members=` does not add the member to the ordered relation, so that after a save, `#member_ids` contains objects, but `#ordered_member_ids` is empty. Since many parts of the code use `#ordered_members`, the converter should order the members.

### DIFF
--- a/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
+++ b/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
@@ -127,6 +127,22 @@ module Wings
       def respond_to_missing?(_name, _include_private = false)
         true
       end
+
+      # @param valkyrie [Boolean] Should the returned ids be for Valkyrie or AF objects?
+      # @return [Enumerable<Hydra::Works::Work>] The works this work contains
+      # NOTE: This method avoids using the Hydra::Works version of parent_works because of Issue #361
+      def parent_works(valkyrie: false)
+        af_child = Wings::ActiveFedoraConverter.new(resource: self).convert
+        af_parents = af_child.member_of_works
+        return af_parents unless valkyrie
+        af_parents.map(&:valkyrie_resource)
+      end
+
+      # @param valkyrie [Boolean] Should the returned ids be for Valkyrie or AF objects?
+      # @return [Enumerable<String> | Enumerable<Valkerie::ID] The ids of the file sets this work contains
+      def parent_work_ids(valkyrie: false)
+        parent_works(valkyrie: valkyrie).map(&:id)
+      end
     end
   end
 end
--- a/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
+++ b/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
@@ -139,7 +139,7 @@ module Wings
       end
 
       # @param valkyrie [Boolean] Should the returned ids be for Valkyrie or AF objects?
-      # @return [Enumerable<String> | Enumerable<Valkerie::ID>] The ids of the works this work is contained in
+      # @return [Enumerable<String> | Enumerable<Valkyrie::ID>] The ids of the works this work is contained in
       def parent_work_ids(valkyrie: false)
         parent_works(valkyrie: valkyrie).map(&:id)
       end
--- a/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
+++ b/lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
@@ -139,7 +139,7 @@ module Wings
       end
 
       # @param valkyrie [Boolean] Should the returned ids be for Valkyrie or AF objects?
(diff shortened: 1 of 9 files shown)


spec/wings/active_fedora_converter_spec.rb
```diff
@@ -81,6 +81,10 @@
         it 'converts member_of_collection_ids back to af_object' do
           expect(converter.convert.members.map(&:id)).to match_array [work2.id, work3.id]
```
elrayle: If you want to demonstrate the difference between members and ordered_members, you can use match_array here and reverse the order of the ids...
```
          expect(converter.convert.members.map(&:id)).to match_array [work3.id, work2.id]
```

spec/wings/active_fedora_converter_spec.rb
```diff
@@ -81,6 +81,10 @@
...
         end
+
+        it 'preserves order across conversion' do
+          expect(converter.convert.ordered_member_ids).to match_array [work2.id, work3.id]
```
elrayle: And this should use `.to eq`

```
expect(converter.convert.ordered_member_ids).to eq [work2.id, work3.id]
```

spec/wings/active_fedora_converter_spec.rb
rotated8: 👍

spec/wings/active_fedora_converter_spec.rb
rotated8: 👍

elrayle:
Exploring the fileset failures...
* The failure is on calls to fileset.parent_works which does not exist for resources in Hyrax and therefore, converts the resource to an af_object and passes it to Hydra::Works through the missing_method override for Hyrax resources
* The expected results are `[wk1, wk2]` and the actual results are `[wk1, wk2, wk2]`
* This happens because parent_works in Hydra::Works combines `in_works + member_of_works`.  `in_works` returns `[wk2]` and `member_of_works` returns `[wk1, wk2]`
* Testing with the original af_object before it is converted to the resource, and it had the same problem.  This implies that the problem my be at the Hydra::Works level.

What is not clear is why this did not happen before this branch and does with this branch.  What is also not clear is why wk2 is repeated, but wk1 is not.

A simple fix would be to apply `.uniq` to `parent_works` in Hydra::Works, but I'd like to understand why changing from setting the members with `members` to setting them with `ordered_members` would cause this error.

elrayle:
Since Hyrax is only interested in one direction, changed parent_works in Hyrax to not depend on parent_works in Hydra::Works.  This fixes the problem where parent_works for works and filesets was returning repeated parents.

lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
```diff
@@ -127,6 +127,22 @@ def method_missing(name, *args, &block)
...
+      # NOTE: This method avoids using the Hydra::Works version of parent_works because of Issue #361
+      def parent_works(valkyrie: false)
+        af_child = Wings::ActiveFedoraConverter.new(resource: self).convert
+        af_parents = af_child.member_of_works
+        return af_parents unless valkyrie
+        af_parents.map(&:valkyrie_resource)
+      end
+
+      # @param valkyrie [Boolean] Should the returned ids be for Valkyrie or AF objects?
+      # @return [Enumerable<String> | Enumerable<Valkerie::ID] The ids of the file sets this work contains
```
jrgriffiniii: Comment should be rewritten `# @return [Enumerable<String> | Enumerable<Valkyrie::ID>] The ids of the file sets this work contains`

lib/wings/hydra/works/models/concerns/work_valkyrie_behavior.rb
```diff
@@ -24,17 +24,17 @@ def file_set?
...
+      # # @return [Enumerable<Hydra::Works::Work>] The works this work is in
+      # def parent_works(valkyrie: false)
+      #   parent_objects(valkyrie: valkyrie).select(&:work?)
+      # end
+      #
+      # # @param valkyrie [Boolean] Should the returned ids be for Valkyrie or AF objects?
+      # # @return [Enumerable<String> | Enumerable<Valkerie::ID] The ids of the works this work is in
+      # def parent_work_ids(valkyrie: false)
+      #   parent_works(valkyrie: valkyrie).map(&:id)
+      # end
```
jrgriffiniii: This commented block needs to be removed.

lib/wings/hydra/pcdm/models/concerns/pcdm_valkyrie_behavior.rb
```diff
@@ -139,7 +139,7 @@ def parent_works(valkyrie: false)
       end
 
       # @param valkyrie [Boolean] Should the returned ids be for Valkyrie or AF objects?
-      # @return [Enumerable<String> | Enumerable<Valkyrie::ID>] The ids of the file sets this work contains
+      # @return [Enumerable<String> | Enumerable<Valkerie::ID>] The ids of the works this work is contained in
```
laritakr: Valkerie?

PR #3703, opened 2019-03-22
Revert "Move instance variable out of actor and into environment"

### LINKED ISSUE #3549: CreateWithRemoteFilesOrderedMembersActor test is unstable

hackmastera:
This reverts commit 53967b39cf5535f7036a17f95bfbb5ea70401975.

closes #3549


cjcolvar:
I haven't seen it break in the build in a while but that might be due to timing not hitting the issue.  If it isn't breaking the build, maybe it would be okay to leave it in for now?

hackmastera:
@cjcolvar sounds good! I'm not invested in seeing this merged. If it is eventually closed then #3722 should be closed, as well.

Issue #3604, opened 2019-03-11
work#reload refactor to operate in Valkyrie context


revgum:
### Descriptive summary

Calls to `#reload` should use the `valkyrie_metadata_adapter` to fetch/reload the resource from the backend. 

There are many of these calls throughout the specs, but there are a few found in the application as well.

https://github.com/samvera/hyrax/blob/master/app/jobs/create_derivatives_job.rb#L14
https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/file_set_actor.rb#L74

### Proposed Solution

Instead of these `#reload` calls, replace with a local call to something like the following:

```
def reload
  adapter = Valkyrie::MetadataAdapter.find(Hyrax.config.valkyrie_metadata_adapter)
  resource = adapter.query_service.find_by(id: id)
  adapter.resource_factory.from_resource(resource: resource)
end
```

no-reply:
instead of reworking `#reload` itself, we should simply replace the existing calls with a more explicit fetch of the valkyrie resource.

for the test suite cases, the best thing may be to refactor the test whenever the production code changes to support valkyrie.

PR #3704, opened 2019-03-22
add  _id attrs to wings roundtrip

lsat12357:
refs #3604

Add representative_id and thumbnail_id from object reflections to the list of attributes in the ModelTransformer following the current syntax (currently contains embargo_id and lease_id); their presence is tested in file_set_actor_spec.
Adds reflection ids if they contain values in ActiveFedoraConverter


lsat12357:
adding one line to set visibility: although it's in the attrs, it's not getting set.

lsat12357:
also, adding just the representative and thumbnail manually is a no go without testing that they're valid for the class; ended up refactoring and adding all _id attributes if they're in the class reflections

PR #3707, opened 2019-03-22
prefers to use internal test app from cache

### LINKED ISSUE #3706: CircleCI Build Job improvement
### Descriptive summary

Improve the speed of the test suite by caching the `.internal_test_app` and only regenerating it when necessary. 

- [x] Anytime code changes in one of the `lib/generators/hyrax/*` files
- [x] If the `.internal_test_app` hasn't already been generated

This should eliminate the build job from being run most of the time. There's always the ability to use CircleCi rebuild without cache to force the process to start over (because .internal_test_app/Gemfile wouldn't exist in cache)

revgum:
Fixes #3706

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,14 @@ jobs:
 
     - run:
         name: Generate test app
-        command: (git diff --name-only master | grep -qG 'generators\/hyrax') || bundle exec rake engine_cart:generate
+        command: |
+          has_generator_changes=`(git diff --name-only master | grep -G 'generators\/hyrax')`
+          if [ $? == 0 ] || [ ! -f .internal_test_app/Gemfile ]; then
+            bundle exec rake engine_cart:generate
+            exit $?
+          else
+            exit 0
+          fi
 
     - save_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -98,7 +98,14 @@ jobs:
 
     - run:
         name: Generate test app
-        command: (git diff --name-only master | grep -qG 'generators\/hyrax') || bundle exec rake engine_cart:generate
+        command: |
+          has_generator_changes=`(git diff --name-only master | grep -G 'generators\/hyrax')`
+          if [ $? == 0 ] || [ ! -f .internal_test_app/Gemfile ]; then
```
revgum: This checks that the previous command (git diff and grepping for any generators/hyrax file) exitted with a success (it found a changed file).. or if the .internal_test_app/Gemfile does not exist (because the cache has been cleared or it hasn't yet been cached). 

then it exits with the same status code from the engine cart generate so that the build job either succeeds or fails depending on the outcome of generation.

...

Or this just exits with a successful code because there are no generator file change and the .internal_test_app/Gemfile exists already.

revgum:
Cancelling this PR as we're likely to move to Circle Orb soon.

PR #3714, opened 2019-03-24
add navigator services to find child members for various model parts

### LINKED ISSUE #3713: create navigator services to find pcdm/works model children
### Descriptive summary

Create 'Navigator' services that use the existing valkyrie query service methods to find members by type (e.g. find_child_collections, find_child_works, find_child_filesets).

Alternate approach is to create pure Valkyrie custom queries ([documentation](https://github.com/samvera/valkyrie/wiki/Queries#custom-queries)) to implement these methods.

### Rationale

Moving PCDM/Works modeling to be closer aligned to Valkyrie.

### Expected behavior

```
  adapter = Wings::Valkyrie::MetadataAdapter.new
  query_service = Wings::Valkyrie::QueryService.new(adapter: adapter)

  # call custom query
  navigator = Wings.ChildCollectionsNavigator.new(query_service: query_service)
  navigator.find_child_collections(resource: parent_resource) # returns only collection resources from members
  navigator.find_child_collection_ids(resource: parent_resource) # returns only ids of collection resources from members

  navigator = Wings.ChildWorksNavigator.new(query_service: query_service)
  navigator.find_child_works(resource: parent_resource) # returns only work resources from members
  navigator.find_child_work_ids(resource: parent_resource) # returns only ids of work resources from members

  navigator = Wings.ChildFilesetsNavigator.new(query_service: query_service)
  navigator.find_child_filesets(resource: parent_resource) # returns only fileset resources from members
  navigator.find_child_fileset_ids(resource: parent_resource) # returns only ids of fileset resources from members
```

elrayle:
Fixes #3713

* find_child_collections/collection_ids
* find_child_works/work_ids
* find_child_filesets/fileset_ids

These use the query service methods to get members.  As the query service moves to pure valkyrie, these methods will become pure valkyrie without any changes.

----
*Design Question*

Right now, all children are stored in a single members relationship.  Because of this, to get_works, you instantiate all member resources and then keep only the work resources.

A more efficient approach would be to have work members and fileset members.  Then you can get only works or only filesets.  Additionally, in this approach, if you only want ids, you won’t have to instantiate any resources to get just work ids or just fileset ids.

### DIFF
--- a/lib/wings/navigators/child_collections_navigator.rb
+++ b/lib/wings/navigators/child_collections_navigator.rb
@@ -1,0 +1,35 @@
+# Navigate from a resource to the child collections in the resource.
+module Wings
+  class ChildCollectionsNavigator
+    # Define the queries that can be fulfilled by this navigator.
+    def self.queries
+      [:find_child_collections, :find_child_collection_ids]
+    end
+
+    attr_reader :query_service
+
+    def initialize(query_service:)
+      @query_service = query_service
+    end
+
+    # Find child collections of a given resource, and map to Valkyrie Resources
+    # @param [Valkyrie::Resource]
+    # @return [Array<Valkyrie::Resource>]
+    # TODO: By storing all children in a single relationship, it requires that the full resource be constructed for all children
+    #       and then selecting only the children of a particular type to return.
+    def find_child_collections(resource:)
+      query_service.find_members(resource: resource).select(&:collection?)
+    end
+
+    # Find the ids of child collections of a given resource, and map to Valkyrie Resources
+    # @param [Valkyrie::Resource]
+    # @return [Array<Valkyrie::ID>]
+    # TODO: By storing all children in a single relationship, it requires that the full resource be constructed for all children
+    #       and then selecting only the children of a particular type.  If we stored collections in a collections relationship and
+    #       works in a works relationship, then a request for IDs would return all ids from the relationship and not instantiate
+    #       any resources.
+    def find_child_collection_ids(resource:)
+      find_child_collections(resource: resource).map(&:id)
+    end
+  end
+end
--- a/lib/wings/valkyrie/query_service.rb
+++ b/lib/wings/valkyrie/query_service.rb
@@ -89,17 +89,11 @@ module Wings
       # @param [Valkyrie::ResourceClass]
       # @return [Array<Valkyrie::Resource>]
       def find_members(resource:, model: nil)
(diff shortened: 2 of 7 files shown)


lib/wings/navigators/child_collections_navigator.rb
```diff
@@ -0,0 +1,35 @@
+# Navigate from a resource to the child collections in the resource.
+module Wings
+  class ChildCollectionsNavigator
+    # Define the queries that can be fulfilled by this navigator.
```
elrayle: These set of classes and methods define navigator services.  An alternate pattern would be to define valkyrie custom queries.  

Considerations:
* valkyrie custom queries is an established pattern
* navigators use a naming scheme and setup similar to custom queries
* navigators do not make any direct queries of the backend.  They only use methods in the primary QueryService.

lib/wings/valkyrie/query_service.rb
```diff
@@ -82,17 +82,11 @@ def find_by_alternate_identifier(alternate_identifier:)
       # @param [Valkyrie::ResourceClass]
       # @return [Array<Valkyrie::Resource>]
       def find_members(resource:, model: nil)
+        return [] unless resource.respond_to?(:member_ids) && resource.member_ids.present?
+        all_members = find_many_by_ids(ids: resource.member_ids)
+        return all_members unless model
```
elrayle: This was rewritten to...
* return as early as possible to avoid expensive processing
* avoid conversions back and forth between resources and af object
* avoid processing of models unless a model filter was passed in
* use `find_many_by_ids` such that when that method is pure valkyrie, this method will also operate as pure valkyrie

lsat12357:
https://github.com/samvera/hyrax/blob/515f8a0edbd0cafb5ab3ecf2834c747b1e9422c3/lib/wings/valkyrie/query_service.rb#L87
do you need to test model on 88?

lib/wings/navigators/child_collections_navigator.rb
laritakr: I really like this pattern.

lib/wings/valkyrie/query_service.rb
laritakr: Great revision. We definitely didn't optimize the method, as we were just trying to get the queries working.

elrayle:
@lsat12357 I saw your question after merging based on the approved review.  

The test `unless model` is necessary.  Otherwise, the test two lines down for `== find_model` would be tested against `nil` and none of the results will be selected instead of all of the results which is what is wanted when a model is not passed in.

PR #3716, opened 2019-03-25
Display full geonames location

### LINKED ISSUE #1065: Index full place name from geonames
When I choose "Sheffield, South Yorkshire, United Kingdom" and it's indexed as 'Sheffield' ... and ditto for "Sheffield, Tasmania, Australia" 
screen shot 2017-05-26 at 10 24 16 am

lfarrell:
Fixes #1065

Adds geonames api call to allow for indexing of full geonames location: place, state, country

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/controlled_vocabularies/location.rb
+++ b/lib/hyrax/controlled_vocabularies/location.rb
@@ -4,9 +4,9 @@ module Hyrax
       configure rdf_label: ::RDF::Vocab::GEONAMES.name
 
       # Return a tuple of url & label
-      def solrize
+      def solrize(full_label)
         return [rdf_subject.to_s] if rdf_label.first.to_s.blank? || rdf_label.first.to_s == rdf_subject.to_s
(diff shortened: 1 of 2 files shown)


lib/hyrax/controlled_vocabularies/location.rb
```diff
@@ -4,9 +4,9 @@ class Location < ActiveTriples::Resource
       configure rdf_label: ::RDF::Vocab::GEONAMES.name
 
       # Return a tuple of url & label
-      def solrize
+      def solrize(full_label)
```
cjcolvar: Giving the new parameter a default value of `rdf_label.first` would make this change backwards-compatible:
```suggestion
      def solrize(full_label=rdf_label.first)
```

elrayle:
@lfarrell There may be an alternative way to approach this problem using the geonames linked data authority.  I'd be willing to work with you to explore what that might look like.  It could potentially avoid making the additional network calls in the deep indexing process.

PR #3723, opened 2019-03-28
use caches for bundle and internal test app

revgum:
use bundle and test application caches rather than defaulting to maybe using bundler to fetch gems

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
 
     - restore_cache:
         keys:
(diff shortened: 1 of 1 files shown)


no-reply:
seems like this falls down on spec runs. i expect the issue has to do with the multiple bundles in play: the engine's and the applications.

what about restoring

```
    - run:
        name: Ensure top-level Gemfile.lock is valid
        command: bundle check || bundle install

revgum:
giving up on this effort, I think the better solution will just to be to convert to using the samvera-labs CCI orb

PR #3726, opened 2019-04-01
Add find_parents query to wings query service

### LINKED ISSUE #3705: wings query service: find_parents
child of #3570

laritakr:
resolves #3705

Implements `find_parents` query in query service.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/valkyrie/query_service.rb
+++ b/lib/wings/valkyrie/query_service.rb
@@ -131,6 +131,19 @@ module Wings
         end
       end
 
+      # Find all parents of a given resource.
+      # @param resource [Valkyrie::Resource] The resource whose parents are being searched
+      #   for.
+      # @return [Array<Valkyrie::Resource>] All resources which are parents of the given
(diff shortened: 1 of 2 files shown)


lib/wings/valkyrie/query_service.rb
```diff
@@ -131,6 +131,19 @@ def find_inverse_references_by(resource: nil, id: nil, property:)
...
       end
 
+      # Find all parents of a given resource.
+      # @param resource [Valkyrie::Resource] The resource whose parents are being searched
+      #   for.
+      # @return [Array<Valkyrie::Resource>] All resources which are parents of the given
+      #   `resource`. This means the resource's `id` appears in their `member_ids`
+      #   array.
+      def find_parents(resource:)
+        id = resource.alternate_ids.first
```
jrgriffiniii: This question lies outside of the scope of this PR, but I would be curious as to why the singular `resource.id` cannot instead be used for the Solr query.

PR #3736, opened 2019-04-04
move converter value mappers to a separate class

elrayle:
Separates out the value mapper classes used by ActiveFedoraConverter into ConverterValueMapper.  This follows the same separation pattern setup with ModelTransformer and TransformerValueMapper.


lib/wings/converter_value_mapper.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  class NestedResourceValue < ::Valkyrie::ValueMapper
+    ConverterValueMapper.register(self)
+    def self.handles?(value)
+      value.last.is_a?(Hash)
+    end
+
+    def result
+      # [value.first, ActiveFedoraConverter.new(resource: value.last).convert]
```
jrgriffiniii: If this is a comment, then it should likely be expanded please.  Otherwise, I would please request that it be removed.

elrayle: I just moved this here, so I'm not sure the original purpose of this comment.  I suspect it is left over kruft from the original commit.  @laritakr Do you agree that this is ok to delete?

hackmastera: 👍

PR #3740, opened 2019-04-09
Relabel the links for private child works from 'File'

no-reply:
"File" isn't an accurate description of a work. When a child work isn't readable, just list it as "Private" instead.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -116,7 +116,7 @@ module Hyrax
     end
 
     def link_name
-      current_ability.can?(:read, id) ? to_s : 'File'
+      current_ability.can?(:read, id) ? to_s : 'Private'
     end
 
     def export_as_nt


ShanaLMoore:
We created a private work then switched to a different user to test that we'll see the text 'Private' where there would otherwise be a link to the work. Under this different user, we were not able to locate the private work. What would be a scenario where we could test this? CC @no-reply

laritakr:
You would need to create a public work with a private file.

ShanaLMoore:
As a user that created the public work with both a public and private file, I see:
Screen Shot 2019-04-29 at 12 33 18 PM

As a different user, I only see the public file and not the private one:
Screen Shot 2019-04-29 at 12 32 47 PM

As the different user, is it expected to see the public and private files listed with the private file showing 'Private' instead of the link?

laritakr:
Hmmm actually I'm not sure. I think private files used to be shown in at least some places, but that changed at some point. Maybe there is an admin toggle option to control it?

laritakr:
Screen Shot 2019-04-29 at 3 57 27 PM

Issue #3568, opened 2019-03-11
FileSetActor rework to decouple tight dependency on ActiveFedora


revgum:
### Descriptive summary

The FileSetActor has a couple `.save` calls as well as tight coupling to ActiveFedora, this needs to be reworked such that it can use Wings for making these persistence calls.

### Rationale

This change makes it possible for persistence to happen by way of Wings and into whichever backend is configured.

A couple of spots to look at:

https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/file_set_actor.rb#L24
https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/file_set_actor.rb#L81
https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/file_set_actor.rb#L104
https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/file_set_actor.rb#L162

PR #3742, opened 2019-04-10
add wings to file_set_actor

lsat12357:
refs #3604 #3568

Adds reload and destroy methods that go through the Wings module instead of directly using ActiveFedora methods. Similarly adds a save method which is currently still WIP and blocked from being used.

In order to correctly maintain the ActiveFedora::Aggregation::Proxy.count, this also changes the persister to use destroy! rather than delete, and operates on the found object.

One of the tests for actor.destroy uses the ActiveFedora reload and looks for an exception that is raised by AF reload when object.destroyed? returns true. Since the Wings conversion process requires that the destroy operation happen on another copy of the object, the original object.destroyed? returns false, and therefore, a different exception is raised by AF reload. Since this test's function seems to be confirming that the object was indeed destroyed after actor.destroy, the spec is changed to accept the Ldp::Gone exception.

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -1,5 +1,7 @@
+require 'wings'
 module Hyrax
   module Actors
+    # rubocop:disable Metrics/ClassLength
     # Actions are decoupled from controller logic so that they may be called from a controller or a background job.
     class FileSetActor
       include Lockable
@@ -21,7 +23,7 @@ module Hyrax
         # If the file set doesn't have a title or label assigned, set a default.
         file_set.label ||= label_for(file)
         file_set.title = [file_set.label] if file_set.title.blank?
-        return false unless file_set.save # Need to save to get an id
+        return false unless save(file_set) # Need to save to get an id
         if from_url
           # If ingesting from URL, don't spawn an IngestJob; instead
           # reach into the FileActor and run the ingest with the file instance in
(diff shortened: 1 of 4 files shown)


app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -71,14 +73,15 @@ def create_metadata(file_set_params = {})
...
+          work = reload(work) unless work.new_record?
           file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
+          save(file_set)
           work.ordered_members << file_set
           work.representative = file_set if work.representative_id.blank?
           work.thumbnail = file_set if work.thumbnail_id.blank?
           # Save the work so the association between the work and the file_set is persisted (head_id)
           # NOTE: the work may not be valid, in which case this save doesn't do anything.
-          work.save
+          save(work)
```
laritakr: This actor needs to be set up to use the parameter `use_valkyrie: false` like was done in https://github.com/samvera/hyrax/blob/094f5284d33252d356eaada1f18e7e1a2d3e3bd6/app/actors/hyrax/actors/base_actor.rb#L22

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -159,8 +162,37 @@ def unlink_from_work
...
-          work.save!
+          save(work, use_valkyrie: false)
         end
+
+        def save(af_object, use_valkyrie: false)
+          af_object.save unless use_valkyrie
+
+          adapter = Hyrax.config.valkyrie_metadata_adapter
+          resource = adapter.persister.save(resource: af_object.valkyrie_resource)
+          Wings::ActiveFedoraConverter.new(resource: resource).convert
```
elrayle: Is the conversion back to an af_object necessary?  It isn't assigned to anything and the method returns true.

elrayle:
Are there plans to add tests for the methods with use_valkyrie = true?  Right now, the valkyrie code in the save method looks untested.

For testing purposes, you could add a variable to the initialize method...
```
      def initialize(file_set, user, use_valkyrie: false)
        @file_set = file_set
        @user = user
        @use_valkyrie = use_valkyrie
      end
```

Then replace all the places that do...
```
save(file_set, use_valkyrie: false)
```
with
```
save(file_set, use_valkyrie: @use_valkyrie)
```

app/actors/hyrax/actors/file_set_actor.rb
lsat12357: oh shoot, forgot about this one. @revgum asked me to put this in.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -101,7 +105,7 @@ def update_metadata(attributes)
 
       def destroy
         unlink_from_work
-        file_set.destroy
+        file_set_destroy
```
no-reply: does this call a valkyrie destroy, regardless of `@use_valkyrie`?

Issue #3422, opened 2018-11-16
uv fails rendering png with transparent alpha channel 


jmfrenzel:
### Descriptive summary

Universal viewer (uv) fails rendering png with transparent alpha channel as well as uv download is providing corrupt jpg version 

### Expected behavior
uv should provide original png download link and render (if needed) jpg correctly

### Actual behavior
https://nurax-dev.curationexperts.com/concern/images/hd76s024f?locale=en#?c=0&m=0&s=0&cv=0&xywh=-2%2C0%2C290%2C99

### Steps to reproduce the behavior

1. https://nurax-dev.curationexperts.com/concern/images/hd76s024f?locale=en#?c=0&m=0&s=0&cv=0&xywh=-2%2C0%2C290%2C99Do this

### Related work

Link to related tickets or prior related work here.

no-reply:
Someone is going to need to look into this to determine if this is a UV issue, a RIIIF issue, a Hyrax issue, or some combination. After that analysis, we'll probably need to scope a ticket for Hyrax-side handling.

geekscruff:
I've confirmed that this is a problem with RIIIF and `convert`.

dlim87:
The UV isn't working for me in general, it's just rendering a black box for all the works I opened. In safari, chrome, and firefox

dlim87:
@dlpierce ^

dlpierce:
@dlim87 Did you regenerate the test app or update the iiif_image_builder lambda like this: https://github.com/samvera/hyrax/pull/3760#discussion_r312195728

dlim87:
I kinda dropped the ball on following up with this, since I was trying to check up on it through nurax. @dlpierce would further action need to be done on nurax as per that previous comment? ^

Unlike the past month, the UV is actually rendering now; but unfortunately, it's exhibiting the same problem as this ticket was created to address: it's displaying (and downloading) a malformed jpeg version, instead of the png version with the proper transparency layers.

MariaStarPower:
**Fail** - the Universal Viewer is not rendering anything.

![image](https://user-images.githubusercontent.com/32848057/101082656-a5d0f700-3560-11eb-901e-a083bb55f373.png)

![image](https://user-images.githubusercontent.com/32848057/101082710-bc774e00-3560-11eb-9546-2b80e3ee3c27.png)

mark-dce:
[...]

PR #3752, opened 2019-04-17
Add mime_type to the Riiif info_service

geekscruff:
Part of the fix for #3422;  This [Riiif PR](https://github.com/curationexperts/riiif/pull/125) provides the other part of the fix

In order to deal with #3422 we need the mime_type to be passed with the Riiif::Image.info_service

In Riiif we can then filter by png and then deal with the problematic alpha channel, reported in [riiif#124](https://github.com/curationexperts/riiif/issues/124)

* add mime_type to the response for `Riiif::Image.info_service`

* call `Riiif::Image.info_service.call(file_set_id, nil)` with an image file_set id

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/templates/config/initializers/riiif.rb
+++ b/lib/generators/hyrax/templates/config/initializers/riiif.rb
@@ -9,7 +9,7 @@ Riiif::Image.info_service = lambda do |id, _file|
   resp = ActiveFedora::SolrService.get("id:#{fs_id}")
   doc = resp['response']['docs'].first
   raise "Unable to find solr document with id:#{fs_id}" unless doc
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
@geekscruff do we expect to see the same "fedora" image viewer as we do in the thumbnail ? Is the error that they look different? See screenshot: 
Screen Shot 2019-04-29 at 11 14 10 AM

geekscruff:
@ShanaLMoore Hi Shana, this was only a part of what's needed to fix this, there is another PR coming with the full fix, so what you are seeing is expected, but will be fixed soon!

PR #3755, opened 2019-04-17
Fix template url and update rails version

cjcolvar:
`https://raw.githubusercontent.com/samvera/hyrax/v3.0.0-beta2/template.rb` returns a 404.  This PR fixes that url and also updates the rails version used.

@samvera/hyrax-code-reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ NOTE: The steps need to be done in order to create a new Hyrax based app.
 Generate a new Rails application using the template.
 
 ```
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
@cjcolvar Is the expected behavior to see the following screenshot when visiting "https://raw.githubusercontent.com/samvera/hyrax/v3.0.0.pre.beta2/template.rb" ?
Screen Shot 2019-04-29 at 11 05 51 AM

cjcolvar:
@ShanaLMoore Yes, that looks good to me.

PR #3757, opened 2019-04-19
Active fedora converter members

### LINKED ISSUE #3756: ActiveFedoraConverter convert_members method increases proxy count
### Descriptive summary

When converting a Valkyrie resource to an ActiveFedora object,  convert_members is using af_object.ordered_members = [wk1, wk2] which increases ActiveFedora::Aggregation::Proxy.count every time the converter runs. 

### Rationale
Not exactly sure how important this is? But the Proxy.count is tested in file_set_actor_spec.

### Expected behavior
after running ActiveFedoraConverter(resource: resource).convert, ActiveFedora::Aggregation::Proxy.count does not change
### Actual behavior
after running ActiveFedoraConverter(resource: resource).convert, ActiveFedora::Aggregation::Proxy.count increases by # of members

lsat12357:
Fixes #3756 

In ActiveFedoraConverter, setting the members in an ActiveFedora object being assembled from a Valkyrie Resource increases the object's proxy count each time the converter runs. Although the current code looks as though it ought to avoid increasing the member count maintained by ActiveFedora, ActiveFedora behaves as if items are being added.

This additional code adds the members to a temporary object and then copies from the temp object's proxy_association for ordered_members and members to the ActiveFedora object being assembled.

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -107,16 +107,23 @@ module Wings
     private
 
       def convert_members(af_object)
-        return unless resource.respond_to?(:member_ids) && resource.member_ids
-        return if af_object.ordered_members.ids == resource.member_ids
+        return unless resource.respond_to?(:member_ids) && !resource.member_ids.blank?
         # TODO: It would be better to find a way to add the members without resuming all the member AF objects
+        temp_object = assemble_members(af_object)
+        af_object.ordered_member_proxies.association.replace temp_object.ordered_member_proxies.association.to_a
+        af_object.members.proxy_association.target.concat temp_object.members.proxy_association.target
+        af_object
+      end
+
+      def assemble_members(af_object)
         temp_object = af_object.class.new
         resource.member_ids.each do |valkyrie_id|
           temp_object.ordered_members << ActiveFedora::Base.find(valkyrie_id.id)
         end
-        temp_object.id = resource.alternate_ids.first.to_s if resource.respond_to?(:alternate_ids) && resource.alternate_ids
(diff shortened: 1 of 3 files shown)


lib/wings/active_fedora_converter.rb
```diff
@@ -107,13 +107,23 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
     private
 
       def convert_members(af_object)
-        return unless resource.respond_to?(:member_ids) && resource.member_ids
+        return unless resource.respond_to?(:member_ids) && !resource.member_ids.blank?
```
jrgriffiniii:
```suggestion
        return unless resource.respond_to?(:member_ids) && resource.member_ids.present?
```
This would avoid double negation (https://api.rubyonrails.org/classes/Object.html#method-i-present-3F)

lib/wings/active_fedora_converter.rb
```diff
@@ -107,13 +107,23 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
...
         # TODO: It would be better to find a way to add the members without resuming all the member AF objects
-        ordered_members = []
+        temp_object = assemble_members(af_object)
+        af_object.ordered_member_proxies.association.replace temp_object.ordered_member_proxies.association.to_a
+        af_object.members.proxy_association.target.replace temp_object.members.proxy_association.target
```
jrgriffiniii: Not a criticism, but leaving `temp_object` for the GC is something I would imagine maintainers would wish to avoid.  I can't see any alternative to your approach, but it might still be worth noting.

Issue #125, opened 2016-12-21
0 files listed in Edit Work


mjgiarlo:
<a href="https://github.com/elrayle"><img src="https://avatars.githubusercontent.com/u/6855473?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [elrayle](https://github.com/elrayle)**
_Friday Apr 29, 2016 at 18:47 GMT_
_Originally opened as https://github.com/projecthydra/sufia/issues/1943_

----

### Descriptive summary

Sufia 7.0.0.beta1; Ruby 2.2.2

Editing a work should list files already uploaded.  This was the intention of the UI Working Group.  Has this been changed?
### Expected behavior

Per UI Working Group designs, I expected the previously uploaded files to be listed on the Files tab.
### Actual behavior

Files tab appears to be a place to add new files, but not effect existing files.  There are 0 files listed on the Files tab.
### Steps to reproduce the behavior
1. Works -> New Work
2. Metadata Tab:  Set Title
3. File Tab:  Add one file
4. Save
5. Confirm Title is correct and added file is listed in Items section
6. Edit
7. Files Tab:  NOTICE 0 files are listed
### Related work
#1942

mjgiarlo:
<a href="https://github.com/jcoyne"><img src="https://avatars.githubusercontent.com/u/92044?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jcoyne](https://github.com/jcoyne)**
_Monday May 02, 2016 at 14:03 GMT_

----

> Editing a work should list files already uploaded. This was the intention of the UI Working Group. Has this been changed?

I didn't understand your intention from the wireframes.  This sounds reasonable to me.  +1 on slotting this for 7.1.0.

@elrayle can we remove the "question" and "needs feedback" label now?

PR #3760, opened 2019-04-26
Universal viewer (UV) fails rendering png with transparent alpha channel

### LINKED ISSUE #3422: uv fails rendering png with transparent alpha channel

geekscruff:
Fixes #3422; required a new version of Riiif: [#125](https://github.com/curationexperts/riiif/pull/125) released as 2.1.

This PR, combined with a change in Riiif will fix an issue whereby images with an alpha channel are incorrectly rendered by Universal Viewer backed by Riiif.

The incorrect rendering occurs because Riiif is asked for a 'jpg', and jpegs do not support alpha channels. 

One complication is that Universal Viewer requests 'jpg' even if png is specified in the manifest. The changes proposed in Riiif help to get around that by returning a png if an alpha channel present, irrespective of the request format.

Changes in Hyrax are:

1) Extend the characterize job to run Imagemagick identify and extract channel information (via the Riiif::ImageInfoExtractor) ... we need channel information to pass to Riiif, and we cannot get this from FITS AFAIK. Channel information could also be useful for preservation purposes, so it is useful to collect it. This channel information is added to the characterization_proxy and indexed into solr.

2) Change the Riiif info_service to return format AND alpha_channels; Riiif will then check for an alpha channel and preserve this, returning a png instead of a jpg.

@samvera/hyrax-code-reviewers


cjcolvar:
I wonder if it will become a problem having this property called "channels" when audio files have "channels" as well.
https://github.com/samvera/hydra-works/blob/master/lib/hydra/works/characterization/schema/audio_schema.rb#L4

geekscruff:
@cjcolvar good point, I'll rename it alphachannels ... I want to get this finished this week if at all possible

cjcolvar:
It would be good for someone else to review this because I don't work with images and don't feel confident enough.

geekscruff:
@cjcolvar thanks! I need to get the tests passing, and do a bit of manual QA of my own, then I'll ping a couple of reviewers

geekscruff:
@no-reply Hi Tom, any chance of a review?

app/jobs/characterize_job.rb
```diff
@@ -9,11 +9,26 @@ class CharacterizeJob < Hyrax::ApplicationJob
...
+      Hydra::Works::CharacterizationService.run(file_set.characterization_proxy, filepath)
+      Rails.logger.debug "Ran characterization on #{file_set.characterization_proxy.id} (#{file_set.characterization_proxy.mime_type})"
+      file_set.characterization_proxy.alpha_channels = channels(filepath) if file_set.image? && Hyrax.config.iiif_image_server?
+      file_set.characterization_proxy.save!
+      file_set.update_index
+      file_set.parent&.in_collections&.each(&:update_index)
+    end
+
+    def channels(filepath)
+      ch = Riiif::ImageMagickInfoExtractor.new(filepath).extract[:channels]
```
cjcolvar: Maybe it makes sense to avoid the dependency on Riiif and instead pull this code into Hyrax or where ever characterization code lives.

app/presenters/hyrax/displays_image.rb
```diff
@@ -35,5 +37,9 @@ def iiif_endpoint(file_id)
           profile: Hyrax.config.iiif_image_compliance_level_uri
         )
       end
+
+      def image_format(channels)
+        channels.find { |c| c.include?('rgba') }.nil? ? 'jpg' : 'png'
```
cjcolvar: It looks like this will always return a jpg or png but do other IIIF image servers or IIIF clients make use of or expect anything different?  Maybe not.

dlpierce: jpg is the only format required at all compliance levels, with png required for level 2. See https://iiif.io/api/image/2.1/compliance/#format

lib/generators/hyrax/riiif_generator.rb
```diff
@@ -33,7 +33,7 @@ def mount_route
 
   def override_image_url_builder_in_hyrax_config
     insert_into_file 'config/initializers/hyrax.rb', before: /^  # config.iiif_image_url_builder/ do
-      "  config.iiif_image_url_builder = lambda do |file_id, base_url, size|\n" \
+      "  config.iiif_image_url_builder = lambda do |file_id, base_url, size, format|\n" \
```
cjcolvar: Does this change in the lambda signature require updating documentation or at least need to be called out in the 3.0 migration docs?

dlpierce: Yes, I believe it will.

PR #3761, opened 2019-04-26
Improve the universal viewer deployment

geekscruff:
Based on feedback from @little9 over at [nurax#336](https://github.com/curationexperts/nurax/pull/336)

The uv install can be run with `yarn install` or `rake assets:precompile`.

I'll update the Hyrax Management Guide accordingly.

@samvera/hyrax-code-reviewers


no-reply:
This looks great.

@geekscruff is there any action that needs to be taken post-merge for `nurax`?

geekscruff:
@no-reply thanks! nothing for nurax, as I have a PR in to put the new files in place (those come during install which isn't re-run for nurax).

no-reply:
ah, yeah. do we need a rake task to move files around during upgrade?

geekscruff:
@no-reply good idea, I'll get that in early next week

PR #3764, opened 2019-04-26
Index versioned file URI for IIIF, improve thumbnail cache behavior.

### LINKED ISSUE #2910: Universal Viewer does not show new file version

dlpierce:
Fixes #2910 
Use versioned URI for IIIF image requests, and instruct browser to check for new versions of thumbnails. 

* Upload a new version of a file in a fileset
* Work show page should show new version in UV and a matching thumbnail.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/local_file_downloads_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/local_file_downloads_controller_behavior.rb
@@ -29,6 +29,7 @@ module Hyrax
       end
 
       def send_local_file_contents
+        return unless stale?(last_modified: local_file_last_modified, template: false)
         self.status = 200
         prepare_local_file_headers
         # For derivatives stored on the local file system
@@ -65,7 +66,7 @@ module Hyrax
         response.headers['Content-Type'] = local_file_mime_type
         response.headers['Content-Length'] ||= local_file_size.to_s
         # Prevent Rack::ETag from calculating a digest over body
-        response.headers['Last-Modified'] = local_file_last_modified.utc.strftime("%a, %d %b %Y %T GMT")
+        response.headers['Last-Modified'] ||= local_file_last_modified.httpdate
         self.content_type = local_file_mime_type
       end
 
--- a/app/controllers/concerns/hyrax/local_file_downloads_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/local_file_downloads_controller_behavior.rb
@@ -66,7 +66,7 @@ module Hyrax
         response.headers['Content-Type'] = local_file_mime_type
         response.headers['Content-Length'] ||= local_file_size.to_s
         # Prevent Rack::ETag from calculating a digest over body
-        response.headers['Last-Modified'] ||= local_file_last_modified.httpdate
+        response.headers['Last-Modified'] = local_file_last_modified.httpdate
         self.content_type = local_file_mime_type
       end
 
--- a/app/indexers/hyrax/file_set_indexer.rb
+++ b/app/indexers/hyrax/file_set_indexer.rb
@@ -42,11 +42,7 @@ module Hyrax
 
       def original_file_id
         return unless object.original_file
-        if object.original_file.versions.present?
-          ActiveFedora::File.uri_to_id(object.current_content_version_uri)
-        else
-          object.original_file.id
-        end
+        Hyrax::VersioningService.versioned_file_id object.original_file
       end
 
       def file_format
--- a/app/indexers/hyrax/file_set_indexer.rb
+++ b/app/indexers/hyrax/file_set_indexer.rb
@@ -28,11 +28,21 @@ module Hyrax
         solr_doc['sample_rate_tesim']       = object.sample_rate
         solr_doc['original_checksum_tesim'] = object.original_checksum
         solr_doc['alpha_channels_ssi']      = object.alpha_channels
+        solr_doc['current_file_version_ssi'] = latest_version_id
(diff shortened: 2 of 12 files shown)


Dananji:
@dlpierce I was trying to test this. The thumbnail in the works page doesn't show. And UV gives the following pop-up message when I go to the work show page;
UV fails in Hyrax

This is not related to PR, because it happens even when I checkout `master` branch.
I'm running Hyrax locally following the instructions [here](https://github.com/samvera/hyrax/wiki/Hyrax-Development-Guide#work-with-test-app-in-the-browser).
Thanks!

dlpierce:
It was determined this was due to a missing FITS install.

spec/support/factory_helpers.rb
```diff
@@ -18,7 +18,9 @@ def mock_file_factory(opts = {})
                  original_checksum: opts.fetch(:original_checksum, []),
                  digest:            opts.fetch(:digest, []),
                  duration:          opts.fetch(:duration, []),
-                 sample_rate:       opts.fetch(:sample_rate, []))
+                 sample_rate:       opts.fetch(:sample_rate, []),
+                 versions:          opts.fetch(:versions, []),
+                 uri:               opts.fetch(:versions, []))
```
cjcolvar: Was this deliberate or a copy/paste issue?
```suggestion
                 uri:               opts.fetch(:uri, []))
```

app/presenters/hyrax/displays_image.rb
```diff
@@ -35,5 +41,10 @@ def iiif_endpoint(file_id)
           profile: Hyrax.config.iiif_image_compliance_level_uri
         )
       end
+
+      def unindexed_current_file_version
+        Rails.logger.warn "Indexed current_file_version for #{id} not found, falling back to Fedora."
```
cjcolvar: I couldn't find `current_file_version` defined anywhere.  Maybe something different could be used to make this warning clearer?

app/presenters/hyrax/displays_image.rb
```diff
@@ -38,8 +41,17 @@ def iiif_endpoint(file_id)
         )
       end
 
+      def lookup_original_file_id
+        result = original_file_id
+        if result.blank?
+          Rails.logger.warn "original_file_id for #{id} not found, falling back to Fedora."
+          result = ActiveFedora::File.uri_to_id(::FileSet.find(id).current_content_version_uri)
```
cjcolvar: Does this need a fallback in the case that `original_file.versions.present?` would fail?  `FileSetPresenter#original_file_id` in this PR is doing that at least.
I also wonder if both of these cases are getting tested.

app/controllers/concerns/hyrax/local_file_downloads_controller_behavior.rb
```diff
@@ -29,6 +29,7 @@ def send_range_for_local_file
       end
 
       def send_local_file_contents
+        return unless stale?(last_modified: local_file_last_modified, template: false)
```
cjcolvar: This should probably be tested.

app/controllers/concerns/hyrax/local_file_downloads_controller_behavior.rb
```diff
@@ -65,7 +66,7 @@ def prepare_local_file_headers
         response.headers['Content-Type'] = local_file_mime_type
         response.headers['Content-Length'] ||= local_file_size.to_s
         # Prevent Rack::ETag from calculating a digest over body
-        response.headers['Last-Modified'] = local_file_last_modified.utc.strftime("%a, %d %b %Y %T GMT")
+        response.headers['Last-Modified'] ||= local_file_last_modified.httpdate
```
cjcolvar: This should probably be tested.
I'm also unsure about the `||=`.  What would be previously setting `Last-Modified` and should it really be preferred over `local_file_last_modified.httpdate`?  The comment above leads me to believe that it shouldn't but maybe the comment just needs to be updated.

dlpierce: The call to `stale?` will add a Last-Modified header before this code executes, but I think you're right that the conditional assignment isn't really needed.

app/presenters/hyrax/displays_image.rb
dlpierce: I DRYed this up by moving it to the VersioningService.

app/indexers/hyrax/file_set_indexer.rb
```diff
@@ -41,7 +41,11 @@ def digest_from_content
 
       def original_file_id
         return unless object.original_file
-        object.original_file.id
+        if object.original_file.versions.present?
```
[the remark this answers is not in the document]
dlpierce: Added a test.

PR #3769, opened 2019-04-30
Extract more strings for i18n

j-dornbusch:
A few strings extracted for i18n, mostly confirmation messages.
The same strings were present in dedicated flash message templates ( app/views/hyrax/file_sets/_asset_deleted_flash.html.erb,  app/views/hyrax/file_sets/_asset_saved_flash.html.erb) and in controllers : it is difficult to understand the reason. I think that templates are never used without being sure. I translated both.
String concatenation in  app/assets/javascripts/hyrax/collections.js is not a robust solution: the syntax varies from one language to another. We have no idea of the result in Chinese. I only translated the part that was left in English but it would be necessary to change the whole approach.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/collections.js
+++ b/app/assets/javascripts/hyrax/collections.js
@@ -249,11 +249,13 @@ Blacklight.onLoad(function () {
     var tableRows = $('#documents table.collections-list-table tbody tr');
     var checkbox = null;
     var numRowsSelected = false;
+    var $modal = $('#selected-collections-delete-modal');
+    var $deleteWordingTarget = $modal.find('.pluralized');
     var deleteWording = {
-      plural: 'these collections',
-      singular: 'this collection'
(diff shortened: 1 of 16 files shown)


j-dornbusch:
@cjcolvar @adamjarling 
I have already used the data-attribute's approach to solve the problem of client-side i18n in https://github.com/samvera/hyrax/pull/3564. This approach is implicit in the way hydra-editor's manage_repeating_fields.js builds and uses the options object.

cjcolvar:
@j-dornbusch 👍 I had forgotten about that change.

cjcolvar:
@j-dornbusch I think this will be good if you rebase the latest from master.

j-dornbusch:
@cjcolvar 
I rebased as requested and the tests passed. Should I also squash the commits?

PR #3776, opened 2019-05-16
fix current logic error in displaying/hiding feature/unfeature buttons

### LINKED ISSUE #3169: Feature/unfeature buttons don't work when there are five works already featured
### Descriptive summary

In 2.1 Hyrax, the feature/unfeature button at some point splits into 2 buttons (supposed to be one) and then stops working entirely. I can't find a good way to reproduce this with steps consistently, but try going to a work and feature and unfeature it, go back to the home page and back to the work. 

We think it might be if you hit 5 featured works, then it breaks. 

here is a screenshot from nurax.

screen shot 2018-07-25 at 1 35 23 pm

### Done when

Users no longer have the option to feature works when there are already five works features on the home page (?). Open to other suggestions.

yf508:
Fixes #3169 

The current implementation of showing/hiding feature/unfeature buttons seems to be incorrect. 

e.g. for feature button:

<%= link_to "Feature", hyrax.featured_work_path(presenter, format: :json),
          data: { behavior: 'feature' },
          class: presenter.display_unfeature_link? ? 'btn btn-default collapse' : 'btn btn-default' %>

It says, "if display_unfeature_link returns true, then hide feature button, OTHERWISE, show feature button". While display_unfeature_link returns false, it doesn't necessarily mean feature button needs to be shown.  For example, when 5 works have been featured already, both feature/unfeature buttons should be hidden for new objects. Therefore, to make it simple and working:

<%= link_to t('.feature'), hyrax.featured_work_path(presenter, format: :json),
          data: { behavior: 'feature' },
          class: presenter.display_feature_link? ? 'btn btn-default' : 'btn btn-default collapse' %>

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_show_actions.html.erb
+++ b/app/views/hyrax/base/_show_actions.html.erb
@@ -9,11 +9,11 @@
     <% if presenter.work_featurable? %>
       <%= link_to t('.feature'), hyrax.featured_work_path(presenter, format: :json),
           data: { behavior: 'feature' },
-          class: presenter.display_unfeature_link? ? 'btn btn-default collapse' : 'btn btn-default' %>
+          class: presenter.display_feature_link? ? 'btn btn-default' : 'btn btn-default collapse' %>
 
       <%= link_to t('.unfeature'), hyrax.featured_work_path(presenter, format: :json),
           data: { behavior: 'unfeature' },
-          class: presenter.display_feature_link? ? 'btn btn-default collapse' : 'btn btn-default' %>
+          class: presenter.display_unfeature_link? ? 'btn btn-default' : 'btn btn-default collapse' %>
     <% end %>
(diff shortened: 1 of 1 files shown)


no-reply:
This looks good to me.

Added a Check CLA label.

cjcolvar:
@yf508 Thanks for the pull request! In order for us to merge this code, we'll need to have signed Contributor License Agreements on file for you as an individual and your employer. (If you work for University of York then it looks like we already have the cCLA.) You can read more about the Samvera community's CLA processes here:
https://wiki.duraspace.org/display/samvera/Samvera+Community+Intellectual+Property+Licensing+and+Ownership

yf508:
I have sent the iCLA via email.

cjcolvar:
:+1: Looks like it has been received.  Thanks!

PR #3777, opened 2019-05-17
Allows Hyrax to use prerelease versions of Valkyrie.

rotated8:
After 2.0.0rc2 was yanked, the specification for Valkyrie in the gemspec could not be satisfied.


jrgriffiniii:
https://github.com/samvera/hyrax/pull/3781 should resolve this problem for now, I'm still investigating what will be needed to support more recent release candidates.

PR #3782, opened 2019-05-21
Create derivatives job reload

### LINKED ISSUE #3604: work#reload refactor to operate in Valkyrie context

lsat12357:
Fixes #3604 (together with PR #3742)

Adds a reload method using wings; uses the same code as in PR#3742 file_set_actor#reload.
Removes from the spec the expectation that file_set will receive reload.

### DIFF
--- a/app/jobs/create_derivatives_job.rb
+++ b/app/jobs/create_derivatives_job.rb
@@ -1,3 +1,4 @@
+require 'wings'
 class CreateDerivativesJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
 
@@ -11,7 +12,7 @@ class CreateDerivativesJob < Hyrax::ApplicationJob
     file_set.create_derivatives(filename)
 
     # Reload from Fedora and reindex for thumbnail and extracted text
-    file_set.reload
+    file_set = reload(file_set)
     file_set.update_index
     file_set.parent.update_index if parent_needs_reindex?(file_set)
   end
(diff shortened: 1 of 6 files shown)


app/jobs/create_derivatives_job.rb
```diff
@@ -22,4 +23,14 @@ def parent_needs_reindex?(file_set)
     return false unless file_set.parent
     file_set.parent.thumbnail_id == file_set.id
   end
+
+  def reload(af_object)
+    return af_object unless af_object.persisted?
+
+    adapter = Hyrax.config.valkyrie_metadata_adapter
```
elrayle: If this is the same code as in FileSetActor, I would recommend pulling it out of both into a service (e.g. `lib/wings/services/af_object_reloader_service.rb`).

Issue #3293, opened 2018-09-24
Map license to dcterms:license


ruthtillman:
### Descriptive summary

Map field "license" to dcterms:license instead of DCT:rights.

### Rationale

Right now “license” maps to DCT:rights. Rights and licensing are related but fundamentally different data, as the license conveys information about how the object may be used. Map instead to 
dcterms:license. This will require data migration.

From the Metadata IG Hyrax review project. UCSD, IU, Tufts and Oregon State in favor of this change.

### Related work

See also comment https://github.com/samvera/hyrax/issues/808#issuecomment-296028015

jenyoung:
Also, the list of Creative Commons licenses needs to be updated to 4.0 licenses. Current list is for 3.0

no-reply:
I like the idea of moving to `dcterms:license` but want to acknowledge that we don't currently meet the range restriction http://dublincore.org/documents/dcmi-terms/#terms-license. Should we try to wrap the license update, nested data, and predicate change all into this ticket?

kdid:
@jenyoung @ruthtillman @no-reply - can someone clarify what needs to be done here? I'm not sure I'm following.

jlhardes:
Updating issue title to reflect that property should be mapped to 
http://purl.org/dc/terms/license

jenyoung:
Also updated ticket to reflect the above

fritzfreiheit:
@kellyChex -- Here's what I've worked out on how to QA this. It's all done from your localhost as you will need to run a rake task do verification through the rails console.

* Using a local install, start up and run Hyrax with the latest from head.
* Create a GenericWork and add a license metadata field (using `Additional Fields`)
* Note the id of the GenericWork
* Edit the file `app/models/concerns/hyrax/basic_metadata.rb`
* Replace the following line (this will allow the creation of works with the old license predicate) :
`property :license, predicate: ::RDF::URI.new('http://dublincore.org/documents/dcmi-terms/#terms-license'), multiple: true`
with
`property :license, predicate: ::RDF::Vocab::DC.rights, multiple: true`
* Restart Hyrax
* Add another GenericWork with a license field and note the id
* Start a rails console and run the following:
`predicate_from = ::RDF::Vocab::DC.rights`
`predicate_to = ::RDF::URI.new('http://dublincore.org/documents/dcmi-terms/#terms-license')`
[...]

PR #3785, opened 2019-05-23
HYRAX-#3293 - Map license to dcterms:license

fritzfreiheit:
Fixes HYRAX-#3293

Change the basic meta data for `license` to the url pointing to the `dcterms:license`

### DIFF
--- a/app/models/concerns/hyrax/basic_metadata.rb
+++ b/app/models/concerns/hyrax/basic_metadata.rb
@@ -20,7 +20,7 @@ module Hyrax
       property :abstract, predicate: ::RDF::Vocab::DC.abstract
       property :keyword, predicate: ::RDF::Vocab::SCHEMA.keywords
       # Used for a license
-      property :license, predicate: ::RDF::Vocab::DC.rights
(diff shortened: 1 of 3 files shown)


no-reply:
I think we need to provide a migration script for this, yes? Or else make this configurable, and retain the existing behavior in applications that were generated prior to 3.0.0?

no-reply:
@fritzfreiheit could you provide a commit/PR message for this?

See https://chris.beams.io/posts/git-commit/

fritzfreiheit:
> I think we need to provide a migration script for this, yes? Or else make this configurable, and retain the existing behavior in applications that were generated prior to 3.0.0?

Sorry, but I really don't know the code well enough to do more than nod in agreement. If I had an example of a migration script and configuration to work off of I would be able to answer more intelligently and do the work. Probably someone else should have done the coding for this, as you can see my code was far too simplistic and totally lacks the depth of knowledge to have done it intelligently.

fritzfreiheit:
@no-reply -- You can see the driver for the migration here:
https://github.com/samvera/hyrax/blob/master/app/services/hyrax/works/migration_service.rb

PR #3795, opened 2019-05-28
Verifiy the validity of the the Link URL supplied for the collection …

geekscruff:
Users creating a new collection can provide a Link URL for the logo icon. There is currently no validation on the link.


no-reply:
@geekscruff @cjcolvar it looks to me like the failures are due to neglected maintenance on this branch's CI config (`chromedriver` related; cc: @jrgriffiniii).

I'd suggest we not wait for the fix to push this through.

geekscruff:
Merging as test failure is unrelated.

PR #3796, opened 2019-05-28
[wings] Avoid double setting visibility in Wings converter

no-reply:
`#visibility` in `Hydra::AccessControls` is a projection on the ACLs. Since the
converter sets the ACL users and groups manually, there's no need to set
`visibility` separately; doing so may complicate special cases like embargo and
lease.

This is a non-functional change, isolated to Wings.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -34,7 +34,7 @@ module Wings
     end
 
     ##
-    # Accesses and parses the attributes from the resource through ConvetrerValueMapper
+    # Accesses and parses the attributes from the resource through ConverterValueMapper
     # @return [Hash]
     def attributes
       @attribs ||= begin
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -34,7 +34,7 @@ module Wings
     end
 
     ##
-    # Accesses and parses the attributes from the resource through ConverterValueMapper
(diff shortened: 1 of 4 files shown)


lib/wings/active_fedora_converter.rb
```diff
@@ -34,7 +34,7 @@ def initialize(resource:)
     end
 
     ##
-    # Accesses and parses the attributes from the resource through ConverterValueMapper
+    # Accesses and parses the attributes from the resource through ConvetrerValueMapper
```
jcoyne: Typo?

revgum:
fixes #3806

lib/wings/active_fedora_converter.rb
no-reply: i suppose so, but not one introduced in this PR.

no-reply:
@revgum rebased.

lib/wings/active_fedora_converter.rb
jcoyne: why is GH showing it as a change then?

no-reply: oh huh. i assumed it was a whitespace change, but i guess i must have done a stealth letter transpose in my editor.

sorry for the confusion, and thanks for the catch!

jrgriffiniii:
The failing test appears to be unrelated (I've encountered this on my branches: https://circleci.com/gh/samvera/hyrax/4287#tests/containers/2), I'll restart the build.

Issue #3800, opened 2019-05-29
setup solr indexers to use valkyrie/wings


elrayle:
### Descriptive summary

Setup solr service and indexers in hyrax to use when persisting via valkyrie.  This will create and register a solr indexing adapter that can be used with any backend.  Wings uses ActiveFedora for solr indexing.  There will need to be some thought put into how the indexing adapter will be used when the metadata/storage adapters are configured to use Wings.  Once we move away from AF / wings, having an indexing adapter will allow new indexing adapters to be created for other indexing schemes (e.g. elastic search) that can be swapped out. 

#### Hyrax indexers

The various Hyrax indexers live in [app/indexers](https://github.com/samvera/hyrax/blob/master/app/indexers).

The valkyrie version of the indexers, at least for now, will live in [lib/wings/indexers](https://github.com/samvera/hyrax/blob/master/lib/wings/indexers).

### Creating Valkyrie indexers

REF: https://github.com/samvera/valkyrie/issues/598

Valkyrie does not currently have the concept of an indexing adapter, but there have been discussions about creating one (see issue https://github.com/samvera/valkyrie/issues/598).  The work being done in Hyrax will serve as a test case for this concept.

#### Key Concepts and Functionality:

* The indexer is write-only.  As such, it can create light weight solr docs including only the required metadata and can control additional information going into the solr doc to support relationships between objects.
* There is a chain of indexers that builds up the solr document.  See PR #3866 which adds a basic indexer and proposes indexers Hyrax::Indexers::WorkIndexer and Hyrax::Indexers::CollectionIndexer (see registration of indexers at [indexing_adapter_initializer.rb](https://github.com/samvera/hyrax/blob/master/config/initializers/indexing_adapter_initializer.rb#L11-L12)). The last to write to a field has the final say on the value of a particular field in the solr doc.
* The chain of indexers will also need to address the nested indexing functionality which creates the hierarchy of collections within the solr documents for each collection in the hierarchy.
* Indexing chains need to be customizable to allow apps to further control the creation of solr docs.  If a field should have a different name or be dropped from the index, an app can overwrite a provided indexer to modify the desired behavior.
* There needs to be a solr_read_service that supports the connections to solr required for reading documents since the indexer is write-only.
* Currently there are two solr cores set up:  1) active fedora core, 2) valkyrie core.  Need to write to both until all uses of the active fedora core are addressed, then the active fedora core can be removed.
[...]

PR #3826, opened 2019-06-18
[wings] Introduce a Hyrax::IndexingService to replace the AF class

no-reply:
For now, this service subclasses the ActiveFedora version, but can be used
throughout the local codebase, allowing us to deprecate direct inheritance of
`ActiveFedora::IndexingService`, and any of its features that we would prefer
not to support going forward.

Connected to #3800.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/indexers/hyrax/work_indexer.rb
+++ b/app/indexers/hyrax/work_indexer.rb
@@ -1,5 +1,5 @@
 module Hyrax
-  class WorkIndexer < ActiveFedora::IndexingService
+  class WorkIndexer < Hyrax::IndexingService
     include Hyrax::IndexesThumbnails
     include Hyrax::IndexesWorkflow
 
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -15,6 +15,7 @@ require 'browse-everything'
 require 'hydra/works'
 require 'hyrax/engine'
 require 'hyrax/version'
+require 'hyrax/indexing_service'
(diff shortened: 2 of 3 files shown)


no-reply:
@hackmastera do you think this is still useful in light of your current work?

hackmastera:
@no-reply I'm not sure. should we close it for now and re-open if needed?

hackmastera:
@no-reply ack sorry this is a PR i was reading it as an issue. one sec.

hackmastera:
Could this object be used to essentially serve as both an AF indexer and a Valkyrie indexer at the same time? Considerations:
 - valkyrie indexers don't have to inherit from a superclass, so there's no conflict there
 -  if we use the composite indexer, which seems like the right approach, valkyrie indexers expect to be initialized with a resource keyword (https://github.com/samvera/valkyrie/blob/master/lib/valkyrie/persistence/solr/composite_indexer.rb). AF indexers just pass the object into the initializer w/ no keyword. Probably we don't want to write a different / alternate composite indexer?
- Otherwise, the implementations generating the hashes are different. AF indexers use `super` to get the hash so far, while the Valkyrie composite indexer just just through them and merges everything into one hash. There might be a way to pull out the hash itself into a helper that you invoke from both `generate_solr_doc` (AF) and `to_solr` (Valkyrie)

So I think probably no, given the initializer difference.

no-reply:
Closing for now.

Issue #3790, opened 2019-05-28
Valkyrize calls to ActiveFedora::Base.find (removes 27 refs to ActiveFedora)


elrayle:
### Descriptive summary

Replace all calls to ActiveFedora::Base.find to use valkyrie query_service to find the object (except calls made in the lib/wings directories).  At this writing there are 27 uses of the ActiveFedora::Base #find method.

### ActiveFedora code

Typical call looks like...

```
ActiveFedora::Base.find(id)
```

search for `ActiveFedora::Base.find` to find all occurrences.

### Valkyrie replacement code could look something like...

Create `app/queries/hyrax/find_query.rb`.

Assumes id is the active_fedora id
```
      resource = query_service.find_by(id: Valkyrie::ID.new(id.to_s))
      af_object = Wings::ActiveFedoraConverter.new(resource: resource).convert # TODO: WINGS - Remove when Valkyrie fully integrated.
```

_NOTE: There is one instance that uses ActiveFedora::Base.find with two arguments, the second of which is cast.  This does not address that usage._

_NOTE: There is at least one instance where an array of ids is passed in instead of a single id._

Reference: Replaces https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/relation/finder_methods.rb#L28-L49

### Considerations

* Wings::Valkyrie::QueryService #find_by looks for the alternate_identifier which is the ActiveFedora object's ID
* Since we are still working with ActiveFedora objects in the rest of the code, the Valkyrie resource has to be converted back to an ActiveFedora object.
* There is a Wings::IdConverterService that gets the Valkyrie id by using ActiveFedora to find the AF object and then getting the valkyrie id from it.  Can't use it in this case because it is a circular call since it uses the ActiveFedora::Base.find method we are trying to remove.  
* Do NOT replace this call if found in lib/wings directories.

### Impact on Testing

* Not expected to require changes to existing tests.  All tests should pass with the replacement code.

elrayle:
There was discussion in slack this morning.  I believe we have landed on an approach that will ultimately end with native Valkyrie code in each location where `find` is being replaced.

The summary of that discussion follows.

Update query_service.find_by to allow it to return a resource or an AF object.  By default, it will return a resource.

```
QueryService.find_by(id: id, use_valkyrie: true)
```

Then all instances calling `find` can be changed to…

```
query_service.find_by(id: id, use_valkyrie: false)
```

[...]

Issue #3821, opened 2019-06-17
Replace calls to ActiveFedora::Base.uri_to_id and .id_to_uri (removes 6 refs to ActiveFedora)


elrayle:
### Descriptive summary

`ActiveFedora::Base.uri_to_id` and `ActiveFedora::Base.id_to_uri` are used to translate back and forth between the AF uri and AF id. 

These calls will be replaced in two steps.  The first step will replace the ActiveFedora::Base calls with the same calls defined in Hyrax::Base in app/models/hyrax/base.rb.  This Hyrax::Base versions of these methods will simply call the ActiveFedora::Base versions.  This will isolate the refs to ActiveFedora to a single class.

The second step will rewrite the methods to not have a dependency on ActiveFedora.

_NOTE: If app/models/hyrax/base.rb does not exist, create it as part of this issue._

### ActiveFedora code

Active Fedora call being replaced...

Example from app/services/hyrax/list_source_exporter.rb
```
          parent_id = ActiveFedora::Base.uri_to_id(parent_url)
```

To locate, search code for `ActiveFedora::Base.uri_to_id`.

#### First pass

At least initially, define a service class to do these transformations defined in Hyrax::Base (see above).

```
module Hyrax
  class Base
    def self.uri_to_id(uri)
      ActiveFedora::Base.uri_to_id uri
    end

    def self.id_to_uri(id)
      ActiveFedora::Base.id_to_uri id
    end
  end
end
```

Example from app/services/hyrax/list_source_exporter.rb
```
          parent_id = Hyrax::Base.uri_to_id(parent_url)
```

### Second pass

How to change this into Valkyrie native code?  Potential locations for the actual transformations...

* wings metadata adapter defines `uri_to_id` and `id_to_uri` methods
    * Challenge - Other metadata adapters (e.g. postgres) do not define these methods.  Note that the Valkyrie defined Fedora adapter does, but presumably are only used by the Fedora adapters/query service.
    * If the methods are moved to the wings metadata adapter, can the usage be isolated to the wings adapters/query service and removed from Hyrax general code?
* resource defines `uri_to_id` and `id_to_uri` methods
    * Challenge - What will these do when the metadata adapter is postgres or some other adapter without URIs?
* somewhere else?

For the wings AF metadata adapter, it can continue to make the ActiveFedora::Base calls to these methods.  It is not clear how this will be handled for other metadata adapters.  This may involve a translation between the valkyrie URI and the alternate id, and vice versa.

[...]

Issue #3038, opened 2018-05-07
Fix randomly failing feature tests associated with select_member_of_collection


elrayle:
### Descriptive summary

A number of feature tests are randomly failing since the addition of `select_member_of_collection` feature which uses jquery select2 to search for a collection by name.

### Expected behavior

REMOVE optional line from each test...
```
      optional 'ability to get capybara to find css select2-result (see Issue #3038)' if ENV['TRAVIS']
```

All tests in spec/features/collection_multi_membership_spec.rb and spec/features/work_show_spec.rb pass consistently when the optional line is removed from the tests.

### Actual behavior

6 tests in collection_multi_membership_spec.rb and 1 in spec/features/work_show_spec.rb (all marked optional for Travis) randomly fail when running in Travis.

NOTE: These tests pass locally and only fail on Travis.  They are currently marked optional.  The goal of this Issue is to remove the optional statement and have them run without failures in Travis.

The error occurs on the `select_member_of_collection` line which calls the method in spec/support/selectors/rb.  The failing line in that method is the attempt to click `collection.title.first` within `select2-result`.

Attempts to fix included...
* adding `expect(page).to have_css('.select2-result') which causes some wait time through Malificient
* adding sleep time
* changing the expect to look for `select2-results-label` instead of `select2-result`

There are similar tests that use `select_collection` also defined in spec/support/selectors.rb that pass consistently.

### Related work

PR #3027 Change radio buttons into autocomplete field for add works to collection modal

jrgriffiniii:
https://circleci.com/gh/jrgriffiniii/hyrax/17 is queued to run a possible solution to this.  If this works, I can merge this into https://github.com/samvera/hyrax/pull/3493

PR #3828, opened 2019-06-18
Replace ActiveFedora::Base.find

laritakr:
references https://github.com/samvera/hyrax/issues/3790

Replace the majority of direct calls to ActiveFedora::Base.find with calls to Valkyrie query service find_by_alternate_identifier.

Note: Does not address following cases:
* find is using a batch of ids rather than a single id
* nesting index adapter

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/persist_directly_contained_output_file_service.rb
+++ b/app/services/hyrax/persist_directly_contained_output_file_service.rb
@@ -24,7 +24,7 @@ module Hyrax
     def self.retrieve_file_set(directives)
       uri = URI(directives.fetch(:url))
       raise ArgumentError, "#{uri} is not an http(s) uri" unless uri.is_a?(URI::HTTP)
-      ActiveFedora::Base.find(ActiveFedora::Base.uri_to_id(uri.to_s))
+      Valkyrie.config.metadata_adapter.query_service.find_by_alternate_identifier(alternate_identifier: ActiveFedora::Base.uri_to_id(uri.to_s), use_valkyrie: false)
     end
     private_class_method :retrieve_file_set
 
--- a/lib/wings/valkyrie/query_service.rb
+++ b/lib/wings/valkyrie/query_service.rb
@@ -87,8 +87,8 @@ module Wings
         object = ::ActiveFedora::Base.find(alternate_identifier.to_s)
         return object if use_valkyrie == false
         resource_factory.to_resource(object: object)
-        rescue ::ActiveFedora::ObjectNotFoundError, Ldp::Gone => e
-          raise use_valkyrie == true ? ::Valkyrie::Persistence::ObjectNotFoundError : e
+      rescue ::ActiveFedora::ObjectNotFoundError, Ldp::Gone => e
+        raise use_valkyrie == true ? ::Valkyrie::Persistence::ObjectNotFoundError : e
       end
 
       # Find all members of a given resource, and map to Valkyrie Resources
--- a/lib/wings/valkyrie/query_service.rb
+++ b/lib/wings/valkyrie/query_service.rb
@@ -76,12 +76,19 @@ module Wings
         resources
       end
 
-      def find_by_alternate_identifier(alternate_identifier:)
+      # Find a record using an alternate ID, and map it to a Valkyrie Resource
+      # @param [Valkyrie::ID, String] id
+      # @param [boolean] optionally return ActiveFedora object/errors
+      # @return [Valkyrie::Resource]
+      # @raise [Valkyrie::Persistence::ObjectNotFoundError]
+      def find_by_alternate_identifier(alternate_identifier:, use_valkyrie: true)
         alternate_identifier = ::Valkyrie::ID.new(alternate_identifier.to_s) if alternate_identifier.is_a?(String)
         validate_id(alternate_identifier)
-        resource_factory.to_resource(object: ::ActiveFedora::Base.find(alternate_identifier.to_s))
-      rescue ::ActiveFedora::ObjectNotFoundError, Ldp::Gone
-        raise ::Valkyrie::Persistence::ObjectNotFoundError
+        object = ::ActiveFedora::Base.find(alternate_identifier.to_s)
+        return object if use_valkyrie == false
+        resource_factory.to_resource(object: object)
+        rescue ::ActiveFedora::ObjectNotFoundError, Ldp::Gone => e
+          raise use_valkyrie == true ? ::Valkyrie::Persistence::ObjectNotFoundError : e
       end
 
       # Find all members of a given resource, and map to Valkyrie Resources
--- a/spec/features/create_work_spec.rb
+++ b/spec/features/create_work_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe 'Creating a new Work', :js, :workflow do
+RSpec.describe 'Creating a new Work', :js, :workflow, :clean_repo do
   let(:user) { create(:user) }
   let!(:ability) { ::Ability.new(user) }
   let(:file1) { File.open(fixture_path + '/world.png') }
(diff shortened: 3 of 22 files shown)


laritakr:
Fixes two randomly failing specs from #3038. They needed additional time after the save so the content could become available. Commented them out to retain the reference to the issue in case their passing was a fluke. Comments can be removed in a later PR if they continue to pass.

app/services/hyrax/persist_directly_contained_output_file_service.rb
```diff
@@ -24,7 +24,7 @@ def self.call(content, directives)
     def self.retrieve_file_set(directives)
       uri = URI(directives.fetch(:url))
       raise ArgumentError, "#{uri} is not an http(s) uri" unless uri.is_a?(URI::HTTP)
-      ActiveFedora::Base.find(ActiveFedora::Base.uri_to_id(uri.to_s))
+      Valkyrie.config.metadata_adapter.query_service.find_by_alternate_identifier(alternate_identifier: ActiveFedora::Base.uri_to_id(uri.to_s), use_valkyrie: false)
```
jeremyf: Do we need to cast the URI? Can we use `uri` instead of `ActiveFedora::Base.uri_to_id(uri.to_s)`?

laritakr: uri_to_id is being addressed by another issue so I left this as is.

laritakr: https://github.com/samvera/hyrax/issues/3821

lib/wings/valkyrie/query_service.rb
```diff
@@ -76,12 +76,19 @@ def find_many_by_ids(ids:)
         resources
       end
 
-      def find_by_alternate_identifier(alternate_identifier:)
+      # Find a record using an alternate ID, and map it to a Valkyrie Resource
+      # @param [Valkyrie::ID, String] id
+      # @param [boolean] optionally return ActiveFedora object/errors
+      # @return [Valkyrie::Resource]
+      # @raise [Valkyrie::Persistence::ObjectNotFoundError]
```
elrayle: Nice adding in the documentation!

spec/features/create_work_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe 'Creating a new Work', :js, :workflow do
+RSpec.describe 'Creating a new Work', :js, :workflow, :clean_repo do
```
elrayle: Did something in the PR cause the need for `:clean_repo`?

spec/features/static_pages_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe "The static pages" do
+RSpec.describe "The static pages", :clean_repo do
```
elrayle: Did something in the PR cause the need for `:clean_repo`?

spec/features/work_show_spec.rb
```diff
@@ -55,7 +55,7 @@
     end
 
     it "allows adding work to a collection", clean_repo: true, js: true do
-      optional 'ability to get capybara to find css select2-result (see Issue #3038)' if ci_build?
+      # optional 'ability to get capybara to find css select2-result (see Issue #3038)' if ci_build?
```
elrayle: Can you elaborate on why this change and the addition of `sleep 5` below are needed?

spec/features/work_show_spec.rb
```diff
@@ -107,12 +108,13 @@
...
```
elrayle: same question here

spec/wings/valkyrie/query_service_spec.rb
```diff
@@ -152,6 +152,24 @@ class Image < ActiveFedora::Base
       expect(found.id).to eq resource.id
       expect(found).to be_persisted
     end
+
+    context 'when use_valkyrie: false' do
+      it 'returns an ActiveFedora object' do
+        resource = resource_class.new
```
elrayle: Any reason not to set `id` outside the test using...
```
let(:id) do
  resource.alternate_ids = [Valkyrie::ID.new('p9s0xfj')]
  resource = persister.save(resource: resource)
  resource.alternate_ids.first
end
```

spec/features/static_pages_spec.rb
laritakr: No this has been a randomly failing spec for me. It was failing due to some missing thumbnail in the "recent objects" list due to an object that didn't get created fully elsewhere.

spec/features/work_show_spec.rb
laritakr: After a save is issued, there can be some delay as it processes before the object is available for the subsequent commands. If you debug, it slows things down and works, but when run without debug, it fails. Sleep allows the page to be fully available for the subsequent tests.

spec/features/work_show_spec.rb
laritakr: (same answer) :)

spec/features/create_work_spec.rb
laritakr: If this fails once, then it continues to fail. The work gets created with the same title as previous attempts, so the page has multiple works with the same title, and a later find fails.

spec/features/static_pages_spec.rb
laritakr: (Randomly failing both locally and in CI)

spec/features/work_show_spec.rb
no-reply: okay to just remove this line, rather than commenting?

spec/wings/valkyrie/query_service_spec.rb
laritakr: I copied the code from an earlier spec and changed it for the "use_valkyrie" parameter. I'm not positive, but I believe that I split it out from the code elsewhere to placate Rubocop who thought line 163 was too long when it included `found = query_service.find_by_alternate_identifier(alternate_identifier: resource.alternate_ids.first, use_valkyrie: false)'. Naming it id is probably a bit unclear, given all of the other ids referenced in that spec.

spec/features/work_show_spec.rb
laritakr: I left the comment temporarily, to keep the reference to the related issue until I was sure it fixed the problem.  I intended to remove it later, but if this PR has other changes needed, I'll remove it here.

spec/features/static_pages_spec.rb
elrayle: Thanks for the clarification.  And nice to see changes that make the tests more stable.

spec/wings/valkyrie/query_service_spec.rb
elrayle: No problem.  This is more of a style thing.  I generally try to put setup code that isn't part of what is being tested outside the test itself.  Makes what is being tested easier to see at a glance.

spec/features/work_show_spec.rb
elrayle: My worry would be that these kinds of comments can end up being out of sight, out of mind and never get deleted until someone stumbles across it.  And that person may not be sure it is ok to remove.  

Perhps you could add a comment referencing the issue near the `sleep` statement.  

_Caveat: I'm not familiar with `optional` in rspec tests and a quick google search did not enlighten me.  So my suggestion may not make sense if I guessed its function incorrectly._

Something like...
```
# This attempts to address issue #3038.  Previously, this test was marked optional.

laritakr: I could, but honestly, I intend to remove the comment entirely in my next PR. I just wanted to leave it when I pushed up the code, to make sure it fixed it on Travis as well as locally, so it would be easier to find if it didn't work. Is it worth removing it or adding a different comment in this PR? or should i just do it as part of my next one?

elrayle: That seems fine if you are doing it in the next PR.

PR #3830, opened 2019-06-19
[wings] Add Hyrax finder interface

no-reply:
This interface replaces `ActiveFedora::Base.find` in a way that can support a
clean changeover to Valkyrie.

Existing calls to `ActiveFedora::Base.find` can be swapped for
`Hyrax::ActiveFedoraFinder.find` calls. This class returns the same types, and
raises the same errors as the `AF::Base` version. That can later be swapped for
`Hyrax::ResourceFinder`, which is a light wrapper around
`Valkyrie` `#find_by`.

Some consideration was given to using a Valkyrie custom query, but since the
end-goal for this particular swap is to bundle a globally configured
query adapter and its existing `find_by` method, there seemed to be little
point.

Related to #3790.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/collections_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/collections_controller_behavior.rb
@@ -25,7 +25,7 @@ module Hyrax
     end
 
     def show
-      @curation_concern ||= ActiveFedora::Base.find(params[:id])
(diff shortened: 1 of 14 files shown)


elrayle:
There have been discussions about the best way to handle the replacement for find.  I believe we have landed on a path that moves us toward a natural migration to fully native valkyrie code.  Check out the comment summarizing the slack discussion in issue #3790.

Issue #3794, opened 2019-05-28
Replace calls to ActiveFedora.index_field_mapper.solr_name in app (removes 54 refs to ActiveFedora)


elrayle:
### Descriptive summary

This issue involves removing references to `ActiveFedora.index_field_mapper.solr_name`.

### ActiveFedora code

Active Fedora call being replaced...

Example 1:  has_model from app/models/concerns/hyrax/solr_document_behavior.rb
```
      first(ActiveFedora.index_field_mapper.solr_name('has_model', :symbol)).constantize
```

Example 2: depositor from app/models/concerns/hyrax/solr_document_behavior.rb
```
      val = first(Solrizer.solr_name('depositor'))
```

To locate, search code for `ActiveFedora::SolrQueryBuilder.construct_query(` in `app/search_builders`.

### Replacement code

Example 1: has_model from app/models/concerns/hyrax/solr_document_behavior.rb
Switch to use Valkyrie::Persistence::Solr::Queries constant
```
      first(Valkyrie::Persistence::Solr::Queries::MODEL).constantize
```

Example 2: depositor from app/models/concerns/hyrax/solr_document_behavior.rb
Switch to using Solrizer to generate the name.
```
      val = first(Solrizer.solr_name('depositor'))
```

For some instances, you can compare [master app/search_builders](https://github.com/samvera/hyrax/tree/master/app/search_builders) with [valkryie-6 app/search_builders](https://github.com/samvera/hyrax/tree/valkyrie-6/app/search_builders) to see how it was changed in that branch.

More exploration may be needed to insure the field being converted to a solr_name is supported by Solrizer.

### Considerations

* I only found one instance matching the pattern shown in example 1.  The remaining matched that pattern in example 2.
* Not all diffs between master and valkryie-6 branch are necessarily related to removing references to ActiveFedora.  The valkyrie-6 branch has not been kept in sync with master.
* For the ActiveFedora.index_field_mapper.solr_name, many of the calls at master didn't exist when the valkyrie-6 branch was active.
* Do NOT replace this call if found in lib/wings directories.

### Impact on Testing

Not expected to require test changes.  All tests should pass with the replacement code.

cjcolvar:
I wouldn't recommend using `Solrizer`: https://github.com/samvera/maintenance/issues/19

no-reply:
Agreed about Solrizer.

Instead, we should consider in-lining solr name generation to Hyrax.

jrgriffiniii:
[...]

PR #3831, opened 2019-06-19
Use actual solr field names instead of index_field_mapper.set_field

hackmastera:
advances #3794

### DIFF
--- a/app/indexers/hyrax/admin_set_indexer.rb
+++ b/app/indexers/hyrax/admin_set_indexer.rb
@@ -7,7 +7,7 @@ module Hyrax
     def generate_solr_document
       super.tap do |solr_doc|
         # Makes Admin Sets show under the "Admin Sets" tab
-        Hyrax.config.index_field_mapper.set_field(solr_doc, 'generic_type', 'Admin Set', :facetable)
+        solr_doc['generic_type_sim'] = ['Admin Set']
       end
     end
   end
--- a/app/indexers/hyrax/collection_indexer.rb
+++ b/app/indexers/hyrax/collection_indexer.rb
@@ -11,9 +11,9 @@ module Hyrax
     def generate_solr_document
       super.tap do |solr_doc|
         # Makes Collections show under the "Collections" tab
-        Hyrax.config.index_field_mapper.set_field(solr_doc, 'generic_type', 'Collection', :facetable)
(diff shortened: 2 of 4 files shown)


app/indexers/hyrax/admin_set_indexer.rb
```diff
@@ -7,7 +7,7 @@ class AdminSetIndexer < ActiveFedora::IndexingService
     def generate_solr_document
       super.tap do |solr_doc|
         # Makes Admin Sets show under the "Admin Sets" tab
-        Hyrax.config.index_field_mapper.set_field(solr_doc, 'generic_type', 'Admin Set', :facetable)
+        solr_doc['generic_type_sim'] = ['Admin Set']
```
jcoyne: How about setting these in a lookup table?  Like `Hyrax.config.solr_fields = { type: 'generic_type_sim' }`?

hackmastera: Maybe. I think you might start working your way back to something like solrizer, though, because there will often be more than one solr field for a given model property. I'm advocating to move toward "just let everyone use the solr field they want to use". The dynamic fields are pretty well documented in the solr config. And I think this is a direction the community has been more or less comfortable with for a while.

jcoyne: Let's walk through that "just let everyone use the solr field they want to use" scenario.  What if I want to use `primary_type_sim` instead of `generic_type_sim`.  How would I go about that?

hackmastera: Write an indexer that adds the field you want. And then either stop using the one you don't want or have your indexer remove it I guess.

hackmastera: @jcoyne are you trying to say that this field in particular is uniquely in need of being overridden downstream? Could you flesh out the use case?

jcoyne: Okay, so the plan is "write your own (or override the default indexers)". I can live with that.

no-reply: :+1: these are already hard coded in many places, including adjacent to changes in this diff. It seems fine to just normalize on "you customize index keys by changing the indexer to use different keys".

We're going to get something to that effect into the release notes.

PR #3832, opened 2019-06-19
[wings] Introduce a converter actor for reworking the actor stack

no-reply:
These are support actors, designed to create a boundary in the stack between
actors that handle valkyrie resources, and code that uses ActiveFedora.

By using both actors adjacent to each other anywhere in the actor stack, a two
way conversion can be achieved. Actors below the boundary will accept Valkyrie
resources as their `env.curation_concern` input and can be designed to act only
on Valkyrie objects. Actors above the boundary can continue to work with
ActiveFedora models.

Alternatively, these actors may be used independantly, to swap input/output for
a specific actor.

The idea is that this boundary can begin at the bottom of the stack, and actors
can be reworked or replaced to move the boundary up the stack.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/actors/hyrax/actors/active_fedora_to_valkyrie_spec.rb
+++ b/spec/actors/hyrax/actors/active_fedora_to_valkyrie_spec.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+RSpec.describe Hyrax::Actors::ActiveFedoraToValkyrie do
+  let(:ability)    { :FAKE_ABILITY }
+  let(:attrs)      { {} }
+  let(:env)        { Hyrax::Actors::Environment.new(work, ability, attrs) }
+  let(:spy)        { middleware.next_actor }
+  let(:terminator) { Hyrax::Actors::Terminator.new }
+  let(:work)       { FactoryBot.build(:work) }
+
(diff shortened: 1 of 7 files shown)


elrayle:
Is there a companion actor that converts the ActiveFedora curation_concern object into a Valkyrie resource?

I sort of envisioned, at the top of the actor stack, include an actor that converts the af object into a valkyrie resource.  Then all the following actor in the stack can perform all their operations on that valkryie resource instead of the af object.

How do you envision this actor plugging into the actor stack processing?

no-reply:
@elrayle does the updated commit info answer your questions?

spec/actors/hyrax/actors/active_fedora_to_valkyrie_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    stack.build(terminator)
+  end
+
+  shared_examples 'casts to Valkyrie' do |method|
+    it 'returns true' do
+      expect(middleware.public_send(method, env)).to be true
+    end
+
+    it 'casts to ActiveFedora' do
```
elrayle: Should this be `it 'casts to Valkyrie Resource'`?

PR #3834, opened 2019-06-20
[wings] Inline #query, #post, and #get into solr_service from AF

hackmastera:
refs #3800

This is a small step toward having a full solr indexing and query service directly in hyrax. It sets us up to start working on bringing the connection itself into hyrax and testing an approach to a samvera/valkyrie#598

### DIFF
--- a/app/models/solr_hit.rb
+++ b/app/models/solr_hit.rb
@@ -1,0 +1,23 @@
+class SolrHit < Delegator
+  def __getobj__
+    @document # return object we are delegating to, required
+  end
+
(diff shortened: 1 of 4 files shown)


app/models/solr_hit.rb
```diff
@@ -0,0 +1,29 @@
+class SolrHit < Delegator
+  def self.for(hit)
+    return hit if hit.is_a? ActiveFedora::SolrHit
```
no-reply: should this be just `SolrHit`?

Issue #3835, opened 2019-06-20
Replace ActiveFedora::Base.find calls which receive a batch of ids


laritakr:
### Descriptive summary

Refer to https://github.com/samvera/hyrax/issues/3790

This issue refers to only the remaining calls which receive a batch of ids rather than a single id. This happens when working with collections, leases, and embargoes.

### Related work

PR https://github.com/samvera/hyrax/pull/3828 replaced the majority of direct calls to `ActiveFedora::Base.find`.

PR #3841, opened 2019-06-21
Replace ActiveFedora::Base.find(ids) calls

laritakr:
Create and use custom query "find_many_by_alternate_ids".
Remove obsolete comments for specs fixed in PR #3828

Fixes https://github.com/samvera/hyrax/issues/3835

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/custom_queries/wings.rb
+++ b/app/services/hyrax/custom_queries/wings.rb
@@ -1,0 +1,34 @@
+module Hyrax
+  module CustomQueries
+    class Wings
+      # Holds custom queries for wings
+      # Use:
+      # Hyrax.query_service.custom_queries.find_many_by_alternate_ids(alternate_ids: ids, use_valkyrie: true)
+
+      def self.queries
+        [:find_many_by_alternate_ids]
+      end
+
+      attr_reader :query_service
+      delegate :resource_factory, to: :query_service
+
+      def initialize(query_service:)
+        @query_service = query_service
+      end
+
+      # implements a combination of two Valkyrie queries:
+      # => find_many_by_ids & find_by_alternate_identifier
+      # @param [Enumerator<#to_s>] ids
+      # @param [boolean] defaults to true; optionally return ActiveFedora::Base objects if false
+      # @return [Array<Valkyrie::Resource>, Array<ActiveFedora::Base>]
+      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: true)
+        af_objects = ActiveFedora::Base.find(alternate_ids.map(&:to_s))
+        return af_objects unless use_valkyrie == true
+
+        af_objects.map do |af_object|
+          resource_factory.to_resource(object: af_object)
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
@@ -26,7 +26,8 @@ module Hyrax
     def update
       filter_docs_with_edit_access!
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] }
-      ActiveFedora::Base.find(batch).each do |curation_concern|
+      af_objects = Hyrax.query_service.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
+      af_objects.each do |curation_concern|
         Hyrax::Actors::EmbargoActor.new(curation_concern).destroy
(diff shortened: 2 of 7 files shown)


app/custom_queries/hyrax/wings/custom_query_service.rb
```diff
@@ -0,0 +1,44 @@
...
+      def initialize(query_service:)
+        @query_service = query_service
+        @resource_factory = query_service.adapter.resource_factory
+      end
+
+      # implements a combination of two Valkyrie queries:
+      # => find_many_by_ids & find_by_alternate_identifier
+      # @param [Array<Valkyrie::ID, String>] ids
+      # @param [boolean] defaults to true; optionally return ActiveFedora::Base objects if false
+      # @return [Valkyrie::Resource or ActiveFedora::Base]
```
no-reply: `[Array<Valkyrie::Resource>, Array<ActiveFedora::Base>]`?

app/custom_queries/hyrax/wings/custom_query_service.rb
```diff
@@ -0,0 +1,44 @@
...
+      # NOTE: Ignores non-existent ids without raising an error
+      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: true)
```
no-reply: It seems like it would be good to implement this as an [`Enumerator`](https://ruby-doc.org/core-2.2.0/Enumerator.html), to avoid loading large collections into memory before iterating over them.

app/custom_queries/hyrax/wings/custom_query_service.rb
```diff
@@ -0,0 +1,44 @@
...
+        @query_service = query_service
+        @resource_factory = query_service.adapter.resource_factory
+      end
+
+      # implements a combination of two Valkyrie queries:
+      # => find_many_by_ids & find_by_alternate_identifier
+      # @param [Array<Valkyrie::ID, String>] ids
+      # @param [boolean] defaults to true; optionally return ActiveFedora::Base objects if false
+      # @return [Valkyrie::Resource or ActiveFedora::Base]
+      # NOTE: Ignores non-existent ids without raising an error
```
no-reply: Does this match the current `ActiveFedora::Base.find` behavior, or will this change current behavior?

app/custom_queries/hyrax/wings/custom_query_service.rb
```diff
@@ -0,0 +1,44 @@
...
+      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: true)
+        alternate_ids.each do |id|
+          id = ::Valkyrie::ID.new(id.to_s) if id.is_a?(String)
+          raise ArgumentError, 'id must be a Valkyrie::ID' unless id.is_a? ::Valkyrie::ID
+        end
+        resources = []
+        alternate_ids.uniq.map(&:to_s).each do |id|
+          begin
+            af_object = ActiveFedora::Base.find(id)
```
no-reply: Could we reuse the existing bulk find, instead of querying each object individually?

Something like:

```ruby
af_objects = ActiveFedora::Base.find(alternate_ids.map(&:to_s))

return af_objects unless use_valkyrie

af_objects.map do |obj|
  @resource_factory.to_resource(object: af_object)
end
```

app/custom_queries/hyrax/wings/custom_query_service.rb
laritakr: I'm not sure how it avoids loading anything, but it does make the code easier to read.

app/services/hyrax/custom_queries/wings.rb
```diff
@@ -0,0 +1,37 @@
...
+      # implements a combination of two Valkyrie queries:
+      # => find_many_by_ids & find_by_alternate_identifier
+      # @param [Enumerator<#to_s>] ids
+      # @param [boolean] defaults to true; optionally return ActiveFedora::Base objects if false
+      # @return [Array<Valkyrie::Resource>, Array<ActiveFedora::Base>]
+      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: true)
+        af_objects = ActiveFedora::Base.find(alternate_ids.map(&:to_s))
+        return af_objects unless use_valkyrie == true
+        resources = []
+        af_objects.each do |af_object|
```
no-reply: This is going to eagerly loop through all the objects?

Why not:

```ruby
      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: true)
        af_objects = ActiveFedora::Base.find(alternate_ids.map(&:to_s))
        return af_objects unless use_valkyrie == true

        af_objects.map do |af_object|
          resource_factory.to_resource(object: af_object)
        end
```

PR #3845, opened 2019-06-21
Short-circuit the actor stack when the user can't edit works

no-reply:
In the case that a request to the actor stack tries to add a work to a parent
that the current ability cannot edit, we want to avoid doing so.

Previously, we allowed the stack to carry on creating the new work, only to fail
at the end--with an incomplete work. This leaves the application data in an
undesirable state.

Here, we eagerly fail by checking the permissions of the target works before
continuing with the stack. If we don't have the correct permissions, we avoid
creating any works.

@samvera/hyrax-code-reviewers


no-reply:
this would benefit from a backport to `2.1.x`

PR #3846, opened 2019-06-22
WIP [wings] Use valkyrie resources for BatchEditForm

no-reply:
Removes the implicit direct `ActiveFedora` dependency from `BatchEditForm`. This
class now uses Valkyrie objects, except where behavior is inherited from
`Hyrax::Forms::WorkForm`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/batch_edit_form.rb
+++ b/app/forms/hyrax/forms/batch_edit_form.rb
@@ -9,7 +9,9 @@ module Hyrax
       self.required_fields = []
       self.model_class = Hyrax.primary_work_type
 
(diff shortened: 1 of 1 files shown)


no-reply:
Not sure what's going on here. This would be a good one for someone to look into.

no-reply:
We determined that this is an issue with the tests (stubs are against ActiveFedora internals, so the tests are dependent on specific internal calls in AF). I'm picking it back up to work on infrastructure for Valkyrie-native models, and related testing.

VivianChu:
@no-reply - LGTM.  Is there a test for this change? Thanks

PR #3854, opened 2019-06-26
Add save to the valkyrie solr indexing adapter

hackmastera:
Requires running another solr wrapper for local tests; adding another
solr core in circleci

advances #3800

### DIFF
--- a/lib/valkyrie/indexing/solr/indexing_adapter.rb
+++ b/lib/valkyrie/indexing/solr/indexing_adapter.rb
@@ -5,6 +5,7 @@ module Valkyrie
   module Indexing
     module Solr
       class IndexingAdapter
+        COMMIT_PARAMS = { softCommit: true }.freeze
         ##
         # @!attribute [r] connection
         #   @return [RSolr::Client]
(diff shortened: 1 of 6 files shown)


lib/valkyrie/indexing/solr/indexing_adapter.rb
```diff
@@ -21,8 +22,35 @@ def initialize(connection: default_connection, resource_indexer: default_indexer
...
 
+          def persist(resources)
+            documents = resources.map do |resource|
+              solr_document(resource)
+            end
+            add_documents(documents)
+          end
+
+          def solr_document(resource)
+            base_hash(resource).merge(resource_indexer.new(resource: resource).to_solr)
```
no-reply: would it be good to just push the base hash info into the indexer as well?

hackmastera: That's a good idea once we have a base indexer. currently the only indexer is the Null indexer I'm using from Valkyrie. Would you like me to add a base indexer in this PR or migrate this in a later PR when that's introduced?

no-reply: later is fine. approving.

PR #3857, opened 2019-06-26
[wings] Add switch to use valkyrie in the solr service

hackmastera:
advances #3800


hackmastera:
@no-reply updated with a global switch as we discussed. let me know if it's not what you had imagined.

PR #3859, opened 2019-06-27
[wings] Add save_all, delete, and wipe! to solr indexing adapter

hackmastera:
Also adds some testing infrastructure for working with valkyrie resources (`create_for_repository`).

advances #3800

### DIFF
--- a/spec/support/create_for_repository.rb
+++ b/spec/support/create_for_repository.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+# Factories in Factory Bot don't return the value returned via `to_create`.
+# This custom strategy is necessary to enable data mapper patterns in Factory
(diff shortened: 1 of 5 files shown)


spec/support/create_for_repository.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+# Factories in Factory Bot don't return the value returned via `to_create`.
+# This custom strategy is necessary to enable data mapper patterns in Factory
+# Bot. Without it, when you create the resource you get back doesn't have an id.
+#
+# Copied from:
+#   https://github.com/thoughtbot/factory_bot/issues/565
+class CreateStategyForRepositoryPattern
```
no-reply: :+1: thanks for this!

Issue #3860, opened 2019-06-27
FileSet's original_file.id is not indexed in Solr


little9:
### Descriptive summary

Right now you need to get the id from the `FileSet`
object, which is not ideal. There are other samvera gems 
that use this property and their performance could be improved
by having this property in solr. 

### Rationale

For example, the `iiif_manifest` gem queries the
`FileSet` object for this directly, which causes
performance problems when generating a manifest for a
work that has child works.

If you had a `Book` that had child `Page` objects you
need the `original_file.id` for each `Page`'s `FileSet`.

If a `Book` has a thousand pages `::FileSet.original_file.id`
is going to be called 1000 times. Having this property
indexed in solr would allow for a solr query that gets that
information at once with a list of `Work` ids.

### Expected behavior

```ruby
doc = SolrDocument.find('123abc')
doc[:original_file_id_ssi]
>  '1-23-abc'
```

### Actual behavior
```ruby
file_set = ::FileSet.find('123abc').original_file
file_set.id
> '1-23-abc'
```
### Steps to reproduce the behavior

```ruby
doc = SolrDocument.find('123abc')
doc[:original_file_id_ssi]
> nil
```

PR #3864, opened 2019-06-27
Add original_file.id to FileSetIndexer

little9:
Connected to samvera/hyrax#3860

This modifies the `FileSetIndexer` so that
the `original_file.id` of the `FileSet` is
indexed in solr as an indexed single value
string.

This will allow for accessing the property
without using the `FileSet` object directly.

@samvera/hyrax-code-reviewers


no-reply:
@little9 do you want to backport to 2.x?

little9:
@no-reply That would be awesome :+1:

Issue #3869, opened 2019-06-28
Support managing embargos and leases in native Valkyrie


no-reply:
### Descriptive summary

Given an existing `Hyrax::Resource`, it should be possible to manage Embargos and leases.

PR #3870, opened 2019-06-28
[wings] Introduce valkyrie-native Embargos

no-reply:
Adds a `Hyrax::Embargo` as the valkyrie base class for embargos. The existing
Wings embargo handling is updated to use this base class.

A `NullEmbargo` is introduced as a static empty/inactive embargo. Whenever an
embargo is absent in Valkyrie driven code, we should use this `NullEmbargo`
class to represent a non-existent embargo on the object.

Additionally, an `EmbargoManager` is added to handle operations on the Embargo
as it relates to a given `Hyrax::Resource`.

Connected to #3869.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/embargo_manager.rb
+++ b/app/services/hyrax/embargo_manager.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Manages an embargo for a `Hyrax::Resource`
+  class EmbargoManager
+    ##
+    # @!attribute [rw] resource
+    #   @return [Hyrax::Resource]
+    attr_accessor :resource
+
+    ##
+    # @!attribute [r] query_service
+    #   @return [#find_by]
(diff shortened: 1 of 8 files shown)


app/services/hyrax/embargo_manager.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Manages an embargo for a `Hyrax::Resource`
```
hackmastera: Can you say more in this comment about when/how this class is useful? I think I figured out what you're intending but it wasn't immediately obvious to me.

no-reply: could we push this off to a follow up? this class is developing in conjunction with #3870. i suspect it will be easier to document with some usage examples in place.

i've mainly submitted this as-is because of the changes to `ModelTransformer`.

PR #3878, opened 2019-07-05
Custom dashboard sidebar partials rendered using sidebar_partials class attribute

cjcolvar:
We have at least two use cases of adding links into the dashboard sidebar.  This PR is an attempt to create a seam where these partials can be injected without having to override any partials.  Feedback is welcome especially if you know of a better pattern for this.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/dashboard/sidebar/_activity.html.erb
+++ b/app/views/hyrax/dashboard/sidebar/_activity.html.erb
@@ -31,3 +31,7 @@
       <span class="fa fa-bar-chart" aria-hidden="true"></span> <span class="sidebar-action-text"><%= t('hyrax.admin.sidebar.statistics') %></span>
     <% end %>
   <% end %>
+
+  <% Hyrax::DashboardController.sidebar_partials[:activity].each do |partial| %>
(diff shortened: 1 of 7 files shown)


app/views/hyrax/dashboard/sidebar/_activity.html.erb
```diff
@@ -31,3 +31,7 @@
       <span class="fa fa-bar-chart" aria-hidden="true"></span> <span class="sidebar-action-text"><%= t('hyrax.admin.sidebar.statistics') %></span>
     <% end %>
   <% end %>
+
+  <% Hyrax::DashboardController.sidebar_partials[:activity].each do |partial| %>
```
no-reply: what do you think about moving this repeated view code to a separate view?

no-reply:
Looks reasonable to me. Are there existing sidebar elements in base Hyrax that can be refactored to use this injection pattern?

cjcolvar:
@no-reply Sure I could refactor the existing partials since this would be going into 3.0, but I have some reservation since I'm not sure if it is the right pattern to overhaul everything with.  What do you think?

PR #3881, opened 2019-07-11
Fixes HYRAX-#3775

### LINKED ISSUE #3775: Fileset page is viewable to the public (it should be hidden) in mediated review.
### Descriptive summary

Tested on nurax-stable May 14th. 2.4.1 

When depositing a work into a mediated deposit workflow, the work view page is hidden to the public, but you can download the file via direct link to the fileset page.

### Rationale

Fileset page and file download should NOT be visible until it is in the approved state. 

1. deposit a work into an admin set with MEDIATED REVIEW WORKFLOW
2. While logged out Access the deposited work https://nurax-stable.curationexperts.com/concern/generic_works/sn009x97b?locale=en
3. While logged out Access the fileset page via direct URL https://nurax-stable.curationexperts.com/concern/parent/sn009x97b/file_sets/ft848q77t

fritzfreiheit:
## Recommended updates.

1efe93929285985751cc270675c243f628cf31ca

The above title is from the original commit.  What follows is the
description put in the PR message:

Add parent_presenter to file_set_presenter and use it to enforce access
restrictions from workflow mediated review. If public view not available
generate an unavailable page similar to a work's unavailable page.

Fixes #3775 ; refs #3881

Update to prevent direct navigation to file sets under works that are
public but restricted by workflow. When a user navigates to public file
sets in a workflow without and that user does not have edit privileges,
an unavailable screen is displayed.  From a work in the workflow that is
pending review, the file sets will not be
editable/deletable/downloadable.  Direct navigation to a file set for
users without edit privileges will result in an unavailable screen.

Guidance for testing, such as acceptance criteria or new user interface
behaviors:

* As a non-admin user, create a work with an attached file and place it
  in a mediate workflow (such as one step mediated workflow)
* From the work, you should not be able to edit/delete/download the file
* Navigate to the file set, it will show an unavailable file set
* Copy the url of file set
* Log out
* Directly navigate to the file set url, it will show unavailable

## Updating spec to use decorator

4fccf2b57ea282e4da0a57a77c7cecaae1675e09

This commit builds on the rebased commit of
e9697e4641ac50068fbda72e0bed94fc8adbdba4

The rebased commit was from an aging PR (#3881) that addressed #3775.
In the time between the submission of the original PR and the
revisitation, we updated some underlying aspects of Hyrax.

This is a quick fix to get a build working.  I'll rely on reviewers to
determine if its appropriate.

A question remains, should our `Hyrax::SolrDocument::OrderedMembers`
"be_a" SolrDocument?

## Favoring method usage instead of instance_variable

d6b83ad87b83514583aa2ff5536bc8061e09ddcf

It's odd to have both a method and a like named instance variable.  This
tidies that up.

## Addressing deprecation warning for success?

5a011e0a616d69a34570f113cf5cd8201d3b13ec

```
DEPRECATION WARNING: The success? predicate is deprecated and will
be removed in Rails 6.0. Please use successful? as provided by
Rack::Response::Helpers. (called from block (4 levels) in
<top (required)> at
./spec/controllers/hyrax/file_sets_controller_spec.rb:435)
```

## Compressing logic around workflow restriction

5463ea0aef511e3988783bf73cc754bfcb790c91

This commit attempts to normalize and consolidate the underlying logic
introduced in the Hyrax::FileSetsController.

As commented, there are application states in which we call the
guard_for_workflow_restriction_on! method:

- show action: when we start from a `presenter`
- edit/update/delete: when we start from a `curation_concern`

Both of those objects have a `parent` object but of different types.
The guard_for_workflow_restrion_on! now handles both of those parent
object types.

In doing so, the method begins to look like another method introduced in
a presenter (as noted).

To ensure that this works in the controller, I added the `#suppressed?`
method to the WorkShowPresenter.  Perhaps an oversight in delegation.

## Favoring helper method over presenter method

35f7d6f090e10256031dccdf00e888abd331e200

In discussions with @no_reply, I'm looking to trim the size of
presenters.  As envisioned, presenters should be more about gathering up
data to present.  Helpers shine when used for conditional rendering.

Since we don't ask presenters to "render themself" this inflection makes
sense; Let the helpers decide render logic.

This is an interstitial commit.  It preserves the backwards behavior.

## Fleshing out logic of workflow_restriction? helper

1a2e0ba05ce3df0e5fb06f6b7a87ecab35a99d41

As I noodle on this, I'm fairly certain we should be registering a
CanCan ability check.  I believe in promoting this to a helper method it
will be easier to incorporate this into an ability.

However, that is presently outside the scope of what I consider
necessary.

## Swiching Hyrax::FileSetsController to use helper

cb21570fadcea0b8d1dfd0b7cffecf5135c1ea76

The echoed the logic of `Hyrax::Workflows#workflow_restriction?`, this
change leverages that single source of logic for evaluation.

## Refactoring view logic to use helper method

cbdf4f6eabba0f131d4983075e11f8b629788f22

This also involves introducing a deprecation on a single use method.

## Fixing broken specs based on prior refactors

5d19677e8d668270ef82d3a10c1e559001120ad6

## Echoing past logic around `@presenter.try(:parent)`

38b3640a0467f818ab48cec03f646531e8213bb4

The `./spec/features/batch_edit_spec.rb` failed as the SolrDocuments did
not have a parent concept.

I'm uncertain what to do about this, but am offering it up for consideration.

### DIFF
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 module Hyrax
   class FileSetsController < ApplicationController
+    rescue_from WorkflowAuthorizationException, with: :render_unavailable
+
     include Blacklight::Base
     include Blacklight::AccessControls::Catalog
     include Hyrax::Breadcrumbs
@@ -42,6 +44,7 @@ module Hyrax
 
     # GET /concern/parent/:parent_id/file_sets/:id
     def show
+      workflow_check(parent_id: params[:parent_id])
       respond_to do |wants|
         wants.html { presenter }
         wants.json { presenter }
@@ -52,12 +55,15 @@ module Hyrax
     # DELETE /concern/file_sets/:id
     def destroy
       parent = curation_concern.parent
+      workflow_check(parent_id: parent.id)
       actor.destroy
       redirect_to [main_app, parent], notice: view_context.t('hyrax.file_sets.asset_deleted_flash.message')
     end
 
     # PATCH /concern/file_sets/:id
     def update
+      parent = curation_concern.parent
+      workflow_check(parent_id: parent.id)
       if attempt_update
         after_update_response
       else
@@ -146,9 +152,22 @@ module Hyrax
 
     def initialize_edit_form
       @parent = @file_set.in_objects.first
+      workflow_check(parent_id: @parent.id)
       original = @file_set.original_file
       @version_list = Hyrax::VersionListPresenter.new(original ? original.versions.all : [])
       @groups = current_user.groups
+    end
+
+    def workflow_check(parent_id: nil)
+      return if current_ability.can?(:edit, @solr_document)
+      if parent_id.nil?
+        curation_concern = ::FileSet.find(params[:id])
+        parent = curation_concern.parent
+        parent_id = parent.id
+      end
+      doc = ::SolrDocument.find(parent_id)
+      return if current_ability.can?(:edit, doc)
+      raise WorkflowAuthorizationException if doc.suppressed?
     end
 
     def actor
@@ -161,8 +180,10 @@ module Hyrax
 
     def presenter
       @presenter ||= begin
-        show_presenter.new(curation_concern_document, current_ability, request)
-      end
+                       presenter = show_presenter.new(curation_concern_document, current_ability, request)
+                       raise WorkflowAuthorizationException if presenter.parent.blank?
+                       presenter
+                     end
     end
 
     def curation_concern_document
@@ -199,6 +220,36 @@ module Hyrax
                  'dashboard'
                end
       File.join(theme, layout)
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def render_unavailable
+      message = I18n.t("hyrax.workflow.unauthorized_parent")
+      respond_to do |wants|
+        wants.html do
+          unavailable_presenter
+          flash[:notice] = message
+          render 'unavailable', status: :unauthorized
+        end
+        wants.json do
+          render plain: message, status: :unauthorized
+        end
+        additional_response_formats(wants)
+        wants.ttl do
+          render plain: message, status: :unauthorized
+        end
+        wants.jsonld do
+          render plain: message, status: :unauthorized
+        end
+        wants.nt do
+          render plain: message, status: :unauthorized
+        end
+      end
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    def unavailable_presenter
+      @presenter ||= show_presenter.new(::SolrDocument.find(params[:id]), current_ability, request)
     end
   end
 end
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -44,10 +44,11 @@ module Hyrax
 
     # GET /concern/parent/:parent_id/file_sets/:id
     def show
-      workflow_check(parent_id: params[:parent_id])
+      presenter
+      guard_for_workflow_restriction_on!(parent: presenter.parent)
       respond_to do |wants|
-        wants.html { presenter }
-        wants.json { presenter }
+        wants.html
+        wants.json
         additional_response_formats(wants)
       end
     end
@@ -55,7 +56,7 @@ module Hyrax
     # DELETE /concern/file_sets/:id
     def destroy
       parent = curation_concern.parent
-      workflow_check(parent_id: parent.id)
+      guard_for_workflow_restriction_on!(parent: parent)
       actor.destroy
       redirect_to [main_app, parent], notice: view_context.t('hyrax.file_sets.asset_deleted_flash.message')
     end
@@ -63,7 +64,7 @@ module Hyrax
     # PATCH /concern/file_sets/:id
     def update
       parent = curation_concern.parent
-      workflow_check(parent_id: parent.id)
+      guard_for_workflow_restriction_on!(parent: parent)
       if attempt_update
         after_update_response
       else
@@ -152,22 +153,35 @@ module Hyrax
 
     def initialize_edit_form
       @parent = @file_set.in_objects.first
-      workflow_check(parent_id: @parent.id)
+      guard_for_workflow_restriction_on!(parent: @parent)
       original = @file_set.original_file
       @version_list = Hyrax::VersionListPresenter.new(original ? original.versions.all : [])
       @groups = current_user.groups
     end
 
-    def workflow_check(parent_id: nil)
-      return if current_ability.can?(:edit, @solr_document)
-      if parent_id.nil?
-        curation_concern = ::FileSet.find(params[:id])
-        parent = curation_concern.parent
-        parent_id = parent.id
-      end
-      doc = ::SolrDocument.find(parent_id)
-      return if current_ability.can?(:edit, doc)
-      raise WorkflowAuthorizationException if doc.suppressed?
+    # @see Hyrax::WorkShowPresenter#workflow_restriction?
+    # @see Hyrax::FileSetShowPresenter#workflow_restriction?
+    # @see #presenter
+    #
+    # Note, this method echoes the logic of the #workflow_restriction?
+    # in presenter classes.  My preference would be to rely on that
+    # method.  However, in specs the :parent parameter is of two
+    # different types:
+    #
+    # - Hyrax::WorkShowPresenter (for show actions)
+    # - GenericWork (for edit/delete/update actions)
+    #
+    # @todo How is this related to the presenter logic around building the presenter?
+    #
+    # @param parent [Hyrax::WorkShowPresenter, GenericWork, #suppressed?] an
+    #        object on which we check if the current can take action.
+    #
+    # @return true if we did not encounter any workflow restrictions
+    # @raise WorkflowAuthorizationException if we encountered some workflow_restriction
+    def guard_for_workflow_restriction_on!(parent:)
+      return true if current_ability.can?(:edit, parent)
+      return true unless parent.suppressed?
+      raise WorkflowAuthorizationException
     end
 
     def actor
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -159,28 +159,14 @@ module Hyrax
       @groups = current_user.groups
     end
 
-    # @see Hyrax::WorkShowPresenter#workflow_restriction?
-    # @see Hyrax::FileSetShowPresenter#workflow_restriction?
-    # @see #presenter
-    #
-    # Note, this method echoes the logic of the #workflow_restriction?
-    # in presenter classes.  My preference would be to rely on that
-    # method.  However, in specs the :parent parameter is of two
-    # different types:
-    #
-    # - Hyrax::WorkShowPresenter (for show actions)
-    # - GenericWork (for edit/delete/update actions)
-    #
-    # @todo How is this related to the presenter logic around building the presenter?
-    #
+    include WorkflowsHelper # Provides #workflow_restriction?, and yes I mean include not helper; helper exposes the module methods
     # @param parent [Hyrax::WorkShowPresenter, GenericWork, #suppressed?] an
     #        object on which we check if the current can take action.
     #
     # @return true if we did not encounter any workflow restrictions
     # @raise WorkflowAuthorizationException if we encountered some workflow_restriction
     def guard_for_workflow_restriction_on!(parent:)
-      return true if current_ability.can?(:edit, parent)
-      return true unless parent.suppressed?
+      return true unless workflow_restriction?(parent, ability: current_ability)
       raise WorkflowAuthorizationException
     end
 
--- a/app/presenters/hyrax/file_set_presenter.rb
+++ b/app/presenters/hyrax/file_set_presenter.rb
@@ -34,6 +34,10 @@ module Hyrax
              :original_file_id,
              to: :solr_document
 
+    def workflow
+      nil
+    end
+
     def single_use_links
       @single_use_links ||= SingleUseLink.where(item_id: id).map { |link| link_presenter_class.new(link) }
     end
@@ -93,6 +97,10 @@ module Hyrax
       current_ability.can?(:edit, id) || current_ability.can?(:destroy, id) || current_ability.can?(:download, id)
     end
 
+    def workflow_restriction?
+      parent_presenter.try(:workflow_restriction?)
+    end
+
     private
 
     def link_presenter_class
@@ -103,6 +111,11 @@ module Hyrax
       ids = Hyrax::SolrService.query("{!field f=member_ids_ssim}#{id}", fl: Hyrax.config.id_field)
                               .map { |x| x.fetch(Hyrax.config.id_field) }
       Hyrax.logger.warn("Couldn't find a parent work for FileSet: #{id}.") if ids.empty?
+      ids.each do |id|
+        doc = ::SolrDocument.find(id)
+        next if current_ability.can?(:edit, doc)
+        raise WorkflowAuthorizationException if doc.suppressed? && current_ability.can?(:read, doc)
+      end
       Hyrax::PresenterFactory.build_for(ids: ids,
                                         presenter_class: WorkShowPresenter,
                                         presenter_args: current_ability).first
--- a/app/presenters/hyrax/file_set_presenter.rb
+++ b/app/presenters/hyrax/file_set_presenter.rb
@@ -94,11 +94,8 @@ module Hyrax
     end
 
     def user_can_perform_any_action?
+      Deprecation.warn("We're removing Hyrax::FileSetPresenter.user_can_perform_any_action? in Hyrax 4.0.0; Instead use can? in view contexts.")
       current_ability.can?(:edit, id) || current_ability.can?(:destroy, id) || current_ability.can?(:download, id)
-    end
-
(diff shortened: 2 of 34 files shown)


.gitignore
```diff
@@ -22,6 +22,9 @@ tmtags
 \#*
 .\#*
 
+## RubyMine
```
no-reply: if you'd like to include this, could we make it a separate PR? it doesn't seem connected to this change.

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -33,26 +35,36 @@ class FileSetsController < ApplicationController
     # GET /concern/file_sets/:id
     def edit
       initialize_edit_form
+    rescue WorkflowAuthorizationException
```
no-reply: this this redundant with line 3?

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -33,26 +35,36 @@ class FileSetsController < ApplicationController
...
 
     # GET /concern/parent/:parent_id/file_sets/:id
     def show
+      workflow_check(parent_id: params[:parent_id])
       respond_to do |wants|
         wants.html { presenter }
         wants.json { presenter }
         additional_response_formats(wants)
       end
+    rescue WorkflowAuthorizationException
```
no-reply: as in https://github.com/samvera/hyrax/pull/3881/files#r305449920

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -33,26 +35,36 @@ class FileSetsController < ApplicationController
...
+      render_unavailable
     end
 
     # DELETE /concern/file_sets/:id
     def destroy
       parent = curation_concern.parent
+      workflow_check(parent_id: parent.id)
       actor.destroy
       redirect_to [main_app, parent], notice: view_context.t('hyrax.file_sets.asset_deleted_flash.message')
+    rescue WorkflowAuthorizationException
```
no-reply: as in https://github.com/samvera/hyrax/pull/3881/files#r305449920

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -62,6 +74,8 @@ def update
       flash[:error] = error.message
       logger.error "FileSetsController::update rescued #{error.class}\n\t#{error.message}\n #{error.backtrace.join("\n")}\n\n"
       render action: 'edit'
+    rescue WorkflowAuthorizationException
```
no-reply: as in https://github.com/samvera/hyrax/pull/3881/files#r305449920

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -137,11 +151,24 @@ def search_builder_class
...
+      def workflow_check(parent_id: nil)
+        return if current_ability.can?(:edit, @solr_document)
+        if parent_id.nil?
+          curation_concern = ::FileSet.find(params[:id])
+          parent = curation_concern.parent
+          parent_id = parent.id
+        end
+        doc = ::SolrDocument.find(parent_id)
+        return if current_ability.can?(:edit, doc)
+        raise WorkflowAuthorizationException if doc.suppressed? && current_ability.can?(:read, doc)
```
no-reply: i'm not following this line. do we need to check `current_ability.can?(:read, doc)`?

what happens if ` doc.suppressed?` is `true` but the current ability *cannot* read the doc?

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -155,7 +182,9 @@ def presenter
           _, document_list = search_results(params)
           curation_concern = document_list.first
           raise CanCan::AccessDenied unless curation_concern
-          show_presenter.new(curation_concern, current_ability, request)
+          pres = show_presenter.new(curation_concern, current_ability, request)
+          raise WorkflowAuthorizationException if pres.parent_presenter.nil?
```
no-reply: `raise WorkflowAuthorizationException if pres.parent_presenter.blank?` might be more robust?

app/views/hyrax/file_sets/_show_actions.html.erb
```diff
@@ -3,7 +3,8 @@
     <%= link_to t('.analytics'), @presenter.stats_path, id: 'stats', class: 'btn btn-default' %>
   <% end %>
 
-  <% if @presenter.editor? %>
+  <% no_workflow_restrictions = !( @presenter.respond_to?( :workflow_restriction? ) and @presenter.workflow_restriction? ) %>
```
no-reply: In what cases does `@presenter` not respond to `#workflow_restriction?`? Would it be possible to just ensure that all appropriate presenters have this method?

Alternatively `<% if @presenter.editor? && !@presenter.try(:workflow_restriction?) %>` could be easier to follow.

Do we need a unit test for the new case in https://github.com/samvera/hyrax/pull/3881/files#diff-c82771edaa1f35232e9c45b621988f7c?

no-reply: Oh, i'm just realizing how many times this `no_workflow_restrictions` code is being repeated. Let's definitely figure out how to move this into the presenters. Duplicating this code in many views is going to be very fragile, I fear.

config/locales/hyrax.en.yml
```diff
@@ -1248,6 +1248,7 @@ en:
       load:
         state_error: 'The workflow: %{workflow_name} has not been updated.  You are removing a state: %{state_name} with %{entity_count} entity/ies.  A state may not be removed while it has active entities!'
       unauthorized: The work is not currently available because it has not yet completed the approval process
+      unauthorized_parent: The file is not currently available because its parent work has not yet completed the approval process
```
no-reply: @samvera/hyrax-repo-managers do any of you have feedback about this text?

spec/controllers/hyrax/file_sets_controller_spec.rb
```diff
@@ -68,8 +68,14 @@
     end
 
     describe "#update" do
+      let(:parent) do
+        create(:work, :public, user: user)
+      end
       let(:file_set) do
-        create(:file_set, user: user)
+        create(:file_set, user: user).tap do |file_set|
```
no-reply: instead of explicitly creating the `parent` and manually saving the fileset as a member, could we use the `:file_with_work` factory?

```ruby
let(:file_set) do
  create(:file_with_work, user: user)
end
```

?

no-reply: Same question for instances of this below.

app/views/hyrax/base/unavailable.html.erb
```diff
@@ -1,5 +1,5 @@
 <h1>
-    <%= @presenter %> <%= @presenter.workflow.badge %>
+    <%= @presenter %> <%= @presenter.workflow.badge if @presenter.respond_to?(:workflow) && @presenter.workflow.present? %>
```
no-reply: When does `@presenter` not respond to `#workflow`?

It might be nicer just to ensure that `#workflow` and `#badge` produce reasonable defaults in the null case. Otherwise, perhaps `@presenter.try(:workflow).try(:badge)` would be more concise and easier to read.

no-reply (dismissed):
I haven't quite made it all the way through this, but I think some of it needs tidying up.

I've added some initial comments.

spec/controllers/hyrax/file_sets_controller_spec.rb
fritzfreiheit: On your recommendation I tried this, but it doesn't add the file set to the ordered members.

app/controllers/hyrax/file_sets_controller.rb
fritzfreiheit: It is. I had problems with this at first and thought it was a bug with 'rescue_from' (which I researched), but it isn't. Removed.

.gitignore
fritzfreiheit: Will do.

app/controllers/hyrax/file_sets_controller.rb
fritzfreiheit: It is. I had problems with this at first and thought it was a bug with 'rescue_from' (which I researched), but it isn't. Removed.

app/controllers/hyrax/file_sets_controller.rb
fritzfreiheit: It is. I had problems with this at first and thought it was a bug with 'rescue_from' (which I researched), but it isn't. Removed.

app/controllers/hyrax/file_sets_controller.rb
fritzfreiheit: I copied this code from elsewhere. And yes, it looks like it is redundant and/or misleading. I've removed the trailing `current_ability.can?(:read, doc)`

app/controllers/hyrax/file_sets_controller.rb
fritzfreiheit: Done.

app/views/hyrax/file_sets/_show_actions.html.erb
fritzfreiheit: I've updated the code to rely on a method on the presenters, but not all presenters have this method, and there are a number of spec tests that don't bother with presenters and I chose not track them all down and fix and instead used a try on the method call.

spec/controllers/hyrax/file_sets_controller_spec.rb
fritzfreiheit: The create file with work does not seem to add the file set to the works ordered members.

app/views/hyrax/base/unavailable.html.erb
fritzfreiheit: Did the presenter try try as suggested.

fritzfreiheit:
@no-reply / @jeremyf -- Please review and approve.

jeremyf:
## Updating spec to use decorator

53a305b82ee573b985d5988900f46d148edf42e7

This commit builds on the rebased commit of
e9697e4641ac50068fbda72e0bed94fc8adbdba4

The rebased commit was from an aging PR (#3881) that addressed #3775.
In the time between the submission of the original PR and the
revisitation, we updated some underlying aspects of Hyrax.

This is a quick fix to get a build working.  I'll rely on reviewers to
determine if its appropriate.

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -147,11 +153,24 @@ def search_builder_class
 
     def initialize_edit_form
       @parent = @file_set.in_objects.first
+      workflow_check(parent_id: @parent.id)
       original = @file_set.original_file
       @version_list = Hyrax::VersionListPresenter.new(original ? original.versions.all : [])
       @groups = current_user.groups
     end
 
+    def workflow_check(parent_id: nil)
```
jeremyf, line 162: I'm contemplating the value of extracting this as a Service class.

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -162,11 +181,13 @@ def attributes
 
     def presenter
       @presenter ||= begin
-        _, document_list = search_results(params)
-        curation_concern = document_list.first
-        raise CanCan::AccessDenied unless curation_concern
-        show_presenter.new(curation_concern, current_ability, request)
-      end
+                       _, document_list = search_results(params)
```
jeremyf, line 184: There's some echoes of commonality between this and the above comment.  Not a blocker, but one I'm noodling on.

app/presenters/hyrax/file_set_presenter.rb
```diff
@@ -100,9 +112,15 @@ def link_presenter_class
     def fetch_parent_presenter
```
jeremyf, line 112: This is looking like we may want a service object; There's a lot of logic packed into this.

app/presenters/hyrax/file_set_presenter.rb
```diff
@@ -100,9 +112,15 @@ def link_presenter_class
...
                               .map { |x| x.fetch(Hyrax.config.id_field) }
-      Hyrax::PresenterFactory.build_for(ids: ids,
-                                        presenter_class: WorkShowPresenter,
-                                        presenter_args: current_ability).first
+      ids.each do |id|
+        doc = ::SolrDocument.find(id)
+        next if current_ability.can?(:edit, doc)
+        raise WorkflowAuthorizationException if doc.suppressed? && current_ability.can?(:read, doc)
+      end
+      rv = Hyrax::PresenterFactory.build_for(ids: ids,
```
no-reply, line 120: is this variable assignment doing something?

app/controllers/hyrax/file_sets_controller.rb
no-reply: i feel like it would really help to extract this to a service. not least: it would make the exact behavior of this check easier to test robustly and and much easier to change later if we didn't get it perfect this time.

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -147,11 +153,24 @@ def search_builder_class
...
+      workflow_check(parent_id: @parent.id)
       original = @file_set.original_file
       @version_list = Hyrax::VersionListPresenter.new(original ? original.versions.all : [])
       @groups = current_user.groups
     end
 
+    def workflow_check(parent_id: nil)
+      return if current_ability.can?(:edit, @solr_document)
+      if parent_id.nil?
+        curation_concern = ::FileSet.find(params[:id])
```
no-reply, line 165: i'm *slightly* worried about the cost of this lookup, and not easily understanding when it's triggered.

in what cases would i pass a `nil` argument here? could this branch be separated and called explicitly, instead of branching on the `.nil?` type check?

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -38,6 +40,7 @@ def edit
 
     # GET /concern/parent/:parent_id/file_sets/:id
     def show
+      workflow_check(parent_id: params[:parent_id])
```
no-reply, line 43: do we need to check workflow permissions on `show`?

app/views/hyrax/file_sets/media_display/_office_document.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if Hyrax.config.display_media_download_link? %>
+<% if Hyrax.config.display_media_download_link? && !@presenter.try(:workflow_restriction?) %>
```
no-reply, line 1: is this check a candidate for a helper method?

app/presenters/hyrax/file_set_presenter.rb
jeremyf: I'm removing that local variable in a later commit.

jeremyf:
## Updating spec to use decorator

328f7f4f591cd3339ab7722683adc0ce68d328c0

This commit builds on the rebased commit of
e9697e4641ac50068fbda72e0bed94fc8adbdba4

The rebased commit was from an aging PR (#3881) that addressed #3775.
In the time between the submission of the original PR and the
revisitation, we updated some underlying aspects of Hyrax.

This is a quick fix to get a build working.  I'll rely on reviewers to
determine if its appropriate.

A question remains, should our `Hyrax::SolrDocument::OrderedMembers`
"be_a" SolrDocument?

## Removing unnecessary local variable

cf2d5ff1657be93029c38dd4426ec91610006cb4


## Favoring method usage instead of instance_variable

e02f3755f66a4492d7d6971e7bf84bddcdee4c7a

It's odd to have both a method and a like named instance variable.  This
tidies that up.

## Addressing deprecation warning for success?

f06d5ecacb884e332bd47089f7299c7106688c60

```
DEPRECATION WARNING: The success? predicate is deprecated and will
be removed in Rails 6.0. Please use successful? as provided by
Rack::Response::Helpers. (called from block (4 levels) in
<top (required)> at
./spec/controllers/hyrax/file_sets_controller_spec.rb:435)
```

## Compressing logic around workflow restriction

09451a943164cfc7fc3eba5c31a493d33bf94075

This commit attempts to normalize and consolidate the underlying logic
introduced in the Hyrax::FileSetsController.

As commented, there are application states in which we call the
guard_for_workflow_restriction_on! method:

- show action: when we start from a `presenter`
- edit/update/delete: when we start from a `curation_concern`

Both of those objects have a `parent` object but of different types.
The guard_for_workflow_restrion_on! now handles both of those parent
object types.

In doing so, the method begins to look like another method introduced in
a presenter (as noted).

To ensure that this works in the controller, I added the `#suppressed?`
method to the WorkShowPresenter.  Perhaps an oversight in delegation.

app/controllers/hyrax/file_sets_controller.rb
jeremyf: In my reading of #3775 (see block quote), I believe it is required.

> When depositing a work into a mediated deposit workflow, the work view page is hidden to the public, but you can download the file via direct link to the fileset page.

jeremyf:
@no-reply I pushed up some refinements that helped me find a very common thread.  I have not yet begun the service class extraction because I want to make sure I'm moving in the right direction.

Maybe we can hop on a call and hash this out?

jeremyf:
Also, I have updated the PR description to include commit logs to help contextualize the entire commit.

jeremyf:
Ugh, the build broke because of precarious test constructs.  I'm reviewing now.  I suspect the Service class is going to be more important to insulate against these banana-grams.

jlhardes:
@no-reply you assigned this to yourself recently. Is there anything I can do to help resolve this?

jlhardes:
This is still ready for review.

no-reply:
thanks. i'm going to try to get through this today.

dlpierce (approved):

PR #3892, opened 2019-07-22
Don't short-circut actor Environment when interpreting visibility

no-reply:
`InterpretVisibilityActor` previously short-circuited `env` by initializing a completely new copy for use further down the stack. Changes to the environment after that point would fail to have any impact further up the stack, where actors are already using the older `env`.

The original reason the `InterpretVisibilityActor` did this was to alter the attributes in batch; i.e. this was a work around for `#attributes=`. We add support for that method, then we can use it instead.

The `CreateFilesActor`, by contrast, was counting on those attributes not to change. Since any actor further down the stack may change them, it's better to cache them immediately.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/create_with_files_actor.rb
+++ b/app/actors/hyrax/actors/create_with_files_actor.rb
@@ -6,16 +6,26 @@ module Hyrax
       # @return [Boolean] true if create was successful
       def create(env)
         uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
-        files = uploaded_files(uploaded_file_ids)
-        validate_files(files, env) && next_actor.create(env) && attach_files(files, env)
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+
+        validate_files(files, env) &&
+          next_actor.create(env) &&
+          attach_files(files, env.curation_concern, attributes)
       end
 
       # @param [Hyrax::Actors::Environment] env
       # @return [Boolean] true if update was successful
       def update(env)
         uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
-        files = uploaded_files(uploaded_file_ids)
-        validate_files(files, env) && next_actor.update(env) && attach_files(files, env)
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+
+        validate_files(files, env) &&
+          next_actor.update(env) &&
+          attach_files(files, env.curation_concern, attributes)
       end
 
       private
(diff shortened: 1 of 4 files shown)


app/actors/hyrax/actors/create_with_files_actor.rb
```diff
@@ -6,16 +6,26 @@ class CreateWithFilesActor < Hyrax::Actors::AbstractActor
       # @return [Boolean] true if create was successful
       def create(env)
         uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
-        files = uploaded_files(uploaded_file_ids)
-        validate_files(files, env) && next_actor.create(env) && attach_files(files, env)
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
```
hweng: `#clone` creates a shallow copy of an `env.attributes` object, which means if the `attributes` object contains other objects, then those objects will be shared not copied so it still can be mutated.
 What is the structure of the `attributes`?   I'm not familiar with the codebase yet, maybe it is fine.

no-reply: in general, mutations to the attributes in the stack amount to addition/deletion of key/value pairs. still, it might not hurt to use `#deep_dup` just to be safe. i'll make the change.

hweng: Yeah it's safer to use other approach.
I tired key/value pair (hash type). The original value can be mutated at deep level if using `#clone`:

```
irb(main):008:0> hsh1 = { k1: "a", k2: "b" }
=> {:k1=>"a", :k2=>"b"}
irb(main):009:0> hsh2 = hsh1.clone
=> {:k1=>"a", :k2=>"b"}
irb(main):010:0> hsh2[:k1].upcase!
=> "A"
irb(main):011:0> hsh1
=> {:k1=>"A", :k2=>"b"}
```

hweng: But if it is only addition/deletion, it might be fine.

no-reply:
@hweng i pushed the change to use `#deep_dup`

hweng:
@no-reply `#deep_copy` failed on CircleCI . Do you want to try `Marshal.load( Marshal.dump(env.attributes) )` ?  It works to make deep copies of hash, maybe not perfect though:

```
irb(main):001:0> hsh1 = { k1: "a", k2: "b" }
=> {:k1=>"a", :k2=>"b"}
irb(main):002:0> hsh2 = Marshal.load( Marshal.dump(hsh1) )
=> {:k1=>"a", :k2=>"b"}
irb(main):003:0> hsh2[:k1].upcase!
=> "A"
irb(main):004:0> hsh1
=> {:k1=>"a", :k2=>"b"}
```

Or if you think there is only addition/deletion in the stack, it might be fine to just keep `#clone`.

no-reply:
blergh, yeah. let's go ahead and use `#clone` for now. i think there will be an opportunity to fix this up further later in the valkyrie transition. for now, this seems like an improvement from the status quo even with `#clone`.

hweng:
Yeah, approve for now!

PR #3895, opened 2019-07-23
[wings] Use the model's id minting logic in the transformer

no-reply:
The ActiveFedora model provides an `#assign_id` method, which can be used
directly instead of reaching into the id minter for assigning ids. This
reinstates compatibility between wings and models that are using Hyrax's
prescribed interfaces for minting identifiers.

By default, this only mints ids when the noid minter is enabled.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/model_transformer.rb
+++ b/lib/wings/model_transformer.rb
@@ -66,9 +66,13 @@ module Wings
       klass = ResourceClassCache.instance.fetch(pcdm_object) do
         self.class.to_valkyrie_resource_class(klass: pcdm_object.class)
       end
-      pcdm_object.id = minted_id if pcdm_object.id.nil?
+
+      mint_id unless pcdm_object.id
+
       attrs = attributes.tap { |hash| hash[:new_record] = pcdm_object.new_record? }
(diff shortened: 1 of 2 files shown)


lib/wings/model_transformer.rb
```diff
@@ -201,8 +205,9 @@ def self.run(obj, keys)
 
     private
 
-      def minted_id
-        ::Noid::Rails.config.minter_class.new.mint
+      def mint_id
```
cjcolvar: I know this is a private method but do you mind putting in some comments to explain what it is doing and what it is supposed to be returning?  I find line 210 hard to read and understand.

no-reply: does the updated version, without the control-flow `&&`, read more clearly?

cjcolvar: I think this is helpful.  I guess `assign_id` is the source of my confusion and that carries over into this method.

no-reply: yeah, it's unfortunate that `#assign_id` can return `nil`. especially since `#id=` errors when `nil` is passed. but there we are :|

PR #3897, opened 2019-07-24
Add an actor stack spec for persisting permissions

no-reply:
The actor stack should persist arbitrary permissions. This behavior is untested
elsewhere, and may be at risk during the transition to Wings.

* No testing needed; this simply adds an integration test.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/actor_stack_spec.rb
+++ b/spec/features/actor_stack_spec.rb
@@ -55,6 +55,18 @@ RSpec.describe Hyrax::DefaultMiddlewareStack, :clean_repo do
       end
     end
 
+    context 'when adding permissions' do
+      before do
+        work.permissions.build(name: 'discover_user', type: 'person', access: 'discover')
+      end
+
+      it 'persists arbitrary ACL permissions' do
+        expect { actor.create(env) }
+          .to change { env.curation_concern.permissions }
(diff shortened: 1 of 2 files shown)


spec/features/actor_stack_spec.rb
```diff
@@ -55,6 +55,18 @@ def create(env)
...
 
+    context 'when adding permissions' do
+      before do
+        work.permissions.build(name: 'user2', type: 'person', access: 'discover')
+      end
+
+      it 'persists arbitrary ACL permissions' do
+        expect { actor.create(env) }
+          .to change {  env.curation_concern.permissions.map(&:mode).map(&:first) }
+          .to include RDF::URI('http://projecthydra.org/ns/auth/acl#Discover')
```
cjcolvar: Would it make sense to make this test a bit more strict by checking for the name and type along with the mode?  I'm afraid of an error that would masked by having a different discover permission already on the object.

no-reply: probably a matcher is a good idea.

no-reply: @cjcolvar what do you think about the updated version with permission matcher?

cjcolvar: Looks good to me!

PR #3899, opened 2019-07-25
Support and test with Rails 5.2

cjcolvar:


cjcolvar:
@no-reply I had to split things into two separate workflows.  It would be good to work through all of the deprecation warnings but that can be done in a follow up PR.

dlim87:
Going to leave this in this column until the end, since doing a manual test run is prohibitively expensive. I'll run a the full test plan once all the work for 3.0 is done

PR #3900, opened 2019-07-25
Adds Valkyrie native permissions and ACL handling

no-reply:
Adds `Hyrax::Permission` and `Hyrax::AccessControlList` as an initial implementation of ACLs and read/edit/discover grants for objects in Valkyrie.

A simple grant/revoke DSL is added to encapsulate the most common practice of granting or revoking a particular access mode for a given user or group.

For now, this is only working against the abstract Valkyrie interface; further work is needed to allow the Wings adapter to translate the Valkyrie native approach to `ActiveFedora`-compatible data/operations.

This moves work forward toward #3584.

* Build out low-level infrastructure for Valkyrie handling of permissions

* none; impact of this work is limited to internals.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/access_control_list.rb
+++ b/app/services/hyrax/access_control_list.rb
@@ -3,6 +3,9 @@
 module Hyrax
   ##
   # ACLs for `Hyrax::Resource` models
+  #
+  # Allows managing `Hyrax::Permission` entries referring to a specific
+  # `Hyrax::Resource` using a simple add/delete model.
   class AccessControlList
     ##
     # @!attribute [rw] resource
@@ -16,6 +19,8 @@ module Hyrax
 
     ##
     # @param resource [Valkyrie::Resource]
+    # @param persister [#save]
+    # @param query_service [#find_inverse_references_by]
     def initialize(resource:, persister: Hyrax.persister, query_service: Hyrax.query_service)
       self.resource  = resource
       @persister     = persister
@@ -47,6 +52,15 @@ module Hyrax
     end
 
     ##
+    # @example
+    #    user = User.find('user_id')
+    #
+    #    acl.grant(:read).to(user)
+    def grant(mode)
+      ModeGrant.new(self, mode)
+    end
+
+    ##
     # @return [Boolean]
     def pending_changes?
       additions.any? || deletions.any?
@@ -58,6 +72,15 @@ module Hyrax
       Set.new(query_service.find_inverse_references_by(resource: resource, property: :access_to)) -
         deletions |
         additions
+    end
+
+    ##
+    # @example
+    #    user = User.find('user_id')
+    #
+    #    acl.revoke(:read).from(user)
+    def revoke(mode)
+      ModeRevoke.new(self, mode)
     end
 
     ##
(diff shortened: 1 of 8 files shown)


no-reply:
> Is this deliberate to mirror hydra-access-controls?

More or less. We definitely want `Permissions` (or ACLs) to point back to resources, to avoid writing to curation objects when application-specific permissions change. I'm hypothesizing for now that leaving `AccessControlList` as a utility service, rather than a model, will make life easier for Wings. I'm working on that now.

> If it doesn't need to mirror the existing fedora modeling, would collapsing them into the ACL so there is only one ACL object persisted for all Permissions be more performant?

Possibly. If so, this benefit won't be accessible until after Wings is deprecated (Wings must continue to save `Permissions` in the same way `ActiveFedora` does). If performance issues materialize, it should be easy to convert `AccessControlList` to a model and save permissions as nested resources. I think this decision is best delayed until the optimization becomes concrete.

cjcolvar:
The switch to nested resources in the future might require a data migration though hence my interest in pushing for the ideal at the outset.  But it is probably too early to see what is going to be best and my suggestion might just be premature optimization.

no-reply:
> The switch to nested resources in the future might require a data migration though hence my interest in pushing for the ideal at the outset.

i think whether this is the case will depend on the specific adapter. for now, with only Wings supported, the data layout will be the same regardless.

do you feel confident the change will be an optimization for, e.g., the `sequel` adapter, in normal/real-world usage patterns? i'm not sure how to get a feel for that until some of the production code is ported and we can support additional adapters, but if you feel sure it will have an eventual payoff, i'm open to making the change now.

cjcolvar:
Maybe it would be useful to ask some other valkyrie developers for their opinion?

no-reply:
it seems like maybe i should just develop this further before submitting?

it doesn't feel productive to me to explore optimizations at this point, since this approach doesn't yet support round-tripping via Wings, so can't be used in production.

app/services/hyrax/access_control_list.rb
```diff
@@ -0,0 +1,150 @@
...
+    ##
+    # @param permission [Hyrax::Permission]
+    #
+    # @return [Boolean]
+    def <<(permission)
+      permission.access_to = resource.id
+
+      additions << permission
+
+      true
```
hweng: Why return boolean value instead of `additions` array?

app/services/hyrax/access_control_list.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    ##
+    # @param permission [Hyrax::Permission]
+    #
+    # @return [Boolean]
+    def delete(permission)
+      additions.delete(permission)
+      deletions << permission
+
+      true
```
hweng: Same question, why return boolean value instead of `deletions` array?

cjcolvar:
> it seems like maybe i should just develop this further before submitting?
> 
> it doesn't feel productive to me to explore optimizations at this point, since this approach doesn't yet support round-tripping via Wings, so can't be used in production.

:+1: I don't want to block Wings work over this.  Maybe we could review this / think about optimization when it is getting closer to production-ready?

app/services/hyrax/access_control_list.rb
no-reply: `additions` and `deletions` are internal implementation details that i don't expect to survive sustained development. it seems likely this code will use a `ChangeSet` or similar eventually.

For this reason, exposing them via the API seemed fragile.

PR #3902, opened 2019-07-31
fix match_valkyrie_ids_with_af_ids

elrayle:
This fixes several bugs...

* matcher checking valkyrie ids against active fedora ids passes when it shouldn't
* when using members instead of ordered members, model transformer fails to copy member ids to valkyrie resource
* csv constant not found

#### matcher
This matcher was incorrectly returning true in certain circumstances.  It now extends contain_exactly builtin matcher and overrides the match method to map the valkyrie ids there instead.

Tests were added to confirm the matcher is operating as expected.

#### members
Fixing the matcher revealed that the tests for `navigators/child_works_navigator_spec.rb` were not correctly finding work members.  This was happening because the model transformer discards member ids in preference for ordered member ids.  But if members were not added as ordered, then ordered_member_ids returns an empty array.  The second commit adds code to try and get ordered members first, but if none are found, it tries to get the unordered member ids.

Test was added to confirm that members are added regardless of whether they are ordered or unordered.

#### csv
This code hasn't been touched in 2 years, so I'm not sure why this suddenly doesn't work, but it was failing locally and on CCI.  The fix was simply to add `require 'csv'` in the test.

### DIFF
--- a/lib/wings/model_transformer.rb
+++ b/lib/wings/model_transformer.rb
@@ -237,7 +237,14 @@ module Wings
           read_users: pcdm_object.try(:read_users),
           edit_groups: pcdm_object.try(:edit_groups),
           edit_users: pcdm_object.try(:edit_users),
-          member_ids: pcdm_object.try(:ordered_member_ids) } # We want members in order, so extract from ordered_members.
+          member_ids: member_ids }
+      end
+
+      # Prefer ordered members, but if ordered members don't exist, use non-ordered members.
+      def member_ids
+        ordered_member_ids = pcdm_object.try(:ordered_member_ids)
+        return ordered_member_ids if ordered_member_ids.present?
+        pcdm_object.try(:member_ids)
       end
 
       def append_embargo(attrs)
--- a/spec/support/matchers/match_valkyrie_ids_with_af_ids.rb
+++ b/spec/support/matchers/match_valkyrie_ids_with_af_ids.rb
@@ -1,5 +1,19 @@
-RSpec::Matchers.define :match_valkyrie_ids_with_active_fedora_ids do |_active_fedora_ids|
-  match do |valkyrie_ids|
-    contain_exactly(valkyrie_ids.map(&:id))
+module RSpec
+  module Matchers
+    module BuiltIn
+      class MatchValkyrieIdsWithActiveFedoraIds < RSpec::Matchers::BuiltIn::ContainExactly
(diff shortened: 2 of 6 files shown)


spec/support/matchers/match_valkyrie_ids_with_af_ids.rb
```diff
@@ -1,5 +1,19 @@
-RSpec::Matchers.define :match_valkyrie_ids_with_active_fedora_ids do |_active_fedora_ids|
-  match do |valkyrie_ids|
-    contain_exactly(valkyrie_ids.map(&:id))
+module RSpec
+  module Matchers
+    module BuiltIn
```
no-reply: this is a minor nitpick, but shouldn't this module be reserved for the built-in RSpec matchers? any reason not to put `MatchValkyrieIdsWithActiveFedoraIds` in the top-level namespace?

lib/wings/model_transformer.rb
```diff
@@ -237,7 +237,14 @@ def additional_attributes
           read_users: pcdm_object.try(:read_users),
           edit_groups: pcdm_object.try(:edit_groups),
           edit_users: pcdm_object.try(:edit_users),
-          member_ids: pcdm_object.try(:ordered_member_ids) } # We want members in order, so extract from ordered_members.
+          member_ids: member_ids }
+      end
+
+      # Prefer ordered members, but if ordered members don't exist, use non-ordered members.
+      def member_ids
```
no-reply: looks pretty reasonable to me. :+1:

spec/support/matchers/match_valkyrie_ids_with_af_ids.rb
elrayle: I tried putting it in a different namespace and it wasn’t picked up by RSpec.  I’m not sure how it brings in matchers.

no-reply: hmmm... i assumed that https://github.com/samvera/hyrax/pull/3902/files#diff-c7c616878ab53fcfcec45311b43d0da2R13-R17 took care of registering this with RSpec? maybe i'm missing something?

do you mind if i give it a try?

elrayle: Found a fix for the namespace issue.  Thanks for the push back.  This is better.

Issue #3904, opened 2019-08-01
Relationships that have ids but not an id or ids method are not converted to the valkyrie resource


elrayle:
### Descriptive summary

Relationships that have `_ids` appended for the valkyrie resource attribute but do not have an equivalent `_id` or `_ids` property on the active fedora object, fail to copy over the IDs.  The potential relationships that are effected are determined by method `relationship_keys_for`.

### Rationale

All relationships should be converted over to the Valkyrie resource.

### Expected behavior

When an ActiveFedora FileSet with files and original file set is transformed into a Valkyrie resource, the resource should have file_ids and original_file_ids set to the ids from the ActiveFedora FileSet.  For both of these, I believe they should be converted to Valkyrie IDs.

### Actual behavior

When an ActiveFedora FileSet with files and original file set is transformed into a Valkyrie resource, the resource has `file_ids=[]` and `original_file_ids=[]`.

### Steps to reproduce the behavior

Need to write test to check these tranformations.

1. Use test to setup the ids in an ActiveFedora FileSet for files and original_file by adding a file to the file set
2. Use test to check that the ids were transformed into the Valkyrie resource

### Related work

PR #3902 Fixes a problem in the valkyrie id to fedora id matcher and needs to be merged prior to fixing this issue.

elrayle:
For a fileset, this produces...

| resource attribute | af _ids | af _id | af .map(&:id) | af .id | pluralize attrname | convert to Valkyrie ID? |
| ------------------ | ------ | ------ | ------------- | ----- | ---------- | -- |
| "list_source_ids" |  |  | | f23a69e2-7d02-4b55-b4f8-6ed2fb4b0890/list_source  | | ? |
| "ordered_member_proxy_ids" |  |  | ? |  | YES | ? |
| "related_object_ids" | ? |  | ? |  | YES | ? |
| "member_of_collection_ids" | ? |  | ? |  | YES | ? |
| "file_ids" |  |  | "f23a69e2-7d02-4b55-b4f8-6ed2fb4b0890/files/3bcbc7a3-2ee9-4950-b8c7-c5be38d2a043" |  | YES | ? |
| "original_file_ids" |  |  |  | "f23a69e2-7d02-4b55-b4f8-6ed2fb4b0890/files/3bcbc7a3-2ee9-4950-b8c7-c5be38d2a043" | NO | ? |
| "thumbnail_ids" |  |  | ? | ? | NO | ? |
[...]

PR #3905, opened 2019-08-01
Copy ids from attributes ending with _ids where the id is accessed through an object

elrayle:
Ref Issue #3904

This addresses the need to convert ids from an active fedora property where the id is accessed through a method of the property value.

This PR specifically addresses the following for filesets...
* files
* original_file
* thumbnail
* extracted_text

and the following for works...
* member_of_collections

It is likely that it addresses others in the list for #3904, but they do not have tests at this time.  So that issue will be left open to document future changes that may be required.


spec/wings/model_transformer_spec.rb
```diff
@@ -184,6 +184,72 @@
...
+      it 'has the correct reflection ids' do
+        resource = described_class.new(pcdm_object: file_set).build
+        expect(resource.file_ids).to match_valkyrie_ids_with_active_fedora_ids(file_set.files.map(&:id))
+        expect(resource.original_file_ids).to match_valkyrie_ids_with_active_fedora_ids([file_set.original_file.id])
+        expect(resource.thumbnail_ids).to match_valkyrie_ids_with_active_fedora_ids([file_set.thumbnail.id])
+        expect(resource.extracted_text_ids).to match_valkyrie_ids_with_active_fedora_ids([file_set.extracted_text.id])
+      end
+    end
+
+    context 'with members' do
```
elrayle: 'with members' was moved up here because it is actually a contextual test for the build method.  There is a minor change renaming `:parent_work` to `:work` so it can utilize the subject factory the same as other tests for the build method.

spec/wings/model_transformer_spec.rb
```diff
@@ -184,6 +184,72 @@
...
+          work.members << child_work2
+        end
+
+        it 'sets member_ids to the ids of the unordered members' do
+          expect(subject.build.member_ids).to match_valkyrie_ids_with_active_fedora_ids(['cw1', 'cw2'])
+        end
+      end
+    end
+
+    context 'with parent collections' do
```
elrayle: 'with parent collections' is new to test parent collections on a work is correctly translated.

spec/wings/model_transformer_spec.rb
```diff
@@ -184,6 +184,72 @@
         expect(subject.build.lease.id.id).to eq work.lease.id
       end
     end
+
+    context 'with files and derivatives in fileset' do
```
elrayle: 'with files and derivatives in fileset' is new to test that files on a fileset are correctly translated.

spec/wings/model_transformer_spec.rb
```diff
@@ -274,37 +340,4 @@
       end
     end
   end
-
-  context 'with members' do
```
elrayle: moved to be context under the build method

lib/wings/model_transformer.rb
```diff
@@ -196,11 +196,37 @@ def internal_resource
...
+
+      def self.attr_ids_for_name_minus_ids_pluralized(obj, attr_name)
+        real_attr_name = attr_name_minus_ids(attr_name) + 's'
+        attr_value = obj.try(real_attr_name)
+        return [attr_value.id] if attr_value.respond_to? :id
+        return attr_value.map(&:id) if attr_value.respond_to? :each
+      end
+
+      def self.attr_name_minus_ids(attr_name)
+        return unless attr_name.ends_with? '_ids'
```
no-reply: why do we want to return `nil`? if we can tolerate returning the string in the case the string doesn't end with `"_ids"`, this is simply `attr_name.chomp('_ids')`.

lib/wings/model_transformer.rb
```diff
@@ -196,11 +196,37 @@ def internal_resource
...
+            TransformerValueMapper.for(attr_ids_for_name_minus_ids(obj, attr_name)).result ||
+            TransformerValueMapper.for(attr_ids_for_name_minus_ids_pluralized(obj, attr_name)).result || []
+        end
+        keys.each_with_object(attrs) do |attr_name, mem|
+          next unless obj.respond_to?(attr_name) && !mem.key?(attr_name.to_sym)
           mem[attr_name.to_sym] = TransformerValueMapper.for(obj.public_send(attr_name)).result
         end
       end
+
+      def self.attr_ids_for_name_minus_ids(obj, attr_name)
```
no-reply: i'm finding these method names hard to read, and noticing that most of the method code is duplicated.

what do you think about a refactor like:

```ruby
# ...
        attrs = keys.select { |k| k.to_s.end_with? '_ids' }.each_with_object({}) do |attr_name, mem|
          mem[attr_name.to_sym] =
            TransformerValueMapper.for(obj.try(attr_name)).result ||
            TransformerValueMapper.for(attribute_ids_for(name: attr_name.chomp('_ids'), obj: obj)).result ||
            TransformerValueMapper.for(attribute_ids_for(name: attr_name.chomp('_ids').pluralize, obj: obj)).result || []
        end
# ...
def attribute_ids_for(name:, obj:)
  attribute_value = obj.try(name)
  Array(attribute_value).map(&:id) # i think this is equivalent to L216-217 and L222-223
end
```

lib/wings/model_transformer.rb
```diff
@@ -196,11 +196,37 @@ def internal_resource
...
+        real_attr_name = attr_name_minus_ids(attr_name)
+        attr_value = obj.try(real_attr_name)
+        return [attr_value.id] if attr_value.respond_to? :id
+        return attr_value.map(&:id) if attr_value.respond_to? :each
+      end
+
+      def self.attr_ids_for_name_minus_ids_pluralized(obj, attr_name)
+        real_attr_name = attr_name_minus_ids(attr_name) + 's'
```
no-reply: i think this will fail on attribute names with non-standard pluralizations. we should use `attr_name_minus_ids(attr_name).pluralize` instead.

lib/wings/model_transformer.rb
no-reply: keeping `#attr_name_minus_ids` as a separate method also seems like a good option (even if it can be refactored to just `chomp('_ids')`). but if we keep it, renaming to something that scans a little easier would be helpful. something like  `#attribute_name_without_ids` or `#drop_ids_from_name(name)`/`#drop_ids_from(name:)`?

elrayle: I like it!

lib/wings/model_transformer.rb
```diff
@@ -54,7 +54,8 @@ def self.relationship_keys_for(reflections:)
                       .keys
                       .reject { |k| k.to_s.include?('id') }
                       .map { |k| k.to_s.singularize + '_ids' }
-      relationships.delete(:member_ids) # Remove here.  Members will be extracted as ordered_members in attributes method.
+      relationships.delete('member_ids') # Remove here.  Members will be extracted as ordered_members in attributes method.
```
elrayle: This wasn't actually being deleted since the keys are strings, not symbols.

lib/wings/model_transformer.rb
```diff
@@ -54,7 +54,8 @@ def self.relationship_keys_for(reflections:)
...
+      relationships.delete('ordered_member_proxy_ids') # This does not have a Valkyrie equivalent.
```
elrayle: 'ordered_member_proxy_ids' needs to be deleted too.  It does not follow the same patters as the other '_ids' attributes.  And I don't believe there is a Valkyrie equivalent for this as members are ordered in Valkyrie and don't use a proxy to do it.

lib/wings/model_transformer.rb
```diff
@@ -196,11 +197,24 @@ def internal_resource
...
+        end
+        keys.each_with_object(attrs) do |attr_name, mem|
+          next unless obj.respond_to?(attr_name) && !mem.key?(attr_name.to_sym)
           mem[attr_name.to_sym] = TransformerValueMapper.for(obj.public_send(attr_name)).result
         end
       end
+
+      def self.attribute_ids_for(name:, obj:)
+        attribute_value = obj.try(name)
+        return unless attribute_value.present?
```
elrayle: Have to return nil if there isn't a value; otherwise, this returns `[]` TransformerValueMapper.for([]

```
TransformerValueMapper.for([]).result    # []
TransformerValueMapper.for(nil).result   # nil
[] || [3]    # []
nil || [3]   # [3]
```

lib/wings/model_transformer.rb
elrayle: Definitely.  This is fixed in the other refactor you requested.

lib/wings/model_transformer.rb
no-reply: huh. do we know if it needed to be? i suppose your new tests fail without this line?

PR #3906, opened 2019-08-02
Fixing the issue in #show when remove use_valkyrie: false.

### LINKED ISSUE #3865: Set `use_valkyrie` to true in CollectionControllerBehavior#show
### Descriptive summary

In `CollectionControllerBehavior#show`, remove `use_valkyrie: false` from `Hyrax.query_service.find_by_alternate_identifier` and fix related code to advance the use of Valkyrie throughout Hyrax.

hweng:
Fixes #3865 ; refs #3865

- Removed `use_valkyrie: false` from `uery_service.find_by_alternate_identifier` method call. 
- Added customized code to parse out ids for json output for collections page so that fix the failed test that caused by `Valkyrie::ID` object when remove `use_valkyrie: false`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/collections_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/collections_controller_behavior.rb
@@ -25,7 +25,8 @@ module Hyrax
     end
 
     def show
-      @curation_concern ||= Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id], use_valkyrie: false)
+      valkyrie_obj = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id])
+      @curation_concern ||= Wings::ActiveFedoraConverter.new(resource: valkyrie_obj).convert
       presenter
       query_collection_members
     end
--- a/app/controllers/concerns/hyrax/collections_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/collections_controller_behavior.rb
@@ -25,8 +25,7 @@ module Hyrax
     end
 
     def show
-      valkyrie_obj = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id])
-      @curation_concern ||= Wings::ActiveFedoraConverter.new(resource: valkyrie_obj).convert
+      @curation_concern ||= Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id])
       presenter
       query_collection_members
     end
--- a/app/views/hyrax/base/show.json.jbuilder
+++ b/app/views/hyrax/base/show.json.jbuilder
@@ -1,2 +1,17 @@
 json.extract! @curation_concern, *[:id] + @curation_concern.class.fields.reject { |f| [:has_model].include? f }
 json.version @curation_concern.etag
+
+if @curation_concern.is_a? Valkyrie::Resource
+  json.id @curation_concern.id.id
+  json.access_control_id @curation_concern.access_control_id.id
+  json.representative_id @curation_concern.representative_id.id
+  json.thumbnail_id @curation_concern.thumbnail_id.id
+  json.alternate_ids @curation_concern.alternate_ids.map(&:id)
+  json.member_of_collection_ids @curation_concern.member_of_collection_ids.map(&:id)
+  json.member_ids @curation_concern.member_ids.map(&:id)
+  json.access_control_ids @curation_concern.access_control_ids.map(&:id)
+  json.list_source_ids @curation_concern.list_source_ids.map(&:id)
+  json.related_object_ids @curation_concern.related_object_ids.map(&:id)
+  json.representative_ids @curation_concern.representative_ids.map(&:id)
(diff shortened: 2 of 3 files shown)


app/controllers/concerns/hyrax/collections_controller_behavior.rb
```diff
@@ -25,7 +25,8 @@ module CollectionsControllerBehavior
     end
 
     def show
-      @curation_concern ||= Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id], use_valkyrie: false)
+      valkyrie_obj = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id])
+      @curation_concern ||= Wings::ActiveFedoraConverter.new(resource: valkyrie_obj).convert
```
no-reply: This is a good start.

Have you looked at how possible it might be to allow `presenter` and/or `query_collection_members` to work using the Valkyrie::Resource? It would be good to push the `#convert` call as far from the retrieve as possible, or at least to identify the specific blockers to doing so.

hweng: Very good question!
I looked into #present method and #query_collection_members, and could confirm that it is not relevant to this issue.
The only issue exist in @curation_concern
And i also checked the process where the @curation_concern got parsed to response body is out of the scope of hyrax codebase, which means we can not modify it.

Please see the debugging infor and let me know your thoughts: https://github.com/samvera/hyrax/issues/3865#issuecomment-517474571

no-reply: > I looked into #present method and #query_collection_members, and could confirm that it is not relevant to this issue.

i think it's really core to the issue to identify the changes that would be needed to these methods to make them work with valkyrie models. i understand that to be the main point of the work.

> And i also checked the process where the @curation_concern got parsed to response body is out of the scope of hyrax codebase, which means we can not modify it.

we're going to need to modify it sooner or later to complete the Wings work. can we at least document where we reach past the Hyrax codebase and open an issue for resolving it?

hweng: I will double check it.

no-reply: to be clear: the ultimate goal is to avoid needing the `Wings::ActiveFedoraConverter.new(resource: valkyrie_obj).convert` call anywhere, electing to use the `Valkyrie::Resource` result everywhere instead. I expect that there will be places where this won't be practical yet, but the ticket requires pushing as far as is reasonable, and documenting the current boundaries as upcoming work.

hweng: There is no difference between use_valkyrie true or false for `presenter` object.

But you are right, I found this one is relevant: 
https://github.com/samvera/hyrax/blob/master/app/views/hyrax/base/show.json.jbuilder

```
(byebug) json.extract! @curation_concern, *[:id] + @curation_concern.class.fields.reject { |f| [:has_model].include? f }
[:id, :id, :internal_resource, :created_at, :updated_at, :read_groups, :read_users, :edit_users, :edit_groups, :alternate_ids, :embargo, :lease, :member_of_collection_ids, :member_ids, :create_date, :modified_date, :depositor, :title, :date_uploaded, :date_modified, :head, :tail, :collection_type_gid, :alt_title, :label, :relative_path, :import_url, :resource_type, :creator, :contributor, :description, :abstract, :keyword, :license, :rights_notes, :rights_statement, :access_right, :publisher, :date_created, :subject, :language, :identifier, :based_near, :related_url, :bibliographic_citation, :source, :access_control_ids, :list_source_ids, :related_object_ids, :representative_ids, :thumbnail_ids, :access_control_id, :representative_id, :thumbnail_id]
```
But i'm not sure why it returns only key instead of key value. I will dig into it next sprint.

no-reply: > There is no difference between use_valkyrie true or false for presenter object.

this is fantastic news!

hweng:
@samvera/hyrax-code-reviewers I found an approach to define the json attribute names dynamically,  and refactored the hardcoded json path with dynamical one. The PR is updated and all tests are green. It's ready to review. Thanks!

app/views/hyrax/base/show.json.jbuilder
```diff
@@ -1,2 +1,12 @@
-json.extract! @curation_concern, *[:id] + @curation_concern.class.fields.reject { |f| [:has_model].include? f }
+j = json.extract! @curation_concern, *[:id] + @curation_concern.class.fields.reject { |f| [:has_model].include? f }
 json.version @curation_concern.etag
+
+if @curation_concern.is_a? Valkyrie::Resource
```
no-reply: should we push this logic into a service or helper somewhere?

it seems like it might be useful to avoid having this code live in the view, especially since it seems likely to churn a bit as the details get sorted out.

what do you think is best?

no-reply: nevermind, i think this is fine in the jbuilder file

PR #3907, opened 2019-08-06
DO NOT MERGE -- trivial commit to README to cause tests to run on CCI

elrayle:


cjcolvar:
@elrayle Can we close this now that https://github.com/samvera/maintenance/issues/27 was created?

PR #3908, opened 2019-08-07
allow transfer of ownership of works even when proxy is disabled.

### LINKED ISSUE #3836: Transfer ownership has a dependency on proxy deposit
### Descriptive summary

If proxy deposits is disabled, trying to transfer ownership of a work will fail with an authorisation error. Presumably this is happening here https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/transfers_controller.rb#L12

### Rationale

I don't believe there should be a dependency between these two features. Nominating a proxy and transferring ownership are two distinct actions.

### Expected behavior

I should be able to transfer ownership.

### Actual behavior

I am redirected to the homepage with an authorisation error.

1. Disable proxies in the feature flipper
2. Go to 'my works' choose a work and hit 'transfer ownership' in the dropbown.

blancoj:
Fixes #3836 

If a user disables the proxy feature, and then from the dashboard tries to transfer ownership of one of his/her works, prior to this change, he/she would receive an unauthorized message and not be allowed to transfer ownership of the work.  This PR allows transfer of work to be done even if proxy is disabled.

* Disable proxy and enable transfer from the features setting page.
* Go to the works/my page and transfer ownership of one of your works.  This should now work.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -113,7 +113,7 @@ module Hyrax
           end
         end
 
-        can :create, ProxyDepositRequest if Flipflop.proxy_deposit? && registered_user?
+        can :create, ProxyDepositRequest if (Flipflop.proxy_deposit? || Flipflop.transfer_works?) && registered_user?
 
         can :accept, ProxyDepositRequest, receiving_user_id: current_user.id, status: 'pending'
         can :reject, ProxyDepositRequest, receiving_user_id: current_user.id, status: 'pending'
--- a/spec/abilities/proxies_and_transfer_abilities_spec.rb
+++ b/spec/abilities/proxies_and_transfer_abilities_spec.rb
@@ -8,13 +8,15 @@ RSpec.describe 'Proxy and Transfer Abilities' do
 
   describe "Transfers" do
     before do
-      allow(Flipflop).to receive(:transfer_works?).and_return(enabled)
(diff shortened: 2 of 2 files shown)


app/models/concerns/hyrax/ability.rb
```diff
@@ -113,7 +113,7 @@ def proxy_deposit_abilities
           end
         end
 
-        can :create, ProxyDepositRequest if Flipflop.proxy_deposit? && registered_user?
+        can :create, ProxyDepositRequest if (Flipflop.proxy_deposit? || Flipflop.transfer_works?) && registered_user?
```
Dananji: @blancoj Why do we have a dependency on `proxy_deposit` for a user to be able to transfer ownership of a work? As I understand, only enabling `transfer_works` should give the user the ability to transfer ownership of their work. I have so little knowledge on these, so why do we have a check for both these?

blancoj: @Dananji  The `ProxyDepositRequest` model seems to be shared by both proxy and transfer functionality.  My guess is that at one time they were inter dependent.  https://github.com/samvera/hyrax/blob/3af966faf937f10446e4983ca1f384df5ca9c2da/app/models/proxy_deposit_request.rb#L1

no-reply:
@blancoj do you mind rebasing this and restarting the build? I think the configuration is out of date in this branch.

PR #3909, opened 2019-08-07
warn user that he can't make himself a proxy.

### LINKED ISSUE #3710: Users are able to add themselves as a proxy depositor
### Descriptive summary

Observed in Hyrax 2.3.3 and Nurax. Users are able to add themselves as proxy depositors from the manage proxies page.  It looks like this issue was already addressed, there is a before_action method called validate users in the depositors controller.

```
def validate_users
    head :ok if params[:user_id] == params[:grantee_id]
end
```

### Rationale

A user should deposit a work as themselves rather than a proxy of themselves in every circumstance.

### Expected behavior

When a user attempts to add themselves as a proxy they should see a flash message informing them they cannot be a proxy for themselves.

### Actual behavior

When a user adds themselves as a proxy, it is successful.

1. Login as yourself
2. Navigate to manage proxies
3. Add yourself as a proxy.

blancoj:
Fixes #3710

This PR takes the place of PR 3762, which has become stale.

If user logged in tries to make himself a proxy, don't allow this, and put out an alert letting him know that she/he can't do that.

Login and go to the Manage Proxy page
Assign some other user to be a proxy
Verify that user shows up in the list shown at the right of the page
Assign yourself as a proxy
Verify that you get an Alert letting you know that is not permitted.

@samvera/hyrax-code-reviewers


app/controllers/hyrax/depositors_controller.rb
```diff
@@ -15,14 +14,18 @@ def index
     end
 
     def create
-      grantor = authorize_and_return_grantor
-      grantee = ::User.from_url_component(params[:grantee_id])
-      if grantor.can_receive_deposits_from.include?(grantee)
-        head :ok
+      if (params[:user_id].sub '.', '-dot-') == (params[:grantee_id].sub '.', '-dot-')
```
cjcolvar: Can you explain why the string substitution is necessary here?

app/controllers/hyrax/depositors_controller.rb
```diff
@@ -15,14 +14,18 @@ def index
...
+        render json: { name: "FAILURE" }
```
cjcolvar: Maybe it would be good to return a different http status code here since it is a failure?

app/controllers/hyrax/depositors_controller.rb
blancoj: @cjcolvar params[:user_id] comes in with a "-dot-", so just to safe, I do the substitution to both sides.

(byebug) params[:user_id]
"blancoj@umich-dot-edu"
(byebug) params[:grantee_id]
"user1@umich.edu"

app/controllers/hyrax/depositors_controller.rb
blancoj: @cjcolvar I will try status: :bad_request

app/controllers/hyrax/depositors_controller.rb
cjcolvar: Ok.  How about going the opposite direction using `::User.from_url_component(params[:user_id])` which is used on line 41?
```suggestion
      if (params[:grantee_id] == ::User.from_url_component(params[:user_id]))
```

app/controllers/hyrax/depositors_controller.rb
```diff
@@ -15,14 +14,18 @@ def index
     end
 
     def create
-      grantor = authorize_and_return_grantor
-      grantee = ::User.from_url_component(params[:grantee_id])
-      if grantor.can_receive_deposits_from.include?(grantee)
-        head :ok
+      if (params[:user_id].sub '.', '-dot-') == (params[:grantee_id].sub '.', '-dot-')
+        render json: { name: "FAILURE", status: :bad_request }
```
cjcolvar: How about returning the error message in the json and making the status be the http response code instead of inside the json?  It looks like Hyrax already provides a controller method to do this.  Also I think 422 is probably a better fit here than 400.
```suggestion
        return render_json_response(response_type: :unprocessable_entity, message: view_context.t('hyrax.dashboard.proxy_add_deny'))
```
Your JS would have to handle this differently since it wouldn't be a success(200) but I think this is more in line with what Hyrax is already doing and allows your modal to be more generic since it can use whatever the json response message was.

app/controllers/hyrax/depositors_controller.rb
```diff
@@ -15,14 +14,18 @@ def index
     end
 
     def create
-      grantor = authorize_and_return_grantor
-      grantee = ::User.from_url_component(params[:grantee_id])
-      if grantor.can_receive_deposits_from.include?(grantee)
-        head :ok
+      if params[:grantee_id] == ::User.from_url_component(params[:user_id]).email
```
cjcolvar: Looking at this again, I wonder if it makes more sense (and maybe saves a little bit of processing in the normal case) to move this test into the if/else that starts on line 22.  Then it could be something like:
```
if grantor == grantee
  render_json_response(response_type: :unprocessable_entity, message: view_context.t('hyrax.dashboard.proxy_add_deny'))
elsif grantor.can_receive_deposits_from.include?(grantee)
...
```

cjcolvar:
@blancoj I'm guessing that the build is failing due to a cached version fo the test app.  You could try bumping the cache key version to force circleci to rebuild the test app (`v1` -> `v2`).
https://github.com/samvera/hyrax/blob/master/.circleci/config.yml#L72

PR #3926, opened 2019-08-27
increase limit for google analytics stats downloads

### LINKED ISSUE #3925: google analytics stats downloads are limited to 1,000 rows by default
### Descriptive summary

The current setup for downloading stats from google analytics uses the default limit of 1,000 rows of data (~ 2.74 years).  In our local instance of hyrax, we have reached this limit for migrated works and no recent stats are being downloaded or displayed.  We have decided to increase the limit locally to the maximum limit of 10,000 (~ 27.4 years) which should fix the problem for now.  This applies to all versions of hyrax.

Google analytics allows paging results which would allow more than 10,000 rows of data to be downloaded, which is something to consider for the future.

### Rationale

This is a problem for anyone with works more than 1,000 days old or is planning on migrating works and the GA data for those works.  This will start to become a problem for everyone within 3 years of creating works in hyrax.

### Expected behavior

All analytics information for the life of a record should be displayed on the analytics page.

### Actual behavior

Only the first 1,000 days of analytics data are being downloaded and displayed.

1. visit the analytics page for a work created at least 1,000 days ago for which you have analytics data
2. the last data point will be 1,000 days from creation instead of yesterday

screenshot taken 08/27/2019:
image

### Proposed changes

Change the GA query for work record stats here: https://github.com/samvera/hyrax/blob/master/app/models/hyrax/statistic.rb#L36 to `profile.hyrax__pageview(sort: 'date', start_date: start_date, end_date: Date.yesterday, limit: 10_000).for_path(path)`

And change the GA query for file record stats here: https://github.com/samvera/hyrax/blob/master/app/models/file_download_stat.rb#L14 to `profile.hyrax__download(sort: 'date', start_date: start_date, end_date: Date.yesterday, limit: 10_000).for_file(file.id)`

smithjp:
Fixes #3925

Increases GA download limit to the maximum of 10,000 per page.  This does not implement paging of results.

* increase GA download limits for works and files to 10,000
* add end date to work stats downloads

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/file_download_stat.rb
+++ b/app/models/file_download_stat.rb
@@ -13,7 +13,8 @@ class FileDownloadStat < Hyrax::Statistic
       end
       profile.hyrax__download(sort: 'date',
                               start_date: start_date,
-                              end_date: Date.yesterday)
+                              end_date: Date.yesterday,
+                              limit: 10_000)
              .for_file(file.id)
     end
 
--- a/app/models/hyrax/statistic.rb
+++ b/app/models/hyrax/statistic.rb
@@ -33,7 +33,11 @@ module Hyrax
           Rails.logger.error("Google Analytics profile has not been established. Unable to fetch statistics.")
           return []
         end
-        profile.hyrax__pageview(sort: 'date', start_date: start_date).for_path(path)
+        profile.hyrax__pageview(sort: 'date',
+                                start_date: start_date,
+                                end_date: Date.yesterday,
+                                limit: 10_000)
+               .for_path(path)
       end
 
       private


no-reply:
@smithjp thanks for this!

I'm :+1: on this as is, but I'm not sure whether this is a stop-gap or whether it's a good-enough solution for the long run. In your judgement, do we need to implement paging at some point?

Do you have thoughts about how we can test this behavior?

smithjp:
@no-reply I think it would be a good idea to implement paging analytics downloads at some point, at least for correctness, but increasing the limit to 10,000 means that analytics will be downloaded for each repository item for 27 years after deposit, so it is probably not urgent.

I am not sure how to write an automated test for this beyond checking the values of the query parameters since it would probably require using a live connection to GA.  I suppose using a mocked response could work, but I'm not sure how valuable the tests would be.  You might be able to test manually by adjusting the start date for downloads to more than 1,000 days ago.

PR #3935, opened 2019-08-29
Replace calls to ActiveFedora::Base.search_by_id

### LINKED ISSUE #3822: Replace calls to ActiveFedora::Base.search_by_id (removes 2 refs to ActiveFedora)
#### First Pass

#3823 -defines Hyrax::ObjectNotFoundError

#### Second Pass

Depending on the implementation selected, the second pass may have a dependency on #3800.

----
### Descriptive summary

`ActiveFedora::Base.search_by_id` is used to get a single value from a solr doc given the solr document id.

I wonder if it is necessary to use solr to get these values.  

### ActiveFedora code

Active Fedora call being replaced...

Example from app/models/concerns/hyrax/ability.rb
```
          doc = ActiveFedora::Base.search_by_id(t.work_id, fl: 'depositor_ssim')
```

To locate, search code for `ActiveFedora::Base.search_by_id`.

#### First pass

Define Hyrax::Base wrapper method...

```
module Hyrax
  class Base
    def self.search_by_id(id, opts={})
      ActiveFedora::Base.search_by_id(id, opts)
    rescue ActiveFedora::ObjectNotFoundError => e
      raise Hyrax::ObjectNotFoundError.new e.message
    end
end
```

_NOTE: If app/models/hyrax/base.rb does not exist, create it as part of this issue._

Example from app/models/concerns/hyrax/ability.rb
```
          doc = Hyrax::Base.search_by_id(t.work_id, fl: 'depositor_ssim')
```

### Second pass

Does this method need to use Solr for optimization, or can it be replaced with direct calls to wings query_service to find the Valkyrie resource and get the values from that.

There are only two usages of this method and it is not clear that they need the optimization of getting the value from solr.  It should be noted that if this is changed to use query_service, at this time, it will read in the ActiveFedora object to get the value which is costly.

It could be that when #3800 is complete, transitioning this to use the new Valkyrie solr index metadata adapter will be quite simple.

**Exploration required.**

### Considerations

* Do NOT replace this call if found in `lib/wings` directories.

### Impact on Testing

Not expected to require test changes.  All tests should pass with the replacement code.

### Related Work

PR #3804 does a similar first part creating a wrapper for ActiveFedora for ActiveFedora::SolrService.  The second part of replacing the ActiveFedora code has not yet been addressed, at the time of the writing of this issue.

There will be similar issues created that are also replacing ActiveFedora::Base methods.

lsitu:
Fixes #3822

Replace calls to ActiveFedora::Base.search_by_id with Hyrax::Base.search_by_id.

@samvera/hyrax-code-reviewers


app/models/hyrax/base.rb
```diff
@@ -0,0 +1,11 @@
+module Hyrax
+  class Base
+    def self.search_by_id(id, opts = {})
+      opts = opts.merge(rows: 1)
+      result = Hyrax::SolrService.query("id:#{id}", opts)
+
+      raise ActiveFedora::ObjectNotFoundError, "Object '#{id}' not found in solr" if result.empty?
```
no-reply: Hyrax::ObjectNotFoundError?

app/models/hyrax/base.rb
```diff
@@ -0,0 +1,11 @@
+module Hyrax
+  class Base
```
no-reply: does it make sense for this to live in a `Base` class? What about pushing this into a search service?

I think it would be worth looking about in `services/hyrax` to see if there's already a natural home for this functionality.

no-reply: Alternatively, both existing calls seem like they're weird optimizations, designed to avoid loading data from fedora. I wonder if we couldn't refactor them away?

lsitu: @no-reply I've moved it to Hyrax::SolrService class since the behavior of the method is related to SOLR search. And yes, it seems like something that we could refactor at some point later.

PR #3937, opened 2019-08-30
Update wings persister to make a pending test pass

### LINKED ISSUE #3671: wings persister: can find that resource again
make this pending spec pass

VivianChu:
Fixes #3671 

Make the "wings persister: can find that resource again" pending spec pass.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/wings/valkyrie/persister_spec.rb
+++ b/spec/wings/valkyrie/persister_spec.rb
@@ -116,13 +116,13 @@ RSpec.describe Wings::Valkyrie::Persister do
         attribute :ordered_authors, Valkyrie::Types::Array.of(Valkyrie::Types::Anything).meta(ordered: true)
         attribute :ordered_nested, Valkyrie::Types::Array.of(CustomResource).meta(ordered: true)
       end
-      
+
       class Custom < ActiveFedora::Base
         property :title,    predicate: ::RDF::URI('http://example.com/ns/title')
         property :author,   predicate: ::RDF::URI('http://example.com/ns/author')
         property :member_ids, predicate: ::RDF::URI.new('http://www.example.com/member_ids'), multiple: true
         property :nested_resource, predicate: ::RDF::URI("http://example.com/nested_resource"), class_name: CustomResource
-        property :depositor,     predicate: ::RDF::URI("http://example.com/ns/depositor")
+        property :depositor, predicate: ::RDF::URI("http://example.com/ns/depositor")
         property :ordered_authors, predicate: ::RDF::Vocab::DC.creator
         property :ordered_nested, predicate: ::RDF::URI("http://example.com/ordered_nested")
         accepts_nested_attributes_for :nested_resource
--- a/spec/wings/valkyrie/persister_spec.rb
+++ b/spec/wings/valkyrie/persister_spec.rb
@@ -108,7 +108,6 @@ RSpec.describe Wings::Valkyrie::Persister do
     let(:query_service) { adapter.query_service }
     before do
       class CustomResource < Valkyrie::Resource
-        include Valkyrie::Resource::AccessControls
         attribute :title
         attribute :author
         attribute :member_ids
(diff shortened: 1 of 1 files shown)


spec/wings/valkyrie/persister_spec.rb
```diff
@@ -341,10 +341,10 @@ class CustomResource < Valkyrie::Resource
     # not sure how to fix this one. When a resource wasn't ever in AF, it is persisted as
     # internal_resource="Wings::ActiveFedoraConverter::DefaultWork"
     # so the CustomResource defined above will not be persisted as such.
-    xit "can find that resource again" do
+    it "can find that resource again" do
       id = persister.save(resource: resource).id
       item = query_service.find_by(id: id)
-      expect(item).to be_kind_of resource_class
+      expect(item.internal_resource.to_s).to eq(resource_class.to_s)
```
no-reply: does this test change reflect the goal of the test?

cc: @hackmastera

hackmastera: To conform to valkyrie expectations, `query_service.find_by` should return a valkyrie resource. I agree with @no-reply I don't think this change to the test makes sense.

VivianChu: @no-reply @hackmastera - Before the current change, the resource is persisted as internal_resource="Wings::ActiveFedoraConverter::DefaultWork".
After the current change, the resource is persisted as internal_resource="CustomResource"

Here are the screen shot of before and after making current change.
//Before
beforeCustomResourceChange
//After
afterCustomResourceChange

query_service.find always returns unnamed Class object from ModelTransformer L157. That's why I compare the internal_resource instead of be_kind_of. 
https://github.com/samvera/hyrax/blob/master/lib/wings/valkyrie/query_service.rb#L89
https://github.com/samvera/hyrax/blob/master/lib/wings/model_transformer.rb#L157

If I dynamically set the name for the result class of resource_factory.to_resource(object: object) L157 in ModelTransformer class 
- Object.const_set('CustomResource', Class.new(base_for(klass: klass)) do ...
Then the check expect(item).to be_kind_of CustomResource passes.
However the object doesn't look correct to me.  The result of ::ActiveFedora::Base.find is ActiveFedora::Base object instead of CustomResource like other 2 previous screen shots. The data is also shorter.  
Here's the screen shot.
afterHardCodeCustomResource

Do you have any suggestion what I should do in this case?  Thanks

VivianChu: @no-reply - I was able to make the original pending test passed https://github.com/samvera/hyrax/pull/3937/files#diff-73b6a48d0bc3e2171e944e4056f92a6aR347
However, I got 3 failed tests in query_service_spec.  They have the same problem.

For example - https://github.com/samvera/hyrax/blob/wing-issue-3671/spec/wings/valkyrie/query_service_spec.rb#L63
With my change, query_service.find_by returns the correct object the first time, and always failed the 2nd time. 
First time, it returns correctly
`=> #<Book id: "109c3941-f22f-4f27-9c43-31406b12b394", head: [], tail: [], depositor: nil, title: ["Foo"], date_uploaded: nil, date_modified: nil, state: nil, proxy_depositor: nil, on_behalf_of: nil, arkivo_checksum: nil, owner: nil, a_member_of: [], an_ordered_member_of: [], access_control_id: "f1a8fc0b-e931-43ae-a6c4-a73da1865012", representative_id: nil, thumbnail_id: nil, rendering_ids: [], admin_set_id: nil, embargo_id: nil, lease_id: nil>`

and second time, it returns
`=> #<ActiveFedora::Base id: "109c3941-f22f-4f27-9c43-31406b12b394">`

It's related to the best_model method in ModelClassifier class in ActiveFedora
https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/model_classifier.rb#L53
First time, correct 
`[1] pry(#<ActiveFedora::ModelClassifier>)> model_value
=> Book
[2] pry(#<ActiveFedora::ModelClassifier>)> best_model_match
=> ActiveFedora::Base
[3] pry(#<ActiveFedora::ModelClassifier>)> best_model_match > model_value
=> true`

2nd time, incorrect
`[1] pry(#<ActiveFedora::ModelClassifier>)> model_value
=> Book
[2] pry(#<ActiveFedora::ModelClassifier>)> best_model_match
=> ActiveFedora::Base
[3] pry(#<ActiveFedora::ModelClassifier>)> best_model_match > model_value
=> nil`

Do you have any idea why it returns nil for the check best_model_match > model_value ? 

Thanks

VivianChu: @no-reply - It's ready to review. Thanks

PR #3939, opened 2019-09-03
wings wrapper for Hydra:Works:AddFileToFileSet

elrayle:
### Change in this PR

In this PR, AddFileMetadataToFileSet, which performed the file attachment and attempted to set valkyrie specific metadata, was changed to a straight up wrapper of Hydra:Works:AddFileToFileSet.  This included a name change from Wings::Hydra::AddFileMetadataToFileSet to reflect this change in purpose.

### Testing

Tests for Wings::Hydra::AddFileToFileSet were added as part of this PR.

Tests for original_file and thumbnail both pass.

The test for extracted text and for RDF::URI uses fail.  The code related to those tests is already failing.  It just wasn't known until the tests were written.  They are currently marked pending with a note that the failing specs should be fixed.  It seems fine, and good, to merge this PR with the failing pending tests as the tests only reveal a problem that already exists in the code.

Additionally, there are 3 versioning related tests extracted from the Hydra:Works testing of this functionality.  Those are marked pending awaiting the valkyrization of the versioning process.

### Next Steps

The intent of this PR is to isolate any valkyrie specific setting of attributes.  These should be set in the conversion process from af_fileset to valkyrie resource fileset.  When the af_fileset is converted back into a valkyrie resource, the conversion process will setup the applicable FileMetadata resources.


lib/wings/model_transformer.rb
```diff
@@ -212,7 +212,7 @@ def self.run(obj, keys)
 
       def self.attribute_ids_for(name:, obj:)
         attribute_value = obj.try(name)
-        return unless attribute_value.present?
+        return if attribute_value.nil?
```
elrayle: For a File, when attribute_value.content is an empty string, attribute_value.present? returns false even though attribute_value is_a ActiveFedora::File object.  Switching to check for nil catches this and should still appropriately fail the test obj.try returns nil.

lib/wings/hydra/works/services/add_file_to_file_set.rb
```diff
@@ -0,0 +1,54 @@
...
+      #   When set to false, always creates a new file within file_set.files.
+      # @param [Boolean] versioning whether to create new version entries (only applicable if +type+ corresponds to a versionable file)
+      def call(file_set:, file:, type:, update_existing: true, versioning: true)
+        raise ArgumentError, 'supplied object must be an instance of Valkyrie::Resource' unless file_set.is_a?(Valkyrie::Resource) && file_set.file_set?
+        raise ArgumentError, 'supplied file must respond to read' unless file.respond_to? :read
+
+        af_type = normalize_type(type)
+        raise ArgumentError, "supplied type (#{type}) is not supported" unless af_type.present?
+
+        # TODO: required as a workaround for https://github.com/samvera/active_fedora/pull/858
```
no-reply: can you clarify the nature of the work-around?

is this necessary in all cases, or only for certain object states?

elrayle: That got copied over from Hydra::Works and really doesn't apply here with the simple wrapper.  It applies in the Hydra::Works method being wrapped.  I removed it here.

https://github.com/samvera/hydra-works/blob/master/lib/hydra/works/services/add_file_to_file_set.rb#L14

lib/wings/model_transformer.rb
elrayle: Added requested tests.

PR #3941, opened 2019-09-04
Use valkyrie for `BaseActor#create`

no-reply:
Changes to support saving permissions in the valkyrie branch of `BaseActor#save` are included. The new permission models in Valkyrie expect a manual save. Since we don't fully round-trip them, we need to ensure we set them with the proper `#access_to` id and request a save for the `AccessControl` object.

Now unnecessary manual saves of Embargo and Lease are removed, and a refactor to drop valkyrie save behavior into its own method is included.

The use of valkyrie for sensitive operations is moved behind an environment variable `HYRAX_VALKYRIE`. This should help low-risk production environments avoid its use, but make testing easy for users.

@samvera/hyrax-code-reviewers


no-reply:
blocked by #3938

PR #3943, opened 2019-09-04
Move SolrService deprecation target to 4.0

no-reply:
The new interfaces will first be released in 3.0.0, so we can't remove the old
ones until 4.0.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/solr_service.rb
+++ b/app/services/hyrax/solr_service.rb
@@ -110,7 +110,7 @@ module Hyrax
       end
 
       def rsolr_call_warning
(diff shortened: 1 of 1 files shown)


no-reply:
I suspect there's something up with the build cache. Other PRs are failing with similar errors.

I can't look into it until later today, though.

no-reply:
If my assumption is true, a bump to `.regen` should do the trick.

no-reply:
actually, i'm pretty sure that's the deal.

rebase in (and `.regen` bumped as a result). let's see if it builds.

PR #3945, opened 2019-09-05
if admin, show all groups in sharing tab, ALSO removed html tags from page

### LINKED ISSUE #3534: Allow admin users to share works with groups they aren't members of
### Descriptive summary

Hyrax Version: 2.4.1

### Rationale

Admin users (and potentially other users) may need to assign permissions to groups that they are not a member of

### Expected behavior

The drop down box for groups on the 'Sharing' tab of the New Work page includes all groups in the system, at least for admin users

### Actual behavior

The drop down box for groups on the 'share' tab of the New Work page only shows the groups that the user is a member of

1. Log in
2. Navigate to the 'New Work' page
3. Click on the 'Sharing' tab
4. Select the drop down box next to 'Add group'
5. Only the groups you are a member of will be listed

### Related work

blancoj:
Fixes #3534
When creating or editing a work, and user goes to the sharing tab, if the user is an admin, a complete list of all the groups should be displayed in the group selection widget.  If the user is not an admin, then the list will only list all the groups for which the user is a member.

Also, the Sharing page had some html tags being displayed.  That was cleaned up.

* Create a few groups in the role_map.yml file, some groups will have an admin user, and some will have other users.
* Login as an admin, and go edit a work.
* Verify that the groups listed in the group widget of the sharing tab is a complete list of all the groups.
* Login as a non admin, and go edit a work.
* Verify that the groups listed in the group widget of the sharing tab is just the groups for which this user is a member.
* Verify that the Sharing page does not display any html attributes as part of the text on the page.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_form_share.html.erb
+++ b/app/views/hyrax/base/_form_share.html.erb
@@ -3,7 +3,7 @@
 
 <% depositor = f.object.depositor %>
 
-<div class="alert alert-info hidden" id="save_perm_note"><%= t('.permissions_save_note') %></div>
+<div class="alert alert-info hidden" id="save_perm_note"><%= raw t('.permissions_save_note') %></div>
 
 <div class="alert alert-warning hidden" role="alert" id="permissions_error">
   <span id="permissions_error_text"></span>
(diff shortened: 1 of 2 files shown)


no-reply:
this needs a rebase.

no-reply:
@blancoj this looks good to me, but i think we need a more descriptive commit message than this. Could you put some of the info from the PR title and description into the commit message?

Issue #3946, opened 2019-09-06
Issues in Readme – Hyrax Setup Guide


julianklotz:
### Descriptive summary

The Hyrax repository contains a README file that explains how to set up a Hyrax installation for development.
Some of the instructions did not work for me. Relates to using the installation template v3.0.0.pre.beta2 and 2.5.1.

### Rationale

I followed the README setup guide on https://github.com/samvera/hyrax, using a fresh MacOS install running OS 10.14.6.

1. Installing ffmpeg: the options `--with-fdk-aac --with-libvpx --with-libvorbis` are not valid options in current versions of Homebrew (tested on September 2nd, 2019)

2. Following the instructions using the installation template `v3.0.0.pre.beta2/template.rb`, the installation on MacOS fails. There are dependencies that can’t be resolved. Tested with Ruby 2.6.4 and Rails 5.1. I switched to using the 2.5.1 template, which worked like a charme.

3. When running `bin/rails hydra:server`, Solr 8 will be installed automatically. Running `hydra:server` fails, telling you that the fedora-development core couldn’t be created. This is because [Solr 8 isn’t supported at the moment](https://github.com/samvera/hyrax/wiki/Hyrax-Development-Guide):  “[…] Please be aware that the Solr field schema bundled with Hyrax does not yet support Solr 8.0.x releases. […]”. I had to specify the Solr version by adding `version: 7.7.2` to the `.solr_wrapper` file.
    4. In the quickstart section, Hyrax Vagrant is mentioned, but this project is deprecated.

no-reply:
The template issue (2) and Solr 8 support are fixed by using the current `master` instructions, which is to be released as `beta3` in #3949.

no-reply:
Thanks for this!

As far as I know, this is the first report of the `ffmpeg` issue. I can't diagnose it, since I don't have access to an OS X machine. Can someone try to find a working `ffmpeg` install process, or should we just drop the homebrew instructions?

no-reply:
for (4), I'm just dropping the **Quickstart** section. If someone has a proposal for another easy launch (`docker-compose`?), I'd be happy to re-add it.

julianklotz:
Quick check concerning the [ffmpeg Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ffmpeg.rb):

* `--with-fdk-aac` isn’t part of the formular anymore
* `--with-libvpx` is now `--enable-libvpx`
[...]

PR #3951, opened 2019-09-09
Update homebrew ffmpeg install instructions

no-reply:
Connected to #3946 

@samvera/hyrax-code-reviewers


brianmcbride:
Tested the brew command on OS X 10.14.6 and ffmpeg and libvpx were installed successfully.

PR #3952, opened 2019-09-11
[wings] Remove method missing

no-reply:
There's no reason to pass through these methods, we'd prefer they fail. The
`original_file` case is used internally, but should also be removed.

@samvera/hyrax-code-reviewers


VivianChu:
@no-reply - 👍 LGTM .  I rebased it with the master branch, tried to rerun the workflow twice and the tests still failed.  I'll approve it when the tests pass.  Could you take a look at the build? Thanks

no-reply:
@VivianChu sorry, we're expecting this one to fail right now. @lsitu is looking into it

lsitu:
@no-reply @VivianChu All those failed tests that are relying on the removed method `method_missing` are passed now.

PR #3991, opened 2019-09-13
Upgrade Valkyrie to 2.0

no-reply:
The main impact of this is that `Dry::Validations` and `Dry::Struct` are
updated. Neither is backward compatible.

The `WorkflowSchema` is rewritten using the new syntax, and it is now slightly
more strict. It expects `notifications:` to have names that can be constantized
to a `Class`, not just to any registered constant.

`Wings::ConverterValueMapper` has to handle ids and nested structs/`Valkyrie::Resources`
slightly more aggressively, since types are more aggressively coerced, and aren't
squashed into hashes when nested in a parent's `#attributes`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/workflow/workflow_schema.rb
+++ b/app/services/hyrax/workflow/workflow_schema.rb
@@ -14,48 +14,41 @@ module Hyrax
     # @see Sipity::Notification
     # @see Sipity::Method
     # @see ./lib/generators/hyrax/templates/workflow.json.erb
-    WorkflowSchema = Dry::Validation.Schema do
-      configure do
-        def self.messages
-          Dry::Validation::Messages.default.merge(
-            en: { errors: { constant_name?: 'must be an initialized Ruby constant' } }
-          )
-        end
+    class WorkflowSchema < Dry::Validation::Contract
+      class Types
+        include Dry::Types()
 
-        def constant_name?(value)
-          value.constantize
-          true
-        rescue NameError
-          false
+        Constant = Types::Class.constructor do |v|
+          begin
+            v.constantize
+          rescue NameError => _err
+            v
+          end
         end
       end
 
-      required(:workflows).each do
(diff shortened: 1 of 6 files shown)


app/services/hyrax/workflow/workflow_schema.rb
```diff
@@ -14,48 +14,41 @@ module Workflow
...
-              schema do
-                required(:names) { array? { each(:str?) } } # Sipity::WorkflowState#name
-                required(:roles) { array? { each(:str?) } } # Sipity::Role#name
-              end
+      params do
+        required(:workflows).array(:hash) do
+          required(:name).filled(:string) # Sipity::Workflow#name
+          optional(:label).filled(:string) # Sipity::Workflow#label
+          optional(:description).filled(:string) # Sipity::Workflow#description
+          optional(:allows_access_grant).filled(:bool?) # Sipity::Workflow#allows_access_grant?
```
cjcolvar: I don't really know how this works, but looking at the switch from `:str?` to `:string` is there an analogous `:bool?` to `:boolean`?

no-reply: maybe? the shift is from an explicit predicate to a type spec: https://dry-rb.org/gems/dry-schema/basics/type-specs/

i couldn't find any reference to `:boolean` in the `dry-schema` code. but i imagine that it (or maybe `:bool`) would correspond to the built-in `Types::Bool` type, and map correctly to the predicate. without a clear reference point in external code or an explicit schema test for this value, i decided not to mess with it.

i'd be happy to revisit, though.

no-reply: looks like `:bool` works.

hweng:
LGTM 👍

jeremyf:
👍

PR #4012, opened 2019-09-17
[wings] Round trip registered valkyrie native classes

### LINKED ISSUE #3671: wings persister: can find that resource again

no-reply:
Fixes #3671; supersedes #3937 

Users will want to define their models with native Valkyrie. This supports a transition by allowing them to define Valkyrie analogues for their existing ActiveFedora models.

New Valkyrie models need to be registered with `Wings` to cast to the correct ActiveFedora models within Wings. Wings continues to return meta-programmed model objects, but these now subclass the user's native Valkyrie models.

* Add a registry for Wings models
* support and test round-tripping of native Valkyrie models in Wings

* Documentation should be added to https://github.com/samvera/hyrax/wiki/Hyrax-Valkyrie-Usage-Guide
* This has no end-user facing changes.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings.rb
+++ b/lib/wings.rb
@@ -15,11 +15,50 @@
 # However, these dependencies should be considered temprorary: this code will
 # be deprecated for removal in a future release.
 #
+# @example casting an ActiveFedora model to Valkyrie
+#   work     = GenericWork.create(title: ['Comet in Moominland'])
+#   resource = work.valkyrie_resource
+#
+#   resource.title # => ["Comet in Moominland"]
+#   resource.title = ["Mumintrollet på kometjakt"]
+#
+#   Hyrax.persister.save(resource: resource)
+#
+#   work.reload
+#   work.title # => ["Mumintrollet på kometjakt"]
+#
+# @example defining a native Valkyrie model for use with Wings
+#   # given an `ActiveFodora` model like
+#   class Book < ActiveFedora::Base
+#     property :author, predicate: ::RDF::URI('http://example.com/ns/author')
+#     property :title,  predicate: ::RDF::URI('http://example.com/ns/title')
+#   end
+#
+#   # define a `Valkyrie` model with matching attributes,
+#   class BookResource < Hyrax::Resource
+#     attribute :author, Valkyrie::Types::String
+#     attribute :title,  Valkyrie::Types::String
+#   end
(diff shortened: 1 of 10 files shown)


lib/wings.rb
```diff
@@ -15,11 +15,50 @@
...
+#     attribute :author, Valkyrie::Types::String
+#     attribute :title,  Valkyrie::Types::String
+#   end
+#
+#   # and register the relationship with `Wings`
+#   Wings::ModelRegistry.register(BookResource, Book)
+#
+#   # `Wings` will cast the `BookResource` to a `Book` to persist via `ActiveFedora`
+#   resource = BookResource.new(author: 'Tove Jansson', title: 'Comet in Moominland')
+#   adapter  = Wings::Valkyrie::MetadataAdapter.new
```
no-reply: i chose to use `Wings::Valkyrie::MetadataAdapter.new` here to show this working *explicitly* with the wings adapter. i guess we could change either this or the `Hyrax.persister` on L25, but honestly I'm not thinking it matters. It might be good to show the ways of selecting persisters, and just get better inline documentation for `Hyrax.persister` in at the Hyrax level.

no-reply:
@elrayle @cjcolvar i've made the suggested changes, and rebased on @lsitu's `ModelTransformer` refactor

spec/wings/valkyrie/query_service_spec.rb
```diff
@@ -5,31 +5,35 @@
...
-    class Image < ActiveFedora::Base
-      include Hydra::AccessControls::Permissions
-      property :title, predicate: ::RDF::Vocab::DC.title, multiple: true
     end
   end
 
   after do
-    Object.send(:remove_const, :Book)
-    Object.send(:remove_const, :Image)
+    Hyrax::Test.send(:remove_const, :QueryService)
```
elrayle: I like this!  Makes for easier removal of constants defined in tests.

VivianChu:
👍

spec/wings/valkyrie/query_service_spec.rb
no-reply: yeah, we should probably start using Hyrax::Test everywhere test constants are needed

PR #4014, opened 2019-09-18
add file_converter_service which converts the metadata between AF:File and FileMetadata resource

elrayle:
Provides conversions between AF:Files and FileMetadata resources.

* af_file_to_resource(af_file:)
* resource_to_af_file(metadata_resource:)

### DIFF
--- a/lib/wings/services/file_converter_service.rb
+++ b/lib/wings/services/file_converter_service.rb
@@ -1,0 +1,66 @@
+# Conversion service for going between files on a valkyrie resource and files on an active fedora object
+module Wings
+  class FileConverterService
+    class << self
+      def af_file_to_resource(af_file:)
+        return unless af_file&.id.present?
+        attrs = base_af_file_attributes(af_file: af_file)
+        attrs = metadata_node_to_attributes(metadata_node: af_file.metadata_node,
+                                            attributes: attrs)
+        Hyrax::FileMetadata.new(attrs)
+      end
+
+      def resource_to_af_file(metadata_resource:)
+        return unless metadata_resource&.alternate_id.present?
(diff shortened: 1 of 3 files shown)


lib/wings/services/file_converter_service.rb
```diff
@@ -0,0 +1,66 @@
+# Conversion service for going between files on a valkyrie resource and files on an active fedora object
+module Wings
+  class FileConverterService
```
no-reply: where do we plan to call this from?

elrayle: These methods will be called from ModelTransformer and ActiveFedoraConverter.  That is what I'm working on next.  In master now, this conversion of the file metadata is not happening.

lib/wings/services/file_converter_service.rb
```diff
@@ -0,0 +1,66 @@
...
+            if af_file.respond_to? mname
+              af_file.send(mname, v)
+            elsif af_file.metadata_node.respond_to? mname
+              af_file.metadata_node.send(mname, v)
+            end
+          end
+        end
+
+        def content(metadata_resource)
+          return nil if metadata_resource.content.nil? || metadata_resource.content.size.zero?
```
no-reply: are we expecting `metadata_resource.content` to be `nil` or an `Array`? i'm wondering about the impact of `size.zero?` vs `empty?`. `size` is normally going to `0` or `1` i'd guess?

will `return nil if metadata_resource.content.blank?` work here?

elrayle: This is used to set the content in the AF:File.  When there is no content (i.e. nil or []), then the content in the AF:File will be nil.  If there is an empty string (i.e. ['']), then the content in the AF:File should be an empty string.

I confirmed that blank? works for all 3 of these conditions.

PR #4017, opened 2019-09-18
Define a schema sharing pattern for Hyrax/Valkyrie

no-reply:
Allow schema reuse for Valkyrie resources with `include
Hyrax::Schema(:schema_name)`.

Resolution of schema names is hidden behind an abstraction, with the idea that
we can support schema loaders for
[Houndstooth](https://github.com/samvera-labs/houndstooth), but also support
simple schema definitions. For now, the schemas are hard-coded and only specify
attribute-type pairs.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/metadata/basic.rb
+++ b/app/models/hyrax/metadata/basic.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Metadata
+    ##
+    # @api private
+    class Basic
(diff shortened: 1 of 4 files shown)


no-reply:
cc: @geekscruff @tpendragon

no-reply:
Some further ideas about where this might be going are explored in https://github.com/samvera/hyrax/compare/data-driven-metadata

VivianChu:
👍

PR #4020, opened 2019-09-19
refactor wings overrides of custom queries

elrayle:
This refactor places wings overrides of custom queries in the wings directory structure.

#### Hyrax version of custom query

The default hyrax implementation of the custom query should not have any adapter specific calls.  It should only use standard valkyrie query service calls.

* Write standard valkyrie version of custom queries at `app/services/hyrax/custom_queries/your_custom_query_name.rb`. 

NOTE: There are several find_* custom queries that provide a pattern for how to write a custom query.  Reference:  [Valkyrie documentation for custom queries](https://github.com/samvera/valkyrie/wiki/Queries) 

* Register the custom query in `lib/wings.rb` by updating the `custom_queries` array
* Write a test for the standard valkyrie custom query at `spec/services/hyrax/custom_queries/your_custom_query_name_spec.rb`

#### Wings override of custom query

You only need to write this version of the custom query if the custom query requires wings adapter specific calls.  If wings needs to override the custom query...

* Write the override of the custom query in `lib/wings/services/custom_queries/your_custom_query.rb`.
* Register the override custom query in `lib/wings.rb` by replacing the Hyrax version with the Wings version.  For example...
```
custom_queries = 
  [...
    Wings::CustomQueries::FindAccessControl, # override Hyrax::CustomQueries::FindAccessControl
   ...]
```
* Write a test for your custom query override at `spec/wings/services/custom_queries/your_custom_query_name_spec.rb`


no-reply:
I wonder if it wouldn't be a good start to just lift and shift `custom_queries/wings.rb` into `lib/wings`, without the remainder of the changes?

PR #4023, opened 2019-09-19
Add find_file_metadata custom queries

elrayle:
Adds custom queries...
* find_file_metadata_by
* find_file_metadata_by_alternate_id

Update add_file_to_file_set tests to use find_file_metadata_by custom query.

This includes a change to FileMetadata attribute alternate_id changed to alternate_ids.  This is a standard Valkyrie field and needs to use the Valkyrie attribute name to work with Valkyrie query services for all adapters.

### DIFF
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -3,7 +3,7 @@
 module Hyrax
   class FileMetadata < Valkyrie::Resource
     attribute :file_identifiers, ::Valkyrie::Types::Set # id of the file stored by the storage adapter
-    attribute :alternate_id, ::Valkyrie::Types::Set # id of the Hydra::PCDM::File which holds metadata and the file in ActiveFedora
+    attribute :alternate_ids, Valkyrie::Types::Set.of(Valkyrie::Types::ID) # id of the Hydra::PCDM::File which holds metadata and the file in ActiveFedora
     attribute :file_set_id, ::Valkyrie::Types::ID # id of parent file set resource
 
     # all remaining attributes are on AF::File metadata_node unless otherwise noted
--- a/spec/services/hyrax/custom_queries/find_file_metadata_spec.rb
+++ b/spec/services/hyrax/custom_queries/find_file_metadata_spec.rb
@@ -1,11 +1,104 @@
-# rubocop:disable RSpec/EmptyExampleGroup
 RSpec.describe Hyrax::CustomQueries::FindFileMetadata do
-  let(:query_service) { Hyrax.query_service }
-  let(:valk_id) { double } # TODO: Stubbed waiting for tests
-  subject { query_service.custom_queries.find_file_metadata_by(id: valk_id) }
+  before do
+    query_service.custom_queries.register_query_handler(described_class)
+  end
+  let(:metadata_adapter) { Valkyrie::Persistence::Memory::MetadataAdapter.new }
+  let(:query_service) { metadata_adapter.query_service }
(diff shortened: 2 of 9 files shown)


spec/services/hyrax/custom_queries/find_file_metadata_spec.rb
```diff
@@ -0,0 +1,11 @@
+# rubocop:disable RSpec/EmptyExampleGroup
+RSpec.describe Hyrax::CustomQueries::FindFileMetadata do
+  let(:query_service) { Hyrax.query_service }
+  let(:valk_id) { double } # TODO: Stubbed waiting for tests
+  subject { query_service.custom_queries.find_file_metadata_by(id: valk_id) }
+
+  describe '.find_file_metadata_by' do
+    # TODO: Need tests for this custom query once inMemory adapter is supported
```
no-reply: if we're not going to write the tests, i think i'd prefer to either drop the file altogether, or write a few pending tests we intend to fill in later:

```ruby
describe '.find_file_metadata_by' do
  it 'does thing we know we want to test'
  it 'does other thing'
end
```

this way it's clear what we *should* be testing here, and what tech debt we have taken on.

no-reply: that said, i think it would be really nice to just slot in tests using the memory adapter and its query service.

elrayle: @no-reply I went ahead and wrote the in-memory tests for this custom query

app/models/hyrax/file_metadata.rb
```diff
@@ -3,7 +3,7 @@
 module Hyrax
   class FileMetadata < Valkyrie::Resource
     attribute :file_identifiers, ::Valkyrie::Types::Set # id of the file stored by the storage adapter
-    attribute :alternate_id, ::Valkyrie::Types::Set # id of the Hydra::PCDM::File which holds metadata and the file in ActiveFedora
+    attribute :alternate_ids, Valkyrie::Types::Set.of(Valkyrie::Types::ID) # id of the Hydra::PCDM::File which holds metadata and the file in ActiveFedora
```
elrayle: Required to be alternate_ids which is a standard defined Valkyrie attribute used by the query service.

spec/wings/hydra/works/services/add_file_to_file_set_spec.rb
```diff
@@ -33,11 +33,9 @@
         expect(ids.first).to be_a Valkyrie::ID
         expect(ids.first.to_s).to start_with "#{file_set.id}/files/"
 
-        # file_metadata = Hyrax::query_service.find_by(id: ids.first) # TODO: Can't do this b/c Wings tries to get an ActiveFedora::Base object instead of FileMetadata Resource
-        # expect(file_metadata.content).to start_with('%PDF-1.3')
-        af_file_set = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: file_set.id, use_valkyrie: false)
-        expect(af_file_set.original_file.content).to start_with('%PDF-1.3')
-        expect(af_file_set.original_file.mime_type).to eq(pdf_mimetype)
+        file_metadata = Hyrax.query_service.custom_queries.find_file_metadata_by(id: ids.first)
```
elrayle: These tests were updated to use the custom query so they are fully valkyrized.

app/models/hyrax/file_metadata.rb
no-reply: oh! glad we caught this

PR #4025, opened 2019-09-20
Use valkyrie for BaseActor#create when ENV['HYRAX_VALKYRIE'] is true

### LINKED ISSUE #3572: BaseActor replace #save with Valkyrie

no-reply:
Fixes #3572

Changes to support saving permissions in the valkyrie branch of BaseActor#save are included. The new permission models in Valkyrie expect a manual save. Since we don't fully round-trip them, we need to ensure we set them with the proper #access_to id and request a save for the AccessControl object.

Now unnecessary manual saves of Embargo and Lease are removed, and a refactor to drop valkyrie save behavior into its own method is included.

The use of valkyrie for sensitive operations is moved behind an environment variable HYRAX_VALKYRIE. This should help low-risk production environments avoid its use, but make testing easy for users.

@samvera/hyrax-code-reviewers

* This is a non-functional change.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,7 @@ jobs:
     resource_class: medium+
     environment:
       RAILS_VERSION: << parameters.rails_version >>
+      HYRAX_VALKYRIE: true
       NOKOGIRI_USE_SYSTEM_LIBRARIES: true
     steps:
       - samvera/cached_checkout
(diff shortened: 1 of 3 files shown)


no-reply:
closing this briefly due to a known issue

PR #4026, opened 2019-09-20
Data-driven Metadata

no-reply:
A whole bunch of configurable/data-driven/houndstooth (but not exactly houndstooth) metadata work here. This allows basic attribute -> type maps, as well as application specific indexing behavior, to be defined in application configuration.

Special attention is paid to defining an extensible architecture, allowing adopters to build clearly defined and modular-as-needed systems for defining metadata-dependent behavior. We'll need to continue to pay attention to this as form work progresses.

See individual commit messages for more extensive expressions of the direction of this work.


config/metadata/basic_metadata.yaml
```diff
@@ -0,0 +1,67 @@
+attributes:
+  abstract:
+    type: string
+    multiple: true
```
elrayle: I am really liking defining metadata in a config file!!!

Have you had any thoughts on other parts of adding metadata beyond the attribute definitions...
* in WorkForm subclass... (for new/edit form)
  * terms determines which properties are shown on the new/edit form (and order I think)
  * required_fields determines which properties are required on the new/edit form
  * primary_terms determines which properties are above the fold
  * secondary_terms determines which properties are below the fold
* in WorkShowPresenter subclass... (for show page)
  * delegate solr retrieval for each term to solr_document (methods defined in solr_document.rb)
* in app/models/solr_document.rb... (for show page)
  * add method to extract value from solr_document for each property to be retrieved by the presenter
* copy in and update `app/views/curation_concerns/base/_attribute_rows.html.erb`... (for show page)
  * add `<%= presenter.attribute_to_html(:_property_) %>`  for each property
  * optionally use a `render_as` parameter to `attribute_to_html` to control how the html is written out for the property
* configure blacklight... (for discovery)

elrayle: Perhaps expand the yml to include some of this information...
```
  description:
    type: string
    multiple: true
    editable:                    # if editable not included, not editable
      fold: :primary | :secondary
      required: true | false
      index: _integer_       # sort order of fields within the primary/secondary fold)
   showable:
      render_as: :attribute | :date_attribute | :external_link_attribute | :faceted | etc. (and custom renderers)
      index: _integer_       # sort order of fields on the show page
```

The pieces this doesn't address are the solr_document extraction methods and delegators, and the blacklight config.  There may be other stuff I'm not remembering.  But this would go a long way toward making a simpler process.

elrayle: And questions in general for how these aspects are being handles in the Valkyrized Hyrax...
* controlling how a property is indexed - previously done on the property statement with `index.as`
* predicate associated with the property

elrayle:
This is a great start in the direction of simplifying custom metadata.  I made some suggestions for where we might go next with this.

config/metadata/basic_metadata.yaml
no-reply: See also: https://github.com/samvera-labs/houndstooth/blob/master/examples/hyrax.yaml

My notion is that something like the `SimpleSchemaLoader` used here could be defined for other input formats: `HoundstoothSchemaLoader`, `MyCustomDatabaseSchemaLoader`, etc...

In the meanwhile, `SimpleSchemaLoader` could be extended to add support for validation specs, indexing rules, etc..., along the lines of what you're suggesting above.

no-reply: > controlling how a property is indexed - previously done on the property statement with index.as

notably, we've officially deprecated this as of today. The present way of handling this is to define a custom indexer and set it for the model. This pattern has been available since Hyrax 1.0 (i believe).

I expect to keep it, with a new implementation, for Valkyrie. If that's the case, it should be fairly easy to adopt a module builder pattern like this one for those indexers, and read the schema config (or perhaps a separate config document?) to define indexing behavior.

geekscruff:
I'm working on implementing something for Indiana that's based on the m3/houndstooth specification. I'm hoping that it will be either contributed to Hyrax or become a standalone gem. I'll keep an eye on this work too - would be good to align if possible @no-reply

no-reply:
@geekscruff, @elrayle  i'm thinking to revisit this again, following #4141, specifically to extend it to add indexing behavior.

if you have feedback about the patterns used so far, or the direction this is heading, it would be good to get that out there soon.

no-reply:
@geekscruff @orangewolf: 

> Indiana came up with an additonal requirement around metadata in a particualr context, eg. two departments wanting a different schema for a particular work type

this is an interesting requirement. i think in Fedora terms, the model type (`info:fedora/fedora-system:def/model#hasModel`) is meant to be tightly bound to the schema. i think the concept of "Work Type" in Hyrax/Hydra Works might blend this with some other issues: the work type modal, form selection, indexing behavior, etc... (for that matter, Houndstooth also does this)

do you have a good sense for which things Indiana identifies "Work Type" with other than the schema?

no-reply:
Okay, this now depends on  #4141, but also now includes a basic indexing implementation.

I'm hoping that I can get a wide range of reviews on this, with a focus on the inline documentation, the correctness and completeness of the `@api` declarations, and the usability of the resulting interfaces.

no-reply:
Actually, i'd like for this to just supersede  #4141. I'm closing that accordingly.

config/metadata/basic_metadata.yaml
no-reply: Prospective implementations for indexing are now included in this PR.

no-reply:
> Have you considered using a standard such as JSON schema to declare the fields?

This was considered at some length (and came down against my personal opinion) as part of the Houndstooth/M3 work. That's not to say we couldn't do something JSON schema based as the core Hyrax implementation, but my feeling is that the community is going in a different direction. This is leading me to want to do the lightest-weight thing on hand for "replace ruby declarations", and my sense is that that's yaml configs to line up with other rails/application configuration.

Other ideas are exceedingly welcome. I'm not more than superficially happy with the status quo.

no-reply:
> I wonder if this is a good time to indicate that “multiple” is ordered or not.

Good question. I think the status quo is "it's not" but implementation decisions on the forward looking implementation are basically "maybe". Is this adequate?

My preference would be to avoid serious reconsideration of AF models, and Wings has that limitation in-hand.

laritakr:
I don't have a lot of time to look at this in detail, but I really like the direction in general. It looks like a solid beginning to expand on.

Regarding @geekscruff comment regarding multiple schema for a work type, we have a similar idea in our current Curate system, where we have model Document with multiple types, and various types of documents can have different metadata. It's a super messy and problematic implementation, but one benefit is that someone can submit a work and then subsequently change the type, something that can't happen at the model level. We have discussed the idea of creating one model containing all possible metadata, and then using type to distinguish which metadata items apply to a specific work type.

app/services/hyrax/simple_schema_loader.rb
```diff
@@ -0,0 +1,123 @@
...
+
+    class UndefinedSchemaError < ArgumentError; end
+
+    private
+
+      ##
+      # @param [#to_s] schema_name
+      # @return [Enumerable<AttributeDefinition]
+      def definitions(schema_name)
+        schema_config(schema_name)['attributes'].map do |name, config|
```
jeremyf: It would be useful to include schema validation for the schema_config. I believe that may make sense as a new issue in Hyrax (and not something to block this PR).

no-reply: yes i think so. i haven't pursued it yet, thinking it's premature.

i'm not sure the end of this exploration won't just be a Houndstooth -> Hyrax engine; and i think Houndstooth has/should have its own validation tooling?

the value of schema validation for workflows has been somewhat questionable, in my experience: it certainly prevents some classes of bad workflow configurations from being loaded into production, but it also has arguably resulted in less interpretable error messages for folks trying to write workflows than just allowing bad schemas to run during development. i'd want to take care to avoid that outcome.

geekscruff:
@no-reply @laritakr the work we've done for Indiana adds a concept called 'context' which is tied to the admin set. So there's a default 'context' and 'schema' for a work type, and then there can be additional 'contextual' ones dependent on the Admin Set a work is assigned to.

At the model class level, the default schema is applied, meaning all properties defined for that work type are available. But at the Form / View level, this can be contextualised to remove properties and change labels. 

The work I've been doing is here https://github.com/notch8/essi

PR #4033, opened 2019-09-23
Provide a default unauthorized.png

no-reply:
Provide `unauthorized.png` from the engine. This requires a backport to 2.x-stable

Fixes #3887 

Does this make the image copy in the generator unnecessary?

* Provide `unauthorized.png` from the engine.

@samvera/hyrax-code-reviewers


no-reply:
cc: @mark-dce

mark-dce:
👍 & agreed that with this in place, I don't think the image copy in the generator is necessary.  If someone wants a specific image for unauthorized, they'd put one in their local repo like this https://github.com/curationexperts/laevigata/blob/master/app/assets/images/unauthorized.png; otherwise, you just get whatever the gem provides.

PR #4039, opened 2019-09-23
Save file_set using valkyrie and pass use_valkyrie to file_actor

elrayle:
This is the first of 3 PRs to update the file_set_actor to be valkyrized.  The expected PRs are...
* save file_set using valkyrie and pass use_valkyrie to file_actor
* save work using valkyrie when attaching file set to the work
* destroy file_set using valkyrie

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -117,7 +117,7 @@ module Hyrax
         end
 
         def build_file_actor(relation)
-          file_actor_class.new(file_set, relation, user)
+          file_actor_class.new(file_set, relation, user, use_valkyrie: use_valkyrie)
         end
 
         # uses create! because object must be persisted to serialize for jobs
@@ -164,15 +164,16 @@ module Hyrax
           work.save!
         end
 
-        # method to use when fully valkyrized
+        # save a valkyrie resource returning false if FailedSaveError is raised
+        # TODO: Change calls to `#perform_save` to call `#save` instead when env passes resources instead of active fedora objects
         def save(resource)
           Hyrax.persister.save(resource: resource)
-        rescue StandardError
+        rescue FailedSaveError
           false
         end
 
         # switches between using valkyrie to save or active fedora to save
-        # TODO: Remove this method when fully valkyrized
+        # TODO: Remove this method when env passes resources instead of active fedora objects
         def perform_save(object)
           obj_to_save = object_to_act_on(object)
           if valkyrie_object?(obj_to_save)
@@ -187,14 +188,14 @@ module Hyrax
         end
 
         # if passed a resource or if use_valkyrie==true, object to act on is the valkyrie resource
-        # TODO: Remove this method when fully valkyrized
+        # TODO: Remove this method when env passes resources instead of active fedora objects
         def object_to_act_on(object)
           return object if valkyrie_object?(object)
           use_valkyrie ? object.valkyrie_resource : object
         end
 
         # determine if the object is a valkyrie resource
-        # TODO: Remove this method when fully valkyrized
+        # TODO: Remove this method when env passes resources instead of active fedora objects
         def valkyrie_object?(object)
           object.is_a? Valkyrie::Resource
         end
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -1,11 +1,12 @@
 module Hyrax
   module Actors
     # Actions are decoupled from controller logic so that they may be called from a controller or a background job.
-    class FileSetActor
+    class FileSetActor # rubocop:disable Metrics/ClassLength
       include Lockable
-      attr_reader :file_set, :user, :attributes
+      attr_reader :file_set, :user, :attributes, :use_valkyrie
 
-      def initialize(file_set, user)
+      def initialize(file_set, user, use_valkyrie: false)
+        @use_valkyrie = use_valkyrie
         @file_set = file_set
         @user = user
       end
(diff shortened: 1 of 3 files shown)


spec/actors/hyrax/actors/file_set_actor_spec.rb
```diff
@@ -3,345 +3,350 @@
...
-    describe '#assign_visibility?' do
-      let(:viz) { Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC }
-
-      it 'without params, returns false' do
-        expect(actor.send(:assign_visibility?)).to eq false
-      end
-      it 'with string-keyed or symbol-keyed visibility, returns true' do
-        expect(actor.send(:assign_visibility?, visibility: viz)).to eq true
-        expect(actor.send(:assign_visibility?, 'visibility' => viz)).to eq true
+  [false, true].each do |use_valkyrie|
```
elrayle: This spec hasn't really changed.  Adding the line to run all tests with use_valkyrie=true and again with false, changed the indentation, which makes it look like the file had substantial changes when it didn't.

spec/actors/hyrax/actors/file_set_actor_spec.rb
```diff
@@ -3,345 +3,350 @@
...
+          actor.create_content(fixture_file_upload(file1))
+          actor.create_content(fixture_file_upload('hyrax_generic_stub.txt'))
+          actor.file_set.reload
+        end
 
-    it "restores the first versions's content and metadata" do
-      actor.revert_content(version1)
-      expect(restored_content.original_name).to eq file1
+        it "restores the first versions's content and metadata" do
+          pending 'implementation of valkyrized versioning of files' if use_valkyrie
```
elrayle: This is the only change to this spec.  Versioning is not yet implemented using valkyrie, so this test fails when use_valkyrie=true.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -115,7 +117,7 @@ def ability
         end
 
         def build_file_actor(relation)
-          file_actor_class.new(file_set, relation, user)
+          file_actor_class.new(file_set, relation, user, use_valkyrie: use_valkyrie)
```
elrayle: pass use_valkyrie to file_actor

no-reply:
@elrayle do you mean for this to be assigned to you, or is it ready for review?

no-reply:
I'll go ahead and review.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -21,7 +22,8 @@ def create_content(file, relation = :original_file, from_url: false)
         # If the file set doesn't have a title or label assigned, set a default.
         file_set.label ||= label_for(file)
         file_set.title = [file_set.label] if file_set.title.blank?
-        return false unless file_set.save # Need to save to get an id
+        @file_set = perform_save(file_set) # Need to save to get an id # TODO: change this to `save(file_set)` when fully valkyrized
```
no-reply: what's holding this back at this point? it might be helpful to replace "fully valkyrized" with a description of the gap.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -161,6 +163,41 @@ def unlink_from_work
           work.rendering_ids -= [file_set.id]
           work.save!
         end
+
+        # method to use when fully valkyrized
```
no-reply: if this is a public method, i think it needs more descriptive documentation.

app/actors/hyrax/actors/file_set_actor.rb
elrayle: #perform_save does the switching between AF and Valkyrie approaches.  #save is pure valkyrie.  So when we are at a point where we are passing valkyrie resources and no longer need to switch between the two methods, the changes would be...

* change the method call here from `perform_save(file_set)` to `save(file_set)`.  For `perform_save`, file_set can be either an AF object or a resource.  For `save`, file_set must be a resource.
* remove any private methods marked `TODO: Remove this method when fully valkyrized`

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -161,6 +163,41 @@ def unlink_from_work
...
+        def save(resource)
+          Hyrax.persister.save(resource: resource)
+        rescue StandardError
```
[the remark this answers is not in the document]
elrayle: This was leftover from the previous work on file_set_actor.  Two options...

OPTION 1: remove the rescue and let exceptions be raised
OPTION 2: catch call exceptions and return false

The reason for returning false is that af_object.save returns false if the save fails.  As long as we don't quietly fail, then I wouldn't anticipate unexpected consequences from changing the behavior from return false to raise an exception.

How would you like to handle this?  Other parts of the code will follow this same pattern when saving.

app/actors/hyrax/actors/file_set_actor.rb
elrayle: It is a private method.  That description probably should be a TODO.

no-reply: i think my concern is about the phrase "fully valkyrized". this doesn't help me understand when i should use this method, or when i should look to remove the others.

app/actors/hyrax/actors/file_set_actor.rb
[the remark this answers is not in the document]
no-reply: why not catch errors specific to persistence failures?

app/actors/hyrax/actors/file_set_actor.rb
elrayle: I had switched the comment to...
```
# TODO: Use this method where ever perform_save is called when fully valkyrized
```

but if your hesitation is 'fully valkyrized' (which is used in many comments), then perhaps a more accurate description would be...

```
# TODO: Use this method where ever perform_save is called when env passes resources instead of active fedora objects.
```

app/actors/hyrax/actors/file_set_actor.rb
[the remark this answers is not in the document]
elrayle: The only one I see in persister is FailedSaveError.  Are there others?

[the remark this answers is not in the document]
elrayle: PR updated to catch FailedSaveError and return false only for that exception.

app/actors/hyrax/actors/file_set_actor.rb
elrayle: Comment updated to be more precise...

`# TODO: Change calls to `#perform_save` to call `#save` instead when env passes resources instead of active fedora objects`

PR #4040, opened 2019-09-23
Use valkyrie for `BaseActor#create`

no-reply:
At the base level of the actor stack, use valkyrie for `#create`. This is a
simple flag swich; all the blockers to this swap have been cleared in prior
work.

The use of valkyrie for sensitive operations is moved behind an environment
variable HYRAX_VALKYRIE. This should help low-risk production environments avoid
its use, but make testing easy for users.

The new permission models in Valkyrie expect a manual save. Since we don't fully
round-trip them, we need to ensure we set them with the proper #access_to id and
request a save for the AccessControl object. Unnecessary manual saves of Embargo
and Lease are removed, and a refactor to drop valkyrie save behavior into its
own method is included.

Fixes #3941 

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,6 +88,9 @@ jobs:
       bundler_version:
         type: string
         default: 1.17.3
+      hyrax_valkyrie:
+        type: string
+        default: "false"
     executor:
       name: 'samvera/ruby_fcrepo_solr_redis'
       ruby_version: << parameters.ruby_version >>
@@ -95,6 +98,7 @@ jobs:
     parallelism: 10
     environment:
       COVERALLS_PARALLEL: true
+      HYRAX_VALKYRIE: << parameters.hyrax_valkyrie >>
       VALKYRIE_SOLR_PORT: 8985
     steps:
       - attach_workspace:
@@ -161,5 +165,11 @@ workflows:
       - test:
           name: "ruby2-6-2"
           ruby_version: "2.6.2"
+          requires:
+            - build
+      - test:
+          name: "ruby2-6-2-valkyrie"
+          ruby_version: "2.6.2"
+          hyrax_valkyrie: "true"
           requires:
             - build
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -19,7 +19,7 @@ module Hyrax
         apply_creation_data_to_curation_concern(env)
         apply_save_data_to_curation_concern(env)
 
-        save(env, use_valkyrie: false) &&
+        save(env, use_valkyrie: Hyrax.config.use_valkyrie?) &&
           next_actor.create(env) &&
           run_callbacks(:after_create_concern, env)
       end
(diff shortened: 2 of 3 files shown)


app/actors/hyrax/actors/base_actor.rb
```diff
@@ -113,6 +114,15 @@ def remove_blank_attributes!(attributes)
         def multivalued_form_attributes(attributes)
           attributes.select { |_, v| v.respond_to?(:select) && !v.respond_to?(:read) }
         end
+
+        def valkyrie_save(resource:)
+          permissions = resource.permission_manager.acl.permissions
+          resource    = Hyrax.persister.save(resource: resource)
+
+          resource.permission_manager.acl.permissions = permissions
```
elrayle: I notice that you get permissions from the resource, save the resource, reapply the permissions to the resource and then save the permissions.

Is it because the permissions do not last through a resource being saved?

.circleci/config.yml
```diff
@@ -95,6 +95,7 @@ jobs:
     parallelism: 10
     environment:
       COVERALLS_PARALLEL: true
+      HYRAX_VALKYRIE: true
```
elrayle: Is this creating another test run that will run all the tests again with a config stating that use_valkyrie=true?

no-reply: this turns on valkyrie use for all test runs.

we may want to consider adding this to the matrix. if so, which builds do we need to duplicate?

app/actors/hyrax/actors/base_actor.rb
no-reply: > Is it because the permissions do not last through a resource being saved?

not exactly. the object that `resource` references on L119 still has the permissions (until it's garbage collected). but we've set `resource = Hyrax.persister.save(resource: resource)`, and *this* new object knows nothing about unsaved permissions.

in general, it's desirable to just build the permissions and save them on a completely different lifecycle from `resource`, but since this is the first time `resource` gets saved, we need to save the resource first.

there might be a more illustrative implementation for this... i could think on it

.circleci/config.yml
elrayle: It seems like, until we have removed all references to AF and tests are updated to create and pass valkyrie objects, then we are still supporting AF in our code and need the tests to run with AF and again with Valkyrie.

For file_set_actor_spec, I run all the tests with 
```
[true, false].each do |use_valkyrie|
```

I would think something similar here would work.

no-reply: Lets move to running the full matrix with `HYRAX_VALKYRIE=0` and just the Ruby 2.6/Rails 5.2 build with `HYRAX_VALKYRIE=1` for now.

As we get closer to native-valkyrie throughout, we can flip the order.

`[true, false].each do |use_valkyrie|` won't work here, since the actor initializer and `#create` method signatures are defined elsewhere. There's no nice way to inject this.

Besides, I'm keen to have an environment variable so folks can start testing the more experimental valkyrie-native work in real applications.

Issue #3885, opened 2019-07-12
Hyrax should provide a default unauthorized.png


mark-dce:
### Descriptive summary

I'm running v2.5.1 and I get a deprecation warning asking me to copy `unauthorized.png` but because I generated my application prior to 90a5dc299451c6d4a6da0ba4079f18c7afbf10f9, I don't have that image anywhere in my app. 

### Rationale

 If the gem just provided the file in the correct location, like it does for https://github.com/samvera/hyrax/blob/master/app/assets/images/default.png, I wouldn't have to do anything to my code and my application would handle the deprecation gracefully.

### Expected behavior
 The gem should provide a default version of `unuathorized.png` for institutions that generated their app prior to the release that added the deprecations.

### Actual behavior

>DEPRECATION WARNING: redirect_to default_image is deprecated and will be removed from Hyrax 3.0 (copy unauthorized.png image to directory assets/images instead). (called from authorize_download! at /opt/laevigata/shared/bundle/ruby/2.4.0/gems/hyrax-2.5.1/app/controllers/hyrax/downloads_controller.rb:43)

### Steps to reproduce the behavior

1. Request the thumbnail for a files that is private without logging in
2. Then check your web server error log (or stdout)

### Related work

Link to related tickets or prior related work here.

PR #4044, opened 2019-09-24
Remove unauthorized.png from generator

no-reply:
We now provide a default `unauthorized.png` in the engine. Generating one into
the application isn't needed. This is a follow-up to #4033.

Connected to #3885

* Rely on the engine's `unauthorized.png` instead of a generated one in the application.

* Generate a new application (or delete the local `unauthorized.png` from an existing one.
* Deposit a work with a private file
* Log out
* Request the thumbnail for the file and ensure the file in `unauthorized.png` is retrieved.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/assets_generator.rb
+++ b/lib/generators/hyrax/assets_generator.rb
@@ -25,10 +25,6 @@ class Hyrax::AssetsGenerator < Rails::Generators::Base
     end
   end
 
-  def copy_image_file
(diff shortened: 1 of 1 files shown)


no-reply:
This and #4033 should be backported to 2.x-stable, once merged.

cc: @mark-dce

mark-dce:
Thanks!

PR #4045, opened 2019-09-25
create methods for attaching file_set to work using valkyrie and using active fedora

elrayle:
### Support for valkyrie

There are multiple steps in the attach_to_work method that must change to work with valkyrie.  If each step is separated into its own method, the conversion between active fedora and valkyrie would have to happen multiple times.  So to valkyrize attach_to_work, the entire method is written once for active fedora and again for valkyrie.  A switching method, ‘perform_attach_to_work’ is used to determine which to call.

### Modified behavior

For new works to be able to complete the attach_to_work process when using Valkyrie, the updated work needs to be returned to the caller.  This is because the passed in work does not have an id and the resource save process returns a new instance of the work with the id.  The original work object has no way to reload the values saved by the valkyrie persister.

### Delayed work

Marked pending the test for versions as the valkyrie resource is not maintaining the versions.  This work will be addressed separately.

### DIFF
--- a/app/actors/hyrax/actors/create_with_remote_files_actor.rb
+++ b/app/actors/hyrax/actors/create_with_remote_files_actor.rb
@@ -68,7 +68,7 @@ module Hyrax
           ::FileSet.new(import_url: import_url, label: file_name) do |fs|
             actor = Hyrax::Actors::FileSetActor.new(fs, env.user)
             actor.create_metadata(visibility: env.curation_concern.visibility)
-            actor.attach_to_work(env.curation_concern)
+            env.curation_concern = actor.attach_to_work(env.curation_concern)
             fs.save!
             if uri.scheme == 'file'
               # Turn any %20 into spaces.
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -22,7 +22,7 @@ module Hyrax
         # If the file set doesn't have a title or label assigned, set a default.
         file_set.label ||= label_for(file)
         file_set.title = [file_set.label] if file_set.title.blank?
-        @file_set = perform_save(file_set) # Need to save to get an id # TODO: change this to `save(file_set)` when fully valkyrized
+        @file_set = perform_save(file_set) # Need to save to get an id
         return false unless file_set
         if from_url
           # If ingesting from URL, don't spawn an IngestJob; instead
@@ -70,19 +70,10 @@ module Hyrax
 
       # Adds a FileSet to the work using ore:Aggregations.
       # Locks to ensure that only one process is operating on the list at a time.
+      # @param [Valkyrie::Resource, ActiveFedora::Base] the work
+      # @param [Hash] params being set on file_set
       def attach_to_work(work, file_set_params = {})
-        acquire_lock_for(work.id) do
-          # Ensure we have an up-to-date copy of the members association, so that we append to the end of the list.
-          work.reload unless work.new_record?
-          file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
-          work.ordered_members << file_set
-          work.representative = file_set if work.representative_id.blank?
-          work.thumbnail = file_set if work.thumbnail_id.blank?
-          # Save the work so the association between the work and the file_set is persisted (head_id)
-          # NOTE: the work may not be valid, in which case this save doesn't do anything.
-          work.save
-          Hyrax.config.callback.run(:after_create_fileset, file_set, user)
-        end
+        perform_attach_to_work(work, file_set_params)
       end
       alias attach_file_to_work attach_to_work
       deprecation_deprecate attach_file_to_work: "use attach_to_work instead"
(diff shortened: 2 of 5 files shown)


app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -22,7 +22,7 @@ def create_content(file, relation = :original_file, from_url: false)
         # If the file set doesn't have a title or label assigned, set a default.
         file_set.label ||= label_for(file)
         file_set.title = [file_set.label] if file_set.title.blank?
-        @file_set = perform_save(file_set) # Need to save to get an id # TODO: change this to `save(file_set)` when fully valkyrized
+        @file_set = perform_save(file_set) # Need to save to get an id
```
elrayle: The TODO comment was moved to the perform_save method.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -70,19 +70,10 @@ def create_metadata(file_set_params = {})
 
       # Adds a FileSet to the work using ore:Aggregations.
       # Locks to ensure that only one process is operating on the list at a time.
+      # @param [Valkyrie::Resource, ActiveFedora::Base] the work
+      # @param [Hash] params being set on file_set
       def attach_to_work(work, file_set_params = {})
-        acquire_lock_for(work.id) do
```
elrayle: This method was moved as is to af_attach_to_work.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -189,16 +214,61 @@ def perform_save(object)
...
+            work_to_return = valkyrie_object?(work) ? updated_work : convert_to_af_object(updated_work)
+          else
+            work_to_return = af_attach_to_work(work, file_set_params)
+          end
+          work_to_return
+        end
+
+        # TODO: Remove this method when env passes resources instead of active fedora objects
+        def af_attach_to_work(work, file_set_params)
+          acquire_lock_for(work.id) do
```
elrayle: This is the copy of the original attach_to_work method that operates on active fedora objects.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -189,16 +214,61 @@ def perform_save(object)
...
+            file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
+            work.ordered_members << file_set
+            work.representative = file_set if work.representative_id.blank?
+            work.thumbnail = file_set if work.thumbnail_id.blank?
+            # Save the work so the association between the work and the file_set is persisted (head_id)
+            # NOTE: the work may not be valid, in which case this save doesn't do anything.
+            work.save
+            Hyrax.config.callback.run(:after_create_fileset, file_set, user)
+          end
+          work
```
elrayle: This is the only change to the original attach_to_work.  It now returns the updated work.

app/actors/hyrax/actors/file_set_ordered_members_actor.rb
```diff
@@ -10,6 +10,7 @@ def attach_to_work(work, file_set_params = {})
         file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
         work.representative = file_set if work.representative_id.blank?
         work.thumbnail = file_set if work.thumbnail_id.blank?
+        work
```
elrayle: To be consistent, this version of the `attach_to_work` method also needs to return the updated work.

app/actors/hyrax/actors/create_with_remote_files_actor.rb
```diff
@@ -68,7 +68,7 @@ def create_file_from_url(env, uri, file_name, auth_header = {})
           ::FileSet.new(import_url: import_url, label: file_name) do |fs|
             actor = Hyrax::Actors::FileSetActor.new(fs, env.user)
             actor.create_metadata(visibility: env.curation_concern.visibility)
-            actor.attach_to_work(env.curation_concern)
+            env.curation_concern = actor.attach_to_work(env.curation_concern)
```
no-reply: i'm worried about this, because i've seen nasty failures outside the actor stack when trying to change the referent of `env.curation_concern`.

Are there any end-to-end tests of the actor stack that attach a remote file (and exercise this code)? I think i'd want to see something like that in `spec/features/actor_stack_spec.rb` and/or something that calls this code through a controller before moving forward with this.

I think, e.g., a `actor.create(env); env.curation_concern.persisted?` will be false with this, and lots of dependent code will fail as a result.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -172,14 +163,48 @@ def save(resource)
           false
         end
 
-        # switches between using valkyrie to save or active fedora to save
+        # copy visibility from the work resource to the file_set resource
+        def propagate_visibility(work, file_set)
```
no-reply: is this already handled by the asynchronous `VisibilityCopyJob`? or perhaps `ResourceVisibilityPropagator` should be called directly here?

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -172,14 +163,48 @@ def save(resource)
...
+          file_set.visibility = work.visibility
+          file_set.permission_manager.acl.save
+          Hyrax.persister.save(resource: file_set)
```
no-reply: does the `file_set` need to be saved here?

PR #4046, opened 2019-09-25
add parent works navigator custom query

elrayle:
Add custom navigator for finding parent works with queries...
* find_parent_works - returns work resources
* find_parent_work_ids - returns valkyrie ids of work resources

### DIFF
--- a/app/services/hyrax/custom_queries/navigators/parent_works_navigator.rb
+++ b/app/services/hyrax/custom_queries/navigators/parent_works_navigator.rb
@@ -1,0 +1,33 @@
+# Navigate from a resource to the child works in the resource.
+module Hyrax
+  module CustomQueries
+    module Navigators
+      class ParentWorksNavigator
+        # Define the queries that can be fulfilled by this navigator.
+        def self.queries
+          [:find_parent_works, :find_parent_work_ids]
+        end
+
+        attr_reader :query_service
+
+        def initialize(query_service:)
+          @query_service = query_service
+        end
+
+        # Find parent works of a given resource, and map to Valkyrie Resources
+        # @param [Valkyrie::Resource]
+        # @return [Array<Valkyrie::Resource>]
+        def find_parent_works(resource:)
+          query_service.find_inverse_references_by(resource: resource, property: :member_ids).select(&:work?)
+        end
+
+        # Find the ids of parent works of a given resource, and map to Valkyrie Resources
+        # @param [Valkyrie::Resource]
+        # @return [Array<Valkyrie::ID>]
+        def find_parent_work_ids(resource:)
+          find_parent_works(resource: resource).map(&:id)
+        end
+      end
+    end
+  end
+end
--- a/lib/wings.rb
+++ b/lib/wings.rb
@@ -92,6 +92,7 @@ Valkyrie.config.storage_adapter = :active_fedora
 custom_queries = [Hyrax::CustomQueries::Navigators::ChildCollectionsNavigator,
                   Hyrax::CustomQueries::Navigators::ChildFilesetsNavigator,
                   Hyrax::CustomQueries::Navigators::ChildWorksNavigator,
+                  Hyrax::CustomQueries::Navigators::ParentWorksNavigator,
                   Wings::CustomQueries::FindAccessControl, # override Hyrax::CustomQueries::FindAccessControl
                   Wings::CustomQueries::FindFileMetadata, # override Hyrax::CustomQueries::FindFileMetadata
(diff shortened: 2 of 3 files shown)


app/services/hyrax/custom_queries/navigators/parent_works_navigator.rb
```diff
@@ -0,0 +1,33 @@
...
+        # @return [Array<Valkyrie::Resource>]
+        def find_parent_works(resource:)
+          query_service.find_inverse_references_by(resource: resource, property: :member_ids).select(&:work?)
+        end
+
+        # Find the ids of parent works of a given resource, and map to Valkyrie Resources
+        # @param [Valkyrie::Resource]
+        # @return [Array<Valkyrie::ID>]
+        def find_parent_work_ids(resource:)
+          find_parent_works(resource: resource).map(&:id)
```
no-reply: do we need this query?

i can't think of any easy optimizations, but i'm grimacing about pulling the entire object out of the data store to check whether an ID belongs to a work or not.

is there a known user for this query, and if so, could we explore optimizations there instead?

app/services/hyrax/custom_queries/navigators/parent_works_navigator.rb
```diff
@@ -0,0 +1,33 @@
...
+
+        def initialize(query_service:)
+          @query_service = query_service
+        end
+
+        # Find parent works of a given resource, and map to Valkyrie Resources
+        # @param [Valkyrie::Resource]
+        # @return [Array<Valkyrie::Resource>]
+        def find_parent_works(resource:)
+          query_service.find_inverse_references_by(resource: resource, property: :member_ids).select(&:work?)
```
no-reply: who is implementing `#work?` here? i only see this method defined in `Wings`-specific behavior. if that's right, we can't count on it being available here

elrayle:
@no_reply

### Purpose

I added the custom query for find_parent_works because I need it for the file_set_actor test which tests for a specific parent.

### work? method

This is included in the converted work resource at https://github.com/samvera/hyrax/blob/master/lib/wings/orm_converter.rb#L51-L53 from https://github.com/samvera/hyrax/blob/master/lib/wings/hydra/works/models/concerns/work_valkyrie_behavior.rb#L17-L19.

I can see the argument that this is defined in wings and as such is wings specific.  But a counter argument is that these are the only resources we define, and at this point, they serve as the resources use in all of hyrax.

An alternative is to check the type of class.  This would require that all collections, works, and file_sets inherit from a base collection, base work, and base fileset classes, respectively, all of which inherits from Valkyrie::Resource.

### Efficiency issues

I’ve raised concerns previously about putting all of our relationships (e.g. collection-child_collections, collection-child_works, work-child_works, work-child_filesets) in a single relationship (i.e. members).  Since valkyrie only keeps the ids of these relationships, it requires creating the full object for all query_service results in order to test the type of resource to be able to limit the results to a specific object types in the model.

If each relationship is in a separate attribute, then getting the list of ids for a particular relationship is trivial.  You only incur expensive calls if you specifically request the actual resources instead of the ids.  This can have implications for improving the efficiency of indexing.

There are TODOs in each of the child navigators pointing out these issues, but I did not copy them into the parent work navigator.

### Challenge with multiple relationships

We are taking advantage of the first class support for members.  For example, Valkyrie query_service supports `find_members` method which exists for all adapters.

But, it is not difficult to create a custom query that would work for each of the relationships that would provide `find_related_ids_for(resource:, relationship:)` and `find_related_for(resource:, relationship:)` types of queries that use standard valkyrie queries to fulfill these requests.

### Future

The question is how to proceed going forward.

It seems ok for now to stick with what we have since the members part of the resources is generated from the AF use of members.  But long term, I recommend a separate relationship for each part of the model.  It allows for more efficient queries and it makes the model more transparent.

elrayle:
@no-reply To demonstrate a custom query for finding related resources that supports each relationship being defined in a separate attribute, I created find_related custom query in PR #4049

no-reply:
I'd like to take data model changes off the table as part of this process. It would be challenging to slot those in without far-reaching breaking changes and data migrations. Besides, the PCDM/Hydra Works model is one that was arrived at through extensive community process, so changes to it probably deserve a concrete public proposal and a considerable communication process.

no-reply:
>  But a counter argument is that these are the only resources we define, and at this point, they serve as the resources used in all of hyrax.

I don't think this is true. We have `Hyrax::Resource` already, and are working actively toward advising users about native models. This really needs to be moved out of Wings to be used.

PR #4048, opened 2019-09-26
Use a custom Indexer for AdminSets

no-reply:
Avoid deprecated `index.as` syntax for `AdminSet` data.

Admin Sets should be able to index using the AdminSetIndexer, rather than the
old-style `index.as` syntax.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/indexers/hyrax/admin_set_indexer.rb
+++ b/app/indexers/hyrax/admin_set_indexer.rb
@@ -7,7 +7,12 @@ module Hyrax
     def generate_solr_document
       super.tap do |solr_doc|
         # Makes Admin Sets show under the "Admin Sets" tab
-        solr_doc['generic_type_sim'] = ['Admin Set']
+        solr_doc['generic_type_sim']  = ['Admin Set']
+        solr_doc['alt_title_tesim']   = object.alt_title
+        solr_doc['creator_ssim']      = object.creator
+        solr_doc['description_tesim'] = object.description
+        solr_doc['title_tesim']       = object.title
+        solr_doc['title_sim']         = object.title
       end
     end
   end
--- a/spec/indexers/hyrax/admin_set_indexer_spec.rb
+++ b/spec/indexers/hyrax/admin_set_indexer_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Hyrax::AdminSetIndexer do
   describe 'alt_title' do
     it 'is stored searchable' do
       expect(indexer.generate_solr_document)
-        .to match a_hash_including('alt_title_sim' => admin_set.alt_title)
+        .to match a_hash_including('alt_title_tesim' => admin_set.alt_title)
     end
   end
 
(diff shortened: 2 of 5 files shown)


app/indexers/hyrax/admin_set_indexer.rb
```diff
@@ -8,6 +8,12 @@ def generate_solr_document
       super.tap do |solr_doc|
         # Makes Admin Sets show under the "Admin Sets" tab
         solr_doc['generic_type_sim'] = ['Admin Set']
+
+        solr_doc['alt_title_sim']     = object.alt_title
```
cjcolvar: Are the changes in the suffix here deliberate?  From what I can tell `stored_searchable` should map to `_tesim` which `alt_title` and `description` were both previously indexed as.

no-reply: hmmmm... no, i don't think changes are intended. When working on this I generated a `SolrDocument` using the old approach, and re-implemented (i think?) identically.

Maybe we need an integration test?

no-reply: I finally got around to shimming in a model test for the indexing behavior.

good catch on the keys, not sure how i got that wrong initially.

spec/indexers/hyrax/admin_set_indexer_spec.rb
```diff
@@ -12,4 +12,37 @@
       expect(indexer.generate_solr_document).to match a_hash_including(doc)
     end
   end
+
+  describe 'alt_title' do
+    it 'is stored searchable' do
+      expect(indexer.generate_solr_document)
+        .to match a_hash_including('alt_title_sim' => admin_set.alt_title)
```
cjcolvar:
```suggestion
        .to match a_hash_including('alt_title_tesim' => admin_set.alt_title)
```

spec/indexers/hyrax/admin_set_indexer_spec.rb
```diff
@@ -12,4 +12,37 @@
...
+    it 'is indexed as a symbol' do
+      expect(indexer.generate_solr_document)
+        .to match a_hash_including('creator_ssim' => admin_set.creator)
+    end
+  end
+
+  describe 'description' do
+    it 'is stored searchable' do
+      expect(indexer.generate_solr_document)
+        .to match a_hash_including('description_sim' => admin_set.description)
```
cjcolvar:
```suggestion
        .to match a_hash_including('description_tesim' => admin_set.description)
```

cjcolvar:
@no-reply Maybe this needs a rebase to fix the valkyrie tests?  Or invalidating the cache?

no-reply:
@cjcolvar seems like the failure is an unrelated random error:

`expect(VisibilityCopyJob).to have_been_enqueued`

I've seen this one before, and it looks like a test order issue to me.

PR #4049, opened 2019-09-26
Add find_related custom queries

elrayle:
Add a custom queries to find related ids and related resources for another resource given an attribute that holds ids to the other related resources.

* find_related_ids_for(resource:, relationship:) - returns ids for the identified relationship
* find_related_for(resource:, relationship:) - returns resources for the identified relationship

This can be used to define relationships in separate attributes instead of putting them all in the members relationship.  This allows for efficient retrieval of ids for a single relationship without having to bring the entire related objects into memory.

The spec shows defining a work with a works relationship (for child works) and file_sets relationship (for child file sets).

### DIFF
--- a/app/services/hyrax/custom_queries/find_related.rb
+++ b/app/services/hyrax/custom_queries/find_related.rb
@@ -1,0 +1,31 @@
+# Navigate from a resource to the child works in the resource.
+module Hyrax
+  module CustomQueries
+    class FindRelated
+      # Define the queries that can be fulfilled by this custom query.
+      def self.queries
+        [:find_related_for, :find_related_ids_for]
+      end
+
+      attr_reader :query_service
+
+      def initialize(query_service:)
+        @query_service = query_service
+      end
+
+      # Find related resources, and map to Valkyrie Resources
+      # @param [Valkyrie::Resource]
+      # @return [Array<Valkyrie::Resource>]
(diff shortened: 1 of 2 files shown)


no-reply:
When should I use these queries, vs. `query_service.find_references_by(resource: resource, property: :some_relationship_ids`, and `query_service.find_inverse_references_by(resource: resource, property: :some_relationship_ids`?

In particular, the `.map { |id| query_service.find_by(id: id) }` seems like it avoids the existing abstraction leading to an n+1 query problem that is (hopefully) avoided by existing implementations of the above.

app/services/hyrax/custom_queries/find_related.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      def initialize(query_service:)
+        @query_service = query_service
+      end
+
+      # Find related resources, and map to Valkyrie Resources
+      # @param [Valkyrie::Resource]
+      # @return [Array<Valkyrie::Resource>]
+      def find_related_for(resource:, relationship:)
+        find_related_ids_for(resource: resource, relationship: relationship).map { |id| query_service.find_by(id: id) }
```
no-reply: I have a broader question about how these queries refine the built-in Valkyrie queries. Whatever the outcome of that discussion, I think we need to refine this line prior to merge.

PR #4053, opened 2019-09-27
Introduce a `SimpleWork` for testing, and a factory for it

no-reply:
This "Simple Work" is meant for testing native-Valkyrie works. It has only the
core Hyrax metadata for now, but should be extended with the minumum Work
model (members, etc...; i.e. it should inherit `Hyrax::Work` or something along
those lines).

A factory is added with some basic visibility traits.

For Wings support, `SimpleWorkLegacy`, a work with `CoreMetadata` and
`WorkBehavior` is provided.

* This is a test suite only PR

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/support/simple_work.rb
+++ b/spec/support/simple_work.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Test
+    class SimpleWork < Hyrax::Resource
+      # use the *private* initalizer here to ensure the module gets loaded
+      # use `include Hyrax::Schema(:core_metadata)
+      include Hyrax::Schema(:core_metadata)
+    end
+
+    class SimpleWorkLegacy < ActiveFedora::Base
+      include WorkBehavior
+      include CoreMetadata
+    end
+  end
+end
+
+Wings::ModelRegistry.register(Hyrax::Test::SimpleWork, Hyrax::Test::SimpleWorkLegacy)
--- a/spec/support/simple_work.rb
+++ b/spec/support/simple_work.rb
@@ -2,6 +2,14 @@
 
 module Hyrax
   module Test
+    ##
+    # A generic PCDM Work, with only Hyrax "core" (required) metadata.
+    #
+    # @example building with FactoryBot
+    #   work = FactoryBot.build(:hyrax_work, :public, title: ['Comet in Moominland'])
+    #
+    # @example creating with FactoryBot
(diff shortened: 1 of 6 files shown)


spec/support/simple_work.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Test
+    ##
+    # A generic PCDM Work, with only Hyrax "core" (required) metadata.
+    #
+    # @example building with FactoryBot
+    #   work = FactoryBot.build(:work, :public, title: ['Comet in Moominland'])
```
hweng: Why not use `:hyrax_work`, but instead using `work` here?

spec/support/simple_work.rb
```diff
@@ -0,0 +1,26 @@
...
+    #
+    # @example creating with FactoryBot
+    #   work = FactoryBot.valkyrie_create(:work, :public, title: ['Comet in Moominland'])
```
hweng: Is there some syntactic sugar so that the `valkyrie_create` doesn't need to call `result` method? https://github.com/samvera/hyrax/blob/master/spec/factories/strategies/valkyrie_resource.rb#L2

spec/support/simple_work.rb
no-reply: good catch. i'll fix this.

spec/support/simple_work.rb
no-reply: The class you linked is used by Factory Bot, and conforms to the [Bulid Strategies](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-strategies) interface. The strategy is registered in `spec/spec_helper.rb`.

spec/support/simple_work.rb
no-reply: updated.

spec/support/simple_work.rb
hweng: Gotcher, thanks!

spec/support/simple_work.rb
hweng: but i didn't see there is `title` defined in `:hyrax_work` factory: https://github.com/samvera/hyrax/pull/4053/files#diff-82cb20478049021cf266e1d389b3ebea
Did i miss something?

no-reply: fields only need to be defined on factories to provide default/factory values. `#title` is defined on `SimpleWork`, and that is sufficient here.

hweng: oh, I see. Thanks!

PR #4055, opened 2019-09-30
Convert file ids in resource fileset to pcdm files in AF fileset

elrayle:
### Description

Add `convert_files` to the process that converts a resource to af object

Related to Issue #4060 Wings: Round trip conversion of files does not maintain file relationships

### Description of Issue with this PR

When round tripping from AF to Valk to AF, the existing file is assigned a new ID while being set on the FileSet

### Behaviors 

The described behaviors assume the same starting `fileset1` object created in Issue #3406 

```
fileset1.class # FileSet
fileset1.id #  abc12345
fileset1.original_file.id # ab/c1/23/45/abc12345/files/xyz67890
```

`rt_fileset` is the round trip fileset from the conversion from `fileset1` to a valkyrie `resource` and back to an ActiveFedora based fileset.

#### Expected behavior

```
rt_fileset.class # FileSet
rt_fileset.id # abc12345
rt_fileset.original_file # <Hydra::PCDM::file uri="http://127.0.0.1:8986/rest/test/ab/c1/23/45/abc12345/files/xyz67890">
rt_fileset.filter_files_by_type(RDF::URI.new('http://pcdm.org/use#OriginalFile')) # [<Hydra::PCDM::File uri="http://127.0.0.1:8986/rest/test/ab/c1/23/45/abc12345/files/xyz67890">]
rt_fileset.original_file.id # "ab/c1/23/45/abc12345/files/xyz67890"
rt_fileset.original_file.uri # "http://127.0.0.1:8986/rest/test/ab/c1/23/45/abc12345/files/xyz67890"
rt_fileset.files # [#<Hydra::PCDM::File uri="http://127.0.0.1:8986/rest/test/ab/c1/23/45/abc12345/files/xyz67890" >]
rt_fileset.original_file.id == fileset1.original_file.id # true
```

#### Actual behavior using this PR's `ActiveFedoraConverter #convert_file` method

```
rt_fileset.class # FileSet
rt_fileset.id # abc12345
rt_fileset.original_file # <Hydra::PCDM::file uri="http://127.0.0.1:8986/rest/test/ab/c1/23/45/abc12345/files/DIFFERENT_ID">
rt_fileset.filter_files_by_type(RDF::URI.new('http://pcdm.org/use#OriginalFile')) # [<Hydra::PCDM::file uri="http://127.0.0.1:8986/rest/test/ab/c1/23/45/abc12345/files/DIFFERENT_ID">]
[...]

PR #4056, opened 2019-09-30
[WIP] allow characterize_job to receive valkyrie resource

elrayle:
**[WIP]**

**PR #4057 needs to be merged and this PR rebased before this one is merged.**

NOTE: Passing a valkyrie resource will not work until PR #4055 is fixed.  It is safe to merge this without this fix because the error is only seen when using valkyrie.  Normal functioning with an active fedora object works the same as before this PR.

----

If CharacterizeJob receives a Valkyrie::Resource immediately convert it to an AF object.  This allows callers of this job to be fully valkyrized.

### Prerequisite

* PR #4055 Convert file ids in resource fileset to pcdm files in AF fileset
* PR #4057 add valkyrie_transition_helper to help with the transition

### TODO for Valkyrie Transition

Make CharacterizeJob able to process a Valkyrie::Resource.

### Related Work

This supports full valkyrization of new service Hyrax::Files::IngestFileService (replacing FileActor #ingest_file method).

### DIFF
--- a/app/jobs/characterize_job.rb
+++ b/app/jobs/characterize_job.rb
@@ -3,10 +3,14 @@ class CharacterizeJob < Hyrax::ApplicationJob
 
   # Characterizes the file at 'filepath' if available, otherwise, pulls a copy from the repository
   # and runs characterization on that file.
-  # @param [FileSet] file_set
+  # @param [FileSet, Valkyrie::Resource] file_set
   # @param [String] file_id identifier for a Hydra::PCDM::File
   # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
   def perform(file_set, file_id, filepath = nil)
+    # STATUS: Valkyrie Transistion - currently converts a Valkyrie::Resource to a FileSet as it enters this method.
+    # TODO: Valkyrie Transistion - Next step is to make the characterization process operate on a Valkyrie::Resource.
+    file_set = ValkyrieTransitionHelper.to_active_fedora(file_set)
+
     raise "#{file_set.class.characterization_proxy} was not found for FileSet #{file_set.id}" unless file_set.characterization_proxy?
     filepath = Hyrax::WorkingDirectory.find_or_retrieve(file_id, file_set.id) unless filepath && File.exist?(filepath)
     characterize(file_set, file_id, filepath)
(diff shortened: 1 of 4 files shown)


app/jobs/characterize_job.rb
```diff
@@ -3,10 +3,14 @@ class CharacterizeJob < Hyrax::ApplicationJob
 
   # Characterizes the file at 'filepath' if available, otherwise, pulls a copy from the repository
   # and runs characterization on that file.
-  # @param [FileSet] file_set
+  # @param [FileSet, Valkyrie::Resource] file_set
   # @param [String] file_id identifier for a Hydra::PCDM::File
   # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
   def perform(file_set, file_id, filepath = nil)
```
no-reply: It might be nicer to existing applications to follow the pattern used in `VisibilityCopyJob`:

  - extract the existing `#perform` implementation to a service, making the Job class single responsibility;
  - call the service from a factory like `Hyrax::Charactarizer.for(file_set: file_set).characterize`;
  - add a new implementation, reachable from the factory, for the valkyrie file_set.

This keeps the existing implementation from changing (apart from the extraction), and isolates branching logic to the factory.

elrayle: I am not trying to valkyrize this class at this time.  I am trying to valkyrize another class and just want this one not to barf when it is passed a valkyrie resource instead of a fileset.  This is reflected in the STATUS and TODO comments.

no-reply: if we're not wanting to change this job's behavior, why not do the conversion in the caller? That avoids expanding the method signature.

The Job is also already deserializing from a `GlobalId`. It ought to be possible to queue the job from just an ID and deserialize directly from `ActiveFedora` with no changes here.

elrayle: I am admittedly trying to take the most expedient approach with minimal change. The caller will be valkyrized and I do not want to touch the caller again for the valkyrization transition.  Whereas this class has not yet been evaluated as to what is best to do for the valkyrization transition.  Making this very minor change here allows the other work to be completed.  It is assumed that this class will be evaluated in the future to determine if additional changes are required.

PR #4057, opened 2019-09-30
add valkyrie_transition_helper to help with the transition

elrayle:
During the transition to valkyrie, methods might receive either an ActiveFedora object or a Valkyrie::Resource.  The methods in ValkyrieTransitionHelper handle common actions that may be applied to either type of object and calls the correct method based on the type.

### DIFF
--- a/app/helpers/valkyrie_transition_helper.rb
+++ b/app/helpers/valkyrie_transition_helper.rb
@@ -1,0 +1,87 @@
+class ValkyrieTransitionHelper
+  class << self
+    # Save an object
+    # @param object [ActiveFedora::Base | Valkyrie::Resource] the object to be saved
+    # @return [ActiveFedora::Base | Valkyrie::Resource | FalseClass] the saved object or false if save fails
+    # @raise [ArgumentError]
+    def save(object:)
+      raise ArgumentError, "Object argument must be a Valkyrie::Resource or an ActiveFedora object" unless valkyrie_object?(object) || active_fedora_object?(object)
+      return object.save if active_fedora_object?(object)
+      save_resource(resource: object)
+    end
+
+    # Save a resource
+    # @param resource [Valkyrie::Resource] the resource to be saved
+    # @return [Valkyrie::Resource | FalseClass] the saved resource or false if save fails
+    # @raise [ArgumentError]
+    def save_resource(resource:)
+      raise ArgumentError, "Resource argument must be a Valkyrie::Resource" unless valkyrie_object? resource
+      Hyrax.persister.save(resource: resource)
+    rescue Wings::Valkyrie::Persister::FailedSaveError
+      false
+    end
+
+    # Reload an object
+    # @param object [ActiveFedora::Base | Valkyrie::Resource] the object to be reloaded
+    # @return [ActiveFedora::Base | Valkyrie::Resource | FalseClass] the reloaded object
+    # @raise [Hyrax::ObjectNotFoundError, ArgumentError]
+    def reload(object:)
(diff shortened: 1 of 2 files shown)


no-reply:
@elrayle can you say more about how you'd expect to see this used?

Anything that is going to call this is going to need to add an explicit dependency on this helper, which will then need to be removed in favor of the valkyrie native approach, yes?

If that's right, what's the advantage of using this over factoring native valkyrie in immediately? Doesn't this avoid grappling with our main objective?

elrayle:
Example use case:

I have a method that is fully valkyrized except that it calls CharacterizeJob which requires passing in an ActiveFedora object.  I'm not yet ready to fully valkyrize CharacterizeJob.  But I want it to be able to receive a Valkyrie resource or an ActiveFedora object so it can continue to work in other contexts and allow my method to work completely using valkyrie.  To do this, I add this as the first line of CharacterJob...

```
file_set = ValkyrieTransitionHelper.to_active_fedora(file_set)
```

This ensures that the object acted on in CharacterizeJob is an ActiverFedora object.  Any doesn't require the logic of checking which kind of object was passed in to be repeated all through the code.  When the time comes to make CharacterizeJob only receive a Valkyrie resource, it is easy and clear to identify the code as transitional and safe to remove.

no-reply:
> This ensures that the object acted on in CharacterizeJob is an ActiverFedora object. 

Isn't this what `Wings::ActiveFedoraConverter` does?

I guess I'm thinking that if this Helper is being used, the transition pattern might need work (see also: https://github.com/samvera/hyrax/pull/4056#discussion_r329741673).

If the goal is to provide a "convert me to ActiveFedora if i'm not already ActiveFedora", and a "convert me to valkyrie if i'm not already valkyrie", that seems okay---though I'd want it to live in `lib/wings` and `Wings` so it can be searched/removed with the rest of the transition code. `#save` and `#reload` seem less compelling, though.

elrayle:
I was under the impression that ActiveFedoraConverter only takes a resource.  If you pass it an active fedora object, it would not work.

elrayle:
The transition helper lets you pass in an object that can be either an AF object or a Valkyrie resource and it will return an AF object.  So the caller doesn't have to check what they have before calling.  The transition helper will return the AF object without change if it is passed in and will call the ActiveFedoraConverter if passed a resource.

elrayle:
All these methods are expected to go away as the transition proceeds.  For example, CharacterizeJob in PR 4056 is considered to be 'in-transition'.  So it uses .to_active_fedora in the helper.  When it is fully valkyrized, the reference to the helper class will be removed.

When the transition to valkyrie is complete, there should be no references to this class.

no-reply:
> I was under the impression that ActiveFedoraConverter only takes a resource. If you pass it an active fedora object, it would not work.

Sorry, i suppose it's `ModelTransformer` that's relevant here.

----

I'm feeling a bit anxious at the idea that the caller shouldn't need to know which model code it's converting from. *We* know, as we do the work. Hiding it in the code seems like it would make it harder to pick up the work later.

The existing `use_valkyrie: true/false` pattern has the virtue of making this explicit on both sides: once all the `false` callers are removed, the whole interface can be deprecated. I think that pattern has the downside that it requires the interface to handle both inputs, leading to complex classes. In cases where that's hurting us, I think we should reach for entirely new classes (e.g. `ResourceCharactarizationJob`, requiring the caller to choose explicitly) or patterns like factories that represent long-term extensibility for selecting concrete implementations of abstract behavior.

I'm pretty soft on this as concerns the converter methods, as long as they are in `Wings`: i think all the above is true, but am also good to do what we need to do to move forward. For `#save` and `#reload`, i would be worried about the soundness of any code calling them. I definitely think we should skip implementing them.

PR #4063, opened 2019-10-04
Use Clamby gem for ClamAV integration

cjcolvar:
New versions of ClamAV appear to break the `clamav` gem.  This PR replaces `clamav` with `clamby` for integrating with ClamAV.

### DIFF
--- a/app/models/hyrax/virus_scanner.rb
+++ b/app/models/hyrax/virus_scanner.rb
@@ -1,10 +1,9 @@
-# The default virus scanner Hyrax::Works, ported from hydra_works.
-# If ClamAV is present, it will be used to check for the presence of a virus. If ClamAV is not
-# installed or otherwise not available to your application, Hyrax::Works does no virus checking
+# If Clamby is present, it will be used to check for the presence of a virus. If Clamby is not
+# installed or otherwise not available to your application, Hyrax does no virus checking
 # add assumes files have no viruses.
 #
 # To use a virus checker other than ClamAV:
-#   class MyScanner < Hyrax::Works::VirusScanner
+#   class MyScanner < Hyrax::VirusScanner
 #     def infected?
 #       my_result = Scanner.check_for_viruses(file)
 #       [return true or false]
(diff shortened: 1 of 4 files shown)


app/models/hyrax/virus_scanner.rb
```diff
@@ -30,12 +30,12 @@ def initialize(file)
     # Override this method to use your own virus checking software
     # @return [Boolean]
     def infected?
-      defined?(ClamAV) ? clam_av_scanner : null_scanner
+      defined?(Clamby) ? clam_av_scanner : null_scanner
```
no-reply: Changing the generators immediately seems good, but I think we need a transition plan for existing applications.

I guess technically we could sneak this into 3.0.0, but it seems late in the game for dropping support for the `ClamAV` gem.

What do you think about supporting both?

spec/models/hyrax/virus_scanner_spec.rb
```diff
@@ -10,37 +10,37 @@
...
-      before { allow(ClamAV.instance).to receive(:scanfile).with('/tmp/path').and_return(0) }
+      before { allow(Clamby.instance).to receive(:virus?).with('/tmp/path').and_return(0) }
       it 'returns false with no warning' do
         expect(Hyrax.logger).not_to receive(:warn)
         is_expected.not_to be_infected
       end
     end
     context 'with an infected file' do
-      before { allow(ClamAV.instance).to receive(:scanfile).with('/tmp/path').and_return(1) }
+      before { allow(Clamby.instance).to receive(:virus?).with('/tmp/path').and_return(1) }
```
jeremyf: Should this be `true` instead of `1`?

spec/models/hyrax/virus_scanner_spec.rb
```diff
@@ -10,37 +10,37 @@
...
         end
       end
     end
     after do
-      Object.send(:remove_const, :ClamAV)
+      Object.send(:remove_const, :Clamby)
     end
     context 'with a clean file' do
-      before { allow(ClamAV.instance).to receive(:scanfile).with('/tmp/path').and_return(0) }
+      before { allow(Clamby.instance).to receive(:virus?).with('/tmp/path').and_return(0) }
```
jeremyf: Should this be `false` instead of `0`.

cjcolvar:
Closed in favor of #4127.

PR #4074, opened 2019-10-09
Start to generalize Work Controllers for Valkyrie

no-reply:
Working under the hypothesis that `WorkControllerBehavior` can reasonably
support Valkyrie native Works as well as ActiveFedora ones, begin to test
behavior for Valkyrie works in `works_controller_behavior_spec.rb`.

In the past, we've tested controller behavior using generated work
controllers (`GenericWorkController`) in the test application. Locating the
tests for this sharable module in a file named for a specific implementation
makes it hard to find, and encourages sloppy testing. So we are shifting away
from this approach, and instead bootstrapping a controller and routes directly
for use in testing---if we need additional controllers to test other code paths,
they should be easy to add following this pattern.

A `ValkyrieCanCanAdapter` is added to resolve Resource objects from
ids. Inheriting `CanCan::ModelAdapters::AbstractAdapter` automatically registers
this adapter for use based on `.for_class?`.

A simple version of `Hyrax::Ability` support is added via a specialized
`#extract_subjects` case: whenever the `subject` is a `Hyrax::Resource`, cast it
to an ID and check permissions based on this. This will often (always? )involve
querying permissions from Solr within `Ability`, but should work for most
ability checks. As usage in Controllers grows, we'll need to support Valkyrie
native ability checking.

* Expand Valkyrie support into `WorkControllerBehavior` and `Ability`, in a very preliminary way.
* No functional changes.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/valkyrie_can_can_adapter.rb
+++ b/lib/hyrax/valkyrie_can_can_adapter.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # A `CanCan::ModelAdapter` for valkyrie resources
+  class ValkyrieCanCanAdapter < CanCan::ModelAdapters::AbstractAdapter
+    ##
+    # @param [Class] member_class
+    def self.for_class?(member_class)
+      member_class == Hyrax::Resource ||
+        member_class < Hyrax::Resource
+    end
+
+    ##
+    # @param [Class] model_class
+    # @param [String] id
+    #
+    # @return [Hyrax::Resource]
+    #
+    # @raise Hyrax::ObjectNotFoundError
+    def self.find(_model_class, id)
+      Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: id)
+    end
+  end
+end
--- a/spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
+++ b/spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+RSpec.describe Hyrax::WorksControllerBehavior, type: :controller do
+  let(:paths) { Rails.application.routes.url_helpers }
+
+  controller(ApplicationController) do
+    include Hyrax::WorksControllerBehavior
+
+    self.curation_concern_type = Hyrax::Test::BookResource
+  end
+
+  shared_context 'with a logged in user' do
+    let(:user) { create(:user) }
+
+    before { sign_in user }
+  end
+
+  describe '#edit' do
+    let(:work) { FactoryBot.valkyrie_create(:hyrax_work, :public, alternate_ids: [id]) }
+    let(:id)   { '123' }
+
+    before { Hyrax.persister.save(resource: work) }
+
+    it 'gives a 404 for a missing object' do
+      expect { get :edit, params: { id: 'missing_id' } }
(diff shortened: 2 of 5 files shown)


spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+RSpec.describe Hyrax::WorksControllerBehavior, type: :controller do
+  subject(:controller) { controller_class.new }
+  let(:paths)          { Rails.application.routes.url_helpers }
+  routes               { Rails.application.routes }
+
+  let(:controller_class) do
```
cjcolvar: Does it make sense to use a rspec-defined controller instead of fully declaring it?  Rspec takes care of the cleanup of the classes then.  Here's an example with custom routes: https://relishapp.com/rspec/rspec-rails/v/3-9/docs/controller-specs/anonymous-controller#draw-custom-routes-for-defined-controllers

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    context 'with a logged in user' do
+      include_context 'with a logged in user'
+
+      it 'gives unauthorized' do
+        get :edit, params: { id: id }
+
+        expect(response.status).to eq 401
+      end
```
cjcolvar: Could a happy path test be added to ensure that access can be granted in the proper cases?

lib/hyrax/valkyrie_can_can_adapter.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # A `CanCan::ModelAdapter` for valkyrie resources
+  class ValkyrieCanCanAdapter < CanCan::ModelAdapters::AbstractAdapter
```
cjcolvar: Does it make sense to have specific tests for this class or just let it be tested through ability tests?

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
no-reply: probably. I didn't know about this RSpec feature and can take a look.

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
no-reply: Yes, but it will fail for now. :joy:

I've been waiting on review for this before digging in too deep on `Ability`

lib/hyrax/valkyrie_can_can_adapter.rb
no-reply: unit tests are probably good. i'll look into what's needed here.

no-reply: i pushed up unit tests for this.

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
no-reply: so, it's not clear to me how to define a controller with a mixin and draw routes more succinctly using this interface. it looks to me like we would still need to define a named controller(?)

can you help prod me in the right direction?

PR #4076, opened 2019-10-15
Implement Valkyrie native Collections as `PcdmCollection` 

no-reply:
Add a PCDM style collection in native Valkyrie, with support for
membership. Extract Work membership tests into shared examples for use across
PCDM/Works models.

We name the Collection model `Hyrax::PcdmCollection` to avoid application layer
collisions with `::Collection` when referring to it from a Hyrax module. That
is: we don't want to introduce a `Hyrax::Collection` without warning adopters.

@samvera/hyrax-code-reviewers


hweng:
Are you planning to use dry-transactions + pub/sub style, dry-validation and etc. to replace those functions?

app/models/hyrax/pcdm_collection.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Valkyrie model for Collection domain objects in the Hydra Works model.
+  class PcdmCollection < Hyrax::Resource
+    include Hyrax::Schema(:core_metadata)
+
+    attribute :collection_type_gid, Valkyrie::Types::String
+    attribute :member_ids, Valkyrie::Types::Array.of(Valkyrie::Types::ID).meta(ordered: true)
```
hweng: Is that fine to add `collection_type`, `collection_type=` in the model?

no-reply:
> Is this your intention to keep the model as short as only include attributes?

That's my notion, yes. Attributes and code explicitly involved in managing them.

Obviously, we inherit `Valkyrie::Resource` so other code directly impacting the concerns is bundled there (e.g. `Draper` decoration, we choose to use it). Is this your intention to keep the model as short as only include attributes?

no-reply:
> Are we planning to use dry-transactions + pub/sub style for callbacks?and push indexer, member_object, permission templates to app/services/hyrax/collection? Correct me if i am wrong.

Indexing is already a separate service in Hyrax (see: `app/indexers`), though there's some work we need to do to figure out the ideal approach for Valkyrie. see: https://github.com/samvera/valkyrie/issues/598

I think not `dry-transaction`, since development has been discontinued. I'm playing with its replacement at https://github.com/samvera/hyrax/compare/transaction-to-monads. That's one pattern we could consider using.

----

Re: Callbacks, I think this might be a bit case-by-case.

The *Do notation* approach above could probably be used for many of these cases. the Figgy-style `ChangeSetPersister` approach is probably also worth considering. For some callbacks (e.g. those in `Hyrax.config.callback`), `dry-events` style pub-sub seem good.

For others (e.g. those in `CollectionNesting`), just an external service that needs to be called explicitly (and only sometimes) is probably fine.

----

Membership is already handled here. I'm not sure what else is needed on this front. Any ideas?

----

Permissions are handled by `Hyrax::AccessControl` and `Hyrax::AccessControlList`, while template application is also an external service in Hyrax already. I think some further work on admin sets and permission templates in native Valkyrie is still called for.

----

@hweng does this help answer your questions? I realize there's still a lot to decide here, but I think it's best to take these things one at a time.

app/models/hyrax/pcdm_collection.rb
no-reply: We don't supply a `collection_type` setter without the `_gid` suffix in the current `CollectionBehavior` implementation. There *is* a ticket to refine that implementation, though it might be a `wontfix` until after the ActiveFedora behavior is removed.

I'm not necessarily against putting them in now, but I'm a bit worried about the impact on `Wings`. Keeping symmetry between Valkyrie and ActiveFedora properties simplifies things a bit.

hweng: Gotcha. Sounds reasonable.

hweng:
@no-reply Your approach https://github.com/samvera/hyrax/compare/transaction-to-monads looks good to me.

Regarding the callbacks, do you consider to add them as steps in Transactions?
For instance, `save` lib/hyrax/transactions/steps/save.rb

```
module Hyrax
  module Transactions
    module Steps
       class Save
         include Dry::Monads[:result]

         step :before_save
         Step :call
         step :after_save

        def before_save(&block); end 
    
        def after_save(&block); end 
        .....
```
Or if you have any concern about it?

hweng:
Regarding the `Permissions`:

There are some local `Permissions` code in collection_behavior.rb and admit_set.rb, do you think it is safe to remove or should put it somewhere?
https://github.com/samvera/hyrax/blob/master/app/models/concerns/hyrax/collection_behavior.rb#L131-L162

hweng:
For Membership, Do we still want to keep these methods `add_member_objects`, `member_objects`:

https://github.com/samvera/hyrax/blob/master/app/models/concerns/hyrax/collection_behavior.rb#L55-L83

app/models/hyrax/pcdm_collection.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Valkyrie model for Collection domain objects in the Hydra Works model.
+  class PcdmCollection < Hyrax::Resource
```
elrayle: I am concerned that we are establishing inconsistent naming of the Works Model classes.  See my comments in PR #4084 RE: name of PcdmFileSet

https://github.com/samvera/hyrax/pull/4084/files/a5ab3ca30ca078c8c4ced3da56f5ff6c9b9922d9#r335705853

no-reply: See my follow-up on the linked discussion.

hweng:
@no-reply This PR looks good to me.  I agree with you that there is still a lot to decide, might be best to discuss in the separate tickets.

PR #4082, opened 2019-10-15
Test custom query `FindManyByAlternateIds`

no-reply:
Add tests for the `find_many_by_alternate_ids(alternate_ids:)` query.

Additionally, refactor the query implementation to use an enumerator based
implementation. This avoids querying individually for unused items.

* Adds missing tests for a Valkyrie custom query

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/custom_queries/find_many_by_alternate_ids.rb
+++ b/app/services/hyrax/custom_queries/find_many_by_alternate_ids.rb
@@ -16,11 +16,15 @@ module Hyrax
 
       # implements a combination of two Valkyrie queries:
       # => find_many_by_ids & find_by_alternate_identifier
+      #
       # @param alternate_ids [Enumerator<#to_s>] list of ids
-      # @return [Array<Valkyrie::Resource>, Array<ActiveFedora::Base>]
+      # @return [Enumerable<Valkyrie::Resource>, Enumerable<ActiveFedora::Base>]
       def find_many_by_alternate_ids(alternate_ids:)
-        alternate_ids.uniq.map(&:to_s).each_with_object([]) do |id, resources|
-          resources << query_service.find_by_alternate_identifier(alternate_identifier: id)
+        return enum_for(:find_many_by_alternate_ids, alternate_ids: alternate_ids) unless
+          block_given?
+
+        alternate_ids.uniq do |id|
+          yield query_service.find_by_alternate_identifier(alternate_identifier: id.to_s)
         end
       end
     end
--- a/spec/services/hyrax/custom_queries/find_many_by_alternate_ids_spec.rb
+++ b/spec/services/hyrax/custom_queries/find_many_by_alternate_ids_spec.rb
@@ -1,17 +1,41 @@
-# rubocop:disable RSpec/EmptyExampleGroup
-RSpec.describe Wings::CustomQueries::FindManyByAlternateIds do
-  let(:query_service) { Hyrax.query_service }
-  let(:id_list) { double } # TODO: Stubbed waiting for tests
-  subject { query_service.custom_queries.find_many_by_alternate_ids(alternate_ids: id_list) }
+# frozen_string_literal: true
 
-  describe '.find_many_by_alternate_ids' do
+RSpec.describe Hyrax::CustomQueries::FindManyByAlternateIds do
+  subject(:query_handler) { described_class.new(query_service: query_service) }
+  let(:adapter)           { Valkyrie::MetadataAdapter.find(:test_adapter) }
+  let(:persister)         { adapter.persister }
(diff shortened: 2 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -151,11 +151,14 @@ def main_app
...
-Valkyrie::MetadataAdapter
-  .find(:test_adapter)
-  .query_service
-  .custom_queries
-  .register_query_handler(Hyrax::CustomQueries::FindAccessControl)
+[Hyrax::CustomQueries::FindAccessControl,
+ Hyrax::CustomQueries::FindManyByAlternateIds].each do |handler|
+  Valkyrie::MetadataAdapter
+    .find(:test_adapter)
+    .query_service
```
elrayle: I prefer the approach in PR #4084 as you only have to find the test_adapter once.  Probably not a huge cost but why incur it if not needed.

https://github.com/samvera/hyrax/blob/a5ab3ca30ca078c8c4ced3da56f5ff6c9b9922d9/spec/spec_helper.rb#L155-L158

no-reply: happy to make the change.

app/services/hyrax/custom_queries/find_many_by_alternate_ids.rb
```diff
@@ -16,11 +16,15 @@ def initialize(query_service:)
...
       # => find_many_by_ids & find_by_alternate_identifier
+      #
       # @param alternate_ids [Enumerator<#to_s>] list of ids
-      # @return [Array<Valkyrie::Resource>, Array<ActiveFedora::Base>]
+      # @return [Enumerable<Valkyrie::Resource>, Enumerable<ActiveFedora::Base>]
       def find_many_by_alternate_ids(alternate_ids:)
-        alternate_ids.uniq.map(&:to_s).each_with_object([]) do |id, resources|
-          resources << query_service.find_by_alternate_identifier(alternate_identifier: id)
+        return enum_for(:find_many_by_alternate_ids, alternate_ids: alternate_ids) unless
+          block_given?
```
elrayle: None of the tests use a block with this.  Seems like it should be tested if we are going to support this.

Is the thought that you might want to use the block to process all the returned resources?

no-reply: this is a pretty basic enumerator implementation. `enum_for` uses the block to yield the items, and the tests hit `enum_for`.

all the existing tests hit both this line and the `yield` code below.

spec/spec_helper.rb
no-reply: how does the new version look?

elrayle:
Philosophical question here...

When would we want to call this?  Won't the relationships have the primary valkyrie id of objects they are related to in the member_ids list?  I know initially id = alternate_id (for wings).  But that won't necessarily be the case when other adapters are used.

As an alternate example...
non-Wings Valkyrie query services - find_by(id: id) looks for a matching primary id
Wings query services - find_by(id: id) assumes id is the alternate id and calls find_by_alternate_id

elrayle:
Following on, wouldn't an implementation of find_many_by_ids...
non-Wings Valkyrie query services - find_by(id: id) looks for a matching primary id
Wings query services - find_by(id: id) assumes id is the alternate id and calls find_by_alternate_id

elrayle:
Following on this, wouldn't it make more sense to redefine find_many_by_ids to be
* non-Wings Valkyrie query services - find_many_by_ids(ids:) looks for a matching primary id
* Wings query services - find_many_by_ids(ids:) should assume ids are alternate ids and should call custom query find_many_by_alternate_ids.

Then all references to find_many_by_alternate_ids in non-wings code could be switched to find_many_by_ids.

This decision may impact our long term goal of working with any valkyrie adapter.

no-reply:
> When would we want to call this? Won't the relationships have the primary valkyrie id of objects they are related to in the member_ids list? I know initially id = alternate_id (for wings). But that won't necessarily be the case when other adapters are used.

I'm not sure this question is in-scope for this PR... this just adds tests for a long-standing query and refactors it for efficiency.

no-reply:
> Then all references to find_many_by_alternate_ids in non-wings code could be switched to find_many_by_ids.

I'm not following this at all.

We use `alternate_ids` to support bespoke ids for non-wings adapters, since:
  - Hyrax supports shortened/custom id formats, and needs to continue to do so; and
  - Valkyrie adapters don't always allow custom ids in the `id` field.

Relying on `find_many_by_ids` outside of Wings would require dropping support for Noids and any existing application-side id minters.

elrayle:
Moved discussion to Issue #4085

no-reply:
rebased and re-pushed

PR #4086, opened 2019-10-16
Make `Hyrax::FileSet` a class

no-reply:
`Hyrax::FileSet` will be the name of a `Valkyrie::Resource` model. This begins
that transition by moving it from a module to a class inheriting `Resource`.

@samvera/hyrax-code-reviewers


app/services/hyrax/statistics/file_sets/by_format.rb
```diff
@@ -1,6 +1,6 @@
 module Hyrax
   module Statistics
-    module FileSets
+    class FileSets
```
elrayle: Is this one needed since it is under the Statistics module?

PR #4090, opened 2019-10-18
Add a Valkyrie-based Admin Set model (`AdministrativeSet`)

no-reply:
Implement a basic Valkryie Admin Set model. Add support for admin sets to
Valkyrie native Work models.

We played with the idea of making `AdministrativeSet` a subclass of
`PcdmCollection` but the assumption of ordered membership in collection is
problematic, here. We don't want order in Admin Sets, and the AF models won't
support it.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/administrative_set.rb
+++ b/app/models/hyrax/administrative_set.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Valkyrie model for Admin Set domain objects.
+  class AdministrativeSet < Hyrax::Resource
+    attribute :alt_title,   Valkyrie::Types::Set.of(Valkyrie::Types::String)
+    attribute :creator,     Valkyrie::Types::Set.of(Valkyrie::Types::String)
+    attribute :description, Valkyrie::Types::Set.of(Valkyrie::Types::String)
+    attribute :title,       Valkyrie::Types::Set.of(Valkyrie::Types::String)
+  end
+end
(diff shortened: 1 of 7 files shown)


app/models/hyrax/administrative_set.rb
```diff
@@ -0,0 +1,12 @@
...
+module Hyrax
+  ##
+  # Valkyrie model for Admin Set domain objects.
+  class AdministrativeSet < Hyrax::Resource
+    attribute :alt_title,   Valkyrie::Types::Set.of(Valkyrie::Types::String)
+    attribute :creator,     Valkyrie::Types::Set.of(Valkyrie::Types::String)
+    attribute :description, Valkyrie::Types::Set.of(Valkyrie::Types::String)
+    attribute :title,       Valkyrie::Types::Set.of(Valkyrie::Types::String)
+  end
+end
```
hweng: Doesn't it need the following attributes:

date_created: Valkyrie::Types::DateTime,
date_modified: Valkyrie::Types::DateTime,

no-reply: oh, probably!

I can come back review the current adminset models a bit more thoroughly.

no-reply: Turns out `AdminSet` doesn't provide a Hyrax managed `date_created` or `date_modified`.

I think there is more work to do on AdminSet models (representative and thumbnail, definitely). Should we evaluate this as-is and see about getting the rest done on a future Pull request?

hweng: 👍

PR #4107, opened 2019-10-21
Implement Resource state in a model agnostic way

no-reply:
Extract resource state/status to `Hyrax::ResourceStatus`. This service is
agnostic to Valkyrie/ActiveFedora models, but respects Fedora "resource status"
URIs. https://fedora.info/definitions/1/0/access/2016/06/02/ObjState

Additionally, add a `#state` attribute to `Hyrax::Work`. This readds an
attribute previously handled by `Hyrax::WorkBehavior`.

Though Hyrax only uses/respects Fedora's "inactive" status, the utility we
provide supports three statuses from the Fedora Resource Status vocabulary:
active, inactive, and deleted.

* Refactors `Suppressible` to rely on a more complete implementation of the Fedora `objState` ontology.
* The new implementation supports ActiveFedora and Valkyrie models interchangeably.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/suppressible.rb
+++ b/app/models/concerns/hyrax/suppressible.rb
@@ -14,9 +14,7 @@ module Hyrax
     #
     # Override this method if you have some criteria by which records should not display in the search results.
     def suppressed?
-      return false if state.nil?
-
-      state == Vocab::FedoraResourceStatus.inactive
+      Hyrax::ResourceStatus.new(resource: self).inactive?
     end
 
     ##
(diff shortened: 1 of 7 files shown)


no-reply:
The CI failures are due to https://github.com/samvera/valkyrie/issues/772

no-reply:
this now depends on https://github.com/samvera/valkyrie/pull/810.

or we can retain the hack in 30389f1. if we do the latter, we should maybe make the hack dependent on the valkyrie version?? or just include the entire RDF override in Hyrax instead of valkyrie? i guess we have several options, but i think the best might be to press/wait for a release with https://github.com/samvera/valkyrie/pull/810

no-reply:
samvera/valkyrie#810 is merged, but not released

i'm now thinking it would be best to move forward with this regardless.

no-reply:
this should now be rebased on top of #4180.

no-reply:
rebased.

this now has no special work arounds, and i think it's just a clean implementation of the desired logic.

mjgiarlo:
Good change! Thanks, @no-reply

PR #4110, opened 2019-10-21
Rewrite Arel.sql calls to other arel methods

aploshay:
It looks like cases of
```ruby
.pluck(Arel.sql('DISTINCT field_name'))
```
can be equivalently rewritten as
```ruby
.select(:field_name).distinct.pluck(:field_name)
```
to avoid the raw SQL.

@geekscruff

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -73,7 +73,9 @@ module Hyrax
         ids = PermissionTemplateAccess.for_user(ability: self,
                                                 access: ['deposit', 'manage'])
                                       .joins(:permission_template)
-                                      .pluck(Arel.sql('DISTINCT source_id'))
+                                      .select(:source_id)
+                                      .distinct
+                                      .pluck(:source_id)
         query = "_query_:\"{!raw f=has_model_ssim}AdminSet\" AND {!terms f=id}#{ids.join(',')}"
         Hyrax::SolrService.count(query).positive?
       end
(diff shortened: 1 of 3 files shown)


aploshay:
@no-reply Didn't mean to dismiss your review, was just trying to force a rebuild now that I _think_ I have successfully changed the organizations's security settings to allow 3rd party orbs...

no-reply:
The issue breaking this now is a known issue. I'm working on patching it now.

no-reply:
Fixed in #4122

no-reply:
This should be rebased now that #4122 is merged

aploshay:
> This should be rebased now that #4122 is merged

Rebased!

cjcolvar:
@aploshay It looks like the build failed because it is running with with the hyrax circleci configured 10x parallelism but your fork's circleci plan only supports 4 concurrent containers.  Do you have permissions to push this to a branch on hyrax and create the PR from there?

aploshay:
@cjcolvar Look like I _don't_ have permissions to push this to `samvera/hyrax` and create the PR from there.  Should I request permissions, or do you just want to do it, or something else?  Looks like we can't bump up the concurrent container limit for `iublibtech` without paying monies?

cjcolvar:
@aploshay I just invited you to the samvera github org so you should be able to push this as a branch on hyrax.
@no-reply We should figure out how to have the 10x parallelism be optional so we can still accept PRs from forks.

aploshay:
Closing, replacing with #4126 equivalent based on `samvera/hyrax` instead of `IUBLibTech/hyrax`

PR #4112, opened 2019-10-21
Compile javascript into an NPM package

jcoyne:
This will allow us to support webpacker as well as continue support of sprockets in the manner used by Blacklight.

The javascript is built by npm from sources in `app/javascript` into a bundle in `app/assets/javascripts/hyrax.js`. This file should not be edited by hand as any changes would be overwritten. When any of the javascript components in the gem are changed, this bundle should be rebuild with the following steps:

1. [Install npm](https://www.npmjs.com/get-npm)
1. run `npm install` to download dependencies
1. run `npm run js-compile-bundle` to build the bundle
1. run `npm publish` to push the javascript package to https://npmjs.org/package/hyrax-frontend

@samvera/hyrax-code-reviewers


package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "hyrax-frontend",
+  "version": "3.0.0-alpha",
+  "description": "[![Build Status](https://travis-ci.org/samvera/hyrax.png?branch=master)](https://travis-ci.org/projectblacklight/blacklight) [![Gem Version](https://badge.fury.io/rb/blacklight.png)](http://badge.fury.io/rb/blacklight) [![Coverage Status](https://coveralls.io/repos/github/projectblacklight/blacklight/badge.svg?branch=master)](https://coveralls.io/github/projectblacklight/blacklight?branch=master)",
```
jcoyne: This is wrong.

Issue #4102, opened 2019-10-21
Allow GrantReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work


elrayle:
### Descriptive summary

Convert GrantReadToMembersJob to allow it to receive either an ActiveFedora based FileSet or a Valkyrie::Resource equivalent Work.

### Rationale

Allow jobs to execute correctly with existing AF implementation or the wings Valkyrie implementation.

### Expected behavior

GrantReadToMembersJob receives a Valkyrie::Resource and performs GrantReadJob for each member FileSet.

To get the list of FileSet ids from a work resource, use...

```
file_set_ids = Hyrax.query_service.custom_queries.find_child_fileset_ids(resource: work_resource)
```

Coordinate with Issue #4105 to pass `use_valkyrie: true` when a Valkyrie::Resource was received and false if not.

### Actual behavior

GrantReadToMembersJob raises an exception (perhaps: NoMethodFound for #search_with_conditions) when passed a Valkyrie::Resource.

### Approach for Conversion

See [Pattern for Jobs Receiving Parameter that may be an AF Object or a Valkyrie Resource](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727)

### Steps to reproduce the behavior

1. Need to add tests that validate that the same tests pass whether a Valkyrie::Resource or an AF object is passed.

See example of processing both objects in...  [file_set_actor_spec.rb](https://github.com/samvera/hyrax/blob/master/spec/actors/hyrax/actors/file_set_actor_spec.rb#L6)

For testing this job, you can use something like...
```
RSpec.describe GrantReadToMembersJob do
  [true, false].each do |test_valkyrie|
    context "when test_valkyrie is #{test_valkyrie}" do
      let(:af_work) { create(:work) }
      let(:valk_file_set) { af_work.valkyrie_resource }
      let(:work) { test_valkyrie ? valk_file_set : af_work }
```
### Related work

* Issue #4104 Process GrantEditJob with either an AF object or a Valkyrie resource based on use_valkyrie parameter
* Issue #4105 Process GrantReadJob with either an AF object or a Valkyrie resource based on use_valkyrie parameter
* Issue #4106 Process RevokeEditJob with either an AF object or a Valkyrie resource based on use_valkyrie parameter
[...]

PR #4117, opened 2019-10-21
Valkyrize grant_edit_job

bkeese:
refs #4101 
refs #4104

Adds valkyrie support to grant_edit_members_job and grant_edit_job

@samvera/hyrax-code-reviewers

PR #4118, opened 2019-10-21
Valkyrize grant_read_job, grant_read_to_members_job

### LINKED ISSUE #4105: Process GrantReadJob with either an AF object or a Valkyrie resource based on use_valkyrie parameter
### Descriptive summary

Add use_valkyrie parameter to GrantReadJob #perform method.  When true, use Valkyrie to get the file set object and set permissions on the valkyrie resource.  Otherwise, continue to process the ActiveFedora fileset.

### Rationale

Allow jobs to execute correctly with existing AF implementation or the wings Valkyrie implementation.

### Expected behavior

GrantReadJob #perform method receives parameters `file_set_id, user_key, use_valkyrie: true | false`, finding a valkyrie resource when use_valkyrie is true or  an AF object if use_valkyrie is false, and in both cases correctly grants read access to the user for the fileset.  Parameter `use_valkyrie` defaults to `: Hyrax.config.use_valkyrie?`

To find a fileset by id for a resource, use...

```
id = Valkyrie::ID.new(file_set_id)
file_set_resource = Hyrax.query_service.find_by(id: file_set_id)
```

The process for setting the read access is the same for both types of objects.

To save a fileset resource, use...

```
Hyrax.persister.save(resource: file_set_resource)
```

### Actual behavior

GrantReadJob #perform method raises an exception (e.g. Unknown Argument) when use_valkyrie parameter is passed in.

### Approach for Conversion

See [Pattern for Jobs Receiving Parameter that may be an AF Object or a Valkyrie Resource](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727)

1. Need to add tests that validate that the same tests pass whether a Valkyrie::Resource or an AF object is passed.

See example of processing both objects in...  [file_set_actor_spec.rb](https://github.com/samvera/hyrax/blob/master/spec/actors/hyrax/actors/file_set_actor_spec.rb#L6)

For testing this job, you can use something like...
```
RSpec.describe GrantReadJob do
  [true, false].each do |use_valkyrie|
    context "when use_valkyrie is #{use_valkyrie}" do
      ...
          described_class.perform_now(file_set.id, depositor.user_key, use_valkyrie: use_valkyrie)
```
### Related work

* Issue #4104 Process GrantEditJob with either an AF object or a Valkyrie resource based on use_valkyrie parameter
* Issue #4101 Allow GrantEditToMembersJob to receive either an AF object or a Valkyrie resource as the Work
* Issue #4102 Allow GrantReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work
* Issue #4103 Allow RevokeReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work
* Issue #4100 Allow CharacterizeJob to receive either an AF object or a Valkyrie resource as the FileSet
* Issue #4099 Add GlobalID::Identification for resources passed to jobs

rotated8:
Fixes #4105, #4102  by adding valkyrie support to the grant_read_job and grant_read_to_members_job

### DIFF
--- a/app/jobs/hyrax/grant_read_job.rb
+++ b/app/jobs/hyrax/grant_read_job.rb
@@ -1,14 +1,23 @@
 module Hyrax
   # Grants the user's read access on the provided FileSet
-  class GrantReadJob < ApplicationJob
+  class GrantReadJob < Hyrax::ApplicationJob
     queue_as Hyrax.config.ingest_queue_name
 
     # @param [String] file_set_id - the identifier of the object to grant access to
     # @param [String] user_key - the user to add
-    def perform(file_set_id, user_key)
-      file_set = ::FileSet.find(file_set_id)
-      file_set.read_users += [user_key]
-      file_set.save!
+    # @param [Boolean] use_valkyrie - use valkyrie resources for this operation?
+    def perform(file_set_id, user_key, use_valkyrie: Hyrax.config.use_valkyrie?)
+      if use_valkyrie
+        file_set_id = Valkyrie::ID.new(file_set_id)
+        file_set_resource = Hyrax.query_service.find_by(id: file_set_id)
+        acl = Hyrax::AccessControlList.new(resource: file_set_resource)
+        acl.grant(:read).to(::User.find_by_user_key(user_key))
+        acl.save
+      else
+        file_set = ::FileSet.find(file_set_id)
+        file_set.read_users += [user_key]
+        file_set.save!
+      end
     end
   end
 end
--- a/app/jobs/hyrax/grant_read_to_members_job.rb
+++ b/app/jobs/hyrax/grant_read_to_members_job.rb
@@ -1,15 +1,16 @@
 module Hyrax
   # Grants read access for the supplied user for the members attached to a work
-  class GrantReadToMembersJob < ApplicationJob
+  class GrantReadToMembersJob < Hyrax::ApplicationJob
     queue_as Hyrax.config.ingest_queue_name
 
     # @param [ActiveFedora::Base] work - the work object
     # @param [String] user_key - the user to add
-    def perform(work, user_key)
+    # @param [Boolean] use_valkyrie - use valkyrie objects for this operation?
+    def perform(work, user_key, use_valkyrie: Hyrax.config.use_valkyrie?)
       # Iterate over ids because reifying objects is slow.
       file_set_ids(work).each do |file_set_id|
         # Call this synchronously, since we're already in a job
-        GrantReadJob.perform_now(file_set_id, user_key)
+        GrantReadJob.perform_now(file_set_id, user_key, use_valkyrie: use_valkyrie)
       end
     end
 
@@ -18,7 +19,12 @@ module Hyrax
       # Filter the member ids and return only the FileSet ids (filter out child works)
       # @return [Array<String>] the file set ids
       def file_set_ids(work)
-        ::FileSet.search_with_conditions(id: work.member_ids).map(&:id)
+        case work
+        when ActiveFedora::Base
+          ::FileSet.search_with_conditions(id: work.member_ids).map(&:id)
+        when Valkyrie::Resource
+          Hyrax.query_service.custom_queries.find_child_fileset_ids(resource: work)
+        end
       end
   end
 end
--- a/spec/jobs/hyrax/grant_read_job_spec.rb
+++ b/spec/jobs/hyrax/grant_read_job_spec.rb
@@ -1,11 +1,23 @@
 RSpec.describe Hyrax::GrantReadJob do
   let(:depositor) { create(:user) }
-  let(:file_set) { build(:file_set) }
 
-  it 'grants a user read access to a FileSet' do
-    expect(FileSet).to receive(:find).with(file_set.id).and_return(file_set)
-    expect(file_set).to receive(:read_users=).with(array_including(depositor.user_key))
-    expect(file_set).to receive(:save!)
-    described_class.perform_now(file_set.id, depositor.user_key)
+  context "when use_valkyrie is false" do
+    let(:file_set) { create(:file_set, user: depositor) }
+
+    it 'grants a user read access to a FileSet' do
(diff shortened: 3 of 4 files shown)


app/jobs/hyrax/grant_read_job.rb
```diff
@@ -5,10 +5,19 @@ class GrantReadJob < ApplicationJob
...
     # @param [String] file_set_id - the identifier of the object to grant access to
     # @param [String] user_key - the user to add
-    def perform(file_set_id, user_key)
-      file_set = ::FileSet.find(file_set_id)
-      file_set.read_users += [user_key]
-      file_set.save!
+    # @param [Boolean] use_valkyrie - use valkyrie resources for this operation?
+    def perform(file_set_id, user_key, use_valkyrie: Hyrax.config.use_valkyrie?)
+      if use_valkyrie
+        file_set_id = Valkyrie::ID.new(file_set_id)
```
no-reply: Looks good. There's a comment at https://github.com/samvera/hyrax/pull/4118 that shows off a little grant DSL that may be appropriate for this case.

As I said there, I'll leave it up to y'all whether that DSL is useful for this case. 

See also: https://github.com/samvera/hyrax/wiki/Hyrax-Valkyrie-Usage-Guide#grantrevoke

spec/jobs/hyrax/grant_read_job_spec.rb
```diff
@@ -1,11 +1,22 @@
...
+  context "when use_valkyrie is not set" do
+    it 'grants a user read access to a FileSet' do
+      described_class.perform_now(file_set.id, depositor.user_key)
+      file_set.reload
+      expect(file_set.read_users).to include depositor.user_key
+    end
+  end
+
+  [true, false].each do |use_valkyrie|
+    context "when use_valkyrie is #{use_valkyrie}" do
```
no-reply: Do we think it's worth refactoring tests to have a valkyrie native version?

Sooner or later we'll need to do this, so maybe now's the time???

elrayle: Here's my take on the loop for tests as we are in transition from AF to valkyrie.  In a lot of cases, the tests and their outcomes are the same for both approaches.  By using the same tests for both, the AF tests serve as regression tests showing the valkyrie implementation has the same effect.

Also, for these jobs, the tests are super short, but for other classes, the list of tests can be quite long.  Maintaining a set of regression tests for both approaches can be long.

In some cases, the valkyrie test is significantly different and we will have to write a different test.  But when that is not the case, we benefit from the tests remaining the same.

app/jobs/hyrax/grant_read_job.rb
rotated8: This comment? https://github.com/samvera/hyrax/pull/4117#discussion_r337204897

spec/jobs/hyrax/grant_read_job_spec.rb
```diff
@@ -1,11 +1,28 @@
...
+    it 'grants a user read access to a FileSet' do
+      described_class.perform_now(file_set.id, depositor.user_key)
+      file_set.reload
+      expect(file_set.read_users).to include depositor.user_key
+    end
+  end
+
+  context "when use_valkyrie is false" do
+    it 'grants a user read access to a FileSet' do
+      described_class.perform_now(file_set.id, depositor.user_key, use_valkyrie: false)
```
elrayle: To be a pure valkyrie test, create a valkyrie file_set using...
```
let(:valk_file_set) { valkyrie_create(:hydra_file_set) }
```
and then in the `perform_now` use `valk_file_set.id`

rotated8: `let(:valk_file_set) { valkyrie_create(:hyrax_file_set) }` works for me.

spec/jobs/hyrax/grant_read_job_spec.rb
elrayle: Because these job tests are so small, we are going with 2 tests... 1) pure AF, 2) pure valkyrie

app/jobs/hyrax/grant_read_to_members_job.rb
```diff
@@ -1,15 +1,16 @@
...
     # @param [ActiveFedora::Base] work - the work object
     # @param [String] user_key - the user to add
-    def perform(work, user_key)
+    # @param [Boolean] use_valkyrie - use valkyrie objects for this operation?
+    def perform(work, user_key, use_valkyrie: Hyrax.config.use_valkyrie?)
       # Iterate over ids because reifying objects is slow.
       file_set_ids(work).each do |file_set_id|
         # Call this synchronously, since we're already in a job
-        GrantReadJob.perform_now(file_set_id, user_key)
+        GrantReadJob.perform_now(file_set_id, user_key, use_valkyrie)
```
elrayle: This line should be...
```
        GrantReadJob.perform_now(file_set_id, user_key, use_valkyrie: use_valkyrie)
```

spec/jobs/hyrax/grant_read_to_members_job_spec.rb
```diff
@@ -1,16 +1,32 @@
...
-    file_set_ids.each do |file_set_id|
-      expect(Hyrax::GrantReadJob).to receive(:perform_now).with(file_set_id, depositor.user_key).once
+  context "when use_valkyrie is true" do
+    let(:file_set1) { valkyrie_create(:hyrax_file_set) }
+    let(:file_set2) { valkyrie_create(:hyrax_file_set) }
+    let(:work) { valkyrie_create(:hyrax_work, member_ids: [file_set1.id, file_set2.id]) }
+
+    it 'loops over FileSet IDs, spawning a job for each' do
+      work.member_ids.each do |file_set_id|
+        expect(Hyrax::GrantReadJob).to receive(:perform_now).with(file_set_id, depositor.user_key, true).once
```
elrayle: With the recommended named parameter change in the original code, you can use `use_valkyrie: true` here.

elrayle:
Cherry picked into PR #4123 extract out common permission job code.

PR #4120, opened 2019-10-21
Update hydra editor in 2.x stable

geekscruff:
Upgrade hydra-editor and simple_form, and fix failing tests caused by changes to simple_form.

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -112,5 +112,5 @@ SUMMARY
   #
   # Pin sass-rails to 5.x because rails 5.x apps have this same dependency in their generated Gemfiles
   # See https://github.com/samvera/hyrax/issues/3919
-  spec.add_dependency 'sass-rails',  '~> 5.0'
+  spec.add_dependency 'sass-rails', '~> 5.0'
 end
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -49,7 +49,7 @@ SUMMARY
   spec.add_dependency 'flot-rails', '~> 0.0.6'
   spec.add_dependency 'font-awesome-rails', '~> 4.2'
   spec.add_dependency 'hydra-derivatives', '~> 3.3'
-  spec.add_dependency 'hydra-editor', '>= 3.3', '< 5.0'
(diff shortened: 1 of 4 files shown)


hyrax.gemspec
```diff
@@ -49,7 +49,7 @@ SUMMARY
   spec.add_dependency 'flot-rails', '~> 0.0.6'
   spec.add_dependency 'font-awesome-rails', '~> 4.2'
   spec.add_dependency 'hydra-derivatives', '~> 3.3'
-  spec.add_dependency 'hydra-editor', '>= 3.3', '< 5.0'
+  spec.add_dependency 'hydra-editor', '~> 5.0'
```
no-reply: is it possible to do this without dropping `hydra-editor` 3.3 and 4.x support?

geekscruff: I've pushed an update ... would the removal of the to_model method in `app/forms/hyrax/forms/workflow_responsibility_form.rb` be a problem?

geekscruff: well the tests still pass

hyrax.gemspec
```diff
@@ -49,7 +49,7 @@ SUMMARY
   spec.add_dependency 'flot-rails', '~> 0.0.6'
   spec.add_dependency 'font-awesome-rails', '~> 4.2'
   spec.add_dependency 'hydra-derivatives', '~> 3.3'
-  spec.add_dependency 'hydra-editor', '>= 3.3', '< 5.0'
+  spec.add_dependency 'hydra-editor', '>= 3.0', '~> 5.0'
```
cjcolvar: Does this allow a 3.x hydra-editor?  I would think that the `~> 5.0` would force to only 5.x releases.  Maybe `< 6` would work?

geekscruff: The idea was to allow for 3.x yes so I've made that change. Do you think the change that removes to_model from the WorkflowResponsiblity would be problematic?

geekscruff: Ideally I'd like to force the upgrade on people ;)

geekscruff: Tests still passing with that change @cjcolvar

PR #4121, opened 2019-10-21
Removing an out of date comment

jeremyf:
In reviewing the Gemspec, I noticed a comment that did not match the
underlying reality.


no-reply:
If we can get #4122 in, i'd be happy to merge this over CI objections.

It's failing due to the issue patched there, along with the bump to the enginecart cache owing to the gemspec change.

PR #4119, opened 2019-10-21
Valkyrizes job that revokes :edit permissions

afred:
Fixes #4102 

Modifies the job for revoking edit permissions for a given user on a given resource to be compatible when using Valkyrie resources.

PR #4123, opened 2019-10-22
extract out common permission job code

elrayle:
Replaces PR #4117, #4118, #4119 - The commits of those PRs are cherry picked into this PR.

The additional commit in this PR primarily extracts out common code for the permission jobs into Hyrax::PermissionJob and Hyrax::MembersPermissionJob.  There are also a few minor tweaks to make the code in the two types of permission jobs match each other.

### DIFF
--- a/app/jobs/concerns/hyrax/members_permission_job_behavior.rb
+++ b/app/jobs/concerns/hyrax/members_permission_job_behavior.rb
@@ -1,6 +1,12 @@
 module Hyrax
   # Grants read access for the supplied user for the members attached to a work
   module MembersPermissionJobBehavior
+    extend ActiveSupport::Concern
+
+    included do
+      queue_as Hyrax.config.ingest_queue_name
+    end
+
     private
 
       # Filter the member ids and return only the FileSet ids (filter out child works)
--- a/app/jobs/concerns/hyrax/members_permission_job_behavior.rb
+++ b/app/jobs/concerns/hyrax/members_permission_job_behavior.rb
@@ -1,8 +1,6 @@
 module Hyrax
   # Grants read access for the supplied user for the members attached to a work
-  class MembersPermissionJob < ApplicationJob
-    queue_as Hyrax.config.ingest_queue_name
-
+  module MembersPermissionJobBehavior
     private
 
       # Filter the member ids and return only the FileSet ids (filter out child works)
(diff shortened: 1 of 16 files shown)


elrayle:
@afred @rotated8 @bkeese Can you review to be sure this looks good to you?

elrayle:
@bkeese 
> The param and its type are in the wrong order here

I don't see this referencing a specific line, but I'm guessing you are referring to these param comments.

```
    # @param [String] file_set_id - the identifier of the object to grant access to
    # @param [String] user_key - the user to add
    # @param use_valkyrie [Boolean] whether to use valkyrie support
    def perform(file_set_id, user_key, use_valkyrie: Hyrax.config.use_valkyrie?)
```

I've always used (and seen others use)...
* `[type] description` for unnamed parameters
* `name [type] description` for named parameters

But I looked it up at [yardoc](https://rubydoc.info/gems/yard/file/docs/Tags.md#param) and they always put the name first whether it is a named parameter or not.

app/jobs/hyrax/members_permission_job.rb
```diff
@@ -0,0 +1,23 @@
+module Hyrax
+  # Grants read access for the supplied user for the members attached to a work
+  class MembersPermissionJob < ApplicationJob
```
no-reply: wondering if these might not be better as module mixins, since they aren't queue-able jobs in their own right?

app/jobs/concerns/hyrax/permission_job_behavior.rb
```diff
@@ -1,8 +1,6 @@
 module Hyrax
   # Grants the user's edit access on the provided FileSet
-  class PermissionJob < ApplicationJob
-    queue_as Hyrax.config.ingest_queue_name
-
+  module PermissionJobBehavior
```
elrayle: same here RE: queue_as

app/jobs/concerns/hyrax/members_permission_job_behavior.rb
```diff
@@ -1,8 +1,6 @@
 module Hyrax
   # Grants read access for the supplied user for the members attached to a work
-  class MembersPermissionJob < ApplicationJob
-    queue_as Hyrax.config.ingest_queue_name
-
+  module MembersPermissionJobBehavior
```
elrayle: @bkeese I think we still need the queue_as here to have these jobs go into the ingest_queue. That or the statement needs to go back into each of the jobs.

bkeese: @elrayle Good point. I just assumed those statements were still in the jobs themselves and didn't look. Which place do you think makes more sense?

bkeese: I made a guess and decided to put it in the mixin.

elrayle:
@no-reply The final changes look good to me.  Do you have any other questions?

PR #4127, opened 2019-10-24
Only override logger for VirusScanner

cjcolvar:
This PR pulls in hydra-works master to include https://github.com/samvera/hydra-works/pull/372 in order to allow upgrading to `clamby` for ClamAV integration.  It then subclasses `Hydra::Works::VirusScanner` and removes overrides except for the logging since the rest is supplied in super.  The change to the generator is a guess at a default Clamby configuration based upon a suggestion in slack.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/virus_scanner.rb
+++ b/app/models/hyrax/virus_scanner.rb
@@ -14,39 +14,7 @@
 # Then set Hyrax::Works to use your scanner either in a config file or initializer:
 #   Hyrax::Works.default_system_virus_scanner = MyScanner
 module Hyrax
-  class VirusScanner
-    attr_reader :file
-
-    # @api public
-    # @param file [String]
-    def self.infected?(file)
-      new(file).infected?
-    end
-
-    def initialize(file)
-      @file = file
-    end
-
-    # Override this method to use your own virus checking software
-    # @return [Boolean]
-    def infected?
-      defined?(ClamAV) ? clam_av_scanner : null_scanner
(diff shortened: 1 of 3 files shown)


app/models/hyrax/virus_scanner.rb
```diff
@@ -14,39 +14,7 @@
...
-    end
-
-    # Always return zero if there's nothing available to check for viruses. This means that
-    # we assume all files have no viruses because we can't conclusively say if they have or not.
-    def null_scanner
-      warning "Unable to check #{file} for viruses because no virus scanner is defined"
-      false
-    end
-
+  class VirusScanner < Hydra::Works::VirusScanner
```
no-reply: :+1:

no-reply:
I think we originally inlined the VirusScanner out of a desire to remove the dependency on `Hydra::Works`. 

Does this code maybe have a better home than that gem. It doesn't seem to depend on anything in the AF, Works, PCDM stack.

Is anyone except Hyrax using it?

cjcolvar:
@no-reply I think those are good questions.  Could those be discussed in a separate issue?  That way this PR won't get held up when it allows Hyrax users to use the latest ClamAV and not be forced to use an old version or override in their local app.

cjcolvar:
@no-reply Would you like me to inline the whole scanner again to avoid the dependency (and the need to wait until a release of `hydra-works`)?

PR #4128, opened 2019-10-24
[WIP] Only do 4x parallelism when the build is from a fork

cjcolvar:
When a github user who isn't a member of the samvera github org forks hyrax, the circleci build will fail because it attempts to run with 10x parallelism when a user's default circleci plan only allows 4x.
(https://github.com/samvera/hyrax/pull/4110#issuecomment-545267345).  This PR creates duplicate test jobs for each workflow with different parallelism config based upon if it is a fork or not.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,11 +91,14 @@ jobs:
       hyrax_valkyrie:
         type: string
         default: "false"
+      parallelism:
+        type: integer
+        default: 10
     executor:
       name: 'samvera/ruby_fcrepo_solr_redis'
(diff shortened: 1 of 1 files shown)


no-reply:
What's the remaining work here?

I'd be happy to help get this to done, but don't know where we stand.

cjcolvar:
@no-reply I think the problem is that I haven't been able to determine a way to identify a fork.  Some example I found said to check the branch name which is what I tried:
```
# Forked pull requests have CIRCLE_BRANCH set to pull/XXX
              ignore: /pull\/[0-9]+/

PR #4129, opened 2019-10-25
Turn off Batch Upload features by default

no-reply:
The batch upload features that ship by default aren't very useful, and may be
error prone. Turning them off by default is a first step to removing them and
replacing with a different batch implementation.

Several alternative implementations are present in the community at the
moment. Hopefully we can bring one of these into Hyrax during the 3.0 series.

cc: @mark-dce @orangewold

@samvera/hyrax-code-reviewers


no-reply:
This needs an addition to Release Notes ASAP upon approval.

orangewolf:
Release notes added to the 3.0 draft

PR #4132, opened 2019-10-26
Replace custom Hyrax callbacks with `dry-events`

no-reply:
Hyrax has had its own internal events API since Hyrax late 2015. The approach
taken has a few drawbacks:

  - only one subscriber (as a block) to each event;
  - the subscriber registry, though mutable, isn't threadsafe.

We'll want to change the individual subscribers as we move toward Valkyrie. We
may also want to add new event types. Before continuing to develop against this
internal API, we should replace it with the externally maintained and more
robust `Dry::Events` tooling.

Since the existing callbacks make use of dependency inversion (i.e. adopters may
have custom listeners/callbacks defined) we can't stop calling their hooks. As a
first step, we implement the `Dry::Events` replacement and call them from the
callbacks.

We name the application events publisher `Hyrax::Publisher` to avoid confusion
with `Hyrax::Event`. Possibly more naming consideration is called for here.

An explicit dependency is added for `dry-events`, since this is the first time
we depend on it directly.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/listeners/batch_event_listener.rb
+++ b/app/services/hyrax/listeners/batch_event_listener.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Listeners
+    ##
+    # Listens for events related to batch activity
+    class BatchEventListener
+      ##
+      # @param event [Dry::Event]
+      def on_batch_created(event)
+        case event[:result]
+        when :success
+          Hyrax::BatchCreateSuccessService
+            .new(event[:user])
+            .call
+        when :failure
+          Hyrax::BatchCreateFailureService
+            .new(event[:user], event[:messages])
+            .call
+        end
+      end
+    end
+  end
+end
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -55,13 +55,13 @@ module Hyrax
   def self.primary_work_type
     config.curation_concerns.first
   end
-   
+
   ##
   # @return [Valkyrie::IndexingAdapter]
   def self.index_adapter
     config.index_adapter
   end
-  
+
   ##
   # @return [Dry::Events::Publisher]
   def self.publisher
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -57,6 +57,12 @@ module Hyrax
   end
 
   ##
+  # @return [Dry::Events::Publisher]
+  def self.publisher
+    config.publisher
+  end
+
+  ##
   # The Valkyrie persister used for PCDM models throughout Hyrax
   #
   # @note always use this method to retrieve the persister when data
--- a/lib/hyrax/publisher.rb
+++ b/lib/hyrax/publisher.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # This is an application-wide publisher for Hyrax's Pub/Sub interface.
+  #
+  # Hyrax publishes events on a variety of streams. The streams are namespaced
+  # using `dry-rb`'s dot notation idiom to help with organization. Namespaces
+  # reflect the kinds of resources the event applied to.
+  #
+  #   - `batch`: events related to the performance of `BatchCreateJob`
+  #   - `file_set`: events related to the lifecycle of Hydra Works FileSets
+  #   - `object`: events related to the lifecycle of all PCDM Objects
+  #
+  # Applications *SHOULD* publish events whenever the relevant actions are
+  # performed. While Hyrax provides certain out-of-the-box listeners to power
+  # (e.g.) notifications, event streams are useful for much more: implementing
+  # local logging or instrumentation, adding application-specific callback-like
+  # handlers, etc... Ensuring events are consistently published is key to their
+  # usefulness.
+  #
+  # @note this API replaces an older `Hyrax::Callbacks` interface, with added
+  #   thread safety and capacity for many listeners on a single publication
+  #   stream.
+  #
+  # @note we call this "Publisher" to differentiate from the `Hyrax::Event`
+  #   model. This class is a `Dry::Events` publisher.
+  #
+  # @todo audit Hyrax code (and dependencies!) for places where events should be
+  #   published, but are not.
+  #
+  # @example publishing an event
+  #   publisher = Hyrax::Publisher.instance
+  #
+  #   publisher.publish('object.deposited', object: deposited_object, user: depositing_user)
+  #
+  # @example use Hyrax.publisher
+  #   Hyrax.publisher.publish('object.deposited', object: deposited_object, user: depositing_user)
+  #
+  # @example subscribing to an event type/stream with a block handler
+  #   publisher = Hyrax::Publisher.instance
+  #
+  #   publisher.subscribe('object.deposited') do |event |
+  #     do_something(event[:object])
+  #   end
+  #
+  # @see https://dry-rb.org/gems/dry-events/0.2/
+  # @see Dry::Events::Publisher
+  class Publisher
+    include Singleton
+    include Dry::Events::Publisher[:hyrax]
+
+    register_event('batch.created')
+    register_event('file.set.audited')
+    register_event('file.set.attached')
+    register_event('file.set.url.imported')
+    register_event('file.set.restored')
+    register_event('object.deleted')
+    register_event('object.deposited')
+    register_event('object.metadata.updated')
+  end
+end
--- a/app/services/hyrax/listeners/batch_notification_listener.rb
+++ b/app/services/hyrax/listeners/batch_notification_listener.rb
@@ -3,9 +3,11 @@
 module Hyrax
   module Listeners
     ##
-    # Listens for events related to batch activity
-    class BatchEventListener
+    # Listens for events related to batch activity and creates notifications
+    class BatchNotificationListener
       ##
+      # Notify requesting users of batch success/failure
+      #
       # @param event [Dry::Event]
       def on_batch_created(event)
         case event[:result]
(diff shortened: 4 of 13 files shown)


no-reply:
I've proposed discussing this at the Samvera Tech call on 2019-10-30 https://wiki.duraspace.org/display/samvera/Samvera+Tech+Call+2019-10-30

no-reply:
> What's the use case for Events over modifying the actor stack or adding to the actor stack?

This is a fairly good point. The existing infrastructure could be in-lined into the callers without much harm. The main feature they add is the hook to override.

The use case for events is mainly to have many-to-many publisher/listener relationships. The idea is that the listeners should be configurable sets of things that should always happen whenever, e.g. an object is created, regardless of the subsystem that created it. They are currently used mainly to populate `Hyrax::Event`, triggering notifications, etc... 

There *are* a few existing callbacks with multiple callers, so inlining would deny them the abstraction. But also there are probably a lot of existing behaviors that meet the above requirements, but don't use the existing callbacks. Depending on how you look at it, that calls for either rolling back the `Hyrax::Callbacks` system and going about our day, or normalizing the handling throughout the code base.

In part, I'm attracted to the latter because it provides a powerful tool for applications. But I could definitely be talked down.

no-reply:
> Is there a need to instantiate the publisher to use it if it includes Singleton?

The class-level subscription methods don't support the style of listeners we're using here. Having specific listener classes has the advantage of making it easy to unsubscribe (and arguably code organization benefits).

Class-level publishing isn't a thing, as far as I can tell.

lib/hyrax/publisher.rb
```diff
@@ -0,0 +1,59 @@
...
+  #   stream.
+  #
+  # @note we call this "Publisher" to differentiate from the `Hyrax::Event`
+  #   model. This class is a `Dry::Events` publisher.
+  #
+  # @todo audit Hyrax code (and dependencies!) for places where events should be
+  #   published, but are not.
+  #
+  # @example publishing an event
+  #   publisher = Hyrax::Publisher.new
```
no-reply: need to update this documentation. it's out of date wrt Hyrax::Publisher being a singleton.

no-reply:
@hweng: at this point, this only aims to replace the `Hyrax::Callbacks` functions. `Hyrax::Callbacks` provides a registry callback actions for a set of methods that need to be manually triggered.

This wouldn't replace more traditional Rails/ActiveModel style callbacks in `CollectionBehavior`, etc... just the current set of methods that already rely on manual publishing.

The next step here, after appropriate deprecation, would be to replace existing manual calls to the trigger methods, e.g. `after_create_concern`, with explicit `Hyrax.publisher.publish` calls.

> Can we parse a method reference(symbol) or call back objects in the callback?

Can you say more about what you mean here?

hweng:
@no-reply Thanks for clarification.

>Can you say more about what you mean here?

For instance, when you add a callbacks `after_destroy` in your class, it can take a parameter of a method reference(symbol)  such as:

`after_destroy :destroy_permission_template`

Or call back objects:

`after_destroy SomeCallBack.new`

no-reply:
> @no-reply Thanks for clarification.
> 
> > Can you say more about what you mean here?
> 
> For instance, when you add a callbacks `after_destroy` in your class, it can take a parameter of a method reference(symbol) such as:
> 
> `after_destroy :destroy_permission_template`
> 
> Or call back objects:
> 
> `after_destroy SomeCallBack.new`

Ah, thanks. I'm following now.

The `Hyrax::Callbacks` approach for this is/was:

```ruby
Hyrax.config.callback.set(:my_callback_name) do |my_args|
  # callback handler code here
end
```

---

The proposed solution here affords a couple of approaches:

At the class level, block handlers can be defined:

```ruby
Hyrax::Publisher.subscribe(:my_event_type) do |event|
  # insert handler code here
end
```

At the publisher instance level, block handlers can also be defined:

```ruby
publisher = Hyrax::Publisher.new

publisher.subscribe(:my_event_type) do |event|
  # insert handler code here
end
```

And lastly, custom listeners can be setup and subscribed/unsubscribed to publisher instances (something like your `SomeCallBack.new` example):

```ruby
class MyListener
  def on_my_event_type(event)
    # insert handler code here
  end
end

publisher = Hyrax::Publisher.new
listener    = MyListener.new

publisher.subscribe(listener) # publisher.unsubscribe(listener)
```

The argument for all of these approaches is always a `Dry::Event`, which has some event metadata and a `payload` `Hash`; you can see the payload in use in the listener code here, e.g. https://github.com/samvera/hyrax/pull/4132/files#diff-02253bc5f9810688b4113ead8189ea5bR27-R28

hweng:
@no-reply Thank you for the explanation. Do you mean it can use the following pattern, for instance:

```
publisher.subscribe(:after_destroy) do |event|
  insert destroy_permission_template code
end
```

hweng:
👍

app/services/hyrax/listeners/batch_event_listener.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Listeners
+    ##
+    # Listens for events related to batch activity
+    class BatchEventListener
```
no-reply: Add unit tests for Listeners.

app/services/hyrax/listeners/batch_event_listener.rb
```diff
@@ -0,0 +1,24 @@
...
+      ##
+      # @param event [Dry::Event]
+      def on_batch_created(event)
+        case event[:result]
+        when :success
```
no-reply: Should this use the `dry-monads` `Result` (`Success`|`Failure`)?

lib/hyrax/publisher.rb
```diff
@@ -0,0 +1,62 @@
...
+  # @example subscribing to an event type/stream with a block handler
+  #   publisher = Hyrax::Publisher.instance
+  #
+  #   publisher.subscribe('object.deposited') do |event |
+  #     do_something(event[:object])
+  #   end
+  #
+  # @see https://dry-rb.org/gems/dry-events/0.2/
+  # @see Dry::Events::Publisher
+  class Publisher
```
no-reply: I considered adding unit tests for this, but just don't see the point.

None of the behavior is ours, except the names of the registered events. Everything is tested through the callers.

no-reply:
I'm wanting slightly to block this on more feedback and unit tests for Listeners, but can't actually do that myself.

If anyone has reservations about going down this road, I'd like them to put in requested changes (even if the changes requested are vague and open ended) on this for the time being.

hweng:
@no-reply I have more questions about the workflow, but it's fine if you think it should be answered in the next step(separate ticket):

For instance:
>```
>publisher.subscribe(:after_destroy) do |event|
>  insert destroy_permission_template code
>end
>```

Where do you insert the above code?  Is that inserted in the Transactions::Steps::DestroyWork as a step after call method: https://github.com/samvera/hyrax/blob/master/lib/hyrax/transactions/steps/destroy_work.rb

If it is, What if it need to parse in different code instead of destroy_permission_template code?

no-reply:
@hweng i think the best place for code like this is in an initializer, e.g. https://github.com/samvera/hyrax/pull/4132/files#diff-fa645789498fcdfac6345237199e7b14R1-R5

If we move forward with the transactions/do-notation approach as an actor stack replacement, we would probably publish messages as part of the transaction steps. For now, i think we just need to get to a place where we can replace the existing `Hyrax.config.callbacks` calls in the actor stack and elsewhere.

no-reply:
Thinking more about the use cases for this:

I've been [working on indexers](https://github.com/samvera/hyrax/pull/4141) for Valkyrie and thinking that the approach I'd like to take to pushing data to the index is to use a listener (or a series of them) to handle events that need to be indexed:

```ruby
module Hyrax
  module Listeners
    class IndexingListener
       def on_object_deposited(event)
         Hyrax.index_adapter.save(resource: event[:object])
       end

       def on_batch_succeeded(event)
         Hyrax.index_adapter.save_all(resources: event[:objects])
       end

       def on_other_index_triggered(event)
         Hyrax.index_adapter.save(resource: event[:object])
       end
   end
end
```

no-reply:
i think this is in a place where i'd like to move forward with it. do we have any reviewers?

cc: @samvera/hyrax-code-reviewers

lib/hyrax.rb
```diff
@@ -55,12 +55,18 @@ def self.logger
   def self.primary_work_type
     config.curation_concerns.first
   end
-
+
```
no-reply:
```suggestion
 
```

lib/hyrax.rb
```diff
@@ -55,12 +55,18 @@ def self.logger
...
```
no-reply:
```suggestion

```

no-reply:
@hweng i think i've fixed up the PR for merge. Can you give this one last review?

hweng:
@no-reply couldn't merge because the ci/circleci ruby2-6-2 failed. I triggered to rerun the job to see if it can pass this time.

hweng:
@no-reply still got it failed on ruby2-6-2:
Failed examples:
rspec './spec/actors/hyrax/actors/file_set_actor_spec.rb[1:1:4:1]' # Hyrax::Actors::FileSetActor when use_valkyrie is false #create_content when from_url is true calls ingest_file and kicks off jobs

no-reply:
@hweng the 2.6.2 build seems like it succeeded to me.

as far as i can tell, failures on this build have been unrelated to the change.

no-reply:
@orangewolf 

> This seems good to me, however. Are we removing existing code in a separate PR?

yes. the deprecation process here is challenging because of the inverted dependencies. the existing code needs to not just hang around, but *be called* by hyrax until downstream apps change. Definitely planning to remove all the existing code for the 4.0.0 release

PR #4135, opened 2019-10-30
Return empty string if APAFormatter authors_list is empty

geekscruff:
The citations link will throw an error if no authors are present. The error is in the APAFormatter where there is an assumption that there will be at least one author. The ChicagoFormatter already handles this by returning an empty string for an empty authors_list. This PR applies the same change to the APAFormatter.

In a Hyrax application where creator is required, this will never happen. But if a Hyrax application is customised to remove creator from the required files, this error could then arise.

* Remove creator from required fields
* Add work without a creator
* The citations render

@samvera/hyrax-code-reviewers


geekscruff:
@cjcolvar I did that, and it passes, and the failure is something else :(

Issue #3972, opened 2019-09-12
Display bugs on Work form >  Sharing tab


dlim87:
### Descriptive summary
There are 3 display bugs on the New/edit work page, sharing tab: 
- In the alert box, raw HTML and ASCII codes are not being translated
- In the "add groups" section, the button has raw HTML
- On the access list, the current user has a (translation?) display bug

![Image 2019-09-12 at 9 21 43 AM](https://user-images.githubusercontent.com/5492162/64802119-ca494980-d53e-11e9-902b-59aea65bafcd.png)

### Rationale

HTML should be properly rendered

### Expected behavior
Html should be properly rendered
### Actual behavior
HTML tags are showing up
### Steps to reproduce the behavior

1. Log in
2. Click on share your work
3. Make sure it's in an admin set that allows sharing (the default admin set should work)
3. Click on the sharing tab

### Related work
Another HTML error: https://github.com/samvera/hyrax/issues/3959
The original work (and thus files) for this view: https://github.com/samvera/hyrax/pull/2237

Tested on Mac and windows - Chrome 76.0.3809.132,
Mac - Safari 12.1.2,
Windows - Firefox 69.0

<details><summary>hyrax info</summary>
Hyrax Version: 3.0.0.pre.rc1
Hyrax Branch: master
Hyrax Revision: 48ed925940130bcc5e75092c974555f4f1654390

Nurax Version
ENV: production
SHA: 40b47b0d71af1b958e016ebe49212c1740683133
Branch: 2019-09-12-10-00
Updated: 12 Sep 2019 15:00:37</details>

dlim87:
This is rendering properly
![Image 2019-11-06 at 9 57 12 AM](https://user-images.githubusercontent.com/5492162/68324313-fb945080-007b-11ea-9b9f-6d84442df858.png)

PR #4138, opened 2019-10-30
Fix for HYRAX-#3972

fritzfreiheit:
Fix for HYRAX-#3972

The edit works share tab displays raw html at several locations.

Expected behavior:
The edit works share tab does not display raw html.

* Modify form erb to use "_html" from I18n.t values
* Change hyrax.*.yml entries to use "_html" for form values

* Log in as Administrator
* Edit a work
* Click on "Sharing" tab
* Under header "Currently Shared With":
** should display "Depositor(<some email>)"
* Under header "Add Sharing" add a group permission:
** by selecting from "Group" and "Choose Access" pull downs, then "Add"
** The message 'Permissions are not saved until the "Save" button is
   pressed at the bottom of the page.' appears and does not contain raw html.

@samvera/hyrax-code-reviewers


spec/views/hyrax/base/_form_share.erb_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  it "renders the permissions save note" do
+    expect(page).to have_selector('div#save_perm_note', visible: false)
+    expect(rendered_save).to include I18n.t("hyrax.base.form_share.permissions_save_note_html")
+  end
+  it "renders the add this group" do
+    expect(rendered_save).to include I18n.t("hyrax.base.form_share.add_this_group_html")
+  end
+end
```
no-reply:
```suggestion
end

```

PR #4141, opened 2019-11-01
Use a factory method (`.for`) to get valkyrie indexers

no-reply:
We'll want to get different indexers for different kinds of resources, so we can
prepare them for indexing according to a customized schema. Switching to a
factory method instead of an initializer allows us to create a new indexer
instance based on the properties of the resource.

In Hyrax, we'll likely build from the factory depending on class, but
applications can easily override to build depending on schema/attributes,
present values, etc..., by providing a different `ValkyrieIndexer` to their
`IndexingAdapter`, with a custom `.for` implementation.

Also adds documentation for the method composition style for indexing
This style is mainly borrowed from Hydra/Samvera history. `#to_solr` composition
has always been used for indexers, from the time when an on-model indexer
`MyBook#to_solr` was the whole index builder interface.

Hyrax continues to use a similar style for is
`IndexingService#generate_solr_document` methods; always `super.tap`. The
pitfalls of putting `#to_solr` directly on the model (namely: breaking
encapsulation by inspecting deep model internals), were largely avoided by this,
but we still had pretty complex indexing behavior (giving rise to comments like
"This class gets called by
ActiveFedora::IndexingService#solrize_rdf_assertions").

The main difference in our approach here is to keep the composition chain
clean: call `super`, transform its output, return the result. Using this as _the
one pattern_ for Valkyrie-based indexers and documenting it clearly up-front
should help keep this behavior understandable.

Connected to #3800.

@samvera/hyrax-code-reviewers


app/indexers/hyrax/valkyrie_indexer.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    ##
+    # @api private
+    def self.registry
+      @registry ||= {}
+    end
+
+    ##
+    # @api public
```
no-reply:
```suggestion
    # @api private
```

per @jeremyf's suggestion

PR #4143, opened 2019-11-01
Put an upper limit on `browse_everything` support

no-reply:
We support BrowseEverything 0.16+ and 1.x; but can't support 2.x yet (it doesn't
exist and will likely have extensive breaking changes).

@samvera/hyrax-code-reviewers


no-reply:
cc: @jrochkind

PR #4146, opened 2019-11-05
Make `Skip to content` link focus on main content

### LINKED ISSUE #3774: Skip to Content link doesn't always set keyboard focus
### Descriptive summary

In Hyrax 2.5.0 and Firefox, I can press the tab key and keyboard focus goes to the "Skip to Content" link.  I can press enter, and the visual focus of the screen does move down to the content of the page.  However, when I press tab again, keyboard focus goes right back to the "Skip to Content" link.

### Rationale

Currently, users who navigate hyrax apps using a keyboard aren't reliably well-served by the "Skip to Content" link.  Depending on the browser they use, it could skip to content for visual purposes, but not for navigation purposes.

### Expected behavior

I expect that the "Skip to Content" link also sets keyboard focus to the content, regardless of browser.

### Actual behavior

The "Skip to Content" link does not set keyboard focus to the content in Firefox.  It does in Chrome.

Go to Firefox, and press Tab - Enter - Tab.

### Related info

An interesting article here, including mobile considerations: https://axesslab.com/skip-links/

Another interesting article: https://www.bignerdranch.com/blog/web-accessibility-skip-navigation-links/

Dananji:
Fixes #3774 

When a user use tab key on `Skip to content` link in Firefox, it does not shift the keyboard to focus to main content, even though the view is focused. The user has to go through all the navigation links to reach the main content.

* Use JavaScript on the `click` event of `Skip to content` link to change keyboard focus to the main content

* Open Hyrax in Firefox
* Press `Tab` key and then `Enter`
* Press `Tab` again to see the focus is skipped to main content

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/layouts/hyrax.html.erb
+++ b/app/views/layouts/hyrax.html.erb
@@ -25,7 +25,6 @@
 
       <a name="skip-to-content" id="skip-to-content"></a>
       <%= content_for?(:content) ? yield(:content) : yield %>
-  
     </div><!-- /#content-wrapper -->
     <%= render 'shared/footer' %>
     <%= render 'shared/ajax_modal' %>
--- a/app/views/layouts/hyrax.html.erb
+++ b/app/views/layouts/hyrax.html.erb
@@ -23,11 +23,26 @@
         </div>
       <% end %>
 
-        <a name="skip-to-content" id="skip-to-content"></a>
-        <%= content_for?(:content) ? yield(:content) : yield %>
-
+      <a name="skip-to-content" id="skip-to-content"></a>
+      <%= content_for?(:content) ? yield(:content) : yield %>
+  
     </div><!-- /#content-wrapper -->
     <%= render 'shared/footer' %>
     <%= render 'shared/ajax_modal' %>
   </body>
+  <script type="text/javascript">
+    $( document ).ready(function() {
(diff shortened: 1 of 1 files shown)


app/views/layouts/hyrax.html.erb
```diff
@@ -23,11 +23,26 @@
         </div>
       <% end %>
 
-        <a name="skip-to-content" id="skip-to-content"></a>
-        <%= content_for?(:content) ? yield(:content) : yield %>
-
+      <a name="skip-to-content" id="skip-to-content"></a>
+      <%= content_for?(:content) ? yield(:content) : yield %>
+
```
no-reply:
```suggestion

```

dlim87:
I'm not 100% sure what the intended behavior for this ticket is, but it is behaving in a way that seems to be working in most cases:

On the home page: it takes you to right before "share your work", skipping the navbar
on a works page: it takes you to right before "add to collection", skipping the thumbnails to be right after the work's title
on a collection page: it takes you to right before the "search subcollections and works" search bar, right after the title
On a search page, it takes you to right before the pagination controls at the top of the page, skipping the sidebar

However, I'm not sure it's really working in the dashboard, since it looks like for the pages there, it takes you out of the page, since the next tab is interacting with my bookmarks toolbar

PR #4148, opened 2019-11-06
Make the valkyrie index adapter configurable

no-reply:
Add configuration for the indexing adapter to make access consistent with other
valkyrie adapters. The one-true-way to access the indexing adapter is now
`Hyrax.indexing_adapter`.

We'll likely only really support the Solr indexing adapter (or things that very
closely conform to its behavior) for the forseeable future.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/solr_service.rb
+++ b/app/services/hyrax/solr_service.rb
@@ -142,7 +142,7 @@ module Hyrax
     private
 
       def valkyrie_index
-        Valkyrie::IndexingAdapter.find(:solr_index)
+        Hyrax.index_adapter
       end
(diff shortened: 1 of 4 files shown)


VivianChu:
@no-reply - LGTM 👍   However coverage/coveralls failed because coverage decreased 6.9%.  Do you need to add test to make that increase the coverage?

no-reply:
@VivianChu coveralls routinely reports coverage varying by this much. it's unrelated to this PR.

we made coveralls a non-required check for this reason, and have plans to remove it entirely.

PR #4157, opened 2019-11-08
Deprecate `Hyrax.config.callback`

### LINKED ISSUE #4154: Deprecate `Hyrax::Callbacks` setting behavior
Connected to #4132

no-reply:
Emit deprecation warnings on `#set` and `#run`.

A `warn:` flag is added, defaulting to true. Internal Hyrax callers pass it
`false` to avoid emitting warnings. These callers will still need to register
and invoke callbacks until downstream applications are using `Hyrax.publisher`.

Closes #4154 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -46,7 +46,7 @@ module Hyrax
       private
 
         def run_callbacks(hook, env)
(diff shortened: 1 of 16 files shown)


no-reply:
sorry @cjcolvar, i think i've got it this time.

Issue #3385, opened 2018-11-07
direct links to works, filesets, and analytics pages have no breadcrumbs


vantuyls:
### Descriptive summary

if one drops a link directly to a work showpage, fileset showpage, or analytics showpage, there are no breadcrumbs, which make it impossible to contextualize and navigate the relationships among these views. Providing breadcrumbs or some other navigation will help users navigate. 

### Expected behavior
some kind of navaigation capability from these pages to their related pages (eg. the work for which analytics are being displayed, the work the fileset is part of)

### Actual behavior
from a direct link, there is no navigation interface

### Steps to reproduce the behavior

1. go to a work
2. go to the analytics page for the work
3. see breadcrumbs
4. go to a work
4b. go to the work's analytics page
5. copy the URL for the analytics page
6. open a new browser tab
7. paste in URL
8. see no breadcrumbs

adamjarling:
@vantuyls This has to do with the `breadcrumbs_for_...` builders in the controllers I think.  If someone is more familiar with that part of the app or Ruby, might be better suited for it than I.

no-reply:
I think this is by design? Since this view (like others) is accessible from various contexts, the breadcrumbs should provide the user a trail back through the path they came from. If they dropped a link in, which set of breadcrumbs should we display?

I'm honestly a bit confused about how the `dashboard` breadcrumb finds its way in when I find a work from the catalog.

blancoj:
@no-reply  we actually flagged this as an issue in our local instance and made the change.  I will put in a PR with our change, it is pretty straight forward.

dlim87:
It is now retaining the breadcrumbs (except for "return to search results", which is expected to be dropped) when the URL is c/p from the work show page, file show page, or analytics page. 

However, there was some weird behavior with the breadcrumbs. Like was noted earlier in this ticket, it adds in "Dashboard" to the breadcrumbs, even when you're navigating from elsewhere.

fritzfreiheit:
I've tested this on my local host and it appears to be behaving correctly.

dlim87:
I'm not really sure what the movement in this ticket was for, it doesn't appear like any action was taken? It still has the same behaviors as a couple months ago: the breadcrumbs _are_ displaying even when following a direct work.

It just says Home / work-name if you're a guest or Home / Dashboard / Works / work-name if you're logged in.

no-reply:
[...]

PR #4158, opened 2019-11-08
Fix for HYRAX-#3385

fritzfreiheit:
Modified breadcrumbs to check referer and skip unless it is this website

Expected behavior:

For works, file sets, and collections: when navigating from within the website,
display breadcrumbs. When the referer is blank or some other website, then
don't display breadcrumbs.

In order to test:

* Go to the main page.
* Browse to a work or collection.
* Breadcrumbs should be displayed.
* Browse to a file set from a work or to a work from a collection.
* Breadcrumbs should be displayed.
* Copy the url.
* Open a new browser tab.
* Paste the url into the browser tab.
* No breadcrumbs should be displayed.
* Repeat for work, collection, file set, etc.
* Repeat for dashboard.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/breadcrumbs.rb
+++ b/app/controllers/concerns/hyrax/breadcrumbs.rb
@@ -4,12 +4,18 @@ module Hyrax
     extend ActiveSupport::Concern
 
     def build_breadcrumbs
-      if request.referer
-        trail_from_referer
-      else
-        default_trail
-        add_breadcrumb_for_controller if user_signed_in?
-        add_breadcrumb_for_action
+      return if build_breadcrumbs_skip
+      trail_from_referer
+    end
+
+    def build_breadcrumbs_skip
+      return true if request.referer.blank?
+      referer_path = URI(request.referer).path
+      begin
+        Rails.application.routes.recognize_path(referer_path)
+        false
+      rescue ActionController::RoutingError
+        true
       end
     end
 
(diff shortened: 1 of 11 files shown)


no-reply:
cc: @samvera/hyrax-repo-managers 

comments on desired breadcrumb behaviors would be extremely welcome.

fritzfreiheit:
@no-reply / @jeremyf -- Please review and approve.

app/controllers/concerns/hyrax/breadcrumbs.rb
```diff
@@ -3,12 +3,18 @@ module Breadcrumbs
...
-        add_breadcrumb_for_action
+      return if build_breadcrumbs_skip
+      trail_from_referer
+    end
+
+    def build_breadcrumbs_skip
+      return true if request.referer.blank?
+      referer_path = URI(request.referer).path
+      begin
+        Rails.application.routes.recognize_path(referer_path)
```
jeremyf, line 14: To make sure I'm following the logic, if Rails recognizes the referer path, then we don't want to skip breadcrumbs.  Otherwise, we do want to skip them.

fritzfreiheit: Jeremy,
Right. The requirements are that breadcrumbs are displayed from internal references, but not from external references.
Thanks!
Fritz.

jeremyf:
I went ahead and rebased. Will merge once tests pass.

jeremyf:
I believe this requires #4401

jeremyf:
Ok, so this is stuck and needs further break-down of permissions.  https://github.com/samvera/hyrax/issues/4405

jeremyf:
@fritzfreiheit I'm rebasing this against master. Hopefully the specs run clean (after #4414)

jeremyf:
Gah, I misremembered. This is related to something else.

PR #4159, opened 2019-11-08
use :access_rights instead of :access_right

ebenenglish:
Changes `:access_right` property to `:access_rights` so property name more closely matches predicate and semantics.

* change `:access_right` metadata property to `:access_rights`

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/work_form.rb
+++ b/app/forms/hyrax/forms/work_form.rb
@@ -20,9 +20,9 @@ module Hyrax
                :member_ids, :alternative_title, to: :model
 
       self.terms = [:title, :alternative_title, :creator, :contributor, :description, :abstract,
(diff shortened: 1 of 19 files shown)


no-reply:
@ebenenglish this looks like it will need some resolution since the `alternative_title` PR got merged.

Looks good to me, at a glance. I'll save a more thorough review for after rebase.

ebenenglish:
@no-reply 
Hmm, looks like this might be a dead end.

This property causes `ActiveFedora::Attributes#define_active_triple_accessor` to throw an error, because the `protected_property_name?` check (from `ActiveTriples::Properties`) returns `true`.

There is already an `access_rights` property defined by `Hydra::AccessControls::WithAccessRight` from `hydra-head`.

Let me know if you want to pursue a different alternative to `access_right`? Open to suggestions.

PR #4175, opened 2019-12-26
Write Metadata and ACL changes to valkyrie index with listeners

no-reply:
when an object's metadata is updated, update the Valkyrie-driven index via the
configured indexing adapter.

Also: When an ACL change event is published, write to the currently configured
Hyarx (valkyrie) indexing adapter. This ensures that when ACLs are changed
independently of the object, the permissions cached in Solr will stay up to
date.

Hyrax relies on permissions to be cached correctly in the index for various
`Ability` behaviors. This is perhaps not ideal for a long-term valkyrie future;
`ActiveFedora` basically requires this of us, though, unless we're willing to
take the performance hit of looking up the ACLs from Fedora per-request.

Note that both of these are presently no-op by default, since we use a null indexing
adapter. Changes to setup this behavior are included in earlier commits.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/solr_service.rb
+++ b/app/services/hyrax/solr_service.rb
@@ -116,9 +116,15 @@ module Hyrax
     private
 
       ##
-      # @api private
+      # @private
+      # Return the valkyrie solr index.
+      #
+      # Since this module depends closely on RSolr internals and makes use
+      # of `#connection`, it will always need to connect to a Solr index. Other
+      # valkyrie indexers used here would, at minimum, need to provide a
+      # functioning `rsolr` connection.
       def valkyrie_index
-        Hyrax.index_adapter
+        Valkyrie::IndexingAdapter.find(:solr_index)
       end
 
       ##
--- a/app/services/hyrax/solr_service.rb
+++ b/app/services/hyrax/solr_service.rb
@@ -9,7 +9,12 @@ module Hyrax
   class SolrService
     COMMIT_PARAMS = { softCommit: true }.freeze
 
+    ##
+    # @!attribute [r] use_valkyrie
+    #   @private
     attr_reader :use_valkyrie
+
+    delegate :commit, to: :connection
 
     def initialize(use_valkyrie: Hyrax.config.query_index_from_valkyrie)
       @old_service = ActiveFedora::SolrService
@@ -42,12 +47,8 @@ module Hyrax
       solr_path = args.delete(:path) || Hyrax.config.solr_select_path
       args = args.merge(q: query) unless query.blank?
 
-      if use_valkyrie
-        valkyrie_index.connection.get(solr_path, params: args)
-      else
-        args = args.merge(qt: 'standard') unless query.blank?
-        self.class.instance.conn.get(solr_path, params: args)
-      end
+      args = args.merge(qt: 'standard') unless query.blank? || use_valkyrie
+      connection.get(solr_path, params: args)
     end
 
     # Wraps rsolr post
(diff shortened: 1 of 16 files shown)


app/services/hyrax/solr_service.rb
```diff
@@ -141,8 +115,23 @@ def search_by_id(id, opts = {})
...
+      # @private
+      # Return the valkyrie solr index.
+      #
+      # Since this module depends closely on RSolr internals and makes use
+      # of `#connection`, it will always need to connect to a Solr index. Other
+      # valkyrie indexers used here would, at minimum, need to provide a
+      # functioning `rsolr` connection.
       def valkyrie_index
-        Hyrax.index_adapter
+        Valkyrie::IndexingAdapter.find(:solr_index)
```
orangewolf: I do agree that on the whole this service (and everything that calls it) has to be refactored before a elastic search or something indexer can be implemented, but I disagree with making this a pure solr lock here. What about a null indexer? it would be useful to start down the path of making this generic, not locking it even tighter to have to do more refactoring in the future.

spec/support/clean_solr.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 RSpec.configure do |config|
   config.before(:each, clean_index: true) do
-    client = Hyrax.index_adapter.connection
+    client = Valkyrie::IndexingAdapter.find(:solr_index).connection
```
orangewolf: again I feel like we should leave it open indexer wise.

orangewolf: you can definitely make the case for a tag ability to use the null indexer

app/services/hyrax/solr_service.rb
no-reply: the issue i'm trying to avoid here is that `Hyrax.index_adapter` *can* vary (e.g. it may be a null adapter), but this one should always return an `rsolr` connection.

the pattern we've landed on for `Valkyrie::IndexingAdapter` is a ["write only adapter"](https://github.com/samvera/valkyrie/issues/598), and i guess the real issue here is that there's no corresponding pattern for the read side, other than "just use RSolr".

is the alternative to just leave this as `Hyrax.index_adapter`, accepting that this will raise errors when `use_valkyrie` is on and a non-RSolr powered indexer is configured?

spec/support/clean_solr.rb
no-reply:
```suggestion
    Hyrax.index_adapter.wipe!
```

spec/support/clean_solr.rb
```diff
@@ -1,7 +1,7 @@
...
     client.delete_by_query("*:*", params: { softCommit: true })
```
no-reply:
```suggestion
```

spec/support/clean_solr.rb
no-reply: how about https://github.com/samvera/hyrax/pull/4175/files#r362945930
??

orangewolf: I dig it

app/services/hyrax/solr_service.rb
orangewolf: Ah, I see, I retract to comment then. I didn't fully understand the write only idea.

no-reply:
> In general I think the case for a better system for permissions that doesn't require frequent large scale reindexing is a good idea

agreed. i think we're more or less stuck with indexing permissions as long as we're using `Blacklight::AccessControls` and, maybe more significantly, as long as our Fedora support involves storing ACLs as Fedora resources.

one design consideration for the new Valkyrie-powered `Hyrax::AccessControlList` was that we could choose to store ACLs using a separate adapter from the one used for digital object metadata. if we get there, we may be able to stop relying on the index to hold this data. that seems desirable to me, in part because it reduces the risk of inappropriate access because of, e.g. an indexing error.

in either case, i think i'd keep the `:object_acl_updated` event and we could unsubscribe the index listener.

no-reply:
ughhh b369557 breaks it. reverting

PR #4178, opened 2020-01-03
Remove call to dead ActiveFedora code

no-reply:
This conditional was needed to support ActiveFedora 11. We have since dropped
support, so the check is no longer needed.

@samvera/hyrax-code-reviewers


jeremyf:
@no-reply what is the procedure for a drop in coverage? The build says its broken.

no-reply:
@jeremyf coveralls has been chronically unreliable on this project. There's a known TODO item to remove it.

It is already turned off as a required build on Github's settings, which unblocks merge, but doesn't prevent the :x: from showing. :(

PR #4179, opened 2020-01-03
Add tests for `WorksControllerBehavior#show`, `#create`, & `#new`

no-reply:
Adds module level tests for `WorksControllerBehavior#show` using a
controller registered for a Valkyrie model. A minor changes to `GraphExporter`
is needed to make some of these tests pass (the changed branch is otherwise
untested).

The outline of tests for `#create` and `#new` are also put in place, but these
require more work (on Actors and Forms, respectively) to pass. Accordingly, they
are left as pending tests.

This is a step toward full `WorksController` support for Valkyrie.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -231,19 +231,5 @@ module Hyrax
                .where('sipity_workflow_roles.role_id' => approving_role.id).any?
         end
       end
-
-      ##
-      # @api private
-      #
-      # Overwrite extract subjects to map permissions
-      #
-      # @note this shims in support for using existing abilities when passing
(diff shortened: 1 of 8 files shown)


app/models/concerns/hyrax/ability.rb
```diff
@@ -231,19 +231,5 @@ def can_review_submissions?
                .where('sipity_workflow_roles.role_id' => approving_role.id).any?
         end
       end
-
-      ##
```
no-reply: this turns out to hurt more than help. We'll want to just implement the correct permissions as we go, and this obfuscates which ones are missing.

PR #4181, opened 2020-01-10
Add a `ChildTypes` to manage selection display during work attachment

no-reply:
With `ActiveFedora` models, we use `MyWork.valid_child_concerns` to determine
which types the user can select to attach to an existing work.

Since this is a display-only issue, we break this out into a new service. For
legacy models, we delegate the configuration back to the model, avoiding any
need for configuration changes.

For Valkyrie models, we provide a default config (a work can nest works of its
own type), but punt on configuration issues. There are other configuration
matters to discuss.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/child_types.rb
+++ b/app/services/hyrax/child_types.rb
@@ -23,6 +23,7 @@ module Hyrax
 
     ##
     # @params [Class] parent
+    # @return [Enumerable<Class>] a list of classes that are valid as child types for `parent`
     def self.for(parent:)
       return new(parent.valid_child_concerns) if
         parent.respond_to?(:valid_child_concerns)
--- a/app/services/hyrax/child_types.rb
+++ b/app/services/hyrax/child_types.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # A list of child work types a user may choose to attach for a given work type
+  #
+  # These lists are used when users select among work types to attach to an
+  # existing work, e.g. from Actions available on the show view.
+  #
+  # @example
+  #   child_types = Hyrax::ChildTypes.for(parent: MyWorkType)
+  #
+  class ChildTypes
+    include Enumerable
+    extend Forwardable
+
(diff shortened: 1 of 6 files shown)


app/services/hyrax/child_types.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def_delegators :@types, :each
+
+    ##
+    # @!attribute [r] types
+    #   @return [Array<Class>]
+    attr_reader :types
+
+    ##
+    # @params [Class] parent
```
jeremyf: I believe this should have an expected return object type.

From a coverage stand-point, should we use:

```ruby
if parent.respond_to?(:valid_child_concerns)
  new(parent.valid_child_concerns)
else
  new([parent])
end
```

no-reply:
```suggestion
    # @params [Class] parent
    # @return [Enumerable<Class>] a list of classes that are valid as child types for `parent`
```

no-reply: i added the YARD for the return type.

i'm not sure i follow the coverage question. can you say more?

jeremyf: With my proposed code, we have three lines that can be covered the `if` the logic within the `if` and the logic in the `else`. In the committed code, the guard structure knows that the line `new(parent.valid_child_concerns) if parent.respond_to?(:valid_child_concerns)` evaluated, regardless if the parent responds to valid_child_concerns (or not). So our code coverage does not know if we called `new(parent.valid_child_concerns)` (which could be tested in specs, but I'd prefer to see the coverage check the two points of branching logic).

no-reply: we use guard clauses pretty extensively throughout the codebase, and haven't had reliable coverage metrics in over a year. with those factors in mind, i'd suggest we not worry too much about this detail.

jeremyf: With that context, keep the guard clause.

PR #4185, opened 2020-01-13
Remove Coveralls

no-reply:
Coveralls reliably reports nonsense from CircleCI, and it's just the worst.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -6,13 +6,8 @@ def ci_build?
 end
 
 require 'simplecov'
-require 'coveralls'
-Coveralls.wear!
 SimpleCov.root(File.expand_path('../..', __FILE__))
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-                                                                 Coveralls::SimpleCov::Formatter,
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -7,7 +7,6 @@ def ci_build?
 
 require 'simplecov'
 require 'coveralls'
-Coveralls.wear!
 SimpleCov.root(File.expand_path('../..', __FILE__))
 SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
                                                                  Coveralls::SimpleCov::Formatter,
```
rotated8: What is this formatter used for after Coveralls is removed?

no-reply: it seemed better to not change the simplecov formatter just to stop reporting failures.

no-reply: on the other hand, dropping `Coveralls.wear!` doesn't seem to have fixed the issue at all.

no-reply: rebuilding with coveralls dropped entirely.

PR #4187, opened 2020-01-16
Update work generator usage documentation

no-reply:
This documentation was *far* out of date, and barely listed half of what the
generator actually does in practice.

@samvera/hyrax-code-reviewers


no-reply:
it's the order they are output by the generator, except with the model itself (and its spec) moved to the top of the groups listing `app/*` and `spec/*`

PR #4188, opened 2020-01-16
Add a `hyrax:work_resource` generator for valkyrie native works

no-reply:
The new generator creates a valkyrie-driven `Hyrax::Work` subclass, with a
controller and a spec file. The specs include the Hyrax shared examples (and the
valkyrie shared examples) by default.

Since a custom search builder is needed to ensure Valkyrie works can be found by
their legacy `ActiveFedora` model names, we include a class builder for that,
and register the custom class with the generated `Controller`. This can be
removed when Wings is deleted; applications that are ready to move to
valkyrie-only index behavior will need to drop this search builder in favor of
one that uses the valkyrie index at a later date.

* in a new hyrax app (built from this branch) do: `rails g hyrax:work_resource Monograph`; when it's done, run the new spec

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/form_factory.rb
+++ b/app/services/hyrax/form_factory.rb
@@ -5,7 +5,7 @@ module Hyrax
   # A factory class for Hyrax forms.
   #
   # @note We use a factory class (rather than a factory method like `.for`) to
-  #   provide compatibility with the legacy `Hyrax::WorkFormService`.y
+  #   provide compatibility with the legacy `Hyrax::WorkFormService`.
   #
   # @example
   #   factory = Hyrax::FormFactory.new
--- a/app/services/hyrax/form_factory.rb
+++ b/app/services/hyrax/form_factory.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # A factory class for Hyrax forms.
+  #
+  # @note We use a factory class (rather than a factory method like `.for`) to
+  #   provide compatibility with the legacy `Hyrax::WorkFormService`.y
+  #
+  # @example
+  #   factory = Hyrax::FormFactory.new
+  #
+  #   form_change_set = factory.build(work)
+  #
+  # @since 3.0.0
+  class FormFactory
+    ##
+    # @param model      [Object]
+    # @param ability    [Hyrax::Ability]
+    # @param controller [ApplicationController]
+    def build(model, _ability, _controller)
+      Hyrax::ChangeSet.for(model)
+    end
+  end
+end
--- a/lib/generators/hyrax/work_resource/templates/work.html.erb_spec.rb.erb
+++ b/lib/generators/hyrax/work_resource/templates/work.html.erb_spec.rb.erb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+RSpec.describe '<%= plural_file_name %>/<%= file_name %>.html.erb', type: :view do
+
+end
--- a/lib/generators/hyrax/work_resource/templates/work.html.erb_spec.rb.erb
+++ b/lib/generators/hyrax/work_resource/templates/work.html.erb_spec.rb.erb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
-
+# Generated via
+#  `rails generate hyrax:work <%= class_name %>`
 RSpec.describe '<%= plural_file_name %>/<%= file_name %>.html.erb', type: :view do
 
 end
--- a/lib/generators/hyrax/work_resource/templates/work_spec.rb.erb
+++ b/lib/generators/hyrax/work_resource/templates/work_spec.rb.erb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# Generated via
+#  `rails generate hyrax:work <%= class_name %>`
+require 'rails_helper'
+require 'hyrax/specs/shared_specs/hydra_works'
+
+RSpec.describe <%= class_name %> do
+  subject(:work) { described_class.new }
+
+  it_behaves_like 'a Hyrax::Work'
+end
--- a/lib/generators/hyrax/work_resource/work_resource_generator.rb
+++ b/lib/generators/hyrax/work_resource/work_resource_generator.rb
@@ -1,0 +1,58 @@
+require 'rails/generators'
+require 'rails/generators/model_helpers'
+
+class Hyrax::WorkResourceGenerator < Rails::Generators::NamedBase
+  # ActiveSupport can interpret models as plural which causes
+  # counter-intuitive route paths. Pull in ModelHelpers from
+  # Rails which warns users about pluralization when generating
+  # new models or scaffolds.
+  include Rails::Generators::ModelHelpers
+
+  source_root File.expand_path('../templates', __FILE__)
+
+  argument :attributes, type: :array, default: [], banner: 'field:type field:type'
+
+  def banner
+    if revoking?
+      say_status("info", "DESTROYING VALKYRIE WORK MODEL: #{class_name}", :blue)
+    else
+      say_status("info", "GENERATING VALKYRIE WORK MODEL: #{class_name}", :blue)
+    end
+  end
+
+  def create_controller
+    template('controller.rb.erb', File.join('app/controllers/hyrax', class_path, "#{plural_file_name}_controller.rb"))
+  end
+
+  def create_model
+    template('work.rb.erb', File.join('app/models/', class_path, "#{file_name}.rb"))
+  end
+
+  def create_model_spec
+    template('work_spec.rb.erb', File.join('spec/models/', class_path, "#{file_name}_spec.rb")) if
+      rspec_installed?
+  end
+
+  def create_views
+    create_file File.join('app/views/hyrax', class_path, "#{plural_file_name}/_#{file_name}.html.erb") do
+      "<%# This is a search result view %>\n" \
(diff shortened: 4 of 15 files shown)


spec/services/hyrax/form_factory_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+RSpec.describe Hyrax::FormFactory do
+  subject(:factory) { described_class.new }
+  let(:model)       { FactoryBot.build(:hyrax_work) }
+
+  describe '.build' do
+    let(:ability)    { :FAKE_ABILITY }
+    let(:controller) { :FAKE_CONTROLLER }
+
+    it 'returns a change set' do
+      expect(factory.build(model, ability, controller)).to be_a Hyrax::ChangeSet
```
jeremyf: Would it be out of scope to have named parameters for the `#build` method?

app/services/hyrax/form_factory.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # A factory class for Hyrax forms.
+  #
+  # @note We use a factory class (rather than a factory method like `.for`) to
+  #   provide compatibility with the legacy `Hyrax::WorkFormService`.y
```
cjcolvar:
```suggestion
  #   provide compatibility with the legacy `Hyrax::WorkFormService`.
```

lib/generators/hyrax/work_resource/templates/work.html.erb_spec.rb.erb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
```
cjcolvar: Should we also add the `Generated via` comment that is in the other generated files here?

spec/services/hyrax/form_factory_spec.rb
jeremyf: Based on our conversation, @no-reply stated that we are conforming to the `#build` method from `Hyrax::WorkFormService`.

See

https://github.com/samvera/hyrax/blob/1b0afbd6749f35ea771f88da47cf4cf32a0a92fa/app/services/hyrax/work_form_service.rb#L3

lib/generators/hyrax/work_resource/work_resource_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+    create_file File.join('app/views/hyrax', class_path, "#{plural_file_name}/_#{file_name}.html.erb") do
+      "<%# This is a search result view %>\n" \
+      "<%= render 'catalog/document', document: #{file_name}, document_counter: #{file_name}_counter  %>\n"
+    end
+  end
+
+  def create_view_spec
+    return unless rspec_installed?
+    template('work.html.erb_spec.rb.erb',
+             File.join('spec/views/', class_path, "#{plural_file_name}/_#{file_name}.html.erb_spec.rb"))
```
cjcolvar: It makes for a really long line but would the guard clause version be good for consistency with `create_model_spec`?  I think rubocop prefers the guard clause version.

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
```diff
@@ -19,7 +19,8 @@
     include Hyrax::WorksControllerBehavior
 
     self.curation_concern_type = Hyrax::Test::SimpleWork
-    self.search_builder_class  = Hyrax::Test::SimpleWorkSearchBuilder
+    self.search_builder_class  = Wings::WorkSearchBuilder(Hyrax::Test::SimpleWork)
+    self.work_form_service     = Hyrax::FormFactory.new
```
jeremyf: Would we need an instance for this? It appears that you could have a module_function on `Hyrax::FormFactory.build`

lib/generators/hyrax/work_resource/templates/work.html.erb_spec.rb.erb
jeremyf: I think so.

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
jeremyf: Per @no-reply discussion, there is no "strong" need for an instance method. It does allow for greater extensibility.

lib/generators/hyrax/work_resource/templates/work.html.erb_spec.rb.erb
no-reply: :+1:

no-reply: :+1:

cjcolvar:
```suggestion
# Generated via
#  `rails generate hyrax:work <%= class_name %>`
```

lib/generators/hyrax/work_resource/templates/work_spec.rb.erb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Generated via
+#  `rails generate hyrax:work <%= class_name %>`
+require 'rails_helper'
+require 'hyrax/specs/shared_specs/hydra_works'
```
mjgiarlo: IT LIVES :dragon:

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
```diff
@@ -84,10 +85,16 @@
...
+        get :new
+
+        expect(assigns[:form]).to be_a Valkyrie::ChangeSet
+      end
+
+      it 'renders form' do
         get :new
 
-        expect(assigns[:form]).to be_kind_of Hyrax::WorkForm
+        expect(controller).to render_template('hyrax/base/new')
```
jeremyf: As this is an open question on what the form builder will return, we are looking to test does the page render (instead of the assumption of the type of object that is assigned). Note the above `it 'assigns a change_set as the form'` replicates the behavior.

lib/generators/hyrax/work_resource/work_resource_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+require 'rails/generators/model_helpers'
+
+class Hyrax::WorkResourceGenerator < Rails::Generators::NamedBase
+  # ActiveSupport can interpret models as plural which causes
+  # counter-intuitive route paths. Pull in ModelHelpers from
+  # Rails which warns users about pluralization when generating
+  # new models or scaffolds.
+  include Rails::Generators::ModelHelpers
+
+  source_root File.expand_path('../templates', __FILE__)
```
mjgiarlo: Kind of a meh, but: could also do `File.expand_path('templates', __dir__)` ?

lib/generators/hyrax/work_resource/work_resource_generator.rb
mjgiarlo: Alternatively, consider changing `#create_model_spec` to use `return unless rspec_installed?`

lib/wings.rb
```diff
@@ -55,7 +55,25 @@
...
+module Wings
+  ##
+  # @api public
+  #
+  # Provides a search builder for new valkyrie types that are indexed as their
+  # corresponding legacy ActiveFedora classes.
+  #
+  # @example
+  #   builder = Wings::WorkSearchBuilder(Monograph)
+  def self.WorkSearchBuilder(work_type) # rubocop:disable Naming/MethodName
```
mjgiarlo: Should we re-enable these cops down around line 75 or so?

lib/generators/hyrax/work_resource/templates/work_spec.rb.erb
no-reply: yesssss!

spec/services/hyrax/form_factory_spec.rb
mjgiarlo: And we're not open to changing that API at this point.

lib/generators/hyrax/work_resource/work_resource_generator.rb
no-reply: i think this line is just ripped from the old work generator. i'm open to whatever refactors here

lib/wings.rb
no-reply: i only recently learned this, but the end-of-line disable only applies to the one line! super handy

mjgiarlo: oh! that is super nice. TIL. thank you.

lib/generators/hyrax/work_resource/work_resource_generator.rb
mjgiarlo: k, consistency is more important then IMO. `wontfix`

no-reply:
> Is it worth documenting the existence of this new generator in the README, perhaps near the section on generating work types?

definitely. i'm not sure i know how to provide the relevant context, at this point, since this isn't a generator most users should run (yet). ideas?

mjgiarlo:
@no-reply :speech_balloon: 
> > Is it worth documenting the existence of this new generator in the README, perhaps near the section on generating work types?
> 
> definitely. i'm not sure i know how to provide the relevant context, at this point, since this isn't a generator most users should run (yet). ideas?

I fear I'm too far out-of-date to help w/ that. Happy to move forward without this in the README. Is there documentation external to this repo that could/should be amended to include this? Presumably it should be documented *somewhere* in the interim.

PR #4189, opened 2020-01-16
Undeprecate `index.as`

no-reply:
I think it turns out there is no advantage to deprecating this approach, given
where we've actually landed on Valkyrie indexing mechanisms (they'll be
completely parallel to the ActiveFedora ones). Additionally, `ActiveFedora` uses
this syntax internally in places, resulting in spammy deprecation warnings.

@samvera/hyrax-code-reviewers


cjcolvar:
The build is failing because there is still `spec/wings/indexing_configuration_spec.rb` which needs to be removed alongside `lib/wings/indexing_configuration.rb`.

no-reply:
@cjcolvar fixed

PR #4199, opened 2020-01-22
[WIP] Provide a docker-compose file for dev/test dependencies

aaron-collier:
This is an early pass at setting up docker-compose to remove the need for `solr_wrapper`, `fcrepo_wrapper`, and running `redis` on the command line.


docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+version: '3.6'
+
+services:
+  solr-test:
+    image: solr:7
```
no-reply: `solr:8` ?

PR #4206, opened 2020-01-23
Wings/add extracted text

### LINKED ISSUE #4018: WINGS: extracted text fails to copy content via LDP
### Descriptive summary

lib/wings/hydra/works/services/add_file_to_file_set correctly adds content for an original file and thumbnail image, but LDP fails to copy content when use is extracted text.  The same file can be successfully added when use is original file.

This was traced down to the LDP put request.  It is confirmed that the file that is part of the LDP request does have content.  The rest of the LDP put request looks the same as it does when use is original file.

It was also validated, that the same test passes when it is run directly under Hydra::Works instead of going through the wings wrapper.

It is unclear why this fails.

### Related code

FAILING TEST: https://github.com/samvera/hyrax/blob/master/spec/wings/hydra/works/services/add_file_to_file_set_spec.rb#L44-L61

LDP PUT: https://github.com/samvera/ldp/blob/master/lib/ldp/client/methods.rb#L86-L98

### Related work

PR #3939 wings wrapper for Hydra:Works:AddFileToFileSet

elrayle:
Fixes #4018 

### Changes

* pin active-fedora to 13.1.2
* remove pending marker for add_file_to_file_set for extracted text

The release of active-fedora v13.1.2 includes a bug fix to always rewind context.  Before this release, files added with the extracted text association were getting consumed prior the ldp put call that added them to Fedora resulting in the content being empty.  Other file associations may have also been effected.

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -31,7 +31,7 @@ SUMMARY
   # http://guides.rubyonrails.org/maintenance_policy.html
   spec.add_dependency 'rails', '~> 5.0'
 
-  spec.add_dependency 'active-fedora', '~> 13.0'
+  spec.add_dependency 'active-fedora', '~> 13.1.2'
   spec.add_dependency 'almond-rails', '~> 0.1'
   spec.add_dependency 'awesome_nested_set', '~> 3.1'
   spec.add_dependency 'blacklight', '~> 6.14'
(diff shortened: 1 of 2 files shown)


hyrax.gemspec
```diff
@@ -31,7 +31,7 @@ SUMMARY
   # http://guides.rubyonrails.org/maintenance_policy.html
   spec.add_dependency 'rails', '~> 5.0'
 
-  spec.add_dependency 'active-fedora', '~> 13.0'
+  spec.add_dependency 'active-fedora', '~> 13.1.2' # pin for required bug fix to always rewind context
```
no-reply:
```suggestion
  spec.add_dependency 'active-fedora', '~> 13.1.2'
```
i think it's okay to just let the commit message and the fact of the pin speak for themselves here.

no-reply:
Approved, with a suggestion about the comment.

I'm also wondering about whether these commits need to be re-ordered. the previously pending test will fail with older AF versions, right? will that failure cause confusion in `git bisect` scenarios?

Issue #4100, opened 2019-10-21
Allow CharacterizeJob to receive either an AF object or a Valkyrie resource as the FileSet


elrayle:
### Descriptive summary

Convert CharacterizeJob to allow it to receive either an ActiveFedora based FileSet or a Valkyrie::Resource equivalent FileSet.

### Rationale

Allow jobs to execute correctly with existing AF implementation or the wings Valkyrie implementation.

### Expected behavior

CharacterizeJob receives a Valkyrie::Resource and performs the characterization.

### Actual behavior

CharacterizeJob raises an exception when passed a Valkyrie::Resource.

### Approach for Conversion

See [Pattern for Jobs Receiving Parameter that may be an AF Object or a Valkyrie Resource](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727)

### Steps to reproduce the behavior

1. Need to add tests that validate that the same tests pass whether a Valkyrie::Resource or an AF object is passed.

See example of processing both objects in...  [file_set_actor_spec.rb](https://github.com/samvera/hyrax/blob/master/spec/actors/hyrax/actors/file_set_actor_spec.rb#L6)

For testing this job, you can use something like...
```
RSpec.describe CharacterizeJob do
  [true, false].each do |test_valkyrie|
    context "when test_valkyrie is #{test_valkyrie}" do
      let(:af_file_set) { create(:file_set) }
      let(:valk_file_set) { af_file_set.valkyrie_resource }
      let(:file_set) { test_valkyrie ? valk_file_set : af_file_set }
```

OR you can write tests that are specific to each object type, similar to the tests in [spec/jobs/hyrax/revoke_edit_job_spec.rb](https://github.com/samvera/hyrax/blob/master/spec/jobs/hyrax/revoke_edit_job_spec.rb).  This approach doubles the number of tests and doesn't provide certainty that the new code works exactly the same as the old code if the two versions of the tests diverge.  The approach to take is at the discretion and judgement of the developer.

### Related work

These have all be modified to support AF object or Valkyrie resource using the [established pattern]((https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727). 

* Issue #4101 Allow GrantEditToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
[...]

PR #4207, opened 2020-01-23
Remove collection size feature

mjgiarlo:
While looking into #4100, we noticed that `CharacterizeJob` not only saves a file set's `original_file`, but also reindexes the file set, *and* reindexes every collection to which the file set's work belongs. It appears that this feature was added way back in the Sufia days in order to support the use case of displaying on the collection show page the total size of all files uploaded to a collection's works.

Why is this problematic?

At large scale, this means touching a potentially huge portion of the repository—iterating over all files in all works in a collection—in order to display an arguably useful string to a subset of repository users who care about such information. Doing this in `CharacterizeJob` introduces significant potential delay in the ingest pipeline, increasing how long it takes for a deposit to finish and render in the Hyrax UI (e.g., by delaying the creation of derivatives).

We propose removing this feature for now and encourage folks who need this feature to contribute it back with a better-performing, more streamlined design.

### DIFF
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -111,6 +111,17 @@ module Hyrax
       end
     end
 
+    # @deprecated to be removed in 4.0.0; this feature was replaced with a
+    #   hard-coded null implementation
+    # @return [Fixnum] 0
+    def bytes
+      Deprecation.warn('#bytes has been deprecated for removal in Hyrax 4.0.0; ' \
+                       'The implementation of the indexed Collection size ' \
+                       'feature is extremely inefficient, so it has been removed. ' \
+                       'This method now returns a hard-coded `0` for compatibility.')
+      0
+    end
+
     # @api public
     # Retrieve the permission template for this collection.
     # @return [Hyrax::PermissionTemplate]
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -111,20 +111,6 @@ module Hyrax
       end
     end
 
-    # Compute the sum of each file in the collection using Solr to
-    # avoid having to access Fedora
-    #
-    # @return [Fixnum] size of collection in bytes
-    # @raise [RuntimeError] unsaved record does not exist in solr
-    def bytes
-      return 0 if member_object_ids.empty?
-
-      raise "Collection must be saved to query for bytes" if new_record?
-
-      # One query per member_id because Solr is not a relational database
-      member_object_ids.collect { |work_id| size_for_work(work_id) }.sum
-    end
-
     # @api public
     # Retrieve the permission template for this collection.
     # @return [Hyrax::PermissionTemplate]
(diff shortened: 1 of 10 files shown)


mjgiarlo:
> Do we need to add logic to deprecate the `#size` method usage? (e.g. have a `#size` method for collections that returns `n/a`)

I defer to @no-reply and others who have a hyrax in this hunt.

no-reply:
:100: to this! thanks.

i do think it would be nice to make the public methods emit deprecation warnings and return hard-coded & compatible null-ish values

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -111,18 +111,15 @@ def collection_type_gid_document_field_name
...
+    #   hard-coded null implementation
+    # @return [Fixnum] 0
     def bytes
-      return 0 if member_object_ids.empty?
-
-      raise "Collection must be saved to query for bytes" if new_record?
-
-      # One query per member_id because Solr is not a relational database
-      member_object_ids.collect { |work_id| size_for_work(work_id) }.sum
+      Deprecation.warn("#bytes has been deprecated for removal in Hyrax 4.0.0; " \
```
mjgiarlo: I'm a little surprised Rubocop didn't flag this for using double-quotes. I don't care deeply about this, but thought I'd raise it to see if you do.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -111,18 +111,15 @@ def collection_type_gid_document_field_name
...
+                       "The implementation of the indexed Collection size " \
+                       "feature extremely inefficient, so it has been removed. " \
```
mjgiarlo: Typo. I think we need an `is ` between `feature` and `extremely inefficient`

app/presenters/hyrax/collection_presenter.rb
```diff
@@ -41,17 +41,20 @@ def collection_type
...
       self.class.terms.select { |t| self[t].present? }
     end
 
+    ##
+    # @param [Symbol] key
+    # @return [Object]
     def [](key)
       case key
-      when :size
+      when size
```
mjgiarlo: I believe this ought to be a symbol, `:size`.

app/presenters/hyrax/collection_presenter.rb
```diff
@@ -60,8 +63,16 @@ def [](key)
...
     end
 
+    # @deprecated to be removed in 4.0.0; this feature was replaced with a
+    #   hard-coded null implementation
+    # @return [String] 'unknown'
     def size
-      number_to_human_size(@solr_document['bytes_lts'])
+      Deprecation.warn("#size has been deprecated for removal in Hyrax 4.0.0; " \
+                       "The implementation of the indexed Collection size " \
+                       "feature extremely inefficient, so it has been removed. " \
```
mjgiarlo: Same comments as above re: double- vs. single-quotes and the need to add `is `.

spec/presenters/hyrax/collection_presenter_spec.rb
```diff
@@ -126,12 +122,6 @@
     it { is_expected.to eq ['adc12v'] }
   end
 
-  describe "#size", :clean_repo do
```
mjgiarlo: Should we tweak these tests to verify that the hard-coded value is returned? And/or that `Deprecation` receives `:warn`?

app/presenters/hyrax/collection_presenter.rb
jeremyf: Agreed, this should be `:size`

app/models/concerns/hyrax/collection_behavior.rb
jcoyne: Bixby don't care about how many quotes you use.

elrayle:
I know this is already merged, but I just want to add my kudos to @mjgiarlo (and any others working on this PR).  Always happy to see performance enhancements.

PR #4213, opened 2020-01-24
Extract characterization service to allow ActiveFedora and Valkyrie filesets

### LINKED ISSUE #4100: Allow CharacterizeJob to receive either an AF object or a Valkyrie resource as the FileSet

aaron-collier:
Fixes #4100  

This extracts the majority of the current CharacterizerJob into a Characterizer that implements individual characterize methods per FileSet type.

The pattern followed here was based on suggestion by @no-reply 

Note: the remainder of characterizer_job_spec can likely be removed after #4207

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/characterize_job.rb
+++ b/app/jobs/characterize_job.rb
@@ -1,33 +1,10 @@
+# Characterizes the file at 'filepath' if available, otherwise, pulls a copy from the repository
+# and runs characterization on that file.
+
 class CharacterizeJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
 
-  # Characterizes the file at 'filepath' if available, otherwise, pulls a copy from the repository
-  # and runs characterization on that file.
-  # @param [FileSet] file_set
-  # @param [String] file_id identifier for a Hydra::PCDM::File
-  # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
-  def perform(file_set, file_id, filepath = nil)
-    raise "#{file_set.class.characterization_proxy} was not found for FileSet #{file_set.id}" unless file_set.characterization_proxy?
-    filepath = Hyrax::WorkingDirectory.find_or_retrieve(file_id, file_set.id) unless filepath && File.exist?(filepath)
-    characterize(file_set, file_id, filepath)
-    CreateDerivativesJob.perform_later(file_set, file_id, filepath)
+  def perform(work)
+    Hyrax::Characterizer.for(source: work).characterize
   end
-
-  private
-
-    def characterize(file_set, _file_id, filepath)
-      Hydra::Works::CharacterizationService.run(file_set.characterization_proxy, filepath)
-      Rails.logger.debug "Ran characterization on #{file_set.characterization_proxy.id} (#{file_set.characterization_proxy.mime_type})"
-      file_set.characterization_proxy.alpha_channels = channels(filepath) if file_set.image? && Hyrax.config.iiif_image_server?
-      file_set.characterization_proxy.save!
-      file_set.update_index
-    end
-
-    def channels(filepath)
-      ch = MiniMagick::Tool::Identify.new do |cmd|
-        cmd.format '%[channels]'
-        cmd << filepath
-      end
-      [ch]
-    end
 end
--- a/app/jobs/characterize_job.rb
+++ b/app/jobs/characterize_job.rb
@@ -1,10 +1,9 @@
 # Characterizes the file at 'filepath' if available, otherwise, pulls a copy from the repository
 # and runs characterization on that file.
-
 class CharacterizeJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
 
-  def perform(work)
-    Hyrax::Characterizer.for(source: work).characterize
+  def perform(file_set)
+    Hyrax::Characterizer.for(source: file_set).characterize
   end
 end
--- a/app/models/hyrax/file_set.rb
+++ b/app/models/hyrax/file_set.rb
@@ -7,6 +7,8 @@ module Hyrax
   # @see https://wiki.duraspace.org/display/samvera/Hydra%3A%3AWorks+Shared+Modeling
   class FileSet < Hyrax::Resource
     include Hyrax::Schema(:core_metadata)
+    include Hyrax::FileSet::Characterization
+    include Hydra::Works::MimeTypes
 
     attribute :file_ids, Valkyrie::Types::Array.of(Valkyrie::Types::ID) # id for FileMetadata resources
     attribute :original_file_id, Valkyrie::Types::ID # id for FileMetadata resource
--- a/app/services/hyrax/characterizer.rb
+++ b/app/services/hyrax/characterizer.rb
@@ -11,9 +11,8 @@ module Hyrax
       case source
       when Hyrax::FileSetBehavior # ActiveFedora
         FileSetCharacterizer.new(source: source)
-      when Hyrax::Resource # Valkyrie
-        byebug
-        ResourceVisibilityPropagator.new(source: source)
+      when Hyrax::FileSet # Valkyrie
+        FileSetCharacterizer.new(source: source)
       end
     end
   end
--- a/app/services/hyrax/characterizer.rb
+++ b/app/services/hyrax/characterizer.rb
@@ -12,7 +12,7 @@ module Hyrax
       when Hyrax::FileSetBehavior # ActiveFedora
         FileSetCharacterizer.new(source: source)
       when Hyrax::FileSet # Valkyrie
-        FileSetCharacterizer.new(source: source)
+        ResourceCharacterizer.new(source: source)
       end
     end
   end
--- a/app/services/hyrax/characterizer.rb
+++ b/app/services/hyrax/characterizer.rb
@@ -1,0 +1,20 @@
+module Hyrax
+  ##
+  # @abstract Propogates visibility from a provided object (e.g. a Work) to some
+  # group of its members (e.g. file_sets).
+  class Characterizer
+    ##
+    # @param source [#visibility] the object to propogate visibility from
+    #
+    # @return [#propogate]
+    def self.for(source:)
+      case source
+      when Hyrax::FileSetBehavior # ActiveFedora
+        FileSetCharacterizer.new(source: source)
+      when Hyrax::Resource # Valkyrie
+        byebug
+        ResourceVisibilityPropagator.new(source: source)
+      end
+    end
+  end
+end
--- a/app/services/hyrax/characterizer.rb
+++ b/app/services/hyrax/characterizer.rb
@@ -1,12 +1,14 @@
+# frozen_string_literal: true
+
 module Hyrax
   ##
-  # @abstract Propogates visibility from a provided object (e.g. a Work) to some
-  # group of its members (e.g. file_sets).
+  # Determines which characterizer to run based on the file_set type
+  # allowing implementation of Valkyrie file_sets
   class Characterizer
     ##
-    # @param source [#visibility] the object to propogate visibility from
+    # @param source: the object to run a characterizer on
     #
-    # @return [#propogate]
+    # @return [#characterize]
     def self.for(source:)
       case source
       when Hyrax::FileSetBehavior # ActiveFedora
--- a/app/services/hyrax/file_set_characterizer.rb
+++ b/app/services/hyrax/file_set_characterizer.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Propogates visibility from a given Work to its FileSets
+  class FileSetCharacterizer 
+    ##
+    # @!attribute [rw] source
+    #   @return [#visibility]
+    attr_accessor :source
+
+    ##
+    # @param source [#visibility] the object to propogate visibility from
+    def initialize(source:)
+      self.source = source
+    end
+
+    ##
+    # @return [void]
+    #
+    # @raise [RuntimeError] if visibility propogation fails
+    def characterize
+      Hydra::Works::CharacterizationService.run(source.characterization_proxy, filepath)
+      Rails.logger.debug "Ran characterization on #{source.characterization_proxy.id} (#{source.characterization_proxy.mime_type})"
+      source.characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
+      source.characterization_proxy.save!
+      source.update_index
+      source.parent&.in_collections&.each(&:update_index)
+      CreateDerivativesJob.perform_later(source, source.original_file.id, filepath)
+    end
+
+    private
+
+      def filepath
+        Hyrax::WorkingDirectory.find_or_retrieve(source.original_file.id, source.id)
+      end
+
+      def channels(path)
+        ch = MiniMagick::Tool::Identify.new do |cmd|
+          cmd.format '%[channels]'
+          cmd << path
+        end
+        [ch]
+      end
+  end
+end
--- a/app/services/hyrax/file_set_characterizer.rb
+++ b/app/services/hyrax/file_set_characterizer.rb
@@ -20,6 +20,7 @@ module Hyrax
     #
     # @raise [RuntimeError] if visibility propogation fails
     def characterize
+      byebug
       Hydra::Works::CharacterizationService.run(source.characterization_proxy, filepath)
       Rails.logger.debug "Ran characterization on #{source.characterization_proxy.id} (#{source.characterization_proxy.mime_type})"
       source.characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
--- a/app/services/hyrax/file_set_characterizer.rb
+++ b/app/services/hyrax/file_set_characterizer.rb
@@ -20,16 +20,21 @@ module Hyrax
     #
     # @raise [RuntimeError] if visibility propogation fails
     def characterize
-      Hydra::Works::CharacterizationService.run(source.characterization_proxy, filepath)
-      Rails.logger.debug "Ran characterization on #{source.characterization_proxy.id} (#{source.characterization_proxy.mime_type})"
-      source.characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
-      source.characterization_proxy.save!
+      Hydra::Works::CharacterizationService.run(characterization_proxy, filepath)
+      Rails.logger.debug "Ran characterization on #{characterization_proxy.id} (#{characterization_proxy.mime_type})"
+      characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
+      characterization_proxy.save!
       source.update_index
       source.parent&.in_collections&.each(&:update_index)
       CreateDerivativesJob.perform_later(source, source.original_file.id, filepath)
     end
 
     private
+
+      def characterization_proxy
+        raise "#{source.class.characterization_proxy} was not found for FileSet #{source.id}" unless source.characterization_proxy?
+        source.characterization_proxy
+      end
 
       def filepath
         Hyrax::WorkingDirectory.find_or_retrieve(source.original_file.id, source.id)
--- a/app/services/hyrax/file_set_characterizer.rb
+++ b/app/services/hyrax/file_set_characterizer.rb
@@ -2,38 +2,33 @@
 
 module Hyrax
   ##
-  # Propogates visibility from a given Work to its FileSets
+  # Characterizes an ActiveFedora based FileSet
   class FileSetCharacterizer
     ##
     # @!attribute [rw] source
-    #   @return [#visibility]
+    #   @return [#characterize]
     attr_accessor :source
 
     ##
-    # @param source [#visibility] the object to propogate visibility from
+    # @param source the object to characterize
     def initialize(source:)
-      self.source = source
+      @source = source
     end
 
     ##
     # @return [void]
     #
-    # @raise [RuntimeError] if visibility propogation fails
+    # @raise [RuntimeError] if FileSet is missing the characterization_proxy
     def characterize
       Hydra::Works::CharacterizationService.run(characterization_proxy, filepath)
       Rails.logger.debug "Ran characterization on #{characterization_proxy.id} (#{characterization_proxy.mime_type})"
       characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
       characterization_proxy.save!
-      update_source
+      source.update_index
       CreateDerivativesJob.perform_later(source, source.original_file.id, filepath)
     end
 
     private
-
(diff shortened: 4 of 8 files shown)


app/jobs/characterize_job.rb
```diff
@@ -1,34 +1,10 @@
+# Characterizes the file at 'filepath' if available, otherwise, pulls a copy from the repository
+# and runs characterization on that file.
+
```
mjgiarlo: A trivial thing: would you consider dropping the newline between the class documentation and the `class` declaration?

app/jobs/characterize_job.rb
```diff
@@ -1,34 +1,10 @@
...
-  # @param [FileSet] file_set
-  # @param [String] file_id identifier for a Hydra::PCDM::File
-  # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
-  def perform(file_set, file_id, filepath = nil)
-    raise "#{file_set.class.characterization_proxy} was not found for FileSet #{file_set.id}" unless file_set.characterization_proxy?
-    filepath = Hyrax::WorkingDirectory.find_or_retrieve(file_id, file_set.id) unless filepath && File.exist?(filepath)
-    characterize(file_set, file_id, filepath)
-    CreateDerivativesJob.perform_later(file_set, file_id, filepath)
+  def perform(work)
+    Hyrax::Characterizer.for(source: work).characterize
```
mjgiarlo: How would you feel about renaming `work` on lines 7 and 8 to `file_set`?

app/models/hyrax/file_set.rb
```diff
@@ -7,6 +7,8 @@ module Hyrax
   # @see https://wiki.duraspace.org/display/samvera/Hydra%3A%3AWorks+Shared+Modeling
   class FileSet < Hyrax::Resource
     include Hyrax::Schema(:core_metadata)
+    include Hyrax::FileSet::Characterization
```
mjgiarlo: Both @aaron-collier and I **believe** this is the smallest possible set of includes to make sure resource-backed file sets respond to messages that are sent in the course of characterizing them.

`Hyrax::FileSet::Characterization` is needed to set up the characterization proxy, and `Hydra::Works::MimeTypes` is needed to ask a file set what sort of files it wraps (*e.g.*, `#image?`), based on mime types.

app/services/hyrax/characterizer.rb
```diff
@@ -0,0 +1,19 @@
+module Hyrax
+  ##
+  # @abstract Determins which characterizer to run based on the file_set type
```
mjgiarlo: Typo: `Determins`

mjgiarlo: I'm not sure I'd consider this class to be an abstract class, so I'd be inclined to drop the `@abstract` tag.

app/services/hyrax/characterizer.rb
```diff
@@ -0,0 +1,19 @@
+module Hyrax
```
mjgiarlo: I am guessing this file also needs the `frozen_string_literal` pragma.

app/services/hyrax/characterizer.rb
```diff
@@ -0,0 +1,19 @@
...
+  # @abstract Determins which characterizer to run based on the file_set type
+  # allowing implementation of Valkyrie file_sets
+  class Characterizer
+    ##
+    # @param source: the object to run a characterizer on
+    #
+    # @return [#characterize]
+    def self.for(source:)
+      case source
+      when Hyrax::FileSetBehavior # ActiveFedora
```
jcoyne: would this ever be the case?  FileSetBehavior is a module not a class. https://github.com/samvera/hyrax/blob/addb09f382feb86ab32f937d1c41feaf8d7a1537/app/models/concerns/hyrax/file_set_behavior.rb#L2

app/services/hyrax/file_set_characterizer.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def update_source
+        source.update_index
+        source.parent&.in_collections&.each(&:update_index)
+      end
+
+      def characterization_proxy
+        raise "#{source.class.characterization_proxy} was not found for FileSet #{source.id}" unless source.characterization_proxy?
+        source.characterization_proxy
+      end
```
mjgiarlo: This method is called five times, I think, which means a bunch of unnecessary checks to see if the proxy exists. Would you be open to inlining this back up in `#characterize`?

app/services/hyrax/file_set_characterizer.rb
```diff
@@ -0,0 +1,55 @@
...
+      Rails.logger.debug "Ran characterization on #{characterization_proxy.id} (#{characterization_proxy.mime_type})"
+      characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
+      characterization_proxy.save!
+      update_source
+      CreateDerivativesJob.perform_later(source, source.original_file.id, filepath)
+    end
+
+    private
+
+      def update_source
```
mjgiarlo: I'm not clear on this method. It's updating the source and updating the parent's collections, but it's not updating the characterization proxy. I might be in favor of either putting this code back in `#characterize` until a clearer abstraction is found.

app/services/hyrax/characterizer.rb
aaron-collier: That is the case when passing a `FileSet` and not a `Hyrax::FileSet`. We see the same style here: https://github.com/samvera/hyrax/blob/master/app/services/hyrax/visibility_propagator.rb#L12. We verified this when stepping through this.

app/services/hyrax/file_set_characterizer.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Characterizes an ActiveFedora based FileSet
+  class FileSetCharacterizer
+    ##
+    # @!attribute [rw] source
+    #   @return [#characterize]
```
mjgiarlo: I'm a bit fuzzy on this. Is this saying that the thing in the `source` attr responds to `#characterize`. Does it? I would expect `source` to return an instance of an AF-based  file set.

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Characterizes a Valkyrie based FileSet
+  class ResourceCharacterizer
+    ##
+    # @!attribute [rw] source
+    #   @return [#characterize]
```
mjgiarlo: Same as above. I would expect `source` to return an instance of a Valkyrie-based file set.

app/services/hyrax/file_set_characterizer.rb
```diff
@@ -0,0 +1,55 @@
...
+    attr_accessor :source
+
+    ##
+    # @param source the object to characterize
+    def initialize(source:)
+      self.source = source
```
mjgiarlo: Should this be `@source = source`?

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,54 @@
...
+    attr_accessor :source
+
+    ##
+    # @param source the object to characterize
+    def initialize(source:)
+      self.source = source
```
mjgiarlo: Should this be `@source = source`?

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    ##
+    # @return [void]
+    #
+    # @raise [RuntimeError] if visibility propogation fails
```
mjgiarlo: I think the text here needs to change.

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,54 @@
...
+    def characterize
+      Hydra::Works::CharacterizationService.run(characterization_proxy, filepath)
+      Rails.logger.debug "Ran characterization on #{characterization_proxy.id} (#{characterization_proxy.mime_type})"
+      characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
+      persist
```
mjgiarlo: Similar to the above comment, I wonder about the utility of moving the persistence code into this method.

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def persist
+        Hyrax.persister.save(resource: characterization_proxy)
+        Hyrax.persister.save(resource: source)
+      end
+
+      def characterization_proxy
+        raise "#{source.class.characterization_proxy} was not found for FileSet #{source.id}" unless source.characterization_proxy?
+        source.characterization_proxy
+      end
```
mjgiarlo: Same comment on this method as above.

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def filepath
+        Hyrax::WorkingDirectory.find_or_retrieve(source.original_file.id, source.id.id)
```
mjgiarlo: what would you think about tossing a comment here calling out the double id call is intentional (and required)

app/services/hyrax/characterizer.rb
mjgiarlo: @jcoyne yeah. @aaron-collier and I verified this yesterday.

spec/jobs/characterize_job_spec.rb
```diff
@@ -1,66 +1,34 @@
 RSpec.describe CharacterizeJob do
-  let(:file_set_id) { 'abc12345' }
-  let(:filename)    { Rails.root.join('tmp', 'uploads', 'ab', 'c1', '23', '45', 'abc12345', 'picture.png').to_s }
-  let(:file_set) do
-    FileSet.new(id: file_set_id).tap do |fs|
-      allow(fs).to receive(:original_file).and_return(file)
-      allow(fs).to receive(:update_index)
+  context 'when the work is an ActiveFedora FileSet' do
```
mjgiarlo: Is this a context we want to keep here? the job is not switching behavior anymore.

spec/jobs/characterize_job_spec.rb
```diff
@@ -1,66 +1,34 @@
...
-    let(:collection) { build(:collection_lw) }
-
-    before do
-      allow(file_set).to receive(:parent).and_return(work)
-      allow(work).to receive(:in_collections).and_return([collection])
-    end
-    it "reindexes the collection" do
-      expect(collection).to receive(:update_index)
-      described_class.perform_now(file_set, file.id)
+      before do
```
mjgiarlo: I think this test should be removed, since the characterizejob no longer handles this behavior. I believe the test should only ensure that the characterizer factory returns a thing that `#characterize` gets called on.

spec/services/hyrax/file_set_characterizer_spec.rb
```diff
@@ -0,0 +1,42 @@
+RSpec.describe Hyrax::FileSetCharacterizer do
+  subject(:characterizer) { described_class.new(source: file_set) }
```
mjgiarlo: I have a slight trivial preference for adding a newline between subjects and lets but `/shrug`

spec/services/hyrax/file_set_characterizer_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(file_set).to receive(:update_index)
+      expect(CreateDerivativesJob).to receive(:perform_later).with(file_set, file.id, filename)
+      characterizer.characterize
+    end
+  end
+
+  context 'when the characterization proxy content is absent' do
+    before { allow(file_set).to receive(:characterization_proxy?).and_return(false) }
+    it 'raises an error' do
+      expect { characterizer.characterize }.to raise_error(StandardError, /original_file was not found/)
```
mjgiarlo: I am wondering if this can/should change to `RuntimeError`

app/services/hyrax/file_set_characterizer.rb
jeremyf: I personally prefer using a setter method, as it allows checks/coercion.

spec/services/hyrax/resource_characterizer_spec.rb
```diff
@@ -0,0 +1,32 @@
+RSpec.describe Hyrax::ResourceCharacterizer do
+  subject(:characterizer) { described_class.new(source: resource) }
+  let(:resource) { FactoryBot.create(:file_set).valkyrie_resource }
```
mjgiarlo: I think this is dead code since each context below has its own resource let block

spec/services/hyrax/resource_characterizer_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+  end
+
+  context 'without a complete Valkyrie FileSet' do
+    let(:resource) { FactoryBot.create(:file_set).valkyrie_resource }
+
+    before { allow(resource).to receive(:characterization_proxy?).and_return(false) }
+
+    it 'runs Hydra::Works::CharacterizationService and creates a CreateDerivativesJob' do
+      expect { characterizer.characterize }.to raise_error(StandardError, /original_file was not found/)
```
mjgiarlo: RuntimeError?

app/models/hyrax/file_set.rb
randalldfloyd: +1 on including `MimeTypes`, the derivatives job also needs to know that for a resource-backed file set.

app/services/hyrax/characterizer.rb
mjgiarlo: @jcoyne see:

```ruby
2.5.5 :001 > file_set = FileSet.new
 => #<FileSet id: nil, head: [], tail: [], depositor: nil, title: [], date_uploaded: nil, date_modified: nil, alternative_title: [], label: nil, relative_path: nil, import_url: nil, resource_type: [], creator: [], contributor: [], description: [], abstract: [], keyword: [], license: [], rights_notes: [], rights_statement: [], access_right: [], publisher: [], date_created: [], subject: [], language: [], identifier: [], based_near: [], related_url: [], bibliographic_citation: [], source: [], access_control_id: nil, embargo_id: nil, lease_id: nil> 

2.5.5 :002 > Hyrax::FileSetBehavior === file_set
 => true 
```

jcoyne: This seems really unclear and easy to violate this assumption.

jcoyne: Could you use `ActiveFedora::Base` instead?

no-reply: i'm :+1: on using `ActiveFedora::Base`, with the idea that as long as we're in a position where we need to do type checks we should do the least strict type checks that will work.

in theory, folks can implement a valid AF File Set class that doesn't include `Hyrax::FileSetBehavior`, right?

app/models/hyrax/file_set.rb
no-reply: if these are required `FileSet` interface, let's get them into the `'a Hyrax::FileSet'` shared specs in `lib/hyrax/specs/shared_specs/hydra_works.rb`?

i'm tempted to look for a way around making this the interface, though. so far we've been mostly successful in avoiding putting business logic in the models. while there's definitely space for compromise on that, it seems worth at least poking at other possibilities. any ideas?

spec/services/hyrax/resource_characterizer_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe Hyrax::ResourceCharacterizer do
+  subject(:characterizer) { described_class.new(source: resource) }
+
+  before do
+    allow(Hyrax::FileSet).to receive(:find).with(resource.id).and_return(resource)
```
no-reply: can `Hyrax::FileSet` do this? is this coming from one of the includes??

spec/services/hyrax/resource_characterizer_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    allow(Hydra::Works::CharacterizationService).to receive(:run)
+    allow(Hyrax.persister).to receive(:save)
+    allow(CreateDerivativesJob).to receive(:perform_later)
```
no-reply: is this just to avoid running the job? if so, it shouldn't be needed since we use the test queue adapter by default in the test suite.

spec/services/hyrax/resource_characterizer_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe Hyrax::ResourceCharacterizer do
+  subject(:characterizer) { described_class.new(source: resource) }
+
+  before do
+    allow(Hyrax::FileSet).to receive(:find).with(resource.id).and_return(resource)
+    allow(Hydra::Works::CharacterizationService).to receive(:run)
+    allow(Hyrax.persister).to receive(:save)
```
no-reply: okay to just actually call save on the persister?

if we want to avoid round tripping wings, we could do `allow(Hyrax).to receive(:metadata_adapter).and_return(Valkyrie::MetadataAdapter.find(:test_adapter))` to use the test/metadata adapter for the duration.

i've considered rigging this into some RSpec metadata, but it hasn't made it's way into the codebase yet.

spec/services/hyrax/resource_characterizer_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    allow(CreateDerivativesJob).to receive(:perform_later)
+  end
+
+  context 'with a complete Valkyrie FileSet' do
+    let(:resource) { FactoryBot.create(:file_set, content: File.open(fixture_path + '/world.png')).valkyrie_resource }
+
+    it 'runs Hydra::Works::CharacterizationService and creates a CreateDerivativesJob' do
+      expect(Hydra::Works::CharacterizationService).to receive(:run).with(resource.original_file, anything)
+      expect(CreateDerivativesJob).to receive(:perform_later).with(resource, resource.original_file.id, anything)
+      expect(Hyrax.persister).to receive(:save).twice
```
no-reply: twice `with_args`? or, again, consider just actually saving the objects to memory and expecting them to be saved.

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,52 @@
...
+    def initialize(source:)
+      @source = source
+    end
+
+    ##
+    # @return [void]
+    #
+    # @raise [RuntimeError] if FileSet is missing the characterization_proxy
+    def characterize
+      Hydra::Works::CharacterizationService.run(characterization_proxy, filepath)
```
no-reply: what do you think about injecting this dependency?

```ruby
##
# @param service [#run]
def characterize(service: Hydra::Works::CharacterizationService)
  service.run(characterization_proxy, filepath)
  ...
end
```

in this case, we could inject a spy service in the test suite and check its arguments, instead of hard-stubbing the class method.

app/services/hyrax/file_set_characterizer.rb
```diff
@@ -0,0 +1,50 @@
...
+    # @param source the object to characterize
+    def initialize(source:)
+      @source = source
+    end
+
+    ##
+    # @return [void]
+    #
+    # @raise [RuntimeError] if FileSet is missing the characterization_proxy
+    def characterize
```
no-reply: same method signature/dependency injection suggestion as https://github.com/samvera/hyrax/pull/4213/files#r370888617

app/services/hyrax/resource_characterizer.rb
```diff
@@ -0,0 +1,52 @@
...
+      Rails.logger.debug "Ran characterization on #{characterization_proxy.id} (#{characterization_proxy.mime_type})"
+      characterization_proxy.alpha_channels = channels(filepath) if source.image? && Hyrax.config.iiif_image_server?
+      Hyrax.persister.save(resource: characterization_proxy)
+      Hyrax.persister.save(resource: source)
```
no-reply: should this trigger a reindex of `source` in some way, or is it not needed?

app/models/hyrax/file_set.rb
no-reply: re: `MimeTypes`,

`Hyrax::FileSet` doesn't (yet) have a `#mime_type` property, so i think the include is injecting a bunch of methods that will currently raise errors?

no-reply: aho noticing that `::FileSet` *also* doesn't have a `mime_type` property and just delegates it to `#characterization_proxy`. it seems like we should follow suit and leave `#mime_type` at the file level?

no-reply: wondering about this as a possible solution to the `#mime_type` issue https://github.com/samvera/hyrax/pull/4215

PR #4215, opened 2020-01-25
Create `FileSetDescription` to find characterization for FileSets

### LINKED ISSUE #4226: use a Dry::Struct/Types default for FileMetadata type attribute
### Descriptive summary

In file `app/models/hyrax/file_metadata.rb`, method `#for` sets the default of type to `Hyrax::FileSet::ORIGINAL_FILE`.  It is preferred to use a Dry::Struct/Types default in the type definition instead of in the `#for` method.

This would look something like...

```
attribute :type, ::Valkyrie::Types::Set, default: [Hyrax::FileSet.original_file_use]
```

Need to determine the impact this will have on setting the `type` in the `#for` method.

### Related work

PR #4225 [Requested change](https://github.com/samvera/hyrax/pull/4225#discussion_r371544309)

no-reply:
**For discussion as possible input for #4213** 

Valkyrie-based FileSet objects don't have direct access to characterization
information. In ActiveFedora, we used a `#characterization_proxy` to provide
that information. Here, we provide a small service that can find the "primary"
file (`FileMetadata`) for a given FileSet, and ask it about its description.

Closes #4226

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -100,29 +100,28 @@ module Hyrax
         end
 
         def normalize_relation_for_active_fedora(relation)
-          return relation if relation.is_a? Symbol
           return relation.to_sym if relation.respond_to? :to_sym
 
-          # TODO: whereever these are set, they should use FileSet.*_use... making the casecmp unnecessary
-          return :original_file if relation.to_s.casecmp(Hyrax::FileMetadata::Use::ORIGINAL_FILE.to_s)
-          return :extracted_file if relation.to_s.casecmp(Hyrax::FileMetadata::Use::EXTRACTED_TEXT.to_s)
-          return :thumbnail_file if relation.to_s.casecmp(Hyrax::FileMetadata::Use::THUMBNAIL.to_s)
-          :original_file
+          case relation
+          when Hyrax::FileMetadata::Use::ORIGINAL_FILE
+            :original_file
+          when Hyrax::FileMetadata::Use::EXTRACTED_TEXT
+            :extracted_file
+          when Hyrax::FileMetadata::Use::THUMBNAIL
+            :thumbnail_file
+          else
+            :original_file
+          end
         end
 
+        ##
+        # @return [RDF::URI]
         def normalize_relation_for_valkyrie(relation)
-          # TODO: When this is fully switched to valkyrie, this should probably be removed and relation should always be passed
-          #       in as a valid URI already set to the file's use
-          case relation.to_s.to_sym
-          when :original_file
-            Hyrax::FileMetadata::Use::ORIGINAL_FILE
-          when :extracted_file
-            Hyrax::FileMetadata::Use::EXTRACTED_TEXT
-          when :thumbnail_file
-            Hyrax::FileMetadata::Use::THUMBNAIL
-          else
-            Hyrax::FileMetadata::Use::ORIGINAL_FILE
-          end
+          return relation if relation.is_a?(RDF::URI)
+
+          Hyrax::FileMetadata::Use.uri_for(use: relation.to_sym)
+        rescue ArgumentError
+          Hyrax::FileMetadata::Use::ORIGINAL_FILE
         end
     end
   end
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -104,9 +104,9 @@ module Hyrax
           return relation.to_sym if relation.respond_to? :to_sym
 
           # TODO: whereever these are set, they should use FileSet.*_use... making the casecmp unnecessary
-          return :original_file if relation.to_s.casecmp(Hyrax::FileSet::ORIGINAL_FILE_USE.to_s)
-          return :extracted_file if relation.to_s.casecmp(Hyrax::FileSet::EXTRACTED_TEXT_USE.to_s)
-          return :thumbnail_file if relation.to_s.casecmp(Hyrax::FileSet::THUMBNAIL_USE.to_s)
+          return :original_file if relation.to_s.casecmp(Hyrax::FileMetadata::Use::ORIGINAL_FILE.to_s)
+          return :extracted_file if relation.to_s.casecmp(Hyrax::FileMetadata::Use::EXTRACTED_TEXT.to_s)
+          return :thumbnail_file if relation.to_s.casecmp(Hyrax::FileMetadata::Use::THUMBNAIL.to_s)
           :original_file
         end
 
@@ -115,13 +115,13 @@ module Hyrax
           #       in as a valid URI already set to the file's use
           case relation.to_s.to_sym
           when :original_file
-            Hyrax::FileSet::ORIGINAL_FILE_USE
+            Hyrax::FileMetadata::Use::ORIGINAL_FILE
           when :extracted_file
-            Hyrax::FileSet.EXTRACTED_TEXT_USE
+            Hyrax::FileMetadata::Use::EXTRACTED_TEXT
           when :thumbnail_file
-            Hyrax::FileSet::THUMBNAIL_USE
+            Hyrax::FileMetadata::Use::THUMBNAIL
           else
-            Hyrax::FileSet::ORIGINAL_FILE_USE
+            Hyrax::FileMetadata::Use::ORIGINAL_FILE
           end
         end
     end
--- a/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
@@ -27,7 +27,7 @@ module Hyrax
       filter_docs_with_edit_access!
       copy_visibility = []
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] } if params[:embargoes]
-      af_objects = Hyrax.query_service.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
+      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
       af_objects.each do |curation_concern|
         Hyrax::Actors::EmbargoActor.new(curation_concern).destroy
         # if the concern is a FileSet, set its visibility and visibility propagation
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -10,7 +10,7 @@ module Hyrax
     module Use
       ORIGINAL_FILE = ::Valkyrie::Vocab::PCDMUse.OriginalFile
       EXTRACTED_TEXT = ::Valkyrie::Vocab::PCDMUse.ExtractedText
-      THUMBNAIL = ::Valkyrie::Vocab::PCDMUse.Thumbnail
+      THUMBNAIL = ::Valkyrie::Vocab::PCDMUse.ThumbnailImage
 
       ##
       # @param use [Symbol]
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -21,7 +21,7 @@ module Hyrax
     attribute :label, ::Valkyrie::Types::Set
     attribute :original_filename, ::Valkyrie::Types::Set
     attribute :mime_type, ::Valkyrie::Types::String.default(GENERIC_MIME_TYPE)
-    attribute :type, ::Valkyrie::Types::Set # AF::File type
+    attribute :type, ::Valkyrie::Types::Set.default([Use::ORIGINAL_FILE])
     attribute :content, ::Valkyrie::Types::Set
 
     # attributes set by fits
@@ -86,8 +86,7 @@ module Hyrax
     def self.for(file:)
       new(label: file.original_filename,
           original_filename: file.original_filename,
-          mime_type: file.content_type,
-          type: file.try(:type) || [Use::ORIGINAL_FILE])
+          mime_type: file.content_type)
     end
 
     ##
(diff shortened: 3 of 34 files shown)


app/services/hyrax/characterization/file_set_description.rb
```diff
@@ -0,0 +1,42 @@
...
+      #   used for characterization
+      def initialize(file_set:, primary_file: :original_file)
+        self.file_set = file_set
+        @primary_file = primary_file
+      end
+
+      ##
+      # @api public
+      # @return [Hyrax::FileMetadata] the member file to use for characterization
+      def primary_file
```
jeremyf: Something of this sits a bit odd. Setting the `primary_file` based on a symbol, but returning an `Hyrax::FileMetadata` object. It feels odd to have an assigned instance_variable that is different from the returned value from the method named the same as the instance_variable.

Do we want / need memoization for the primary_file (as FileMetadata)?

no-reply: i think my expectation was that callers would want to use a symbol since this is how the existing `characterization_proxy` class attribute is handled.

do you have suggestions about alternate approaches?

jeremyf: If I had my druthers, I might look at renaming the initialize parameter to `primary_file_type`. Though in typing that I'm not overly satisfied. However, to keep backwards API compatability, I think what you have is reasonable. The file/class is not that difficult to follow, as there is little going on.

no-reply: :+1: i definitely get the awkwardness, but don't have a better solution at hand.

jeremyf: I'm good with leaving as is. The compact nature of the class keeps this cognitive load to what I think is reasonable.

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -100,29 +100,26 @@ def normalize_relation(relation)
...
-          return relation if relation.is_a? Symbol
           return relation.to_sym if relation.respond_to? :to_sym
 
-          # TODO: whereever these are set, they should use FileSet.*_use... making the casecmp unnecessary
-          return :original_file if relation.to_s.casecmp(Hyrax::FileSet::ORIGINAL_FILE_USE.to_s)
-          return :extracted_file if relation.to_s.casecmp(Hyrax::FileSet::EXTRACTED_TEXT_USE.to_s)
-          return :thumbnail_file if relation.to_s.casecmp(Hyrax::FileSet::THUMBNAIL_USE.to_s)
-          :original_file
+          case relation
+          when Hyrax::FileMetadata::Use::ORIGINAL_FILE
```
no-reply:
```suggestion
```

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -100,29 +100,26 @@ def normalize_relation(relation)
...
+            :original_file
```
no-reply:
```suggestion
```

app/actors/hyrax/actors/file_actor.rb
no-reply: as noted by suggestions, this branch could be dropped and fall through to the `else` instead. i think the existing way is more clear, but only slightly.

lib/wings/hydra/works/services/add_file_to_file_set.rb
```diff
@@ -37,9 +37,9 @@ def association_type(type)
         end
 
         def type_to_association_type(type)
-          return :original_file if type.to_s.casecmp?(Hyrax::FileSet::ORIGINAL_FILE_USE.to_s)
-          return :extracted_text if type.to_s.casecmp?(Hyrax::FileSet::EXTRACTED_TEXT_USE.to_s)
-          return :thumbnail if type.to_s.casecmp?(Hyrax::FileSet::THUMBNAIL_USE.to_s)
+          return :original_file if type.to_s.casecmp?(Hyrax::FileMetadata::Use::ORIGINAL_FILE.to_s)
```
no-reply: these probably don't need `casecmp?` either? is `type` going to be an `RDF::URI`?

app/models/hyrax/file_metadata.rb
```diff
@@ -2,15 +2,45 @@
...
   class FileMetadata < Valkyrie::Resource
+    GENERIC_MIME_TYPE = 'application/octet-stream'
+
+    ##
+    # Constants for PCDM Use URIs; use these constants in place of hard-coded
+    # URIs in the `::Valkyrie::Vocab::PCDMUse` vocabulary.
+    module Use
+      ORIGINAL_FILE = ::Valkyrie::Vocab::PCDMUse.OriginalFile
+      EXTRACTED_TEXT = ::Valkyrie::Vocab::PCDMUse.ExtractedText
+      THUMBNAIL = ::Valkyrie::Vocab::PCDMUse.Thumbnail
```
elrayle: In looking back through code trying to track down why case compare was used, I noticed that this should be...

```
THUMBNAIL_IMAGE = ::Valkyrie::Vocab::PCDMUse.ThumbnailImage
```

I can fix that if you want since I introduced it.

lib/wings/services/file_metadata_builder.rb
```diff
@@ -39,18 +39,30 @@ def attach_file_metadata(file_metadata:, file_set:)
...
         saved_file_metadata
       end
 
+      ##
+      # @api private
+      # @return [Hyrax::FileMetadata, nil]
+      def current_original_file(file_set)
+        Hyrax.query_service
+          .custom_queries
+          .find_many_file_metadata_by_use(resource: file_set, use: Hyrax::FileMetadata::Use::ORIGINAL_FILE)
```
elrayle: Part of what I liked about the #original_file method was that it was simple and had the knowledge that there should be only one, so it was ok to grab the first one returned.  I'd hate to see this code block repeated in multiple places.  If it doesn't belong in FileSet (for single responsibility reasons), is there somewhere else it can live (to keep it DRY)?  And if not, should it be a service.  Getting the original file is something that happens a lot in Hyrax.

app/services/hyrax/custom_queries/find_file_metadata.rb
```diff
@@ -9,7 +9,8 @@ class FindFileMetadata
       def self.queries
         [:find_file_metadata_by,
          :find_file_metadata_by_alternate_identifier,
-         :find_many_file_metadata_by_ids]
+         :find_many_file_metadata_by_ids,
+         :find_many_file_metadata_by_use]
```
elrayle: That was a nice catch. I'm surprised the tests passed.  I thought queries had to be registered to be used through custom_queries.

lib/wings/hydra/works/services/add_file_to_file_set.rb
no-reply: yeah, agreed on all points. i'm thinking the issue of the symbols is out-of-scope for now.

lib/wings/services/file_metadata_builder.rb
no-reply: ~~I think the idea of a service is a good one. Maybe there should be a navigator?~~
There already is one. I'll look to make sure it works well and refactor around it.

app/services/hyrax/custom_queries/find_file_metadata.rb
no-reply: seems like they worked okay with the wings adapter, but not the memory adapter (?)

not sure why the wings one worked okay, but this is the first time this gets used with a different adapter.

app/models/hyrax/file_metadata.rb
no-reply: ah. i'll add a patch here.

app/services/hyrax/characterization/file_set_description.rb
```diff
@@ -0,0 +1,45 @@
...
+          Hyrax::FileMetadata.new
+      end
+
+      private
+
+        ##
+        # @api private
+        def queries
+          Hyrax.query_service.custom_queries
+        end
```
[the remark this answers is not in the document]
no-reply: :+1: to `Hyrax.custom_queries`

app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
```diff
@@ -27,7 +27,7 @@ def update
       filter_docs_with_edit_access!
       copy_visibility = []
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] } if params[:embargoes]
-      af_objects = Hyrax.query_service.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
+      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
```
elrayle: Excellent!  Glad to see Hyrax.custom_queries.

app/services/hyrax/characterization/file_set_description.rb
```diff
@@ -0,0 +1,45 @@
...
+    # @api public
+    class FileSetDescription
+      include Hydra::Works::MimeTypes
+
+      ##
+      # @!attribute [rw] file_set
+      #   @return [Hyrax::FileSet]
+      attr_accessor :file_set
+
+      delegate :mime_type, to: :primary_file
```
elrayle: Can you expand on the reason for changing from the concept of original_file to primary_file?

app/services/hyrax/characterization/file_set_description.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      ##
+      # @param [Hyrax::FileSet] file_set
+      # @param [Symbol] primary_file  a symbol mapping to the file_set member
+      #   used for characterization
+      def initialize(file_set:, primary_file: :original_file)
+        self.file_set = file_set
+
+        @primary_file_type_uri =
+          Hyrax::FileMetadata::Use.uri_for(use: primary_file)
```
elrayle: In other places, the file type is sometimes passed as a symbol (e.g. :original_file, :extracted_file, :thumbnail_image), and sometimes passed as a URI for custom types.  Do you think that might happen here too?  FileActor which calls Hydra::Works::AddFileToFileSet handles both symbol and URI.

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -100,29 +100,28 @@ def normalize_relation(relation)
...
+          when Hyrax::FileMetadata::Use::EXTRACTED_TEXT
+            :extracted_file
+          when Hyrax::FileMetadata::Use::THUMBNAIL
+            :thumbnail_file
```
elrayle: As I was exploring inconsistencies documented in Issue #4228, I noticed that these are using symbols that I think are incorrect.  I'm pretty sure these should be :extracted_text and :thumbnail.

This part of why I propose putting any conversions between symbol and RDF::URI in a single place (e.g. FileMetadata::Use module).

no-reply: i also noticed that the symbols were inconsistent across modules. i'm not sure that means they are incorrect(?). these are the symbols that pre-date this PR, right?

unless they got mixed up somewhere else recently, they should be the ones callers are already passing in.

in any case, i do think we should hold changes to behavior around these symbols for resolution as part of #4228. for now, the behavior exists in two places because it behaves slightly differently across them, and the goal here is to preserve that faithfully.

lib/wings/services/file_metadata_builder.rb
no-reply: can we hold this for a later refactor? i promise to get to it this week, if so.

elrayle: I am good with putting this in another PR.  Created an issue to track this...  #4233

app/actors/hyrax/actors/file_actor.rb
elrayle: It looks like these were incorrectly entered in PR #2746 which was one of the first PRs for valkyrizing files.  I don't think they have surfaced as a problem because Hyrax generally already passes relation as a Symbol unless it is a custom relation, so the conversion probably has never been used in a Hyrax app.

elrayle: Looking back on PR #2746 for this part handling the conversion from URI to Symbol and vice versa, I am somewhat uncomfortable with the default being original file.  I'm wondering what happens if a custom file type is passed in for relation.

I'm can update #4228 with this question or create a separate issue.  It seems a bit beyond the refactoring for consistency issue.

app/services/hyrax/characterization/file_set_description.rb
elrayle: @no-reply This is an open question for me.

no-reply: i'm not sure i understand the question. this is a new interface, so callers will presumably use it as defined.

app/actors/hyrax/actors/file_actor.rb
no-reply: re: `:original_file` i think this is just a straight refactor, and i'd prefer not to let its scope creep.

i'm in favor of wrapping any remaining issues into #4228.

app/services/hyrax/characterization/file_set_description.rb
elrayle: Custom types do not have a Symbol equivalent.  So what happens here if someone wants to pass http://pcdm.org/use#remastered?

app/actors/hyrax/actors/file_actor.rb
elrayle: Done... #4228 is updated.

app/services/hyrax/characterization/file_set_description.rb
elrayle: Based on recommendations in Issue #4228 addressing inconsistencies, I would suggest that this new interface accept RDF::URI for primary_file.  And I would recommend renaming primary_file to primary_use to be consistent with the naming recommendations in #4228.

jeremyf:
👍

PR #4217, opened 2020-01-25
Use the configured storage adapter

no-reply:
We weren't respecting storage adapter configuration for valkyrie! This fixes that.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -100,11 +100,11 @@ module Hyrax
     end
 
     def valid?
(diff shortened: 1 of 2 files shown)


no-reply:
@elrayle this fixes an issue in `Hyrax::FileMetadata`, but i think there's a bigger problem, which is that `#file` doesn't seem to mean anything at all(?).

it looks like this method, and `#valid?` are with respect to an arbitrary id from `#file_identifiers`? can these just be removed? or should `#file_identifiers` be replaced with a singular `#file`?

PR #4218, opened 2020-01-25
Index listener

geekscruff:
Adds `Hyrax.publisher.subscribe(Hyrax::Listeners::MetadataIndexListener.new)` to `config/initializers/listeners.rb`

Adds publishing an `object.metadata.updated` event to the `after_create_concern` callback.

### DIFF
--- a/config/initializers/listeners.rb
+++ b/config/initializers/listeners.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 Hyrax.publisher.subscribe(Hyrax::Listeners::AclIndexListener.new)
+Hyrax.publisher.subscribe(Hyrax::Listeners::MetadataIndexListener.new)
 Hyrax.publisher.subscribe(Hyrax::Listeners::BatchNotificationListener.new)
 Hyrax.publisher.subscribe(Hyrax::Listeners::ObjectLifecycleListener.new)
 Hyrax.publisher.subscribe(Hyrax::Listeners::FileSetLifecycleListener.new)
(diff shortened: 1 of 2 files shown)


app/services/hyrax/listeners/metadata_index_listener.rb
```diff
@@ -16,7 +16,17 @@ class MetadataIndexListener
...
-        Hyrax.index_adapter.save(resource: event[:object])
+        Hyrax.index_adapter.save(resource:
+          Hyrax::ValkyrieIndexer.for(resource: event[:object]).resource)
+      end
+
+      ##
+      # Index the resource.
+      #
+      # @param event [Dry::Event]
+      def on_object_metadata_created(event)
```
no-reply: is `created` a subtype of `updated` for the purpose of indexing?
i think i'd expect the application to publish metadata updated events in both cases, and that the indexer should only need to listen for updates in general, since it's behavior is the same for created cases? thoughts?

app/services/hyrax/listeners/metadata_index_listener.rb
```diff
@@ -16,7 +16,17 @@ class MetadataIndexListener
       #
       # @param event [Dry::Event]
       def on_object_metadata_updated(event)
-        Hyrax.index_adapter.save(resource: event[:object])
+        Hyrax.index_adapter.save(resource:
+          Hyrax::ValkyrieIndexer.for(resource: event[:object]).resource)
```
no-reply: i'm not following this. the index adapter does this factory lookup, right? the client should only need to pass in the resource.

i feel like i might be missing something :)

geekscruff: In my console testing it seemed like you had to pass the resource like this, but I'm not 100% on that so let me do more testing.

app/services/hyrax/listeners/metadata_index_listener.rb
geekscruff: I did wonder the same, and wanted feedback. I'll refactor the new method out and setup the calls to update.

geekscruff:
@no-reply I backed out the additional method, and passing the Indexer to the adapter. The problem with the latter that I was seeing in the console is the same issue with the `ValykrieIndexer.for` not retrieving the right Indexer that I'm hoping to fix in #4221.

PR #4219, opened 2020-01-25
Add indexer to work_resource generator

geekscruff:
Adds an indexer to the work_resource generator, eg. `MonographIndexer`. The indexer is configured to register the resource-specific Indexer, and adds a commented out example `to_solr`. Registers the new indexer in `config/initializers/indexers.rb`.

Corrects `rails generate hyrax:work` to `rails generate hyrax:work_resource` in other templates.

Adds `config/initializers/indexers.rb` to the install generator (via hyrax:config generator).

Adds a test to ensure the correct Indexer is returned by `Hyrax::ValkyrieIndexer#for` for a generated resource.


geekscruff:
Test failure seems unrelated to my change, and only on one of the circles.

geekscruff:
I *think* the failure here is because I need to create a new .internal_test_app (circle orb?). Help?!

no-reply:
@geekscruff you need to bump the number in `.regen`.

this is our current, slightly hacky---suggestions welcome!---way to handle changes that require bumps for reasons not detected by the circle config.

jeremyf:
I pushed the `.regen` bump to get this going.

lib/generators/hyrax/work_resource/work_resource_generator.rb
```diff
@@ -33,6 +33,18 @@ def create_model_spec
...
   end
 
+  def create_indexer
+    template('indexer.rb.erb', File.join('app/indexers/', class_path, "#{file_name}_indexer.rb"))
+  end
+
+  def register_indexer
+    config = 'config/initializers/indexers.rb'
+    register_line = "  Hyrax::ValkyrieIndexer.register #{class_name}Indexer, as_indexer_for: #{class_name}\n"
+    inject_into_file config, after: "# Register Indexers\n" do # rubocop:disable Style/MultilineIfModifier
```
no-reply: could avoid the rubocop exception with a guard clause?

```ruby
return if File.read(config).include?(register_line)
inject_into_file config, after: "# Register Indexers\n" { register_line }
```

???
```

geekscruff: Oh that should work, I tried various permutations that all had different rubocop errors!

PR #4221, opened 2020-01-25
Valkyrie Indexer Setup and Specs

geekscruff:
Note: includes a failing test that need a fix (see below).

Indexer:
* Add :basic_metadata Indexer to `ValkyrieWorkIndexer`
* Add indexing rules to `config/metadata/basic_metadata.yml`
* Add a basic spec for a generated Indexer (`MonographIndexer`)

ActiveFedoraConverter
* Fixes issue with creating a DefaultWork for Valkyrie native resources (ie. those generated via the generator)
* Specs updated for ModelTransformer, ActiveFedoraConverter and Persister
* This persister test now fails and needs fixing: `Wings::Valkyrie::Persister When passing a Valkyrie::Resource that was never an ActiveFedora::Base can save nested resources` with `undefined method `nested_resource_attributes=' for #<Wings::CustomResource::GeneratedResourceSchema:xxxx>`


no-reply:
this is looking great! thanks for pushing this forward! i'll give a closer review when we get #4219 in and rebase.

straleyb:
Since #4219 is in now, im going to go ahead and publish this PR. Its ready for review.

geekscruff:
I expect one or two tests to fail, so next step is to fix 'em.

geekscruff:
@no-reply I did move the indexer registration to an initializer in the generate indexer PR. I left the statement in the individual indexers for clarity, even though they aren't strictly necessary now. 

So, the failure relating to loading the indexer has been fixed, but the current failure is related to the refactoring of `ActiveFedoraConverter` so I/we do need to have a look at that.

no-reply:
@geekscruff thanks. i'm taking a look

no-reply:
The issue is to do with inheritance of class variables, which impacts `accepts_nested_attributes_for`. i'm working on approaches, but may also just end up writing `DefaultWork` out of existence. it's probably time for that :)

no-reply:
seems connected to https://github.com/samvera/active_fedora/issues/1343

straleyb:
Well, I rebased this onto master and it seems like it picked up on a whole slew of stuff. Not sure why that happened exactly. I can push a new one if needed though since I borked it-ish.

PR #4222, opened 2020-01-27
Removes configurability for resource indexer in solr indexing adapter

straleyb:

This PR pushes up some changes that had been discussed at Solar Vortex. There is not much of a reason to have a configurable indexer for the Indexing adapters. This is for the fact that indexing adapters will be the piece that is created and registered. So if someone wishes to use a new resource indexer in a specific adapter, they would either swap it or register a new indexer. 

There are no tests that tested the configurability of the adapter as it was a way to create an ambiguous enough interface for testing purposes. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/valkyrie/indexing/solr/indexing_adapter.rb
+++ b/lib/valkyrie/indexing/solr/indexing_adapter.rb
@@ -9,7 +9,7 @@ module Valkyrie
         ##
         # @!attribute [r] connection
         #   @return [RSolr::Client]
-        attr_reader :connection, :resource_indexer
+        attr_reader :connection
 
         ##
         # @param connection [RSolr::Client] The RSolr connection to index to.
--- a/lib/valkyrie/indexing/solr/indexing_adapter.rb
+++ b/lib/valkyrie/indexing/solr/indexing_adapter.rb
@@ -9,17 +9,13 @@ module Valkyrie
         ##
         # @!attribute [r] connection
         #   @return [RSolr::Client]
-        # @!attribute [r] resource_indexer
-        #   @return [Class]
         attr_reader :connection, :resource_indexer
 
         ##
         # @param connection [RSolr::Client] The RSolr connection to index to.
-        # @param resource_indexer [Class, #to_solr] An indexer which is able to
-        #   receive a `resource` argument and then has an instance method `#to_solr`
-        def initialize(connection: default_connection, resource_indexer: default_indexer)
+        def initialize(connection: default_connection)
           @connection = connection
-          @resource_indexer = resource_indexer
+          @resource_indexer = default_indexer
         end
 
         def save(resource:)
--- a/lib/valkyrie/indexing/solr/indexing_adapter.rb
+++ b/lib/valkyrie/indexing/solr/indexing_adapter.rb
@@ -48,7 +48,7 @@ module Valkyrie
           end
 
           def solr_document(resource)
-            resource_indexer.for(resource: resource).to_solr
+            @resource_indexer.for(resource: resource).to_solr
           end
 
           def add_documents(documents)
--- a/lib/valkyrie/indexing/solr/indexing_adapter.rb
+++ b/lib/valkyrie/indexing/solr/indexing_adapter.rb
@@ -15,7 +15,6 @@ module Valkyrie
         # @param connection [RSolr::Client] The RSolr connection to index to.
         def initialize(connection: default_connection)
           @connection = connection
-          @resource_indexer = default_indexer
         end
 
         def save(resource:)
@@ -48,7 +47,7 @@ module Valkyrie
           end
 
           def solr_document(resource)
-            @resource_indexer.for(resource: resource).to_solr
+            resource_indexer.for(resource: resource).to_solr
           end
 
           def add_documents(documents)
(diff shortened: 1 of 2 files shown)


lib/valkyrie/indexing/solr/indexing_adapter.rb
```diff
@@ -9,17 +9,13 @@ class IndexingAdapter
...
 
         ##
         # @param connection [RSolr::Client] The RSolr connection to index to.
-        # @param resource_indexer [Class, #to_solr] An indexer which is able to
-        #   receive a `resource` argument and then has an instance method `#to_solr`
-        def initialize(connection: default_connection, resource_indexer: default_indexer)
+        def initialize(connection: default_connection)
           @connection = connection
-          @resource_indexer = resource_indexer
+          @resource_indexer = default_indexer
```
no-reply: looks like you either need to make the default `Hyrax::ValkyrieIndexer` or move the `.for` implementation somewhere that `NullIndexer` has a copy of the registry.

straleyb: That was sort of my question. I used `Hyrax::ValkyrieIndexer` as the default indexer here, but if we would prefer using the null indexer still, I can have it respond to the `.for` method and stub it to return the right indexer. 

That just depends on if we want this system as a whole to still utilize the null indexer or not.

straleyb:
Side question here. This test doesn't utilize these two lines. Was the goal of this to get it to use the persister to call save instead of call save on the adapter?

https://github.com/samvera/hyrax/blob/d646a36267bc8b7af9cf1b333917b39f57a6bad5/spec/valkyrie/indexing/solr/indexing_adapter_spec.rb#L7-L8

lib/valkyrie/indexing/solr/indexing_adapter.rb
no-reply: i think it's okay for us to use `Hyrax::Indexer` locally, and punt the `NullIndexer` question over to Valkyrie.

no-reply:
> Side question here. This test doesn't utilize these two lines. Was the goal of this to get it to use the persister to call save instead of call save on the adapter?
> 
> https://github.com/samvera/hyrax/blob/d646a36267bc8b7af9cf1b333917b39f57a6bad5/spec/valkyrie/indexing/solr/indexing_adapter_spec.rb#L7-L8

looks like cruft to me. introduced here: https://github.com/samvera/hyrax/commit/57450b50fd8058278be348efcefa1d61035a6d7b#diff-25d623799bda9453d7d1af9e8047932f

straleyb:
Not sure why ruby 2.5.5 failed and 2.4.7 is hanging. Ill keep an eye on this for any updates, but I think it should be good to go now.

straleyb:
@no-reply Should I run the tests again for the failed 2.5.5? The failures look unrelated.

lib/valkyrie/indexing/solr/indexing_adapter.rb
```diff
@@ -9,17 +9,13 @@ class IndexingAdapter
         ##
         # @!attribute [r] connection
         #   @return [RSolr::Client]
-        # @!attribute [r] resource_indexer
-        #   @return [Class]
         attr_reader :connection, :resource_indexer
```
no-reply:
```suggestion
        attr_reader :connection
```

straleyb:
@no-reply Got this back up to speed.

lib/valkyrie/indexing/solr/indexing_adapter.rb
```diff
@@ -9,17 +9,13 @@ class IndexingAdapter
...
 
         ##
         # @param connection [RSolr::Client] The RSolr connection to index to.
-        # @param resource_indexer [Class, #to_solr] An indexer which is able to
-        #   receive a `resource` argument and then has an instance method `#to_solr`
-        def initialize(connection: default_connection, resource_indexer: default_indexer)
+        def initialize(connection: default_connection)
           @connection = connection
-          @resource_indexer = resource_indexer
+          @resource_indexer = default_indexer
```
no-reply:
```suggestion
          @resource_indexer = default_indexer
```

lib/valkyrie/indexing/solr/indexing_adapter.rb
```diff
@@ -94,7 +90,7 @@ def default_connection
           end
 
           def default_indexer
```
no-reply:
```suggestion
          def resource_indexer
```

lib/valkyrie/indexing/solr/indexing_adapter.rb
```diff
@@ -52,7 +48,7 @@ def persist(resources)
           end
 
           def solr_document(resource)
-            resource_indexer.for(resource: resource).to_solr
+            @resource_indexer.for(resource: resource).to_solr
```
no-reply:
```suggestion
            resource_indexer.for(resource: resource).to_solr
```

straleyb:
@no-reply I updated it to use that method to return the indexer. Definitely more hyrax-y and doesn't keep an unnecessary instance variable around. Good call on this one.

PR #4223, opened 2020-01-27
Implement Valkyrie indexing for Collections

no-reply:
Add an indexer for valkyrie-based `Hyrax::PcdmCollection` and index "generic type" and "thumbnail".

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/thumbnail_path_service.rb
+++ b/app/services/hyrax/thumbnail_path_service.rb
@@ -32,7 +32,7 @@ module Hyrax
         # @return the network path to the thumbnail
         # @param [FileSet] thumbnail the object that is the thumbnail
         def thumbnail_path(thumbnail)
-          Hyrax::Engine.routes.url_helpers.download_path(tuhmbnail.id,
+          Hyrax::Engine.routes.url_helpers.download_path(thumbnail.id,
                                                          file: 'thumbnail')
         end
 
--- a/app/services/hyrax/thumbnail_path_service.rb
+++ b/app/services/hyrax/thumbnail_path_service.rb
@@ -1,10 +1,10 @@
 module Hyrax
   class ThumbnailPathService
     class << self
-      # @param [Work, FileSet] object - to get the thumbnail for
+      # @param [#id] object - to get the thumbnail for
       # @return [String] a path to the thumbnail
       def call(object)
-        return default_image unless object.thumbnail_id
+        return default_image unless object.try(:thumbnail_id)
 
         thumb = fetch_thumbnail(object)
 
@@ -32,7 +32,7 @@ module Hyrax
         # @return the network path to the thumbnail
         # @param [FileSet] thumbnail the object that is the thumbnail
         def thumbnail_path(thumbnail)
-          Hyrax::Engine.routes.url_helpers.download_path(thumbnail.id,
+          Hyrax::Engine.routes.url_helpers.download_path(tuhmbnail.id,
                                                          file: 'thumbnail')
         end
 
--- a/spec/search_builders/hyrax/collection_member_search_builder_spec.rb
+++ b/spec/search_builders/hyrax/collection_member_search_builder_spec.rb
@@ -1,52 +1,50 @@
 RSpec.describe Hyrax::CollectionMemberSearchBuilder do
+  subject(:builder) { described_class.new(scope: context, collection: collection, search_includes_models: include_models) }
   let(:context) { double("context", blacklight_config: CatalogController.blacklight_config) }
   let(:solr_params) { { fq: [] } }
   let(:include_models) { :both }
   let(:collection) { build(:collection_lw, id: '12345') }
-  let(:builder) { described_class.new(scope: context, collection: collection, search_includes_models: include_models) }
 
   describe ".default_processor_chain" do
     subject { builder.default_processor_chain }
 
     it { is_expected.to include :member_of_collection }
+    it { is_expected.to include :filter_models }
   end
 
   describe '#member_of_collection' do
-    let(:subject) { builder.member_of_collection(solr_params) }
-
     it 'updates solr_parameters[:fq]' do
+      expect { builder.member_of_collection(solr_params) }
+        .to change { solr_params[:fq] }
+        .to include("#{builder.collection_membership_field}:#{collection.id}")
       subject
-      expect(solr_params[:fq]).to include("#{builder.collection_membership_field}:#{collection.id}")
     end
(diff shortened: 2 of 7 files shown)


app/services/hyrax/thumbnail_path_service.rb
```diff
@@ -1,10 +1,10 @@
 module Hyrax
   class ThumbnailPathService
     class << self
-      # @param [Work, FileSet] object - to get the thumbnail for
+      # @param [#id] object - to get the thumbnail for
```
straleyb: Im wondering if this could be misleading. Its not an ID object right? and if its a method call you are referencing, you are calling #thumbnail_id. If Im understanding this right, this is a thumbnail object or a FileSet object right?

app/services/hyrax/thumbnail_path_service.rb
```diff
@@ -32,7 +32,7 @@ def fetch_thumbnail(object)
         # @return the network path to the thumbnail
         # @param [FileSet] thumbnail the object that is the thumbnail
         def thumbnail_path(thumbnail)
-          Hyrax::Engine.routes.url_helpers.download_path(thumbnail.id,
+          Hyrax::Engine.routes.url_helpers.download_path(tuhmbnail.id,
```
straleyb: Typo

spec/search_builders/hyrax/collection_member_search_builder_spec.rb
```diff
@@ -1,52 +1,89 @@
...
+          .to include("#{builder.collection_membership_field}:#{collection.id}")
+      end
     end
-  end
 
-  describe '#models' do
-    let(:work_classes) { [GenericWork] }
-    let(:collection_classes) { [Collection] }
-    let(:subject) { builder.models }
+    describe '#filter_models' do
```
straleyb: Somehow I feel like these names are backwards. I feel like #filter_models would filter out models and #include_models would include the models in the search. Im not sure if this was the way the interface was defined in hyrax originally, but something that stuck out here.

spec/services/hyrax/thumbnail_path_service_spec.rb
```diff
@@ -3,6 +3,35 @@
 
   subject { described_class.call(object) }
 
+  context 'with a Valkyrie resource' do
+    subject(:service) { described_class }
+    let(:object) { FactoryBot.build(:hyrax_resource) }
```
straleyb: Should this be changed at some time to be ':valkyrie_resource` and the other to be `:af_resource`? Nothing needs to be changed here, just an idea for clarity.

spec/services/hyrax/thumbnail_path_service_spec.rb
```diff
@@ -3,6 +3,35 @@
...
+
+    describe '#call' do
+      it 'returns the default image path' do
+        expect(service.call(object)).to include('/assets/default', '.png')
+      end
+
+      context 'when it is a Work with a thumbnail_id' do
+        it 'returns the thumbnail path for the file_set'
```
straleyb: missing do block?

spec/services/hyrax/thumbnail_path_service_spec.rb
```diff
@@ -3,6 +3,35 @@
...
+        let(:object) { build(:hyrax_file_set) }
+
+        it 'returns the default image path' do
+          expect(service.call(object)).to include('/assets/default', '.png')
+        end
+
+        context 'when it has a thumbnail id' do
+          let(:object) { build(:hyrax_file_set, thumbnail_id: '123') }
+
+          it 'returns the thumbnail file path'
```
straleyb: missing do block

straleyb:
I left some comments here if anyone takes a peek at this. Im also going to add an in progress tag, so people who look at it know its not ready yet. Im not sure this is deviating, but ill add [WIP] on this too since there is still some stuff going on with it.

app/services/hyrax/thumbnail_path_service.rb
no-reply: the work of these changes is to ensure that the only interface for the object passed in is an `#id` instance method. in YARD convention, that's what this means. in the case of `#thumbnail_id`, this PR changes the service to give the default thumbnail path if it's not implemented.

in practice, what we plan to pass in is any ActiveFedora::Base or any Valkyrie::Resource object. but again, the goal of this change is to codify the contract of this service as requiring `#id` only.

app/services/hyrax/thumbnail_path_service.rb
no-reply:
```suggestion
          Hyrax::Engine.routes.url_helpers.download_path(thumbnail.id,
```

spec/services/hyrax/thumbnail_path_service_spec.rb
no-reply: nope. this is just a pending test

spec/services/hyrax/thumbnail_path_service_spec.rb
no-reply: another pending test

spec/services/hyrax/thumbnail_path_service_spec.rb
no-reply: i think the naming is aimed at being resilient to the eventual removal of AF/Wings. `:hyrax_resource` seems like a reasonable name for the factory that builds `Hyrax::Resource` objects? i would expect `:valkyrie_resource` to build a `Valkyrie::Resource`, i think

no-reply:
@straleyb normally we use `[WIP]` to indicate PR's that might not be ready for review. is it your opinion that that applies here?

app/services/hyrax/thumbnail_path_service.rb
no-reply: fixed, thanks!

no-reply:
ah yeah, it looks like i'm still stuck on some of the shared examples. I'll poke at this for some more, and let's leave the `[WIP]` in place.

thanks!

straleyb:
@no-reply Thanks for filling me in on some of these conventions. I didnt know that `it 'things and stuff happen'` was a valid convention for a pending test. Threw me off seeing that and made me think some `block not given` error would pop up or something. 

Once this has been updated, whether I take a stab at getting those tests to pass or you, this looks good to get in.

no-reply:
@straleyb i think this should pass now

straleyb:
Looks pretty index-y to me! :merge:

PR #4225, opened 2020-01-27
handle conversion of custom use types for files

### LINKED ISSUE #4203: Wings: Round trip custom files in filesets
### Descriptive summary

Support fileset files with custom associations (i.e. associations other than original_file, extracted_text, and thumbnail).  The filesets with these files should be able to round trip from ActiveFedora to Valkyrie Resource and back.

### Rationale

Maintain support for custom associations between a fileset and files.

### Expected behavior

Whether a fileset is created directly as a Valkyrie::Resource or an AF object, it should allow for adding files with relationships other than original file, extracted text and thumbnail, such as, http://pcdm.org/use#remastered.

### Actual behavior

Not sure.

#### app/models/hyrax/file_metadata.rb #1193 

```
file.use = RDF::URI.new('http://pcdm.org/use#remastered')
file_metadata = Hyrax::FileMetadata.for(file: fixture)
file_metadata.use # should return <RDF::URI 'http://pcdm.org/use#remastered'>
```

#### app/wings/hydra/works/service/services/add_file_to_file_set.rb

Should allow `type:` to be passed in as `RDF::URI.new('http://pcdm.org/use#remastered')`

Tested in spec/wings/hydra/works/services/add_file_to_file_set_spec.rb at `context 'when :use is NOT the name of an association type'`

#### app/wings/active_fedora_converter.rb #convert_files 

#convert_files currently only supports conversion of files with use original file, extracted text, and thumbnail.  Needs to be expanded to support all file associations.  The use is stored in the FileMetadata resource in the use attribute.  The FileSet has the list of files.

#### app/actors/hyrax/actors/file_actor.rb #normalize_relation_for_valkyrie

Appears to only handle original_file, extracted_text, and thumbnail_file.  

The relation tests in file_actor_spec.rb when processing using valkyrie are marked `pending 'implementation of Wings::Valkyrie::Persister #save_file_metadata'`.  I'm not sure that still applies. 

The first test starting at `context 'relation'` (~ln 149) uses relation `http://pcdm.org/use#remastered` which makes it a good test of this functionality.
 

### Related work

Issue #4060 - added support for standard uses of original_file, extracted_text, and thumbnail (PR #4055)

elrayle:
NOTE: PR #4215 is working toward similar issues around use types.  Both approaches need to be reviewed for compatibility.

----

Fixes #4203

Co-authored-by: cjcolvar <cjcolvard@indiana.edu>

### Conversion of files with custom use type

Conversion of the 3 primary use types were handled in PR #4055. That made use of the AF association target.   That does not work for custom file.  This PR includes a generalized approach that works for all relationships.

### Refactor:  type instead of use

There is no conversion path from `pcdm_file.metadata_node.type` as used in AF to `fiile_metadata.use`  in FileMetadata.  

metadata_node.type is an ActiveTriples::Relation that includes all RDF types for the file, not just the use type.  For example...

```
 [#<RDF::URI:0x3fc9c096fe3c URI:http://fedora.info/definitions/v4/repository#Binary>,
  #<RDF::URI:0x3fc9c096f938 URI:http://fedora.info/definitions/v4/repository#Resource>, 
  #<RDF::URI:0x3fc9c096f234 URI:http://pcdm.org/models#File>, 
  #<RDF::URI:0x3fc9c096fa04 URI:http://www.w3.org/ns/ldp#NonRDFSource>,
  #<RDF::URI:0x3fc9c096fb83 URI:http://pcdm.org/use#OriginalFile>]
```

The only part of this that is the use is the last one ending in `#OriginalFile`.  We could have search the types for one beginning with `http://pcdm.org/use`, but there is not a guarantee that a site wouldn't have custom uses defined that do not begin with this.

For backward compatibility, we decided to remove the `use` attribute from FileMetadata and replace it with `type`.

To support functionality like #original_file, we added a custom query `find_many_file_metadata_by_use(resource:, use:)` to FindFileMetadata.  Then we defined methods #original_file, #extracted_text, and #thumbnail in Hyrax::FileSet using the new custom query to locate the file with the corresponding use.

### Refactor: Centralize definition of 3 primary file use types

In making changes to for this PR, we noticed that there were differing URIs used for the 3 primary file use types  (i.e., original_file, extracted_text, and thumbnail).  To address this, we centralized the definition of the primary uses in Hyrax::FileSet.  All references to the associated use URIs were updated to use Hyrax::FileSet.original_file_use, Hyrax::FileSet.extracted_text_use, and Hyrax::FileSet.thumbnail.

### DIFF
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -103,10 +103,10 @@ module Hyrax
           return relation if relation.is_a? Symbol
           return relation.to_sym if relation.respond_to? :to_sym
 
-          # TODO: whereever these are set, they should use Valkyrie::Vocab::PCDMUse... making the casecmp unnecessary
-          return :original_file if relation.to_s.casecmp(Valkyrie::Vocab::PCDMUse.original_file.to_s)
-          return :extracted_file if relation.to_s.casecmp(Valkyrie::Vocab::PCDMUse.extracted_file.to_s)
-          return :thumbnail_file if relation.to_s.casecmp(Valkyrie::Vocab::PCDMUse.thumbnail_file.to_s)
+          # TODO: whereever these are set, they should use FileSet.*_use... making the casecmp unnecessary
+          return :original_file if relation.to_s.casecmp(Hyrax::FileSet.original_file_use.to_s)
+          return :extracted_file if relation.to_s.casecmp(Hyrax::FileSet.extracted_text_use.to_s)
+          return :thumbnail_file if relation.to_s.casecmp(Hyrax::FileSet.thumbnail_use.to_s)
           :original_file
         end
 
@@ -114,10 +114,10 @@ module Hyrax
           # TODO: When this is fully switched to valkyrie, this should probably be removed and relation should always be passed
           #       in as a valid URI already set to the file's use
           relation = relation.to_s.to_sym
-          return Valkyrie::Vocab::PCDMUse.original_file if relation == :original_file
-          return Valkyrie::Vocab::PCDMUse.extracted_file if relation == :extracted_file
-          return Valkyrie::Vocab::PCDMUse.thumbnail_file if relation == :thumbnail_file
-          Valkyrie::Vocab::PCDMUse.original_file
+          return Hyrax::FileSet.original_file_use if relation == :original_file
+          return FileSet.extracted_file_use if relation == :extracted_file
+          return FileSet.thumbnail_file_use if relation == :thumbnail_file
+          Hyrax::FileSet.original_file_use
         end
     end
   end
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -113,11 +113,16 @@ module Hyrax
         def normalize_relation_for_valkyrie(relation)
           # TODO: When this is fully switched to valkyrie, this should probably be removed and relation should always be passed
           #       in as a valid URI already set to the file's use
-          relation = relation.to_s.to_sym
-          return Hyrax::FileSet::ORIGINAL_FILE_USE if relation == :original_file
-          return FileSet.extracted_file_use if relation == :extracted_file
-          return FileSet.thumbnail_file_use if relation == :thumbnail_file
-          Hyrax::FileSet::ORIGINAL_FILE_USE
+          case relation.to_s.to_sym
+          when :original_file
+            Hyrax::FileSet::ORIGINAL_FILE_USE
+          when :extracted_file
+            Hyrax::FileSet.EXTRACTED_TEXT_USE
+          when :thumbnail_file
+            Hyrax::FileSet::THUMBNAIL_USE
+          else
+            Hyrax::FileSet::ORIGINAL_FILE_USE
+          end
         end
     end
   end
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -104,9 +104,9 @@ module Hyrax
           return relation.to_sym if relation.respond_to? :to_sym
 
           # TODO: whereever these are set, they should use FileSet.*_use... making the casecmp unnecessary
-          return :original_file if relation.to_s.casecmp(Hyrax::FileSet.original_file_use.to_s)
-          return :extracted_file if relation.to_s.casecmp(Hyrax::FileSet.extracted_text_use.to_s)
-          return :thumbnail_file if relation.to_s.casecmp(Hyrax::FileSet.thumbnail_use.to_s)
+          return :original_file if relation.to_s.casecmp(Hyrax::FileSet::ORIGINAL_FILE_USE.to_s)
+          return :extracted_file if relation.to_s.casecmp(Hyrax::FileSet::EXTRACTED_TEXT_USE.to_s)
+          return :thumbnail_file if relation.to_s.casecmp(Hyrax::FileSet::THUMBNAIL_USE.to_s)
           :original_file
         end
 
@@ -114,10 +114,10 @@ module Hyrax
           # TODO: When this is fully switched to valkyrie, this should probably be removed and relation should always be passed
           #       in as a valid URI already set to the file's use
           relation = relation.to_s.to_sym
-          return Hyrax::FileSet.original_file_use if relation == :original_file
+          return Hyrax::FileSet::ORIGINAL_FILE_USE if relation == :original_file
           return FileSet.extracted_file_use if relation == :extracted_file
           return FileSet.thumbnail_file_use if relation == :thumbnail_file
-          Hyrax::FileSet.original_file_use
+          Hyrax::FileSet::ORIGINAL_FILE_USE
         end
     end
   end
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -77,7 +77,7 @@ module Hyrax
         def perform_ingest_file_through_valkyrie(io)
           # Skip versioning because versions will be minted by VersionCommitter as necessary during save_characterize_and_record_committer.
           unsaved_file_metadata = io.to_file_metadata
-          unsaved_file_metadata.use = relation
+          unsaved_file_metadata.type = [relation]
           begin
             saved_file_metadata = file_metadata_builder.create(io_wrapper: io, file_metadata: unsaved_file_metadata, file_set: file_set)
           rescue StandardError => e # Handle error persisting file metadata
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -76,19 +76,19 @@ module Hyrax
       new(label: file.original_filename,
           original_filename: file.original_filename,
           mime_type: file.content_type,
-          use: file.try(:use) || [::Valkyrie::Vocab::PCDMUse.OriginalFile])
+          use: file.try(:use) || [Hyrax::FileSet.original_file_use])
     end
 
     def original_file?
-      use.include?(::Valkyrie::Vocab::PCDMUse.OriginalFile)
+      use.include?(Hyrax::FileSet.original_file_use)
     end
 
     def thumbnail_file?
-      use.include?(::Valkyrie::Vocab::PCDMUse.ThumbnailImage)
+      use.include?(Hyrax::FileSet.thumbnail_use)
     end
 
     def extracted_file?
-      use.include?(::Valkyrie::Vocab::PCDMUse.ExtractedImage)
+      use.include?(Hyrax::FileSet.extracted_text_use)
     end
 
     def title
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -76,19 +76,19 @@ module Hyrax
       new(label: file.original_filename,
           original_filename: file.original_filename,
           mime_type: file.content_type,
-          type: file.try(:type) || [Hyrax::FileSet.original_file_use])
+          type: file.try(:type) || [Hyrax::FileSet::ORIGINAL_FILE_USE])
     end
 
     def original_file?
-      type.include?(Hyrax::FileSet.original_file_use)
+      type.include?(Hyrax::FileSet::ORIGINAL_FILE_USE)
     end
 
     def thumbnail_file?
-      type.include?(Hyrax::FileSet.thumbnail_use)
+      type.include?(Hyrax::FileSet::THUMBNAIL_USE)
     end
 
     def extracted_file?
-      type.include?(Hyrax::FileSet.extracted_text_use)
+      type.include?(Hyrax::FileSet::EXTRACTED_TEXT_USE)
     end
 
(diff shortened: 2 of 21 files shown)


lib/wings/services/custom_queries/find_file_metadata.rb
```diff
@@ -68,6 +70,23 @@ def find_many_file_metadata_by_ids(ids:, use_valkyrie: true)
...
+      # @param use_valkyrie [boolean] defaults to true; optionally return ActiveFedora::File objects if false
+      # @return [Array<Hyrax::FileMetadata, Hydra::PCDM::File>] or empty array if there are no files with the requested use
+      # @example
+      #   Hyrax.query_service.find_file_metadata_by_use(use: ::RDF::URI("http://pcdm.org/ExtractedText"))
+      def find_many_file_metadata_by_use(resource:, use:, use_valkyrie: true)
+        pcdm_files = find_many_file_metadata_by_ids(ids: resource.file_ids, use_valkyrie: false)
+        pcdm_files.select! do |pcdm_file|
+          pcdm_file.metadata_node.type.include?(use)
+        end
+        # pcdm_files.select { |pcdm_file| pcdm_file.metadata_node.type.include?(use) }
```
straleyb: Does this need to stick around?

straleyb:
I looked through this and it looked good to me. Only thing I could find was there maybe a different way to populate an array of values possibly using #each_with_object and populating the new object with the files. But thats more of a style like preference. 

https://github.com/samvera/hyrax/pull/4225/files#diff-01ac1642e5ca1afd32dd5eef2a819d22R142

Thats the line I was looking at. If someone else looks at it and finds anything, otherwise I say good to go after my one comment is addressed.

lib/wings/active_fedora_converter.rb
```diff
@@ -120,51 +120,38 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
...
       end
 
       # Normalizes the attributes parsed from the resource
       #   (This ensures that scalar values are passed to the constructor for the
       #   ActiveFedora::Base Class)
       # @return [Hash]
       def normal_attributes
-        normalized = {}
-        attributes.each_pair do |attr, value|
+        attributes.each_with_object(P{}) do |(attr, value), hash|
```
straleyb: I think this is a typo P here which threw off the linting and some tests.

no-reply:
> WIP Needs reconciliation of approaches with PR #4215
let's not block on this. rather i can rebase #4215 on this and we can work to resolve any differences in architecture of approach at that point

lib/wings/active_fedora_converter.rb
```diff
@@ -120,54 +120,38 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
...
```
no-reply:
```suggestion
        attributes.each_with_object({}) do |(attr, value), hash|
```

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -103,21 +103,21 @@ def normalize_relation_for_active_fedora(relation)
...
 
         def normalize_relation_for_valkyrie(relation)
           # TODO: When this is fully switched to valkyrie, this should probably be removed and relation should always be passed
           #       in as a valid URI already set to the file's use
           relation = relation.to_s.to_sym
-          return Valkyrie::Vocab::PCDMUse.original_file if relation == :original_file
-          return Valkyrie::Vocab::PCDMUse.extracted_file if relation == :extracted_file
-          return Valkyrie::Vocab::PCDMUse.thumbnail_file if relation == :thumbnail_file
-          Valkyrie::Vocab::PCDMUse.original_file
+          return Hyrax::FileSet.original_file_use if relation == :original_file
```
no-reply: would switch statements be more readable here?

```ruby
case relation
when :original_file
  Valkyrie::Vocab::PCDMUse.original_file
#...
end
```

app/models/hyrax/file_metadata.rb
```diff
@@ -10,7 +10,7 @@ class FileMetadata < Valkyrie::Resource
     attribute :label, ::Valkyrie::Types::Set
     attribute :original_filename, ::Valkyrie::Types::Set
     attribute :mime_type, ::Valkyrie::Types::Set
-    attribute :use, ::Valkyrie::Types::Set # AF::File type
+    attribute :type, ::Valkyrie::Types::Set # AF::File type
```
no-reply: :+1: i like this rename.

app/models/hyrax/file_metadata.rb
```diff
@@ -76,19 +76,19 @@ def self.for(file:)
       new(label: file.original_filename,
           original_filename: file.original_filename,
           mime_type: file.content_type,
-          use: file.try(:use) || [::Valkyrie::Vocab::PCDMUse.OriginalFile])
+          type: file.try(:type) || [Hyrax::FileSet.original_file_use])
```
no-reply: could we use a Dry::Struct/Types default here? something like `attribute :type, ::Valkyrie::Types::Set, default: [Hyrax::FileSet.original_file_use]`?

app/models/hyrax/file_set.rb
```diff
@@ -24,5 +24,54 @@ def pcdm_object?
...
+    # @return [FileMetadata] the FileMetadata resource of the extracted text
+    def extracted_text
+      Hyrax.query_service.custom_queries.find_many_file_metadata_by_use(resource: self, use: Hyrax::FileSet.extracted_text_use).first
+    end
+
+    ##
+    # Gives file metadata for the file filling the http://pcdm.org/Thumbnail use
+    # @return [FileMetadata] the FileMetadata resource of the thumbnail
+    def thumbnail
+      Hyrax.query_service.custom_queries.find_many_file_metadata_by_use(resource: self, use: Hyrax::FileSet.thumbnail_use).first
```
no-reply: i'm interested in refactoring to reduce duplication, but :+1: to moving forward with this as written.

lib/wings/active_fedora_converter.rb
```diff
@@ -120,54 +120,38 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
       def convert_members(af_object)
         return unless resource.respond_to?(:member_ids) && resource.member_ids
         # TODO: It would be better to find a way to add the members without resuming all the member AF objects
-        ordered_members = []
-        resource.member_ids.each do |valkyrie_id|
-          ordered_members << ActiveFedora::Base.find(valkyrie_id.id)
-        end
-        af_object.ordered_members = ordered_members
+        af_object.ordered_members = resource.member_ids.each_with_object([]) { |valkyrie_id, arr| arr << ActiveFedora::Base.find(valkyrie_id.id) }
```
no-reply:
```suggestion
        af_object.ordered_members = resource.member_ids.each_with_object([]) do |valkyrie_id, arr|
          arr << ActiveFedora::Base.find(valkyrie_id.id)
        end
```

or maybe just a `.map { ActiveFedora::Base.find(valkyrie_id.id) }`??

lib/wings/active_fedora_converter.rb
```diff
@@ -120,54 +120,38 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
...
 
       def convert_member_of_collections(af_object)
         return unless resource.respond_to?(:member_of_collection_ids) && resource.member_of_collection_ids
         # TODO: It would be better to find a way to set the parent collections without resuming all the collection AF objects
-        member_of_collections = []
-        resource.member_of_collection_ids.each do |valkyrie_id|
-          member_of_collections << ActiveFedora::Base.find(valkyrie_id.id)
-        end
-        af_object.member_of_collections = member_of_collections
+        af_object.member_of_collections = resource.member_of_collection_ids.each_with_object([]) { |valkyrie_id, arr| arr << ActiveFedora::Base.find(valkyrie_id.id) }
```
no-reply:
```suggestion
        af_object.member_of_collections = resource.member_of_collection_ids.each_with_object([]) do |valkyrie_id, arr|
          arr << ActiveFedora::Base.find(valkyrie_id.id)
        end
```

lib/wings/active_fedora_converter.rb
```diff
@@ -120,54 +120,38 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
...
-      def convert_file(af_object, relation)
-        resource_relation = "#{relation}_ids".to_sym
-        related_file_ids = resource.try(resource_relation)
-        return unless related_file_ids.present?
-        pcdm_file = Hydra::PCDM::File.new(related_file_ids.first.to_s)
-        af_object.association(relation).target = pcdm_file
+        return unless resource.respond_to? :file_ids
+        af_object.files = resource.file_ids.each_with_object([]) do |fid, arr|
+          pcdm_file = Hydra::PCDM::File.new(fid.id)
+          af_object.association(:original_file).target = pcdm_file if pcdm_file.metadata_node.type.include? RDF::URI.new('http://pcdm.org/use#OriginalFile')
```
no-reply: maybe a switch again here?

lib/wings/active_fedora_converter.rb
```diff
@@ -120,54 +120,38 @@ def initialize(uri = RDF::Node.new, _parent = ActiveTriples::Resource.new)
...
-      end
-
-      def convert_file(af_object, relation)
-        resource_relation = "#{relation}_ids".to_sym
-        related_file_ids = resource.try(resource_relation)
-        return unless related_file_ids.present?
-        pcdm_file = Hydra::PCDM::File.new(related_file_ids.first.to_s)
-        af_object.association(relation).target = pcdm_file
+        return unless resource.respond_to? :file_ids
+        af_object.files = resource.file_ids.each_with_object([]) do |fid, arr|
```
no-reply: `.map`?

app/models/hyrax/file_set.rb
```diff
@@ -24,5 +24,54 @@ def pcdm_object?
...
+    # @return [Enumerable<FileMetadata>] the FileMetadata resources
+    # @example
+    #   filter_files_by_type(::RDF::URI("http://pcdm.org/ExtractedText"))
+    def filter_files_by_type(uri)
+      Hyrax.query_service.custom_queries.find_many_file_metadata_by_use(resource: self, use: uri)
+    end
+
+    ##
+    # @return [RDF::URI] RDF Type for original file use
+    def self.original_file_use
```
no-reply: would these be better as constants?

app/actors/hyrax/actors/file_actor.rb
elrayle: I agree, this is much better as a switch.  Still have to check the relation against the RDF::URI because that is how the types are represented in `pcdm_file.metadata_node.type`

app/services/hyrax/custom_queries/find_file_metadata.rb
```diff
@@ -53,6 +53,16 @@ def find_many_file_metadata_by_ids(ids:)
         results = query_service.find_many_by_ids(ids: ids)
         results.select { |resource| resource.is_a? Hyrax::FileMetadata }
       end
+
+      # Find file metadata for files within a resource that have the requested use.
+      # @param use [RDF::URI] uri for the desired use Type
+      # @return [Array<Hyrax::FileMetadata] or empty array if there are no files with the requested use
+      # @example
+      #   Hyrax.query_service.find_file_metadata_by_use(use: ::RDF::URI("http://pcdm.org/ExtractedText"))
+      def find_many_file_metadata_by_use(resource:, use:)
```
no-reply: :100: i'm eager to have this in!

app/models/hyrax/file_set.rb
no-reply: i.e., i think rebasing #4215 would resolve this, and moving forward with this will facilitate that refactor

lib/wings/services/custom_queries/find_file_metadata.rb
elrayle: Nope - removed

app/models/hyrax/file_set.rb
elrayle: Done

lib/wings/active_fedora_converter.rb
elrayle: Done

app/models/hyrax/file_set.rb
elrayle: Done

lib/wings/active_fedora_converter.rb
elrayle: rewrote to use .map

lib/wings/active_fedora_converter.rb
elrayle: rewrote to use .map

lib/wings/active_fedora_converter.rb
elrayle: rewrote as .map

app/models/hyrax/file_metadata.rb
elrayle: @no-reply This is the only requested change I didn't make.  I don't have time to work on this today, so I'd like to punt this requested change to later.  Open issue #4226 to track this request.

PR #4235, opened 2020-01-30
Support custom URI use/type in `FileSetDescription`

no-reply:
When constructing a description of a Valkyrie description, prefer to use URIs.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -13,12 +13,14 @@ module Hyrax
       THUMBNAIL = ::Valkyrie::Vocab::PCDMUse.ThumbnailImage
 
       ##
-      # @param use [Symbol]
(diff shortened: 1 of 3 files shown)


jeremyf:
I would love to restart the failed ruby2-6-2-valkyrie instance, but I don't have Restart Privileges on CircleCI

no-reply:
@jeremyf thanks, i've kicked it.

I've heard about the restart problem from a handful of folks, but i haven't been able to get anyone to show me what they are experiencing. Maybe you can shoot me some sceenshots and I can see if i can figure out how to grant permissions?

jeremyf:
@no-reply I did some further poking around, and it looks as though I was signed in, but had not associated with Samvera. Once I did that, I had a restart button (and a different UI)

PR #4236, opened 2020-01-31
Use the `FileMetadata#type` not `FileSet#original_file_id` for use 

no-reply:
FileSet's use "relation" should be controlled exclusively by the related
FileMetadata's `#type` property. Move toward removing the `original_file_id` and
other `FileSet` attributes by reworking the existing queries.

@samvera/hyrax-code-reviewers


app/services/hyrax/custom_queries/navigators/find_files.rb
```diff
@@ -1,13 +1,13 @@
 module Hyrax
   module CustomQueries
     module Navigators
+      ##
+      # @example
+      #   Hyrax.custom_queries.find_files(file_set: file_set_resource)
+      #   Hyrax.custom_queries.find_original_file(file_set: file_set_resource)
+      #   Hyrax.custom_queries.find_extracted_text(file_set: file_set_resource)
+      #   Hyrax.custom_queries.find_thumbnail(file_set: file_set_resource)
```
jeremyf: Good catch on documentation. Those examples would've been attached at a weird spot.

app/services/hyrax/custom_queries/navigators/find_files.rb
```diff
@@ -39,40 +39,48 @@ def find_files(file_set:)
...
+          # @api private
+          #
+          # @return [Hyrax::FileMetadata]
+          # @raise [Valkyrie::Persistence::ObjectNotFoundError]
+          def find_exactly_one_file_by_use(file_set:, use:)
+            files =
+              query_service.custom_queries.find_many_file_metadata_by_use(resource: file_set, use: use)
+
+            raise Valkyrie::Persistence::ObjectNotFoundError, "FileSet #{file_set.id}'s #{use.fragment} is missing." if
+              files.blank?
```
jeremyf: `#blank?` seems a bit of an odd method to check; I was reading it and expecting to see `#empty?`. Cognitively, I don't think of a "blank" Array. I could also see `unless files.present?` which would read similarly.

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
```diff
@@ -1,19 +1,23 @@
...
-      let!(:file_metadata1) { FactoryBot.create_using_test_adapter(:hyrax_file_metadata) }
-      let!(:file_metadata2) { FactoryBot.create_using_test_adapter(:hyrax_file_metadata) }
-      let!(:fileset) { FactoryBot.create_using_test_adapter(:hyrax_file_set, files: [file_metadata1, file_metadata2]) }
+      let(:file_metadata1) { FactoryBot.valkyrie_create(:hyrax_file_metadata) }
+      let(:file_metadata2) { FactoryBot.valkyrie_create(:hyrax_file_metadata) }
+      let(:fileset) { FactoryBot.valkyrie_create(:hyrax_file_set, files: [file_metadata1, file_metadata2]) }
+
       it 'returns file metadata resource' do
-        expect(query_handler.find_files(file_set: fileset).map(&:id).map(&:to_s)).to match_array [file_metadata1.id.to_s, file_metadata2.id.to_s]
+        expect(query_handler.find_files(file_set: fileset).map(&:id).map(&:to_s))
```
jeremyf: Inefficient using chained maps, but I like how it reads.

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
```diff
@@ -22,78 +26,93 @@
...
     end
 
     context 'when files do not exist' do
-      let!(:fileset) { FactoryBot.build(:hyrax_file_set) }
+      let(:fileset) { FactoryBot.build(:hyrax_file_set) }
+
       it 'raises error' do
         expect { query_handler.find_original_file(file_set: fileset) }
-          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError, "File set's original file is blank"
+          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError
```
jeremyf: I like removing the second parameter to the `raise_error`, as that should be evident from the spec context.

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
```diff
@@ -22,78 +26,93 @@
...
       it 'raises error' do
         expect { query_handler.find_extracted_text(file_set: fileset) }
-          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError, "File set's extracted text is blank"
+          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError
       end
     end
 
-    context 'when file_set does not respond to extracted text' do
-      let!(:fileset) { FactoryBot.build(:hyrax_resource) }
+    context 'when file_set class does not does not have any files' do
```
jeremyf: Did you mean "file_set class" or "file_set"? I see that as a bit confusing.

app/services/hyrax/custom_queries/navigators/find_files.rb
no-reply: i was thinking `if files.blank?` (or the equivalent `unless files.present?`) is necessary because some adapters may return `nil` instead of an empty array; but the query called is ours and we can just be strict about its return type.

i'll change to `#empty?`

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
no-reply: yeah, probably not what i would have done, but it's fine and not originating here.

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
no-reply: :+1:  it seemed like it was making the test fragile.

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
no-reply: uhhh yeah, this wording is awkward. i'll rephrase.

app/services/hyrax/custom_queries/navigators/find_files.rb
no-reply:
```suggestion
              files.empty?
```

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
no-reply: i think the `to_s` won't be needed much longer; since valkyrie deals with id<->string equality more robustly now

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
no-reply:
```suggestion
    context 'when file_set does not respond to #file_ids' do
```

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
```diff
@@ -22,78 +26,93 @@
...
+      let(:fileset) { FactoryBot.build(:hyrax_file_set) }
+
       it 'raises error' do
         expect { query_handler.find_thumbnail(file_set: fileset) }
-          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError, "File set's thumbnail is blank"
+          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError
       end
     end
 
     context 'when file_set does not respond to thumbnail' do
```
no-reply:
```suggestion
    context 'when file_set does not respond to #file_ids' do
```

spec/services/hyrax/custom_queries/navigators/find_files_spec.rb
```diff
@@ -22,78 +26,93 @@
...
+      let(:fileset) { FactoryBot.build(:hyrax_file_set) }
+
       it 'raises error' do
         expect { query_handler.find_original_file(file_set: fileset) }
-          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError, "File set's original file is blank"
+          .to raise_error ::Valkyrie::Persistence::ObjectNotFoundError
       end
     end
 
     context 'when file_set does not respond to original file' do
```
no-reply:
```suggestion
     context 'when file_set does not respond to #file_ids' do
```

jeremyf:
This error looks weird given other Ruby builds work:

```console
1) Hyrax::CollectionIndexer#generate_solr_document with block yields the document that includes our fields
     # ActiveFedora pattern must become extensible
     Failure/Error: expect { |b| indexer.generate_solr_document(&b) }.to yield_with_args(a_hash_including(doc))
     
       expected given block to yield with arguments, but yielded with unexpected arguments
       expected: [(a hash including {"generic_type_sim" => ["Collection"], "thumbnail_path_ss" => "/downloads/1234?file=thumbnail", "member_of_collection_ids_ssim" => ["col1", "col2"], "member_of_collections_ssim" => ["col1 title", "col2 title"], "visibility_ssi" => "restricted"})]
            got: [{"collection_type_gid_ssim"=>["gid://internal/hyrax-collectiontype/1"], "depositor_ssim"=>["user134@...020-01-31T19:10:52Z", "title_sim"=>["Collection Title 29"], "title_tesim"=>["Collection Title 29"]}]
```

From https://circleci.com/gh/samvera/hyrax/17173#tests/containers/2

PR #4237, opened 2020-01-31
Backport to 2 6 remove collection size feature hyrax 4207

fritzfreiheit:
Backport of #4207 to v2.6.* / 2.x-stable

See: https://github.com/samvera/hyrax/pull/4207

@samvera/hyrax-code-reviewers


jeremyf:
@fritzfreiheit thanks for the work on this. One question I have is if this is something to push to `2.x-stable`; My understanding would be that we want to have a branch that starts at tag `v2.6.0` (which I recently made as `2.6-stable`). We would then craft a PR onto the `2.6-stable` branch. @no-reply does that reconcile with your understanding? If so, do we have documentation speaking to that? If not, I'm more than happy to write up procedures for back porting.

PR #4242, opened 2020-01-31
Attempting to squash a flappy test

jeremyf:
## Fixing spec that intermittently fails

accad29808b702ed8f16fa6476d6f303f97abb79

Alas, the hardest of PRs to review, as its addressing an occassional
but regular failure. However, bear with me.

The error occurred in this [CircleCI Build][1], which had a spec error
as follows.

```log
1) Hyrax::Actors::FileSetActor when use_valkyrie is false \
       #create_content when from_url is true calls ingest_file and \
       kicks off jobs
     Failure/Error: expect(VisibilityCopyJob).to have_been_enqueued

       expected to enqueue exactly 1 jobs, but enqueued 7
```

This lead me to believe that the queues were not being properly cleared
out (or even if that is expected behavior). However, as we are relying
on the Rspec matcher `have_been_enqueued`, there may be something
related to the ActiveJob queue.

Based on the phrase of the failure (e.g. "exactly") and the
[code for building][2] that phrase in `have_been_enqueued`, we have
been relying on the previous code state to clear the queues only after
a `:perform_enqueued` example was run.

This meant that we had a queue that was filling up with more and more
of the same type of jobs, and thus the implicit "exactly" match of
`have_been_enqueued` raised the error.

To replicate the fix, you will need to checkout SHA da3b4632b45a8bf22100f691612d299a0ac79448
of the codebase. You will then need to run the specs using the seed
2186 (e.g. with all services running, use `rspec --seed 2186`).

I believe there is an implicit assumption that we had a cleared out
queue after each run. This change ensures that prior to each run, we
clear the queue.

[1]:https://circleci.com/gh/samvera/hyrax/17187#tests/containers/5
[2]:https://github.com/rspec/rspec-rails/blob/v3.9.0/lib/rspec/rails/matchers/active_job.rb#L86

## Addressing spec warnings concerning `raise_error`

82be9beff3709e1268656f64ec784b711fa07f22

This also involved tidying up the documentation to reflect the expected
parameter type.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/wings/orm_converter_spec.rb
+++ b/spec/wings/orm_converter_spec.rb
@@ -48,7 +48,7 @@ RSpec.describe Wings::OrmConverter do
 
     context 'when given a non-ActiveFedora class' do
       it 'raises an exception' do
-        expect { described_class.to_valkyrie_resource_class(klass: String) }.to raise_error
+        expect { described_class.to_valkyrie_resource_class(klass: String) }.to raise_error(NoMethodError)
       end
     end
 
--- a/lib/wings/orm_converter.rb
+++ b/lib/wings/orm_converter.rb
@@ -33,7 +33,7 @@ module Wings
     end
 
     ##
-    # @param klass [String] an `ActiveFedora` model
+    # @param klass [Class] an `ActiveFedora` model class
     #
     # @return [Class] a dyamically generated `Valkyrie::Resource` subclass
     #   mirroring the provided `ActiveFedora` model
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -83,8 +83,6 @@ Capybara.javascript_driver = :selenium_chrome_headless_sandboxless # This is slo
 # FIXME: Pin to older version of chromedriver to avoid issue with clicking non-visible elements
 Webdrivers::Chromedriver.version = '72.0.3626.69'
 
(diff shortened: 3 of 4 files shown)


jeremyf:
There is a bit of discussion around this PR. Namely, should we be clearing the queue after every run (or at the start its roughly the same).

jeremyf:
I opted for `around` as that could isolate the specific interacts for the job, and keep those  concerns together.

spec/wings/orm_converter_spec.rb
```diff
@@ -48,6 +48,7 @@
 
     context 'when given a non-ActiveFedora class' do
       it 'raises an exception' do
+        # TODO: What error are we expecting to raise
```
no-reply: would it be worth just putting in whatever error is actually raised now, as a first step?

jeremyf: It would be…except I didn't have band-width to fix it. I was more interested in getting this flappy test, which has raised its head quite a bit today, resolved.

jeremyf: With it resolved, I can look at the exceptions (though I'm trying to tackle why CircleCI failed in a different spot and required a restart)

jeremyf: I went ahead and pushed those changes.

PR #4243, opened 2020-02-01
Refactoring for hopefully making 4195 easier

jeremyf:
This morning, I thought I'd tackle one of the jobs. However, it is not immediately obvious how I should proceed (my Hyrax w/ Valkyrie skills are academic at best). These commits reflect a bit of work to get things into a state that I _could_ make it easier to incorporate Valkyrie.

What would be helpful is to see what the next work should be. I attempted to follow #4213, but did not find that helpful in the limited time that I had to look into the problem. My hope is that this is a "leaves the code-base a little better than before I touched it" set of commits.

I'm very curious what the next steps are regarding Valkyrizing this particular job. I feel once I see a pattern, I may be able to crank out a few more.

Below are the two commits and their messages.

## Refactoring ImportUrlJob to ease Valkyrization

369f318b98dc4081e1d94db72172e35cbbfe31bd

The code changes reflect a simple re-structuring that will make future
work easier:

1) Extract method from perform into context specific private method
2) Extract additional attr_reader to allow for common means of using
   instance variables

Refactoring related to #4195

## Refactoring object to leverage instance vars

bfd316e23b5bfbf8d811e8a8dd582c1a2529887b

Prior to this commit, there were parameters passed that could be
instead assigned at the beginning of the `#perform` method. In doing,
we have less chatter on the method calls.

This also involved extract additional instance variables that were
commonly used throughout.

Refactoring related to #4195

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/import_url_job.rb
+++ b/app/jobs/import_url_job.rb
@@ -8,7 +8,7 @@ require 'browse_everything/retriever'
 # and CreateWithRemoteFilesActor when files are located in some other service.
 class ImportUrlJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
-  attr_reader :file_set, :operation
+  attr_reader :file_set, :operation, :headers
 
   before_enqueue do |job|
     operation = job.arguments[1]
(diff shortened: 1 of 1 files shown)


jeremyf:
CircleCI failed (https://circleci.com/gh/samvera/hyrax/17299) which would likely be solved with #4242

PR #4244, opened 2020-02-01
Begin to rework Hyrax's `Dry::Transactions` work to use Do notation

no-reply:
`Dry::Transaction` development has been sunsetted, replaced by a new "Do notation" syntax in `Dry::Monad`. `Do` is more flexible than the older `Transaction` interfaces, but is missing some of the `Dry` ecosystem features like `Dry::Container` driven step resolution.

Our goal in transitioning this nascent feature set to Do notation is to replace the parts of this that have direct value for Hyrax.

To this end, some base behavior on top of Do is added to `Hyrax::Transactions::Transaction`. This base class can either be used directly by passing a list of steps in on init, or subclassed to provide ready-to-use function-specific transactions.

A few of these are provided:
  - `ApplyChangeSet` is intended for general use when applying a `Valkyrie::ChangeSet`, ensuring appropriate results wrt Hyrax's system-managed metadata.
  - `WorkCreate` is an early-stage replacement for the Actor Stack's `#create` stack. There is much work left to do here.

In general, the plan is for Transactions to provide composable, customizable behavior sets for accomplishing sequential tasks, with a clear interface for error handling (`Dry::Monads::Result`). This is meant to show that working and begin to move us away from the false-start `Dry::Transaction` approach.

---
Adds a `Dry::Transaction`-like step argument setup via
`with_step_arguments`. When no arguments are passed for a given step, we pass
only the value wrapped by the previous step. But when step arguments are given,
we pass them into the specified step as supplemental data.

This makes it easy for individual steps to require pieces of information without
other steps having to handle them. In the actor stack, we struggled with
environment parameters mutating throughout the stack, causing brittle order and
behavior dependencies throughout the stack. Step arguments should avoid this
pitfall by ensuring callers pass step-specific data only to the step that
requires them.

To aid debugging and prevent running misconfigured transactions, we fail eagerly
when arguments are passed for steps that don't exist in the transaction.

---

~~One missing feature from `Dry::Transaction` is "step arguments". For example, one could do:~~

```ruby
transaction.with_step_args(step_name: [:additional, :arguments]).call(value)
```

~~This was handy for injecting step-specific arguments without needing to pass them through the other steps, risking mutation and a resulting strong inter-dependency between steps. e.g. `tx.with_step_args(save: {user: current_user})` would save some of the awkwardness about `depositor` in the current implementation.~~

~~What do folks think about reimplementing this in our local `Transaction` base?~~
@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/apply_change_set.rb
+++ b/lib/hyrax/transactions/apply_change_set.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+require 'hyrax/transactions/transaction'
+
+module Hyrax
+  module Transactions
+    ##
+    # Applies and saves a ChangeSet.
+    #
+    # @since 3.0.0
+    #
+    # @example Applying a ChangeSet to a Work
+    #   work = Hyrax::Work.new
+    #   change_set = Hyrax::ChangeSet.for(work)
+    #   change_set.title = ['Comet in Moominland']
+    #
+    #   transaction = Hyrax::Transactions::ApplyChangeSet.new
+    #   result = transaction.call(change_set)
+    #
+    #   result.bind(&:persisted?) => true
+    #
+    #   persisted = result.value_or { raise 'oh no!' } # safe unwrap
+    #   persisted.title      # => ['Comet in Moominland']
+    #
+    class ApplyChangeSet < Transaction
+      DEFAULT_STEPS = ['change_set.set_modified_date',
+                       'change_set.set_uploaded_date',
+                       'change_set.validate',
+                       'change_set.save'].freeze
+
+
+      ##
+      # @see Hyrax::Transactions::Transaction
+      def initialize(container: Container, steps: DEFAULT_STEPS)
+        super
+      end
+    end
+  end
+end
--- a/lib/hyrax/transactions/apply_change_set.rb
+++ b/lib/hyrax/transactions/apply_change_set.rb
@@ -33,7 +33,7 @@ module Hyrax
     #
     class ApplyChangeSet < Transaction
       DEFAULT_STEPS = ['change_set.set_modified_date',
-                       'change_set.set_uploaded_date',
+                       'change_set.set_uploaded_date_unless_present',
                        'change_set.validate',
                        'change_set.save'].freeze
 
--- a/lib/hyrax/transactions/apply_change_set.rb
+++ b/lib/hyrax/transactions/apply_change_set.rb
@@ -4,7 +4,17 @@ require 'hyrax/transactions/transaction'
 module Hyrax
   module Transactions
     ##
-    # Applies and saves a ChangeSet.
+    # Applies and saves a `ChangeSet`.
+    #
+    # This transaction is intended to ensure appropriate results for a Hyrax
+    # model when saving changes from a `ChangeSet`. For example: it will set the
+    # system-managed metadata like modified date.
+    #
+    # If your application has custom system managed metadata, this is an
+    # appropriate place to inject that behavior.
+    #
+    # This will also validate the `ChangeSet`. Which validations to use is
+    # delegated on the `ChangeSet` itself.
     #
     # @since 3.0.0
     #
--- a/lib/hyrax/transactions/apply_change_set.rb
+++ b/lib/hyrax/transactions/apply_change_set.rb
@@ -27,7 +27,6 @@ module Hyrax
                        'change_set.validate',
                        'change_set.save'].freeze
 
-
       ##
       # @see Hyrax::Transactions::Transaction
       def initialize(container: Container, steps: DEFAULT_STEPS)
--- a/lib/hyrax/transactions/apply_change_set.rb
+++ b/lib/hyrax/transactions/apply_change_set.rb
@@ -24,7 +24,7 @@ module Hyrax
     #   change_set.title = ['Comet in Moominland']
     #
     #   transaction = Hyrax::Transactions::ApplyChangeSet.new
-    #   result = transaction.call(change_set)
+    #   result = transaction.call(change_set) # => Success(#<Hyrax::Work ...>)
     #
     #   result.bind(&:persisted?) => true
     #
--- a/lib/hyrax/transactions/container.rb
+++ b/lib/hyrax/transactions/container.rb
@@ -26,15 +26,35 @@ module Hyrax
       require 'hyrax/transactions/steps/destroy_work'
       require 'hyrax/transactions/steps/ensure_admin_set'
       require 'hyrax/transactions/steps/ensure_permission_template'
+      require 'hyrax/transactions/steps/save'
       require 'hyrax/transactions/steps/save_work'
       require 'hyrax/transactions/steps/set_default_admin_set'
       require 'hyrax/transactions/steps/set_modified_date'
       require 'hyrax/transactions/steps/set_uploaded_date'
+      require 'hyrax/transactions/steps/validate'
 
       extend Dry::Container::Mixin
 
       # Disable BlockLength rule for DSL code
       # rubocop:disable Metrics/BlockLength
+      namespace 'change_set' do |ops|
+        ops.register 'save' do
+          Steps::Save.new
+        end
+
+        ops.register 'set_modified_date' do
+          Steps::SetModifiedDate.new
+        end
+
+        ops.register 'set_uploaded_date' do
+          Steps::SetUploadedDate.new
+        end
+
+        ops.register 'validate' do
+          Steps::Validate.new
+        end
+      end
+
       namespace 'work' do |ops|
         ops.register 'apply_collection_permission_template' do
           Steps::ApplyCollectionPermissionTemplate.new
--- a/lib/hyrax/transactions/container.rb
+++ b/lib/hyrax/transactions/container.rb
@@ -32,7 +32,7 @@ module Hyrax
       require 'hyrax/transactions/steps/save_work'
       require 'hyrax/transactions/steps/set_default_admin_set'
       require 'hyrax/transactions/steps/set_modified_date'
-      require 'hyrax/transactions/steps/set_uploaded_date'
+      require 'hyrax/transactions/steps/set_uploaded_date_unless_present'
       require 'hyrax/transactions/steps/validate'
 
       extend Dry::Container::Mixin
@@ -60,8 +60,8 @@ module Hyrax
           Steps::SetModifiedDate.new
         end
 
-        ops.register 'set_uploaded_date' do
-          Steps::SetUploadedDate.new
+        ops.register 'set_uploaded_date_unless_present' do
+          Steps::SetUploadedDateUnlessPresent.new
         end
 
         ops.register 'validate' do
@@ -106,8 +106,8 @@ module Hyrax
           Steps::SetModifiedDate.new
         end
 
-        ops.register 'set_uploaded_date' do
-          Steps::SetUploadedDate.new
+        ops.register 'set_uploaded_date_unless_present' do
+          Steps::SetUploadedDateUnlessPresent.new
         end
       end
       # rubocop:enable Metrics/BlockLength
--- a/lib/hyrax/transactions/container.rb
+++ b/lib/hyrax/transactions/container.rb
@@ -18,8 +18,10 @@ module Hyrax
     #
     # @see https://dry-rb.org/gems/dry-container/
     class Container
+      require 'hyrax/transactions/apply_change_set'
       require 'hyrax/transactions/create_work'
       require 'hyrax/transactions/destroy_work'
+      require 'hyrax/transactions/work_create'
       require 'hyrax/transactions/steps/apply_collection_permission_template'
       require 'hyrax/transactions/steps/apply_permission_template'
       require 'hyrax/transactions/steps/apply_visibility'
@@ -38,6 +40,10 @@ module Hyrax
       # Disable BlockLength rule for DSL code
       # rubocop:disable Metrics/BlockLength
       namespace 'change_set' do |ops|
+        ops.register 'apply' do
+          ApplyChangeSet.new
+        end
+
         ops.register 'save' do
           Steps::Save.new
         end
--- a/lib/hyrax/transactions/container.rb
+++ b/lib/hyrax/transactions/container.rb
@@ -44,8 +44,16 @@ module Hyrax
           ApplyChangeSet.new
         end
 
+        ops.register 'ensure_admin_set' do
+          Steps::EnsureAdminSet.new
+        end
+
         ops.register 'save' do
           Steps::Save.new
+        end
+
+        ops.register 'set_default_admin_set' do
+          Steps::SetDefaultAdminSet.new
         end
 
         ops.register 'set_modified_date' do
--- a/lib/hyrax/transactions/steps/ensure_admin_set.rb
+++ b/lib/hyrax/transactions/steps/ensure_admin_set.rb
@@ -3,19 +3,19 @@ module Hyrax
   module Transactions
     module Steps
       ##
-      # A `dry-transaction` step that ensures the input `work` has an AdminSet.
+      # A step that ensures the input object has an `#admin_set_id`.
       #
       # @since 2.4.0
       class EnsureAdminSet
-        include Dry::Transaction::Operation
+        include Dry::Monads[:result]
 
         ##
-        # @param [Hyrax::WorkBehavior] work
+        # @param [#admin_set_id] obj
         #
         # @return [Dry::Monads::Result] `Failure` if there is no `AdminSet` for
         #   the input; `Success(input)`, otherwise.
-        def call(work)
-          work.admin_set_id ? Success(work) : Failure(:no_admin_set_id)
+        def call(obj)
+          obj.admin_set_id ? Success(obj) : Failure(:no_admin_set_id)
         end
       end
     end
--- a/lib/hyrax/transactions/steps/save.rb
+++ b/lib/hyrax/transactions/steps/save.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+require 'dry/monads'
+
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # Saves a given work, returning a Result (Success|Failure)
+      class Save
+        include Dry::Monads[:result]
+
+        ##
+        # @params [#save] persister
+        def initialize(persister: Hyrax.persister)
+          @persister = persister
+        end
+
+        ##
+        # @param [Hyrax::ChangeSet] change_set
+        #
+        # @return [Dry::Monads::Result] `Success(work)` if the change_set is
+        #   applied and the resource is saved;
+        #   `Failure([#to_s, change_set.resource])`, otherwise.
+        def call(change_set)
+          change_set.sync
+
+          Success(@persister.save(resource: change_set.resource))
+        rescue StandardError => err
+          Failure([err.message, change_set.resource])
+        end
+      end
+    end
+  end
+end
--- a/lib/hyrax/transactions/steps/save.rb
+++ b/lib/hyrax/transactions/steps/save.rb
@@ -23,17 +23,17 @@ module Hyrax
 
         ##
         # @param [Hyrax::ChangeSet] change_set
+        # @param [::User, nil] user
         #
         # @return [Dry::Monads::Result] `Success(work)` if the change_set is
         #   applied and the resource is saved;
         #   `Failure([#to_s, change_set.resource])`, otherwise.
-        def call(change_set)
+        def call(change_set, user: nil)
           saved = @persister.save(resource: change_set.sync)
 
-          depositor = ::User.find_by_user_key(saved.try(:depositor))
           Hyrax.publisher.publish('object.metadata.updated',
                                   object: saved,
-                                  user: depositor)
+                                  user: user)
 
           Success(saved)
         rescue StandardError => err
--- a/lib/hyrax/transactions/steps/save.rb
+++ b/lib/hyrax/transactions/steps/save.rb
@@ -5,7 +5,13 @@ module Hyrax
   module Transactions
     module Steps
       ##
-      # Saves a given work, returning a Result (Success|Failure)
+      # Saves a given work from a change_set, returning a `Dry::Monads::Result`
+      # (`Success`|`Failure`).
+      #
+      # If the save is successful, publishes an `object.metadata.updated` event
+      # for the affected resource.
+      #
+      # @see https://dry-rb.org/gems/dry-monads/1.0/result/
       class Save
         include Dry::Monads[:result]
 
@@ -22,9 +28,14 @@ module Hyrax
         #   applied and the resource is saved;
         #   `Failure([#to_s, change_set.resource])`, otherwise.
         def call(change_set)
-          change_set.sync
+          saved = @persister.save(resource: change_set.sync)
 
-          Success(@persister.save(resource: change_set.resource))
+          depositor = ::User.find_by_user_key(saved.try(:depositor))
+          Hyrax.publisher.publish('object.metadata.updated',
+                                  object: saved,
+                                  user: depositor)
+
+          Success(saved)
         rescue StandardError => err
           Failure([err.message, change_set.resource])
         end
--- a/lib/hyrax/transactions/steps/set_default_admin_set.rb
+++ b/lib/hyrax/transactions/steps/set_default_admin_set.rb
@@ -3,21 +3,22 @@ module Hyrax
   module Transactions
     module Steps
       ##
-      # A `dry-transaction` step that sets the `AdminSet` for an input work to
-      # the default admin set, if none is already set. Creates the default
-      # admin set if it doesn't already exist.
+      # A step that sets the `AdminSet` for an input work to the default admin
+      # set if none is already set. Creates the default admin set if it doesn't
+      # already exist.
+      #
+      # @since 2.4.0
       class SetDefaultAdminSet
-        include Dry::Transaction::Operation
+        include Dry::Monads[:result]
 
         ##
-        # @param [Hyrax::WorkBehavior] work
+        # @param [#admin_set_id=] obj
         #
         # @return [Dry::Monads::Result]
-        def call(work)
-          work.admin_set ||=
-            AdminSet.find(AdminSet.find_or_create_default_admin_set_id)
+        def call(obj)
+          obj.admin_set_id ||= AdminSet.find_or_create_default_admin_set_id
 
-          Success(work)
+          Success(obj)
         end
       end
     end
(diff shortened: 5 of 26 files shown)


lib/hyrax/transactions/apply_change_set.rb
```diff
@@ -0,0 +1,35 @@
...
+    #
+    # @since 3.0.0
+    #
+    # @example Applying a ChangeSet to a Work
+    #   work = Hyrax::Work.new
+    #   change_set = Hyrax::ChangeSet.for(work)
+    #   change_set.title = ['Comet in Moominland']
+    #
+    #   transaction = Hyrax::Transactions::ApplyChangeSet.new
+    #   persisted   = transaction.call(change_set).value_or(:error)
```
jeremyf: I'm assuming that the `.value_or(:error)` is something from the Dry world? It reads a bit awkwardly.

lib/hyrax/transactions/apply_change_set.rb
```diff
@@ -0,0 +1,35 @@
...
+    #
+    #   persisted.persisted? # => true
+    #   persisted.title      # => ['Comet in Moominland']
+    #
+    class ApplyChangeSet < Transaction
+      DEFAULT_STEPS = ['change_set.set_modified_date',
+                       'change_set.set_uploaded_date',
```
jeremyf: I understand the "change_set.set_modified_date", "change_set.validate", and "change_set.save"; However the "change_set.set_uploaded_date" doesn't quite make sense. Perhaps further reading will help.

lib/hyrax/transactions/steps/ensure_admin_set.rb
```diff
@@ -3,19 +3,19 @@ module Hyrax
...
+      # A step that ensures the input object has an `#admin_set_id`.
       #
       # @since 2.4.0
       class EnsureAdminSet
-        include Dry::Transaction::Operation
+        include Dry::Monads[:result]
 
         ##
-        # @param [Hyrax::WorkBehavior] work
+        # @param [#admin_set_id] obj
```
jeremyf: I like the improvement of this documentation

lib/hyrax/transactions/steps/save.rb
```diff
@@ -0,0 +1,39 @@
...
+        ##
+        # @param [Hyrax::ChangeSet] change_set
+        #
+        # @return [Dry::Monads::Result] `Success(work)` if the change_set is
+        #   applied and the resource is saved;
+        #   `Failure([#to_s, change_set.resource])`, otherwise.
+        def call(change_set)
+          saved = @persister.save(resource: change_set.sync)
+
+          depositor = ::User.find_by_user_key(saved.try(:depositor))
```
jeremyf: This looks a bit odd…I'm reading this as "depositor" is optional.

lib/hyrax/transactions/steps/set_uploaded_date.rb
```diff
@@ -3,25 +3,44 @@ module Hyrax
...
+          Success(obj)
         end
+
+        private
+
+          def date_uploaded(obj)
+            if obj.try(:date_modified).present?
+              obj.try(:date_modified)
+            else
+              @time_service.time_in_utc
```
jeremyf: I don't think we have a `@time_service` object (as per  lib/hyrax/transactions/steps/set_modified_date.rb)

lib/hyrax/transactions/transaction.rb
```diff
@@ -0,0 +1,42 @@
...
+      def initialize(container: Container, steps:)
+        self.container = container
+        self.steps     = steps
+      end
+
+      ##
+      # @param [Valkyrie::ChangeSet] change_set
+      #
+      # @return [Dry::Monads::Result]
+      def call(change_set)
```
jeremyf: Can I get a comment on what is happening here? Perhaps expand the variable naming of `w` and `s`.

My understanding of `#inject` is that the yielded value will become the `w` for the next `s`.

lib/hyrax/transactions/work_create.rb
```diff
@@ -0,0 +1,22 @@
...
+require 'hyrax/transactions/transaction'
+
+module Hyrax
+  module Transactions
+    ##
+    # Creates a Work from a ChangeSet
+    #
+    # @since 3.0.0
+    class WorkCreate < Transaction
+      DEFAULT_STEPS = ['change_set.set_default_admin_set',
```
jeremyf: I have to say, I really love the composition of these steps. They guide me towards an understanding of what goes into the "WorkCreate" transaction in a clearer means than the ActorStack

spec/hyrax/transactions/apply_change_set_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  describe '#call' do
+    it 'is a success' do
+      expect(tx.call(change_set)).to be_success
+    end
+
+    it 'wraps a saved work' do
+      result = tx.call(change_set)
+
+      expect(result.value!).to be_persisted
```
jeremyf: I'm not used to seeing a comparison of a "bang!" method return value. I'm assuming that's a Dry::Monad idiom.

spec/hyrax/transactions/apply_change_set_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      it 'is a failure' do
+        expect(tx.call(change_set)).to be_failure
+      end
+    end
+
+    context 'when save step fails' do
+      before do
+        tx.container.enable_stubs!
+        tx.container.stub('change_set.save', failure_step)
```
jeremyf: Now this is some spiffy stubbing behavior. I see that we may wish for a spec_helper for failed containers.

lib/hyrax/transactions/apply_change_set.rb
no-reply: yeah, it's a `Dry::Monad` thing

lib/hyrax/transactions/apply_change_set.rb
no-reply: maybe set `set_uploaded_date` could be renamed? it's actually something like `set_uploaded_date_if_not_present`. the transaction assumes you'll never want to set the uploaded date if it has already been set.

the workflow is:
  - set the modified date to now;
  - set the uploaded date to the modified date, if not already set
  - ...

lib/hyrax/transactions/steps/save.rb
no-reply: that's right. i'm still fighting a bit with this flow.

The event payload wants to have the user responsible for the change (or `nil`, if none (?)), but the transaction doesn't already have it.

While Hyrax Core Metadata demands a depositor, it's strongly preferable not to raise an error after saving in this step, regardless of the input. The object has been saved, and we *really* want to publish the appropriate events whatever else may happen. Since the transaction can't guarantee that the `ChangeSet` validation will support depositor, it seems necessary to be resilient to its absence and to the method missing case.

lib/hyrax/transactions/steps/set_uploaded_date.rb
no-reply: i'm not following. `@time_service` is set in the initializer, right?

lib/hyrax/transactions/transaction.rb
no-reply: will rename the variables for clarity, and add a top-level method comment about what this is doing.

spec/hyrax/transactions/apply_change_set_spec.rb
no-reply: it's usual for `Dry::Monad`'s `Result` monads to access with handling for both the `Success` case (do something with the value) *and* the `Failure` case (log an error, recover, and/or raise). this is what's going on in `#value_or` calls; a bit like `Hash#fetch`'s "give me the value, or run this block" approach.

in the test suite, it's handy to just use this `#value!` method ("give met the value, or panic") since we expect a `Success` and are glad to see the test fail eagerly in any other case.

spec/hyrax/transactions/apply_change_set_spec.rb
no-reply: pretty nice, huh?

the `Dry::Container` dependency resolution also makes it super easy to inject specialized behavior for any given step at runtime in production.

no-reply:
> As this is WIP, I have some comments but am neither approving nor requesting changes.

Thanks so much for the early feedback!

lib/hyrax/transactions/transaction.rb
no-reply: maybe the rename and added docs help?

no-reply:
Okay, so this is starting to feel a bit more complete, so i'm taking off `[WIP]`. The PR description has some context and questions.

lib/hyrax/transactions/apply_change_set.rb
jeremyf: The `set_uploaded_date_if_not_present` sounds more descriptive. I prefer that.

lib/hyrax/transactions/steps/save.rb
jeremyf: @no-reply Could you add that to the code comments? I think the context would be extremely helpful for what sounded like working lore of the Core Metadata implementation

lib/hyrax/transactions/steps/set_uploaded_date.rb
jeremyf: You are correct. I must have overlooked the initializer.

spec/hyrax/transactions/apply_change_set_spec.rb
jeremyf: Excellent. Thanks for the insight.

lib/hyrax/transactions/steps/save.rb
no-reply: what do you think about the `with_step_args` option mentioned in the PR comment?

i'm thinking it's very likely we'll completely rip up this line, so i'm not inclined to spend much time refining it or commenting on it. but maybe the short-term solution is to just always pass `nil` as the user until we have a way for the caller to pass that in?

no-reply:
```suggestion
```

lib/hyrax/transactions/steps/save.rb
```diff
@@ -0,0 +1,39 @@
...
+          Hyrax.publisher.publish('object.metadata.updated',
+                                  object: saved,
+                                  user: depositor)
```
no-reply:
```suggestion
                                  user: nil)
```

lib/hyrax/transactions/apply_change_set.rb
no-reply: renamed!

lib/hyrax/transactions/apply_change_set.rb
```diff
@@ -0,0 +1,47 @@
...
+    #
+    # @since 3.0.0
+    #
+    # @example Applying a ChangeSet to a Work
+    #   work = Hyrax::Work.new
+    #   change_set = Hyrax::ChangeSet.for(work)
+    #   change_set.title = ['Comet in Moominland']
+    #
+    #   transaction = Hyrax::Transactions::ApplyChangeSet.new
+    #   result = transaction.call(change_set)
```
straleyb: It would be nice to see what the data type of this result is. It SEEMS like it might be a resource? But it would be nice to see here in the documentation.

no-reply:
```suggestion
    #   result = transaction.call(change_set) # => Success(#<Hyrax::Work ...>)
```

no-reply: @straleyb it's definitely not a resource, so i guess that means there needs to be documentation here. https://github.com/samvera/hyrax/pull/4244/files/2cd191b9d812382eb7a2ecd580809ecaeb5e769c#r374209760

note that the `#call` documentation *does* give an explicit return type, and there's more narrative about this there and in the top-level `Transaction` class documentation.

straleyb:
Very interesting updates here. Im not familiar with Monads and the use of this piece of architecture. My guess is that it sort of replaces the Actor Stack sort of stuff where you create a specific type of transaction and run #call on it. This would then run a list of steps to set and verify changes on a specific work based on the attributes contained in the ChangeSet? 

This is more just me catching up on this idea rather than any sort of commentary on the PR itself.

lib/hyrax/transactions/apply_change_set.rb
straleyb: I think im understanding that the result is either a Success or Failure. Still would like to see that in the documentation here though

no-reply:
> Very interesting updates here. Im not familiar with Monads and the use of this piece of architecture. My guess is that it sort of replaces the Actor Stack sort of stuff where you create a specific type of transaction and run #call on it. This would then run a list of steps to set and verify changes on a specific work based on the attributes contained in the ChangeSet?
> 
> This is more just me catching up on this idea rather than any sort of commentary on the PR itself.

that's the gist. my hope is that [we can _mostly_ avoid talking about "Monads"](http://existentialcomics.com/comic/other/10) outside these PRs. for users, the upshot is that they can expect transactions to return a value "boxed" in a `Success` or a `Failure`, which both have the same interface.

lib/hyrax/transactions/apply_change_set.rb
no-reply: sorry, does the suggested edit satisfy this request? or is there more needed?

lib/hyrax/transactions/steps/save.rb
jeremyf: Reading your current PR post and in light of this comment, the PR and your comment make sense. Good work. I do want @cjcolvar to weigh in as he's often mentioned Dry::Transaction and other actor stack related work that he's done; He's done work closer to this PRs area of concern.

lib/hyrax/transactions/transaction.rb
```diff
@@ -0,0 +1,97 @@
...
+    #
+    #   tx.call(change_set)
+    #   # => Failure([:failed_validation, #<Reform::Form::ActiveModel::Errors ...>]
+    #
+    # @example unwapping values safely with handling for failures
+    #   tx.call(change_set).value_or { |failure| "uh oh: #{failure} }
+    #
+    # @example a pattern for subclassing to create new transactions
+    #   class CustomTransaction < Transaction
+    #     DEFAULT_STEPS ['step.1', 'step.2', 'step.3']
```
cjcolvar:
```suggestion
    #     DEFAULT_STEPS = ['step.1', 'step.2', 'step.3'].freeze
```

no-reply:
now with step arguments!

lib/hyrax/transactions/container.rb
```diff
@@ -18,23 +18,57 @@ module Transactions
...
+
+        ops.register 'set_default_admin_set' do
+          Steps::SetDefaultAdminSet.new
+        end
+
+        ops.register 'set_modified_date' do
+          Steps::SetModifiedDate.new
+        end
+
+        ops.register 'set_uploaded_date_unless_present' do
```
jeremyf: One of the most liberating moments in software development naming is getting to the point where the classes are descriptive and verbose (and even include verbs). I found as I moved towards more descriptive classes, the full breadth of language (and thus symbols) became available. I love this class name!

lib/hyrax/transactions/steps/save.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+require 'dry/monads'
+
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # Saves a given work, returning a Result (Success|Failure)
```
jeremyf: Are `Success` and `Failure` name spaced? I don't know if that might be helpful to include an @see reference to their namespace location.

lib/hyrax/transactions/steps/set_default_admin_set.rb
```diff
@@ -3,21 +3,22 @@ module Hyrax
...
     module Steps
       ##
-      # A `dry-transaction` step that sets the `AdminSet` for an input work to
-      # the default admin set, if none is already set. Creates the default
-      # admin set if it doesn't already exist.
+      # A step that sets the `AdminSet` for an input work to the default admin
+      # set if none is already set. Creates the default admin set if it doesn't
+      # already exist.
+      #
+      # @since 2.4.0
```
jeremyf: As this Step already exists, I'm okay with it. I do like the more conditional naming `SetDefaultAdminSetUnlessAlreadySet` class, but can see maintaining the previous implemented name.

lib/hyrax/transactions/steps/set_uploaded_date_unless_present.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # A step that sets the uploaded date to now for an input resource or
```
jeremyf: Based on the `date_modified` method I believe the documentation could be tightened a bit. Perhaps an `@see`? Given the specificity of these classes, I can see further documenting the specific implementation details.

lib/hyrax/transactions/steps/validate.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+require 'dry/monads'
+
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # Saves a given work, returning a Result (Success|Failure)
```
jeremyf: The documentation does not match the implementation. It appears that we are validating the change_set, and not saving the given work.

lib/hyrax/transactions/transaction.rb
```diff
@@ -0,0 +1,141 @@
...
+      # @!attribute [rw] container
+      #   @return [Container]
+      # @!attribute [rw] steps
+      #   @return [Array<String>]
+      attr_accessor :container, :steps
+
+      ##
+      # @api public
+      #
+      # @param [Container] container
```
jeremyf: Is the Container namespaced?

lib/hyrax/transactions/transaction.rb
```diff
@@ -0,0 +1,141 @@
...
+      end
+
+      private
+
+        ##
+        # @api private
+        # @param [String, Symbol] step_name
+        # @return [Array]
+        def step_arguments_for(step_name)
+          Array.wrap((@_step_args || {})[step_name])
```
jeremyf: There is a bit of visual gymnastics with this one; It is hard to scan like a medium complexity regexp. Perhaps separate the hash lookup/setting:

```ruby
step_args = @_set_args || {}
Array.wrap(step_args[step_name])
```

lib/hyrax/transactions/transaction.rb
```diff
@@ -0,0 +1,141 @@
...
+      def call(value)
+        Success(
+          steps.inject(value) do |val, step_name|
+            yield container[step_name].call(val, *step_arguments_for(step_name))
+          end
+        )
+      end
+
+      ##
+      # @param [Hash<Object, Array>] args
```
jeremyf: Can you speak to the purpose of this method in the documentation? I think the spec will help me understand, but a bit more documentation would be beautiful.

lib/hyrax/transactions/transaction.rb
```diff
@@ -0,0 +1,141 @@
...
+        end
+
+        ##
+        # @api private
+        # @param [Array] keys
+        # @return [String]
+        def key_err_msg(keys)
+          missing_steps = keys.select { |key| !step?(key) }
+
+          "Tried to pass step aguments for unknown steps #{missing_steps.join(', ')}\n" \
```
jeremyf: Typo: "aguments" -> "arguments"

spec/hyrax/transactions/apply_change_set_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  subject(:tx)     { described_class.new }
+  let(:change_set) { Hyrax::ChangeSet.for(resource) }
+  let(:resource)   { build(:hyrax_work) }
+  let(:user)       { create(:user) }
+  let(:xmas)       { DateTime.parse('2018-12-25 11:30').iso8601 }
+
+  before { allow(Hyrax::TimeService).to receive(:time_in_utc).and_return(xmas) }
+
+  describe '#call' do
+    it 'is a success' do
```
jeremyf: I believe lines 17 through 25 would benefit with a context scope similar to line 55 (eg. `context "with a valid changeset" do`)

spec/hyrax/transactions/steps/save_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  before { Hyrax.publisher.subscribe(listener) }
+  after  { Hyrax.publisher.unsubscribe(listener) }
+
+  let(:change_set_class) do
+    Class.new(Hyrax::ChangeSet) { self.fields = [:title] }
+  end
+
+  describe '#call' do
+    it 'is a success' do
```
jeremyf: I believe a wrapping context for lines 23 to 36 for "when the save succeeds" (or some such thing).

spec/hyrax/transactions/steps/validate_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+require 'spec_helper'
+require 'hyrax/transactions/steps/validate'
+
+RSpec.describe Hyrax::Transactions::Steps::Validate do
+  subject(:step)   { described_class.new }
+  let(:change_set) { Hyrax::ChangeSet.for(resource) }
+  let(:resource)   { build(:hyrax_work) }
+
+  describe '#call' do
+    it 'is a success' do
```
jeremyf: Wrap lines 11 through 20 in a context regarding valid change set.

spec/hyrax/transactions/work_create_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+require 'hyrax/transactions'
+require 'dry/container/stub'
+
+RSpec.describe Hyrax::Transactions::WorkCreate do
+  subject(:tx)     { described_class.new }
+  let(:change_set) { Hyrax::ChangeSet.for(resource) }
+  let(:resource)   { build(:hyrax_work) }
+
+  describe '#call' do
+    it 'is a success' do
```
jeremyf: Can you wrap lines 12 through 23 in a context analogue to line 25?

spec/models/hyrax/change_set_spec.rb
```diff
@@ -2,37 +2,50 @@
...
-  it 'applies changeset attributes' do
-    change_set.title = titles
+    it 'does not expose reserved attributes' do
+      expect { change_set.id = 'fake id' }.to raise_error NoMethodError
+    end
 
-    expect { change_set.sync }
-      .to change { resource.title }
-      .to contain_exactly(*titles)
+    it 'does not list attributes' do
```
jeremyf: Consider `it 'does not list reserved attributes as fields'`

lib/hyrax/transactions/steps/save.rb
no-reply: it seems to be conventional in `dry-monads` to use just `Success` and `Failure`.

`include Dry::Monads[:result]` creates an alias for each in the local namespace: htps://github.com/dry-rb/dry-monads/blob/master/lib/dry/monads/result.rb#L296-L299

i'd welcome documentation improvements, especially if they can be applied to steps and transactions consistently.

lib/hyrax/transactions/steps/set_uploaded_date_unless_present.rb
no-reply: thanks i think this can be expanded pretty easily and will push up some changes.

lib/hyrax/transactions/steps/validate.rb
no-reply: oops! good catch.

spec/hyrax/transactions/steps/save_spec.rb
no-reply: i tend to have a practice of leaving some default case ("happy path"?) outside a context in specs. i'm happy to change these ones in particular on your say-so, but it might be worth considering the desired convention across the code base

spec/models/hyrax/change_set_spec.rb
no-reply:
```suggestion
    it 'does not list reserved attributes as fields' do
```

lib/hyrax/transactions/steps/save.rb
jeremyf: I'll take a look at the Yard output to see how it resolves.

jeremyf: Yard does not resolve the constant as something visible in Hyrax's documentation. I think morning Jeremy is ready to resolve what evening Jeremy raised.

lib/hyrax/transactions/transaction.rb
no-reply: this is a little fussy. `Container` needs to extend `Dry::Container::Mixin`, but it's not actually a class and there's not a common inherited ancestor. The best I could figure was to allow this to resolve to `Hyrax::Transaction::Container` (certainly it should register all the names registered there) but leave it a bit vague since I think it wouldn't be out-of-the-question to slot in a new container matching that interface.

I guess in short, it's really `Hyrax::Transaction::Container` that defines the interface here.

lib/hyrax/transactions/transaction.rb
no-reply: Some new docs are in. Hopefully this helps?

jeremyf: It does.

lib/hyrax/transactions/steps/save.rb
```diff
@@ -0,0 +1,45 @@
...
+        # @params [#save] persister
+        def initialize(persister: Hyrax.persister)
+          @persister = persister
+        end
+
+        ##
+        # @param [Hyrax::ChangeSet] change_set
+        # @param [::User, nil] user
+        #
+        # @return [Dry::Monads::Result] `Success(work)` if the change_set is
```
jeremyf: My reading is that this is correct, we return a `Success(work)` but a `Failure(change_set.resource)`. The reason for the comment is that wrapping different objects in the return may look / feel a bit odd (when it is correct behavior).

_ps_ This is a non-blocking observation

no-reply: `Success` will always wrap the object to be used by the next step; here we transition from a `change_set` to a newly saved resource. 

It's normally the case that `Failure` will wrap something different, whatever is needed to communicate the type of failure and handle it. The convention for what `Failure` wraps is a bit under-determined so far. This one does `['message', change_set.resource]` with the idea that relevant errors from the save will be in the active_model errors. I think i'm on the verge of proposing this as the standard format for errors for Hyrax transactions/steps: `Array<#to_s, Object>` where `#to_s` should be human readable as a description of the error, and `Object` should be something particular to the error type that could aid in reporting, diagnosis, and recovery.

PR #4246, opened 2020-02-01
Attempting to quash SQLite3::BusyException

jeremyf:
On occasion, our specs encounter a `SQLite3::BusyException` exception.
This commit, based on some Google-fu (actually DuckDuckGo-ing but
that's not yet in our vernacular), I found a few references that
involve increasing the `RAILS_MAX_THREADS` environment variable and/or
increasing the database timeout.

In CircleCI, for SHA 600753e0a45f4a5682da0b8188063f8178954c11 and rspec
seed 3048, we encountered the problem. I don't assume that we can
duplicate this behavior with that configuration. There is an example of a [Postgresql configuration](https://github.com/CircleCI-Public/circleci-demo-ruby-rails/blob/master/.circleci/config.yml#L10-L19).

The advantage of keeping Sqlite is that local testing is made easier. We could consider introducing a switch for CI to use Postgresql and local testing to use Sqlite.

References

* https://stackoverflow.com/questions/7154664/ruby-sqlite3busyexception-database-is-locked

Based on my understanding of our CircleCI ecosystem, it appears that we are running our specs in parallel. I'm wondering if that means each of those specs share the same SQLite file. If so, we may have a case where we need to use a different database for CircleCI specs (e.g. one that allows more than one process against it).

@samvera/hyrax-code-reviewers


jeremyf:
I have a monkey patch in the wings…but I am loath to bring that to bear. I think I'd rather look into using MySQL for our test ecosystem.

PR #4254, opened 2020-02-03
Revert "Attempting to quash SQLite3::BusyException"

jeremyf:
This reverts commit 00316d0be39dcea3aabe9bb2f99652d3cff4df0a.

Based on CircleCI's response, instead of one of the builds fail, there
were failures of three of the builds. Two of which were
SQLite3::BusyException, and the second a UI test that could've masked
the busy exception.

I believe we need to be looking to switching from SQLite in CI to
Postgresql or MySQL. This will require forking logic, as we have
traditionally allowed SQLite to work for localized development and
testing (one simple dependency in an otherwise baroque labyrinth of
heavy JAVA-esque dependencies).

Thoughts?

@samvera/hyrax-code-reviewers


no-reply:
i think it would be nice to add testing with Postgres, in part because we'll need to do that to add testing for postgresql support via Valkyrie anyway.

apart from that, i'd be inclined not to spend too much time on this problem. in my experience, other flappy tests have been an order of magnitude more impactful than the occasional SQLite error.

PR #4255, opened 2020-02-03
Hyrax Valkyrie Setup and Specs. Take 2

straleyb:
This references the work done here: #4221

This gets the model conversion work, tests and generator stuff back up to speed.


straleyb:
@geekscruff Do you see anything that may be missing here or are we back up to speed? I know this is waiting on some AF changes to get the nested_attributes_for inherited properly when subclassed. So I think it may be on hold until that is updated.

config/metadata/basic_metadata.yaml
```diff
@@ -2,71 +2,131 @@ attributes:
   abstract:
     type: string
     multiple: true
-  access_right:
+    index_keys:
```
jeremyf: It appears that order has been changed. As I'm not familiar with these files, is order significant for the keys?

jeremyf: If no, the order is not significant, I'd prefer to see the keys not re-ordered as it makes the review more complicated. That is not a blocker, but makes the effort of review more significant.

no-reply: i've been trying to keep the key order alphabetical, for ease of reference. it would be good to fix this up to conform to that convention.

straleyb: I altered the order in the pursuit to verify that we had captured all the necessary pieces of info to be indexed. This was purely for cross referencing. I can update this to be alphabetical again. No biggy for me.

geekscruff:
I'm re-opening this as I still want to get to it. Step away, stale-bot.

PR #4256, opened 2020-02-04
Modifying engine cart CI build parameterization

jeremyf:
At present, EngineCart's parsing of `.engine_cart.yml` will overwrite
the `ENGINE_CART_RAILS_OPTIONS` set in CircleCI (see
[cbeer/engine_cart/102][1] for some details). The present overwrite
behavior is reasonable. As the removed `.engine_cart.yml` targeted
pre-Rails 5 behavior, I opted to remove it.

In doing so, this would now "activate" the `.circleci/config.yml`
declaration of `ENGINE_CART_RAILS_OPTIONS`. As we have coffee script in
Hyrax (for better or worse) and we have ActionCable usage, I needed to
revisit the now activated parameters.

Note: This is in preparation for allowing PostgreSQL as the database
for CircleCI.

As this impacts the `.internal_test_app` generation, I also needed to
bump the Regen number.

TL;DR: A 4 year old file that we no longer need is silently trampling
on a CI variable. That CI variable has likely not been used. It now will
be used, but needed modifications to reflect that usage.

_I hope this works, as iterating on the CI process is always time
consuming as it requires coordination of remote resources._

[1]:https://github.com/cbeer/engine_cart/pull/102

@samvera/hyrax-code-reviewers


straleyb:
@jeremyf  When this gets merged, will people have to regenerate their internal_test_apps? Just in case I get to doing some work at some point in time this week.

jeremyf:
You will not need to regenerate your internal test app once this is merged.
It's just cleaning up and aligning explicitly what was implicitly
happening.

PR #4257, opened 2020-02-05
Implement member->collection directional Collection membership

no-reply:
Works should be able to track their collection membership directly, without
writing to the collection to determine membership.

There are a few things to note here:

  - This tracks closely with the `Hydra::PCDM` implementation, which has
    advantages in the short term.
  - It may be possible to simply implement membership as parent->child with
    `#member_ids` across the board, and handle the inversion for `ActiveFedora`
    in `Wings`. Currently thinking this isn't a fruitful exercise though...
  - We already support `member_ids` in Collections; maybe we should drop this?
    It's in `Hydra::PCDM`, too.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/custom_queries/find_access_control.rb
+++ b/app/services/hyrax/custom_queries/find_access_control.rb
@@ -19,6 +19,8 @@ module Hyrax
           .find_inverse_references_by(resource: resource, property: :access_to)
           .find { |r| r.is_a?(Hyrax::AccessControl) } ||
           raise(Valkyrie::Persistence::ObjectNotFoundError)
+      rescue ArgumentError # some adapters raise ArgumentError for missing resources
+        raise(Valkyrie::Persistence::ObjectNotFoundError)
       end
     end
   end
--- a/app/services/hyrax/custom_queries/navigators/collection_members.rb
+++ b/app/services/hyrax/custom_queries/navigators/collection_members.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module CustomQueries
+    module Navigators
+      ##
+      # Find members of collections and collections for members
+      #
+      # @since 3.0.0
+      class CollectionMembers
+        ##
+        # @return [Array<Symbol>
+        def self.queries
+          [:find_child_works, :find_child_work_ids]
+        end
+        ##
+        # @!attribute [r] query_service
+        #   @return [#custom_queries]
+        attr_reader :query_service
+
+        def initialize(query_service:)
+          @query_service = query_service
+        end
+
+        ##
+        # Find child works of a given resource, and map to Valkyrie Resources
+        #
+        # @param [Valkyrie::Resource] resource
+        # @return [Array<Valkyrie::Resource>]
+        def find_collections_for(resource:)
+          query_service
+            .find_references_by(resource: resource, property: :member_of_collection_ids)
+        end
+
+        ##
+        # Find members for the given collection
+        #
+        # @param [Valkyrie::Resource] collection
+        # @return [Array<Valkyrie::Resource>]
+        def find_members_of(collection:)
+          query_service
+            .find_inverse_references_by(resource: collection, property: :member_of_collection_ids)
+        end
+      end
+    end
+  end
+end
--- a/lib/hyrax/specs/shared_specs/hydra_works.rb
+++ b/lib/hyrax/specs/shared_specs/hydra_works.rb
@@ -34,7 +34,7 @@ RSpec.shared_examples 'belongs to collections' do
   let(:collection_class) { Hyrax::PcdmCollection }
 
   describe 'collection membership' do
-    it 'is in no collections default' do
+    it 'is in no collections by default' do
       expect(model.member_of_collection_ids).to be_empty
     end
 
--- a/lib/hyrax/specs/shared_specs/hydra_works.rb
+++ b/lib/hyrax/specs/shared_specs/hydra_works.rb
@@ -26,6 +26,63 @@ RSpec.shared_examples 'a Hyrax::Resource' do
   it { is_expected.to respond_to :work? }
 end
 
+RSpec.shared_examples 'belongs to collections' do
+  subject(:model)        { described_class.new }
+  let(:adapter)          { Valkyrie::MetadataAdapter.find(:test_adapter) }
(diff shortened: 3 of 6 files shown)


app/services/hyrax/custom_queries/find_access_control.rb
```diff
@@ -19,6 +19,8 @@ def find_access_control_for(resource:)
           .find_inverse_references_by(resource: resource, property: :access_to)
           .find { |r| r.is_a?(Hyrax::AccessControl) } ||
           raise(Valkyrie::Persistence::ObjectNotFoundError)
+      rescue ArgumentError # some adapters raise ArgumentError for missing resources
```
jeremyf: Oof…that sounds like an upstream inconsistency. @tpendragon what are your thoughts as this relates to Valkyrie?

app/services/hyrax/custom_queries/navigators/collection_members.rb
```diff
@@ -0,0 +1,47 @@
...
+          query_service
+            .find_references_by(resource: resource, property: :member_of_collection_ids)
+        end
+
+        ##
+        # Find members for the given collection
+        #
+        # @param [Valkyrie::Resource] collection
+        # @return [Array<Valkyrie::Resource>]
+        def find_members_of(collection:)
```
jeremyf: Thank you for naming the parameters…this helps clarify what we are expecting (instead of relying on positional parameters)

lib/hyrax/specs/shared_specs/hydra_works.rb
```diff
@@ -26,6 +26,63 @@
...
 
+RSpec.shared_examples 'belongs to collections' do
+  subject(:model)        { described_class.new }
+  let(:adapter)          { Valkyrie::MetadataAdapter.find(:test_adapter) }
+  let(:persister)        { adapter.persister }
+  let(:query_service)    { adapter.query_service }
+  let(:collection_class) { Hyrax::PcdmCollection }
+
+  describe 'collection membership' do
+    it 'is in no collections default' do
```
jeremyf: The wording of this doesn't quite read right. Do you mean `it 'is in no collections by default'`?

app/services/hyrax/custom_queries/find_access_control.rb
tpendragon: Via `find`? They definitely should not raise `ArgumentError`, and that'd be a bug. I'm pretty certain it's even in the shared_specs.

lib/hyrax/specs/shared_specs/hydra_works.rb
```diff
@@ -26,6 +26,63 @@
...
+      expect(model.member_of_collection_ids).to be_empty
+    end
+
+    it 'can be added to collections' do
+      collection_ids = [Valkyrie::ID.new('coll_1'), Valkyrie::ID.new('coll_2')]
+
+      expect { model.member_of_collection_ids = collection_ids }
+        .to change { model.member_of_collection_ids }
```
jeremyf: Do we need a `.from` assertion as well?

app/services/hyrax/custom_queries/find_access_control.rb
no-reply: via `.find_inverse_references_by`

lib/hyrax/specs/shared_specs/hydra_works.rb
no-reply:
```suggestion
    it 'is in no collections by default' do
```

spec/services/hyrax/custom_queries/navigators/collection_members_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    it 'is empty with no collections' do
+      expect(query_handler.find_collections_for(resource: work))
+        .to be_empty
+    end
+
+    context 'when it is a member of collections' do
+      let(:work) { FactoryBot.valkyrie_create(:hyrax_work, :as_member_of_multiple_collections) }
+
+      it 'finds collections' do
+        expect(query_handler.find_collections_for(resource: work).map(&:id))
```
jeremyf: Do we have an assertion around `:as_member_of_multiple_collections` functioning? I can see that if the factory has an empty array for collection membership, then this spec would succeed.

lib/hyrax/specs/shared_specs/hydra_works.rb
no-reply: it doesn't seem relevant to me, but also would most likely be harmless.

if you make the suggestion i'll commit it.

spec/services/hyrax/custom_queries/navigators/collection_members_spec.rb
no-reply: fair point. i could add a separate spec for `expect(query_handler.find_collections_for(resource: work)).not_to be_empty`

other suggestions?

lib/hyrax/specs/shared_specs/hydra_works.rb
jeremyf: The spec reads well. There is an assumption that change has happened, and I don't quite know how RSpec treats change without the full `.from` and `.to` preamble. I'm okay with it as is.

spec/services/hyrax/custom_queries/navigators/collection_members_spec.rb
jeremyf: One pattern that I had tried to work through was having a factory spec area, where we test the behavior of factories. Some of them can be rather gnarly. I don't know if that is part of the code-base. That also seems like a much larger exercise.

I believe a separate spec (or even assertion within spec) and a comment saying something along the lines of "Verifying working factory" (hmm... Maybe we need an `OSHA` gem?)

no-reply:
I'll let this hang here in case anyone else wants to have a peek, and probably merge in the AM.

jeremyf:
@no-reply Sounds good.

app/services/hyrax/custom_queries/find_access_control.rb
jeremyf: @tpendragon or @no-reply Any further thoughts/discussions/necessary issues to submit?

tpendragon: @no-reply Could you make a ticket in Valkyrie with the details? `find_inverse_references_by` should be returning an empty array.

PR #4263, opened 2020-02-07
Support proxy deposit for Valkryie Work models

no-reply:
Our proxy deposit implementation requires metadata for `proxy_depositor` and
`on_behalf_of`.

There's some question about the need to store this data on the curation
metadata, but our existing metadata holds it, so it's important to keep it here
for now.

@samvera/hyrax-code-reviewers


app/models/hyrax/work.rb
```diff
@@ -1,4 +1,4 @@
-# frozen_string_literal: true
+ # frozen_string_literal: true
```
no-reply:
```suggestion
# frozen_string_literal: true
```

PR #4264, opened 2020-02-07
Use Valkyrie::ID<=>String equality

no-reply:
Valkyrie 2.1.0 overhauled ID equality as a feature-flag change, with the new
behavior to become default in 3.0.0. Turning it on preemptively for Hyrax will
save us and adopters a lot of trouble going forward.

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/initializers/valkyrie_id_equality.rb
+++ b/config/initializers/valkyrie_id_equality.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+Valkyrie.config.id_string_equality = true
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -77,7 +77,7 @@ SUMMARY
   spec.add_dependency 'select2-rails', '~> 3.5'
   spec.add_dependency 'signet'
   spec.add_dependency 'tinymce-rails', '~> 4.1'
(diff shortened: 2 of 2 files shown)


no-reply:
Hold this pending a resolution to https://github.com/samvera/valkyrie/issues/820

jeremyf:
I like! 👍

no-reply:
do you think we need to hold this until a release with some fix for samvera/valkyrie#820, or should we roll forward now and let the valkyrie fix come in its own time?

jeremyf:
@no-reply Given the caveat that you posted about `(A == B) == true` but `(B == A) == false` I think we should hold.

no-reply:
bumping this

jeremyf:
@no-reply I added another commit to this. For the record, I approve of your commit.

## Narrowing Valkyrie dependency range

d102a03a460629cd983654d99bef7b44fe4fac05

For this PR to work for upstream adopters, we need to ensure that our
gemspec includes the [upstream Valkyrie commit][1].

From the valkyrie repository, I ran:

```console
$ git tag --contains 343656c78411cc0788a8f723f8332d2e697213fe
=> v2.1.1
```

[1]:https://github.com/samvera/valkyrie/commit/343656c78411cc0788a8f723f8332d2e697213fe

PR #4265, opened 2020-02-08
Try to get a grip on the interfaces for `Hyrax::Event` 

no-reply:
It's important that we grapple with how these events are used. A simple starting
point is to try to document it. Maybe some tests next.

Also:

Pin the namespacing for events to the model name, not the Ruby class name. In
most cases (excepting adopters who have specificially overridden ActiveModel
naming behavior) this should have no impact, but loosens the dependency on
specific Ruby class names.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/with_events.rb
+++ b/app/models/concerns/hyrax/with_events.rb
@@ -13,6 +13,10 @@ module Hyrax
       model_name.name
     end
 
+    ##
+    # @param [Integer] size  the maximum number of events to fetch from the log.
+    #   Offset by 1; to get one event, use `0`, for two, use `1`, etc...
+    #   `-1` gives all events.
     def events(size = -1)
       event_stream.fetch(size)
     end
--- a/app/models/concerns/hyrax/with_events.rb
+++ b/app/models/concerns/hyrax/with_events.rb
@@ -10,7 +10,7 @@ module Hyrax
 
     # @return [String]
     def event_class
-      self.class.name
+      model_name.name
     end
 
     def events(size = -1)
--- a/app/models/concerns/hyrax/with_events.rb
+++ b/app/models/concerns/hyrax/with_events.rb
@@ -1,9 +1,14 @@
 module Hyrax
+  # A mixin module intended to provide an interface into creating the paper trail
+  # of activity on the target model of the mixin.
+  #
+  # @see Hyrax::Event
   module WithEvents
     def stream
       Nest.new(event_class)[to_param]
     end
 
+    # @return [String]
     def event_class
       self.class.name
     end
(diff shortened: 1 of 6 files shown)


app/models/concerns/hyrax/with_events.rb
```diff
@@ -5,7 +5,7 @@ def stream
     end
 
     def event_class
-      self.class.name
+      model_name.name
```
jeremyf: I'm wondering about an alias of `event_class` and then deprecation of `event_class`. When I see a method `*_class` my assumption is that I have a `Class` object. However, in reading this, that is not the case. We have a String.

app/models/concerns/hyrax/with_events.rb
```diff
@@ -1,11 +1,16 @@
 module Hyrax
+  # A mixin module intended to provide an interface into creating the paper trail
+  # of activity on the target model of the mixin.
+  #
+  # @see Hyrax::Event
   module WithEvents
```
jeremyf: I was looking at the structure of this module. I find it peculiar that the methods of this module call a private method `event_stream` which in turn calls a public method. Can we "privatize" `stream`? I think I'll create a PR on this branch to test that very thing.

app/models/concerns/hyrax/with_events.rb
no-reply: yeah. i expect there's more work to do here. `event_namespace` or `event_type` would make more sense.

but i'm also trying to take things step by step and just understand what's really happening here. these events are (almost?) exclusively published in jobs which have a hard-to-follow inheritance structure.

this one change seemed important to me because i'm worried about tight class name dependencies in the AF->Valkyrie transition.

also, it was end-of-day friday, so i submitted the scope of what i actually had done/ready. :)

no-reply:
what do you think about squashing the two newer documentation commits into the first one and pushing this forward for further iteration

I also have some wIp at https://github.com/samvera/hyrax/tree/redis-events-2 that might be relevant if you're jumping into this part of the codebase.

jeremyf:
I'm all for the squash. I'll defer to you on that, as I've only done those commits via the Github interface.

app/models/concerns/hyrax/with_events.rb
jeremyf: Yeah. I was looking briefly at this, and its a Mixin that would benefit with a class to encapsulate the public behavior. However, as it stands, these PRs move the shared understanding forward, so "Make it so!"

lib/hyrax/redis_event_store.rb
```diff
@@ -15,8 +14,13 @@ def create(action, timestamp)
         nil
       end
 
-      delegate :logger, to: Hyrax
-
+      ##
+      # @api private
```
no-reply:
```suggestion
      # @api public
```

PR #4266, opened 2020-02-08
Privatizing a method with deprecation warning

jeremyf:
Thoughts on this? @no-reply


no-reply:
This seems good. I also have a refactor in progress that would deprecate all of these methods and bump this logic out to a service. 

None of this interface seems to have unit tests, which is making it slower going than i'd like

jeremyf:
A service class would be great. Encodes what we are expecting and starts creating a crisp interface. I'll defer to you.

I'm wrestling with a new computer and frankly I have a lot of stickers to add to my machine. :)

Issue #4193, opened 2020-01-22
Allow ContentEventJob to receive either an ActiveFedora object or a Valkyrie resource


randalldfloyd:
### Descriptive summary
Convert ContentEventJob to allow it to receive either an ActiveFedora based object or a Valkyrie::Resource equivalent object.

### Rationale
Allow jobs to execute correctly with existing AF implementation or the Wings Valkyrie implementation.

### Expected behavior
ContentEventJob receives a Valkyrie::Resource and performs the action.

### Actual behavior
ContentEventJob raises an exception when passed a Valkyrie::Resource.

### Approach for Conversion
[See Pattern for Jobs Receiving Parameter that may be an AF Object or a Valkyrie Resource](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727)

### Related work

These have all be modified to support AF object or Valkyrie resource using the [established pattern](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727). 

* Issue #4101 Allow GrantEditToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4102 Allow GrantReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4103 Allow RevokeReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4099 Add GlobalID::Identification for resources passed to jobs (PR #4124)

jeremyf:
The ContentEventJob has 6 subclasses:

- [ ] ContentUpdateEventJob
- [ ] ContentNewVersionEventJob
- [ ] FileSetAttachedEventJob
- [ ] ContentRestoredVersionEventJob
- [ ] ContentDepositorChangeEventJob
- [ ] ContentDepositEventJob

This requires checking their implementations, but I suspect they are almost entirely a "noop"

jeremyf:
The `Hyrax::WithEvents#log_events` ([see code](https://github.com/samvera/hyrax/blob/a8597884e1ee1802119ee9e82cb7ae9e52ee203f/app/models/concerns/hyrax/with_events.rb#L25-L27)) exposes the common logging method.

Below are the five places that mixin the `Hyrax::WithEvents` module:

[...]

PR #4270, opened 2020-02-18
Valkyrize Event Jobs

### LINKED ISSUE #4202: Allow ContentRestoredVersionEventJob to receive either an ActiveFedora object or a Valkyrie resource
### Descriptive summary
Convert ContentRestoredVersionEventJob to allow it to receive either an ActiveFedora based object or a Valkyrie::Resource equivalent object.

### Rationale
Allow jobs to execute correctly with existing AF implementation or the Wings Valkyrie implementation.

### Expected behavior
ContentRestoredVersionEventJob receives a Valkyrie::Resource and performs the action.

### Actual behavior
ContentRestoredVersionEventJob raises an exception when passed a Valkyrie::Resource.

### Approach for Conversion
[See Pattern for Jobs Receiving Parameter that may be an AF Object or a Valkyrie Resource](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727)

### Related work

These have all be modified to support AF object or Valkyrie resource using the [established pattern]((https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727). 

* Issue #4101 Allow GrantEditToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4102 Allow GrantReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4103 Allow RevokeReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4099 Add GlobalID::Identification for resources passed to jobs (PR #4124)

bkeese:
Fixes #4202, #4193, #4194 

Valkyrizes ContentRestoredVersionEventJob
Valkyrizes ContentUpdateEventJob
Valkyrizes ContentDepositorChangeEventJob

@samvera/hyrax-code-reviewers


app/models/concerns/hyrax/with_depositor.rb
```diff
@@ -0,0 +1,11 @@
+module Hyrax::WithDepositor
+  # Adds metadata about the depositor to the asset and
+  # grants edit permissions to the +depositor+
+  # @param [String, #user_key] depositor
+  def apply_depositor_metadata(depositor)
```
no-reply: i think we should avoid putting this business logic into the model code. maybe the transaction step `SetUserAsDepositor` could be reused here?

spec/jobs/content_depositor_change_event_job_spec.rb
```diff
@@ -14,15 +13,38 @@
...
-                                                .and change { another_user.events.length }.by(1)
-                                                                                          .and change { generic_work.events.length }.by(1)
+  context "when use_valkyrie is false" do
+    let(:generic_work) { create(:generic_work, title: ['BethsMac'], user: user) }
 
-    expect(user.profile_events.first).to eq(event)
-    expect(another_user.events.first).to eq(event)
-    expect(generic_work.events.first).to eq(event)
+    it "logs the event to the proxy depositor's profile, the depositor's dashboard, and the FileSet" do
+      expect do
```
no-reply: could we clean this up? maybe something like

```ruby
expect { described_class.perform_now(generic_work, another_user) }
  .to change { user.profile_events.length }
  .by(1)
  .and change { another_user.events.length }
  .by(1)
  .and change { generic_work.events.length }
  .by(1)
```

spec/jobs/content_depositor_change_event_job_spec.rb
```diff
@@ -14,15 +13,38 @@
...
+      expect(generic_work.events.first).to eq(event)
+    end
+  end
+
+  context "when use_valkyrie is true" do
+    let(:generic_work) { valkyrie_create(:hyrax_work, title: ['BethsMac']) }
+
+    it "logs the event to the proxy depositor's profile, the depositor's dashboard, and the FileSet" do
+      generic_work.apply_depositor_metadata(user)
+      allow_any_instance_of(described_class).to receive(:hyrax_test_simple_work_path).and_return("/concern/generic_works/#{generic_work.id}")
```
no-reply: do we need `allow_any_instance_of`? if we're stubbing this (+0 to that---if it gets around the other issues here, i'm okay with it), could we stub it on the current instance and call `#perform` directly instead of routing through `.perform_now`?

app/models/hyrax/resource.rb
```diff
@@ -32,6 +32,7 @@ module Hyrax
   #
   class Resource < Valkyrie::Resource
     include Hyrax::Naming
+    include Hyrax::WithEvents
```
jeremyf: How does this interact with the pull request? I'm not seeing specs surrounding changes to Hyrax::Resource getting "with_events"

bkeese: All of the event jobs rely on the ability to track and log those events for the affected resources. The Hyrax::WithEvents mixin adds that capability and is central to all of the specs for those jobs.

jeremyf: The answer makes sense. My surprise came when I didn't see a change in spec to `hyrax/resource_spec.rb`

I'm wondering if there is a way to capture the expectation that Hyarx::Resource includes Hyrax::WithEvents. I know this relates to #4265 .

jeremyf: I should add, that this is not a blocker for the PR. Just musing and wondering.

bkeese: Interesting. I somehow missed the whole conversation around 4265. I'll look into adding something that directly tests the new event requirement for Hyrax::Resource.

bkeese:
Bumping

PR #4275, opened 2020-03-05
Adding Caching to IIIF Manifest Generation

jeremyf:
This commit is to push something up for conversation. I don't want to
get too far in the solution without community feedback.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -10,10 +10,11 @@ module Hyrax
       with_themed_layout :decide_layout
       copy_blacklight_config_from(::CatalogController)
 
-      class_attribute :_curation_concern_type, :show_presenter, :work_form_service, :search_builder_class
+      class_attribute :_curation_concern_type, :show_presenter, :work_form_service, :search_builder_class, :iiif_manifest_builder
       self.show_presenter = Hyrax::WorkShowPresenter
       self.work_form_service = Hyrax::WorkFormService
       self.search_builder_class = WorkSearchBuilder
+      self.iiif_manifest_builder = Hyrax::ManifestBuilderService
       attr_accessor :curation_concern
       helper_method :curation_concern, :contextual_path
 
@@ -127,7 +128,7 @@ module Hyrax
     def manifest
       headers['Access-Control-Allow-Origin'] = '*'
 
-      json = Hyrax::ManifestBuilderService.as_json(presenter: presenter)
+      json = iiif_manifest_builder.as_json(presenter: presenter)
 
       respond_to do |wants|
         wants.json { render json: json }
@@ -136,6 +137,10 @@ module Hyrax
     end
 
     private
+
+      def iiif_manifest_builder
+        self.class.iiif_manifest_builder
+      end
 
       def user_collections
         collections_service.search_results(:deposit)
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -127,7 +127,7 @@ module Hyrax
     def manifest
       headers['Access-Control-Allow-Origin'] = '*'
 
-      json = sanitize_manifest(JSON.parse(manifest_builder.to_h.to_json))
+      json = Hyrax::ManifestBuilderService.as_json(presenter: presenter)
 
       respond_to do |wants|
         wants.json { render json: json }
@@ -157,10 +157,6 @@ module Hyrax
 
       def build_form
         @form = work_form_service.build(curation_concern, current_ability, self)
-      end
-
-      def manifest_builder
-        ::IIIFManifest::ManifestFactory.new(presenter)
       end
 
       def actor
@@ -325,22 +321,6 @@ module Hyrax
 
       def permissions_changed?
         @saved_permissions != curation_concern.permissions.map(&:to_hash)
-      end
-
-      def sanitize_manifest(hash)
-        hash['label'] = sanitize_value(hash['label']) if hash.key?('label')
-        hash['description'] = hash['description']&.collect { |elem| sanitize_value(elem) } if hash.key?('description')
-
-        hash['sequences']&.each do |sequence|
-          sequence['canvases']&.each do |canvas|
-            canvas['label'] = sanitize_value(canvas['label'])
-          end
-        end
-        hash
-      end
-
-      def sanitize_value(text)
-        Loofah.fragment(text.to_s).scrub!(:prune).to_s
       end
   end
 end
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -14,7 +14,7 @@ module Hyrax
       self.show_presenter = Hyrax::WorkShowPresenter
       self.work_form_service = Hyrax::WorkFormService
       self.search_builder_class = WorkSearchBuilder
-      self.iiif_manifest_builder = Hyrax::ManifestBuilderService
+      self.iiif_manifest_builder = (Flipflop.cache_work_iiif_manifest? ? Hyrax::CachingIiifManifestBuilder.new : Hyrax::ManifestBuilderService.new)
       attr_accessor :curation_concern
       helper_method :curation_concern, :contextual_path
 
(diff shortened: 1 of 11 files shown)


randalldfloyd:
> I think this looks really good, but I'm wondering if there is any way to do the manifest generation so that the first request doesn't take a really long time. @randalldfloyd said he's seen it take minutes to generate the manifest with a multi-volume work.

We've always just let the first user take the hit after initial creation or subsequent save.  That's usually the work creator when redirected back to the show page, which probably has them wondering at first if something didn't work right. I wonder if we could take care of the first stash of the manifest to cache as part of the delayed jobs in the save cycle.

devanshu-m:
> I think this looks really good, but I'm wondering if there is any way to do the manifest generation so that the first request doesn't take a really long time. @randalldfloyd said he's seen it take minutes to generate the manifest with a multi-volume work.

This is not a complete solution: We have a rake task in place for generating manifests and caching them, however, this code should be moved to a job (https://github.com/emory-libraries/dlp-curate/blob/master/app/lib/create_manifest.rb). This way once the job completes and all work manifests are cached, users will not need to wait for their manifest even for the first request. The challenge with such a rake task would be the ability and request for the presenter. Presenter initialization takes the ability of the current_user and creates the manifest, and since with the rake task we won’t have a current user, we are creating manifest with a read all ability which might not be something everyone would want (https://github.com/emory-libraries/dlp-curate/blob/master/app/models/manifest_ability.rb). One possible solution here would be to create restrictions for manifest access after it has been generated. Another issue would be the `request` object for presenter creation which takes `nil` as default. With the rake task, we will need a way to create the manifest_url, like this: https://github.com/emory-libraries/dlp-curate/blob/master/app/presenters/hyrax/curate_generic_work_presenter.rb#L25
During presenter initialization manifest_url creation is done using the base_url for the request.

jeremyf:
@devanshu-m Thank you for the input. It's certainly something to consider for future work.

orangewolf:
in a similar vain we did the following to first hit caching of iiif on DORIS
https://github.com/notch8/doris-hyku/blob/master/lib/tasks/riiif.rake

```
namespace :riiif do
  desc 'cache riiif images'
  task :cache => [:environment] do
    # load 'config/initializers/riiif.rb'
    progress = ProgressBar.create(total: Image.count)
    logger = ActiveSupport::TaggedLogging.new(Logger.new(Rails.root.join('log', 'riiif.log')))
    Image.find_each do |i|
      begin
        image_params = {"region"=>"full", "size"=>"600,", "rotation"=>"0", "quality"=>"default", "format"=>"jpg"}.with_indifferent_access
        image_id = i&.file_sets&.first&.files&.first&.uri&.to_s&.split('/')&.[](-3..-1)&.join('/')
        if image_id
          r = Riiif::Image.new(image_id)
          r.render(image_params)
        end
      rescue => e
        puts "#{i.id}: #{e.message}"
        logger.error "#{i.id}: #{e.message}\n#{e.backtrace}"
      end
      progress.increment
    end
  end
end
```

orangewolf:
I had plans at one point to implement an actor / background job to do this on file change. But I personally think this is valuable as a set of work w/o that and the automating of cache creation can be a separate ticket.

jeremyf:
Keeping this alive. Don't want the stalebot to drop the hammer.

jeremyf:
Pinging @no-reply

no-reply:
it's not clear to me whether folks are intending to block this caching pending pre-warming of some kind?

i can't see the logic in that, so i'm hoping someone will chime in if there's a reason not to push forward with this.

app/services/hyrax/manifest_builder_service.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    attr_reader :presenter, :manifest_builder, :expires_in, :cache_manifest
+    alias cache_manifest? cache_manifest
+
+    # @api private
+    #
+    # @return [Hash] a Ruby hash representation of a IIIF manifest document
+    def as_json
+      if cache_manifest?
+        Rails.cache.fetch(manifest_cache_key, expires_in: expires_in) do
```
no-reply, line 55: tempted to rework this as a subclass like `CachingManifestBuilderService`, to avoid the branching.

thoughts, @jeremyf?

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -127,7 +127,7 @@ def inspect_work
     def manifest
       headers['Access-Control-Allow-Origin'] = '*'
 
-      json = sanitize_manifest(JSON.parse(manifest_builder.to_h.to_json))
+      json = Hyrax::ManifestBuilderService.as_json(presenter: presenter)
```
no-reply, line 130: let's make this configurable on the controller level?

app/services/hyrax/manifest_builder_service.rb
```diff
@@ -0,0 +1,93 @@
...
+
+module Hyrax
+  # A class responsible for converting a Hyrax::Work like thing into a IIIF
+  # manifest.
+  #
+  # @see !{.as_json}
+  class ManifestBuilderService
+    # @api public
+    #
+    # @param presenter [Hyrax::WorkShowPresenter] the work presenter from which
```
no-reply, line 11: the documentation above says this works for `Hyrax::Work`-like objects. what's the correct type signature?

app/services/hyrax/manifest_builder_service.rb
```diff
@@ -0,0 +1,93 @@
...
+    # @param iiif_manifest_factory [Class] a class that initializes with presenter
+    #        object and returns an object that responds to `#to_h`
+    # @param expires_in [Integer] the number of seconds until the cache expires
+    # @param cache_manifest [Boolean] should we perform Rails level caching of
+    #        the generated manifest
+    #
+    # @note While the :presenter may be a Hyrax::WorkShowPresenter it is likely
+    #       defined by Hyrax::WorksControllerBehavior.show_presenter
+    #
+    # @note Why the public class method? It is easier to perform stubs in
```
no-reply, line 22: it seems a bit rigid compared with

```ruby
  class_attribute :iiif_manifest_builder
  self.iiif_manifest_builder = Hyrax::ManifestBuilderService.new

  ...
  
  def manifest
    self.class.iiif_manifest_builder.as_json(presenter: presenter)
  end
```

in the above case, if i want to configure my builder on the class level, i have the option to pass arguments at the initializer without changing calls in the business logic. 

e.g. if i want to turn off caching for one controller presently (or in the tests), i'd need to pass `cache_manifest: false` every time i call `#as_json`. with an instance, i can do `MyController.iiif_manifest_builder = Hyrax::ManifestBuilderService.new(cache_manifest: false)` and leave other code untouched. the API footprint for a generic controller is reduced here (at least nominally) since the `#as_json` method now doesn't need to know anything about caching. Liskov compliant subclasses are easier to implement, in this case.

orangewolf:
> it's not clear to me whether folks are intending to block this caching pending pre-warming of some kind?
> 
> i can't see the logic in that, so i'm hoping someone will chime in if there's a reason not to push forward with this.

I agree, getting this in and following on with pre-warm makes sense to me.

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
```diff
@@ -172,4 +172,26 @@
       it_behaves_like 'allows show access'
     end
   end
+
+  describe '#manifest' do
+    include_context 'with a logged in user'
+    xit 'resolves json' do
+      # When I attempt to get this working, I encounter the following error:
```
no-reply, line 179: probably worth testing this in `generic_works_controller_spec.rb` instead, rather than holding this feature up on in-progress infrastructure.

app/services/hyrax/manifest_builder_service.rb
jeremyf: I don't think a subclass is worth the effort at this point. There's very logic variance.

jeremyf:
I've reviewed @no-reply's changes, and am good with those. Maybe @orangewolf or @randalldfloyd or @cjcolvar or @devanshu-m could review?

app/services/hyrax/manifest_builder_service.rb
no-reply: i think the answer to this really is `Hyrax::WorkShowPresenter`, at least for now. maybe we need to edit the top level class documentation?

PR #4277, opened 2020-03-23
Gh 4193 content event job

jeremyf:
## Adding explicit parameter to `super` call

dcdb2d42caaa1210beafa87a3d3109eb51c7bb6e

Prior to this commit the perform method had an explicit call to super
with an implicit passing of parameters. This commit adds an explicit
passing of those parameters.

## Refactoring to unify a subsystem of Event behavior

a182c855f52fd4aa5de4ed2d2e9fd4c0173e63ca

Prior to this commit, the descendants of ContentEventJob each
approached writing their `action` string differently. With this
change, I've normalized that behavior. There is still work to be done
to get a handle on the whole event ecosystem.

Relates to: #4193

### DIFF
--- a/app/jobs/content_deposit_event_job.rb
+++ b/app/jobs/content_deposit_event_job.rb
@@ -1,6 +1,6 @@
 # Log a concern deposit to activity streams
 class ContentDepositEventJob < ContentEventJob
   def action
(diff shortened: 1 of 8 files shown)


no-reply:
bumping so i can review

PR #4287, opened 2020-04-16
Make permission level configuration defaults use I18n locale.

### LINKED ISSUE #4285: Permission options cannot be translated
### Descriptive summary

The labels of the permissions options, in the work form, are not locale sensitive. 

### Rationale

The user should see "Choose Access",  "View/Download" , "Edit" when navigating in english and "Choisir l'accès", "Consulter/télécharger", "Modifier" when navigating in french.

### Expected behavior

Label changes when user switches locale.

### Actual behavior

Label is allways in english in vanilla Hyrax or allways in another language if you translate it through configuration files (locale files are not involved).

- Create or update a work
- Go to "sharing" tab
- Change the locale

"User" and "groups" right levels select menus are not translated
 
### Analysis

Permissions labels are hardcoded in https://github.com/samvera/hyrax/blob/625bcfdb7a70c752d19c9e8b9b43e8a9b945a515/lib/hyrax/configuration.rb#L491.

They are directly passed to "options_for_select" without being translated first.
https://github.com/samvera/hyrax/blob/625bcfdb7a70c752d19c9e8b9b43e8a9b945a515/app/views/hyrax/base/_form_share.html.erb#L44

They can be changed through hyrax.rb but english is lost.

```ruby
# Labels for display of permission levels
  config.permission_levels = {"Consulter/Télécharger" => "read", "Modifier" => "edit"}

  # Labels for permission level options used in dropdown menus
  config.permission_options = {"Choisir un niveau de droits" => "none", "Consulter/Télécharger" => "read", "Modifier" => "edit"}

  # Labels for owner permission levels
  config.owner_permission_levels = {"Modifier" => "edit"}
```

cjcolvar:
Any overrides of these defaults in initializers would remove the ability to support multiple languages as pointed out by @j-dornbusch in #4285.  Thus I'm not exactly sure of the value of these configuration methods if the values have been put in the locale, but I left them as-is to be backwards compatible.

Fixes #4285

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.en.yml
+++ b/config/locales/hyrax.en.yml
@@ -1283,6 +1283,15 @@ en:
         terms_page: Terms of Use
       updated: Pages updated.
     passive_consent_to_agreement: By saving this work I agree to the
+    permission_levels:
+      edit: "Edit access"
+      read: "View/Download"
+      owner:
+        edit: "Edit access"
+      options:
+        edit: "Edit"
+        none: "Choose Access"
+        read: "View/Download"
     search:
       button:
         html: <span class="glyphicon glyphicon-search" aria-hidden="true"></span> Go
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -488,20 +488,20 @@ module Hyrax
 
     attr_writer :permission_levels
     def permission_levels
-      @permission_levels ||= { "View/Download" => "read",
-                               "Edit access" => "edit" }
+      @permission_levels ||= { I18n.t('hyrax.permission_levels.read') => "read",
+                               I18n.t('hyrax.permission_levels.edit') => "edit" }
     end
 
     attr_writer :owner_permission_levels
     def owner_permission_levels
-      @owner_permission_levels ||= { "Edit access" => "edit" }
+      @owner_permission_levels ||= { I18n.t('hyrax.permission_levels.owner.edit') => "edit" }
     end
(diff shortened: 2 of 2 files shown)


j-dornbusch:
Here is a proposal for a French translation.

```
    permission_levels:
      edit: "Modification" 
      read: "Consultation" 
      owner:
        edit: "Modification" 
      options:
        edit: "Modifier" 
        none: "Choisir un niveau de droits" 
        read: "Consulter/télécharger" 
```
@jmfrenzel may have the german wording.

no-reply:
rebased and pushed, without the regen, which has been patched in #4289

cjcolvar:
> Here is a proposal for a French translation.
> 
> ```
>     permission_levels:
>       edit: "Modification" 
>       read: "Consultation" 
>       owner:
>         edit: "Modification" 
>       options:
>         edit: "Modifier" 
>         none: "Choisir un niveau de droits" 
>         read: "Consulter/télécharger" 
> ```
> 
> @jmfrenzel may have the german wording.

@j-dornbusch This PR was merged before I addd your French translation.  Do you mind making a Pull Request for your translation or maybe wait to fix all of the Google-translated strings at one time after the i18n script has been run?

no-reply:
i'm realizing that these will still be 'hard coded', just to whatever the locale is when the app starts up. i think we need to rip out these configuration methods altogether.

PR #4288, opened 2020-04-23
Configure Redis with ENV variables if present

no-reply:
adds the `REDIS_HOST` and `REDIS_PORT` ENV variables to allow applications to
configure in keeping with 12-factor. defaults are provided for dev and test
environments, while production fails eagerly if configuration is missing.

this will only affect new applications, since this configuration is generated
into the application when hyrax is installed.

@samvera/hyrax-code-reviewers


no-reply:
next, something sane for solr. :)

PR #4292, opened 2020-04-25
implement a solr health check

no-reply:
this check fails if it can't reach Solr, and succeeds otherwise.

@samvera/hyrax-code-reviewers


lib/hyrax/health_checks/solr_check.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Hyrax
+  module HealthChecks
+    class SolrCheck < OkComputer::Check
+      def initialize(service: Hyrax::SolrService)
+        @service = service
+      end
+
+      def check
+        @service.get
```
cjcolvar, line 11: If this is doing the equivalent of an empty keyword search then it might be more efficient to execute a ping instead: https://lucene.apache.org/solr/guide/6_6/ping.html

In the Avalon solr check, we're doing something like:
```
ActiveFedora::SolrService.instance.conn.get('admin/ping')['status'] == 'OK'
```

no-reply: `instance` is deprecated, to avoid tight coupling with ActiveFedora.

it may be that this service needs a `ping` of its own?

PR #4294, opened 2020-04-28
remove the indexer initializer generator

no-reply:
the code this generator injects isn't really necessary (if we want to register
this indexer for the generic `Hyrax::Work`, we can do it in Hyrax
directly). including it complicates the upgrade process for existing
apps. instead, while we need to manually register indexers for specific
works (maybe we can do this with configuration soon, instead?), inject them into
the hyrax initializer when the work generator runs.

additionally, generate basic metadata into the indexer.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/work_resource/templates/indexer.rb.erb
+++ b/lib/generators/hyrax/work_resource/templates/indexer.rb.erb
@@ -3,7 +3,7 @@
 # Generated via
 #  `rails generate hyrax:work_resource <%= class_name %>`
 class <%= class_name %>Indexer < Hyrax::ValkyrieWorkIndexer
-  Hyrax::ValkyrieIndexer.register self, as_indexer_for: <%= class_name %>
+  include Hyrax::Indexer(:basic_metadata)
 
   # Uncomment this block if you want to add custom indexing behavior:
   #  def to_solr
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -16,6 +16,7 @@ require 'hydra/works'
 require 'hyrax/engine'
 require 'hyrax/version'
 require 'hyrax/inflections'
+require 'hyrax/name'
 require 'hyrax/valkyrie_can_can_adapter'
 require 'kaminari_route_prefix'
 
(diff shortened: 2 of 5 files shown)


lib/hyrax.rb
```diff
@@ -16,6 +16,7 @@
 require 'hyrax/engine'
 require 'hyrax/version'
 require 'hyrax/inflections'
+require 'hyrax/name'
```
jeremyf, line 19: I'm unclear why this is added.

lib/generators/hyrax/work_resource/templates/indexer.rb.erb
```diff
@@ -3,7 +3,7 @@
 # Generated via
 #  `rails generate hyrax:work_resource <%= class_name %>`
 class <%= class_name %>Indexer < Hyrax::ValkyrieWorkIndexer
-  Hyrax::ValkyrieIndexer.register self, as_indexer_for: <%= class_name %>
+  include Hyrax::Indexer(:basic_metadata)
```
jeremyf, line 6: With this change, will those that ran previous generators need to change their registration logic? What is the process for people to do that? Do we have documentation in the upgrade notes?

lib/hyrax.rb
no-reply: it turned out that actually running this generator and starting an app caused `NameError` on this module. it seems like it should have just always been there.

lib/generators/hyrax/work_resource/templates/indexer.rb.erb
no-reply: > With this change, will those that ran previous generators need to change their registration logic?

yes.

> What is the process for people to do that? Do we have documentation in the upgrade notes?

this code has never been released (even in an RC), so i think we might be in the clear here? i'm open to counter views on that point, though.

no-reply: i guess i'll also add: they only need to make this change if they want `:basic_metadata`.

if they *do* want that, it's likely they've already made this change.

PR #4296, opened 2020-05-05
[WIP] Don't override ActiveFedora::Base#to_s

cjcolvar:
This is just a test to see what breaks if we remove these overrides.


no-reply:
some of the behavior that falls down here is worked around in #4297

cjcolvar:
@no-reply Is it worth rebasing this and having the tests rerun?

PR #4298, opened 2020-05-06
use helpers to i18nize configured permission levels/options

no-reply:
these configured options (why do they need to be configurable? how would one
configure them successfully?) need to be i18nized, but the current fix only
caches the i18n strings at app startup (or maybe when the configurations are
first hit?).

to translate on-the-fly we need to hit the i18n key while the user is in the
desired locale.

this is a stop-gap fix. probably we want to make the configuration options just
straight arrays, or remove them completely. discussion on longer term approachs
is scheduled in https://wiki.lyrasis.org/display/samvera/Samvera+Tech+Call+2020-06-05.

related to #4297.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/permission_levels_helper.rb
+++ b/app/helpers/hyrax/permission_levels_helper.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module PermissionLevelsHelper
+    ##
+    # @return [Hash<String, String>] a map from i18nized strings to permission
+    #   level keywords
+    def configured_permission_levels
+      Hyrax.config.permission_levels.values.each_with_object({}) do |level, hsh|
+        hsh[I18n.t("hyrax.permission_levels.#{level}")] = level
+      end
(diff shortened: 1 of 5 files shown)


no-reply:
> I believe this isn't backwards compatible if anyone has been customizing these by doing something like Hyrax.config.permission_levels = {} in an initializer.

why not? the configs are still used by the helpers.

app/helpers/hyrax/permission_levels_helper.rb
```diff
@@ -0,0 +1,30 @@
+module Hyrax
+  module PermissionLevelsHelper
+    ##
+    # @return [Hash<String, String>] a map from i18nized strings to permission
+    #   level keywords
+    def configured_permission_levels
```
elrayle, line 6: Can we get unit tests for the methods in this class?

no-reply: done!

Issue #4301, opened 2020-05-07
attach_members_actor problem since Hyrax 2.7.1


joelakes:
### Descriptive summary

I updated to Hyrax 2.7.1 this week and am seeing incorrect behavior when editing a work. The child objects are re-added each time I save changes. 

### Rationale

This needs to be fixed because we are now incorrectly adding child objects repetitively, every time we edit a work.

### Expected behavior

When I edit a work to, e.g. add new Keyword, that should be the only change saved.

### Actual behavior

When I edit a work to, e.g. add new Keyword, the Keyword change is saved correctly, but all of the child Works are added again. So the list of children builds up and up every time I edit.

### Steps to reproduce the behavior

1. I created this structure: 
    Subseries (Work) with one child: 
     - Folio (Work) 
2. Then I edit the Subseries (Work), and make a minor change, e.g. add another Keyword. When save the structure has now changed to : 
    Subseries (Work) with one child: 
     - Folio (Work) 
     - Folio (Work) 

### Related work

I believe this was caused by this change:
https://github.com/samvera/hyrax/commit/21be50b9bf989c78d9b18a366c1dcf0effdbf5c8

Byebug shows me that the recent change made in app/actors/hyrax/actors/attach_members_actor.rb appears to be the culprit. 

attributes_collection.each do |attributes| 
    next if attributes['id'].blank? 
    byebug 
    **if boolean_type_caster.cast(attributes['_destroy'])** 
      # Likely someone in the UI sought to add the collection, then 
      # changed their mind and checked the "delete" checkbox and posted 
      # their update. 
      next unless existing_works.include?(attributes['id']) 
      remove(env.curation_concern, attributes['id']) 
    else 
      add(env, attributes['id']) 
    end 
end 

in the code snippet above
**if boolean_type_caster.cast(attributes['_destroy'])** is false so we go to the add. 

Reverting the if else to the following resolves the problem: 
attributes_collection.each do |attributes| 
    next if attributes['id'].blank? 
    byebug 
    if existing_works.include?(attributes['id']) 
       remove(env.curation_concern, attributes['id']) if has_destroy_flag?(attributes) 
    else 
       add(env, attributes['id']) 
    end 
end 
![Screenshot from 2020-05-07 10-09-37](https://user-images.githubusercontent.com/44173462/81279135-9cc18b00-904e-11ea-8e86-76c84f621259.png)

no-reply:
for folks who may be encountering this:

[...]

PR #4302, opened 2020-05-07
Addressing case where children are added again

jeremyf:
Prior to this commit, there was no guard in place to prevent adding
children again.

The logic would check "is something marked for delete" if not, then
add it. That needed to change, because when a user would submit a
form changing keywords, then we would re-add all of the already added
members.

Related to #4301 

@samvera/hyrax-code-reviewers


no-reply:
should the tests in `spec/actors/hyrax/actors/attach_members_actor_spec.rb` at #4303 be ported here as well?

jeremyf:
@no-reply Yeah. I'll work on that this afternoon.

orangewolf:
looks like @no-reply comment is resolved, so this seems gtg

PR #4303, opened 2020-05-07
Addressing case where children are added again

### LINKED ISSUE #4301: attach_members_actor problem since Hyrax 2.7.1

jeremyf:
Prior to this commit, there was no guard in place to prevent adding
children again.

The logic would check "is something marked for delete" if not, then
add it. That needed to change, because when a user would submit a
form changing keywords, then we would re-add all of the already added
members.

Closes #4301 (which relates to #4193 and #4278)
Related to #4302

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/file_set_csv_service.rb
+++ b/app/services/hyrax/file_set_csv_service.rb
@@ -1,3 +1,4 @@
+require 'csv'
 module Hyrax
   #
   # Generates CSV from a FileSet
--- a/app/actors/hyrax/actors/attach_members_actor.rb
+++ b/app/actors/hyrax/actors/attach_members_actor.rb
@@ -41,6 +41,7 @@ module Hyrax
               next unless existing_works.include?(attributes['id'])
               remove(env.curation_concern, attributes['id'])
             else
+              next if existing_works.include?(attributes['id'])
               add(env, attributes['id'])
             end
           end
--- a/app/actors/hyrax/actors/collections_membership_actor.rb
+++ b/app/actors/hyrax/actors/collections_membership_actor.rb
@@ -31,7 +31,7 @@ module Hyrax
         # @param env [Hyrax::Actors::Enviornment]
         # @return [Boolean]
         #
-        # rubocop:disable Metrics/MethodLength
+        # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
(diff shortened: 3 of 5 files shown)


app/services/hyrax/file_set_csv_service.rb
```diff
@@ -1,3 +1,4 @@
+require 'csv'
```
no-reply, line 1: is this related?

PR #4306, opened 2020-05-08
scale down CircleCI parallelism for tests

no-reply:
try scaling up parallelism to speed builds.

@samvera/hyrax-code-reviewers


cjcolvar:
It doesn't look like this had much effect for the increase in credits used.  What do you think @no-reply ?

no-reply:
agreed.

it makes me tempted to try scaling down. i don't think we ever really explored what the optimal setting is.

no-reply:
scaled down seems quite a bit slower. leaving this as-is for now

PR #4307, opened 2020-05-12
Build out Hyrax::ResourceForm to (mostly) support create views

no-reply:
Extends Valkyrie forms to handle visibility, lease/embargo, and permissions (called "sharing" by the relevant form partials).

Ignored here are:

  - configuration for primary/secondary terms
  - form class generation (we'll probably want to generate something into the application (`class MonographForm < Hyrax::ResourceForm(Monograph); end`
  - collection membership and member works
  - feature tests

All of these are future work that could benefit from ongoing discussion.

@samvera/hyrax-code-reviewers


spec/forms/hyrax/forms/resource_form_spec.rb
```diff
@@ -24,6 +24,13 @@
     end
   end
 
+  describe '.model_class' do
+    it 'has a model class' do
```
jeremyf, line 28: `it "is the given work's class"` might read better.

spec/views/hyrax/base/_form.html.erb_spec.rb
```diff
@@ -1,80 +1,127 @@
...
-
-  let(:form) do
-    Hyrax::GenericWorkForm.new(work, ability, controller)
-  end
+  let(:controller_action) { 'new' }
+  let(:controller_class)  { Hyrax::MonographsController }
   let(:options_presenter) { double(select_options: []) }
 
   before do
+    # mock the admin set options presenter to avoid hitting Solr
```
jeremyf, line 7: An ovation for removing a significant chunk of `allow` type mocking!

spec/views/hyrax/base/_form.html.erb_spec.rb
```diff
@@ -1,80 +1,127 @@
...
       end
     end
 
-    context 'with batch_upload off' do
-      before do
-        allow(Flipflop).to receive(:batch_upload?).and_return(false)
+    context 'with an existing object' do
+      let(:work) { FactoryBot.valkyrie_create(:monograph) }
+
+      xit 'renders a form' do
```
jeremyf, line 47: Is this expected? Maybe a comment as to why it's a `xit`

no-reply: there's some discussion on this in the commit message. passing this test requires grappling with `#version`/optimistic locking, which i'm considering out of scope here.

no-reply:
the failing tests are because of `Hyrax.config.curation_concerns.first` being used for "primary_concern", which is a huge mess...

orangewolf:
👍 the specs are mad, and Jeremy's comments are spot on, but over all this is a solid next step. Great work Tom

spec/forms/hyrax/forms/resource_form_spec.rb
no-reply: how is that?

jeremyf: As originally written this sounds like a tautology. To me, it seems that my recommendation explains, a bit better, what is happening.

no-reply: sorry. misunderstanding. :)

i changed the text, and was asking if the new version looks okay to you.

PR #4310, opened 2020-05-14
read form definitions from metadata configuration files

no-reply:
adds form configuration to the `core_metadata` and `basic_metadata` yaml. these configurations are then used to populate the base `ResourceForm` for ChangeSet forms.

these forms no longer add properties from the wrapped Work class automatically. instead, properties are configured explicitly by the yaml itself. as with other yaml metadata configs, parsing the form configuration is handled by a swap-able object `SimpleSchemaLoader`. arbitrary form options can be added to the config, giving applications a lot of flexibility about extending forms and the views that use them.

Hyrax respects `:primary` and `:required` flags. all form fields configured through the config are assumed `:displayable` (we may want to consider allowing an override later?). `:multiple` is included in the configuration here, but not actively supported (yet). note that this is different from the top-level `:multiple` flag, which specifies cardinality of the metadata field itself, not the form inputs; in some cases, these differ.

@samvera/hyrax-code-reviewers


jeremyf:
This! This is the most exciting change! Huzzah to further configured work definition

app/forms/hyrax/forms/resource_form.rb
```diff
@@ -67,6 +68,8 @@ class << self
         #   monograph  = Monograph.new
         #   change_set = Hyrax::Forms::ResourceForm.for(monograph)
         def for(work)
+          "#{work.class}Form".constantize.new(work)
```
no-reply, line 71: i was slightly reluctant to re-introduce this name-inference convention, but decided it was a good-enough starting place to get this work surfaced.

thoughts on how to do this better are welcome.

jeremyf: You could ask the `work.class` for its form (set as a class attribute)

jeremyf: Which might help provide a waypoint when someone looks at the model.

no-reply: except i think forms don't (necessarily) have a 1:1 relationship with a work. they're triangulated with a controller.

hackmastera: For the yaml configuration the field can be defined as primary. that configuration would be used for any form associated with the work?

no-reply: > For the yaml configuration the field can be defined as primary. that configuration would be used for any form associated with the work?

this is a good question.

in the current setup, we have a dependency chain:
  - your form class
  - `ResourceForm`
  - `FormFields`
  - `SimpleSchemaLoader`
  - yaml config

if you wanted multiple forms for a specific work, you could either have multiple schemas like: 

```ruby
MainForm.include Hyrax::FormFields(:main_schema)
SpecializedForm.include Hyrax::FormFields(:specialized_schema)
```

or you could have a single schema two separate form configurations, and a custom loader. something like:

```ruby
MainForm.include Hyrax::FormFields(:schema, definition_loader: MyFormLoader.new(form: :main)
SpecializedForm.include Hyrax::FormFields(:schema, definition_loader: MyFormLoader.new(form: :specialized)
```

in either case, you'd then need to select the appropriate form from the controller. at the moment, this `.for` method does that for `WorksControllerBehavior` controllers; so in some sense there's an assumption that `ResourceForm` is for use by those controllers. whether that's something to stick to, i'm unsure.

hackmastera: 👍 great; this might be good documentation to find a more permanent / visible home for. It's tempting to suggest that different forms could be specified in the configuration, but it seems like probably an uncommon use case, which this process will support in a straightforward way.

app/forms/hyrax/forms/resource_form.rb
```diff
@@ -15,9 +15,7 @@ def self.ResourceForm(work_class)
       Class.new(Hyrax::Forms::ResourceForm) do
         self.model_class = work_class
 
-        (work_class.fields - work_class.reserved_attributes).each do |field|
-          property field, default: nil
-        end
+        include Hyrax::FormFields(:core_metadata)
```
jeremyf, line 18: This line is so much nicer! The indirection helps keep abstraction at the same leve

Issue #4308, opened 2020-05-14
Sorting a collection by relevance doesn't make much sense


sandbergja:
### Descriptive summary

On 2.6.0 (confirmed on nurax running today's master), collection pages allow you to sort the works in them.  These sorts are the same sorts defined in the catalog controller and displayed on the search results page.

However, the default sort -- relevance -- doesn't make any sense here.  The user hasn't entered any search terms, so there's nothing for the Works to be relevant to.

### Rationale

The default behavior is not useful for end users.

### Expected behavior

As an end user, I go to a collection page (like https://libarchive.linnbenton.edu/collections/gm80hv33v?locale=en).  The default sort is something meaningful, like date uploaded descending.

### Actual behavior

As an end user, I go to a collection page (like https://libarchive.linnbenton.edu/collections/gm80hv33v?locale=en).  The default sort is not meaningful (although it might happen to be date uploaded descending, just due to internal workings of the system).

no-reply:
thanks for this!
----
some notes for whoever picks this up:

it looks like these views just use the default blacklight config for `active_sort_fields` (search your catalog controller for `config.add_sort_field`. probably this should be a whole separate config with a whole separate helper infrastructure: i.e. instead of `active_sort_fields` from the [blacklight helpers](https://github.com/projectblacklight/blacklight/blob/24e316398c6617eb18c68fc8dfce36a8b62a4e9d/app/helpers/blacklight/configuration_helper_behavior.rb#L13-L15), call a new `collection_member_sort_fields`.

as a first step, the implementation of that helper could be `active_sort_fields.reject { |field| field.last.label == 'relevance' }` (maybe improve this reject block, the label *shouldn't* be hard coded in english, so that's an issue). users could override the helper pending interest in making this directly configurable via something like the blacklight config.

cc: @samvera/hyrax-repo-managers @samvera/hyrax-ui-ux-advisors is it a good idea to remove "relevance" from this drop down? is "Title [A-Z]" a good default? are the other drop down fields for sort good?

no-reply:
custom sort configuration for these pages made it into `master` via #4311. this still leaves the question of whether we want to change the default configuration.

some discussion in Slack questions whether we should treat this as an issue with collections page at all, or whether it's better considered as an issue with with blacklight's approach to "fallback" sorting. BL implicitly falls back on 'updated' when no search term is present, and we'd like this behavior to be explicitly visible to users via the drop down.

PR #4311, opened 2020-05-15
use a `collection_member_sort_fields` helper to manage sort options

no-reply:
collection show views were using the default blacklight config for
`active_sort_fields`. there's not much point in sorting by relevance, and other
changes to how these sort options present may be desirable on a per-application
basis.

to make it easy to make this kind of change, add a
`collection_member_sort_fields` helper. for now, its implementation is just a
call to the existing helper, so actual behavior is fixed for this commit.

connected to #4308.

@samvera/hyrax-code-reviewers


no-reply:
failures look to be due to fragile stubs in the view code.

can someone look into it.

PR #4312, opened 2020-05-15
move #member_of_collections_json to a helper

no-reply:
this behavior is *very inefficiently* implemented in WorkForm. as a first step
to supporting this for ChangeSet-style forms, extract the behavior to a helper.

then, add a valkyrie implementation using ` Hyrax.custom_queries.find_collections_for`. we keep the calls to the object's implementation, if present. otherwise, we use the (still inefficient) loop and query approach.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -57,6 +57,7 @@ module Hyrax
 
       # pcdm relationships
       property :admin_set_id
+      property :member_of_collection_ids, default: []
 
       class << self
         ##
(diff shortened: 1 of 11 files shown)


app/forms/hyrax/forms/resource_form.rb
```diff
@@ -57,6 +57,7 @@ class Permission < Hyrax::ChangeSet
 
       # pcdm relationships
       property :admin_set_id
+      property :member_of_collection_ids
```
no-reply, line 60:
```suggestion
```

PR #4314, opened 2020-05-16
count collection items with the SolrService directly

no-reply:
remove a hard dependency on ActiveFedora by querying solr directly instead of
using the `Relation#count` interface.

there are other methods here using similar queries, but working out the
relationship with `Blacklight::AccessControls::Enforcement` is going to take
some more work. for one thing, we have local `Enforcement` additions that are
being bypassed (inappropriately?) by the existing queries.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/collection_presenter.rb
+++ b/app/presenters/hyrax/collection_presenter.rb
@@ -76,7 +76,7 @@ module Hyrax
     end
 
     def total_items
-      ActiveFedora::Base.where("member_of_collection_ids_ssim:#{id}").count
+      Hyrax::SolrService.new.count("member_of_collection_ids_ssim:#{id}")
     end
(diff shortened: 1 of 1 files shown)


cjcolvar:
Right, it looks like those other ones are using `accessible_by` which I think are only applying the basic gated discovery and not the full search builder context.

no-reply:
> Right, it looks like those other ones are using accessible_by which I think are only applying the basic gated discovery and not the full search builder context.

it sounds like you're implying that's appropriate? if so, i'm not following why. could you help me with whatever i'm overlooking?

cjcolvar:
Sorry, no, I wasn't trying to say that I thought it appropriate.  I'm happy to help as possible.

PR #4316, opened 2020-05-19
add a very simple manifest benchmark spec

no-reply:
Testing out some benchmarks for manifest generation.

Connected to: #4315


no-reply:
continuing to hold this. the rework of manifest generation i'm doing is currently cutting the bar for these two benchmarks from 15/25ms to 8/10ms, and i think with better gains for works with more members/file_sets.

i'm continuing to work out details related to #4315, but am hopeful about the outcome.

PR #4317, opened 2020-05-19
wings: round trip Valkyrie-native models, even when unregistered

no-reply:
this is the beginning of an effort to better generalize model conversion
handling in `Wings` to handle native models without an explicit,
application-side `ActiveFedora` equivalent.

the bar we're seeking to clear here is: get a resource/work class generated with
the Hyrax valkyrie generator to round trip in Wings without any specialized
AF-related handling. this doesn't yet try to assess the quality of metadata
round tripping, but that work should follow.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/active_fedora_classifier.rb
+++ b/lib/wings/active_fedora_classifier.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Wings
+  class ActiveFedoraClassifier < ActiveFedora::ModelClassifier
+    private
+
+      def classify(model_value)
+        if (match = model_value.match(/Wings\((.*)\)/))
+          valkyrie_class = match[1].constantize
+          Wings::ActiveFedoraConverter::DefaultWork(valkyrie_class)
+        else
+          super
+        end
+      end
+  end
+end
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -65,11 +65,15 @@ module Wings
     end
 
     def active_fedora_class
-      klass = resource.internal_resource.constantize
+      klass = begin
+                resource.internal_resource.constantize
+              rescue NameError
+                Wings::ActiveFedoraClassifier.new(resource.internal_resource).best_model
+              end
 
       return klass if klass <= ActiveFedora::Base
 
-      ModelRegistry.lookup(klass) || DefaultWork
+      ModelRegistry.lookup(klass) || self.class.DefaultWork(klass)
     end
 
     ##
@@ -84,6 +88,16 @@ module Wings
       resource.alternate_ids.first.to_s
     end
 
+    def self.DefaultWork(resource_class)
+      Class.new(DefaultWork) do
+        self.valkyrie_class = resource_class
+
+        # nested attributes in AF don't inherit! this needs to be here until
+        # we can drop it completely.
+        accepts_nested_attributes_for :nested_resource
+      end
+    end
+
     # A dummy work class for valkyrie resources that don't have corresponding
     # hyrax ActiveFedora::Base models.
     #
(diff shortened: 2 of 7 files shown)


lib/wings/active_fedora_converter.rb
```diff
@@ -84,6 +84,12 @@ def id
       resource.alternate_ids.first.to_s
     end
 
+    def self.DefaultWork(resource_class)
```
jeremyf, line 87: This method looks a little suspicious. And I'm trying to figure out why you wouldn't want a private instance method instead of using a class method.

jeremyf: Now I'm seeing.

lib/wings/active_fedora_classifier.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Wings
+  class ActiveFedoraClassifier < ActiveFedora::ModelClassifier
+    private
+
+      def classify(model_value)
+        match = model_value.match(/Wings\((.*)\)/)
+        match ? Wings::ActiveFedoraConverter::DefaultWork(match[1].constantize) : super
```
jeremyf, line 9: I think this would help connect the commit for me:

```ruby
if match
  klass = match[1].constantize
  Wings::ActiveFedoraConverter::DefaultWork(klass)
else
  super
end
```

A part of me wants `DefaultWork(for: klass)` but that is a possible idiomatic shift. Some of which it is still somewhat atypical to see method names that are not lower case (even though Ruby has `Array()`)

no-reply: Hyrax has been following the Dry.rb style of `ModuleName()` and `ClassName()` for module and class builder methods; see, e.g. `Hyrax::ChangeSet()` and `Hyrax::Schema()`, so that's the approach followed here.

i like your rephrase above though.

lib/wings/orm_converter.rb
```diff
@@ -29,7 +29,9 @@ def self.relationship_keys_for(reflections:)
     #
     # @return [Class]
     def self.base_for(klass:)
-      ModelRegistry.reverse_lookup(klass) || Hyrax::Resource
+      ModelRegistry.reverse_lookup(klass) ||
+        klass.try(:valkyrie_class) ||
```
jeremyf, line 33: Hmm…this isn't a blocker, but I'd almost assume that if the klass responds to :valkyrie_class, we'd go with that.

I haven't used a lot of `try` recently, but I thought `try` was for when I had a possible nil object. So if `klass` is nil, we'll have a false condition and move on to `Hyrax::Resource`. And if `klass` is not nil we're sending the `:valkyrie_class` message regardless of whether `klass` responds to it.

lib/wings/orm_converter.rb
```diff
@@ -53,7 +59,7 @@ def self.to_valkyrie_resource_class(klass:)
         include Wings::Works::FileSetValkyrieBehavior if klass.included_modules.include?(Hyrax::FileSetBehavior)
 
         # Based on Valkyrie implementation, we call Class.to_s to define the internal resource.
-        @internal_resource = klass.to_s
+        @internal_resource = (klass.try(:valkyrie_class) || klass).to_s
```
jeremyf, line 62: Can we even get here if we have line 49 (`return klass.valkyrie_class if klass.respond_to?(:valkyrie_class)`)

lib/wings/active_fedora_classifier.rb
jeremyf: Yeah. I can certainly get behind method names such as these. Just trying to adjust 15 years of Ruby as informed by Rails.

lib/wings/orm_converter.rb
no-reply: for the basic case of `obj.try(:method_name)` `#try` is functionally `respond_to?(:method_name) && public_send(:method_name)`

so we're returning `.valkyrie_class` if `klass` implements it.

i'm not too picky about whether we respect `ModelRegistry` first, or not. probably we could collapse this behavior to just `klass.try(:valkyrie_class) || Hyrax::Resource`, and do something like this in the Wings setup:

```ruby
class ActiveFedora::Base
  self.valkyrie_class
    ModelRegistry.reverse_lookup(self)
  end
end
```

lib/wings/orm_converter.rb
no-reply: probably not. i think this got retained from some earlier refactoring. it's probably not necessary and i'll try backing it out.

lib/wings/orm_converter.rb
jeremyf: I'm okay with the logic being here. And I learned something new about `try` (which once upon a time behaved a bit differently and less friendly)

PR #4321, opened 2020-05-20
extend `MembershipHelper` and `ResourceForm` for `member_ids`

no-reply:
add a `MembershipHelper#work_members_json` with support for legacy forms as well
as valkyrie resources. extend `ResourceForm` to add support for `member_ids`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -57,6 +57,7 @@ module Hyrax
 
       # pcdm relationships
       property :admin_set_id
+      property :member_ids, default: []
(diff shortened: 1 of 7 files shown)


jeremyf:
I like this direction. It starts breaking apart large objects, factoring that behavior into a helper. Yes…we could pursue an object to wrap the extract behavior, but for now this disentangles the issues.

I'd like others to weigh in.

We discussed this in [Samvera Tech call 2020-05-20](https://wiki.lyrasis.org/display/samvera/Samvera+Tech+Call+2020-05-20)

PR #4322, opened 2020-05-20
links from Profile page show users collections and works

### LINKED ISSUE #3478: "Works created" link on profile page does not return any results
### Descriptive summary

Hyrax v3.0.0-beta1
In the "profile" page of the Dashboard, the number of created works is correctly displayed. But by clicking on it, you obtain a blank list of result.

### Expected behavior

The number of items in the search page should match the number displayed in the profile page.

1. Create a work
2. Go to you profile page
profile_bug
3. Click on the "Created works" link
profile_bug1

The only explanation I found is that the solr request contains a local qf parameter.
With this parameter isolated the Solr response is empty :
```
http://localhost:8983/solr/hydra-development/select?q={!qf=depositor_ssim}dornbusch@ehess.fr
```
With a more basic syntax it works : 
```
http://localhost:8983/solr/hydra-development/select?q=depositor_ssim:dornbusch@ehess.fr
```

### Related work
Once this problem is solved, there will be a second one.
If you change the user key field in hydra.rb : 
```
config.user_key_field = Devise.authentication_keys.first # change this
```
it is not taken into account by the partial : 
https://github.com/samvera/hyrax/blob/b5c6605917c01c51b4504f7e27999146fc584713/app/views/hyrax/users/_vitals.html.erb#L12
The filtering uses `user.to_s`, which is equivalent to email.

blancoj:
This is a resubmission with some changes/improvements for the stale and closed PR:
https://github.com/samvera/hyrax/pull/4024

Fixes #3478

With this change, if a user goes to the Profile page from the dashboard, and then tries to see a listing of works and collections created by the user, the listing should be presented.

The code originally used the params search_field and a value for q to search for the depositor. For example, search_field=depositor&q=user@abc.com and then it appended a facet to indicate if looking for a work or a collection. I could not get this to work. The search_field with q value did not really seem to do anything. So...I tried using facets for depositor value, and this seems to work just fine. 

Since there is a change to the catalog controller generator, you will need to `bundle exec rake engine_cart:generate`, if you do not do this, it will still work, but the label for Depositor will not be right. 
The user should have a few works and collections.
Go to the dashboard and select Profile from Your activity dropdown.
Click on Works created and then Collections created and make sure that the lists are accurate in both cases.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/hyrax_helper_behavior.rb
+++ b/app/helpers/hyrax/hyrax_helper_behavior.rb
@@ -103,7 +103,7 @@ module Hyrax
     # @see Blacklight::SearchState#initialize
     def link_to_field(name, value, label = nil, facet_hash = {})
       label ||= value
-      params = { search_field: name, q: "\"#{value}\"" }
+      params = { search_field: name, q: "\"#{value}\"" } if name != ''
       state = search_state_with_facets(params, facet_hash)
       link_to(label, main_app.search_catalog_path(state))
     end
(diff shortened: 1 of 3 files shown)


app/helpers/hyrax/hyrax_helper_behavior.rb
```diff
@@ -305,11 +305,16 @@ def search_action_for_dashboard
 
       # @param [ActionController::Parameters] params first argument for Blacklight::SearchState.new
       # @param [Hash] facet
-      # @note Ignores all but the first facet.  Probably a bug.
       def search_state_with_facets(params, facet = {})
-        state = Blacklight::SearchState.new(params, CatalogController.blacklight_config)
+        state = Blacklight::SearchState.new([], CatalogController.blacklight_config)
```
no-reply, line 309: looks like dropping `params` here is causing the test failures.

no-reply:
is this ready for re-review, or still WIP?

blancoj:
@no-reply This can be reviewed now

PR #4323, opened 2020-05-20
support `#find_child_work` in the new `ResourceForm`

no-reply:
wire up a view partial test to check this behavior on the new form, then add
support for it.

i'm uncertain whether this should really be necessary, but it seems best to
support the current views as a starting place, and maybe tweak things from there.

@samvera/hyrax-code-reviewers


no-reply:
ehhhh. we should rebase this on #4321, rather than dealing with the merge conflict that would result from considering it separately.

no-reply:
rebased

jeremyf:
Approved, feel free to merge. If you're still noodling on this, then go ahead and update.

PR #4325, opened 2020-05-20
wings: support optimistic locking for Valkyrie resources

no-reply:
adds an optimistic locking implementation to `Wings` using Fedora's
`Last-Modified` property and etags.

the existing optimistic locking in the Actor stack counts on etags, so including
them in the lock tokens has immediate value. we'll need a strategy for removing
them later.

for now, this is a fairly naive implementation for early consideration.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/valkyrie/persister.rb
+++ b/lib/wings/valkyrie/persister.rb
@@ -18,6 +18,9 @@ module Wings
       def save(resource:)
         return save_file(file_metadata: resource) if resource.is_a? Hyrax::FileMetadata
         af_object = resource_factory.from_resource(resource: resource)
+
+        check_lock_tokens(af_object: af_object, resource: resource)
+
         af_object.save!
         resource_factory.to_resource(object: af_object)
       rescue ActiveFedora::RecordInvalid => err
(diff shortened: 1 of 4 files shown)


lib/wings/valkyrie/persister.rb
```diff
@@ -68,6 +71,42 @@ def initialize(msg = nil, obj:)
...
         end
       end
+
+      private
+
+        ##
+        # @return [void]
+        # @raise [::Valkyrie::Persistence::StaleObjectError]
+        def check_lock_tokens(af_object:, resource:)
+          return if !resource.optimistic_locking_enabled? || af_object.new_record?
```
jeremyf, line 81: I find myself preferring to read single step guards. Check my "math" on this, but are they equivalent?

Current:

```ruby
return if !resource.optimistic_locking_enabled? || af_object.new_record?
return if
            etag_lock_token_valid?(af_object: af_object, resource: resource) &&
            last_modified_lock_token_valid?(af_object: af_object, resource: resource)
```

Preferred:

```ruby
return unless resource.optimistic_locking_enabled?
return if af_object.new_record?
return unless etag_lock_token_valid?(af_object: af_object, resource: resource)
return unless last_modified_lock_token_valid?(af_object: af_object, resource: resource)
```

I find my preferred (assuming its correct) to read more quickly. I don't need to mentally juggle `&&` or `! ||`

no-reply: seems reasonable to me. i probably went through 5 refactorings of this to get it where it is. no harm in a few more :)

no-reply: re: your math, i think we still need to evaluate the two lock tokens together; we can't short-circuit until both pass.

the alternative is:
```ruby
raise [...] unless etag_lock_token_valid?(af_object: af_object, resource: resource)
raise [...] unless last_modified_lock_token_valid?(af_object: af_object, resource: resource)
```

but i rejected this due to the duplicated `raise` code.

PR #4326, opened 2020-05-21
WIP: try optimistic locking system-wide

no-reply:
see what happens if we try turning on optimistic locking for all our resources.


no-reply:
closing this, as we won't likely pursue this until post-Wings.

PR #4329, opened 2020-05-21
Update Ruby and Rails versions for CI matrix

botimer:
This is a simple update to refresh the CI version matrix to the most current Ruby and Rails.

@samvera/hyrax-code-reviewers


no-reply:
this is now out of date.

PR #4332, opened 2020-05-22
force upgrade of `reform` and `reform-rails`

no-reply:
this is a proposed fix for failing builds caused by
https://github.com/trailblazer/reform/issues/508
and release of `reform` v2.3.1.

@samvera/hyrax-code-reviewers


seuros:
Let me know if you have no regression so we can release 0.2.0

no-reply:
@seuros we're all green, here.

no-reply:
rebased this and bumped to the upcoming version numbers. this should go green once a `reform-rails` release happens.

seuros:
@no-reply  we released the versions.

PR #4335, opened 2020-05-23
wings: generate metadata dynamically for fallback AF classes

no-reply:
simplify `ActiveFedoraConverter` behavior by stripping out hard coded metadata and Hyrax-specific behaviors in favor of dynamically generated metadata.

there's still a bit more work to do here, to add robust support for Valkyrie-native nested models.

a small stylistic refactor is undertaken in dfdb7c6, because looking at the file was offending my sensibilities.

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -55,7 +55,6 @@ module Wings
       active_fedora_class.new(normal_attributes).tap do |af_object|
         af_object.id = id unless id.empty?
         add_access_control_attributes(af_object)
-        apply_depositor_to(af_object)
         convert_members(af_object)
         convert_member_of_collections(af_object)
         convert_files(af_object)
@@ -194,10 +193,6 @@ module Wings
           property = active_fedora_class.properties[attr.to_s]
           hash[attr] = property&.multiple? ? Array.wrap(value) : value
         end
-      end
-
-      def apply_depositor_to(af_object)
-        af_object.apply_depositor_metadata(attributes[:depositor]) unless attributes[:depositor].blank?
       end
 
       # Add attributes from resource which aren't AF properties into af_object
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -92,24 +92,28 @@ module Wings
       Class.new(DefaultWork) do
         self.valkyrie_class = resource_class
 
+        # extract AF properties from the Valkyrie schema;
+        # skip reserved attributes, proctected properties, and those already defined
+        resource_class.schema.each do |schema_key|
+          next if resource_class.reserved_attributes.include?(schema_key.name)
+          next if protected_property_name?(schema_key.name)
+          next if properties.keys.include?(schema_key.name.to_s)
+
+          property schema_key.name, predicate: RDF::URI("http://hyrax.samvera.org/ns/wings##{schema_key.name}")
+        end
+
         # nested attributes in AF don't inherit! this needs to be here until
         # we can drop it completely.
         accepts_nested_attributes_for :nested_resource
       end
     end
 
-    # A dummy work class for valkyrie resources that don't have corresponding
-    # hyrax ActiveFedora::Base models.
-    #
-    # A possible improvement would be to dynamically generate properties based
-    # on what's found in the resource.
+    ##
+    # A base model class for valkyrie resources that don't have corresponding
+    # ActiveFedora::Base models.
     class DefaultWork < ActiveFedora::Base
       include Hyrax::WorkBehavior
-      property :ordered_authors, predicate: ::RDF::Vocab::DC.creator
-      property :ordered_nested, predicate: ::RDF::URI("http://example.com/ordered_nested")
       property :nested_resource, predicate: ::RDF::URI("http://example.com/nested_resource"), class_name: "Wings::ActiveFedoraConverter::NestedResource"
-      include ::Hyrax::BasicMetadata
-      accepts_nested_attributes_for :nested_resource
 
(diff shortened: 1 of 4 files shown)


lib/wings/active_fedora_converter.rb
```diff
@@ -190,17 +188,10 @@ def assign_association_target(af_object, pcdm_file)
       def normal_attributes
         attributes.each_with_object({}) do |(attr, value), hash|
           property = active_fedora_class.properties[attr.to_s]
-          # This handles some cases where the attributes do not directly map to an RDF property value
-          hash[attr] = value
-          next if property.nil?
-          hash[attr] = Array.wrap(value) if property.multiple?
+          hash[attr] = property&.multiple? ? Array.wrap(value) : value
```
no-reply, line 191: i'm hoping this refactor in-particular is clarifying. previously i was wondering which "some cases" needed this special handling; and why, if this logic is narrowly targeted, it applies to all attributes. after looking at it closely, what we actually do is cast `value` to `Array` if `property.multiple?` is `true`, otherwise set the `value` as-is.

the special case is where `property` is `nil`. for me, this seems like a perfect use of `&.`. we can just treat `nil` properties as non-multiple, and not interrupt the flow of logic for a special check.

lib/wings/active_fedora_converter.rb
```diff
@@ -43,12 +43,10 @@ def initialize(resource:)
...
     # Accesses and parses the attributes from the resource through ConverterValueMapper
-    # @return [Hash]
+    #
+    # @return [Hash] attributes with values mapped for building an ActiveFedora model
     def attributes
-      @attribs ||= begin
-        wrapper = self.class.attributes_class.new(resource.attributes)
-        wrapper.result
-      end
+      @attributes ||= self.class.attributes_class.mapped_attributes(attributes: resource.attributes)
```
jeremyf, line 49: An improvement on "nosy neighbor" behavior. Less chained method calls.

lib/wings/active_fedora_converter.rb
```diff
@@ -82,34 +79,34 @@ def active_fedora_class
...
     end
 
     def self.DefaultWork(resource_class)
       Class.new(DefaultWork) do
         self.valkyrie_class = resource_class
 
-        # nested attributes in AF don't inherit! this needs to be here until
-        # we can drop it completely.
+        resource_class.schema.each do |schema_key|
+          next if resource_class.reserved_attributes.include?(schema_key.name) ||
```
jeremyf, line 93: Idomatically, I like `next if <single boolean>` test. This is not a blocker.

lib/wings/active_fedora_converter.rb
```diff
@@ -82,34 +79,34 @@ def active_fedora_class
...
       resource.alternate_ids.first.to_s
     end
 
     def self.DefaultWork(resource_class)
       Class.new(DefaultWork) do
         self.valkyrie_class = resource_class
 
-        # nested attributes in AF don't inherit! this needs to be here until
-        # we can drop it completely.
+        resource_class.schema.each do |schema_key|
```
jeremyf, line 92: A code comment for this block to state simple intention. This doesn't need the guard clause explanation, just tell me quickly what's happening.

lib/wings/active_fedora_converter.rb
```diff
@@ -129,14 +126,13 @@ def self.to_rdf_representation
       def to_global_id
         GlobalID.create(valkyrie_class.new(id: id))
       end
-
-      # self.indexer = <%= class_name %>Indexer
```
jeremyf, line 133: Huzza removing commented out code.

lib/wings/active_fedora_converter.rb
```diff
@@ -82,34 +79,34 @@ def active_fedora_class
...
       Class.new(DefaultWork) do
         self.valkyrie_class = resource_class
 
-        # nested attributes in AF don't inherit! this needs to be here until
-        # we can drop it completely.
+        resource_class.schema.each do |schema_key|
+          next if resource_class.reserved_attributes.include?(schema_key.name) ||
+                  protected_property_name?(schema_key.name) ||
+                  properties.keys.include?(schema_key.name.to_s)
+          property schema_key.name, predicate: RDF::URI("http://hyrax.samvera.org/ns/wings##{schema_key.name}")
```
jeremyf, line 96: @cjcolvar you had a question in Samvera tech, an you write that up here?

jeremyf:
Let's give @cjcolvar a chance to weigh.

lib/wings/active_fedora_converter.rb
no-reply: added. can you check if it looks okay?

lib/wings/active_fedora_converter.rb
no-reply: the question was, briefly: can we define the predicate URIs through valkyrie configuration (and ultimately through Hyrax's YAML driven metadata configuration?

---

yes, i think we should explore this.

can we consider rolling forward with this and taking that work on as immediate follow-up? are there SemVer concerns about that approach?

cjcolvar: I think this could be done as a follow up if we allow customizing predicates falling back to these default predicates if none customized.

I think we could follow an approach similar to that taken in the fedora adapter in valkyrie.
The metadata adapter there can be initialized with a schema (https://github.com/samvera/valkyrie/blob/master/lib/valkyrie/persistence/fedora/metadata_adapter.rb#L9) which responds to `#predicate_for` and `#property_for`(https://github.com/samvera/valkyrie/blob/master/lib/valkyrie/persistence/fedora/permissive_schema.rb#L74-L97) and used in the ModelConverter and OrmConverter.

PR #4337, opened 2020-05-26
support optimistic locking for `ResourceForm`

no-reply:
introduces a field and prepopulator for the (maybe confusingly named?) `version`
virtual attribute. `Hyrax::WorkForm` implements this by calling `model.etag`,
but we don't have that luxury here (since non-fedora backed valkyrie may not
have an etag at all). the logic added here is specific Wings adapter, pulling
out the etag for compatibility.

it's probably okay to keep this wings-specific indefinitely. other adapters
should support optimistic locking internally (indeed, so does Wings) so we can
phase out application-layer lock validation in favor of leaning on
Valkyrie/backend datastore features.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/views/hyrax/base/_form.html.erb_spec.rb
+++ b/spec/views/hyrax/base/_form.html.erb_spec.rb
@@ -7,6 +7,7 @@ RSpec.describe 'hyrax/base/_form.html.erb', type: :view do
     # mock the admin set options presenter to avoid hitting Solr
     allow(Hyrax::AdminSetOptionsPresenter).to receive(:new).and_return(options_presenter)
     allow(controller).to receive(:current_user).and_return(stub_model(User))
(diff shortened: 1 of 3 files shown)


spec/views/hyrax/base/_form.html.erb_spec.rb
```diff
@@ -39,12 +40,8 @@
     context 'with an existing object' do
       let(:work) { FactoryBot.valkyrie_create(:monograph) }
 
-      xit 'renders a form' do
-        # pending handling for #version/optimistic locking, and maybe other
-        # issues. a good way to make progress on ChangeSet forms is to enable
-        # this test, analyze the stacktrace, and try to patch.
+      it 'renders a form' do
```
jeremyf, line 43: Always great to see a `xit` become a run spec!

PR #4342, opened 2020-05-28
Metadata schema generation and use within Hyrax application

cjcolvar:
Generate schema with work resource based upon attributes passed to the generator and use it in the generated form, indexer, and model.  Also modify schema loader to be able to find the generated schema in the main app instead of the Hyrax engine.

~Unfortunately turning on the usage of the schema in the generated work classes (form, indexer, and model) by default breaks right now because the loaded yaml's attributes is nil and the public methods in SimpleSchemaLoader don't handle that case.~


cjcolvar:
Based upon conversation with @no-reply, I'm going to split this PR into two: allow simple schema loader to search the main app for schemas and generator changes to generate schema attributes in metadata yaml configuration then use them in the form/indexer/model.

cjcolvar:
Closed in favor of #4346 and #4347.

PR #4344, opened 2020-05-28
slight optimization to IIIF Manifest generation 

no-reply:
avoid some Fedora and solr calls when generating IIIF manifests. these presenter
checks call Fedora multiple times (`#exists?` and `can?(:read, id)` each make
separate round trips), per image.

instead, assume any resource we `#can?(:read)` exists, and use the solr document
to check (Blacklight-style) permissions.

on local benchmarks, this cuts manifest generation time by about 20% for moderately complex works. i'm still working on making those benchmarks robust enough to better understand how that scales, but in the meanwhile, here's this optimization. 

@samvera/hyrax-code-reviewers


no-reply:
this is somewhat) related to #4315

PR #4347, opened 2020-05-29
Generate metadata yaml with fields passed as arguments to generator

cjcolvar:
Generate metadata yaml with fields passed as arguments to generator then use this metadata configuration in the generated form, indexer, and model by default.  If no fields are passed to the generator then the generated metadata configuration is an empty hash of attributes and doesn't have any affect but doesn't break anything.

Should there be any tests for this other than using it in the generated internal test app?

Second PR split out from #4342.


elrayle:
I really like where this is going.  Ping me when the tests are passing and I'll do a formal review.

PR #4348, opened 2020-05-29
auto-generate tests for attributes in work resource generator

elrayle:
Confirms that the generated work resource responds to all generated metadata.

Related to PR #4347 

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/work_resource/work_resource_generator.rb
+++ b/lib/generators/hyrax/work_resource/work_resource_generator.rb
@@ -37,8 +37,15 @@ class Hyrax::WorkResourceGenerator < Rails::Generators::NamedBase
   end
 
   def create_model_spec
-    template('work_spec.rb.erb', File.join('spec/models/', class_path, "#{file_name}_spec.rb")) if
-      rspec_installed?
+    return unless rspec_installed?
+    filepath = File.join('spec/models/', class_path, "#{file_name}_spec.rb")
+    template('work_spec.rb.erb', filepath)
+    return unless attributes.present?
+    inject_into_file filepath, after: /it_behaves_like 'a Hyrax::Work'\n/ do
+      "\n  context 'includes schema defined metadata' do\n"\
+      "#{attributes.collect { |arg| "    it { is_expected.to respond_to(:#{arg.name}) }\n" }.join}" \
+      "  end\n"
+    end
   end
 
   def create_form


lib/generators/hyrax/work_resource/work_resource_generator.rb
```diff
@@ -37,8 +37,15 @@ def create_model
...
   def create_model_spec
-    template('work_spec.rb.erb', File.join('spec/models/', class_path, "#{file_name}_spec.rb")) if
-      rspec_installed?
+    return unless rspec_installed?
+    filepath = File.join('spec/models/', class_path, "#{file_name}_spec.rb")
+    template('work_spec.rb.erb', filepath)
+    return unless attributes.present?
+    inject_into_file filepath, after: /it_behaves_like 'a Hyrax::Work'\n/ do
+      "\n  context 'includes schema defined metadata' do\n"\
+      "#{attributes.collect { |arg| "    it { is_expected.to respond_to(:#{arg.name}) }\n" }.join}" \
```
no-reply, line 46: i don't think this needs to be wrapped in string interpolation
```suggestion
      attributes.collect { |arg| "    it { is_expected.to respond_to(:#{arg.name}) }\n" }.join
```
has the same result?

no-reply:
looks good to me. i made a simplifying suggestion.

lib/generators/hyrax/work_resource/work_resource_generator.rb
cjcolvar: Do you think it is possible to put this all inside the template and avoid the injection?

cjcolvar:
I'm excited by a generator that can actually build a fully working work type!

lib/generators/hyrax/work_resource/work_resource_generator.rb
no-reply: the injection is conditional on `attributes.present?`, but maybe it should not be?

i.e. what if we put the outline, including the `context` block in the template, plus a `# it { is_expected.to respond_to(:my_attribute})` line as guidance for app developers. then let the injection take place only for the custom attrs.

no-reply: in either case. i'm approving this. i think we should merge as-is and refine from there.

elrayle: @cjcolvar How would you put that in the template?  The attributes are defined inline with the generator.  You could put an example in the template, but I’m not sure that really buys us anything.  I don’t know another way to add to a template other than using inject_into_file.

cjcolvar: I'm not sure but I'd assume that the arguments would be available for use within the template so you could generate it just like you are now but in ERB style.

PR #4350, opened 2020-05-29
cast ResourceForm#member_ids and #member_of_collection_ids to Array

no-reply:
add types for these two form fields, enabling coercion. this makes the form
better at handling user/browser input.

see: http://trailblazer.to/gems/reform/options.html#coercion

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -69,6 +69,8 @@ module Hyrax
 
       property :visibility # visibility has an accessor on the model
 
+      property :date_modified, readable: false
+      property :date_uploaded, readable: false
       property :agreement_accepted, virtual: true, default: false, prepopulator: ->(_opts) { self.agreement_accepted = !model.new_record }
 
       collection :permissions, virtual: true, default: [], form: Permission, prepopulator: ->(_opts) { self.permissions = Hyrax::AccessControl.for(resource: model).permissions }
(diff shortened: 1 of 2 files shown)


app/forms/hyrax/forms/resource_form.rb
```diff
@@ -84,8 +86,8 @@ class Permission < Hyrax::ChangeSet
 
       # pcdm relationships
       property :admin_set_id
-      property :member_ids, default: []
-      property :member_of_collection_ids, default: []
+      property :member_ids, default: [], type: Valkyrie::Types::Array
+      property :member_of_collection_ids, default: [],  type: Valkyrie::Types::Array
```
no-reply, line 90:
```suggestion
      property :member_of_collection_ids, default: [], type: Valkyrie::Types::Array
```

app/forms/hyrax/forms/resource_form.rb
```diff
@@ -69,6 +69,8 @@ class Permission < Hyrax::ChangeSet
 
       property :visibility # visibility has an accessor on the model
 
+      property :date_modified, readable: false
+      property :date_uploaded, readable: false
```
cjcolvar, line 73: @no-reply Can you explain why these were added?  I didn't see any mention of them in this PR.

no-reply: eghhh... they're mostly unrelated, but were needed to make some related work work out.

okay to merge with them in place?

PR #4351, opened 2020-05-29
wings: refactor model lookup behavior for better fallback

no-reply:
since we introduced proper handling for valkyrie models without AF equivalents,
we've been casting Valkyrie models to `DefaultWork` directly in the
`ActiveFedoraConverter`. this is a bit fragile, since anything that looks them
up (like the `Wings::WorkSearchBuilder`), is going to miss in the registry,
leaking `nil` about where folks might see it! instead, let the registry handle
resolution to the appropriate class even in case of misses.

this points to a further refactor, extracting (and probably renaming)
`DefaultWork` from `Wings::ActiveFedoraConverter`, since it's no longer (and
never really was) neatly encapsulated there.

this fixes an intermittent failure in the test suite.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -70,7 +70,7 @@ module Wings
 
       return klass if klass <= ActiveFedora::Base
 
-      ModelRegistry.lookup(klass) || self.class.DefaultWork(klass)
+      ModelRegistry.lookup(klass)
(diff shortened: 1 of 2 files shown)


no-reply:
ughhh... this breaks a bunch of stuff.

going to need to figure it out next week i guess

Issue #4318, opened 2020-05-19
Address "alternate_title" confusion


jeremyf:
Based on an initial question:

> Question: Why is the alternative title a hidden input field in the works edit form? (see 
https://samvera.slack.com/archives/C0X9STKSP/p1589839711011900)

We have a commit that introduces "alt_title". It appears that the purpose of the "alt_title" is to help us treat "title" as a single field, but not lose any excess entries.

Looking at [c2bc31326d](https://github.com/samvera/hyrax/commit/c2bc31326d), we see handling the excess. Later, in [854d66c0de](https://github.com/samvera/hyrax/commit/854d66c0de) we see that introduced `alt_title` properties is aligned closer to `alternate_title`. However, the `alternate_title` in the form-field does not appear related to `alternative_title` the metadata field.

This needs detangling. Further confounding the issue, when we look at the tags that contain the original @c2bc31326d commit we have:

```
v2.5.2
v3.0.0-rc1
v3.0.0.pre.beta3
v3.0.0.pre.rc1
```
Further digging, highlights that while there is a tag for v2.5.2, there is no minted version in rubygems.org for v2.5.2

What this means is that there are no production releases that include the confusion of the `alt_title`

svogtunh:
A couple of other related issues that are still open: #4286, #2354.  The issue referenced above related to commit c2bc31326d is #3299 and there is no issue related to commit 854d66c0de.

jeremyf:
From @jlhardes on [Slack](https://samvera.slack.com/archives/C0X9STKSP/p1590001145016000?thread_ts=1589839711.011900&cid=C0X9STKSP)
> That makes sense to use dct:alternative for these “extra” titles. There is a single primary title and any other titles are considered alternative titles. It’s not a subtitle or anything really additional. The fact that the field is a hidden field might be more to do with trying to transition away from having multiple title fields. As its own field using a separate property it doesn’t really need to be hidden from view - the hiddenness might be more connected to making the change backwards compatible?

no-reply:
i'm not sure that the addition of `dct:alternative` frees us from the issue of a multivalued `dct:title` field in Core Metadata.

[...]

PR #4354, opened 2020-06-03
[WIP] move alternative title to core metadata to avoid nil error

elrayle:
### Description

Fix bug where `@attributes["alternative_title"] << value` throws an exception because `@attributes["alternative_title"]` is nil.  This happens when a model does not include basic metadata causing alternative_title to not be defined.

### Related work:

* Issue #4318 - reconsidering impact of PR #3554
* PR #3554 - separates multiple titles into title and alternative titles

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/basic_metadata.rb
+++ b/app/models/concerns/hyrax/basic_metadata.rb
@@ -6,8 +6,6 @@ module Hyrax
     extend ActiveSupport::Concern
 
     included do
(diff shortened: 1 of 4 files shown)


jeremyf:
Thank you for putting this step forward and bringing this for conversation and clarification. Based on Samvera Tech discussion, we view #4318 as a blocker, and will close this out.

The Reforms style should simplify this behavior (pre-populators and transformations are natively baked in).

PR #4355, opened 2020-06-03
update USAGE for hyrax:work_resource generator

elrayle:
This updates the USAGE doc for hyrax:work_resource generator to reflect recent changes.

Changes:
* add attributes to the example
* update files to the actual set of files that are generated

Question:
* The view is the only file put in the hyrax namespace. Is this done intentionally?  Or should it be changed outside the hyrax namespace.  Moving it outside would be more consistent with the naming followed for other files.

Recent changes to the generator to add attributes:
* PR #4347 
* PR #4348 

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/work_resource/USAGE
+++ b/lib/generators/hyrax/work_resource/USAGE
@@ -2,12 +2,15 @@ Description:
   This generator creates the necessary files for valkyrie-native work types.
 
 Example:
-  rails generate hyrax:work_resource Monograph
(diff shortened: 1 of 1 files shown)


cjcolvar:
>The view is the only file put in the hyrax namespace. Is this done intentionally? Or should it be changed outside the hyrax namespace. Moving it outside would be more consistent with the naming followed for other files.

I think this is a good question @elrayle!  I'm guessing that Hyrax expects and only looks in app/views/hyrax/<work_type> right now, but I think it would be good to add app/views/<work_type> to its search path for consistency as you noted.

PR #4356, opened 2020-06-05
Manifest presenter rewrite

no-reply:
this complete rewrite of the manifest generation process reduces the footprint
of database (Fedora/Solr) calls involved in the manifest generation process,
fixes a bug in sequence/canvas generation for manifests for nested works, and
generally separates the issue of manifest generation from the
`WorkShowPresenter` (existing code is left in place, to be deprecated and
removed as follow-up).

[`iiif_manifest`](https://www.rubydoc.info/gems/iiif_manifest/1.0.1) expects a
"presenter", with `#manifest_url`, `#manifest_metadata` `#description`,
`#sequence_rendering`, and `#work_presenters`/`#file_set_presenters`. since we
know the presenter provided will be used for the duration of a single request,
we can make heavy use of instance variables to cache children and avoid making
multiple requests for any particular data.

the rewrite also does some work to help clarify the two peices of information
the presenter needs in the request context: the `Ability` and the hostname to
use for URIs within the manifest. we provide naive defaults for both of these,
and implement the behavior more or less without regard for their
context---i.e. the whole presenter infrastructure works without them. when
calling from a Controller, we pass them in explicitly to augment the behavior
for the context. this is important because we may (do) want to reuse these
presenters in other contexts (a Job for cache pre-warming) where this data isn't
available.

the presenters themselves use `Draper::Decorator`'s `delegate_all`, so they're
basically glorified wrappers for a model. they don't care whether the model is a
SolrDocument, or powered by a primary database. a small amount of work is done
to align Valkyrie, ActiveFedora, and SolrDocument-style models, but mostly
things were already interoperable.

a refactor in `DisplaysImage` rearranges some instance-level caching there, for
easier reuse in the new presenter infrastructure.

the short story is: between this and a previous optimization in
`DisplaysImage` (#4344), manifest generation in a makeshift controlled
benchmarking environment sped up 10-fold for works of modest size. it seems
roughly linear between 10 and 750 child FileSets. i'm not confident enough in
that environment (especially its behavior WRT caching) to call this "data" or to
provide anything more solid than this statement, but there's good reason to
believe this will be quite a bit faster/more scalable than the status-quo.

* completely rewrite manifest generation behavior

* check behavior of IIIF viewers on complex works with multiple files and nested works
  * look especially to ensure files the current user shouldn't be able to see aren't included.

@samvera/hyrax-code-reviewers


jeremyf:
👀

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -142,6 +142,13 @@ def iiif_manifest_builder
         self.class.iiif_manifest_builder
       end
 
+      def iiif_manifest_presenter
+        IiifManifestPresenter.new(curation_concern_from_search_results).tap do |p|
```
jeremyf, line 146: Putting a pin in the `tap` call. I'm curious if we can instead expect a block for initialization (or even pass those parameters on initialization). Regardless, this is not a blocker. And I'll find out more in later files.

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -128,7 +128,7 @@ def inspect_work
     def manifest
       headers['Access-Control-Allow-Origin'] = '*'
 
-      json = iiif_manifest_builder.manifest_for(presenter: presenter)
+      json = iiif_manifest_builder.manifest_for(presenter: iiif_manifest_presenter)
```
jeremyf, line 131: I like using a specific presenter instead of the big heavy work presenter. Only "present" the smallest necessary object to complete the request.

app/jobs/iiif_manifest_cache_prewarm_job.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class IiifManifestCachePrewarmJob < Hyrax::ApplicationJob
+  ##
+  # @param work [ActiveFedora::Base]
+  def perform(work)
+    manifest_builder.manifest_for(presenter: Hyrax::IiifManifestPresenter.new(work))
```
jeremyf, line 7: This is splitting hairs, a personal preference of mine is to instantiate the object separate from the passing of that object as a parameter. This is a bit of the command/query separation principle.

I've also found this helps in inserting a debugger if necessary.

```ruby
def perform(work)
  presenter = Hyrax::IiifManifestPresenter.new(work)
  manifest_builder.manifest_for(presenter: presenter)
end
```

app/models/concerns/hyrax/solr_document/metadata.rb
```diff
@@ -51,6 +51,7 @@ def self.coerce(input)
         attribute :read_groups, Solr::Array, ::Ability.read_group_field
         attribute :collection_ids, Solr::Array, 'collection_ids_tesim'
         attribute :admin_set, Solr::Array, "admin_set_tesim"
+        attribute :member_ids, Solr::Array, "member_ids_ssim"
```
jeremyf, line 54: 👍

jeremyf: I'm assuming that this is to help reduce calls to Fedora.

app/models/concerns/hyrax/solr_document_behavior.rb
```diff
@@ -61,17 +61,29 @@ def to_model
...
+
+    ##
+    # @return [Boolean]
     def admin_set?
       hydra_model == ::AdminSet
     end
 
     # Method to return the ActiveFedora model
     def hydra_model
       "::#{first('has_model_ssim')}".constantize
```
jeremyf, line 84: Is this a place for `safe_constantize`?

app/presenters/hyrax/displays_image.rb
```diff
@@ -45,13 +44,18 @@ def image_format(channels)
...
-        end
-        result
+      def latest_file_id
+        @latest_file_id ||=
+          begin
+            result = original_file_id
+
+            if result.blank?
+              Rails.logger.warn "original_file_id for #{id} not found, falling back to Fedora."
+              result = Hyrax::VersioningService.versioned_file_id ::FileSet.find(id).original_file
```
jeremyf, line 54: I know this is a port, but line 54 concerns me. On a quick read with line breaks I read it as:

`Hyrax::VersioningService.versioned_file_id::FileSet.find(id).original_file`

However, in looking at a wider buffer, I see it's actually semantically this: `Hyrax::VersioningService.versioned_file_id(::FileSet.find(id).original_file)`

I again would prefer a local variable `original_file = ::FileSet.find(id)` then pass that to the poorly named `#versioned_file_id` method. I'd go with `find_versioned_file_id_for` as the method name.

However, that is certainly not a refactor that need be done in this PR. Just an observation while reviewing.

app/presenters/hyrax/iiif_manifest_presenter.rb
```diff
@@ -0,0 +1,222 @@
...
+    #   @return [Ability]
+    # @!attribute [w] hostname
+    #   @return [String]
+    attr_writer :ability, :hostname
+
+    class << self
+      ##
+      # @param [Hyrax::Resource, SolrDocument]
+      def for(model)
+        klass = model.file_set? ? DisplayImagePresenter : IiifManifestPresenter
```
jeremyf, line 32: Can I get a bit of an explanation on this method? It's a bit surprising to see the `for` pick this class or another class. It implies that the other class (e.g., DisplayImagePresenter) should have a similar logic switch. Which then leads me to look at the namespace organization of these.

Again, not a blocker, but a surprise.

app/services/hyrax/manifest_builder_service.rb
```diff
@@ -68,7 +66,7 @@ def sanitized_manifest(presenter:)
         hash = JSON.parse(manifest.to_json)
 
         hash['label'] = sanitize_value(hash['label']) if hash.key?('label')
-        hash['description'] = hash['description']&.collect { |elem| sanitize_value(elem) } if hash.key?('description')
+        hash['description'] = Array(hash['description'])&.collect { |elem| sanitize_value(elem) } if hash.key?('description')
```
jeremyf, line 69: Good catch!

spec/presenters/hyrax/iiif_manifest_presenter_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+      expect(builder_service.manifest_for(presenter: presenter))
+        .to include('@context' => 'http://iiif.io/api/presentation/2/context.json')
+    end
+
+    context 'with file set and work members' do
+      let(:work) { create(:work_with_image_files) }
+
+      it 'generates a manifest with nested content' do
+        expect(builder_service.manifest_for(presenter: presenter)['sequences'].first['canvases'].count)
+          .to eq 2
```
jeremyf, line 21: How do we get to this magic #? Is there a property of presenter or work that makes this 2?

spec/presenters/hyrax/iiif_manifest_presenter_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+            .not_to have_key('sequences')
+        end
+
+        it 'includes items with read permissions' do
+          readable = FactoryBot.create(:file_set, :image, user: user)
+          work.members << readable
+          work.save
+
+          expect(builder_service.manifest_for(presenter: presenter)['sequences'].first['canvases'].count)
+            .to eq 1
```
jeremyf, line 41: How do we get to this number 1? Is there a property we can get from some of the involved objects?

spec/presenters/hyrax/iiif_manifest_presenter_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+      let(:work) { create(:work_with_image_files) }
+
+      it 'gives presenters for the file sets' do
+        expect(presenter.file_set_presenters)
+          .to contain_exactly(*work.member_ids.map { |id| have_attributes(id: id) })
+      end
+
+      it 'gives DisplayImagePresenters' do
+        expect(presenter.file_set_presenters.map(&:display_image))
+          .to contain_exactly(an_instance_of(IIIFManifest::DisplayImage),
```
jeremyf, line 89: This appears to be related to the above "magic" number 2. Wondering how to convey that in the spec.

app/controllers/concerns/hyrax/works_controller_behavior.rb
no-reply: i'd like to do more of this, in general.

for me, the power of the presenter pattern is that i can decorate an object for a specific purpose. making presenters monolithic seems like an anti-pattern for me

app/models/concerns/hyrax/solr_document/metadata.rb
no-reply: it's a bit weird that this wasn't already exposed as an attribute. it *was* already indexed, so this is just an ergonomics/model-layer-compatibility issue.

app/presenters/hyrax/displays_image.rb
no-reply: agreed. this `displays_image` refactor is something i tried to do in as fixed a way as possible.

i'm not deeply confident in the code in this module (and it's not unit tested) so i've tried to limit the scope of changes.

app/presenters/hyrax/iiif_manifest_presenter.rb
no-reply: the way i'm thinking of this is `DisplayImagePresenter` is the `IiifManifestPresenter` equivalent for FileSets. this factory method's scope is to give me a presenter for use by `iiif_manifest`, appropriate for the given object.

>  It implies that the other class (e.g., DisplayImagePresenter) should have a similar logic switch. 

can you say more about this?

spec/presenters/hyrax/iiif_manifest_presenter_spec.rb
no-reply: i think what's really needed in this case, and in https://github.com/samvera/hyrax/pull/4356#discussion_r436010991, is a set of matchers for `iiif_manifest`

```ruby
expect(my_manifest).to have_sequence(file_set_1, file_set_2, file_set_3, ...)
```

i'm reticent to spend time belaboring these specs, short of that; it felt to me like more precision here was going to be more fragile (`['sequences'].first['canvases'].count` is already a bit miserable).

one refinement could be `   .to eq work.file_set_ids.count`, but even that doesn't feel particularly helpful; it should really be `work.file_sets.select(:&image?).count` or something that starts to look a whole bunch like the implementation...

app/controllers/concerns/hyrax/works_controller_behavior.rb
no-reply: a Builder might be a thing, but i'm not keen on messing with the initializer, since it's provided by `Draper`.

app/presenters/hyrax/iiif_manifest_presenter.rb
jeremyf: I was assuming that both classes were in the same namespace level and not that the `DisplayImagePresenter` was an inner class. Once I saw it as an inner-class, I'm less concerned about this method. I do believe it highlights that the `IiifManifestPresenter` may in fact make sense to have a namespacing module (e.g. `IiifManifestPresentation`) and move the `IiifManifestPresenter` within that module. Then the `IiifManifestPresentation.for` method, which determines which inner class to initialize.

spec/presenters/hyrax/iiif_manifest_presenter_spec.rb
jeremyf: I think this is a case for a code comment explaining the number two is the above logic. I'd certainly not want to run that query to determine that value.

app/presenters/hyrax/iiif_manifest_presenter.rb
no-reply: ah yeah. `DisplayImagePresenter` is kept as an inner class with the idea that it provides encapsulation. generally, only `iiif_manifest` callers should be unwrapping enough to see the "Display Image" parts of this, so from Hyrax's perspective `IiifManifestPresenter` does all the work.

spec/presenters/hyrax/iiif_manifest_presenter_spec.rb
no-reply: can add comments for both of these.

i'm serious about IIIF Manifest matchers, but thinking it's out-of-scope for now.

PR #4358, opened 2020-06-08
Better model resolution for wings

no-reply:
since we introduced proper handling for valkyrie models without AF equivalents,
we've been casting Valkyrie models to `DefaultWork` directly in the
`ActiveFedoraConverter`. this is a bit fragile, since anything that looks them
up (like the `Wings::WorkSearchBuilder`), is going to miss in the registry,
leaking `nil` about where folks might see it! instead, let the registry handle
resolution to the appropriate class even in case of misses.

this points to a further refactor, extracting (and probably renaming)
`DefaultWork` from `Wings::ActiveFedoraConverter`, since it's no longer (and
never really was) neatly encapsulated there.

this fixes an intermittent failure in the test suite.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -70,7 +70,7 @@ module Wings
 
       return klass if klass <= ActiveFedora::Base
 
-      ModelRegistry.lookup(klass) || self.class.DefaultWork(klass)
+      ModelRegistry.lookup(klass)
     end
 
     ##
--- a/lib/wings/model_registry.rb
+++ b/lib/wings/model_registry.rb
@@ -41,7 +41,10 @@ module Wings
     end
 
     def lookup(valkyrie)
-      @map[valkyrie]
+      valkyrie = valkyrie._canonical_valkyrie_model if
+        valkyrie.respond_to?(:_canonical_valkyrie_model)
+
+      @map.fetch(valkyrie) { ActiveFedoraConverter::DefaultWork(valkyrie) }
     end
 
     def reverse_lookup(active_fedora)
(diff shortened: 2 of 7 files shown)


lib/wings/active_fedora_converter.rb
```diff
@@ -70,7 +70,7 @@ def active_fedora_class
 
       return klass if klass <= ActiveFedora::Base
 
-      ModelRegistry.lookup(klass) || self.class.DefaultWork(klass)
+      ModelRegistry.lookup(klass)
```
jeremyf, line 73: Yes!

lib/wings/model_registry.rb
```diff
@@ -41,7 +41,10 @@ def register(valkyrie, active_fedora)
     end
 
     def lookup(valkyrie)
-      @map[valkyrie]
+      valkyrie = valkyrie._canonical_valkyrie_model if
```
jeremyf, line 44: Also yes (providing an escape hatch in the duck-typing)

lib/wings/orm_converter.rb
```diff
@@ -62,7 +63,13 @@ class << self
           attr_reader :internal_resource
 
           def name
-            ancestors[1..-1].find { |parent| parent < ::Valkyrie::Resource }&.name
+            _canonical_valkyrie_model&.name
+          end
+
+          ##
+          # @api private
+          def _canonical_valkyrie_model
```
jeremyf, line 71: Hmm. Does `@api private` reconcile with the above usage in lookup? I'm thinking it might, but want to double check.

no-reply: i think yes. this tag is meant to communicate to applications that this isn't stable/public API. in this case, i think it's `Wings`-use only.

PR #4359, opened 2020-06-09
support Valkyrie work creation with Hyrax transactions

no-reply:
add Valkyrie work creation handling for `WorksControllerBehavior`. this uses `Hyrax::Transactions` to process the work.

to accomplish this, we first bump `Hyrax::Transactions` into the engine for use with Valkyrie resources.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
+
+##
+# A mixin for Hyrax support methods. This is meant to be included in
+# `ApplicationController`.
+#
+# @note private methods within this module are normally still "public API",
+#   since they are meant to be called by inheriting controllers.
 module Hyrax::Controller
   extend ActiveSupport::Concern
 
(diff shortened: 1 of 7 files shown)


app/controllers/concerns/hyrax/controller.rb
```diff
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
+
+##
+# A mixin for Hyrax support methods. This is meant to be included in
+# `ApplicationController`.
+#
+# @note private methods within this module are normally still "public API",
```
jeremyf, line 7: Excellent comment to clarify!

spec/controllers/concerns/hyrax/works_controller_behavior_spec.rb
```diff
@@ -1,18 +1,29 @@
...
+        devise_for :users
+        resources :solr_documents, only: [:show], path: '/catalog', controller: 'catalog'
+      end
+    end
+  end
 
-  after(:context) do
-    config = Hyrax.config
-    types  = config.registered_curation_concern_types - ["Hyrax::Test::SimpleWork"]
+  before do
```
jeremyf, line 20: Is this working now? I'd imagine extracting a general support module might be helpful.

no-reply: i think it's working! nothing else is yet using similar behavior, but i'm :+1: on extracting before we end up duplicating this elsewhere.

PR #4372, opened 2020-06-16
Removing Capybara Maleficent

jeremyf:
Jeremy here, I introduced Maleficent at Notre Dame as we observed odd
behavior in an older version of Capybara.  I know for certain that
Maleficent is not keeping up with Capybara releases.

At the time of its creation, there were assertions that Capybara was
waiting for DOM resolution.  However, in our observations this was not
always true.

However, this is a place where we could very easily be experiencing
increased fragility (and a slow down in specs).

I'd like to see if things work without this past hack.

NOTE: Today, I have been walking back gems from recently released
versions.  It's slow going, and my observation is that the feature specs
are the fragile ones.  The inner workings of Capybara may have changed
and thus either: 1) removing the need of Maleficent or 2) breaking
Maleficent.  So let's remove Maleficent and see how things fare.

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

*
*
*

*
*
*

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -84,7 +84,6 @@ SUMMARY
   spec.add_dependency 'valkyrie', '>= 2.1.1', "< 3.0"
 
   spec.add_development_dependency "capybara", '~> 3.29'
-  spec.add_development_dependency 'capybara-maleficent', '~> 0.3.0'
   spec.add_development_dependency 'chromedriver-helper', '~> 2.1'
   spec.add_development_dependency 'database_cleaner', '~> 1.3'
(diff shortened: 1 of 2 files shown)


hyrax.gemspec
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
no-reply, line 1: is this needed?

jeremyf: Not that I know of. I'm running the long tests locally right now, so I'll amend when those complete/fail.

PR #4330, opened 2020-05-21
Add support for Rails 5.2 and AF 12

no-reply:
Changes proposed in this pull request:
* Does just what it says in the title!
* Test rails 5.1 and 5.2 on latest ruby patch versions.

This would be Hyrax 2.8.0.

@samvera/hyrax-code-reviewers

PR #4376, opened 2020-06-18
Test with latest rubies adding 2.7.1

cjcolvar:
Update all ruby versions to the latest and add 2.7.1 in 2.x-stable.  This is just ruby versions to separate them from the efforts to update rails in #4330 and #4329.

@samvera/hyrax-code-reviewers


no-reply:
Looks like we need to reset the expected builds after this. but it passes so i'm approving and merging.

PR #4379, opened 2020-06-19
Allow both dry-validation 0.x and 1.x

cjcolvar:
This backports the changes done in master in 5d8819ff265f5ce57c4743cac01efa0950bf8f89 but tries to avoid changing the interface through the use of a class method #call and a Dry::Validation::Result wrapping class using SimpleDelegator that defines #messages.

This allows downstream applications to upgrade to dry-validation 1.x which is necessary if the app uses the `config` gem and wants to use ruby 2.7.  This unblocks Hyku from upgrading to Hyrax 2.7.2 and ruby 2.7 (see https://github.com/samvera/hyku/issues/1630).

I ran the tests for WorkflowSchema and WorkflowImporter locally with dry-validation 0.13.3 (and dry-struct 0.6.0) as well as dry-validation 1.5.1 (and dry-struct 1.3.0).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/workflow/workflow_schema.rb
+++ b/app/services/hyrax/workflow/workflow_schema.rb
@@ -14,47 +14,107 @@ module Hyrax
     # @see Sipity::Notification
     # @see Sipity::Method
     # @see ./lib/generators/hyrax/templates/workflow.json.erb
-    WorkflowSchema = Dry::Validation.Schema do
-      configure do
-        def self.messages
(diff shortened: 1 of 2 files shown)


no-reply:
this looks a bit like it might have broken due to an AF release, so i'm looking into that.

cjcolvar:
Waiting now for #4380 to be merged and this to be rebased on top of it.

PR #4382, opened 2020-06-24
Add description, citation_keywords, and citation_publisher meta tags for Google Scholar

cjcolvar:
This is a port of https://github.com/samvera/hyku/pull/1528 with the following slight modifications:

- Uses the same content as og:description for description
- Adds protection to avoid rendering blank citation_publication_date and citation_pdf_url
- Adds tests

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/shared/_citations.html.erb
+++ b/app/views/shared/_citations.html.erb
@@ -15,12 +15,29 @@
 <% end %>
 
 <% content_for(:gscholar_meta) do %>
+  <meta name="description" content="<%= @presenter.description.first.truncate(200) rescue @presenter.title.first %>" />
   <meta name="citation_title" content="<%= @presenter.title.first %>" />
+
   <% @presenter.creator.each do |creator| %>
   <meta name="citation_author" content="<%= creator %>" />
   <% end %>
+
+  <% unless @presenter.date_created.blank? %>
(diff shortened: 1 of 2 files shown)


no-reply:
!!! yesss

app/views/shared/_citations.html.erb
```diff
@@ -15,12 +15,29 @@
...
+  <% unless @presenter.download_url.blank? %>
   <meta name="citation_pdf_url" content="<%= @presenter.download_url %>" />
+  <% end %>
+
+  <% unless @presenter.keyword.blank? %>
+  <meta name="citation_keywords" content="<%= @presenter.keyword.join('; ') %>" />
+  <% end %>
+
+  <% unless @presenter.publisher.blank? %>
+  <meta name="citation_publisher" content="<%= @presenter.publisher.join %>" />
```
jeremyf, line 38: I'm not aware of the syntax, but do you want a delimiter here?

cjcolvar: Good question.  I don't really know the syntax either and there aren't easily found good docs on these tags that I could find.  Probably safest to follow the same as for keywords (`'; '`) even though this might be an edge case?

jeremyf: I agree that it's an edge case and think `; ` probably makes the most sense

PR #4383, opened 2020-06-25
Release 2.8

no-reply:
prepare for 2.8.0 release.

@samvera/hyrax-code-reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,11 +138,11 @@ Starting up Redis will depend on your operating system, and may in fact already 
 
 ## Rails
 
-Hyrax requires Rails 5. We recommend the latest Rails 5.1 release.
+Hyrax 2.x supports Rails 5.1 and 5.2. We recommend the latest Rails 5.2 release.
 
 ```
 # If you don't already have Rails at your disposal...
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -138,11 +138,11 @@ Starting up Redis will depend on your operating system, and may in fact already
 
 ## Rails
 
-Hyrax requires Rails 5. We recommend the latest Rails 5.1 release.
+Hyrax 2.x supports Rails 5.1 and 5.2. We recommend the latest Rails 5.2 release.
 
 ```
 # If you don't already have Rails at your disposal...
-gem install rails -v 5.1.6
+gem install rails -v 5.2.4
```
cjcolvar, line 145: How about bumping this to 5.2.4.3 since there have been security patches?

no-reply: ah. it looks like we can do `gem install rails -v '~> 5.2.4'` (or `~>5.2`?) which seems better than both?

no-reply:
no go on `~>`

our documentation needs to give the specific version later on. short of reworking completely to work around that, switching back to hard codes on the latest security release.

PR #4385, opened 2020-06-25
Refactoring module to reduce redundant calls

jeremyf:
Because we were not memoizing the `current_work` method call, we could
be issuing two SOLR queries that are identical.  This small refactor now
moves the `::SolrDocument.find` call to a local variable which we pass
to the two methods.

I've also opted to break the guard clauses into separate steps.  I
believe this reads more clearly.

And I updated the documentation and added some points about Sipity and
SOLR integration.

@samvera/hyrax-code-reviewers


jeremyf:
It is also possible that this will provide a speed increase to the test suite; One less reification of a SOLR document loaded via search.

PR #4390, opened 2020-06-25
Reworking method call to improve clarity

jeremyf:
Prior to this commit, it wasn't immediately evident that the
`curation_concern_from_search_result` was meant to run for a single
item (e.g., the given `params[:id]`).  It becomes evident when you look
at the behavior of `search_result_document`, which takes the first item
OR fallsback to a `SolrDocument.find(params[:id])`

In addition, this removes a mutation on the params which COULD have some
odd behavior.  Again, given that we're working with a singular object, I
don't think :page is a likely parameter.

I also chose to write a comment to help capture what I've discovered in
some additional work.  It's an odd thing and I want to make sure other
see it.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -138,7 +138,7 @@ module Hyrax
     end
 
     def iiif_manifest_presenter
-      IiifManifestPresenter.new(curation_concern_from_search_results).tap do |p|
+      IiifManifestPresenter.new(search_result_document(id: params[:id])).tap do |p|
         p.hostname = request.hostname
         p.ability = current_ability
       end
@@ -171,7 +171,7 @@ module Hyrax
     end
 
     def presenter
-      @presenter ||= show_presenter.new(curation_concern_from_search_results, current_ability, request)
+      @presenter ||= show_presenter.new(search_result_document(id: params[:id]), current_ability, request)
     end
 
     def parent_presenter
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -199,13 +199,15 @@ def contextual_path(presenter, parent_presenter)
       ::Hyrax::ContextualPath.new(presenter, parent_presenter).show
     end
 
-    def curation_concern_from_search_results
```
no-reply, line 202: can we deprecate this, rather than dropping it?

i've been considering `private` methods on these mixins public API, since they are only "private" to the scope of controllers which users' applications own directly.

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -199,13 +199,46 @@ def contextual_path(presenter, parent_presenter)
       ::Hyrax::ContextualPath.new(presenter, parent_presenter).show
     end
 
+    # @deprecated
     def curation_concern_from_search_results
-      search_params = params
+      Deprecation.warn("We will remove '##{__method__}' in Hyrax 4.0.  " /
```
cjcolvar, line 204: I'm afraid that `We` might be interpreted as a royal `we` instead of a people's `we`. How about making it third-person?
```suggestion
      Deprecation.warn("'##{__method__}' will be removed in Hyrax 4.0.  " /
```

jeremyf: I can certainly change it. My goal is to remove passive voice; There are implications on non-accountability.  However, it is consistent with other deprecation warnings.

cjcolvar: Good point.  Maybe Samvera could come up with a short style guide for deprecation text?

no-reply: is this a tech call topic?

cjcolvar: :+1: Added.

PR #4394, opened 2020-06-30
support transactional update via `WorksControllerBehavior`

no-reply:
adds a barebones update work transaction and use it to support a very basic update controller action.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -90,7 +90,7 @@ module Hyrax
     end
 
     def update
-      if actor.update(actor_environment)
+      if update_work
         after_update_response
       else
         respond_to do |wants|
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -367,7 +379,13 @@ def additional_response_formats(format)
...
+      @saved_permissions =
+        case curation_concern
+        when ActiveFedora::Base
+          curation_concern.permissions.map(&:to_hash)
+        else
+          Hyrax::AccessControl.for(resource: curation_concern).permissions
+        end
     end
 
     def permissions_changed?
```
cjcolvar, line 391: Does this method need to be updated to work with Hyrax::Resource objects as well?

cjcolvar: Nevermind.  It looks like that will come later when dealing with works with file sets.

no-reply: :+1:

PR #4396, opened 2020-06-30
test transactional update

no-reply:
add some unit tests for transactional update in preparation for continuing to add to its behavior.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/steps/set_uploaded_date_unless_present.rb
+++ b/lib/hyrax/transactions/steps/set_uploaded_date_unless_present.rb
@@ -42,7 +42,7 @@ module Hyrax
           return Failure[:no_date_uploaded_attribute, obj] unless
             obj.respond_to?(:date_uploaded=)
 
-          obj.date_uploaded = date_uploaded(obj)
+          obj.date_uploaded = date_uploaded(obj) if obj.date_uploaded.blank?
 
           Success(obj)
         end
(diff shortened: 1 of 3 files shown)


lib/hyrax/transactions/steps/set_uploaded_date_unless_present.rb
```diff
@@ -42,7 +42,7 @@ def call(obj)
           return Failure[:no_date_uploaded_attribute, obj] unless
             obj.respond_to?(:date_uploaded=)
 
-          obj.date_uploaded = date_uploaded(obj)
+          obj.date_uploaded = date_uploaded(obj) if obj.date_uploaded.blank?
```
jeremyf, line 45: Seems like a good guard clause, and probably fixes a bug.

spec/hyrax/transactions/steps/set_uploaded_date_unless_present_spec.rb
```diff
@@ -50,6 +50,18 @@
           .to xmas
       end
 
+      context 'when an uploaded date exists' do
+        let(:xmas_past) { DateTime.parse('2009-12-25 11:30').iso8601 }
+
+        before { change_set.date_uploaded = xmas_past }
+
+        it 'sets the uploaded date to the modified date' do
```
jeremyf, line 58: I don't think this description matches the tested behavior.

Consider `it "does not overwrite the existing uploaded date"`

no-reply, line 58:
```suggestion
        it 'does not overwrite the existing uploaded date' do
```

PR #4397, opened 2020-06-30
Postgresql in circle

jeremyf:
## Adds optional Postgres setup for test database

cd4c970485f55d630a14221d99135203963f7eca

Our CircleCI orb gives us the ENV variables this setup uses.

## Adding postgresql to the engine cart build

ce1cd77509785ba3673c3b1a251f3f1f6af5ab68

Thanks @rotated8 for the previous leg of the journey.

For both local development and CircleCI, this commit switches the default
database from Sqlite to Postgresql.

Intrepid developer, please note when next you regenerate the
`.internal_test_app`, you will by default be generating it to use a
Postgresql database.  For now, I've included guidance on what to do in
the `./.engine_cart.yml` file.

Once you've regenerated your `.internal_test_app`, you'll need to export
a DATABASE_URL (or update `.internal_test_app/config/database.yml`) to
specify how you're connecting to Postgresql.

This builds on the guidance in the [Rails Guides][1] around using a
DATABASE_URL environment variable.

[1]:https://edgeguides.rubyonrails.org/configuring.html#configuring-a-database
Also, ensuring hyrax installs migrations in CirclCI

At a future point, we'll likely need to update the developer Wiki, but
for now, dear developer, you'll need to review the commit messages.

## Changing column type [Needs Upgrade Notes]

f33c2131c9b851da96d6781e036fd4988b674f9f

I do not believe this change column is likely used by adopters; It has
to do with adding permissions to an individual Sipity::Entity.

The change is necessary because the foreign key was created as a string,
but the associated primary key on the other table was an integer.

This is likely a bad copy paste job from ndlib/sipity


spec/features/create_child_work_spec.rb
```diff
@@ -28,6 +28,7 @@
       fill_in('Title', with: work_title)
       click_on('Save')
     end
+
```
no-reply, line 31: squash out whitespace-only change?

spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -5,10 +5,72 @@ class TestAppGenerator < Rails::Generators::Base
...
+    rake "db:drop:all"
+    rake "db:create:all"
+    rake "db:test:prepare"
+  end
+
   def install_engine
     generate 'hyrax:install', '-f'
   end
 
+
```
no-reply, line 73: is this extra whitespace?

jeremyf: This is likely the "And Jeremy rebased 20 commits into 3" effect.

spec/features/create_child_work_spec.rb
jeremyf: This is likely the "And Jeremy rebased 20 commits into 3" effect.

lib/generators/hyrax/templates/db/migrate/20200617002029_change_sipity_entity_specific_responsibility.rb.erb
```diff
@@ -0,0 +1,11 @@
+class ChangeSipityEntitySpecificResponsibility < ActiveRecord::Migration<%= migration_version %>
```
no-reply, line 1: i'd really like to know whether it's possible for existing applications (ones that are already using postgres successfully) to get by without running this migration. i.e. is this migration optional for existing apps? or is this a hard break?

lib/generators/hyrax/templates/db/migrate/20200617002029_change_sipity_entity_specific_responsibility.rb.erb
```diff
@@ -0,0 +1,11 @@
...
+  def change
+    if defined?(Sipity::EntitySpecificResponsibility)
```
no-reply, line 3: what's up with these conditionals? is there a more generic way of accomplishing this? 
`Sipity::EntitySpecificResponsibility.connect.adapter_name == "PostgreSQL"` seems particularly weird here.

will the up/down features of this `change` work, or is there a risk they'll break if the conditionals evaluate differently?

spec/test_app_templates/lib/generators/test_app_generator.rb
```diff
@@ -5,6 +5,67 @@ class TestAppGenerator < Rails::Generators::Base
   # so the following path gets us to /path/to/hyrax/spec/test_app_templates/
   source_root File.expand_path('../../../../spec/test_app_templates/', __FILE__)
 
+  def conditional_postgresql_for_ci_environment
+    # For now, postgresql is something we are looking at adding to CircleCI
```
no-reply, line 9: could all of this be handled by `DATABASE_URL` instead? https://edgeguides.rubyonrails.org/configuring.html#configuring-a-database

jeremyf:
So, even if these specs pass, there appears to be some "fails more than it passes" builds.

I'm going to look into the DATABASE_URL, as I think that might help someone run both a sqlite and a postgresql instance in their dummy app (based on ENV).

I also believe, at this point, that all Postgresql specific changes to the code-base have been merged via other PRs.

lib/generators/hyrax/templates/db/migrate/20200617002029_change_sipity_entity_specific_responsibility.rb.erb
jeremyf: I suspect that this is something that _should_ be run on those applications.  Without the change, I believe some of the Sipity queries may fail.  Given that this is "in production" it is also possible that people have remediated their applications already. It is certainly a "HEY - This is something to consider"

Triage steps will be:

1) Look at the `sipity_entity_specific_responsibilities` table and see if there are records.  If there are no records, run this without fear.
2) If there are records, you'll want to back up your data before you run it.

My suspicion is that no one has entries in this table, but I'm not certain. And yes, we could remove the table, but then we lose a key function of sipity (which is not fully integrated with Hyrax and is something we need to look at in regards to the general permission ecosystem).

lib/generators/hyrax/templates/db/migrate/20200617002029_change_sipity_entity_specific_responsibility.rb.erb
jeremyf: I removed one of the guard clauses, relying instead on the `ActiveRecord::Base.connection` and assuming that the class would use the same connection.

spec/test_app_templates/lib/generators/test_app_generator.rb
jeremyf: I think we could. I also believe we could then run postgresql OR sqlite locally for specs based on an ENV.  That would be dreamy.

jeremyf: I did some further digging, and those ENV's currently come from CircleCI. Also, based on my read of that configuring a database, you'd be able to use DATABASE_URL for local development.

Perhaps this is something to bump to @rotated8  and samvera's orbs?

no-reply: @jeremyf i think you can set environment variables for a job here, without requiring changes to the orb: https://github.com/samvera/hyrax/blob/master/.circleci/config.yml#L20-L22

i think adding a line like `DATABASE_URL: 'postgresql://hyrax_user:hyrax_password@postgres/hyrax?pool=5'` might do the trick?

jeremyf: @no-reply I'll give it a go on my local machine; I'd love to remove some of that conditional logic.

jeremyf:
I rebased the work based on #4417

lib/generators/hyrax/templates/db/migrate/20200617002029_change_sipity_entity_specific_responsibility.rb.erb
jeremyf: As I've been debugging, this table is most definitely used in `app/services/hyrax/workflow/workflow_factory.rb` with the `assign_specific_roles_to`.

spec/test_app_templates/lib/generators/test_app_generator.rb
jeremyf: @no-reply I'm struggling to unpack this one. In part, what is in place currently works for CI. What I'm proposing is that we merge this, and then I can tinker with refining the CI build process to try the DATABASE_URL.

The iteration on the build process is rather slow locally. I think having an already building Postgresql CI environment would be helpful as I work to fine tune that build process.

All of that said, I could be doing this work on a branch of this branch.

jeremyf: I have confirmed that if I provide a `DATABASE_URL=postgresql://localhost:5432/internal_development?pool=5 bundle exec rails console` on my machine that this works. I have not attempted a username/password.

jeremyf: I'm writing these as notes, as I explore: This is where I'm currently getting the ENVs (https://github.com/samvera-labs/samvera-circleci-orb/blob/master/src/executors/ruby_fcrepo_solr_redis_postgres.yml#L41-L43). I believe I would need to adjust our `.circleci/config.yml`

jeremyf: I switched my config to use a different adapter and database name.

I then ran `DATABASE_URL=postgresql://localhost:5432/internal_development?pool=5 bundle exec rails console` in the .internal_test_app. The environment used the DATABASE_URL.

Next consideration, is can I take the CircleCI environment variables and craft a new environment variable?

no-reply: possibly it's okay to just hard code the CI config in multiple ENV variables as a first step, then refactor the duplication out after?

rotated8: I think I'm hearing that it'd be nice if the orb exported a DATABASE_URL variable. So, samvera-labs/samvera-circleci-orb#42

jeremyf: @rotated8 thanks, I'll get to hard-coding, and see about remediating that particular file. I'm a big fan of how small this PR is becoming.

jeremyf: The last set of commits and rebasing addresses the DATABASE_URL, and opts for the hard-coding as talked about in https://wiki.lyrasis.org/display/samvera/Samvera+Tech+Call+2020-07-15

It's much cleaner and removes several switches (though pushes people to locally use postgresql when they regen their internal test app).

.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         type: string
         default: '5.2.2'
     executor:
-      name: 'samvera/ruby'
+      name: 'samvera/ruby_fcrepo_solr_redis_postgres'
```
cjcolvar, line 17: I think it should be safe to just have the ruby executor since you shouldn't need to have postgres et al running for bundle install.

.circleci/config.yml
```diff
@@ -112,6 +114,12 @@ jobs:
           ruby_version: << parameters.ruby_version >>
           bundler_version: << parameters.bundler_version >>
           project: hyrax
+      - run:
+          command: |
+            cd .internal_test_app
+            bundle install
```
cjcolvar, line 120: Do we need another bundle install here?  It runs at the engine level right before this.

.circleci/config.yml
jeremyf: Cool. I'll poke at the change.

.circleci/config.yml
jeremyf: I don't know; I'll remove and find out. Brace yourselves, rebases are coming!

spec/test_app_templates/lib/generators/test_app_generator.rb
no-reply: oh wow this looks great! thanks for your documentation.

i appreciate the personal nature of your code comments. :)

.engine_cart.yml
```diff
@@ -0,0 +1,40 @@
+# Greetings Hyrax developer, you may be surprised to see this file.
```
no-reply, line 1: great!

these docs are helpful. i may take a pass at a docker version later on.

.circleci/config.yml
```diff
@@ -112,6 +114,9 @@ jobs:
           ruby_version: << parameters.ruby_version >>
           bundler_version: << parameters.bundler_version >>
           project: hyrax
+      - run:
+          command: |
+            cd .internal_test_app
```
cjcolvar, line 119: It looks like you lost the migration installation and db:migrate in the rebase.

jeremyf:
The errors that I'm seeing indicate that I may be missing some necessary build step.

.circleci/config.yml
jeremyf: I want to remove this, but without the step, things fail.

jeremyf: See https://app.circleci.com/pipelines/github/samvera/hyrax/3828/workflows/4011ea3f-393e-484f-a0aa-fceeaa633057/jobs/25909/steps

cjcolvar: Thanks for trying that out.  Not in scope of this PR, but I wonder if this is another place that circleci caching could be utilized.

jeremyf: Yeah, I'd love to not run bundle install more than is necessary. This looks like it shouldn't be necessary, but I leave it up to the next person.

PR #4400, opened 2020-06-30
Refining behavior of UsersController#base_query

jeremyf:
This commit clarifies the behavior, the public @api nature of the
method, and adds another test case around hash based parameters.

As I was exploring Postgresql, I believe the `[nil]` as a query
parameter may have caused an issue.

This change should preserve the previous "public" implementation.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/users_controller.rb
+++ b/app/controllers/hyrax/users_controller.rb
@@ -25,19 +25,32 @@ module Hyrax
     # @param query [String] the query string
     def search(query)
       clause = query.blank? ? nil : "%" + query.downcase + "%"
-      base = ::User.where(*base_query)
+      base = ::User
+      base = base.where(*Array.wrap(base_query))
       base = base.where("#{Hydra.config.user_key_field} like lower(?) OR display_name like lower(?)", clause, clause) if clause.present?
       base.registered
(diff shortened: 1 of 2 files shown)


app/controllers/hyrax/users_controller.rb
```diff
@@ -25,19 +25,33 @@ def show
     # @param query [String] the query string
     def search(query)
       clause = query.blank? ? nil : "%" + query.downcase + "%"
-      base = ::User.where(*base_query)
+      base = ::User
+      base = base.where(*Array.wrap(base_query)) if base_query.present?
```
no-reply, line 29: this feels a bit defensive. do we need both `Array.wrap` and the `#present?` check?

jeremyf: The `Array.wrap` is because we need it to "splat" the return value (to keep the previous interface). The present is in a commit from months ago, I added it out of some necessity. Perhaps postgresql?

It definitely is a bit more defensive.

jeremyf: I believe with the `*Array.wrap` I can remove the base_query.

PR #4401, opened 2020-07-01
Restoring a check for permissions on JSON document

jeremyf:
I'm not certain why this started appearing now, but it has.  The problem
is seen in builds of @f90d703dcd8c65fa0e0d0be0e0fa65b378c76cc8.  When
you run specs against that SHA, we get the following errors:

```shell
rspec ./spec/controllers/hyrax/generic_works_controller_json_spec.rb:24
rspec ./spec/controllers/hyrax/generic_works_controller_json_spec.rb:32
````

The following commit (@696dee503e987ebed726b6e76e6925425f8c892f) removed
the call for `curation_conern`; This left the JSON requests
unauthorized.

Devise did release a new gem on 2020-06-10, so it could be related to
that.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -68,6 +68,7 @@ module Hyrax
 
     # Finds a solr document matching the id and sets @presenter
     # @raise CanCan::AccessDenied if the document is not found or the user doesn't have access to it.
+    # rubocop:disable Metrics/AbcSize
     def show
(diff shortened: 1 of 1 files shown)


no-reply:
do we think this needs a backport to 2.x?

jeremyf:
@no-reply I think this one warrants a backport

jeremyf:
I'll start working on it

jeremyf:
Good news! The previous change has not been released!!!! Yay! No backporting!

PR #4402, opened 2020-07-01
move valkyrie create strategy to shared specs to allow use in apps

elrayle:
## Description

Moves ValkyrieCreateStrategy from specs to lib/specs/shared_specs.  This allows implementing apps the ability to use the same strategy to test Valkyrie::Resources in their tests.

@samvera/hyrax-code-reviewers


elrayle:
In testing this in my local app, it took 10s to create a resource using valkyrie_create and 16s to create 3 more.  It was so slow, at first, I thought it wasn't working.

Testing in Hyrax, it took 4s to create the first resource, and 1s to create 3 more.

no-reply:
looks good. i haven't experienced slowdowns at that scale on `valkyrie_create`, but we're also using the metadata adapter for unit tests.

i'm curious why things are so much slower in your local app.

jeremyf:
I'm wondering if you have a nesting indexer disabled or enabled?

PR #4404, opened 2020-07-01
have sidebar menu behave consistently when switching options.  Fixes #1495

### LINKED ISSUE #1495: Dashboard sidebar should respect collapsed/not collapsed universally
### Descriptive summary

As a user, i can collapse the dashboard sidebar or leave it expanded. If i collapse it, i expect it to remain collapsed in future dashboard navigation, unless i select to expand the dashboard. 

### Expected behavior

when navigating in the dashboard, the collapsed/expanded status of the sidebar should remain as selected throughout use of the dashboard. 

### Actual behavior

when i collapse the sidebar then navigate to another dashboard page, the sidebar expands. I am forced to collapse the sidebar every time i go to a page. 

1. go to dashbaord
2. collapse sidebar
3. navigate to different page on dashboard
4. see the sidebar expanded

blancoj:
Fixes issue #1495

Presently when the user goes to the dashboard and opens up one or two of the accordion options ( User Activity, or Settings ), and then selects a link, like say Collections, the accordions close.  With this change the accordions stay in the position the user has them in as the user switches from one option to another.

I want to point out that in the code I added a javascript called `dontChangeAccordion` that gets called by the non accordion options.  This was needed because clicking on these non accordion options was causing the accordions to switch to the state in which they were in when the page was originally rendered before taking on the expected state.

* Go to the dashboard
* Open  the UserActivity Accordion to open it.
* Clink on the Collections option.
* Note that the accordion remains open
* Close the UserActivity Accordion.
* Go to the Works option
* Notice that the accordion remains closed.
* Do the same thing with the Settings accordion
* There are lots of tests combinations to try out. The main thing is that the accordion(s) remain in the state they were in when the user switches options.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax.js
+++ b/app/assets/javascripts/hyrax.js
@@ -104,6 +104,7 @@
 //= require hyrax/tabbed_form
 //= require hyrax/turbolinks_events
 //= require hyrax/i18n_helper
+//= require hyrax/collapse
 
 // this needs to be after batch_select so that the form ids get setup correctly
 //= require hyrax/batch_edit
--- a/app/assets/javascripts/hyrax/collapse.js
+++ b/app/assets/javascripts/hyrax/collapse.js
@@ -1,0 +1,91 @@
+Blacklight.onLoad(function() {
+    const isCollapsed = getCollapsedActivity();
+    if(isCollapsed){
+        localStorage.setItem('collapsedActivity', 'collapsed');
+    }else{
+        localStorage.setItem('collapsedActivity', 'un-collapsed');
+    }
+    getStatusActivity();
+
+    const isCollapsedSettings = getCollapsedSettings();
+    if(isCollapsedSettings){
+        localStorage.setItem('collapsedSettings', 'collapsed');
+    }else{
+        localStorage.setItem('collapsedSettings', 'un-collapsed');
+    }
+    getStatusSettings();
+});
+
+function getCollapsedActivity() {
+    const state = localStorage.getItem('collapsedActivity');
+    if(state === 'un-collapsed'){
+        return false;
+    }
+    return true;
+}
+
+function getCollapsedSettings() {
+    const state = localStorage.getItem('collapsedSettings');
+    if(state === 'un-collapsed'){
+        return false;
+    }
+    return true;
+}
+
+function getStatusActivity(){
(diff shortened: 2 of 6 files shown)


jeremyf:
@blancoj You may want to rebase (see #4414)

jeremyf:
Hmm…the rebase didn't help, wondering if #4417 may be a fix

jeremyf:
I believe this has a legit failure; All 4 specs fail at the same spot (which is not a regular happening for flappy tests). I'm reviewing now.

jeremyf:
Adding within scope for click_link
    
    On this page, I noticed the following HTML:
    
    ```html
    <td>
      <div class="media">
        <a class="media-left" href="/concern/generic_works/64223392-26a5-44fe-aaf3-e27ae4331646?locale=en">
          <img class="hidden-xs file_listing_thumbnail" alt="My Test Work thumbnail" src="/assets/work-ff055336041c3f7d310ad69109eda4a887b16ec501f35afc0a547c4adb97ee72.png">
        </a>
        <div class="media-body">
          <div class="media-heading">
            <a id="src_copy_link64223392-26a5-44fe-aaf3-e27ae4331646" class="document-title" href="/concern/generic_works/64223392-26a5-44fe-aaf3-e27ae4331646?locale=en">
              <span class="sr-only">Display all details of</span>My Test
              Work
            </a>
            <br>
          </div>
        </div>
      </div>
    </td>
    ```
    
    Notice, at `td div.media a.media-left img` we have an alt attribute of
    "My Test Work thumbnail". Then at `td div.media .media-heading a` we
    have an inner HTML text of "Display all details of My Test Work"
    
    On the page, these are the only two references to My Test Work.  By
    adding the within, this now limits where we look for the link text.
    
    For build https://github.com/samvera/hyrax/pull/4404, I saw all four
    Ruby versions fail in the same way.  When I pulled down the branch, I
    was able to reproduce the error.  With this fix, the broken spec started
    working.

no-reply:
@blancoj @jeremyf are both of you :+1: on this in its current state?

just trying to understand if this is still WIP or if it's now ready to go since its :heavy_check_mark:

blancoj:
@no-reply it's not WIP.  It needs to be reviewed. Thanks!

PR #4410, opened 2020-07-02
generate an basic indexer test for Valkyrie::Resource works

elrayle:
As part of the work resource generation process, this adds a spec test for the generated indexer.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/work_resource/templates/indexer_spec.rb.erb
+++ b/lib/generators/hyrax/work_resource/templates/indexer_spec.rb.erb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# Generated via
+#  `rails generate hyrax:work_resource <%= class_name %>`
+require 'rails_helper'
+require 'hyrax/specs/shared_specs/indexers'
+
+RSpec.describe <%= class_name %>Indexer do
+  let(:indexer_class) { described_class }
+
+  it_behaves_like 'a Hyrax::Resource indexer'
+end
--- a/lib/generators/hyrax/work_resource/USAGE
+++ b/lib/generators/hyrax/work_resource/USAGE
@@ -13,4 +13,5 @@ Example:
     config/initializers/hyrax.rb
     config/metadata/monograph.yaml
     spec/models/monograph_spec.rb
+    spec/indexers/monograph_indexer_spec.rb
     spec/views/monographs/_monograph.html.erb_spec.rb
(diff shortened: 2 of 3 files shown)


elrayle:
The test failure is the same for all builds.  The same test failure does not occur when run by itself locally.  The failure does not appear to be related to this PR.

Failing test:
```
rspec ./spec/features/collection_type_spec.rb:367 # collection_type delete collection type when collections exist of this type shows unable to delete dialog and forwards to All Collections with filter applied
```

Failure:
```
Failure/Error: expect(page).to have_content(deny_delete_modal_text, :all)
     ArgumentError:
       ${@type} is not a valid type for a text query
```

lib/generators/hyrax/work_resource/templates/indexer_spec.rb.erb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Generated via
+#  `rails generate hyrax:work_resource <%= class_name %>`
+require 'rails_helper'
+require 'hyrax/specs/shared_specs/indexers'
+
+RSpec.describe <%= class_name %>Indexer do
+  let(:indexer_class) do
```
no-reply, line 9: doesn't this test the wrong class?

the `describe` says this tests `[GeneratedModel]Indexer`, is there a reason we'd want to test a generic `Hyrax::ResourceIndexer` instead?

no-reply:
looks like the tests are fixed in #4409

lib/generators/hyrax/work_resource/templates/indexer_spec.rb.erb
elrayle: That was likely a copy/paste error.  I'll take a look.

elrayle: @no-reply Updated to correctly use the generated index class.

lib/generators/hyrax/work_resource/templates/indexer_spec.rb.erb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Generated via
+#  `rails generate hyrax:work_resource <%= class_name %>`
+require 'rails_helper'
+require 'hyrax/specs/shared_specs/indexers'
+
+RSpec.describe <%= class_name %>Indexer do
+  let(:indexer_class) { <%= class_name %>Indexer }
```
no-reply, line 9: would `described_class` be better?

```suggestion
  let(:indexer_class) { described_class }
```

elrayle: That makes sense.

elrayle: Done

no-reply: thanks!

no-reply:
thanks for this!

PR #4419, opened 2020-07-10
backport: delegate IIIF Manifest cache key version to presenter

no-reply:
it turns out `etag` isn't a very good cache key because the default
`SolrDocument` doesn't contain this data. using the system/Fedora
`#modified_date` is equivalent (has the same weaknesses) but is available on
both an actual work instance, and on its index proxy.

since this is changing (and should likely change again!), delegate it to the
presenter so there's a clear source going forward.

backports #4418 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/caching_iiif_manifest_builder.rb
+++ b/app/services/hyrax/caching_iiif_manifest_builder.rb
@@ -45,13 +45,9 @@ module Hyrax
       end
 
       ##
-      # @note `etag` is a better option than the solr document `_version_`; the
-      #   latter isn't always available, depending on how the presenter was
-      #   built!
-      #
       # @return [String]
       def version_for(presenter)
-        presenter.etag
+        presenter.version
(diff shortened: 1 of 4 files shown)


app/services/hyrax/caching_iiif_manifest_builder.rb
```diff
@@ -45,13 +45,9 @@ def manifest_cache_key(presenter:)
...
 
       ##
-      # @note `etag` is a better option than the solr document `_version_`; the
-      #   latter isn't always available, depending on how the presenter was
-      #   built!
-      #
       # @return [String]
       def version_for(presenter)
-        presenter.etag
+        presenter.version
```
jeremyf, line 50: I like this explicit method name usage; Instead of relying on an implicit understanding of `etag` (which is reasonable but just a bit less clear).

PR #4421, opened 2020-07-10
Demonstrating version changes when work changes

jeremyf:


spec/presenters/hyrax/iiif_manifest_presenter_spec.rb
```diff
@@ -217,6 +217,10 @@
       expect(presenter.version).to be_a String
     end
 
+    it 'changes when the work changes' do
+      expect { work.save }.to change { presenter.version }
```
no-reply, line 221: i originally wrote a test like this, but found it failed because the object saved within a second of when it was originally created. since i expect this method to continue to change, i decided not to belabor this issue.

it looks like maybe this version is failing for the same reason?

no-reply:
closing this since the base PR is merged.

i'm still keen to have some kind of test here, but i imagine we have plenty of opportunity to revisit.

PR #4422, opened 2020-07-10
backport: VersioningService.versioned_file_id

no-reply:
`Hyrax::DisplaysImage` relies on this `.versioned_file_id` method, but it was
missed in previous backports.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/versioning_service.rb
+++ b/app/services/hyrax/versioning_service.rb
@@ -22,5 +22,14 @@ module Hyrax
       return if version.nil?
       VersionCommitter.create(version_id: version.uri, committer_login: user_key)
     end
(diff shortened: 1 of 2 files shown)


no-reply:
same. i'm looking into it now, but wanted to push this as WIP while i'm at it.

i have someone on the other end testing these backports in a live app.

no-reply:
the reason this didn't get caught in testing is that it's only invoked in a narrow circumstance where `original_file_id` is not indexed (this is the case when it's deployed to an existing app, until all existing FileSet are re-indexed).

since this test is missing from the mainline, i'm thinking we should patch it there first.

PR #4429, opened 2020-07-11
Bumping capybara timeouts

jeremyf:
As someone that's been keeping an eye on the specs and trying to tidy
them up, I'd like to see if some of our assumptions about sleeps/waits
in javascript are not adequate for the CircleCI ecosystem.

This is a bit of a stab in the dark.

@samvera/hyrax-code-reviewers


no-reply:
if this turns out to be helpful, let's remember to backport to `2.x-stable`

PR #4430, opened 2020-07-11
add a `#find_collections_by_type` query

no-reply:
support query of collections by type. as noted, this is a default query
implementation without adapter optimization. we'll want to provide a better
version of this for Wings; and later, probably one for postgres.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/custom_queries/find_collections_by_type.rb
+++ b/app/services/hyrax/custom_queries/find_collections_by_type.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+module Hyrax
+  module CustomQueries
+    # @example
+    #   collection_type = Hyrax::CollectionType.find(1)
+    #
+    #   Hyrax.custom_queries.find_collections_by_type(global_id: collection_type.to_global_id)
+    #
+    # @see https://github.com/samvera/valkyrie/wiki/Queries#custom-queries
+    class FindCollectionsByType
+      def self.queries
+        [:find_collections_by_type]
+      end
+
+      def initialize(query_service:)
+        @query_service = query_service
+      end
+
+      attr_reader :query_service
+      delegate :resource_factory, to: :query_service
+
+      ##
+      # @note this is an unoptimized default implementation of this custom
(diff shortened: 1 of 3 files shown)


app/services/hyrax/custom_queries/find_collections_by_type.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+module Hyrax
+  module CustomQueries
+    # @example
+    #   collection_type = Hyrax::CollectionType.find(1)
+    #
+    #   Hyrax.custom_queries.find_collections_by_type(global_id: collection_type.to_global_id)
+    #
+    class FindCollectionsByType
+      def self.queries
```
jeremyf, line 10: This is unexplained, like I'm missing something.

no-reply: https://github.com/samvera/valkyrie/wiki/Queries#custom-queries

jeremyf: I was expecting to see documentation. (Maybe even a Valkyrie provided builder, but this works).

jeremyf: I think I'd like to see either documentation OR a spec tightening up the connection to Valkyrie.

no-reply: i added a `@see` to the above.

i think the tests are already tightly bound to Valkyrie, since this class will raise errors without a `query_service:` initializer argument.

spec/services/hyrax/custom_queries/find_collections_by_type_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+RSpec.describe Hyrax::CustomQueries::FindCollectionsByType, valkyrie_adapter: :test_adapter do
```
jeremyf, line 2: Are we going to need a clean repo here?

PR #4432, opened 2020-07-13
add optimized `#fid_collections_by_type` query for Wings

no-reply:
this is follow-up to #4430, adding an optimized query to Wings. this should be vastly faster than the default query for the Wings adapter.

@samvera/hyrax-code-reviewers


spec/wings/services/custom_queries/find_collections_by_type_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      let(:collections_with_target_type) do
+        FactoryBot.create_list(:collection_lw, 3, collection_type_gid: type_gid)
+      end
+
+      before do # force create all the models
+        non_collection_model
+        collection_with_default_type
+        collections_with_target_type
+      end
```
cjcolvar, line 31: I'm curious of your thoughts on lazy loading these in a before block vs. using `let!`.

no-reply: i think rubocop bars `let!` and this is my attempt at a slightly more intent-revealing approach for that.

cjcolvar: Interesting.  I tend to think that `let!` is clear (except that the little `!` can be easily overlooked) and I usually find the simple references in a before block odd-looking but your comment makes it better by making the intent explicit.

no-reply:
i fixed the failing test.

spec/wings/services/custom_queries/find_collections_by_type_spec.rb
no-reply: the issue with `let!`, i think, is that it leads folks to write test suites that eagerly create unneeded test objects. my understanding is that the style restriction seeks to make this deliberately more cumbersome/visible.

PR #4433, opened 2020-07-13
remove some stubs on object-under-test for CollectionType

no-reply:
we should never be stubbing the object under test. this cleans up some
violations of that for CollectionType.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/models/hyrax/collection_type_spec.rb
+++ b/spec/models/hyrax/collection_type_spec.rb
@@ -2,6 +2,14 @@
 RSpec.describe Hyrax::CollectionType, type: :model do
   let(:collection_type) { build(:collection_type) }
 
+  shared_context 'with a collection' do
+    let(:collection_type) { create(:collection_type) }
+
+    before do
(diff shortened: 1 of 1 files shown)


spec/models/hyrax/collection_type_spec.rb
```diff
@@ -2,6 +2,14 @@
 RSpec.describe Hyrax::CollectionType, type: :model do
   let(:collection_type) { build(:collection_type) }
 
+  shared_context 'with a collection' do
+    let(:collection_type) { create(:collection_type) }
+
+    before do
+      FactoryBot.valkyrie_create(:hyrax_collection, collection_type_gid: collection_type.gid)
```
cjcolvar, line 9: Is there a reason for using `valkyrie_create` instead of `create(:collection_lw, collection_type_gid: collection_type.gid.to_s)` which is used elsewhere in the test (like line 155)?

no-reply: `:collection_lw` will go away when ActiveFedora does. it's a todo item to replace all the direct Fedora calls with valkyrie equivalents, so whenever we write new tests it's better to use it instead.

PR #4436, opened 2020-07-13
refactor CollectionPresenter specs to use `subject`

no-reply:
just getting this organized to test without AF.

@samvera/hyrax-code-reviewers


no-reply:
sorry for the re-push. this is a case where we stubbed the object under test, but used `subject { presenter }` instead of `subject(:presenter)` to keep rubocop from noticing. :|

PR #4438, opened 2020-07-14
i18nize "items" on Collection show page

no-reply:
remove hard coded text in favor of an i18n key.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/collections/show.html.erb
+++ b/app/views/hyrax/collections/show.html.erb
@@ -35,7 +35,7 @@
           <div class="hyc-bugs">
             <div class="hyc-item-count">
               <b><%= @presenter.total_viewable_items %></b>
-              <%= pluralize(@presenter.total_viewable_items, t('item_count')) %></div>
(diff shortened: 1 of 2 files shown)


app/views/hyrax/collections/show.html.erb
```diff
@@ -35,7 +35,7 @@
           <div class="hyc-bugs">
             <div class="hyc-item-count">
               <b><%= @presenter.total_viewable_items %></b>
-              <%= 'Item'.pluralize(@presenter.total_viewable_items) %></div>
+              <%= pluralize(@presenter.total_viewable_items, t('item_count')) %></div>
```
cjcolvar, line 38: This is failing in CI tests.  Maybe this will fix it?
```suggestion
              <%= pluralize(@presenter.total_viewable_items, t('.item_count')) %></div>
```

no-reply: thanks! just a typo

PR #4439, opened 2020-07-14
simplify collection metadata rendering

no-reply:
Collection metadata had complex helper behavior that led to a cyclical
dependency between view code, helpers, and renderers. This rewrite simplifies
the direction of the dependency, removing an awkward `yield`.

Since these methods handle HTML safety of strings, simplicity is a major benefit
here; it's crucial to avoid confusing HTML saftey behavior.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/collections_helper.rb
+++ b/app/helpers/hyrax/collections_helper.rb
@@ -1,9 +1,29 @@
 # frozen_string_literal: true
 module Hyrax
   module CollectionsHelper
+    ##
+    # @since 3.0.0
+    # @return [#to_s]
+    def collection_metadata_label(collection, field)
+      Hyrax::PresenterRenderer.new(collection, self).label(field)
+    end
+
+    ##
+    # @since 3.0.0
+    # @return [#to_s]
+    def collection_metadata_value(collection, field)
+      Hyrax::PresenterRenderer.new(collection, self).value(field)
+    end
+
+    ##
+    # @deprecated Use #collection_metadata_label and #collection_metadata_value instead.
+    #
     # @param presenter [Hyrax::CollectionPresenter]
     # @param terms [Array<Symbol>,:all] the list of terms to draw
     def present_terms(presenter, terms = :all, &block)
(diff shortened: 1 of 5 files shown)


app/helpers/hyrax/collections_helper.rb
```diff
@@ -1,10 +1,29 @@
...
+    # @return [#to_s]
+    def collection_metadata_label(collection, field)
+      Hyrax::PresenterRenderer.new(collection, self).label(field)
+    end
+
+    ##
+    # @since 3.0.0
+    # @return [#to_s]
+    def collection_metadata_value(collection, field)
+      Hyrax::PresenterRenderer.new(collection, self).value(field)
```
no-reply, line 15: i'm not sure if this "renderer" is actually helpful, over just having these helpers do the work directly:

  - it's not the same as other things we call `Renderer`
  - it says it's a `PresenterRenderer` but it only actually works for collection presenters (due to hard-coded partial resolution; though it's only *tested* for work presenters!).
  - most of the novel logic (now deprecated) is there just to yield itself.

i didn't want to mess with it too much in this PR, since my main goal is to move us toward being able to render Valkyrie collections (with a side goal of improving the story for collection metadata customization). this at least makes the whole thing invisible to the view, which can now just trust the helpers to do their thing.

no-reply:
bumped this just to put the comment above into the commit message.

app/views/hyrax/dashboard/collections/_show_descriptions.html.erb
```diff
@@ -1,7 +1,7 @@
 <h2 class="sr-only"><%= t('.descriptions') %></h2>
 <dl class="dl-horizontal metadata-collections">
-  <% present_terms(@presenter, @presenter.terms_with_values) do |r, term| %>
-      <dt><%= r.label(term) %></dt>
-      <dd><%= r.value(term) %></dd>
+  <% @presenter.terms_with_values.each do |field_name| %>
```
jeremyf, line 3: Definitely like the change for this one.

PR #4440, opened 2020-07-14
Allow for injection of new tabs on edit page

cjcolvar:
This commit adds a new code seam allowing for the modification of tab
ordering and the addition (or removal) of tabs on a form by form basis.
The default tab order is moved out of the guts4form view partial and
into a helper.

To modify the default list of tabs in an application override the `form_tabs_for` helper method and make sure it gets loaded
after Hyrax::HyraxHelperBehavoir (by default included in app/helpers/hyrax_helper.rb). For example:
```
module WorksHelper
  def form_tabs_for(form:)
    super + ["my_new_tab"]
  end
end
```

The share tab has not been included in the default tab order because it
wasn't in the view partial.  Future work is to simplify the guts4form
partial so share can be treated the same as the other tabs and included
in the default list.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/_guts4form.html.erb
+++ b/app/views/hyrax/base/_guts4form.html.erb
@@ -1,5 +1,4 @@
 <% # we will yield to content_for for each tab, e.g. :files_tab %>
-<% tabs ||= %w[metadata files relationships] # default tab order %>
 <div class="row">
   <div class="col-xs-12 col-sm-8">
     <div class="panel panel-default tabs" role="main">
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,4 @@ _yardoc
 lib/bundler/man
 spec/reports
 /spec/examples.txt
+node_modules
--- a/app/helpers/hyrax/hyrax_helper_behavior.rb
+++ b/app/helpers/hyrax/hyrax_helper_behavior.rb
@@ -16,6 +16,7 @@ module Hyrax
     include Hyrax::IiifHelper
     include Hyrax::MembershipHelper
     include Hyrax::PermissionLevelsHelper
+    include Hyrax::WorkFormHelper
 
     # Which translations are available for the user to select
     # @return [Hash<String,String>] locale abbreviations as keys and flags as values
--- a/app/helpers/hyrax/work_form_helper.rb
+++ b/app/helpers/hyrax/work_form_helper.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+module Hyrax
+  module WorkFormHelper
+    def form_tabs_for(form:)
+      if form.instance_of? Hyrax::Forms::BatchUploadForm
+        %w[files metadata relationships]
+      else
+        %w[metadata files relationships]
+      end
+    end
+  end
+end
--- a/app/views/hyrax/base/_form.html.erb
+++ b/app/views/hyrax/base/_form.html.erb
@@ -19,7 +19,7 @@
       <p class="switch-upload-type"><%= t('.batch_upload_hint') %> <%= link_to t('.batch_link'), hyrax.new_batch_upload_path(payload_concern: @form.model.class) %></p>
     <% end %>
   <% end %>
-  <%= render 'hyrax/base/guts4form', f: f %>
+  <%= render 'hyrax/base/guts4form', f: f, tabs: form_tabs_for(form: f.object) %>
 <% end %>
 
 <script type="text/javascript">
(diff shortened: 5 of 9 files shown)


app/views/hyrax/base/_guts4form.html.erb
```diff
@@ -1,5 +1,5 @@
 <% # we will yield to content_for for each tab, e.g. :files_tab %>
-<% tabs ||= %w[metadata files relationships] # default tab order %>
+<% tabs ||= f.object.tabs # default tab order %>
```
no-reply, line 2: i'm thinking about whether the form is the best home for this.

i know the older `WorkForm` serves in some ways as both a configuration object and as a data object/presenter. my experience has been that this has been something of a pain point (e.g. i think it's the main reason we haven't been able to make forms repopulate on failed submissions). i've been trying, only somewhat successfully, to limit our repetition of this pattern in `ResourceForm`. we still have `primary/secondary` terms, for instance, which are purely display configuration.

so i guess i'm open to this, but also wondering whether there's any downside to using a simple helper? `<% form_tabs_for(form: f) %>` with an implementation something like:

```ruby
def form_tabs_for(form:)
  %w[metadata files relationships]
end
```

honestly, this could also be a viable pattern for moving away from `#primary/secondary_terms`?

----

(also, what's going on with `tabs ||=` here?! this should be a separate PR, but let's try just dropping that?)

no-reply: ah. `tabs` is being passed in by the caller in some places. maybe a good idea to just have the caller pass it in always, but definitely out of scope.

on the other hand, this does mean we need to consider what other forms would need to implement this new method, and probably provide a fall back in case applications provide their own.

cjcolvar: I'm not entirely following.  Are you suggesting using a helper or passing the tabs in or both?
My ultimate goal is having an engine plugin that can easily inject a new tab for the create/edit form without having to do any overrides and have this be on a per work-type basis.  How would a plugin add a new tab if an application already has the helper implemented?  With the form approach it seems straightforward with something like:
```
module Plugin
  module FormBehavior
    included do
      self.tabs += ['newtab']
    end
  end
end
```
Although there might be timing issues with that.  I'm open to any approaches you think I should try.

no-reply: sorry, i'll rephrase hopefully with more clarity.

what i understand to be happening now is:
  - the `guts4form` partial accepts a `tabs: [:blah]` argument;
  - if that argument is not passed in by `render`, we use `%w[metadata files relationships]` as the default;

the current state of this PR changes the second point to: `use `f.object.tabs` as the default. that seems fine as far as it goes, but:
  - i'm wondering whether putting this knowledge in the `Form` has advantages for applications over a helper which they could override.
    - i've been otherwise trying to keep the new `ResourceForm` objects as thin as backward compatibility allows, but i'm not strongly committed to that. i can say more about this if it's helpful.
    - i'm suggesting passing the form object into the helper so apps can choose implementations that switch tabs based on the form class, or other attributes of the form object (maybe this isn't needed?)
  - i think the branching behavior in "happening now" above isn't great, and we should always have `render` pass in `tabs: [:something]`, but that's probably out of scope for your goals and i don't want to foist it on you.

cjcolvar:
@no-reply I've switched over to the helper approach and added a note about how to override it in an application.  Do you mind reviewing again when you have a chance?

app/views/hyrax/base/_guts4form.html.erb
```diff
@@ -1,5 +1,4 @@
 <% # we will yield to content_for for each tab, e.g. :files_tab %>
-<% tabs ||= %w[metadata files relationships] # default tab order %>
```
no-reply, line 2: wondering if we should keep this line to support existing partial overrides on downstream apps, or just cut over now pre-3.0.0.

thoughts?

cjcolvar: I'm not sure what keeping it would do.  If it has been overridden then I think it doesn't matter because then the app would have a full copy of the partial with that line (or a modified version of it).  I guess in upgrading they would have to reapply their customization through the new helper method.
Were you thinking that downstream apps may have overrides which are calling guts4form without passing tabs and wanting to continue to support them?
Either way, I guess I'm in favor of cutting over now as part of 3.0 to establish the new pattern which engines could rely upon instead of not knowing if local customizations will work with the engine or not.  Does that seem fair?

no-reply: > Were you thinking that downstream apps may have overrides which are calling guts4form without passing tabs and wanting to continue to support them?

yeah exactly.

:+1: to cutting over.

PR #4444, opened 2020-07-15
allow approvers to approve private works. Fixes issue #4411

### LINKED ISSUE #4411: Give approver edit access for new works [one step mediated deposit]
### Descriptive summary

Assigning approver role does not grant access to newly created works with one step mediated deposit

### Rationale

When a user is assigned as an approver, they do not see the newly deposited works pending for approval. Making the approver admin solved the issue.

### Expected behavior
When a user is made an approver, they should see all the new deposits on their dashboard

### Actual behavior
When a user is made an approver, they DO NOT see all new deposits on their dashboard

1. Assign a user "approver" role 
2. Create a new work
3. Ask the "Approver" user to check their dashboard of they see the new deposit and can approve the deposit

### Related work

blancoj:
Fixes #4411
Before this fix, if a user deposited a `private` work  using mediated deposit, and there were other users assigned the approver role, it was impossible for the users  with the approver role to actually approve the work.  When visiting the work page, they would get an Unauthorized warning.  It was working fine if the work being deposited was `public`, but not with `private` works.  With this change, approvers can approve both `public` and `private` works under mediated deposit.

*  Assign a user ( user1 ) an approver role.
*  Set the Admin set to use one step mediated deposit. 
*  Have an admin or a depositor deposit a work, and make the work private.
* Logout and login as user1
* Go to the dashboard and from the sidebar go to "Review Submissions"
* The private work just deposited should be listed there, click on it.
* Approve the work.

Repeat these steps with a public work.  

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/search_filters.rb
+++ b/app/search_builders/hyrax/search_filters.rb
@@ -13,9 +13,22 @@ module Hyrax
       self
     end
 
+    # This method will return true if the user has an appoval role, and false otherwise.
+    # This is used to for approvers to gain access to a private work.  If this check were not done, the approver
+    # would not be able to access private works.
+    def user_approver?
+      approving_role = Sipity::Role.find_by(name: Hyrax::RoleRegistry::APPROVING)
+      return false unless approving_role
+      Hyrax::Workflow::PermissionQuery.scope_processing_agents_for(user: current_ability.current_user).any? do |agent|
+        agent.workflow_responsibilities.joins(:workflow_role)
+             .where('sipity_workflow_roles.role_id' => approving_role.id).any?
+      end
+    end
+
     # Override Hydra::AccessControlsEnforcement (or Hydra::PolicyAwareAccessControlsEnforcement)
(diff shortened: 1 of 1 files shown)


app/search_builders/hyrax/search_filters.rb
```diff
@@ -13,9 +13,22 @@ def with_access(access)
       self
     end
 
+    # This method will return true if the user has an appoval role, and false otherwise.
```
jeremyf, line 16: This ability check verifies that the user is an approver on any workflow, but not necessarily an approver on the work flow related to a given work.

jeremyf: If the above is an acceptable use case, I believe we need to create a PermissionQuery method that handles answering this question, as the below will run 2+N queries for a single answer.

no-reply:
i'm worried about adding this as a permission exception. wouldn't it be preferable to explicitly add the approver(s) as edit users in the relevant object's ACLs as part of the workflow?

PR #4446, opened 2020-07-16
fixup stubs on object-under-test for presenters

no-reply:
trying to tidy this test suite up enough that we can test it against valkyrie/a
clear set of dependency interfaces.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -4,6 +4,10 @@ module Hyrax
     include ModelProxy
     include PresentsAttributes
 
+    ##
+    # @!attribute [w] member_presenter_factory
+    #   @return [MemberPresenterFactory]
+    attr_writer :member_presenter_factory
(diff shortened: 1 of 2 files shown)


app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -241,12 +235,9 @@ def valid_child_concerns
     private
 
     # list of item ids to display is based on ordered_ids
-    def authorized_item_ids
-      @member_item_list_ids ||= begin
-        items = ordered_ids
-        items.delete_if { |m| !current_ability.can?(:read, m) } if Flipflop.hide_private_items?
-        items
-      end
+    def authorized_item_ids(filter_unreadable: Flipflop.hide_private_items?)
```
jeremyf, line 238: I definitely like this parameterized.

no-reply: NO 👏 HARDCODED 👏 DEPENDENCIES 👏

PR #4447, opened 2020-07-16
ensure iiif manifest order

no-reply:
still thinking about how to do a robust test for this, but this ought to
reinstate careful ordering for IIIF manifests.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/iiif_manifest_presenter.rb
+++ b/app/presenters/hyrax/iiif_manifest_presenter.rb
@@ -87,7 +87,12 @@ module Hyrax
     ##
     # @return [Array<#to_s>]
     def member_ids
-      Array(model.try(:member_ids))
+      case model
+      when Valkyrie::Resource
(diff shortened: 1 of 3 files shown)


app/presenters/hyrax/iiif_manifest_presenter.rb
```diff
@@ -87,7 +87,10 @@ def manifest_url
     ##
     # @return [Array<#to_s>]
     def member_ids
-      Array(model.try(:member_ids))
+      ordered = Array(model.try(:ordered_member_ids))
+      unordered = Array(model.try(:member_ids)) - ordered
+
+      ordered + unordered
```
no-reply, line 93: i think this is equivalent, but i like this version for its explicitness. `Array#&`'s order handling won't be very obvious to readers.

```suggestion
      Array(model.try(:ordered_member_ids)) & Array(model.try(:member_ids))
```

PR #4449, opened 2020-07-16
Stop triggering autocomplete for non-auto complete fields. Support data attribute based autocomplete types.

### LINKED ISSUE #3510: Extend Autocomplete.es6 to generate more LinkedData objects instead of just for based_near
### Descriptive summary
This ticket describes the need to be able to generate Linked Data fields on the form using the LinkedData object found here: https://github.com/samvera/hyrax/blob/master/app/assets/javascripts/hyrax/autocomplete/linked_data.es6

Currently, the only field capable of generating a LinkedData object, which in turn, generates the Select2 html and corresponding javascript to help support linked data, is based_near. This is because it is hardcoded in this file. https://github.com/samvera/hyrax/blob/master/app/assets/javascripts/hyrax/autocomplete.es6#L26

We would like to generate more fields as linked data fields, which support autocomplete from endpoints, which in turn, inject data into the hidden id field in the html and deep index through the deep_indexing_service. 

### Rationale

Being able to generate more fields as a linked data field, would raise the power of linked data in the system and improve overall metadata of the system by exposing more fields to things like alternative labels, broader or narrower terms, and other pieces of useful data. 

### Expected behavior

When a field is expressed as a linked data field, it renders the select2 input on the edit/new page of a work. 

### Actual behavior

The only field that renders a linked data field is based_near

orangewolf:
Moves to a new autocomplete type choosing system that uses data[autocomplete-type] to declaritively select
the type instead of just keying of certain field names. Does not remove the existing functionality, but is
a baby step toward depricating it.

Fixes #3510 ; refs #99 (maybe?)

* There are no specs around the autocomplete js as is, so none were modified or edited
* Only real way to test the autocomplete getting called all the time is with a JS breakpoint or debugger statement. 
    Here's a video that shows it happening on Hyku running on vanilla Hyrax 2.5.1 https://share.getcloudapp.com/d5uWdzZj (sorry about the fan noise my mac is prepping for launch)
* Adding a new field and making it linked data can be done by adding a helper like so to the `app/views/records/edit_fields/_language.html.erb` file for example

```
<%= f.input key,
      as: :multi_value,
      input_html: {
        class: 'form-control',
        data: {
          'autocomplete-type' => 'linked',
          'autocomplete-url' => "/authorities/search/local/languages",
          'autocomplete' => key
        }
      },
 required: f.object.required?(key) %>
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/autocomplete.es6
+++ b/app/assets/javascripts/hyrax/autocomplete.es6
@@ -10,6 +10,34 @@ export default class Autocomplete {
    * @param {string} url - The url for the autocompete search endpoint
    */
   setup (element, fieldName, url) {
+    if(element.data('autocomplete-type') && element.data('autocomplete-type').length > 0) {
+      this.byDataAttribute(element, url)
+    } else {
+      this.byFieldName(element, fieldName, url)
+    }
+  }
+
+  byDataAttribute(element, url) {
+    let type = element.data('autocomplete-type')
+    let exlude = element.data('exclude-work')
+    if(type === 'resource' && exclude.length > 0) {
+      new Resource(
+        element,
+        url,
+        { excluding: exclude }
+      )
+    } else if(type === 'resource' ) {
+      new Resource(
+        element,
+        url)
+    } else if(type === 'linked') {
+      new LinkedData(element, url)
+    } else {
+      new Default(element, url)
+    }
+  }
+
+  byFieldName(element, fieldName, url) {
     switch (fieldName) {
       case 'work':
         new Resource(
@@ -30,4 +58,5 @@ export default class Autocomplete {
         break
     }
   }
+
 }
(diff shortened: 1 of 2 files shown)


jeremyf:
Is this something that should be added to the master branch as well?

no-reply:
@jeremyf:

> Is this something that should be added to the master branch as well?

for future reference, our branching policy is to always merge fixes to the mainline *before* pushing them to maintenance branches. the only exception should be cases where the fix is no longer relevant on the mainline, but resolves a significant issue for the maintenance series. in other words: we should always be backporting, never forward-porting. this is to avoid complicated regression situations, where fixes released on maintenance branches aren't available on current releases.

we should either:
  - immediately forward port this, reconciling the current state; or, if that's not possible
  - revert this merge, rework for `master`, and backport from there using the usual process.

orangewolf:
@no-reply that's my bad, as I didn't know that we were only backporting. That's a bit of a lift, as I don't have anything local running against master at the moment, but I can take a swing at getting this done for master later today if no one else can tackle it.

no-reply:
no worries! i'm delighted (and relieved) to have this fix in. we'll get the branch reconciled in short order.

this might not have been the platonic ideal of the process, but it looks like the system working to me!

PR #4454, opened 2020-07-16
Expose the internal test app's rake tasks and use them to avoid having...

cjcolvar:
…to bundle install within the test app

Alternative approach to #4452 .


jeremyf:
I moved this from "draft" to "yup" as it reduces the build time by 5 minutes.

PR #4455, opened 2020-07-17
move ordered member id querying to a decorator for SolrDocument

no-reply:
instead of isolating this logic `MemberPresenterFactory`, where i need an
Ability and a request context in order to run the query, extract for use with
`SolrDocument`.

since this isn't behavior for accessing `SolrDocument` data, per se, we can
avoid clutter and accidental querying by putting it in a decorator. to use, you
need to explictly `.decorate` the `SolrDocument` instance.

this will actually work for anything that's indexed with an `#id` and that has
`ActiveFedora`-style list proxy indexes. that means it *won't* work for
`Valkyrie` resources, at least as currently setup. maybe we don't need it for
them? work still to do on order there.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/solr_document/ordered_members.rb
+++ b/app/models/concerns/hyrax/solr_document/ordered_members.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+module Hyrax
+  module SolrDocument
+    ##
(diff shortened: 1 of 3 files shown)


no-reply:
see, as motivator: #4447

no-reply:
i had to kick this to fix some style issues, and added some more complete documentation while i was at it.

PR #4460, opened 2020-07-18
reload valkyrie resource after running after_create

elrayle:
## Description of Change

If a resource created using ValkyrieCreateStrategy is changed in the after_create, those changes are lost unless `result` is reloaded with those changes.  This PR uses the query_service to find and reload the resource.

## Example of where this is required

If a work is added to a collection in after_create, it changes the work resource.  If the work is not reloaded using `query_service #find_by(id:)`, the relationship is lost.  The updated work resource needs to be returned by the ValkyrieCreateStrategy; otherwise, the test using the factory created resource will not pick up the changes to the resource.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/specs/shared_specs/factories/strategies/valkyrie_resource.rb
+++ b/lib/hyrax/specs/shared_specs/factories/strategies/valkyrie_resource.rb
@@ -10,13 +10,17 @@ class ValkyrieCreateStrategy
     result = persister.save(resource: evaluation.object)
 
     evaluation.notify(:after_create, result)
-    result
+    query_service.find_by(id: result.id)
   end
 
   private
 
   def persister
     Hyrax.persister
+  end
+
+  def query_service
+    Hyrax.query_service
   end
 end
 
@@ -28,5 +32,9 @@ class ValkyrieTestAdapterCreateStrategy < ValkyrieCreateStrategy
 
   def persister
     Valkyrie::MetadataAdapter.find(:test_adapter).persister
+  end
+
+  def query_service
+    Valkyrie::MetadataAdapter.find(:test_adapter).query_service
   end
 end


elrayle:
In factory...
```
after(:create) do |work, evaluator|
  unless evaluator.collection_factory.nil?
    col = create(evaluator.collection_factory)
    col.add_member_objects [work.id]

    ### DEBUGGING ONLY ###
    cols = Hyrax.custom_queries.find_collections_for(resource: work) # cols includes col 
  end
end
```

In valkyrie_resource.rb (ValkyrieCreateStrategy)
```
  def result(evaluation)
    evaluation.notify(:after_build, evaluation.object)
    evaluation.notify(:before_create, evaluation.object)

    result = persister.save(resource: evaluation.object)

    evaluation.notify(:after_create, result)


    ### DEBUGGING ONLY ###
    cols = Hyrax.custom_queries.find_collections_for(resource: result) # []

    result = query_service.find_by(id: result.id)

    ### DEBUGGING ONLY ###
    cols = Hyrax.custom_queries.find_collections_for(resource: result) # cols includes col 
    result
  end
```

I can create a test that changes the resource in the `after(:create)`.  I would propose putting it in `spec/factory_tests/valkyrie_resource_spec.rb`.

Do you have something in mind for an alternate approach?

no-reply:
> In factory...
> 
> ```
> after(:create) do |work, evaluator|
>   unless evaluator.collection_factory.nil?
>     col = create(evaluator.collection_factory)
>     col.add_member_objects [work.id]
> 
>     ### DEBUGGING ONLY ###
>     cols = Hyrax.custom_queries.find_collections_for(resource: work) # cols includes col 
>   end
> end
> ```

i can't find code like this in the Hyrax code base.

it seems to me like this is complicating minimum the collection membership setup substantially by trying to create the collections after the work. since collection membership data is stored on the work, it would be better to create the collections first, then pass their ids to the work factory as `member_of_collection_ids: `.

this is what the existing `:as_collection_member` and `:as_member_of_multiple_collections` factory traits do (e.g., you can do `FactoryBot.valkyrie_create(:monograph, :as_member_of_multiple_collections)`. if you want to use the transient `:collection_factory` pattern, i'd suggest looking at `after(:build)` as the place to create the collections and assign membership.

elrayle:
Per recommendation in Samvera Tech call today, I am going to try and mutate the work in the `after(:create)` to see if the change is maintained when it returns to ValkyrieCreateStrategy.

jeremyf:
Looking at the code, I suspect mutation may be a challenging path.  Below are the two relevant chunks of code.

First, the `ValkyrieCreateStrategy#result` method and then the `after(:create)` callback which the `#result` method calls via `evaluation.notify(:after_create, result)`.  There are four relevant after create blocks (`:hyrax_work`, `:hyrax_resource`, `:hyrax_collection`, `:hyrax_file_set`).

Were we using ActiveRecord::Base for the `after(:create)`, within the `after(:create)` we could call `work.reload` and mutate the work easily.  For this to work, we would need to update each of the four `after(:create)` callbacks to overwrite the attributes of the `work` object.

The `result` object in the `ValkyrieCreateStrategy#result` and the `work` objected yield in the `after(:create)` callback have the same `#object_id`.  The `evaluation.object` has a different `object_id`.

Which I wonder if it's more appropriate to mutate the `evaluation.object` instead?


```ruby
# ./lib/hyrax/specs/shared_specs/factories/strategies/valkyrie_resource.rb
class ValkyrieCreateStrategy
  def result(evaluation)
    evaluation.notify(:after_build, evaluation.object)
    evaluation.notify(:before_create, evaluation.object)

    result = persister.save(resource: evaluation.object)

    evaluation.notify(:after_create, result)
    # Before the change, instead of `query_service.find_by(id: result.id)` we had `result`
    query_service.find_by(id: result.id)
  end
  # ...
end

# ./spec/factories/hyrax_work.rb
FactoryBot.define do
  factory :hyrax_work, class: 'Hyrax::Test::SimpleWork' do
  # ..
  after(:create) do |work, evaluator|
      if evaluator.visibility_setting
        Hyrax::VisibilityWriter
          .new(resource: work)
      end

      work.permission_manager.edit_groups = evaluator.edit_groups
      work.permission_manager.edit_users  = evaluator.edit_users

    end
  # ...
end
```

elrayle:
### Short Answer to Which I wonder if it's more appropriate to mutate the evaluation.object instead?**:  
* `evaluation.object` is not available in the `after(:create)` block
* as written now, `result` is set to the return value of persisting `evaluation.object` prior to calling `after(:create)`.  They are not the same object.

### Longer Answer

#### Scenario:
* Locally defined work is put into a collection in the factory `after(:create)`  using `valkyrie_create` strategy
* Upon return to spec, the object returned from `valkyrie_create` does not have the relationship to the collection

#### Observations:
* in `ValkyrieCreateStrategy`
  * `evaluation` is available in `ValkyrieCreateStrategy` and holds `object`
  * `result` (which is returned) is set to the return value of persisting `evaluation.object` prior to calling `after(:create)`
  * `result` is passed to `after(:create)`
* in factory `after(:create)
  * receives `result` as the first parameter to the block
  * receives `evaluator` as the second parameter to the block (This holds additional information, but not the object (aka result) itself
  
#### Attempts to rectify:
* FAIL: updating `result` in `after(:create)` has no impact on `result` in `ValkyrieCreateStrategy`
* PASS: in `ValkyrieCreateStrategy`, using Hyrax.query_service to read the object back in after calling `after(:create)` works and is what was proposed in this PR, but has been identified as an undesirable solution for performance reasons

#### Summary
This tested the recommendation at Samvera Tech call and confirmed that the mutating the object in the `after(:create)` method has no impact on the object in `ValkyrieCreateStrategy`

jeremyf:
@elrayle I'll bring this up in our Hyrax daily stand-up at 1pm.

no-reply:
i'm having a hard time following the discussion here, and i think the reason is that i'm not grasping the motivation for this analysis.

this strategy implementation is in broad use here, including to create objects with collection membership. `FactoryBot.valkyrie_create(:hyrax_work, member_of_collection_ids: [my_collection.id])` works for this use case, and the `:hyrax_work` factory provides several examples of object traits that build the collections for the user (see: `trait :as_member_of_collections`).

what is this PR trying to fix? it would be helpful to have an example of a test that fails without this change.

jeremyf:
@no-reply My understanding is that the PR fixes the case where the factory returns an object that does not reflect the actually persisted state.

The spec, I assume would be something along the lines of:

```ruby
resource = factory_bot(:hyrax_work)
expect(resource.attributes).to eq(Hyrax.query_service.find(id: resource.id).attributes)
```

Prior to the above commit, this (or its approximation) would fail.

no-reply:
@jeremyf i don't think that's the case *unless* the author goes out of their way to change the persisted object without changing the in-memory copy.

is there any reason to do that?

no-reply:
> FAIL: updating result in after(:create) has no impact on result in ValkyrieCreateStrategy

i've tried to reproduce this result, without luck: https://github.com/samvera/hyrax/compare/after-create-test

elrayle:
Based on the recommendations here, I changed from `col.add_member_objects(ids)` to `work.member_of_collection_ids`.  With this change, it allows me to move this to `after(:build)` because the work no longer needs an id to set the collection.  This works and is satisfactory for my use case.

PR #4461, opened 2020-07-18
adds support for ActiveModel-safe decoration

no-reply:
we have a few `Draper::Decorators`, which we use to good effect. most recently,
we've introduced a Decorator for `SolrDocument` which layers on support for
querying order. when we decorate `SolrDocument` in this way in a presenter,
however, it breaks `Ability#can?`, url helpers, and more. the reasons for this
are rooted in `#to_model` and `#model_name` both of which we handle specially
within Hyrax. `Draper` tries to address this issue in various ways, but isn't
terribly clear about its own design goals on this front; see, e.g.
 https://github.com/drapergem/draper/pull/357

this is a bit of an inflection point. i'm half tempted to turn around completely
on using `Draper` (though Valkyrie users seem happy enough?). i'm another
fraction tempted to just only use decorators in very narrow, internal
circumstances, never exposing them to presenters or views as a strict rule.

this PR represents the alternative: commit to supporting absolutely transparent
model decoration in a way that works for `Hyrax::Naming`,
`SimpleForm`/`Hydra::Editor`, `SolrDocument`, and `Hyrax::Ability`/
`Hydra::AccessControls`.

probably if we go down this route this class will grow, and we'll have taken on
its maintenance in the bargain. so what do we think?

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -232,5 +232,14 @@ module Hyrax
              .where('sipity_workflow_roles.role_id' => approving_role.id).any?
       end
     end
+
+    def extract_subjects(subject)
+      case subject
+      when Draper::Decorator
+        super(subject.model)
+      else
+        super
+      end
+    end
   end
 end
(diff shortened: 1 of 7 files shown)


no-reply:
i think the test failures are due to rigid `expect().to receive` style tests. i'll hold on fixing them until there have been some other eyes on this.

PR #4463, opened 2020-07-20
allow for extensions of valkyrie indexer's to_solr that depends on resource custom metadata

### LINKED ISSUE #4464: indexer shared specs are not usable outside hyrax when they depend on resource custom fields
### Descriptive summary

All the indexer shared specs create the resource for the spec to use in its testing.  They set values in the resource which are tested for in the shared spec.

If an indexer overrides #to_solr and expects to be able to use a property field from the class of resource that it indexes, the call to #to_solr fail because the resource (e.g. Hyrax::Resource, Hyrax::Work, Hyrax::PcdmCollection) does not have that property defined.  In some cases, the property is required and should fail if missing.  Additionally, sometimes the value of the property is more than a simple value.  If the #to_solr customization expects that property to have a value and uses a part of that property (e.g. my_property.first), it will fail if an appropriate value is not set.

### Rationale

The shared spec cannot know customized requirements of the indexer.  As such, it cannot create a resource that will allow the #to_solr method to complete without error.

### Expected behavior

For a resource indexer with customizations to #to_solr that use a property defined in the resource, the shared spec `it_behaves_like 'a Hyrax::Resource indexer'` should pass.

### Actual behavior

For a resource indexer with customizations to #to_solr that use a property defined in the resource, the shared spec `it_behaves_like 'a Hyrax::Resource indexer'` fails with error similar to...

```
NoMethodError:
       undefined method `isbn' for #<Hyrax::Resource:0x00007f860abe0180>
```

See [when extending with custom metadata](https://github.com/samvera/hyrax/blob/9a8f725302934a6b95e52451c3d8598e0a003b85/spec/indexers/hyrax/valkyrie_work_indexer_spec.rb#L24) test in PR #4463.

NOTE: This effects all shared indexer specs and the solution needs to verify that all shared specs allow for customizations to #to_solr.

elrayle:
Fixes #4464

### Description

Allow shared spec testing of indexers with customizations to #to_solr that depend on properties defined in the resource class that the indexer indexes.

### Resolution

Update shared indexers spec to have a valid resource passed in to the shared spec.  The shared spec sets values it depends on in a before block.  

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/indexers/hyrax/valkyrie_collection_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_collection_indexer.rb
@@ -7,6 +7,7 @@ module Hyrax
     include Hyrax::ResourceIndexer
     include Hyrax::PermissionIndexer
     include Hyrax::VisibilityIndexer
+    include Hyrax::Indexer(:core_metadata)
 
     def to_solr
       super.tap do |index_document|
--- a/app/services/hyrax/thumbnail_path_service.rb
+++ b/app/services/hyrax/thumbnail_path_service.rb
@@ -5,7 +5,7 @@ module Hyrax
       # @param [#id] object - to get the thumbnail for
       # @return [String] a path to the thumbnail
       def call(object)
-        return default_image unless object.try(:thumbnail_id)&.to_s&.present?
+        return default_image if object.try(:thumbnail_id).blank?
 
         thumb = fetch_thumbnail(object)
 
--- a/app/services/hyrax/thumbnail_path_service.rb
+++ b/app/services/hyrax/thumbnail_path_service.rb
@@ -5,7 +5,7 @@ module Hyrax
       # @param [#id] object - to get the thumbnail for
       # @return [String] a path to the thumbnail
       def call(object)
-        return default_image unless object.try(:thumbnail_id)
+        return default_image unless object.try(:thumbnail_id)&.to_s&.present?
 
         thumb = fetch_thumbnail(object)
 
--- a/lib/hyrax/specs/shared_specs/indexers.rb
+++ b/lib/hyrax/specs/shared_specs/indexers.rb
@@ -1,12 +1,21 @@
 # frozen_string_literal: true
 
+# These shared specs test various aspects of valkyrie based indexers by calling the #to_solr method.
+# All tests require two variables to be set in the caller using let statements:
+#   * indexer_class - class of the indexer being tested
+#   * resource - a Hyrax::Resource that defines attributes and values consistent with the indexer
+#
+# NOTE: It is important that the resource has required values that the indexer #to_solr customizations expects to be available.
 RSpec.shared_examples 'a Hyrax::Resource indexer' do
-  subject(:indexer)  { indexer_class.new(resource: the_resource) }
-  let(:the_resource) { defined?(resource) ? resource : Hyrax::Resource.new }
+  before do
+    raise 'indexer_class must be set with `let(:indexer_class)`' unless defined? indexer_class
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::Resource' unless defined?(resource) && resource.kind_of?(Hyrax::Resource)
+    resource.alternate_ids = ids
+  end
+  subject(:indexer)  { indexer_class.new(resource: resource) }
   let(:ids)          { ['id1', 'id2'] }
 
   describe '#to_solr' do
-    before { the_resource.alternate_ids = ids }
     it 'indexes alternate_ids' do
       expect(indexer.to_solr)
         .to include(alternate_ids_sm: a_collection_containing_exactly(*ids))
@@ -15,23 +24,18 @@ RSpec.shared_examples 'a Hyrax::Resource indexer' do
 end
 
 RSpec.shared_examples 'a permission indexer' do
-  subject(:indexer) { indexer_class.new(resource: the_resource) }
-  let(:the_resource) do
-    if defined?(resource)
-      Hyrax::VisibilityWriter.new(resource: resource)
-          .assign_access_for(visibility: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC)
-      resource.permission_manager.edit_groups = edit_groups
-      resource.permission_manager.edit_users  = edit_users
-      resource.permission_manager.read_users  = read_users
-      resource.permission_manager.acl.save
-      resource
-    else
-      FactoryBot.valkyrie_create(:hyrax_work, :public,
-                                 read_users: read_users,
-                                 edit_groups: edit_groups,
-                                 edit_users: edit_users)
-    end
+  before do
+    raise 'indexer_class must be set with `let(:indexer_class)`' unless defined? indexer_class
+    # NOTE: resource must be persisted for these tests to pass
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::Resource' unless defined?(resource) && resource.kind_of?(Hyrax::Resource)
+    Hyrax::VisibilityWriter.new(resource: resource)
+        .assign_access_for(visibility: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC)
+    resource.permission_manager.edit_groups = edit_groups
+    resource.permission_manager.edit_users  = edit_users
+    resource.permission_manager.read_users  = read_users
+    resource.permission_manager.acl.save
   end
+  subject(:indexer) { indexer_class.new(resource: resource) }
   let(:edit_groups) { [:managers] }
   let(:edit_users)  { [FactoryBot.create(:user)] }
   let(:read_users)  { [FactoryBot.create(:user)] }
@@ -52,17 +56,23 @@ RSpec.shared_examples 'a permission indexer' do
 end
 
 RSpec.shared_examples 'a visibility indexer' do
-  subject(:indexer)  { indexer_class.new(resource: the_resource) }
-  let(:the_resource) { defined?(resource) ? resource : FactoryBot.build(:hyrax_work) }
+  before do
+    raise 'indexer_class must be set with `let(:indexer_class)`' unless defined? indexer_class
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::Resource' unless defined?(resource) && resource.kind_of?(Hyrax::Resource)
+    # optionally can pass in default_visibility by setting it with a let statement if your application changes the default; Hyrax defines this as 'restricted'
+    # See samvera/hyrda-head hydra-access-controls/app/models/concerns/hydra/access_controls/access_rights.rb for possible VISIBILITY_TEXT_VALUE_...'
+  end
+  subject(:indexer)  { indexer_class.new(resource: resource) }
 
   describe '#to_solr' do
-    it 'indexes visibility' do
-      expect(indexer.to_solr).to include(visibility_ssi: 'restricted')
+    it 'indexes default visibility as restricted or passed in default' do
+      expected_value = defined?(default_visibility) ? default_visibility : 'restricted'
+      expect(indexer.to_solr).to include(visibility_ssi: expected_value)
     end
 
     context 'when resource is public' do
       before do
-        Hyrax::VisibilityWriter.new(resource: the_resource)
+        Hyrax::VisibilityWriter.new(resource: resource)
           .assign_access_for(visibility: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC)
       end
 
@@ -73,9 +83,37 @@ RSpec.shared_examples 'a visibility indexer' do
   end
 end
 
+RSpec.shared_examples 'a Core metadata indexer' do
+  before do
+    raise 'indexer_class must be set with `let(:indexer_class)`' unless defined? indexer_class
+    # NOTE: The resource's class is expected to have or inherit `include Hyrax::Schema(:core_metadata)`
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::Resource' unless defined?(resource) && resource.kind_of?(Hyrax::Resource)
+    resource.title = titles
+  end
+  subject(:indexer) { indexer_class.new(resource: resource) }
+  let(:titles)      { ['Comet in Moominland', 'Finn Family Moomintroll'] }
+
+  describe '#to_solr' do
+    it 'indexes title as text' do
+      expect(indexer.to_solr)
+        .to include(title_tesim: a_collection_containing_exactly(*titles))
+    end
+
+    it 'indexes title as string' do
+      expect(indexer.to_solr)
+        .to include(title_sim: a_collection_containing_exactly(*titles))
+    end
+  end
+end
+
 RSpec.shared_examples 'a Basic metadata indexer' do
-  subject(:indexer) { indexer_class.new(resource: the_resource) }
-  let(:the_resource) { defined?(resource) ? resource : resource_class.new }
+  before do
+    raise 'indexer_class must be set with `let(:indexer_class)`' unless defined? indexer_class
+    # NOTE: The resource's class is expected to to have or inherit `include Hyrax::Schema(:basic_metadata)`
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::Resource' unless defined?(resource) && resource.kind_of?(Hyrax::Resource)
+    attributes.each { |k, v| resource.set_value(k, v) }
+  end
+  subject(:indexer) { indexer_class.new(resource: resource) }
 
   let(:attributes) do
     {
@@ -84,14 +122,7 @@ RSpec.shared_examples 'a Basic metadata indexer' do
     }
   end
 
-  let(:resource_class) do
-    Class.new(Hyrax::Work) do
-      include Hyrax::Schema(:basic_metadata)
-    end
-  end
-
   describe '#to_solr' do
-    before { attributes.each { |k, v| the_resource.set_value(k, v) } }
     it 'indexes basic metadata' do
       expect(indexer.to_solr)
         .to include(keyword_sim:   a_collection_containing_exactly(*attributes[:keyword]),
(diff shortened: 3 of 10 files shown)


lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -3,7 +3,8 @@
 RSpec.shared_examples 'a Hyrax::Resource indexer' do
   subject(:indexer) { indexer_class.new(resource: resource) }
   let(:ids)         { ['id1', 'id2'] }
-  let(:resource)    { Hyrax::Resource.new(alternate_ids: ids) }
+  let(:the_resource_class) { defined?(resource_class) ? resource_class : Hyrax::Work }
```
elrayle, line 6: Setting resource_class to Hyrax::Work when it isn't defined allows for backward compatibility.

spec/indexers/hyrax/valkyrie_collection_indexer_spec.rb
```diff
@@ -3,7 +3,12 @@
 require 'hyrax/specs/shared_specs'
 
 RSpec.describe Hyrax::ValkyrieCollectionIndexer do
-  subject(:indexer) { described_class.new(resource: resource) }
```
elrayle, line 6: The subject is not used anywhere, so it was removed.

spec/indexers/hyrax/valkyrie_collection_indexer_spec.rb
```diff
@@ -3,7 +3,12 @@
...
+  let(:resource_class) do
+    Class.new(Hyrax::Resource) do
+      # include ::Hyrax::CollectionBehavior # TODO: How is this represented in the Valkyrie::Resource version of a Collection?
```
elrayle, line 8: I could not find any example of a defined or generated Collection that is a Valkyrie::Resource.  The AF version includes CollectionBehavior.  I commented it out and left it here as a placeholder for any behaviors that are expected/required for a Valkyrie version of a Hyrax Collection.  I'd be interested in what others think should be included in this class.  NOTE: Prior to this PR, the Collection indexer was being tested with a Hyrax::Work as the resource passed to the indexer.

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -3,8 +3,7 @@
 require 'hyrax/specs/shared_specs'
 
 RSpec.describe Hyrax::ValkyrieWorkIndexer do
-  let(:indexer) { described_class.new(resource: resource) }
-  let(:resource) { FactoryBot.valkyrie_create(:hyrax_work) }
```
elrayle, line 7: indexer and resource were not used by this test, so they were removed.

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -21,4 +20,11 @@
 
     it_behaves_like 'a Basic metadata indexer'
   end
+
+  context 'when extending with custom metadata' do
```
elrayle, line 24: This context is the spec that is addressing the Actual Behavior vs. Expected Behavior described in the PR comments.

spec/support/book_resource.rb
```diff
@@ -21,6 +21,14 @@ class Book < ActiveFedora::Base
       property :publisher, predicate: ::RDF::URI('http://example.com/ns/publisher')
       property :title,     predicate: ::RDF::URI("http://example.com/ns/title")
     end
+
+    class BookResourceIndexer < Hyrax::ValkyrieWorkIndexer
+      def to_solr
```
elrayle, line 26: This defines an indexer that uses custom metadata from the resource.

lib/hyrax/specs/shared_specs/indexers.rb
elrayle: Passing in the resource_class allows #to_solr to successfully create the solr_doc when the indexer is expecting the resource to have custom metadata defined.

spec/indexers/hyrax/valkyrie_collection_indexer_spec.rb
elrayle: Found it, this should be Hyrax::PcdmCollection

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 RSpec.shared_examples 'a Hyrax::Resource indexer' do
-  subject(:indexer) { indexer_class.new(resource: resource) }
-  let(:ids)         { ['id1', 'id2'] }
-  let(:resource)    { Hyrax::Resource.new(alternate_ids: ids) }
+  subject(:indexer)  { indexer_class.new(resource: the_resource) }
+  let(:ids)          { ['id1', 'id2'] }
+  let(:the_resource) { defined?(resource) ? resource : Hyrax::Resource.new }
```
elrayle, line 6: This is providing backward compatibility.  If the caller does not create a resource, then one is created in the shared spec.  If backward compatibility is not required, this setup would become...

```
   subject(:indexer)  { indexer_class.new(resource: resource) }
   let(:ids)          { ['id1', 'id2'] }
```

NOTE: All references to `the_resource` are providing backward compatibility for each shared spec.  Similar changes could occur if `resource` was required to be defined by the caller.

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -1,11 +1,12 @@
...
 
   describe '#to_solr' do
+    before { the_resource.alternate_ids = ids }
```
elrayle, line 9: Makes sure the alternate_ids are set whether the resource is created in the shared spec or by the caller.

NOTE: Similar use of before blocks in other shared specs to be sure the resource has the values set that are being tested.

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -21,4 +20,13 @@
...
+    let(:resource) { Hyrax::Test::BookResource.new(isbn: '978-3-16-148410-0', publisher: 'The Publisher', title: 'A title') }
+    let(:indexer_class) { Hyrax::Test::BookResourceIndexer }
+
+    it_behaves_like 'a Hyrax::Resource indexer'
+    it_behaves_like 'a permission indexer'
```
elrayle, line 29: permission indexer tests currently failing with...
```
NoMethodError:
     #   undefined method `internal_resource' for nil:NilClass
```

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -21,4 +20,13 @@
...
+    it_behaves_like 'a visibility indexer'
+  end
```
elrayle, line 31: permission indexer tests currently failing with...
```
NoMethodError:
     #   undefined method `internal_resource' for nil:NilClass
```

no-reply:
this seems on the right track, in general. it might be simplified by using `shared_example` parameters as document here: https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -79,9 +91,10 @@
   end
 
   describe '#to_solr' do
+    before { attributes.each { |k, v| the_resource.set_value(k, v) } }
     it 'indexes basic metadata' do
       expect(indexer.to_solr)
-        .to include(keyword_sim: a_collection_containing_exactly(*attributes[:keyword]),
+        .to include(keyword_sim:   a_collection_containing_exactly(*attributes[:keyword]),
```
elrayle, line 97: This change just makes the a_collection... align with a_collection... in the following 2 lines.  No real code change.

spec/support/book_resource.rb
```diff
@@ -7,11 +7,11 @@ module Test
...
-      attribute :author,   Valkyrie::Types::String
-      attribute :created,  Valkyrie::Types::Date
-      attribute :isbn,     Valkyrie::Types::String
-      attribute :pubisher, Valkyrie::Types::String
-      attribute :title,    Valkyrie::Types::String
+      attribute :author,    Valkyrie::Types::String
+      attribute :created,   Valkyrie::Types::Date
+      attribute :isbn,      Valkyrie::Types::String
+      attribute :publisher, Valkyrie::Types::String
+      attribute :title,     Valkyrie::Types::String
```
elrayle, line 14: Most of this is just an alignment change.  The only real change is fixing the typo from pubisher to publisher.

app/services/hyrax/thumbnail_path_service.rb
```diff
@@ -5,7 +5,7 @@ class << self
       # @param [#id] object - to get the thumbnail for
       # @return [String] a path to the thumbnail
       def call(object)
-        return default_image unless object.try(:thumbnail_id)
+        return default_image unless object.try(:thumbnail_id)&.to_s&.present?
```
elrayle, line 8: For collections shared spec, the thumbnail_id was a Valkyrie ID with id="".  I don't know if this would happen in the application, but it doesn't hurt anything to have the extra code.  If thumbnail_id is nil, it will still use the default_image.

elrayle:
@no-reply RE: parameters to shared specs...  I like the idea of making the parameters more transparent.  I don't think it would simplify the code if we still have the requirement of providing a default if the parameter does not have a value.  With a change to parameterized shared specs, would you see it as ok to remove the creation a default resource when one is not provided making parameters required?  That would simplify the code.

elrayle:
Problem with parameters for shared specs.  You cannot use `let` or `before` to do setup on the parameters.  They have to be created on the `it_behaves_like` statement.

WORKS
```
  it_behaves_like 'a Hyrax::Resource indexer', described_class, Hyrax::Work.new
```

DOESN'T WORK  (error loading spec file)
```
  let(:indexer_class) { described_class }
  let(:resource) { Hyrax::Work.new }

  it_behaves_like 'a Hyrax::Resource indexer', indexer_class, resource
```

This example is simple.  But if you needed to make a modification to resource before sending it, there isn't a way to do that.

Also, you can't set up the resource once and have it used for all shared specs.  You have to copy the creation code to every it_behaves_like statement which isn't DRY.

```
  it_behaves_like 'a Hyrax::Resource indexer', described_class, Hyrax::Work.new
  it_behaves_like 'a Core metadata indexer', described_class, Hyrax::Work.new
  it_behaves_like 'a permission indexer', described_class, Hyrax::Work.new
  it_behaves_like 'a visibility indexer', described_class, Hyrax::Work.new
```

Any change to the resource you create requires every `it_behaves_like` to be updated.

elrayle:
Example preprocessing from our app.  We default our works to public instead of private.  So I have to fake this to allow our indexer to pass the visibility indexer shared spec.

```
  context 'FAKE work defaulting to private visibility for shared specs' do
    before do
      Hyrax::VisibilityWriter.new(resource: resource)
        .assign_access_for(visibility: Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE)
    end
    it_behaves_like 'a visibility indexer'
  end
```

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -21,4 +19,32 @@
...
+              end
+            end
+          end
+        end
+      end
+    end
+    after { Hyrax::Test.send(:remove_const, :Custom) }
+
+    let(:resource) { Hyrax.persister.save(resource: Hyrax::Test::Custom::Work.new(broader: ['term1', 'term2'])) }
+    let(:indexer_class) { Hyrax::Test::Custom::WorkIndexer }
```
jeremyf, line 43: Per conversation, this was already supported.

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -21,4 +19,32 @@
...
+                solr_doc['broader_ssim'] = resource.broader.first
+              end
+            end
+          end
+        end
+      end
+    end
+    after { Hyrax::Test.send(:remove_const, :Custom) }
+
+    let(:resource) { Hyrax.persister.save(resource: Hyrax::Test::Custom::Work.new(broader: ['term1', 'term2'])) }
```
jeremyf, line 42: Per tech call, this is the line that the PR is supporting.

spec/indexers/hyrax/resource_indexer_spec.rb
```diff
@@ -3,6 +3,7 @@
 require 'hyrax/specs/shared_specs'
 
 RSpec.describe Hyrax::ResourceIndexer do
+  let(:resource) { Hyrax::Resource.new }
```
jeremyf, line 6: This is added to clarify and ensure the proper behavior. Prior to this, we had misleading successful test.

app/services/hyrax/thumbnail_path_service.rb
jeremyf: This may be a separate commit. There may be other problems involved. @no-reply can you take action on this?

jeremyf: We may need to test that Valkyrie::ID with an empty string is not present.

lib/hyrax/specs/shared_specs/indexers.rb
jeremyf: Based on conversations, the reason for these conditionals is for backwards compatibility.

jeremyf: As this was not released in a Hyrax version, we can remove the conditionals and tighten up the logic. Anyone that's using this as written will have a broken spec.

app/indexers/hyrax/valkyrie_collection_indexer.rb
```diff
@@ -7,6 +7,7 @@ class ValkyrieCollectionIndexer < Hyrax::ValkyrieIndexer
     include Hyrax::ResourceIndexer
     include Hyrax::PermissionIndexer
     include Hyrax::VisibilityIndexer
+    include Hyrax::Indexer(:core_metadata)
```
elrayle, line 10: Collection resource defines core metadata, so seems like the collection indexer should index it.

app/services/hyrax/thumbnail_path_service.rb
```diff
@@ -5,7 +5,7 @@ class << self
       # @param [#id] object - to get the thumbnail for
       # @return [String] a path to the thumbnail
       def call(object)
-        return default_image unless object.try(:thumbnail_id)
+        return default_image unless object.try(:thumbnail_id)&.present?
```
elrayle, line 8: Catches the case where thumbnail_id is nil and when it is a Valkyrie::ID with an empty string.

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -61,9 +82,37 @@
   end
 end
 
+RSpec.shared_examples 'a Core metadata indexer' do
```
elrayle, line 85: This is mostly the same with the similar changes for variable validation and removing the default resource.  It was moved to make the ordering of the various shared specs easier to follow.

spec/indexers/hyrax/resource_indexer_spec.rb
elrayle: resource is now required

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -72,26 +121,40 @@
...
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::Work' unless defined?(resource) && resource.kind_of?(Hyrax::Work)
+    # optionally can pass in default_visibility by setting it with a let statement if your application changes the default; Hyrax defines this as 'restricted'
+    # See samvera/hyrda-head hydra-access-controls/app/models/concerns/hydra/access_controls/access_rights.rb for possible VISIBILITY_TEXT_VALUE_...'
+  end
+  subject(:indexer) { indexer_class.new(resource: resource) }
+
+  it_behaves_like 'a Hyrax::Resource indexer'
+  it_behaves_like 'a Core metadata indexer'
+  it_behaves_like 'a permission indexer'
+  it_behaves_like 'a visibility indexer'
```
elrayle, line 146: The shared spec defines the minimal requirements for a work instead of expecting the know which parts are required.

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -72,26 +121,40 @@
...
 RSpec.shared_examples 'a Collection indexer' do
+  before do
+    raise 'indexer_class must be set with `let(:indexer_class)`' unless defined? indexer_class
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::PcdmCollection' unless defined?(resource) && resource.kind_of?(Hyrax::PcdmCollection)
+  end
   subject(:indexer) { indexer_class.new(resource: resource) }
-  let(:resource)    { Hyrax::PcdmCollection.new }
 
+  it_behaves_like 'a Hyrax::Resource indexer'
+  it_behaves_like 'a Core metadata indexer'
```
elrayle, line 157: The shared spec defines the minimal requirements for a collection instead of expecting the  individual collection tests to know which parts are required.

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -107,21 +170,3 @@
     end
   end
 end
-
-RSpec.shared_examples 'a Core metadata indexer' do
```
elrayle, line 111: Moved higher in the file

spec/indexers/hyrax/valkyrie_collection_indexer_spec.rb
```diff
@@ -3,9 +3,8 @@
 require 'hyrax/specs/shared_specs'
 
 RSpec.describe Hyrax::ValkyrieCollectionIndexer do
-  subject(:indexer) { described_class.new(resource: resource) }
+  let(:resource) { FactoryBot.valkyrie_create(:hyrax_collection) }
   let(:indexer_class) { described_class }
 
-  it_behaves_like 'a Hyrax::Resource indexer'
```
elrayle, line 9: All shared specs representing the minimal required parts of a collection are now defined and tested in the shared spec 'a Collection indexer'.

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -3,22 +3,48 @@
...
 RSpec.describe Hyrax::ValkyrieWorkIndexer do
-  let(:indexer) { described_class.new(resource: resource) }
   let(:resource) { FactoryBot.valkyrie_create(:hyrax_work) }
   let(:indexer_class) { described_class }
 
-  it_behaves_like 'a Hyrax::Resource indexer'
-  it_behaves_like 'a Core metadata indexer'
-  it_behaves_like 'a permission indexer'
-  it_behaves_like 'a visibility indexer'
+  it_behaves_like 'a Work indexer'
```
elrayle, line 9: All shared specs representing the minimal required parts of a work are now defined and tested in the shared spec 'a Work indexer'.

spec/hyrax/indexer_spec.rb
```diff
@@ -13,6 +13,7 @@
   end
 
   context 'with core metadata schema' do
+    let(:resource) { work }
```
elrayle, line 16: resource is required for shared indexer specs

spec/indexers/hyrax/permission_indexer_spec.rb
```diff
@@ -8,6 +8,7 @@
       include Hyrax::PermissionIndexer
     end
   end
+  let(:resource) { Hyrax::Resource.new }
```
elrayle, line 11: resource is required for shared indexer specs

app/indexers/hyrax/valkyrie_collection_indexer.rb
no-reply: :+1:

app/services/hyrax/thumbnail_path_service.rb
no-reply, line 8:
```suggestion
        return default_image unless object.try(:thumbnail_id).present?
```

no-reply: `nil` responds to `#present?`. i doubt this matters beyond visual clutter, but let's skip the `&`

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -1,9 +1,19 @@
 # frozen_string_literal: true
 
+# These shared specs test various aspects of valkyrie based indexers by calling the #to_solr method.
```
no-reply, line 3: the docs are really appreciated!

lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -1,9 +1,19 @@
...
+#   * resource - a Hyrax::Resource that defines attributes and values consistent with the indexer
+#
+# NOTE: It is important that the resource has required values that the indexer #to_solr customizations expects to be available.
 RSpec.shared_examples 'a Hyrax::Resource indexer' do
-  subject(:indexer) { indexer_class.new(resource: resource) }
-  let(:ids)         { ['id1', 'id2'] }
-  let(:resource)    { Hyrax::Resource.new(alternate_ids: ids) }
+  before do
+    raise 'indexer_class must be set with `let(:indexer_class)`' unless defined? indexer_class
+    raise 'resource must be set with `let(:resource)` and is expected to be a kind of Hyrax::Resource' unless defined?(resource) && resource.kind_of?(Hyrax::Resource)
```
no-reply, line 12: i feel like the type checks might be heavy-handed, but let's keep them for now.

PR #4465, opened 2020-07-21
[Backport] Provide patterns and infrastructure for assigning remote identifiers (DOI, Handle, etc.)

cjcolvar:
There are three patterns provided: dispatcher, registrar, and builder
Dispatcher - assigns registered identifer to a given object
Registrar - handles communication with external identifier service
Builder - constructs identifier to submit to external identifier service

Registrar implementations just need to implement the `registrar!`
method.  They can be tested with the provided shared spec and then
registered with Hyrax by the identifier_registrars configuration (which
is generated commented out in the hyrax initializer).
identifier_registrars should be a Hash with Symbol keys and Class values.

A custom builder implementation can be injected into your registrar by
overriding the registrar's initialize method setting the custom builder
as the default value for the builder keyword argument.

def initialize(builder: MyCustomBuilder.new)
  super(builder: builder)
end

With this infrastructure in place, a new remote identifier can be
assigned to a work by calling the dispatcher with the work object.
Assuming a :datacite registrar has been registered in Hyrax's
configuration then this would look like:

Hyrax::Identifier::Dispatcher.for(:datacite).assign_for!(object: work)

This will set the remote identifier in the work's identifier attribute
and save the work.  To avoid saving the object use `assign_for` instead.
If a different attribute is desired the pass the attribute as a symbol
in the :attribute keyword argument to `assign_for!`.

This work is ported from mahonia (which had parts ported from epigaea).
Both of those implementations were done by @no-reply.

Backport of #4458 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/identifier/builder.rb
+++ b/app/services/hyrax/identifier/builder.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+module Hyrax
+  module Identifier
+    ##
+    # Builds an identifier string.
+    #
+    # Implementations must accept a `prefix:` to `#initialize`, and a `hint:` to
+    # `#build`. Either or both may be used at the preference of the specific
+    # implementer or ignored entirely when `#build` is called.
+    #
+    # @example
+    #   builder = Hyrax::Identifier::Builder.new(prefix: 'moomin')
+    #   builder.build(hint: '1') # => "moomin/1"
+    class Builder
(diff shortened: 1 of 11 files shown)


cjcolvar:
Sorry @jeremyf I had a rubocop issue so I had to repush.  I wonder why it flagged here but not on the default branch?

PR #4466, opened 2020-07-22
[backport] ensure iiif manifest order

no-reply:
reinstates careful ordering for IIIF manifests.

backports #4447 and #4455 
supersedes #4448 

@samvera/hyrax-code-reviewers


no-reply:
@cjcolvar i think this unblocks a 2.x release.

PR #4470, opened 2020-07-25
add docker and docker-compose as a supported development environment

no-reply:
our normal development stack involves Postgres, Fedora, Solr, and Redis; rather
than put it on the developer to configure all of this, we can provide containers
configured to run the app.

this creates a parallel approach to `.internal_test_app`, with `.hyrax_app` as
the application path by default and `engine_cart` replaced with `rails new` as
the preferred way to generate it.

the included `Dockerfile` is intended to grow to support downstream
applications; to this end it offers arguments for Ruby version, app path, and a
mechanism to inject system dependencies.

all of this could probably use some work; especially:

  - some Hyrax system dependencies are missing,
  - there's no support for live code updates, because:
    - the engine isn't mounted in the dev containers,
    - the mounted `.:/data` volume isn't updating either (for reasons unknown).
  - admin set creation should be added (i tried but it didn't work immediately);

...but this is progress.

@samvera/hyrax-code-reviewers


no-reply:
https://github.com/samvera/hyrax/tree/cn-2 is a better branch for this work now.

PR #4471, opened 2020-07-25
Less db migration

no-reply:
Hyrax has previously run db migrations and created collection types as part of
the install process. both of these choices are questionable:

for DB migrations, it is almost always going to be the case that the user will
want to change the database configuration after running the install
generator. running the migrations means than `rails new` fails for Hyrax when
the database isn't configured. if the database *is* configured, the migrations
put the data in the user's (hopefully development) database at risk when all
they've done is `rails new`.

there are lots of opportunities for the user to run `db:migrate` later in their
development process. it's best avoided here.

creating the `default_collection_types` has similar issues: it directly impact's
the users DB, before they've really had an opportunity to configure it. it also
fails with respect to internationalization (the script run at this point
actually injects empty key notices into the database!). maybe this can be moved
to seeds?

@samvera/hyrax-code-reviewers


no-reply:
i was hoping that this would free us from a mandatory `db:migrate` on `engine_cart:generate`, which would avoid the need to have a database available on the system when the app is generated.

no luck, though. [`engine_cart` hard codes the migration](https://github.com/cbeer/engine_cart/blob/939ef8b775ce58355e7cccb1dbe1e58ffbd61656/lib/engine_cart/tasks/engine_cart.rake#L111).

this still seems like an improvement, though. the ability to regenerate the test app without destroying development data seems like a feature, rather than something to work around.

PR #4480, opened 2020-08-14
Removes upper limit on QA

### LINKED ISSUE #4479: Restrictive pin on QA
This affects Hyrax 3.0.0-beta2 to HEAD, and does not affect the 2.x series.

In Hyrax 3.0.0-beta2, a new restriction was introduced to keep QA from upgrading to 4.0.0. Last week, the Library of Congress authorities started redirecting http queries to https. QA versions less than 5.5.0 cannot handle this change.

The effect today is that Hyrax implementations using the affected versions are unable to use Library of Congress authorities.

I propose removing the pin.

rotated8:
Fixes #4479- Library of Congress authorities broken in QA versions less than 5.5.0.

Looking at release notes for QA 4+, I didn't see an obvious need for the pin. Dependabot introduced the pin in #3439, so it feels safe to remove.

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -69,7 +69,7 @@ SUMMARY
   spec.add_dependency 'oauth2', '~> 1.2'
   spec.add_dependency 'posix-spawn'
   spec.add_dependency 'power_converter', '~> 0.1', '>= 0.1.2'
-  spec.add_dependency 'qa', '>= 2', '< 4' # questioning_authority
+  spec.add_dependency 'qa', '~> 5.5', '>= 5.5.1' # questioning_authority
   spec.add_dependency 'rails_autolink', '~> 1.1'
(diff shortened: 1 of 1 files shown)


no-reply:
> Fixes #4479- Library of Congress authorities broken in QA versions less than 5.5.0.

it's unfortunate that QA versions in widespread use are broken. is it an option to backport a fix there?

i'm generally hesitant about open-ended dependencies. what are the trade-offs for bumping support to `< 6` vs the change as proposed?

elrayle:
The links are broken in all QA releases except the latest which is now [5.5.1](https://github.com/samvera/questioning_authority/releases/tag/v5.5.1).  The breakage isn’t due to a change in QA.  It is due to a change at LOC which we can’t control.

I can backport to the latest releaes on each major release path going back 2 years?  That would include v3.0.0 and v4.2.5.

Issue #4487, opened 2020-08-26
Solr docs generated by ActiveFedora and Hyrax::ValkyrieWorkIndexer are not compatible


elrayle:
### Descriptive summary

I would expect the solr doc created by Hyrax::ValkyrieWorkIndexer to generate the same set of fields as the ActiveFedora generated solr doc.

### Rationale

With both processes producing the same solr doc fields, existing solr indices will not need to be migrated when switching from ActiveFedora to Valkyrie.

### Expected behavior

Generated fields would have the same field names and values.

### Actual behavior

![image](https://user-images.githubusercontent.com/6855473/91320709-d8ae6680-e78b-11ea-9ba5-a6d9e1061eef.png)

NOTE: Excluded all custom fields in our indexer.

### Steps to reproduce the behavior

In my local app, I created a custom indexer which inherits from Hyrax::ValkyrieWorkIndexer.  I use the following in a spec to create the publication work which is Valkyrie::Resource, save it, and index it.  

```
  let!(:publication) do
    pub = FactoryBot.valkyrie_create(:publicationr, identifier: identifier)
    solr_doc = PublicationrIndexer.new(resource: pub).to_solr
byebug
    Hyrax::SolrService.new(use_valkyrie: true).add(solr_doc, commit: true)
    pub
  end
```

* `valkyrie_create` saves the work through ActiveFedora creating the ActiveFedora solr doc.  
* `Hyrax::SolrService` saves the ValkyrieWorkIndexer generated solr doc to a separate solr core.

elrayle:
Per discussion in Samvera-Tech call today, the diffs reflect work in progress.  The path forward is to create PRs that move toward making the Hyrax::ValkyrieWorkIndexer consistent with the AF generated solr doc.

elrayle:
### Related Issues

PR #4780 Add indexing for Hyrax::FileSet through valkyrie indexing 

Issue #4795 - Hyrax::Works do not show Date Added in Dashboard -> Works
Issue #4788 - Hyrax::Works do not show thumbnail in Dashboard -> Works
Issue #4794 - Hyrax::Works do not index file sets

elrayle:
This is mostly complete.  It appears that permissions is the only part that is not in the Valkyrie solr document.  I opened Issue #4901 to cover just that part.

New comparison of the solr docs created by ActiveFedora and Valkyrie indexing...
[...]

PR #4488, opened 2020-08-26
add human readable type to valkyrie generated solr doc

elrayle:
partial fix for #4487

See also, PR #4489 - adds has_model_tesim

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/specs/shared_specs/indexers.rb
+++ b/lib/hyrax/specs/shared_specs/indexers.rb
@@ -19,6 +19,7 @@ RSpec.shared_examples 'a Hyrax::Resource indexer' do
     it 'indexes base resource fields' do
       expect(indexer.to_solr)
         .to include(has_model_ssim: resource.class.name,
+                    human_readable_type_tesim: resource.human_readable_type,
                     alternate_ids_sim: a_collection_containing_exactly(*ids))
     end
   end
--- a/spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
+++ b/spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
@@ -9,23 +9,22 @@ RSpec.describe Hyrax::ValkyrieWorkIndexer do
   it_behaves_like 'a Work indexer'
 
   context 'when extending with basic metadata' do
-    before do
-      module Hyrax::Test
-        module Basic
-          class Work < Hyrax::Work
-            include Hyrax::Schema(:basic_metadata)
-          end
-        end
-      end
-    end
-    after { Hyrax::Test.send(:remove_const, :Basic) }
-
     let(:indexer_class) do
       Class.new(described_class) do
         include Hyrax::Indexer(:basic_metadata)
       end
     end
-    let(:resource) { Hyrax::Test::Basic::Work.new }
+    let(:resource) do
+      Class.new(Hyrax::Work) do
+        # Included to address a `ArgumentError: Class name cannot be
+        # blank. You need to supply a name argument when anonymous
+        # class given`"
+        def self.model_name
+          ActiveModel::Name.new(self, nil, "TemporaryResource")
+        end
+        include Hyrax::Schema(:basic_metadata)
+      end.new
+    end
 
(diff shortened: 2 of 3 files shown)


lib/hyrax/specs/shared_specs/indexers.rb
```diff
@@ -18,7 +18,8 @@
   describe '#to_solr' do
     it 'indexes alternate_ids' do
```
no-reply, line 19: if we're changing the scope of this test, we should also change its description.

config/metadata/core_metadata.yaml
```diff
@@ -1,4 +1,8 @@
 attributes:
+  human_readable_type:
```
no-reply, line 2: doesn't this add `human_readable_type` to `Hyrax::Schema(:core_metadata)`??

have you explored the impact of this? would it be better to put this in `ResourceIndexer` alongside `alternate_ids`?

elrayle: Yes, I was just looking at this again as I was adding has_model and agree it is better in ResourceIndexer.

spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -9,16 +9,23 @@
...
       Class.new(described_class) do
         include Hyrax::Indexer(:basic_metadata)
       end
     end
-    let(:resource) do
-      Class.new(Hyrax::Work) do
-        include Hyrax::Schema(:basic_metadata)
-      end.new
-    end
+    let(:resource) { Hyrax::Test::Basic::Work.new }
```
elrayle, line 28: This change is required to avoid the following error when attempting to extract the human readable name from the class name.

```
ArgumentError:
       Class name cannot be blank. You need to supply a name argument when anonymous class given
```

no-reply: what's the rest of the stack trace.

i'd rather make the new code resilient to the current test cases than change the expectations on the collaborator objects, all else being equal

jeremyf: Looking at the code, I suspect mutation may be a challenging path.  Below are the two relevant chunks of code.

First, the `ValkyrieCreateStrategy#result` method and then the `after(:create)` callback which the `#result` method calls via `evaluation.notify(:after_create, result)`.  There are four relevant after create blocks (`:hyrax_work`, `:hyrax_resource`, `:hyrax_collection`, `:hyrax_file_set`).

Were we using ActiveRecord::Base for the `after(:create)`, within the `after(:create)` we could call `work.reload` and mutate the work easily.  For this to work, we would need to update each of the four `after(:create)` callbacks to overwrite the attributes of the `work` object.

The `result` object in the `ValkyrieCreateStrategy#result` and the `work` objected yield in the `after(:create)` callback have the same `#object_id`.  The `evaluation.object` has a different `object_id`.

Which I wonder if it's more appropriate to mutate the `evaluation.object` instead?  That would mean updating the `persister.save` method.


```ruby
# ./lib/hyrax/specs/shared_specs/factories/strategies/valkyrie_resource.rb
class ValkyrieCreateStrategy
  def result(evaluation)
    evaluation.notify(:after_build, evaluation.object)
    evaluation.notify(:before_create, evaluation.object)

    result = persister.save(resource: evaluation.object)

    evaluation.notify(:after_create, result)
    # Before the change, instead of `query_service.find_by(id: result.id)` we had `result`
    query_service.find_by(id: result.id)
  end
  # ...
end

# ./spec/factories/hyrax_work.rb
FactoryBot.define do
  factory :hyrax_work, class: 'Hyrax::Test::SimpleWork' do
  # ..
  after(:create) do |work, evaluator|
      if evaluator.visibility_setting
        Hyrax::VisibilityWriter
          .new(resource: work)
      end

      work.permission_manager.edit_groups = evaluator.edit_groups
      work.permission_manager.edit_users  = evaluator.edit_users

    end
  # ...
end
```

PR #4491, opened 2020-08-31
document what exists of our container publishing strategy

no-reply:
adds some documentation about publishing containers to https://hub.docker.com/u/samveralabs

@samvera/hyrax-code-reviewers

### DIFF
--- a/CONTAINERS.md
+++ b/CONTAINERS.md
@@ -32,11 +32,13 @@ It also runs database migrations.
 
 ## Hyrax Image
 
-We also provide a base image which can be reused for Hyrax applications: `hyrax`.
+We also provide a base image which can be reused for your Hyrax applications: `hyrax`.
 
 ```sh
 echo "FROM samveralabs/hyrax" > Dockerfile
 ```
+
+_This is for applications that mount Hyrax and is separate from the docker containers for Hyrax engine development._
 
 ### Maintaining
 
--- a/CONTAINERS.md
+++ b/CONTAINERS.md
@@ -35,17 +35,33 @@ It also runs database migrations.
 We also provide a base image which can be reused for Hyrax applications: `hyrax`.
 
 ```sh
-echo "FROM hyrax" > Dockerfile
+echo "FROM samveralabs/hyrax" > Dockerfile
 ```
 
 ### Maintaining
 
-At the moment, we don't publish hyrax images anywhere (TODO). If you want to use them, you need to build them yourself:
+We publish several Hyrax images to hub.docker.com under the
+[`samveralabs` group][dockerhub-samveralabs]. To build them, do:
 
 ```sh
 # build an image for an app using Postgresql (`gem 'pg'`)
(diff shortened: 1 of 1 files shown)


jeremyf:
I'm going to step through the documentation to verify.

CONTAINERS.md
```diff
@@ -35,17 +35,33 @@ It also runs database migrations.
 We also provide a base image which can be reused for Hyrax applications: `hyrax`.
 
 ```sh
-echo "FROM hyrax" > Dockerfile
+echo "FROM samveralabs/hyrax" > Dockerfile
```
jeremyf, line 38: As someone Docker naive, from where would I run this? I don't think I'd want to run it in the root of my hyrax git repository.

CONTAINERS.md
```diff
@@ -35,17 +35,33 @@ It also runs database migrations.
...
 
-At the moment, we don't publish hyrax images anywhere (TODO). If you want to use them, you need to build them yourself:
+We publish several Hyrax images to hub.docker.com under the
+[`samveralabs` group][dockerhub-samveralabs]. To build them, do:
 
 ```sh
 # build an image for an app using Postgresql (`gem 'pg'`)
-docker build --target hyrax --tag hyrax .
+docker build --target hyrax --tag samveralabs/hyrax:(git rev-parse HEAD) .
+docker push samveralabs/hyrax:(git rev-parse HEAD)
```
jeremyf, line 49: I'm assuming the `docker push` command would push up the image to `samveralabs`.  Is that correct?  Should we document that? What about how to get permission to do that?

CONTAINERS.md
no-reply: in the root of your own hyrax application.

maybe the context of this isn't clear?

CONTAINERS.md
no-reply: that's right. this is to publish the image.

right now, there's no strategy for handing out permissions. i had to bang the drum to get them myself.

CONTAINERS.md
jeremyf: It's not immediately clear, but it's not inscrutable.

CONTAINERS.md
```diff
@@ -32,20 +32,38 @@ It also runs database migrations.
...
 
-We also provide a base image which can be reused for Hyrax applications: `hyrax`.
+We also provide a base image which can be reused for your Hyrax applications: `hyrax`.
 
 ```sh
-echo "FROM hyrax" > Dockerfile
+echo "FROM samveralabs/hyrax" > Dockerfile
 ```
 
+_This is for applications that mount Hyrax and is separate from the docker containers for Hyrax engine development._
```
no-reply, line 41: one of the things i hope this documentation will eventually capture is that these containers are separate, in that we have to take action to maintain them separately; but also that they are closely related. the two images use the same base, and (hopefully) the `hyrax-engine-dev` is the same as the `hyrax` image in every substantial way.

PR #4493, opened 2020-09-02
Restoring anonymous resource declaration in specs

jeremyf:
At issue is "human_readable_type" leverages the resource's model name,
which, by default, uses ActiveModel::Name.  By adding `.model_name` that
returns an instance of ActiveModel::Name, the tests now pass.

I could imagine adding a spec helper method for generating an anonymous
resource.  This method would encapsulate the behavior.  If we were to go
this route, we'd want to update all of our existing `Class.new` declarations.

```ruby
module Hyrax::SpecSupport
  def self.build_anonymous_resource_class(klass, name: "Anonymous", &block)
     Class.new(klass) do
       def self.model_name
         ActiveModel::Name.new(self, nil, name)
       end
       instance_exec(&block)
     end
  end
end

Hyrax::SpecSupport.build_anonymous_resource_class(Hyrax::Work, name: "Anonymouse") do
  include Hyrax::Schema(:basic_metadata)
end
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
+++ b/spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
@@ -9,23 +9,22 @@ RSpec.describe Hyrax::ValkyrieWorkIndexer do
   it_behaves_like 'a Work indexer'
 
   context 'when extending with basic metadata' do
-    before do
-      module Hyrax::Test
-        module Basic
-          class Work < Hyrax::Work
-            include Hyrax::Schema(:basic_metadata)
-          end
-        end
-      end
-    end
-    after { Hyrax::Test.send(:remove_const, :Basic) }
-
     let(:indexer_class) do
       Class.new(described_class) do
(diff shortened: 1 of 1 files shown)


spec/indexers/hyrax/valkyrie_work_indexer_spec.rb
```diff
@@ -9,23 +9,22 @@
...
     end
-    let(:resource) { Hyrax::Test::Basic::Work.new }
+    let(:resource) do
+      Class.new(Hyrax::Work) do
+        # Included to address a `ArgumentError: Class name cannot be
+        # blank. You need to supply a name argument when anonymous
+        # class given`"
+        def self.model_name
+          ActiveModel::Name.new(self, nil, "TemporaryResource")
+        end
```
elrayle, line 24: @no-reply Lines 22-24 represent the change from master to address the ArgumentError.  Does this change fit with your expectation for code resiliency?

PR #4494, opened 2020-09-02
Fixing Works Created Count Discrepency - #3478

jeremyf:
## Adjusting logic to ensure use of local variable

b0b5dc46b6f143406ee3259032b8e0abb00a8d96

Prior to this commit, we could be passing the application controller's
`params` instead of of the local variable `params`.  This could result
in some unexpected behavior.

Note, in SHA @b04c41d2772505d8a548c3fd48774a5642d3ac90 we went from:

```ruby
params = { search_field: name, q: "\"#{value}\"" }
state = search_state_with_facets(params, facet_hash)
```

to

```ruby
params = { search_field: name, q: "\"#{value}\"" } if name != ''
state = search_state_with_facets(params, facet_hash)
```

The prior change meant that instead of always having the local variable
named `params`, we would sometimes fallback to the application's
`params` method.

I don't think this is a major issue, but one that could introduce some
unexpected results.

Related to #4322, #3478

## Introducing where clause for number_of_works()

980dc92ce0611222c3facb0285446fba4832b8f1

Based on some investigation around #3478, I suspect that the
`Hyrax::WorkRelation#where` call that uses only a depositor field and
user key is finding more results than the query run on the catalog
search page.

By adding the where clause, I hope to align the query logic of the
`number_of_works` and the `link_to_field` on the [_vitals.html.erb][1] partial.

Related to #3478

[1]: https://github.com/samvera/hyrax/blob/b04c41d2772505d8a548c3fd48774a5642d3ac90/app/views/hyrax/users/_vitals.html.erb#L11-L12

## Adding where clause to align counts on two pages

212cf164410dbe5c301ba5525a71c320c36a6c91

On the "vitals" page, there is a count that we derive from
`number_of_works`.  When you click on the `link_to_field` that new page
also as a count based on the provided facets.  Prior to this commit, I
believe the underlying queries were not using the same logic.

With this change, I hope to bring this in alignment.

Please note, I don't have tests to confirm this, nor a data state that
would help confirm it.  So, I'm pushing up a change that I believe
should work.

My plan is to write some helper tests at a later date, but for now am
drawing to the end of my work day.

Related to #3478

## Moving where clause into helper

7abf7578d4a01a136aae273a72a5f5e351813bc6

Per discussion on https://github.com/samvera/hyrax/pull/4494, a view
likely should not know about the Solr document field names.  This is
better served in a helper.

### DIFF
--- a/app/views/hyrax/users/_vitals.html.erb
+++ b/app/views/hyrax/users/_vitals.html.erb
@@ -8,8 +8,7 @@
 </div>
 
 <div class="list-group-item">
-  <%# "Note the `{ generic_type_sim: "Work" }` for number_of_works aligns with the link_to_field's {generic_type: 'Work'} key/value pair." %>
-  <span class="badge"><%= number_of_works(user, where: { generic_type_sim: "Work" }) %></span>
+  <span class="badge"><%= number_of_works(user) %></span>
   <span class="glyphicon glyphicon-upload" aria-hidden="true"></span> <%= link_to_field('', '', t("hyrax.dashboard.stats.works"), {generic_type: "Work", depositor: user.to_s}) %>
 
   <ul class="views-downloads-dashboard list-unstyled">
--- a/app/views/hyrax/users/_vitals.html.erb
+++ b/app/views/hyrax/users/_vitals.html.erb
@@ -8,7 +8,8 @@
 </div>
 
 <div class="list-group-item">
-  <span class="badge"><%= number_of_works(user) %></span>
+  <%# "Note the `{ generic_type_sim: "Work" }` for number_of_works aligns with the link_to_field's {generic_type: 'Work'} key/value pair." %>
+  <span class="badge"><%= number_of_works(user, where: { generic_type_sim: "Work" }) %></span>
   <span class="glyphicon glyphicon-upload" aria-hidden="true"></span> <%= link_to_field('', '', t("hyrax.dashboard.stats.works"), {generic_type: "Work", depositor: user.to_s}) %>
 
   <ul class="views-downloads-dashboard list-unstyled">
--- a/app/helpers/hyrax/dashboard_helper_behavior.rb
+++ b/app/helpers/hyrax/dashboard_helper_behavior.rb
@@ -5,7 +5,11 @@ module Hyrax
       params[:controller].match(%r{^hyrax/dashboard|hyrax/my})
     end
 
-    def number_of_works(user = current_user, where: {})
+    # @param user [User]
+    # @param where [Hash] applied as the where clause when querying the Hyrax::WorkRelation
+    #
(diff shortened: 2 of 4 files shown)


cjcolvar:
@jeremyf I didn't do a full dive on your code and I'm a bit fuzzy on this area of the codebase, but I would have thought that using `Hyrax::WorkRelation` would have scoped the results to only works such that you shouldn't need to send that extra where condition.  Other parts of the code are using `Hyrax::WorkRelation` as well so I'm concerned that they're broken as well.

jeremyf:
@cjcolvar That is definitely a possible concern.  I'm not overly familiar with this as well.  A real challenge is that to test "does this work" requires a complex state.  And from that complex state, we could kick the tires of the other Hyrax::WorkRelation calls.

app/views/hyrax/users/_vitals.html.erb
```diff
@@ -8,7 +8,8 @@
 </div>
 
 <div class="list-group-item">
-  <span class="badge"><%= number_of_works(user) %></span>
+  <%# "Note the `{ generic_type_sim: "Work" }` for number_of_works aligns with the link_to_field's {generic_type: 'Work'} key/value pair." %>
+  <span class="badge"><%= number_of_works(user, where: { generic_type_sim: "Work" }) %></span>
```
no-reply, line 12: can we extract this where argument from the view? i'm not a fan of view code knowing about index configuration details.

maybe the `where:` argument could be the default for `number_of_works`?

no-reply (dismissed):
see question about arguments passed from views.

no-reply:
> @jeremyf I didn't do a full dive on your code and I'm a bit fuzzy on this area of the codebase, but I would have thought that using `Hyrax::WorkRelation` would have scoped the results to only works such that you shouldn't need to send that extra where condition. Other parts of the code are using `Hyrax::WorkRelation` as well so I'm concerned that they're broken as well.

my understanding is that the count discrepancy is due to there being two different ways of scoping to "only works" in play: the `generic_type_sim` field vs. the model names + `curation_concerns` configuration.

`WorkRelation` *does* scope to works only, but it does so using a list of registered curation_concerns minus `FileSet` and `Collection`. the search results page limits based on `generic_type`. in theory, these should be equivalent in normal use, but i doubt there's a good reason not to just use `generic_type` in `WorkRelation`.

i'd like to be moving away from relying on `curation_concerns` registration for so many things---i think it makes sense for configuring routes, but there's a lot of complex code forcing this class list into a usable shape for differing purposes. `generic_type` seems better here, at least.

i don't think we should block this on a better resolution, though. i'd be in favor of approving this upon resolution of the discussion here: https://github.com/samvera/hyrax/pull/4494#discussion_r485822062

no-reply (approved):

PR #4495, opened 2020-09-02
Use response code from Hyrax::API

cjcolvar:
The response_type used for 500 errors is `:internal_error` which doesn't map to a valid rails response type (it should be `:internal_server_error`).  Instead of changing Hyrax::API just use the code set in Hyrax::API.  There is probably a better fix for this but this avoids errors for now.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -52,7 +52,7 @@ module Hyrax::Controller
   # render a json response for +response_type+
   def render_json_response(response_type: :success, message: nil, options: {})
     json_body = Hyrax::API.generate_response_body(response_type: response_type, message: message, options: options)
-    render json: json_body, status: response_type
+    render json: json_body, status: Hyrax::API::DEFAULT_RESPONSES[response_type][:code]
   end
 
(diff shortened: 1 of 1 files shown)


no-reply (approved):

app/controllers/concerns/hyrax/controller.rb
```diff
@@ -52,7 +52,7 @@ def set_locale
   # render a json response for +response_type+
   def render_json_response(response_type: :success, message: nil, options: {})
     json_body = Hyrax::API.generate_response_body(response_type: response_type, message: message, options: options)
-    render json: json_body, status: response_type
+    render json: json_body, status: Hyrax::API::DEFAULT_RESPONSES[response_type][:code]
```
jeremyf, line 55: This could easily itself raise an error (e.g. the `response_type` is missing in the map).  I think a `Hyrax::API.response_code_for(response_type)` method would help insulate from a failure that could be avoidable.

cjcolvar: I agree this could be cleaned up.  I created this WIP PR so I wouldn't forget as I was working through something else.  Failing test would also be good to add.  I'll come back to this eventually.

PR #4496, opened 2020-09-03
Helm charts for Hyrax and FCRepo

no-reply:
http://dassie.dld.library.ucsb.edu is deployed from this chart.

This [Helm][helm] chart provides configurable deployments for Hyrax applications
to [Kubernetes][k8s] clusters. It seeks to be a complete but flexible
production-ready setup for Hyrax applications. By default it deploys:

  - A Hyrax-based Rails application
  - Fedora Commons v4.7
  - Postgresql
  - Solr (in a cloud configuration, including Apache Zookeeper)
  - Redis

## A base Hyrax deployment

Because Hyrax is a [Rails Engine][engine]---not a stand-alone application---
deploying it requires us to have a specific application. This chart assumes that
the user has a container image based on `samveralabs/hyrax` (see:
[CONTAINERS.md][containers]) that includes their application. Point the chart at
your image by setting the `image.repository` and `image.tag` values.

By default, the chart deploys [images][dassie-images] for Hyrax's development
application, [`dassie`][dassie].

For application configuration, we take our queues from [12-factor][twelve]
methodology. Applications using environment variables to manage their
configuration can be easily reconfigured across different releases using this
chart; e.g. the same chart can be used to deploy sandbox, staging, and
production environments.

The chart populates the following environment variables:

|-------------------|--------------------------------|------------------------|
| Variable          | Description                    | Condition              |
|-------------------|--------------------------------|------------------------|
| DB_HOST           | Postgresql hostname            | `postgresql.enabled`   |
| DB_PORT           | Postgresql service port        | `postgresql.enabled`   |
| MEMCACHED_HOST    | Memcached host                 | `memcached.enabled`    |
| RACK_ENV          | app environment ('production') | n/a                    |
| RAILS_ENV         | app environment ('production') | n/a                    |
| REDIS_HOST        | Redis service host             | `redis.enabled`        |
| FCREPO_HOST       | Fedora Commons host            | `fcrepo.enabled`       |
| FCREPO_PORT       | Fedora Commons port            | `fcrepo.enabled`       |
| FCREPO_REST_PATH  | Fedora Commons REST endpoint   | `fcrepo.enabled`       |
| SOLR_HOST         | Solr service host              | `solr.enabled`         |
| SOLR_PORT         | Solr service port              | `solr.enabled`         |
| SOLR_URL          | Solr service full URL          | `solr.enabled`         |
|----------------- -|--------------------------------|------------------------|

## For DevOps:

For those interested in trying out or contributing to this Chart, it's helpful
to setup a simple cluster locally. Various projects exist to make this easy; we
recommend [`k3d`][k3d] or [minikube][minikube].

For example, with `k3d`:

```sh
k3d cluster create dev-cluster --api-port 6550 -p 80:80@loadbalancer --agents 3
```

[containers]: ../../CONTAINERS.md#hyrax-image
[dassie]: ../../.dassie/README.md
[dassie-image]: https://hub.docker.com/r/samveralabs/dassie
[engine]: https://guides.rubyonrails.org/engines.html
[helm]: https://helm.sh
[k3d]: https://k3d.io
[k8s]: https://kubernetes.io
[minikube]: https://minikube.sigs.k8s.io/docs/

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/fcrepo/templates/deployment.yaml
+++ b/chart/fcrepo/templates/deployment.yaml
@@ -34,7 +34,7 @@ spec:
             claimName: {{ template "fcrepo.fullname" . }}
       initContainers:
         - name: "remove-lost-found"
-          image: "busybox:1.25.0"
+          image: "busybox:1.32.0"
           command:
             - rm
             - -fr
@@ -49,7 +49,8 @@ spec:
           command:
             - sh
             - -c
-            - PGPASSWORD=$DATABASE_PASSWORD PGHOST=$DATABASE_HOST PGUSER=$DATABASE_USER psql -tc "SELECT 1 FROM pg_database WHERE datname = '$DATABASE_NAME'" | grep -q 1 || createdb -e -w $DATABASE_NAME
+            - export PGPASSWORD=$DATABASE_PASSWORD PGHOST=$DATABASE_HOST PGUSER=$DATABASE_USER
+            - psql -tc "SELECT 1 FROM pg_database WHERE datname = '$DATABASE_NAME'" | grep -q 1 || createdb -e -w $DATABASE_NAME
           envFrom:
             - configMapRef:
                 name: {{ include "fcrepo.fullname" . }}-env
--- a/chart/fcrepo/templates/deployment.yaml
+++ b/chart/fcrepo/templates/deployment.yaml
@@ -42,6 +42,19 @@ spec:
           volumeMounts:
             - name: data
               mountPath: /data
+        {{- if not .Values.postgresql.enabled }}
+        - name: "init-db"
+          image: {{ .Values.postgresql.image.repository | default "bitnami/postgresql" }}:{{ .Values.postgresql.image.tag | default "latest" }}
+          command:
+            - sh
+            - -c
+            - PGPASSWORD=$DATABASE_PASSWORD PGHOST=$DATABASE_HOST PGUSER=$DATABASE_USER psql -tc "SELECT 1 FROM pg_database WHERE datname = '$DATABASE_NAME'" | grep -q 1 || createdb -e -w $DATABASE_NAME
+          envFrom:
+            - configMapRef:
+                name: {{ include "fcrepo.fullname" . }}-env
+            - secretRef:
+                name: {{ include .Values.fcrepoSecretName . }}
+        {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           securityContext:
@@ -52,7 +65,11 @@ spec:
             - configMapRef:
                 name: {{ include "fcrepo.fullname" . }}-env
             - secretRef:
+                {{- if .Values.postgresql.enabled }}
                 name: {{ template "fcrepo.fullname" . }}
+                {{- else }}
+                name: {{ include .Values.fcrepoSecretName . }}
+                {{- end }}
           ports:
             - name: http
               containerPort: 8080
@@ -64,12 +81,12 @@ spec:
             httpGet:
               path: /
               port: http
-            initialDelaySeconds: {{ default .Values.healthCheck.initialDelay 15 }}
+            initialDelaySeconds: {{ default 15 .Values.healthCheck.initialDelay }}
           readinessProbe:
             httpGet:
               path: /
               port: http
-            initialDelaySeconds: {{ default .Values.healthCheck.initialDelay 15 }}
+            initialDelaySeconds: {{ default 15 .Values.healthCheck.initialDelay }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
       {{- with .Values.nodeSelector }}
--- a/chart/fcrepo/templates/deployment.yaml
+++ b/chart/fcrepo/templates/deployment.yaml
@@ -43,6 +43,7 @@ spec:
             - name: data
               mountPath: /data
         {{- if not .Values.postgresql.enabled }}
+        {{- if .Values.fcrepoSecretName }}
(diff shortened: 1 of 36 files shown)


chart/hyrax/values.yaml
```diff
@@ -0,0 +1,97 @@
...
+  hosts:
+    - host: hyrax.local
+      paths: []
+  tls: []
+
+resources: {}
+
+fcrepo:
+  enabled: true
+  fcrepoSecretName: "hyrax.fcrepo.fullname"
```
dunn, line 57: having fcrepo and hyrax share a postgres deployment seems unusual for default behavior, imo

orangewolf:
> at some point i imagine we'll want a separate charts repo for shared components like fcrepo (and possibly solr [helm/charts#22863 (comment)](https://github.com/helm/charts/pull/22863#issuecomment-686586313)); in that case would we want to move application charts like hyrax as well? how hard would that be?

👍 for moving the fcrepo out, but I think having the chart for the application live with the code is a very devops way of handling it. That way simple things like "where is this value used" pick up the infrastructure as code along with the application code.

chart/hyrax/values.yaml
no-reply: it seems a lot better than deploying two separate postgresql instances for one application stack?

are you thinking we'll want to scale these independently, or???

no-reply:
:100: x :+1:  on extracting the `fcrepo` chart into some community charts repository.

i mentioned to @orangewolf that my default mode has become putting everything in the same version control repository (owing to positive experiences with the Surfliner monorepo).

having lock-step updates across charts has been important up to this point, but maybe it's already a good time to look at extracting?

chart/hyrax/values.yaml
dunn: no it makes sense, but it took a while for me to see what was going on and it's not a pattern i've seen often

so maybe just a note in the readme or a comment here to clarify

orangewolf:
@no-reply you know I want to see it separate and published, but I'm wondering if this should go in as is and extract as a next action, just so this work isn't held up at all.

chart/hyrax/values.yaml
no-reply: i don't really love the secret injection thing. ideally the hyrax chart wouldn't know so much about fedora's `JAVA_OPTS`, but i ran out of energy for perfecting it.

no-reply:
i'm finding that the fedora db setup init container isn't as reliable as i'd hope.

sometimes it complains that `local user with ID 1001 does not exist`, and refuses to start fedora. i've tried reproducing by changing the script to `sleep 120`, shelling in and running manually, but can't get the negative behavior that way.

it might be worth considering whether a smaller container than `bitnami/postgresql` is called for for that purpose, anyway?

scottrigby:
Hello! 👋 

If you want to extract it, here are some steps to help (we'll be publishing a post on this soon, but in the meantime): https://github.com/scottrigby/prometheus-helm-charts/issues/3 (in an example from another migrated helm repo, @torstenwalter has combined those steps with a checklist for all the other steps needed in this nice summary issue: https://github.com/grafana/helm-charts/issues/3).

You can see this working demo of Helm GitHub Actions for CI (chart testing), and CD (for releasing chart version packages). The README has detailed steps to emulate them in your own Helm repo: https://github.com/helm/charts-repo-actions-demo

Also you'll want to add your helm repo to the CNCF Artifact Hub (the successor to Helm Hub) for discoverability 😄 

Hope that helps!

no-reply:
@scottrigby thanks for this input! it's very helpful indeed!

db/seeds.rb
```diff
@@ -1,3 +1,5 @@
+ActiveFedora.fedora.connection.send(:init_base_path)
```
jeremyf, line 1: Is this something that should be outside of this commit?

no-reply: i think it's tightly motivated by this environment---other existing environments seem to provide the base path for us(?).

i'm not sure moving it to a separate PR is helpful, since that will move it further away from its motivation. but maybe there are better approaches for the app to ensure its base path is present? maybe it's a whole different rake task that can be called in a more targeted way?

jeremyf: I'm satisfied with the answer.  Keep this as part of the commit.

orangewolf (approved):
Went end to end on deploying this both locally with k3d and to our k8s cloud. It works slick!

chart/fcrepo/templates/deployment.yaml
```diff
@@ -0,0 +1,109 @@
...
+            - name: data
+              mountPath: /data
+        {{- if not .Values.postgresql.enabled }}
+        {{- if .Values.fcrepoSecretName }}
+        - name: "init-db"
+          image: {{ .Values.postgresql.image.repository | default "bitnami/postgresql" }}:{{ .Values.postgresql.image.tag | default "latest" }}
+          command:
+            - sh
+            - -c
+            - export PGPASSWORD=$DATABASE_PASSWORD PGHOST=$DATABASE_HOST PGUSER=$DATABASE_USER
```
dunn, line 52: can these be set in an `environment` block below?

no-reply: probably. i'm going to suggest we merge and tweak from there.

this is now deploying successfully and i'm keen on being able to work both on improving the chart and on getting some fashion of CD in place. `nurax`'s current issues are worrisome since only a few people can work on its infra.

PR #4497, opened 2020-09-03
Lock unauthorized permissions from update

### LINKED ISSUE #3166: User can remove (repo admin or any user) permission in Sharing tab on work

jeremyf:
## Lock unauthorized permissions from update

70ff2ee968c0b4fe3cb9ae7e1df511a4ae6a5c21

Fixes #3166

- user is permitted to update any work permissions coming from
collections they manage
- user is permitted to update non-manager permissions from any
Collections
- user is permitted to update any non-collection permissions

Process used:
- find all of the other managers of collections a user cannot manage who
are not also managers in collections that the user CAN manage.

This gives us the permissions the user is not authorized to update.

EditPermissionsService embeds all of the above logic and uses it to
display all of the work or file set's permissions in either a fixed or
editable format.

The new shared partial results in a slight reformat of the display of
these permissions, as there were previously minor differences between
the two displays.

Additionally, a new partial was created for batch edit, as the above
rules cannot be enforced at a batch level. Dealing with batch editing
sharing issues is beyond the scope of this issue.

Restrict user from changing managers which are in both authorized &
unauthorized collections

## Fixing spec to account for upstream changes

47dc8a906cf74fb15e65f4ec81d2725a512aec5c

The previous commit (fde0590eeb56241cb8ed45ed24f50c5da3f127ef) was
worked on over a year ago, and recently re-opened and rebased against
the main branch.

In that time, aspects of the implementation have changed.

## Moving class documention above opening of class

034ac2c66fa209aa86c14dd888224d99a8d4c851

## Adjusting hash interaction and boolean test

88186e3710f09ececfae19172883c8290e137fdb

Given we are expecting a specific hash structure, favor `fetch` over
`[]` access.  This prevents a null-pointer.

Also, re-arrange the operands for the boolean test.  The
`permission_hash.fetch(:access)` is cheaper/quicker than the other
include test.  This is a small tweak that helps conserve a bit of
computation.

I believe it is important to capture some of @laritakr's intentions.
While this may not be the correct commit to associate, it's close to the
work.  Larita had the [following comment][1]:

> While working on this issue, it became apparent that this change as
> requested could be very confusing and misleading to a work's editor.
>
> A work can be in more than one collection, and if so, we wouldn't know
> which collection affected the permissions, because we don't track
> that. And permission changes on a collection don't carry through to a
> work, at least not at this point. So the tie between the collection and
> the work's permissions is tenuous at best.
>
> Admin set permissions always carry through to the work when it is
> created. Collection permissions only do if you create a new work from
> the button off of the collection's show view. Creating a work and adding
> it to a collection while you create it does not change permissions, nor
> does adding an existing work to a collection.
>
> So a work could have been created, and the depositor/editor could have
> given it permissions that they can now not change because it just
> happens to be the same permissions as the collection had. They also
> could add or attempt to add a permission which would appear to not be
> working because that permission is also on the collection or admin set.
>
> I would recommend showing all permissions but limiting the edit and
> delete access to them, and optionally showing something to identify the
> collection which is limiting the ability to change those permissions.
>
> So the changes I intend to implement to the work's permissions view are:
>
> - No one can view admin permissions on the work, because they should
>   always exist and we don't want to allow them to be edited
> - no one can edit a work's permissions which inherit from the manager
>   role of a collection/admin set unless they are also a manager of that
>   collection/admin set. these permissions will show on the work but not be
>   editable, and the collection which causes them to be frozen from change
>   will be identified
> - other permissions may be edited on the work by anyone with edit rights
>
> (note: An editor could still change these rights by removing the work
> from the collection, removing the rights, and adding it back to the
> collection, because the inheritance of permissions only happens during
> the work's creation.)

[1]:https://github.com/samvera/hyrax/issues/3166#issuecomment-420309316

## Favoring member_of_collection_ids over objects

a244fe224483209ac8e7a77ce27f223141308c0f

Instead of using the equivalent of `member_of_collections.map(&:id)`,
favor the already existing `member_of_collection_ids`

## Refactoring view to remove logic

2dd021ed8d24e0cfc41e7112f00cd91633d0b0c6

Prior to this commit, there was some not-immediate-obvious logic in the
view, namely around building the permission advice.

My goal with this refactor is to include a class that provides hooks for
documentation around this behavior.

## Attempting to fix a CircleCI spec error

53b257bde51de30fd5b035d4c6e9a6587a4b7f39

## Tweaking specs to for Valkyrie tests

3b094c79f6a1a955098cca8f819136d8180d3256

Ensuring that the two people are in the same permission template.

## Adding changes to reflect code review

529ede87811f25bf59884efcf7892de850c9f468

During [the review][1], @no_reply recommended a few tweaks to
documentation as well as removing some unneeded code.

[1]:https://github.com/samvera/hyrax/pull/4497

## Factoring view logic into builder method

d421dfda20d2ee3b7ec55f4243966a8c162c806b

The logic in the view is necessary, but we'd prefer object creation to
not have too much switch logic.

By moving it into the class, the hope is to clarify purpose and utility.

### DIFF
--- a/app/services/hyrax/edit_permissions_service.rb
+++ b/app/services/hyrax/edit_permissions_service.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 module Hyrax
+  # Encapsulates the logic to determine which object permissions may be edited by a given user
+  #  - user is permitted to update any work permissions coming ONLY from collections they manage
+  #  - user is not permitted to update a work permission if it comes from a collection they do not manage, even if also from a managed collection
+  #  - user is permitted to update only non-manager permissions from any Collections
+  #  - user is permitted to update any non-collection permissions
   class EditPermissionsService
-    # Encapsulates the logic to determine which object permissions may be edited by a given user
-    #  - user is permitted to update any work permissions coming ONLY from collections they manage
-    #  - user is not permitted to update a work permission if it comes from a collection they do not manage, even if also from a managed collection
-    #  - user is permitted to update only non-manager permissions from any Collections
-    #  - user is permitted to update any non-collection permissions
     attr_reader :depositor, :unauthorized_collection_managers
 
     # @param [Object] GenericWorkForm (if called for object) or GenericWork (if called for file set)
--- a/app/services/hyrax/edit_permissions_service.rb
+++ b/app/services/hyrax/edit_permissions_service.rb
@@ -1,0 +1,113 @@
+# frozen_string_literal: true
+module Hyrax
+  class EditPermissionsService
+    # Encapsulates the logic to determine which object permissions may be edited by a given user
+    #  - user is permitted to update any work permissions coming ONLY from collections they manage
+    #  - user is not permitted to update a work permission if it comes from a collection they do not manage, even if also from a managed collection
+    #  - user is permitted to update only non-manager permissions from any Collections
+    #  - user is permitted to update any non-collection permissions
+    attr_reader :depositor, :unauthorized_collection_managers
+
+    # @param [Object] GenericWorkForm (if called for object) or GenericWork (if called for file set)
+    # @param [Ability] user's current_ability
+    def initialize(object:, ability:)
+      @object = object
+      @ability = ability
+      @depositor = object.depositor
+      unauthorized = manager_permissions_to_block
+      @unauthorized_managers = unauthorized.unauthorized_managers
+      @unauthorized_collection_managers = unauthorized.unauthorized_collection_managers
+    end
+
+    # @param [Hash] one set of permission fields for object {:name, :access}
+    # @return [Boolean] true if user cannot edit the given permissions
+    def cannot_edit_permissions?(permission_hash)
+      @unauthorized_managers.include?(permission_hash[:name]) && permission_hash[:access] == "edit"
+    end
+
+    # @param [Hash] one set of permission fields for object {:name, :access}
+    # @return [Boolean] true if given permissions are one of fixed exclusions
+    def excluded_permission?(permission_hash)
+      exclude_from_display.include? permission_hash[:name].downcase
+    end
+
+    private
+
+    # Fixed set of users & groups to exclude from "editable" section of display
+    def exclude_from_display
+      [::Ability.public_group_name, ::Ability.registered_group_name, ::Ability.admin_group_name, @depositor]
+    end
+
+    BlockedPermissions = Struct.new(:unauthorized_managers, :unauthorized_collection_managers)
+
+    # find all of the other managers of collections which a user cannot manage
+    #
+    #   Process used:
+    #   - find all of the work's collections which a user can manage
+    #   - find all of the work's collections (of a type which shares permissions) that a user cannot manage
+    #   - find all of the managers of these collections the user cannot manage
+    #   This gives us the manager permissions the user is not authorized to update.
+    #
+    # @return [Struct] BlockedPermissions
+    #   - unauthorized_managers [Array] ids of managers of all collections
+    #   - unauthorized_collection_managers [Array hashes] manager ids & collection_ids [{:name, :id}]
+    def manager_permissions_to_block
+      unauthorized_managers = []
+      unauthorized_collection_managers = []
+      if object_unauthorized_collection_ids.any?
+        object_unauthorized_collection_ids.each do |id|
+          Hyrax::PermissionTemplate.find_by(source_id: id).access_grants.each do |grant|
+            if grant.access == "manage"
+              unauthorized_managers << grant.agent_id
+              unauthorized_collection_managers += Array.wrap({ name: grant.agent_id }.merge(id: id))
+            end
+          end
+        end
+      end
+      BlockedPermissions.new(unauthorized_managers, unauthorized_collection_managers)
+    end
+
+    # find all of the work's collections a user can manage
+    # @return [Array] of collection ids
+    def object_managed_collection_ids
+      @object_managed_collection_ids ||= object_member_of & managed_collection_ids
+    end
+
+    # find all of the work's collections a user cannot manage
+    # note: if the collection type doesn't include "sharing_applies_to_new_works", we don't limit access
+    # @return [Array] of collection ids with limited access
+    def object_unauthorized_collection_ids
+      @object_unauthorized_collection_ids ||= begin
+                                                limited_access = []
+                                                unauthorized_collection_ids = object_member_of - object_managed_collection_ids
+                                                if unauthorized_collection_ids.any?
+                                                  unauthorized_collection_ids.each do |id|
+                                                    collection = ActiveFedora::Base.find(id)
+                                                    limited_access << id if (collection.instance_of? AdminSet) || collection.share_applies_to_new_works?
+                                                  end
+                                                end
+                                                limited_access
+                                              end
+    end
+
+    # find all of the collection ids an object is a member of
+    # @return [Array] array of collection ids
+    def object_member_of
+      @object_member_of ||= begin
+                              belongs_to = []
+                              # get all of work's collection ids from the form
+                              @object.member_of_collections.each do |collection|
+                                belongs_to << collection.id
+                              end
+                              belongs_to << @object.admin_set_id unless @object.admin_set_id.empty?
+                              belongs_to
+                            end
+    end
+
+    # The list of all collections this user has manage rights on
+    # @return [Array] array of all collection ids that user can manage
+    def managed_collection_ids
+      Hyrax::Collections::PermissionsService.source_ids_for_manage(ability: @ability)
+    end
+  end
+end
--- a/app/services/hyrax/edit_permissions_service.rb
+++ b/app/services/hyrax/edit_permissions_service.rb
@@ -8,8 +8,8 @@ module Hyrax
   class EditPermissionsService
     attr_reader :depositor, :unauthorized_collection_managers
 
-    # @param [Object] GenericWorkForm (if called for object) or GenericWork (if called for file set)
-    # @param [Ability] user's current_ability
+    # @param object [#depositor, #admin_set_id, #member_of_collection_ids] GenericWorkForm (if called for object) or GenericWork (if called for file set)
+    # @param ability [Ability] user's current_ability
     def initialize(object:, ability:)
       @object = object
       @ability = ability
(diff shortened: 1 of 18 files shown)


jeremyf:
Ping

app/views/hyrax/base/_currently_shared.html.erb
```diff
@@ -0,0 +1,66 @@
...
+#   - form object.object.model = GenericWork
+#   - use the work itself
+#   For file_sets:
+#   - form object.object.class = FileSet
+#   - use work the file_set is in
+#   No other object types are supported by this view. %>
+<% if f.object.respond_to?(:model) && f.object.model.work?
+  object_acting_upon = f.object
+elsif f.object.file_set?
+  object_acting_upon = f.object.in_works.first
```
jeremyf, line 14: @laritakr This is a serious shot in the dark, but what should the fallback be? (If any?)

I'm okay with a `raise`

laritakr: I don't really remember, but I believe this partial is only used for works and file sets, so anything else going through here would be unexpected behavior. I say raise an exception.

jeremyf:
@no-reply I've completed my review; and based on conversations with @jlhardes the addressed issue (#3166) remains a problem that we should solve.

jeremyf:
Keeping this from aging out of the system

jeremyf:
.

jeremyf:
.

dlpierce (commented):
This new EditPermissionsService seems to only be used to render the form. Should it also be used to validate a request to change permissions on the server side?

jeremyf:
@dlpierce - Ideally, yes.  As the one stewarding this from an old pull request I'd prefer to write up an issue about adding this logic into the form validation as well.

dlpierce (approved):

jeremyf:
Note: I believe the failing Valkyrie test that I'm seeing is a failure in our test suite to account for the proxy deposit AND the inter-relation of the new `app/services/hyrax/edit_permissions_service.rb`

I think I'm going to defer on this until I can get time with @no-reply

jeremyf:
I am able to duplicate the failing Valkyrie on my local machine.  With `docker-compose up` running, I ran the following: `docker-compose exec -w /app/samvera/hyrax-engine app sh -c "HYRAX_VALKYRIE=true bundle exec rspec spec/features/create_work_spec.rb"` (note the `HYRAX_VALKYRIE=true`).  The specs fail.

```console
Failures:

  1) Creating a new Work when the user is a proxy allows on-behalf-of deposit
     Failure/Error: expect(page).to have_content "My Test Work"
       expected to find text "My Test Work" in "Hyrax\nHyrax\nSwitch language\nEnglish\nView\n user4@example.com\nprofile\nuser4@example.com\nACTIVITY\nYour activity\nREPOSITORY CONTENTS\nCollections\nWorks\nHomeDashboardWorks\nWorks\nAdd new work\nManaged Works\nYour Works\n0 works you own in the repository\nSearch Hyrax\nGo\n\n\n\n\nWorks listing\nListing of items you have deposited in Hyrax\nSelect all files to be added to a collection or edited\nSelect\nSelect to access selection option\nTitle Date Added Highlighted Visibility of Work Actions"
     # ./spec/features/create_work_spec.rb:102:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:259:in `block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/webmock-3.11.2/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
```

I'm digging into these failures.

app/views/hyrax/base/_currently_shared.html.erb
```diff
@@ -0,0 +1,63 @@
...
+# form object.class = SimpleForm::FormBuilder
+#   For works (i.e. GenericWork):
+#   - form object.object = Hyrax::GenericWorkForm
+#   - form object.object.model = GenericWork
+#   - use the work itself
+#   For file_sets:
+#   - form object.object.class = FileSet
+#   - use work the file_set is in
+#   No other object types are supported by this view. %>
+<% if f.object.respond_to?(:model) && f.object.model.work?
```
no-reply, line 11: a good follow up might be to refactor this into a helper? it's only called once, and it's a bit worrying to have this branching logic in a view.

or maybe it's not a helper and could be wrapped into the service object initializer.

app/services/hyrax/edit_permissions_service.rb
```diff
@@ -0,0 +1,178 @@
...
+    #   - find all of the managers of these collections the user cannot manage
+    #   This gives us the manager permissions the user is not authorized to update.
+    #
+    # @return [Struct] BlockedPermissions
+    #   - unauthorized_managers [Array] ids of managers of all collections
+    #   - unauthorized_collection_managers [Array hashes] manager ids & collection_ids [{:name, :id}]
+    def manager_permissions_to_block
+      unauthorized_managers = []
+      unauthorized_collection_managers = []
+      if object_unauthorized_collection_ids.any?
```
no-reply, line 122: can this check be dropped? `object_unauthorized_collection_ids.each` will be a no-op if it's empty, yes?

app/services/hyrax/edit_permissions_service.rb
```diff
@@ -0,0 +1,178 @@
...
+    # find all of the work's collections a user cannot manage
+    # note: if the collection type doesn't include "sharing_applies_to_new_works", we don't limit access
+    # @return [Array] of collection ids with limited access
+    def object_unauthorized_collection_ids
+      @object_unauthorized_collection_ids ||= begin
+                                                limited_access = []
+                                                unauthorized_collection_ids = object_member_of - object_managed_collection_ids
+                                                if unauthorized_collection_ids.any?
+                                                  unauthorized_collection_ids.each do |id|
+                                                    # TODO: Can we instead use a SOLR query?  This seems to be somewhat expensive.  However, as this is
```
no-reply, line 150: +1 to punting this now. we'll need to refactor completely for valkyrie anyway.

app/services/hyrax/edit_permissions_service.rb
```diff
@@ -0,0 +1,178 @@
...
+    end
+
+    # find all of the collection ids an object is a member of
+    # @return [Array] array of collection ids
+    def object_member_of
+      @object_member_of ||= begin
+                              belongs_to = []
+                              # get all of work's collection ids from the form
+                              belongs_to += @object.member_of_collection_ids
+                              belongs_to << @object.admin_set_id unless @object.admin_set_id.empty?
```
no-reply, line 167: is this `#empty?` catching a `""` string as id?

app/services/hyrax/edit_permissions_service.rb
```diff
@@ -0,0 +1,178 @@
...
+module Hyrax
+  # Encapsulates the logic to determine which object permissions may be edited by a given user
+  #  - user is permitted to update any work permissions coming ONLY from collections they manage
+  #  - user is not permitted to update a work permission if it comes from a collection they do not manage, even if also from a managed collection
+  #  - user is permitted to update only non-manager permissions from any Collections
+  #  - user is permitted to update any non-collection permissions
+  class EditPermissionsService
+    attr_reader :depositor, :unauthorized_collection_managers
+
+    # @param [Object] GenericWorkForm (if called for object) or GenericWork (if called for file set)
```
no-reply, line 11: is the interface `@param object [#depositor, #admin_set_id, #member_of_collection_ids]`?

app/services/hyrax/edit_permissions_service.rb
```diff
@@ -0,0 +1,178 @@
...
+    # @param [Ability] user's current_ability
```
no-reply, line 12:
```suggestion
    # @param [Ability] ability  user's current_ability
```

app/services/hyrax/edit_permissions_service.rb
```diff
@@ -0,0 +1,178 @@
...
+      @depositor = object.depositor
+      unauthorized = manager_permissions_to_block
+      @unauthorized_managers = unauthorized.unauthorized_managers
+      @unauthorized_collection_managers = unauthorized.unauthorized_collection_managers
+    end
+
+    # @api private
+    # @todo refactor this code to use "can_edit?"; Thinking in negations can be challenging.
+    #
+    # @param [Hash] one set of permission fields for object {:name, :access}
```
no-reply, line 25:
```suggestion
    # @param [Hash] permission_hash  one set of permission fields for object {:name, :access}
```

app/services/hyrax/edit_permissions_service.rb
```diff
@@ -0,0 +1,178 @@
...
+    # @return [Boolean] true if user cannot edit the given permissions
+    def cannot_edit_permissions?(permission_hash)
+      permission_hash.fetch(:access) == "edit" && @unauthorized_managers.include?(permission_hash.fetch(:name))
+    end
+
+    # @api private
+    #
+    # @param [Hash] one set of permission fields for object {:name, :access}
```
no-reply, line 33:
```suggestion
    # @param [Hash] permission_hash  one set of permission fields for object {:name, :access}
```

no-reply (dismissed):
a few small suggestions (to fix doc display, mainly) and some maybe follow-up work to discuss.

this looks like it fixes the issue and is reasonably concise, so let's do one last pass and merge. 

thanks to everyone (especially @laritakr!) who has worked on this. sorry it took so long to get through!

app/services/hyrax/edit_permissions_service.rb
jeremyf: I don't know; I suspect it's a `nil` test. I'm going to switch it to `if @object.admin_set_id.present?` as I always prefer to remove double negations (unless empty feels like I need one more hoop than if present).

app/views/hyrax/base/_currently_shared.html.erb
jeremyf: How about `permission_service = Hyrax::EditPermissionsService.build_service_from(form: f, ability: current_ability)`

dlpierce (approved):

PR #4498, opened 2020-09-04
Apply skip_to_content javascript across app not just client pages

### LINKED ISSUE #3774: Skip to Content link doesn't always set keyboard focus

bkeese:
Fixes #3774 

Skip_to_content javascript code was only used on the client-facing page layouts, not dashboard pages.
This PR applies the javascript to all pages.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax.js
+++ b/app/assets/javascripts/hyrax.js
@@ -105,6 +105,7 @@
 //= require hyrax/turbolinks_events
 //= require hyrax/i18n_helper
 //= require hyrax/collapse
+//= require hyrax/skip_to_content
 
 // this needs to be after batch_select so that the form ids get setup correctly
 //= require hyrax/batch_edit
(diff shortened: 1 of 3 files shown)


no-reply (approved):
fantastic.

let's backport to 2.x?

PR #4502, opened 2020-09-08
Ubiquity auto expiry jobs

BertZZ:
Jobs to find all assets that have expired Embargoes and Leases and remove them automatically


app/jobs/embargo_expiry_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class EmbargoExpiryJob < Hyrax::ApplicationJob
+  def perform
+    records_with_expired_embargos.each do |record|
+      ExpireEmbargoJob.perform_later(record)
+    end
+  end
+
+  def records_with_expired_embargos
```
cjcolvar, line 9: I wonder if this is really necessary and if the presenters returned by `Hyrax::EmbargoService` can be passed directly into `perform_later`.  Could you try that and see if it works?

BertZZ: I believe that I did and it did not but just for you Chris I will try again

cjcolvar: If it doesn't automagically work, then it might be necessary to add a `to_global_id` method to the presenter.  I think that could just be delegated in [ModelProxy](https://github.com/samvera/hyrax/blob/08ffa5aa73ee81b2c812d81635bfa328ca565f53/app/presenters/hyrax/model_proxy.rb#L7).

spec/jobs/embargo_expiry_job_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it 'returns all records with expired embargos' do
+      records = described_class.new.records_with_expired_embargos
+      expect(records).to include work_with_expired_embargo, file_set_with_expired_embargo
+      expect(records).not_to include embargoed_work
+    end
+  end
+
+  describe '#perform' do
+    it "Enqueues an Expire Embargo job for each expired record" do
+      expect { described_class.perform_now }.to change(ActiveJob::Base.queue_adapter.enqueued_jobs, :size).by(described_class.new.records_with_expired_embargos.count)
```
cjcolvar, line 31: I think it may be nicer to use the `have_enqueued_job` matcher instead of `change`.
https://relishapp.com/rspec/rspec-rails/docs/matchers/have-enqueued-job-matcher

spec/jobs/expire_embargo_job_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      described_class.new.perform(work_with_expired_embargo)
+      described_class.new.perform(embargoed_work)
+      described_class.new.perform(file_set_with_expired_embargo)
+      embargoed_work.reload
+      file_set_with_expired_embargo.reload
+      work_with_expired_embargo.reload
+      expect(work_with_expired_embargo.visibility).to eq('open')
+      expect(file_set_with_expired_embargo.visibility).to eq('open')
+      expect(embargoed_work.visibility).to eq('restricted')
+    end
```
cjcolvar, line 32: I think this would be clearer as three separate tests.

app/jobs/embargo_expiry_job.rb
no-reply: it does seem like we need a better pattern here. the map into `AF::Base.find` on L11 seems like it could be very expensive.

i'm not sure adding `to_global_id` to the presenter is great, though. i think `GlobalId` is really meant to uniquely identify the model, and the presenter is a different model from the `AF::Base` one, right?

a fast, code-smelly fix might be 

```ruby
assets = Hyrax::EmbargoService.assets_with_expired_embargoes
assets.map { |presenter| presenter.solr_document.to_model }
```

there's also a valkyrie powered, [`Hyrax::EmbargoManager`](https://github.com/samvera/hyrax/blob/130c4e600318a9194725b35dd0fd6e19e5108dd9/app/services/hyrax/embargo_manager.rb) which this will eventually (soon) need to be rewritten to use. if you're up for it, it would be worth reworking this now rather than later.

app/jobs/embargo_expiry_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class EmbargoExpiryJob < Hyrax::ApplicationJob
```
no-reply, line 2: how comfortable do we feel with jobs that enqueue jobs?

could this be squashed into just one job that loops over the `records_with_expired_embargos` and lifts them directly? what's the advantage of the nested queuing?

spec/jobs/embargo_expiry_job_spec.rb
cjcolvar: Nevermind.  I don't see any examples of this matcher in the Hyrax codebase yet.  Maybe expecting calls to `perform_later` then: https://github.com/samvera/hyrax/blob/130c4e600318a9194725b35dd0fd6e19e5108dd9/spec/config/hyrax_events_spec.rb#L9

app/jobs/embargo_expiry_job.rb
cjcolvar: I suggested this approach.  I was thinking it was easier to reason about and write tests for.  Also the single object nested job would be available to be reused in other contexts.  I know that Hyrax has some jobs that enqueue other jobs in the ingest process that are linear steps, but I think this is different.  One job gathers the work to be done and the nested jobs do the work in parallel in an atomic fashion.  There might be more overhead to this but if anything fails than it might be easier to find the cause and not have it affect the rest of the work.  I'm not wed to the approach but I thought it made sense.

Can you speak more to your hesitations about nested queuing?  I'm really curious.

no-reply: > Can you speak more to your hesitations about nested queuing? I'm really curious.

i have the vague recollection that it's considered bad form form `ActiveJob` to do nested enqueues, but i can't find any actual source for that.

my concrete concern here is that it seems much harder to reason about retries. suppose `EmbargoExpiryJob` fails half-way through, after enqueuing `ExpireEmbargoJob` ~100 times, and goes back into the queue for retry. will it end up enqueuing those same ~100 jobs again (that seems like the likely outcome to me, at least for sidekiq). will the duplicate jobs fail and go into the retry queue themselves? will the reason for that be legible to someone reading worker logs?

no-reply: i do like that there's a single-object embargo/lease lift job, in this. i think it's worth keeping that utility even if the bulk expiry jobs don't reuse it.

cjcolvar: >my concrete concern here is that it seems much harder to reason about retries. suppose EmbargoExpiryJob fails half-way through, after enqueuing ExpireEmbargoJob ~100 times, and goes back into the queue for retry. will it end up enqueuing those same ~100 jobs again (that seems like the likely outcome to me, at least for sidekiq). will the duplicate jobs fail and go into the retry queue themselves? will the reason for that be legible to someone reading worker logs?

That's a really good point.  Is that worse than hitting an error with a specific object and having the rest behind it in line getting stuck?  What about having the atomic jobs check that they are still under embargo/lease before calling the actor?

app/jobs/embargo_expiry_job.rb
BertZZ: @no-reply 

> there's also a valkyrie powered, Hyrax::EmbargoManager which this will eventually (soon) need to be rewritten to use.

When you say this do you mean this method or other parts of this workflow, because I don't see anything in `Hyrax::EmbargoManager` that gives me a list of all assets with expired embargoes, but I can see how we could use it to rewrite `ExpireEmbargoJob` for example

no-reply: yes. sorry if my comment was misleading. the `EmbargoManager` handles the embargo workflow for individual objects. i think the `Hyrax::EmbargoService` is still the best way to query embargo status in bulk.

app/jobs/embargo_expiry_job.rb
no-reply: > That's a really good point. Is that worse than hitting an error with a specific object and having the rest behind it in line getting stuck? 

there may be complications, but i think with a single job you have the advantage that a failure still results in some legitimately partially-complete job:

  1. the bulk clearing job starts;
  1. the embargo service queries solr for a list of `n` active expired embargoes;
  1. `m` (zero or more) objects process, lifting their embargo;
  1. **failure**; re-enqueue; back-off wait
  1. re-enqueued bulk job starts;
  1.  the embargo service queries solr for a list of `n - m` active expired embargoes;
  1. ...

at least on an off-hand analysis, a job like this can be safely re-enqueued indefinitely. if it's consistently failing on the same object, it should be possible to debug that, take manual action to fix the cause, and let the job complete on a future retry. 

> What about having the atomic jobs check that they are still under embargo/lease before calling the actor?

i think this would help. if multiple jobs for the same object go into the queue, at least in this case i'd expect them all to eventually clear with successful statuses. it seems like a good idea to make sure the `Expire*Job`s behave reasonably if they are run with an object that doesn't have an embargo, or doesn't have one that's expired.

especially with the added queue wait it seems possible that the object's embargo/lease status would change between the query and the job run.

my recollection is that `EmbargoManager` is more careful about exposing and respecting the current status than the actor implementation.

app/jobs/embargo_expiry_job.rb
```diff
@@ -0,0 +1,13 @@
...
+  def perform
+    records_with_expired_embargos.each do |id|
+      work = ActiveFedora::Base.find(id)
+      Hyrax::Actors::EmbargoActor.new(work).destroy
+    end
+  end
+
+  def records_with_expired_embargos
```
no-reply, line 10:
```suggestion
  ##
  # @return [Enumerator<String>] ids for all the objects that have expired active embargoes
  def records_with_expired_embargos
```

app/jobs/lease_expiry_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class LeaseExpiryJob < Hyrax::ApplicationJob
+  def perform
+    records_with_expired_leases.each do |id|
+      work = ActiveFedora::Base.find(id)
+      Hyrax::Actors::LeaseActor.new(work).destroy
+    end
+  end
+
+  def records_with_expired_leases
```
no-reply, line 10:
```suggestion
  ##
  # @return [Enumerator<String>] ids for all the objects that have expired active leases
  def records_with_expired_leases
```

no-reply (dismissed):
looks great. thanks for all the work you did on this!

PR #4507, opened 2020-09-09
allow resetting the configured valkyrie index adapter

no-reply:
there was a bug causing `Hyrax.config.index_adapter=` to be a no-op if the index
adapter is already set. this patches it, allowing applications to change their
configured indexing behavior at runtime.

@samvera/hyrax-code-reviewers


elrayle (approved):
Look good.  Thanks!

PR #4508, opened 2020-09-09
move indexing adapter registration to high-precedence

no-reply:
applications will probably want to do indexer setup in
initializers. registration of Hyrax's shipped adapters needs to be in place
before they do this. previously, this was the case only when the application
does setup in a letter after `'i'` (e.g. not `hyrax.rb`).

moving this up in precedence will make things more predictable for apps.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -21,6 +21,9 @@ require 'hyrax/inflections'
 require 'hyrax/name'
 require 'hyrax/valkyrie_can_can_adapter'
 require 'kaminari_route_prefix'
+require 'valkyrie/indexing_adapter'
+require 'valkyrie/indexing/solr/indexing_adapter'
+require 'valkyrie/indexing/null_indexing_adapter'
 
 module Hyrax
   extend ActiveSupport::Autoload
--- a/config/initializers/indexing_adapter_initializer.rb
+++ b/config/initializers/indexing_adapter_initializer.rb
@@ -1,14 +1,9 @@
 # frozen_string_literal: true
-require 'valkyrie/indexing_adapter'
-require 'valkyrie/indexing/solr/indexing_adapter'
-require 'valkyrie/indexing/null_indexing_adapter'
 
-Rails.application.config.to_prepare do
-  Valkyrie::IndexingAdapter.register(
-    Valkyrie::Indexing::Solr::IndexingAdapter.new,
-    :solr_index
-  )
-  Valkyrie::IndexingAdapter.register(
-    Valkyrie::Indexing::NullIndexingAdapter.new, :null_index
-  )
-end
+Valkyrie::IndexingAdapter.register(
+  Valkyrie::Indexing::Solr::IndexingAdapter.new,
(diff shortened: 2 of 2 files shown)


lib/hyrax.rb
```diff
@@ -21,6 +21,9 @@
 require 'hyrax/name'
 require 'hyrax/valkyrie_can_can_adapter'
 require 'kaminari_route_prefix'
+require 'valkyrie/indexing_adapter'
```
elrayle, line 24: I don't know that these requires are necessary.  I was able to use the solr_adapter with just a change to indexing_adapter_initializer.rb.

config/initializers/01_indexing_adapter_initializer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
elrayle, line 1: The rename to `01_...` is not needed.  All gem initializers are run before running application initializers.  The problem was with the `to_prepare` in the old version of this initializer.  It was causing the block within the `to_prepare` to run after all initializers completed.

I recommend keeping the original name and simply move the code in the `to_prepare` block outside.

elrayle (changes requested):
See comments.

elrayle:
I wonder if the `hyrax.rb` initializer in generators should be updated to include the adapter definition.  Perhaps [near where `query_index_from_valkyrie`](https://github.com/samvera/hyrax/blob/master/lib/generators/hyrax/templates/config/initializers/hyrax.rb#L241-244) is defined since the relate.

```
  ## Configure index adapter for Valkyrie::Resources to use solr readonly indexer
  # config.index_adapter = :solr_index
```

lib/hyrax.rb
no-reply: these are being moved from the initializer, here.

are you saying that you were able to drop them from initializer without replacing them? i wonder how this code is being loaded?

elrayle: No I did not remove them. If this was a move, then I was likely picking them up in their original location.

elrayle:
@no-reply I checked in with @hackmastera who wrote this code in Hyrax that uses the `to_prepare` block.  She indicated she was following the pattern from the [Valkyrie README](https://github.com/samvera/valkyrie#configuration).  Checking in with @tpendragon who wrote the Valkyrie code, his response was...

> I think I put it in there so it'd support the environment reloading in dev?

no-reply:
> I think I put it in there so it'd support the environment reloading in dev?

seems like that's not as useful in the engine context.

elrayle:
Agreed.  It is more useful for applications to be able to set up use of the indexer in initializers.  I'd like to see the code moved out of the `to_prepare` block.

elrayle (dismissed):
Looks great, thanks!

elrayle (approved):
All is well.  The problem was that this branch is out of date with master.  I rebased the branch locally with master and my test requiring this update passed.

PR #4509, opened 2020-09-09
ValkyrieIndexer.for - use `class.name` instead of `class` to avoid issues with wings generated classes

elrayle:
### Descriptive summary

Getting the indexer for a resource depends on naming convention by appending Indexer to the end of the classname (e.g. MyWork has indexer MyWorkIndexer).  Classname before using wings will match the resource class name (e.g. w = MyWork.new; puts w.classname # "MyWork").  Classname after persisting with wings adapter has a Wings generated class and classname (e.g. w2 = Hyrax.persister.save(resource: w); puts w2.classname # "Wings(MyWork)").  If it cannot find a matching indexer, it defaults to Valkyrie::Indexer

### Solution

Use class.name instead of class.  It returns the correct string for creating the indexer from the resource.

```
>> w = Hyrax::Work.new
#<Hyrax::Work id=nil internal_resource="Hyrax::Work" ...>
>> puts w.class
Hyrax::Work
nil
>> w.class.name
Hyrax::Work

>> w2 = Hyrax.persister.save(resource: w)
#<Hyrax::Work id=#<Valkyrie::ID:0x00007fb23ed54ab0 @id="1831cj92z"> internal_resource="#<Class:0x00007fb23ecf7428>" ...>
>> puts w2.class
Wings(Hyrax::Work)
nil
>> w2.class.name
Hyrax::Work
```

NOTE: The spec that was added fails with the original code and passes with the updated code.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/indexers/hyrax/valkyrie_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_indexer.rb
@@ -56,17 +56,6 @@ module Hyrax
     attr_reader :resource
 
     ##
-    # @api public
-    # @param [Valkyrie::Resource] resource
-    #
-    # @return [#to_solr]
-    def self.for(resource:)
-      indexer_class = "#{resource.class.name}Indexer".safe_constantize || ValkyrieIndexer
-
-      indexer_class.new(resource: resource)
-    end
-
-    ##
     # @api private
     # @param [Valkyrie::Resource] resource
     def initialize(resource:)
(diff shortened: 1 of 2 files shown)


no-reply (dismissed):
great!

:tada: thanks for surfacing this! thanks for the well formed PR! thanks for the nice test! :tada:

no-reply:
well, this is working a bit too well. `ResourceIndexer` and `Resource` are named so nicely that they follow the convention accidentally.

at least it's well tested! would have been a nasty bug to find later.

elrayle:
Made a second commit to address the problem with Hyrax::Resource trying to use Hyrax::ResourceIndexer.

### Description

Using naming convention, Hyrax::Resource would be indexed by Hyrax::ResourceIndexer.  This class exists but should not be used to index instances of Hyrax::Resource.  Instead the default indexer Valkyrie::Indexer should be used.

The second commit, 'handle special case for Hyrax::Resource',  addresses test failures related to trying to use Hyrax::ResourceIndexer for Hyrax::Resource instances.

app/indexers/hyrax/valkyrie_indexer.rb
```diff
@@ -83,5 +72,30 @@ def to_solr
...
+      #       class is expected to be the class name appended with 'Indexer'.  It will return default ValkyrieIndexer
+      #       if an indexer following the naming convention is not found or the resource is the reserved Hyrax::Resource
+      #       which should use ValkyrieIndexer instead of the base implementation in Hyrax::ResourceIndexer.
+      # @example
+      #     resource class:  Book
+      #     indexer class:   BookIndexer
+      # @return [Valkyrie::Indexer] an instance of ValkyrieIndexer or an inherited class based on naming convention
+      #
+      def for(resource:)
+        indexer_class = resource.class.name == 'Hyrax::Resource' ? ValkyrieIndexer : indexer_from_classname(resource)
```
no-reply, line 90: it's not my favorite situation, but i think it's where we're at. 

TY!

no-reply (approved):

app/indexers/hyrax/valkyrie_indexer.rb
elrayle: I didn't think it would be your favorite.  Always better to avoid hardcoding exceptions to the rule.  I tried to compensate by adding clear method documentation.

PR #4510, opened 2020-09-10
Re-rendering Work params for failed submit

jeremyf:
Prior to this commit, when I would submit an invalid work create form,
the re-rendered form would have blank input values.  With this change
all non-file attributes are re-loaded with the submitted values.

Please note, this needs a bit more time to put full polish on, but I
want to make sure its a path worth pursuing.  If this is a reasonable
path, I'm happy to add the additional specs and inline comments to get
this working.

I'm presently dissatisfied with the changes I needed to make to the case
statement, and would appreciate guidance on how to correct that.

Related to #3978

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -53,12 +53,17 @@ module Hyrax
     end
 
     def create
+      # Caching the original input params in case the form is not valid
+      original_input_params_for_form = params[hash_key_for_curation_concern].deep_dup
       if create_work
         after_create_response
       else
         respond_to do |wants|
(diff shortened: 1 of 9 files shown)


jeremyf:
I'll fix the failures once we have consensus that this is a reasonable/useful approach.

no-reply:
thanks.

the basic shape of this seems okay to me. i'll look closer at the form wrapper implementation shortly.

jeremyf:
@no-reply I've removed the WIP prefix.  This is ready for review and consideration.

no-reply (approved):

no-reply:
this is a candidate for backport to 2.x-stable

PR #4519, opened 2020-09-15
Adding AdminSet#update_index to default admin set

jeremyf:
I have found that when I come back to a development environment, that I
may have an AdminSet in Fedora, but it is not indexed in Solr.  This
remediates that situation by ensuring we have an indexed AdminSet.

@samvera/hyrax-code-reviewers


no-reply:
this seems fine, but have you tried using `seeds.rb` instead? https://github.com/samvera/hyrax/blob/master/db/seeds.rb

no-reply (approved):

jeremyf:
I haven't tried seeds.  I think the issue may still exist in seeds as well (for development edge cases)

PR #4522, opened 2020-09-16
Synchronizing Spec::SpyListener and Publisher

jeremyf:
I added documentation to explain how to add a named listener, instead of
an anonymous block listener.  I think this is useful, as our SpyListener
used a named listener, but that mechanism was not documented in the
Hyrax::Publisher.

In addition to the documentation, I've ensured that the SpyListener's
methods are synchronized with the Hyrax::Publisher registered events.
This synchronization also provides a bit of a test of the implementation
details of Dry::Events; Based on Dry's long-standing volatilitaty of
pre-1.0 gems, I think its a good plan to test a bit of the edges of
Dry::Events.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/specs/spy_listener.rb
+++ b/lib/hyrax/specs/spy_listener.rb
@@ -2,37 +2,44 @@
 
 module Hyrax
   module Specs
+    # A spec support class for assisting in spec testing of the
+    # Hyrax::Publisher pub/sub behavior.
+    #
+    # For each registered event, there are two corresponding instance methods
+    #
+    # 1. `#on_<registered_event>`
+    # 2. `#<registered_event>`
+    #
+    # Then, for any spec you want to make sure Pub/Sub events fire,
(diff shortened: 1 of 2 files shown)


lib/hyrax/specs/spy_listener.rb
```diff
@@ -2,37 +2,44 @@
...
-        @file_set_attached = event
-      end
-
-      def on_file_set_url_imported(event)
-        @file_set_url_imported = event
+      Hyrax::Publisher.events.each_value do |registered_event|
+        listener_method = registered_event.listener_method
+        attr_name = registered_event.listener_method.to_s.sub(/^on_/, '')
+        attr_reader attr_name
+        define_method listener_method do |published_event|
```
no-reply, line 40: thanks for this refactor!

no-reply (approved):

PR #4525, opened 2020-09-16
Bumping regen to fix rubocop-ast caching issue

jeremyf:
Prior to this commit, the CI environment of Hyrax went wonky.  This, we
hope, is addressed in our Bixby implementation (see
https://github.com/samvera-labs/bixby/pull/47).  However, we need to
invalidate the cache.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.regen
+++ b/.regen
@@ -1,1 +1,1 @@
-29omg
+30zoinks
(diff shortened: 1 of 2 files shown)


.regen
```diff
@@ -1 +1 @@
-29omg
+30zoinks
```
no-reply, line 1: :clap:

no-reply (approved):

PR #4527, opened 2020-09-17
Refactoring DefaultAdminSetActor

jeremyf:
## Minor refactoring of DefaultAdminSetActor

a6336febd7b8ae801ad4437cd4779a9547e43734

I'm beginning to poke at the logic in the actor stack.  On my first
pass, I'm looking for candidates that I could extract a service class.

This commit compresses some logic, leaning on Rails `find_or_create_by!`
instead of two separate methods and `||` logic.

In addition, this commit fixes a typo in the specs, and moves from a
`create(:user)` to `build(:user)`.  Anecdotally, the move from create to
build shaved 0.5 seconds from these specs, or dropped the test time to
75% of the original.

As my comments indicate, I'm currently unpacking the
`ensure_admin_set_attribute!` method.

## Extracting "Service" from ActorStack object

ae87c08076d965af3f9b682acccbd38808c50034

The extracted code is something easily repurposible, and could be easily
repurposed in the Hyrax::Transactions.

## Switching spec to reflect behavior

f219ee3491dc7faee8c316666d1645db80b519a2

Prior to this change, when running this spec in isolation, it does not
succeed.  I believe at issue is the fact that specs do not reload the
object.

```console
bundle exec rspec
./spec/hyrax/transactions/steps/set_default_admin_set_spec.rb:46

Failures:

  1) Hyrax::Transactions::Steps::SetDefaultAdminSet#call with an active fedora object sets the default admin_set
     Failure/Error:
       expect { step.call(work) }
         .to change { work.admin_set&.id }
         .from(nil)
         .to(admin_set_id)

       expected `work.admin_set&.id` to have changed from nil to
       "admin_set/default", but did not change
```

I verified that the above spec fails for SHA 08ffa5aa7 (a commit from a Sept 4, 2020).

### DIFF
--- a/app/actors/hyrax/actors/default_admin_set_actor.rb
+++ b/app/actors/hyrax/actors/default_admin_set_actor.rb
@@ -24,6 +24,10 @@ module Hyrax
 
       private
 
+      # This method:
+      #
+      # - ensures that the env.attributes[:admin_set_id] is set
+      # - ensures that the permission template for the admin set is correct
       def ensure_admin_set_attribute!(env)
         if env.attributes[:admin_set_id].present?
           ensure_permission_template!(admin_set_id: env.attributes[:admin_set_id])
@@ -36,16 +40,11 @@ module Hyrax
       end
 
       def ensure_permission_template!(admin_set_id:)
-        Hyrax::PermissionTemplate.find_by(source_id: admin_set_id) || create_permission_template!(source_id: admin_set_id)
+        Hyrax::PermissionTemplate.find_or_create_by!(source_id: admin_set_id)
       end
 
       def default_admin_set_id
         AdminSet.find_or_create_default_admin_set_id
-      end
-
-      # Creates a Hyrax::PermissionTemplate for the given AdminSet
-      def create_permission_template!(source_id:)
-        Hyrax::PermissionTemplate.create!(source_id: source_id)
       end
     end
   end
--- a/app/actors/hyrax/actors/default_admin_set_actor.rb
+++ b/app/actors/hyrax/actors/default_admin_set_actor.rb
@@ -6,8 +6,17 @@ module Hyrax
     # Hyrax::Actors::InitializeWorkflowActor, so that the correct
     # workflow can be kicked off.
     #
+    # @see Hyrax::EnsureWellFormedAdminSetService
+    #
     # @note Creates AdminSet, Hyrax::PermissionTemplate, Sipity::Workflow (with activation)
     class DefaultAdminSetActor < Hyrax::Actors::AbstractActor
+      # Hyrax provides a service that ensures well formed admin sets.
+      # It is possible that downstream implementers might seek to
+      # override this behavior.  The class attribute provicdes a means
+      # to override that behavior.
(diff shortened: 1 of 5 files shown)


app/actors/hyrax/actors/ensure_well_formed_admin_set_actor.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Actors
+    # This "Actor" ensures that we have a well formed AdminSet.
+    #
+    # @note For historical reasons, we lazily apply the default admin
+    #   set to curation concerns that don't already have an admin set.
+    #
+    # @see AdminSet
+    # @see Hyrax::PermissionTemplate
+    # @see Hyrax::Actors::DefaultAdminSetActor
+    module EnsureWellFormedAdminSetActor
```
no-reply, line 12: should this be moved outside `hyrax/actors` and renamed something else? it doesn't conform to the actor interface.

no-reply:
@jeremyf this seems good to me, at a glance.

before reviewing more closely, i want to ask whether it's code churn we want to take on. even though it's clearly an improvement to the structure of the existing code, i want to consider its possible merge in the context of our plan to abandon the actor stack---and this code with it. 

@jeremyf, what are your thoughts? is there a factor motivating this refactor, other than as an analysis exercise, that balances the inherent risk of the code churn?

jeremyf:
There are several Actors that don't conform to the Actor interface (I'm looking at the LeaseActor and the EmbargoActor).  This multi-step refactor helps align the "What is a well formed Permission Template", which is logic found in both transactions and the actor stack.

I believe this provides a nice improvement to remove duplication of knowledge, and extract functionality from a Actor, all while preserving external implementations.

no-reply:
> There are several Actors that don't conform to the Actor interface (I'm looking at the LeaseActor and the EmbargoActor). This multi-step refactor helps align the "What is a well formed Permission Template", which is logic found in both transactions and the actor stack.

okay, i'm persuaded that this should be merged.

i still would like to hold out for a rename on the extracted `Actor`. the misnaming of these other classes/modules is a known issue, and will need to be remedied before the actor stack itself is deprecated. this naming issue causes a great deal of confusion: it's a pitfall for folks injecting actors or otherwise customizing their local stack's behaviors, is a barrier to code discovery, and makes communication about the actor stack itself complicated.

i'd strongly prefer not to expand this mistake here.

jeremyf:
I reworked the object, moving it into the `app/services`.

Ideally, I would like to update https://github.com/samvera/hyrax/blob/f219ee3491dc7faee8c316666d1645db80b519a2/lib/hyrax/transactions/steps/set_default_admin_set.rb#L19 to use this service, however I have some "philosophical/ideological" questions as to whether a file in `lib` should have a dependency on `app`.  There may be a dry-container strategy to consider.

app/actors/hyrax/actors/ensure_well_formed_admin_set_actor.rb
jeremyf: I went ahead and moved the one I created.

no-reply (approved):

PR #4531, opened 2020-10-01
index valkyrie work attributes consistent with AF

elrayle:
### Related Issue

Partially addresses Issue #4487 - Solr docs generated by ActiveFedora and Hyrax::ValkyrieWorkIndexer are not compatible

### Description

Most attributes for Hyrax::Work are defined using `Hyrax::Schema` includes.  A few are defined with `attribute` statements.  Some of these attributes were indexed by ActiveFedora.  Adding in indexing for these to be consistent with ActiveFedora’s indexing.

### Additional Differences

Looking for confirmation on how to handle additional attributes not covered by this PR.

Indexed in AF, but not in ValkyrieWorkIndexer…
* generic_type_sim
* file_set_ids_ssim

Defined as attribute in Hyrax::Work, but not indexed
* on_behalf_of
* proxy_depositor
* state

### Observations

The updates in this PR handle the attributes the same way as in AF.  For admin_set and member_of_collections, this involves getting the title from the related resources.  Will we still need the labels going forward?  I am guessing they are used by show pages to avoid needing to instantiate the related objects as those pages are generated.

Looking for confirmation on this as well.

### DIFF
--- a/app/indexers/hyrax/valkyrie_work_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_work_indexer.rb
@@ -11,35 +11,21 @@ module Hyrax
 
     def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
       super.tap do |solr_doc|
-        admin_set_label = admin_set_title(resource.admin_set_id)
-        solr_doc['admin_set_sim'] = admin_set_label
-        solr_doc['admin_set_tesim'] = admin_set_label
-        solr_doc['member_ids_ssim'] = resource.member_ids
-        solr_doc['member_of_collections_ssim']    = collection_titles(resource.member_of_collection_ids)
+        solr_doc['generic_type_sim'] = ['Work']
+        solr_doc['admin_set_id_ssim'] = [resource.admin_set_id]
         solr_doc['member_of_collection_ids_ssim'] = resource.member_of_collection_ids
+        solr_doc['member_ids_ssim'] = resource.member_ids
 
         # attributes from Hyrax::Work not defined by Hyrax::Schema includes
         # solr_doc['on_behalf_of'] = ''
         # solr_doc['proxy_depositor'] = ''
         # solr_doc['state'] = ''
 
-        solr_doc['generic_type_sim'] = ['Work']
-
         # This enables us to return a Work when we have a FileSet that matches
         # the search query.  While at the same time allowing us not to return Collections
         # when a work in the collection matches the query.
         # solr_doc['file_set_ids_ssim'] = solr_doc['member_ids_ssim'] # TODO: Doesn't this return child works too?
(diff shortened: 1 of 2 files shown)


elrayle:
It looks like generic_type_sim is used in multiple places (e.g. collection_presenter.rb, filter_by_type.rb), so I am adding it in.

elrayle:
`state` is one of ACTIVE, DELETED, or INACTIVE as defined in `app/services/hyrax/resource_status.rb`.  I don't see where it is being indexed, but in our production solr, I see solr field `status_tesim` with values `Active` and `Inactive`.  I don't see this field in the valkyrie solr doc, so I will add it as well.

elrayle:
I don't see `on_behalf_of` indexed anywhere, so it is probably fine to leave it out of the valkyrie index.

There is a reference to `proxy_depositor_ssim` in `lib/generators/hyrax/templates`.  I don't see it used anywhere else.  And I don't see it in our production solr docs.

But, we don't use proxy deposit, so it could be it's not there because we didn't use it.

elrayle:
A deeper look at state and ResourceStatus.  These are created based on the state of the object in Fedora.  I'm wondering if it should carry forward?

app/indexers/hyrax/valkyrie_work_indexer.rb
```diff
@@ -8,5 +8,37 @@ class ValkyrieWorkIndexer < Hyrax::ValkyrieIndexer
...
+    end
+
+    private
+
+    def admin_set_title(id)
+      id.present? ? Hyrax.query_service.find_by(id: id)&.title : ""
+    end
+
+    def collection_titles(ids)
+      Hyrax.query_service.find_many_by_ids(ids: ids).map { |col| col.title&.first }
```
no-reply, line 41: it seems bad that we have to fetch all the collections from the persistence backend to index a work. doesn't this also mean we'll need to reindex all collection members if a collection's title changes?

could we do without this index key, and the `admin_set_*` labels in a post valkyrie world, and just index the collection ids instead?

elrayle: Discussed in Slack.  Decision is to go with ids instead of labels.

no-reply:
> A deeper look at state and ResourceStatus. These are created based on the state of the object in Fedora. I'm wondering if it should carry forward?

For better or worse, Hyrax relies heavily on this resource status, via `suppressed?`. we probably don't need to index the state field itself (is it used?), but `suppressed_bsi` is a thing that probably needs to be retained.

There's a documented model/backend agnostic implementation of the fedora status vocabulary and `state` field at `Hyrax::ResourceStatus`

no-reply:
@elrayle this is looking pretty good to me as is. would it be worth taking out the WIP comments and merging this in?

no-reply (approved):

PR #4536, opened 2020-10-07
support tests in the docker environment

no-reply:
run tests with:

```sh
docker-compose exec -w /app/samvera/hyrax-engine app bundle exec rspec
```

----

maybe undesirably, you need the `-w /app/samvera/hyrax-engine` to run the engine
tests; otherwise you'll be trying to run the (non-existant) `.dassie` specs.

most tests currently fail in this environment because dassie doesn't have the
models and other generated MVC code used by the test suite. we'll remedy this in
follow-on work. for now, this makes it possible to work on running the test
suite against `.dassie`, as well as to run individual specs:

```sh
docker-compose exec -w /app/samvera/hyrax-engine app bundle exec rspec spec/hyrax/transactions/apply_change_set_spec.rb
```

----

for Emacs users: i was able to setup the above to work seamlessly with
`rspec-mode` with

```emacs-lisp
(require 'rspec-mode)
(setq rspec-use-docker-when-possible t)
(setq rspec-docker-command "docker-compose exec -w /app/samvera/hyrax-engine")
(setq rspec-docker-cwd "/app/samvera/hyrax-engine/")
(setq rspec-docker-container "app")
```

@samvera/hyrax-code-reviewers


jeremyf (approved):
I followed the instructions and ran the individual specs.  Things worked well.

Issue #4517, opened 2020-09-11
Updating Embargoes from the Manage Embargoes dashboard tab does not work


mcdowney:
### Descriptive summary

Attempting to update active embargoes from the Manage Embargoes section of the dashboard UI does not succeed in updating the embargoes.

This is happening on 3.0.0.pre.rc2 in Chrome, Safari and Firefox.

### Rationale

This seems to have been introduced into the 3.0 release candidate since the most recent round of testing, and was discovered pursuant to test RA_1.20 in the testing script.

### Expected behavior

Clicking on a work in the list of All Active Embargoes from the Manage Embargoes tab should take you to a form that allows you to update the date of expiration for the active embargo. Clicking on "Update Embargo" should persist the change and take the user to the Work's show page, where the updated embargo date now appear.

### Actual behavior

Changing the date of expiration for the embargo does nothing. The user is taken to the work show page, where the green modal appears with the message "Work 'X' successfully updated," but the date listed under "Embargo release date" in the work's metadata remains the old date of expiration.

### Steps to reproduce the behavior

1. As an admin user, navigate to Dashboard > Tasks > Manage Embargoes.
2. Click on a Generic Work in the list of All Active Embargoes
3. Change the desired embargo date.
4. Click on "Update Embargo".
5. Notice the embargo release date in the work's metadata remains the same.

### Related work

Link to related tickets or prior related work here.

eporter23:
Re-tested in Hyrax 3.5 and 4 beta2 and this issue appears to be resolved.

Issue #4516, opened 2020-09-11
Deactivating Leases from Manage Leases dashboard tab does not work


mcdowney:
### Descriptive summary

Attempting to deactivate expired leases from the Manage Leases section of the dashboard UI does not succeed in deactivating the leases.

This is happening on 3.0.0.pre.rc2 in Chrome, Safari and Firefox.

### Rationale

This seems to have been introduced into the 3.0 release candidate since the most recent round of testing, and was discovered pursuant to test RA_1.19 in the testing script.

### Expected behavior

Selecting one or more Expired Active Leases from the "Expired Active Leases" tab and clicking on "Deactivate Leases for Selected" should remove the embargoes from the selected work(s).

### Actual behavior

The green modal appears, advising user to "Select something first", but the works remain in the list of Expired Active Leases. Works scheduled to revert to private after lease expiration remain public.

### Steps to reproduce the behavior

1. As an admin user, navigate to Dashboard > Tasks > Manage Leases.
2. Click on "Expired Active Leases"
3. Select one or more expired leases to deactivate.
4. Click on "Deactivate Leases for Selected"
5. Find the works still in the Expired Active Leases list

### Related work

Link to related tickets or prior related work here.

rjkati:
Deactivating leases using instructions above now displays an error message in iOS Safari and Chrome:

![Image from iOS (1)](https://user-images.githubusercontent.com/32644254/112487033-ccef4680-8d52-11eb-8623-b42f37eff01f.png)

https://nurax-dev.curationexperts.com/leases?locale=en

no-reply:
this latest seems like a new bug. can we open a new ticket for it?

Issue #4515, opened 2020-09-11
Deactivate Embargoes from Manage Embargoes dashboard tab does not work


mcdowney:
### Descriptive summary

Attempting to deactivate expired embargoes from the Manage Embargoes section of the dashboard UI does not succeed in deactivating the embargoes.

This is happening on 3.0.0.pre.rc2 in Chrome, Safari and Firefox.

### Rationale

This seems to have been introduced into the 3.0 release candidate since the most recent round of testing, and was discovered pursuant to test RA_1.18 in the testing script.

### Expected behavior

Selecting one or more Expired Active Embargoes from the "Expired Active Embargoes" tab and clicking on "Deactivate Embargoes for Selected" should remove the embargoes from the selected work(s). 

### Actual behavior

The green modal declaring "Deactivation succeeded!" appears, but the works remain embargoed and in the list of Expired Active Embargoes.

### Steps to reproduce the behavior

1. As an admin user, navigate to Dashboard > Tasks > Manage Embargoes.
2. Click on "Expired Active Embargoes"
3. Select one or more expired embargoes to deactivate.
4. Click on "Deactivate Embargoes for Selected"
5. Find the works still in the Expired Active Embargoes list

### Related work

Link to related tickets or prior related work here.

jeremyf:
This commit ( 241cbed1df4ad416b8d44763ae321e642443ed88 ) changed the form field names, which in turn broke the batch behavior.  I'm reviewing what needs to be done.

PR #4537, opened 2020-10-07
WIP Ensuring embargo and lease are persisted on update

### LINKED ISSUE #4534: Editing Existing Embargo Date Does Not Persist
### Descriptive summary

In local development, with SHA ace33e13c0d7f851ed1c6c5939bbfa8e5d725b64, updating an existing embargo release date via the work's edit page does not persist the updated release date.

1. Create a work with a future embargo release date (7 days from now)
2. Edit the above created work, changing the future embargo release date to one day prior
3. Submit the form
4. Look at the works metadata, the embargo release date has not changed.

### Related work

I believe these are inter-related: #4517, #4516, #4515 

### Further Sleuthing

The related issues and this issue all route through https://github.com/samvera/hyrax/blob/ace33e13c0d7f851ed1c6c5939bbfa8e5d725b64/app/controllers/concerns/hyrax/works_controller_behavior.rb#L195-L198

Note the create action does correctly assign the initial given Embargo Release Date.

jeremyf:
Prior to this commit, when a user changed the Embargo or Lease in the
UI (and by extension via the ActorStack), Hyrax would not persist those
changes.

With this change, we ensure that when the visibility changes, hyrax
updates the embargo and lease.

In consultation with @no_reply, we believe that this should be handled
further upstream (e.g. within `Hydra::AccessControls`).  We also believe
that there was an upstream change that broke the previous correct
behavior.

We are attempting to track that down.

Closes #4534, #4517, #4516, #4515

@samvera/hyrax-code-reviewers


no-reply:
do we think it's worth pushing this through as a hyrax layer fix? or is it better to go straight for hydra access-controls?

app/models/concerns/hyrax/work_behavior.rb
```diff
@@ -30,6 +30,13 @@ module WorkBehavior
       self.indexer = WorkIndexer
       # Default VirusScanner, configurable for Hyrax work types
       self.default_system_virus_scanner = Hyrax::VirusScanner
+      after_update :conditionally_update_embargo_and_lease
+    end
+
+    def conditionally_update_embargo_and_lease
+      return true unless visibility_changed?
```
no-reply, line 37: is this correct? can't i change, e.g., an embargo release date without changing object visibility?

jeremyf: I'll see if `embargo.changed?` and `lease.changed?` can work.

jeremyf:
My preference is to add the behavior in Hyrax, then perform an upstream port of the behavior.  I'd plan on leaving the ActorStack spec but move the WorkBehavior into Hydra::AccessControls.

As such, I'm removing WIP.

jeremyf:
I changed back to WIP as there's a HydraHead patch that should help resolve this samvera/hydra-head#517

The specs should remain but the production code will go

jeremyf:
I believe this is also something we should backport to Hyrax 2.x

no-reply:
this seems good. let's talk about how best to handle the backport.

no-reply (approved):

PR #4540, opened 2020-10-08
Adding feature test for Collection#destroy

### LINKED ISSUE #3872: "Batch deletion" of collections leaves relationship on generic work
### Descriptive summary

Batch edits controller has a `destroy_collection` method. When running the specs to test the behavior, when destroying a collection which has a generic work in it, the generic work retains its `member_of_collections` relationship after the collection is removed. 

### Expected behavior

The member of collections relationships on the works should be removed along with the collection.

1. Run specs
2. Look at what is in Fedora and Solr

### Related work

See comments on https://github.com/samvera/hyrax/pull/3868

jeremyf:
A little bit about this spec.  It's informed by #3762 which is why it
looks a bit odd.  When I wrote this spec it worked.  I suspect that some
prior change between the reported date of June 28, 2019 and
today (October 8, 2020) fixed this bug.

See [Hyrax::BatchEditsController#destroy_collection][1] for the related
code that informed the spec's implementation details.

I'm not prepared nor willing to bisect when this bug was fixed.  I also
understand that we need not merge this commit as the code-base already
works.  I do, however, think about some of the assumptions that we've
found to be broken; [Looking at you Embargo and Lease cascade behavior][2].

Closes #3872

[1]:https://github.com/samvera/hyrax/blob/0e6e7c4ff72fe508d2af0bb2eb22a1a532cf54b3/app/controllers/hyrax/batch_edits_controller.rb#L40-L47
[2]:https://github.com/samvera/hyrax/issue/4534

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/models/concerns/hyrax/collection_behavior_spec.rb
+++ b/spec/models/concerns/hyrax/collection_behavior_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+RSpec.describe Hyrax::CollectionBehavior, clean_repo: true do
+  let(:collection) { create(:collection_lw) }
+  let(:work) { create(:work) }
+
+  describe "#destroy" do
+    it "removes the collection id from associated members" do
+      collection.add_member_objects([work.id])
+      collection.save
+
+      collection_via_query = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: collection.id, use_valkyrie: false)
+
+      expect { collection_via_query.destroy }
+        .to change { work.class.find(work.id).member_of_collection_ids }
+        .from([collection.id])
+        .to([])
+    end
+  end
+end


spec/models/concerns/hyrax/collection_behavior_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    # looks a bit odd.  When I wrote this spec it worked.  I suspect
+    # that some prior change between the reported date of June 28,
+    # 2019 and today (October 8, 2020) fixed this bug.
+    #
+    # See https://github.com/samvera/hyrax/blob/0e6e7c4ff72fe508d2af0bb2eb22a1a532cf54b3/app/controllers/hyrax/batch_edits_controller.rb#L40-L47 for the related code
+    it "removes the collection id from associated members" do
+      collection.add_member_objects([work.id])
+      collection.save
+
+      collection_via_query = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: collection.id, use_valkyrie: false)
```
no-reply, line 19: better to query via ActiveFedora, since this mixin is AF specific?

spec/models/concerns/hyrax/collection_behavior_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+RSpec.describe Hyrax::CollectionBehavior, clean_repo: true do
+  let(:collection) { create(:collection_lw) }
+  let(:work) { create(:work) }
+
+  describe "#destroy" do
+    # A little bit about this spec.  It's informed by
```
no-reply, line 8: i'd prefer to drop this long comment, since it's in the commit message and this test is relevant outside of the context of the particular regression it guards.

spec/models/concerns/hyrax/collection_behavior_spec.rb
jeremyf: My original work had that, but I wanted it to be close to the controller logic which it was verifying.  I'm mostly ambivalent about which to use.

no-reply (approved):

PR #4542, opened 2020-10-12
Pinning Hyrax internal-test-app to SOLR 8.6.2

jeremyf:
Prior to this commit, the internal test app would build against the
latest version of SOLR.  With SOLR 8.6.3, the upstream maintainers of
SOLR introduced a breaking change.  The schema that we use does not work
in 8.6.3.

We envision this commit to be a work-around.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/config_generator.rb
+++ b/lib/generators/hyrax/config_generator.rb
@@ -13,6 +13,7 @@ class Hyrax::ConfigGenerator < Rails::Generators::Base
     * TinyMCE
     * i18n
     * Valkyrie index
+    * SOLR wrapper shims
        """
 
   source_root File.expand_path('../templates', __FILE__)
@@ -36,6 +37,14 @@ class Hyrax::ConfigGenerator < Rails::Generators::Base
   def configure_redis
     copy_file 'config/redis.yml'
     copy_file 'config/initializers/redis_config.rb'
+  end
+
+  # @note We added these to address the fact that the SOLR schema we
+  # use broke in the SOLR 8.6.3 release.  We envision this to be a
+  # stop-gap.
+  def configure_solr_wrappers
+    copy_file 'config/solr_wrapper_test.yml', force: true
+    copy_file 'config/solr_wrapper_dev.yml', force: true
   end
 
   def configure_valkyrie_index
--- a/.regen
+++ b/.regen
@@ -1,1 +1,1 @@
-30zoinks
+31solr-bleck
--- a/lib/generators/hyrax/templates/config/solr_wrapper_dev.yml
+++ b/lib/generators/hyrax/templates/config/solr_wrapper_dev.yml
@@ -1,0 +1,8 @@
+#config/solr_wrapper_test.yml
+version: 8.6.2
+port: 8983
+instance_dir: tmp/solr-development
+collection:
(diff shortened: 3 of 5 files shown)


lib/generators/hyrax/config_generator.rb
```diff
@@ -38,6 +39,14 @@ def configure_redis
     copy_file 'config/initializers/redis_config.rb'
   end
 
+  # @note We added these to address the fact that the SOLR schema we
+  # use broke in the SOLR 8.6.3 release.  We envision this to be a
+  # stop-gap.
+  def configure_solr_wrappers
+    copy_file 'config/solr_wrapper_test.yml'
```
cjcolvar, line 46: Looks like the build failed due to timing out waiting for user input on copying the solr_wrapper test config since it was already generated by `active_fedora:config:solr`.  So maybe it is best to just force overwriting:
```suggestion
    copy_file 'config/solr_wrapper_test.yml', force: true
```

no-reply:
:+1: on this.

before merging, can we document the nature of the Solr schema failure?

jeremyf:
@cjcolvar do you have any documentation? I was following on what you posted in Slack but I wouldn't consider this enough detail.

no-reply:
the release notes for 8.6.3 say "no changes" :joy:

cjcolvar:
The error I got was:
>*_drs of type DateRangeField is incompatible with omitNorms=false; hardcoded behavior is omitNorms=true.  Remove specification in schema

I don't know enough about solr to know if there is an actual behavior change here or rather if it may have been ignoring the `omitNorms=false` for a while and now it is letting us know.

no-reply:
this looks like an ActiveFedora issue: https://github.com/samvera/active_fedora/commit/82a35a4967da92772b28779cc87a8028a945fbcb

maybe it's time to move the AF schema directly to Hyrax and start maintaining it here? or maybe we just need to figure out whether we're supporting this `dateRange` field?

it doesn't look like any of these fields are actively used by Hyrax or ActiveFedora.

jeremyf:
To be clear, the AF schema is the solr schema.  As we move to break our hard dependency on ActiveFedora, it makes a lot of sense to move the schema we use out of Active Fedora.

no-reply:
@jeremyf, @cjcolvar what do you think of just dropping `dateRange` fields across the board?

cjcolvar:
@jeremyf I think that does make sense to do at some point but I'm afraid of the complexity of duplication since we went through the effort a year or so ago to clean up and centralize the solr schema.

@no-reply Thinking of non-Hyrax users, I think it would be good to get a good working solr schema.  Maybe this would be best surfaced with on the email list and see if anyone is using these fields?  But looking at the [original PR](https://github.com/samvera/hydra-head/issues/417) it looks like it was just hypothetical use cases so I'm :+1: to removing them but maybe after due diligence.

no-reply:
i'm now wondering whether adding `omitNorms="true"` to the `fieldType` definition on L87 would be enough?

can someone who is able to reproduce this test?

no-reply:
#4580 confirms that samvera/active_fedora#1439 and a release will fix this.

i think that's the preferred fix now.

PR #4543, opened 2020-10-12
Allowing passed label option to be used

### LINKED ISSUE #3456: Unable to easily customize contributor field on work show page through attribute_rows partial
### Descriptive summary

I had no idea what to use for a title for this issue, so if someone wants to edit it, please feel free.

We are currently using Hyrax version 2.2.3 and doing active development on it.  We were trying to add a custom label the `contributor` field for a particular work type's show page .  The field was displaying as `Contributors` on the work show page even after we defined a custom `label` for the field in our `_attribute_rows.html.erb` partial.  We were expecting any value passed to `_attribute_rows.html.erb` to override the label.

### Rationale

I am not sure if other folks are trying to use custom labels for `keyword` or `contributors`, but it seems like the code in `hyrax/renderers/attribute_renderer.rb` is causing some problematic behavior for some very specific use cases:
```ruby
      def label
        translate(
          :"blacklight.search.fields.#{work_type_label_key}.show.#{field}",
          default: [:"blacklight.search.fields.show.#{field}",
                    :"blacklight.search.fields.#{field}",
                    options.fetch(:label, field.to_s.humanize)]
        )
      end
```
It seems like it is deferring to the `blacklight.search.fields.show` value over whatever value is passed to it through the `_attribute_rows` partial.   I believe this will only affect folks using a custom label for the `contributor` or `keyword` field, so this is a very narrow use case.

### Expected behavior

While using the `_attribute_rows.html.erb` partial, I expected that any value I passed to the `label` key value would override any other label value.

### Actual behavior

If someone attempts to set a label field for the `_attribute_rows.html.erb` for the `keyword` or `contributor` field, they cannot override it.

We have not yet added internationalization for our field labels, but this is something that we want to add eventually.  In our `_attribute_rows` partial, we added the following logic:
```ruby
<% if presenter.model.model_name.name == "OralHistory" %>
  <%= presenter.attribute_to_html(:creator, render_as: :faceted, label: t('hyrax.labels.oral_history.creator')) %>
  <%= presenter.attribute_to_html(:contributor, render_as: :faceted, label: t('hyrax.labels.oral_history.contributor')) %>
<% else %>
```
To replicate this issue, you can attempt to set a custom label for the work show page for the 'contributor' field.

I am not entirely sure if this will be the final edition of this code, but the basic idea is to use the `locale/oral_history.en.yml` file to store a key/value pair like the following:
```yml
    labels:
      oral_history:
        creator:     'Interviewee'
        contributor: 'Interviewer'
```

The basic idea was that we can add custom labels through the various configuration `.yml` files that were generated when we made a new work type.  This seems like a good way to manage our internationalizations in the future.  

The `creator` field was changing to `Interviewee` correctly, but `contributor` was still displaying as `Contributors`, even though there was no reference in our code.

After some digging, I discovered that the problematic portion of code resides in `hyrax/renderers/attribute_renderer.rb`
```ruby
      def label
        translate(
          :"blacklight.search.fields.#{work_type_label_key}.show.#{field}",
          default: [:"blacklight.search.fields.show.#{field}",
                    :"blacklight.search.fields.#{field}",
                    options.fetch(:label, field.to_s.humanize)]
        )
      end
```
This is using the configuration value of `blacklight.search.fields.show.#{field}"`, which, in our case, returned the 'contributor' entry from the engine's version of [locales/hyrax.en.yml](https://github.com/samvera/hyrax/blob/master/config/locales/hyrax.en.yml#L27).  This was overriding the value that we had defined in `_attribute_rows.html.erb`.

Our workaround was to define a new contributor field in the same yml field with a blank value:
```
blacklight:
        show:
          contributor:
```

This fixed the behavior, allowing our custom label to display as we intended for our Oral History work types.

jeremyf:
Prior to this commit, we didn't expose `options[:label]` as a valid
option.

The following quote comes from #3456

> While using the `_attribute_rows.html.erb` partial, I expected that
> any value I passed to the `label` key value would override any other
> label value.

With this change, you can now do something similar to what the original
issue reporter requested:

```erb
<% if presenter.model.model_name.name == "OralHistory" %>
  <%= presenter.attribute_to_html(:creator, render_as: :faceted, label: t('hyrax.labels.oral_history.creator')) %>
  <%= presenter.attribute_to_html(:contributor, render_as: :faceted, label: t('hyrax.labels.oral_history.contributor')) %>
<% else %>
```

Closes #3456

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/renderers/hyrax/renderers/attribute_renderer.rb
+++ b/app/renderers/hyrax/renderers/attribute_renderer.rb
@@ -14,6 +14,9 @@ module Hyrax
       # @param [Symbol] field
       # @param [Array] values
       # @param [Hash] options
+      # @option options [String] :label The field label to render
+      # @option options [String] :include_empty Do we render if if the values are empty?
+      # @option options [String] :work_type Used for some I18n logic
       def initialize(field, values, options = {})
         @field = field
         @values = values
@@ -53,16 +56,23 @@ module Hyrax
         markup.html_safe
       end
 
+      # Defaults to the label provided in the options, otherwise, it
+      # fallsback to the inner logic of the method.
+      #
       # @return The human-readable label for this field.
       # @note This is a central location for determining the label of a field
       #   name. Can be overridden if more complicated logic is needed.
       def label
-        translate(
-          :"blacklight.search.fields.#{work_type_label_key}.show.#{field}",
-          default: [:"blacklight.search.fields.show.#{field}",
-                    :"blacklight.search.fields.#{field}",
-                    options.fetch(:label, field.to_s.humanize)]
-        )
+        if options&.key?(:label)
+          options.fetch(:label)
+        else
+          translate(
+            :"blacklight.search.fields.#{work_type_label_key}.show.#{field}",
+            default: [:"blacklight.search.fields.show.#{field}",
+                      :"blacklight.search.fields.#{field}",
+                      field.to_s.humanize]
+          )
+        end
       end
 
       private
--- a/spec/renderers/hyrax/renderers/attribute_renderer_spec.rb
+++ b/spec/renderers/hyrax/renderers/attribute_renderer_spec.rb
@@ -95,6 +95,14 @@ RSpec.describe Hyrax::Renderers::AttributeRenderer do
 
   describe "#label" do
     subject { renderer }
+    context 'with label provided as an option' do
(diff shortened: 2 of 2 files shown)


bkeese (approved):
Looks good to me. :+1:

PR #4545, opened 2020-10-13
Add ability to define custom change sets for resources

elrayle:
The custom change set inherits from Hyrax::ChangeSet, so it still get properties generated based on resource attributes.

Validations can be added to the custom change set and will be part of what runs when `valid?` is called.

### DIFF
--- a/app/models/hyrax/change_set.rb
+++ b/app/models/hyrax/change_set.rb
@@ -7,10 +7,14 @@ module Hyrax
   # Build a changeset class for the given resource class. The ChangeSet will
   # have fields to match the resource class given.
   #
+  # To define a custom changeset with validations, use naming convention with "ChangeSet" appended to the end
+  # of the resource class name. (e.g. for BookResource, name the change set BookResourceChangeSet)
+  #
   # @example
   #   Hyrax::ChangeSet(Monograph)
   def self.ChangeSet(resource_class)
-    Class.new(Hyrax::ChangeSet) do
+    klass = (resource_class.to_s + "ChangeSet").safe_constantize || Hyrax::ChangeSet
+    Class.new(klass) do
       (resource_class.fields - resource_class.reserved_attributes).each do |field|
         property field, default: nil
       end
--- a/spec/models/hyrax/change_set_spec.rb
+++ b/spec/models/hyrax/change_set_spec.rb
@@ -48,4 +48,39 @@ RSpec.describe Hyrax::ChangeSet do
         .to have_attributes(title: contain_exactly(*titles))
     end
   end
+
+  describe ".for" do
+    context 'when custom change set does not exist' do
+      it 'returns an instance of described_class' do
+        expect(subject).to be_kind_of described_class
+      end
+    end
+
+    context 'when custom change set does exist' do
+      let(:resource) { Hyrax::Test::BookResource.new }
+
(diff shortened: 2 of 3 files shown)


no-reply:
cool! thanks!

i'm not completely in love with the naming convention/`safe_constantize` pattern, but it's in heavy use throughout Hyrax, and this looks like one of the cleaner uses (since it's hidden behind a factory method).

my one question is: who are the callers?

so far the only `ChangeSet.for` caller within Hyrax is in `AccessControlList` to build an ACL ChangeSet. `ResourceForm` and friends inherit the base change set, but do their own configuration/have their own customization. One option we have with non-form change sets is to recommend users make their own with `class MyCustomChangeSet < Hyrax::ChangeSet(MyResourceClass); end` (we could generate this, if we have patterns for recommended use) and have folks use them by name.

elrayle:
This is in support of application customizations. We have custom validations in the work types we define.  The custom changeset inherits from Hyrax::ChangeSet to get the property definitions and then defines the validations.  When `valid?` is called, the custom validations are run.

no-reply:
> This is in support of application customizations.

i get that. i'm just asking whether your application customizations wouldn't be just as well served by you calling your custom class directly, with less Hyrax-layer magic?

e.g.
```ruby
class BookResourceChangeSet < Hyrax::ChangeSet(BookResource)
  validates :isbn, presence: true
end

# later

cs = BookResourceChangeSet.new(my_book_resource)
cs.isbn = '112342342345234'
Hyrax::Transactions::Container['change_set.update_work'].call(cs).value_or { "Oh no!" }
```

elrayle:
I tried your recommendation...

```
class MyCustomChangeSet < Hyrax::ChangeSet(MyResourceClass); end
```

It gets an error...

```
NoMethodError:
  undefined method `ChangeSet' for Hyrax:Module
```

We could generate the custom changeset as written in this PR.

no-reply:
> I tried your recommendation...
> 
> ```
> class MyCustomChangeSet < Hyrax::ChangeSet(MyResourceClass); end
> ```
> 
> It gets an error...
> 
> ```
> NoMethodError:
>   undefined method `ChangeSet' for Hyrax:Module
> ```
> 
> We could generate the custom changeset as written in this PR.

this seems like a (autoload?) bug, since this is the documented usage for `Hyrax::ChangeSet()`.

app/models/hyrax/change_set.rb
```diff
@@ -7,10 +7,14 @@ module Hyrax
...
   #
+  # To define a custom changeset with validations, use naming convention with "ChangeSet" appended to the end
+  # of the resource class name. (e.g. for BookResource, name the change set BookResourceChangeSet)
+  #
   # @example
   #   Hyrax::ChangeSet(Monograph)
   def self.ChangeSet(resource_class)
-    Class.new(Hyrax::ChangeSet) do
+    klass = (resource_class.to_s + "ChangeSet").safe_constantize || Hyrax::ChangeSet
+    Class.new(klass) do
```
jeremyf, line 17: I like this pattern.

I could see a future PR parameterizing a `:default_change_set_class`.

```ruby
def self.ChangeSet(resource_class, default_change_set_class: Hyrax::ChangeSet)
```

jeremyf (approved):

PR #4548, opened 2020-10-14
TEST: try to cause solr 8 to fail on CI

no-reply:
Trying to reproduce the issue behind #4542

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,9 @@ jobs:
       hyrax_valkyrie:
         type: string
         default: "false"
+      solr_version:
+        type: string
+        default: '8.6.3-alpine'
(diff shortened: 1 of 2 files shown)


no-reply:
i seem to be doing something wrong getting this build to go off with Solr 8.6.3.

still just trying to reproduce the error in #4542

cjcolvar:
@no-reply Support for the `solr_version` param was added in https://github.com/samvera-labs/samvera-circleci-orb/pull/30.  It looks like that hasn't made it's way into a formal release yet: https://circleci.com/developer/orbs/orb/samvera/circleci-orb#executors-ruby_fcrepo_solr_redis_postgres

PR #4550, opened 2020-10-15
generate basic changesets for work resources

elrayle:
### Description

Add a generator to create a basic change set and it's spec for work resources created through... 

```
rails generate hyrax:work_resource MyWorkResource
```

Added a shared spec for change sets to the hydra_works shared spec which calls the Valkyrie shared spec for change sets.  Updated Hyrax::ChangeSet's spec to also call the shared spec.

### Related Work

PR #4545 Add ability to define custom change sets for resources

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/work_resource/templates/change_set.rb.erb
+++ b/lib/generators/hyrax/work_resource/templates/change_set.rb.erb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+# Generated via
+#  `rails generate hyrax:work_resource <%= class_name %>`
+class <%= class_name %>ChangeSet < Hyrax::ChangeSet
+  # add custom validations (e.g. validates :custom_field, presence: true)
+end
--- a/lib/generators/hyrax/work_resource/USAGE
+++ b/lib/generators/hyrax/work_resource/USAGE
@@ -9,9 +9,11 @@ Example:
     app/forms/monograph_form.rb
     app/indexers/monograph_indexer.rb
     app/models/monograph.rb
+    app/models/monograph_change_set.rb
     app/views/hyrax/monographs/_monograph.html.erb
     config/initializers/hyrax.rb
     config/metadata/monograph.yaml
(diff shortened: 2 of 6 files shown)


lib/generators/hyrax/work_resource/templates/change_set.rb.erb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Generated via
+#  `rails generate hyrax:work_resource <%= class_name %>`
+class <%= class_name %>ChangeSet < Hyrax::ChangeSet
```
no-reply, line 5: this should probably come with some extensive documentation about how change sets can be used.

if i'm understanding correctly, in a default application this won't be used at all, and editing this file will do nothing. is that right?

elrayle: You do not have to use this change set.  By default, it will behave exactly like Hyrax::ChangeSet.

I would imagine it would be fairly common for there to be custom validations.  That is how I am using it.

Use case:

We have two work types named Publication and Release.  A publication can have many release.  A release belongs to one and only one publication and cannot exist outside a publication.

One of my validations confirms that release.publication_id has a non-blank value.

elrayle: RE: Documentation - Do you envision links to external documentation for the validation options?

no-reply: i think links to external docs seem important.

but also, i think some example usage would help.

> I would imagine it would be fairly common for there to be custom validations

validations added here won't have any impact on any existing code, only on calls that initialize this specific ChangeSet or use the factory. i think it would be reasonable for a user seeing this generated into their codebase to wonder how its used in the engine. the answer "it isn't" will likely be surprising to folks, and i'd like to head that off.

jeremyf:
@elrayle Who's court is this in?

PR #4551, opened 2020-10-15
support indexing Collections using Valkyrie 

no-reply:
wires `Hyrax::PcdmCollection` generated solr documents into presenter tests
using `SolrDocument`. this serves as a fulcrum to add some support for new
valkyrie style collections at the presenter layer.

adds basic metadata to collections and their `Valkyrie::Indexer`. adds index
keys for basic metadata where they are needed to pass collection presenter
tests. adds (some) share specs for basic metadata; there's still more to add
here.

the only presenter behavior left behind are the `#total_viewable_*` methods,
which still depend on `ActiveFedora::Relation#accessible_by`. this is a major
remaining hard `ActiveFedora` dependency, and the specific behavior of
`#accessible_by` continues to get more confusing. in this case, it doesn't seem
to work with Wings at all. this could use some digging into.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/indexers/hyrax/valkyrie_indexer_spec.rb
+++ b/spec/indexers/hyrax/valkyrie_indexer_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe Hyrax::ValkyrieIndexer do
     context 'for a collection' do
       let(:resource) { build(:hyrax_collection) }
 
-      it 'gives an instance of ValkyrieCollectionIndexer' do
+      it 'gives an instance of PcdmCollectionIndexer' do
         expect(described_class.for(resource: resource))
           .to be_a Hyrax::PcdmCollectionIndexer
       end
--- a/spec/indexers/hyrax/valkyrie_indexer_spec.rb
+++ b/spec/indexers/hyrax/valkyrie_indexer_spec.rb
@@ -9,6 +9,15 @@ RSpec.describe Hyrax::ValkyrieIndexer do
         .to eq described_class
     end
 
+    context 'for a collection' do
+      let(:resource) { build(:hyrax_collection) }
+
+      it 'gives an instance of ValkyrieCollectionIndexer' do
+        expect(described_class.for(resource: resource))
+          .to be_a Hyrax::PcdmCollectionIndexer
+      end
+    end
+
     context 'with a matching indexer by naming convention' do
       let(:resource) { build(:monograph) }
       let(:indexer_class) { MonographIndexer }
--- a/spec/presenters/hyrax/collection_presenter_spec.rb
+++ b/spec/presenters/hyrax/collection_presenter_spec.rb
@@ -2,7 +2,7 @@
 RSpec.describe Hyrax::CollectionPresenter do
   subject(:presenter) { described_class.new(solr_doc, ability) }
   let(:collection) do
-    build(:collection_lw,
+    build(:hyrax_collection,
           id: 'adc12v',
           description: ['a nice collection'],
           based_near: ['Over there'],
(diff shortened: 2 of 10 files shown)


spec/indexers/hyrax/valkyrie_indexer_spec.rb
```diff
@@ -9,6 +9,15 @@
         .to eq described_class
     end
 
+    context 'for a collection' do
+      let(:resource) { build(:hyrax_collection) }
+
+      it 'gives an instance of ValkyrieCollectionIndexer' do
```
elrayle, line 15:
```suggestion
      it 'gives an instance of PcdmCollectionIndexer' do
```

spec/presenters/hyrax/collection_presenter_spec.rb
```diff
@@ -120,29 +111,38 @@
...
+        expect(presenter.total_items).to eq 0
+      end
     end
   end
 
   describe "#total_viewable_items", :clean_repo do
     subject { presenter.total_viewable_items }
     let(:ability) { double(::Ability, user_groups: ['public'], current_user: user) }
     let(:user) { create(:user) }
+    let(:collection) { FactoryBot.create(:collection_lw) }
```
elrayle, line 144: Is it intentional for this to be `collection_lw`?  I would have expected this to be `hyrax_collection` based on other changes in this spec.  Should it be...

```suggestion
    let(:collection) { FactoryBot.create(:hyrax_collection) }
```

spec/presenters/hyrax/collection_presenter_spec.rb
```diff
@@ -190,6 +190,8 @@
     subject { presenter.total_viewable_works }
     let(:ability) { double(::Ability, user_groups: ['public'], current_user: user) }
     let(:user) { create(:user) }
+    let(:collection) { FactoryBot.create(:collection_lw) }
```
elrayle, line 193: Same here RE: `collection_lw`.  Should it be...

```suggestion
    let(:collection) { FactoryBot.create(:hyrax_collection) }
```

spec/presenters/hyrax/collection_presenter_spec.rb
```diff
@@ -225,6 +227,8 @@
     subject { presenter.total_viewable_collections }
     let(:ability) { double(::Ability, user_groups: ['public'], current_user: user) }
     let(:user) { create(:user) }
+    let(:collection) { FactoryBot.create(:collection_lw) }
```
elrayle, line 230: Same here RE: `collection_lw`.  Should it be...

```suggestion
    let(:collection) { FactoryBot.create(:hyrax_collection) }
```

elrayle (changes requested):
This looks really good.  Thanks for the work.  Just had a couple of questions.

jeremyf (commented):
@no-reply I echo @elrayle's suggested commit and clarifying questions

spec/presenters/hyrax/collection_presenter_spec.rb
no-reply: i don't think these two suggestions will pass. as noted in the commit messages, these run afoul of explicit `ActiveFedora::Relation#accessible_by` calls in the presenter implementation, and those need to be addressed first.

that's a bigger project.

jeremyf (approved):

spec/presenters/hyrax/collection_presenter_spec.rb
elrayle: Ah.  I missed that.  How would you like to handle this?  I'm fine merging this now and fixing the other problem afterward.  Is there an issue to reference?

elrayle (approved):

PR #4553, opened 2020-10-16
write some tests for SolrDocumentBehavior; make #to_s return a string

no-reply:
this very central module is missing unit tests. in poking at them, i quickly
found that its `#to_s` doesn't always return a string.

this adds a number of unit tests, and hopefully having some will encourage us to
write more.

@samvera/hyrax-code-reviewers


jeremyf (dismissed):

no-reply:
@jeremyf sorry, i had to satisfy robocop. it recommended delegating `#to_s`, but i think that was less clear so i opted to `rubocop:disable` instead.

jeremyf (approved):

PR #4554, opened 2020-10-19
Test calling update_work multiple times

elrayle:
### Description

When calling transaction update_work, the work returned by result.value! does not have the updated attribute value.

### Expected

```
change_set.attr = new_value
result = Hyrax::Transactions::UpdateWork.new.call(change_set)
result.value!.attr # new_value
```

attr should have new_value

### Actual

```
change_set.attr = new_value
result = Hyrax::Transactions::UpdateWork.new.call(change_set)
result.value!.attr # old_value
```

attr should has old_value

### To reproduce

* Run test: spec/hyrax/transactions/update_work_spec.rb:41

This is the new test added in this PR to test calling update_work multiple times for the same work.  It can be used to test for the expected behavior.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/hyrax/transactions/update_work_spec.rb
+++ b/spec/hyrax/transactions/update_work_spec.rb
@@ -34,5 +34,23 @@ RSpec.describe Hyrax::Transactions::UpdateWork, valkyrie_adapter: :test_adapter 
 
       expect(tx.call(change_set).value!).to have_attributes(date_uploaded: uploaded)
     end
+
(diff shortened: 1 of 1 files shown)


elrayle:
The test fails because it is setting `change_set.monograph_title` but testing for `monograph.title`.  It fails when comparing the values.  I would have expected it to fail when the test tries to set field `monograph_title` which doesn't exist.

CLOSING

PR #4555, opened 2020-10-20
add shared spec for change sets

elrayle:
This was extracted from PR #4550.  There were questions in that PR about the usefulness of generating changesets for generated works.  I would like to use the shared specs from that PR to test my changesets.  Hoping to get this PR merged while we continue to explore the usefulness of generating changesets.

@samvera/hyrax-code-reviewers


lib/hyrax/specs/shared_specs/hydra_works.rb
```diff
@@ -291,3 +290,13 @@
     end
   end
 end
+
+RSpec.shared_examples 'a Hyrax::ChangeSet' do
```
no-reply, line 294: should this get organized into a different file? `lib/hyrax/specs/shared_specs/change_set.rb`?

no-reply (commented):
looks good to me! thank you!

i think it should probably go in a different file, since it doesn't test Hydra Works model compliance. would be handy to be able to reuse this without pulling in the other Hydra Works shared example sets.

lib/hyrax/specs/shared_specs/hydra_works.rb
elrayle: Good idea.  DONE

lib/hyrax/specs/shared_specs/hydra_works.rb
```diff
@@ -130,7 +130,6 @@
 
   it_behaves_like 'a Hyrax::Resource'
   it_behaves_like 'a model with core metadata'
-  it_behaves_like 'a model with basic metadata'
```
no-reply, line 133: i missed this on the first pass. just want to check that this is deliberate.

it seems right to remove it, at first glance, but also unrelated(?). is this a considered change?

elrayle: This was introduced by a rebase against master.  I just missed it when I created the new PR. It was brought in PR #4551 just a few days ago.  I will remove this change.

elrayle: Looking a little closer, this was GitHub's auto-rebase.  I didn't have this change in my local repo.  I'm rebasing locally and will push up again.

elrayle: This is resolved.  That was a good catch.

no-reply (approved):

PR #4560, opened 2020-10-28
Misc. dassie and docker-compose test environment improvements

no-reply:
these are various changes to make the test environment for `docker-compose` more accurate. they consist of generalizing changes to the test suite, and configuration tweaks, especially for the `.dassie` test app.

@samvera/hyrax-code-reviewers


lib/hyrax/configuration.rb
```diff
@@ -307,7 +307,8 @@ def browse_everything?
     attr_writer :analytics
     attr_reader :analytics
     def analytics?
-      @analytics ||= false
+      @analytics ||=
+        ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_ANALYTICS', false))
```
orangewolf, line 311: yay!

spec/lib/hyrax/controlled_vocabulary/importer/language_spec.rb
```diff
@@ -11,7 +11,8 @@
   end
 
   let(:instance) { described_class.new }
-  let(:rdf_path) { Gem.loaded_specs['hyrax'].full_gem_path + "/.internal_test_app/tmp/lexvo_2013-02-09.rdf" }
+  let(:rdf_path) { "#{root_path}/tmp/lexvo_2013-02-09.rdf" }
+  let(:root_path) { ENV.fetch('RAILS_ROOT', "#{Gem.loaded_specs['hyrax'].full_gem_path}/.internal_test_app") }
```
orangewolf, line 15: why not Rails.root here?

spec/spec_helper.rb
```diff
@@ -229,7 +232,7 @@ def main_app
   end
 
   config.after(:each, type: :view) do
-    WebMock.disable_net_connect!(allow_localhost: true, allow: 'chromedriver.storage.googleapis.com')
+    WebMock.disable_net_connect!(allow_localhost: true, allow: allowed_hosts)
```
orangewolf, line 235: where does allowed_hosts come from?

orangewolf (commented):
Good stuff here, what's the current expectation on how much of the spec suite should run in docker from this branch?

spec/lib/hyrax/controlled_vocabulary/importer/language_spec.rb
no-reply: oh, that probably works. i'll give it a try.

spec/spec_helper.rb
no-reply: it's defined further above, in this file. the set had already been applied in initial `WebMock` setup. this reset was causing test order issues whenever view tests are run as part of a larger set.

no-reply:
> Good stuff here, what's the current expectation on how much of the spec suite should run in docker from this branch?

i don't have a full report on how many are failing at hand, but the short answer is "not all".

https://github.com/samvera/hyrax/tree/capybara is WIP follow-up on getting capybara working in the docker env. that should bring us to "the vast majority" with a few things to clean up along the way.

i'm running the full suite locally while CI tries the capybara branch now (i'm not sure if i've broken the circle setup there, yet :clock1: )

spec/lib/hyrax/controlled_vocabulary/importer/language_spec.rb
no-reply: okay, i think i've got this now.

no-reply:
good news is the capybara stuff works, excepting a `element not interactable: element has zero size` error that seems to be newly intermittent.

would you rather review that seprately as follow-up, or should i go ahead and push it here?

no-reply:
whee! i went ahead and pushed everything *except* the actual capybara/selenium setup.

this now gets us down to perhaps 3 failing tests(? still running suite locally)

orangewolf (approved):
this looks great, CI is green, the birds outside are chirping, I saw ship it and we are almost at a fully dockerized dev / test env, which is very exciting

PR #4565, opened 2020-10-29
Introduce Hyrax::SearchService to mirror Blacklight::SearchService 

cbeer:
Blacklight introduces Blacklight::SearchService in Blacklight 7. Adding it to Hyrax now provides a stepping stone to get to Blacklight 7 easily and has all the added benefits by making explicit the context clues used for searches.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -5,6 +5,11 @@ module Hyrax
     class CollectionsController < Hyrax::My::CollectionsController
       include Blacklight::AccessControls::Catalog
       include Blacklight::Base
+
+      configure_blacklight do |config|
+        config.search_builder_class = Hyrax::Dashboard::CollectionsSearchBuilder
+      end
+
       include BreadcrumbsForCollections
       with_themed_layout 'dashboard'
 
@@ -192,10 +197,6 @@ module Hyrax
         render json: result
       end
 
-      def search_builder_class
-        Hyrax::Dashboard::CollectionsSearchBuilder
-      end
-
       private
 
       def default_collection_type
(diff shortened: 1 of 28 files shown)


app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -192,10 +197,6 @@ def files
         render json: result
       end
 
-      def search_builder_class
```
jeremyf, line 195: I'm a little curious about this going away.  But will need to do some more investigating on my local machine.

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -162,13 +161,23 @@ def attributes
 
     def presenter
       @presenter ||= begin
-        _, document_list = search_results(params)
-        curation_concern = document_list.first
-        raise CanCan::AccessDenied unless curation_concern
-        show_presenter.new(curation_concern, current_ability, request)
+        show_presenter.new(curation_concern_document, current_ability, request)
```
jeremyf, line 164: I like this refactor!

jeremyf (commented):
I'm wondering about adding a delegate method to the various places that are losing the `search_builder_class`? I suspect that not many are updating those, but it may be helpful for backwards compatability.

I'm happy to add those delegates (and deprecation warnings)

app/controllers/hyrax/dashboard/collections_controller.rb
cbeer: It's been a few weeks but... this class overrides the method to change the search builder but configuring it in the blacklight config (as above) is a happier path for blacklight. Downstream code shouldn't see any difference because blacklight's going to provide `search_build_class` with the right value 🤷‍♂️

jeremyf (approved):

PR #4566, opened 2020-10-29
configure capybara to run in the `docker-compose` environment

no-reply:
use the `selenium/standalone-chrome` image to provide a browser engine. move
configuration for capybara to a shared location `lib/hyrax/specs` where it can
be reused by applications if desired.

@samvera/hyrax-code-reviewers


jeremyf (approved):

orangewolf:
Missed the window to comment here - but this is fantastic

PR #4570, opened 2020-10-30
Rework Hyrax's `spec_helper`

no-reply:
move a bunch of utility code to places where apps can reuse. refactor `spec_helper` to use RSpec features instead of hand-written branching logic.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -38,6 +38,7 @@ end
 
 ActiveRecord::Migration.maintain_test_schema!
 
+require 'active_fedora/cleaner'
 require 'devise'
 require 'devise/version'
 require 'mida'
@@ -70,11 +71,6 @@ WebMock.disable_net_connect!(allow_localhost: true, allow: allowed_hosts)
 require 'i18n/debug' if ENV['I18N_DEBUG']
 require 'byebug' unless ci_build?
 
-# require 'http_logger'
-# HttpLogger.logger = Logger.new(STDOUT)
-# HttpLogger.ignore = [/localhost:8983\/solr/]
-# HttpLogger.colorize = false
-
 require 'hyrax/specs/shared_specs/factories/strategies/json_strategy'
 require 'hyrax/specs/shared_specs/factories/strategies/valkyrie_resource'
 FactoryBot.register_strategy(:valkyrie_create, ValkyrieCreateStrategy)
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-$VERBOSE = nil # silence loud Ruby 2.7 deprecations
+$VERBOSE = nil unless ENV['RUBY_LOUD'] # silence loud Ruby 2.7 deprecations
```
orangewolf, line 2: loud

this is great

spec/spec_helper.rb
```diff
@@ -139,7 +102,13 @@ def main_app
 
 ActiveJob::Base.queue_adapter = :test
 
-require 'active_fedora/cleaner'
+def clean_active_fedora_repository
+  ActiveFedora::Cleaner.clean!
+  # The JS is executed in a different thread, so that other thread
+  # may think the root path has already been created:
+  ActiveFedora.fedora.connection.send(:init_base_path)
```
orangewolf, line 109: tricky and good to know. thank you for commenting this.

spec/spec_helper.rb
```diff
@@ -151,24 +120,19 @@ def main_app
     c.syntax = :expect
   end
 
-  # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
   config.fixture_path = File.expand_path("../fixtures", __FILE__)
-
   config.use_transactional_fixtures = false
 
   config.before :suite do
+    Hyrax::RedisEventStore.instance.redis.flushdb
```
orangewolf, line 127: do both supported set up methods (docker and ~boring~ classic) make it clear that redis should be separate between test and dev?

spec/spec_helper.rb
```diff
@@ -151,24 +120,19 @@ def main_app
...
     DatabaseCleaner.clean_with(:truncation)
     # Noid minting causes extra LDP requests which slow the test suite.
     Hyrax.config.enable_noids = false
+    # Don't use the nested relationship reindexer. Null is much faster
+    Hyrax.config.nested_relationship_reindexer = ->(id:, extent:) {}
```
orangewolf, line 132: it feels like this should be part of a app test harness as well as I bet it speeds things up a lot. any thoughts on how to do that in an accessible way?

spec/spec_helper.rb
no-reply: i guess Ruby 2.7.2 changes this behavior though. we can't win on this one.

spec/spec_helper.rb
no-reply: i'm not sure. either way this rewrite makes us flush the redis db less frequently. the old setup ran it every test!

spec/spec_helper.rb
no-reply: it would be easy enough to move something like this into `lib/hyrax/spec/`... the tricky bit is how we communicate with folks about what's available there.

there's starting to be a lot of good stuff shipping with Hyrax to support application test suites, and i think it's probably the case that few people really know about it. 

dev congress topic?

spec/spec_helper.rb
orangewolf: sounds like no harm then =-)

orangewolf (approved):

PR #4579, opened 2020-11-11
Pin docker solr to 8.6.2

bkeese:
refs #4542 

Pin SOLR version to 8.6.2 in docker-compose file to address solr schema that is invalid in solr 8.6.3.

@samvera/hyrax-code-reviewers

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -88,7 +88,7 @@ services:
       - redis:/data
 
   solr:
-    image: solr:8
+    image: solr:8.6.2
     ports:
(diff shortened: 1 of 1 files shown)


no-reply:
@bkeese are there steps to reproduce this issue with the docker environment? i haven't been able to make it fail, and would prefer not to block support for Solr 8.6.3 and 8.7.0 if we can come up with a genuine fix.

no-reply:
i was able to reproduce! closing this in favor of #4580

PR #4580, opened 2020-11-12
bump solr support to 8.7

no-reply:
test Hyrax on solr 8.7 in `docker-compose`. the version used in CI is set by the samvera orb, and doesn't seem like it can be overridden from here.

change the `.dassie` solr schema to use `omitNorms="true"` on dateRange. from 8.6.3 Solr fails aggressively on the default behavior(!) https://issues.apache.org/jira/browse/SOLR-14859

for new apps, this needs to be merged into ActiveFedora https://github.com/samvera/active_fedora/pull/1439

supersedes #4579 and #4542 

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/solr/conf/schema.xml
+++ b/.dassie/solr/conf/schema.xml
@@ -84,7 +84,7 @@
     <!-- A PointField based date field for faster date range queries and date faceting. -->
     <fieldType name="tdate" class="solr.DatePointField" docValues="true"/>
     <!-- A DateRange based date field for truly faster date range queries. -->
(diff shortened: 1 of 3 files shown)


cjcolvar:
FWIW the solr version can be configured in the orb but that functionality hasn't made it into a release yet so isn't actually available to CI builds.

bkeese (dismissed):
Looks good to me.

cjcolvar (approved):
This looks good to me.  I was able to manually start up a solr container using the commands in docker-compose and it created the collection/core without any errors.  That's as much as I can test at the moment, but I think it is good to go.

PR #4592, opened 2020-11-16
Long standing missing break, and spelling error

sephirothkod:
The missing `break` for based_near has existed for years, and causes issues when devs add new controlled vocabulary fields for their instances. More experienced devs know right away to add it but many newcomers don't know that it is required.

Also fixed a spelling error for the word `exclude`

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/autocomplete.es6
+++ b/app/assets/javascripts/hyrax/autocomplete.es6
@@ -19,7 +19,7 @@ export default class Autocomplete {
 
   byDataAttribute(element, url) {
     let type = element.data('autocomplete-type')
-    let exlude = element.data('exclude-work')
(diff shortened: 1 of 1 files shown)


no-reply:
@sephirothkod thanks for this.

it looks like it's failing some of the tests on CI. any ideas?

crisr15:
CLA was verified on 2021-03-22.

orangewolf:
@crisr15 can we put someone on fixing these 2 jasmine tests and get this pushed back up, they'll need to coordinate with @sephirothkod to get access since it is a personal fork, but it doesn't have to be Braydon that does the work

orangewolf:
poke

PR #4595, opened 2020-11-16
drop docker-compose file version to 3.7

dunn:
3.8 wasn't added until this February, so users on slower release channels will
see an error: https://github.com/docker/compose/blob/master/CHANGELOG.md#features-3

@samvera/hyrax-code-reviewers

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,5 @@
-version: '3.8'
+version: '3.7'
+
 services:
   app: &app
     build:


cjcolvar (approved):
My ubuntu machine didn't have a new enough version of docker-compose to run the 3.8 docker-compose.yml before but it ran fine with the version downgrade in this PR.

jeremyf (approved):

PR #4599, opened 2020-11-17
Ensuring AdminSet properly indexed during seeds

jeremyf:
I have found that when I come back to a development environment, that I
may have an AdminSet in Fedora, but it is not indexed in Solr.  This
remediates that situation by ensuring we have an indexed AdminSet.

We [merged this change merged][1] into `rake
hyrax:default_admin_set:create` but it was not part of the db:seeds.

[1]:https://github.com/samvera/hyrax/commit/940e1003e35b55a1aab69cdc2a231269fe7f29bf

@samvera/hyrax-code-reviewers

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -11,3 +11,10 @@ abort("Failed to process all workflows:\n  #{errors.join('\n  ')}") unless error
 
 puts "\n== Creating default admin set"
 AdminSet.find_or_create_default_admin_set_id
+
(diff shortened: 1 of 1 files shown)


no-reply (approved):
is this a problem you're actually having? i'm wondering how i could reproduce.

in either case, it seem harmless, so :100:

jeremyf:
This is a problem I'm having.  I'm uncertain why.

PR #4600, opened 2020-11-17
Addressing "removed from the module tree" error

jeremyf:
Previously, when I ran the container version of local Hyrax development,
I (and others), encountered the following exception:

```sh
ArgumentError in Hyrax::GenericWorksController#create

A copy of Hyrax::Listeners::ObjectLifecycleListener has been removed
from the module tree but is still active!
```

[As reported by others][1], it is often unclear what to do.  I attempted
to add the `autoload` directive, and it appears to work.  Here's hoping
it works for you as well.

[1]:https://discuss.rubyonrails.org/t/yet-another-a-copy-of-xxx-has-been-removed-from-the-module-tree-but-is-still-active-can-we-add-an-example/75744

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/listeners.rb
+++ b/app/services/hyrax/listeners.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+module Hyrax
+  # @note
+  #    Did you encounter an exception similar to the following:
+  #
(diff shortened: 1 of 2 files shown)


no-reply (approved):
oh thank you!

Issue #4559, opened 2020-10-28
cannot display Valkyrie Resource saved through new form - raises undefined method `model_name' for nil:NilClass


elrayle:
### Descriptive summary

Cannot display a valkyrie resource work that was created with the valkyrie based form.  Note that the form may be working fine.  I suspect it is something to do with the structure of the resource or the change set that is generated for the resource.

### Rationale

Valkyrie Resources should work with save and display.

### Expected behavior

Attempting to display a valkyrie resource on the show view after it is created through the valkyrie based form should show the valkyrie resource work with the values set and saved in the new form.

### Actual behavior

Attempting to display a valkyrie resource on the show view after it is created through the valkyrie based form raises an exception.

### Steps to reproduce the behavior

1. Create a monograph work using http://localhost:3000/concern/monographs/new
2. Fill in required info and Save

When it forwards to the display page after saving, it raises `undefined method 'model_name' for nil:NilClass` in `app/models/concerns/hyrax/solr_document_behavior.rb`.

I haven't completely traced this, but it appears to have something to do with permissions and missing attribute admin_set.

elrayle:
I can read and write with pure valkyrie calls.

This works...
```
m = Monograph.new                                  # create a new monograph
mcs = Hyrax::ChangeSet.for(m)                      # create a change set for the new monograph
mcs.title = 'My Monograph'                         # set a title
mcs.admin_set_id = 'admin_set/default'             # set the admin set
mid = Hyrax.persister.save(resource: mcs.sync).id  # sync the change set and save the resource
mr = Hyrax.query_service.find_by(id: "2227mp645")  # find the resource
```

This doesn't work...
```
m = Monograph.new                                  # create a new monograph
mcs = Hyrax::ChangeSet.for(m)                      # create a change set for the new monograph
mcs.title = 'My Monograph'                         # set a title
mcs.admin_set_id = 'admin_set/default'             # set the admin set
Hyrax::Transactions::CreateWork.new.call(mcs) # use transaction to save
```

raises undefined method `admin_set'

elrayle:
If I add a method to the change set that returns the admin set, I get past the error thrown by running the transaction. But this only leads me to another error...

```
NoMethodError (undefined method 'edit_groups' for #<#<Class:0x00007fea6ab92f38>:0x00007fea6abc8278>)
```

no-reply:
[...]

PR #4606, opened 2020-11-17
add proxy_depositor and on_behalf_off to resource_form

### LINKED ISSUE #4596: Forms for Valkyrie works do not respond to on_behalf_of
### Descriptive summary

Generate form classes for Valkyrie resource works do not respond to the on_behalf_of method.  For ActiveFedora forms, this method is delegated to the model.  Need to look at all the delegated methods for `Hyrax::Forms::WorkForm` to determine how to delegate those methods to the valkyrie resource model.

### Rationale

Proxies should work the same for Valkyrie resource works as they do for ActiveFedora work.

### Expected behavior

When a user can create works on behalf of another user, the form should load and allow the user to create a work for the other user.

### Actual behavior

When a user can create works on behalf of another user, the form fails to load with error...

```
Error: undefined method `on_behalf_of'
```

1. Create 2 users
2. Grant user 1 as a proxy depositor for user 2
3. login as user 1
4. Dashboard -> Works -> Add new work -> Monograph

The error occurs as the form is being built.

### Related work

Issue #4569 - Feature test: new form - resource new form fields are generated correctly

elrayle:
Fixes #4596

AF work forms delegated proxy_depositor and on_behalf_off to the AF model.  Similarly, this PR delegates those to the Valkyrie resource work model.  There are several other fields that are delegated to the Valkyrie resource work (e.g. :depositor, :human_readable_type).

Coming soon will be a feature test that allows proxy depositors that will demonstrate this change in the context of the UI.  The feature test related to this change already works, but a PR with those changes is awaiting a bug fix unrelated to this PR.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -65,7 +65,7 @@ module Hyrax
 
       class_attribute :model_class
 
-      delegate :depositor, :human_readable_type, to: :model
+      delegate :depositor, :human_readable_type, :on_behalf_of, :proxy_depositor, to: :model
 
       property :visibility # visibility has an accessor on the model
 
--- a/spec/forms/hyrax/forms/resource_form_spec.rb
+++ b/spec/forms/hyrax/forms/resource_form_spec.rb
@@ -177,6 +177,24 @@ RSpec.describe Hyrax::Forms::ResourceForm do
     end
   end
 
+  describe '#proxy_depositor' do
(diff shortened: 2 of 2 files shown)


no-reply (approved):
i think this is only partially finished work (do we need to be able to write these properties from the form?) so i'm going to keep the ticket open. but also merging this through to unstick continued work.

elrayle:
Thanks, I will dig deeper once Issue #4559 is fixed as it is also blocking resolution of new form feature tests.

Issue #4556, opened 2020-10-21
Wings: Roundtrip of single value attribute become an array after saving


elrayle:
### Descriptive summary

In a valkyrie resource, saving the resource converts single value attributes from String to Array.

### Rationale

The roundtrip conversion from resource to AF object and back to resource should not change the value of an attribute.

### Expected behavior

```
resource = MyResource.new(single_value_attr: "one")
resource. single_value_attr           # "one"
updated_resource = Hyrax.persister.save(resource: resource)
updated_resource. single_value_attr   # "one"
```

NOTE: single_value_attr is "one" before and after the save

### Actual behavior

```
resource = MyResource.new(single_value_attr: "one")
resource. single_value_attr           # "one"
updated_resource = Hyrax.persister.save(resource: resource)
updated_resource. single_value_attr   # ["one"]
```

NOTE: single_value_attr is "one" before and an Array ["one"] after the save

### Steps to reproduce the behavior

TBD - creating a test to demonstrate the problem

tpendragon:
@elrayle This happens when `single_value_attr` is defined as a `Valkyrie::Types::String`?

no-reply:
at first glance, i think this is expected `Wings` behavior.

`ActiveFedora`/Fedroa still treats these attributes as multi-valued RDF properties, and Wings doesn't try to cast back to a single value, since that risks silent data loss.

elrayle:
@tpendragon It happens for all single value types for the reason described by @no-reply.

elrayle:
How Valkyrie handles this [@tpendragon in Slack]...

>It’s just if there’s only one value it picks off the first one in the conversion from Fedora->Resource, and then if it’s defined as a Valkyrie::Set the set type conversion takes care of making it an array again.

>Valkyrie’s conversion logic for Fedora is mad complicated, but https://github.com/samvera/valkyrie/blob/master/lib/valkyrie/persistence/fedora/persister/orm_converter.rb#L524-L537 is the hunk of code for this.

>Postgres (and everything else) uses this: https://github.com/samvera/valkyrie/blob/master/lib/valkyrie/persistence/shared/json_value_mapper.rb#L107-L130

elrayle:
Closing... Addressed by PR #4607 - wings: handle single value fields for valkyrie native models

PR #4607, opened 2020-11-17
wings: handle single value fields for valkyrie native models

no-reply:
make TransformerValueMapper/OrmConverter treat single value arrays as singular
value objects by default. change the generated fields for ad-hoc valkyrie models
to be Set.of(Anything), by default. don't override existing fields when
generating valkyrie models.

related to #4556

Co-authored-by: Trey Pendragon <tpendragon@princeton.edu>

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/wings/attribute_transformer_spec.rb
+++ b/spec/wings/attribute_transformer_spec.rb
@@ -24,7 +24,7 @@ RSpec.describe Wings::AttributeTransformer do
 
   it "transform the attributes" do
     expect(subject).to include title: work.title,
-                               contributor: work.contributor,
-                               description: work.description
+                               contributor: work.contributor.first,
+                               description: work.description.first
   end
 end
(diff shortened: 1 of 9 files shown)


spec/wings/attribute_transformer_spec.rb
```diff
@@ -24,7 +24,7 @@
 
   it "transform the attributes" do
     expect(subject).to include title: work.title,
-                               contributor: work.contributor,
-                               description: work.description
+                               contributor: work.contributor.first,
+                               description: work.description.first
```
elrayle, line 28: Looking at GenericWork, it defines `contributor` and `description` without the `multiple:`.  I thought that defaulted to true.  In which case, I would expect these to be multi-valued in the subject.  Am I missing something?

elrayle (commented):
Looks really good.  Just one small question.

spec/wings/attribute_transformer_spec.rb
no-reply: what you're missing is that this module only creates an attribute hash, not a valkyrie object. the `Set.of(Anything)` dry-types spec is what casts back to a set (if appropriate).

elrayle (approved):
Thanks for working on this.

PR #4624, opened 2020-11-18
wings: add benchmark for conversion

no-reply:
squash some logging and add a benchmark for ActiveFedoraConverter

@samvera/hyrax-code-reviewers


tasks/benchmark.rake
```diff
@@ -4,10 +4,33 @@ namespace :wings do
...
+      end
+
+      x.report "convert a derived Valkyrie object back to ActiveFedora" do
+        Wings::ActiveFedoraConverter.convert(resource: generic_work_resource)
+      end
+    end
+  end
+
   task benchmark_save: :environment do
+    # Base Case: 1.4 / Second
```
no-reply, line 26:
```suggestion
```

tpendragon (approved):

PR #4626, opened 2020-11-19
wings: remove dead AF 11 patch

no-reply:
this code says it can be removed. let's remove it!

@samvera/hyrax-code-reviewers


cjcolvar (approved):
:scissors: :tada:

PR #4627, opened 2020-11-19
wings: avoid errant test setup

no-reply:
in the test as written somehow(?) adds the work as a member of itself.

adding this test to the `before` block fails:

```ruby
  work1.reload
  expect(work1.members.map(&:id)).to contain_exactly('wk2', 'wk3', 'fs1', 'fs2')
```

removing the ids avoids the issue and tests the correct behavior.

@samvera/hyrax-code-reviewers


elrayle (approved):
This change is fine.  Using ids was really only for debugging.  They are not necessary.

PR #4628, opened 2020-11-19
wings: use find to build AF objects 

no-reply:
this speeds up active fedora converter by in isolation:

before:
```
convert a Monograph object to ActiveFedora
                          6.984  (±14.3%) i/s -     35.000  in   5.055449s
convert a derived Valkyrie object back to ActiveFedora
                          7.846  (±12.7%) i/s -     39.000  in   5.024344s
```

after:

```
convert a Monograph object to ActiveFedora
                         72.812  (± 6.9%) i/s -    366.000  in   5.052053s
convert a derived Valkyrie object back to ActiveFedora
                         32.600  (± 6.1%) i/s -    165.000  in   5.079857s
```
 and save in general:

before

```
save a Valkyrie object
                          1.718  (± 0.0%) i/s -      9.000  in   5.265884s
```

after

```
save a Valkyrie object
                          4.509  (±22.2%) i/s -     23.000  in   5.192596s
```

@samvera/hyrax-code-reviewers


no-reply:
oh dear. i can't figure out where these `Ldp::Gone` messages are coming from.

the error messages are empty `err.message=> ""`, and capturing them, then doing `af_object.reload; af_object.save` succeeds!

no-reply:
this got to be a rather large diff, but i think it's ready!

```

File: lib/wings/active_fedora_converter/nested_resource.rb [hunk 1]
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Wings
```
no-reply, line 3: this is a straight extraction, without changes.

lib/wings/active_fedora_converter/default_work.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Wings
```
no-reply, line 3: this is a straight extraction, without changes.

lib/wings/active_fedora_converter.rb
```diff
@@ -227,21 +128,35 @@ def normal_attributes
...
-      af_object.permissions = cache.map do |permission|
-        permission.access_to_id = resource.try(:access_to)&.id
-        permission
-      end
+      normal_attributes[:permissions].each { |p| p.access_to_id = resource.access_to&.id }
+      af_object.permissions = normal_attributes[:permissions]
+    end
+
+    # for files, add attributes to metadata_node, plus some other work
+    def add_file_attributes(af_object)
```
no-reply, line 154: special treatment for `ActiveFedora::File` and subclasses is now isolated here.

lib/wings/active_fedora_converter.rb
```diff
@@ -62,26 +64,22 @@ def attributes
...
+      @active_fedora_class ||= # cache the class at the instance level
         begin
           klass = begin
                     resource.internal_resource.constantize
                   rescue NameError
                     Wings::ActiveFedoraClassifier.new(resource.internal_resource).best_model
                   end
 
-          return klass if klass <= ActiveFedora::Base
+          return klass if klass <= ActiveFedora::Common
```
no-reply, line 82: we want to catch both `Base` and `File` types. `Common` is a better umbrella. to be honest, we're not seriously testing the interface it defines (i think it would be hard to draw a boundary around it).

lib/wings/active_fedora_converter.rb
```diff
@@ -62,26 +64,22 @@ def attributes
...
-        convert_files(af_object)
+      instance.tap do |af_object|
+        af_object.id ||= id unless id.empty?
+        apply_attributes_to_model(af_object)
       end
     end
 
     def active_fedora_class
-      @active_fedora_class ||=
+      @active_fedora_class ||= # cache the class at the instance level
```
no-reply, line 74: this cache was already introduced, but this bit of documentation got left out. i think it's nice to call it out as wanted, since there is also a class level cache.

lib/wings/services/custom_queries/find_file_metadata.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
-# Custom query override specific to Wings for finding Hydra::PCDM::File and converting to Hyrax::FileMetadata.
-# @example
-#   Hyrax.custom_queries.find_file_metadata_by(id: valkyrie_id, use_valkyrie: true)
-#   Hyrax.custom_queries.find_file_metadata_by_alternate_identifier(alternate_identifier: id, use_valkyrie: true)
-require 'wings/services/file_converter_service'
 module Wings
   module CustomQueries
+    # Custom query override specific to Wings for finding Hydra::PCDM::File and converting to Hyrax::FileMetadata.
```
no-reply, line 4: this class is largely rewritten, since the basic Wings components now do most of its work. `FileConverterService` is take out of service completely (and deprecated). there are also refactors here. changes are separated across several commits to keep refactors and functional changes apart. see commit messages for more.

lib/wings/services/file_converter_service.rb
```diff
@@ -1,67 +1,18 @@
 # frozen_string_literal: true
 # Conversion service for going between files on a valkyrie resource and files on an active fedora object
 module Wings
+  # @deprecated use generic Wings converters instead.
+  # @see ModelTransformer, ActiveFedoraConverter
```
no-reply, line 5: all this behavior is now delegated to the generic wings converter/transformer pair, and this class is deprecated.

app/models/hyrax/file_metadata.rb
```diff
@@ -140,8 +139,20 @@ def valid?
       file.valid?(size: size.first, digests: { sha256: checksum&.first&.sha256 })
     end
 
+    ##
+    # @deprecated get content from #file instead
+    #
+    # @return [#to_s]
+    def content
+      Deprecation.warn('This convienince method has been deprecated. ' \
```
no-reply, line 147: deprecating this. users should get comfortable using valkyrie's interfaces.

considering deprecating `#file` for the same reason.

app/services/hyrax/versioning_service.rb
```diff
@@ -66,16 +48,8 @@ def perform_create_through_active_fedora(content, user)
...
-        # rubocop:disable Lint/UnreachableCode
-        new_version = content.new(id: nil)
-        new_version.label = "version#{content.member_ids.length + 1}"
-        new_version = indexing_adapter.persister.save(resource: new_version)
-        content.member_ids = content.member_ids + [new_version.id]
-        content = indexing_adapter.persister.save(resource: content)
-        record_committer(content, user) if user
-        # rubocop:enable Lint/UnreachableCode
+      def perform_create_through_valkyrie(content, user) # no-op
+        raise NotImplementedError
```
no-reply, line 52: this already did nothing. better to raise an error and catch it further down than silently swallow the message.

lib/hyrax/configuration.rb
```diff
@@ -542,10 +542,14 @@ def publisher
     attr_writer :translate_uri_to_id
 
     def translate_uri_to_id
-      @translate_uri_to_id ||= lambda do |uri|
-        baseparts = 2 + [(::Noid::Rails.config.template.gsub(/\.[rsz]/, '').length.to_f / 2).ceil, 4].min
-        uri.to_s.sub("#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}", '').split('/', baseparts).last
-      end
+      @translate_uri_to_id ||=
+        begin
+          baseparts = 2 + [(::Noid::Rails.config.template.gsub(/\.[rsz]/, '').length.to_f / 2).ceil, 4].min
```
no-reply, line 547: see #4634. this work should probably be rebased on that.

lib/wings/attribute_transformer.rb
```diff
@@ -5,8 +5,16 @@
 module Wings
   ##
   # Transform AF attributes to Valkyrie::Resource attributes representation.
-  #
   class AttributeTransformer
+    def self.for(obj)
```
no-reply, line 9: we need different behavior for some objects, a factory lets us switch without a lot of branching.

jeremyf (approved):

app/actors/hyrax/actors/file_actor.rb
```diff
@@ -48,6 +48,16 @@ def ==(other)
 
       private
 
+      ##
+      # Wraps the verisoning service with erro handling. if the service's
+      # create handler isn't implemented, we want to accept that quietly here.
+      def create_version(content, user)
+        Hyrax::VersioningService.create(content, user)
+      rescue NotImplementedError
```
elrayle, line 56: This is nice encapsulation.

app/models/hyrax/file_metadata.rb
```diff
@@ -40,10 +40,9 @@ def uri_for(use:)
 
     # all remaining attributes are on AF::File metadata_node unless otherwise noted
     attribute :label, ::Valkyrie::Types::Set
-    attribute :original_filename, ::Valkyrie::Types::Set
+    attribute :original_filename, ::Valkyrie::Types::String
```
elrayle, line 43: A good bit of this was based off the Figgy FileMetadata class.  They had [original_filename defined as a Set](https://github.com/pulibrary/figgy/blob/main/app/resources/nested_resources/file_metadata.rb#L14).  It makes sense to me to have this be a String.  I don't know if @tpendragon can comment on why it is a Set in Figgy.

app/services/hyrax/change_content_depositor_service.rb
```diff
@@ -31,11 +31,14 @@ def self.call_valkyrie(work, user, reset)
         work.permission_manager.acl.permissions = []
         work.permission_manager.acl.save
       end
+
       work.proxy_depositor = work.depositor
       apply_depositor_metadata(work, user)
-      work.file_sets.each do |f|
+
+      Hyrax.custom_queries.find_child_filesets(resource: work).each do |f|
```
elrayle, line 38: Nice change.

lib/hyrax/configuration.rb
```diff
@@ -561,6 +561,8 @@ def translate_id_to_uri
 
     attr_writer :resource_id_to_uri_transformer
     def resource_id_to_uri_transformer
+      Deprecation.warn('Use Hyrax.config.translate_uri_to_id instead.')
+
```
elrayle, line 565: The deprecation statement seems to have the direction of the translation backward.  Should be...
```
Deprecation.warn('Use Hyrax.config.translate_id_to_uri instead.')
```

As another note about this.  The code here was written to remove references to ActiveFedora.  It is an equivalent set of code to what is done in ActiveFedora.  So yes to the deprecation, but `Hyrax.config.translate_id_to_uri` should probably be updated to this code to remove references to ActiveFedora.

elrayle: Any changes here should also be reflected in lib/generators/hyrax/templates/config/initializers/hyrax.rb

elrayle: Probably need similar changes for `translate_uri_to_id`.

lib/generators/hyrax/templates/config/initializers/hyrax.rb
```diff
@@ -232,11 +232,6 @@
   # config.translate_uri_to_id = lambda do |id|
   #                                "#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}/#{Noid::Rails.treeify(id)}"
   #                              end
-  # config.resource_id_to_uri_transformer = lambda do |resource, base_url|
-  #                                           file_id = CGI.escape(resource.file_identifiers.first.to_s)
-  #                                           fs_id = CGI.escape(resource.file_set_id.to_s)
-  #                                           "#{base_url}#{::Noid::Rails.treeify(fs_id)}/files/#{file_id}"
-  #                                         end
```
elrayle, line 239: See comments for lib/hyrax/configuration.rb.

spec/wings/hydra/works/services/add_file_to_file_set_spec.rb
```diff
@@ -34,8 +34,7 @@
         expect(ids.first.to_s).to start_with "#{file_set.id}/files/"
 
         file_metadata = Hyrax.custom_queries.find_file_metadata_by(id: ids.first)
-        expect(file_metadata.content.first).to start_with('%PDF-1.3')
-        expect(file_metadata.mime_type.first).to eq pdf_mimetype
+        expect(file_metadata.mime_type).to eq pdf_mimetype
```
elrayle, line 37: It seems good to remove content from the file_metadata.  I'm wondering if the content is tested somewhere else.  I ask because there were lots of issues to work through related to content and the need to translate it back and forth between the Valkyrie FileMetadata object and the AF metadata_node.  It could be there are tests under the AF converter tests and I just haven't seen them yet.

elrayle (commented):
There is a lot happening in this PR.  I've added some comments and questions in the first part.  I'll continue to review as I have time to dig in deeper into the meaty changes in active_fedora_converter, find_file_metadata, file_metadata_builder, and file storage.

lib/hyrax/configuration.rb
no-reply: ah, i'm following this whole line better after these comments. thanks for this context.

question: since this is configuration for ActiveFedora, and for the Wings storage adapter that is designed for AF compatibility, is it okay to just leave this as-is and remove these methods entirely when Wings is removed?

Issue #4630, opened 2020-11-21
Document process for setting up Hyrax based app for development using Docker


JzGo:
### Descriptive summary

Detail steps required to start developing a Hyrax based app using the Hyrax Dockerfile and `docker-compose`

### Rationale

As a Hyrax based app developer, I would like to be able to set up all the required services within a containerized Docker environment to simplify and standardize my workflow with Hyrax applications

rjkati:
@jlhardes does the documentation in https://github.com/samvera/hyrax/blob/main/CONTAINERS.md satisfy this ticket?

jlhardes:
Yes! Those are the instructions for getting Hyrax up and running with Docker. Closing this issue.

PR #4631, opened 2020-11-21
Add dockerizing an app section

JzGo:
**Addresses #4630** 

Describes process for setting up a Hyrax application for a Docker based workflow utilizing the `samveralabs/hyrax` image and `docker-compose`

I did notice however that some adjustments are required to address solr errors so this PR is WIP atm


documentation/developing-your-hyrax-based-app.md
```diff
@@ -269,3 +270,158 @@ proxy_deposit:
 
 If both options exist, whichever option is set from the Administrative Dashboard
 will take precedence.
+
+## Dockerizing a Hyrax Based App for Development
+It is now possible develop a Hyrax based app using a Docker workflow. The process utilizes **docker-compose** to build and run the services necessary for Hyrax based applications.
+
+### Switch to Postgres
+The image that you will build your app's image on uses **Postgres**. If you have just created your Hyrax-based app, for example, you'll need to update your `Gemfile`
+	- `gem "sqlite3", "~> 1.3.0"` ---> `gem "pg"`
```
no-reply, line 279: it might be worth recommending the addition of `--database = postgresql` to the `rails new` command.

documentation/developing-your-hyrax-based-app.md
```diff
@@ -269,3 +270,158 @@ proxy_deposit:
...
+
+### Building the image
+Next you'll want to build the image that your Hyrax container will run from. For that, you will need to create a Dockerfile, build the image from it, and push it to a repository you can fetch from later.
+
+**Build image**
+- Run `cd <hyrax_app_name>`
+- Visit [samveralabs/hyrax](https://hub.docker.com/r/samveralabs/hyrax/tags) and grab the latest image **tag**
+- Run `echo "FROM samveralabs/hyrax:7<tag>" > Dockerfile` to create the Dockerfile
```
no-reply, line 287: we are going to need an image tagging strategy, and should update this as soon as we have something figured out. some input: https://cloud.google.com/solutions/best-practices-for-building-containers, https://stevelasker.blog/2018/03/01/docker-tagging-best-practices-for-tagging-and-versioning-docker-images/

in the meanwhile, i think the `7` here is a typo?
```suggestion
- Run `echo "FROM samveralabs/hyrax:<tag>" > Dockerfile` to create the Dockerfile
```

documentation/developing-your-hyrax-based-app.md
```diff
@@ -269,3 +270,158 @@ proxy_deposit:
...
+- Run `docker build .`
+
+**Push image to a repo (assumes dockerhub.io)**
```
no-reply, line 290: the push to a web-available registry is needed for the kubernetes/helm deployments, but i think it can be streamlined out here.

it's probably best in most cases to tag locally on build `docker build . -t my_app:dev`, and then use `my_app:dev` in place of `<sha-you-copied>` and `<your-image-location>`.

documentation/developing-your-hyrax-based-app.md
```diff
@@ -269,3 +270,158 @@ proxy_deposit:
...
+Since Hyrax relies on serveral different services to run, you'll need to set up a **docker-compose.yml** to tell Docker how to run all those services together.
+
+- Create a **docker-compose.yml** file your application root
+```
+version: '3.8'
+services:
+  app: &app
+    build:
+      context: .
+      target: hyrax-engine-dev
```
no-reply, line 305: does this work? if so, i think maybe it's getting overwritten by `image:` below?

in the hyrax setup, we specify a target because our `Dockerfile` has a multi-stage build, with multiple possible build targets. then we give an `image: ` name to tag the built image for reuse by `db-migrate:` below.

does it work if you try dropping this `target: ` line and replacing each appearance of `image: <your-image-location>` with `image: something-unique-to-this-file`? if so, it might be possible to drop the whole `**Build image**` step for development.

no-reply:
finally got to the bottom of the test failures in #4649. this should get rebased on that.

a few of the questions are outstanding, but i hope we can push this through early next week.

documentation/developing-your-hyrax-based-app.md
JzGo: That sounds like a reasonable default to me. I can't think of a lot of situations when someone would opt to use sqlite but that's from my limited experience. I **can** see some value in adding a mysql example. But that feels like another discussion.

PR #4634, opened 2020-11-23
make default translate_uri_to_id 2x faster

no-reply:
this code is called very frequently. the old implementation was slow in two
places:

first, it calculated `baseparts` for every call. this involved a memory
lookup (`Noid::Rails.config.template`) and several computations (most notably
`String#gsub`). since lambdas are closures, we can do this just once and have
the value (an Integer) passed around with the lambda. this maybe introduces a
cache invalidation problem (if `Noid::Rails.config` changes during execution,
we'll have the wrong number for `baseparts`). however, `Noid::Rails.config`
shouldn't change during execution; it seems safe for now to not worry about
invalidating this cache.

second, it ran `String#sub` on a long string match. `#split` on
`ActiveFedora.fedora.base_path` is more resilient (e.g. it's not protocol
dependent) and just a whole bunch faster.

in all, i saw a > 2x improvement on this code that's called, in some cases,
dozens of times for relatively simple AF operations.

Warming up --------------------------------------
       translate uri    35.111k i/100ms
       translate uri    356.748k (± 4.8%) i/s -      1.791M in   5.030930s

(i didn't capture the baseline, but it was about `14k` / `145k`)

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/indexers/hyrax/file_set_indexer_spec.rb
+++ b/spec/indexers/hyrax/file_set_indexer_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe Hyrax::FileSetIndexer do
 
   let(:file_set) do
     FileSet.new(
-      id: 'foo123',
+      id: 'foo12345',
       contributor: ['Mohammad'],
       creator: ['Allah'],
       title: ['The Work'],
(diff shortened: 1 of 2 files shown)


jeremyf (approved):

no-reply:
it's a bizarre coincidence that the failing test originally passed. it doing so was contingent on a specific number of slashes being in the hard coded example string :shrug:.

i'm going to fix the test the best way i know how, and we can move this through :100:

spec/indexers/hyrax/file_set_indexer_spec.rb
```diff
@@ -43,7 +43,7 @@
 
   let(:resource_version) do
     ActiveFedora::VersionsGraph::ResourceVersion.new.tap do |v|
-      v.uri = "http://test.example/rest/example/foo123/files/#{mock_file.id}/fcr:versions/version1"
+      v.uri = "#{file_set.uri}/files/#{mock_file.id}/fcr:versions/version1"
```
no-reply, line 46: using an actual id here, and not stubbing one that's deliberately shorter than the real implementation, fixes this test.

cjcolvar (approved):
This is great!  This seems like a candidate for backporting.

PR #4635, opened 2020-12-01
wings: remove unused MultiChecksum resource model class

no-reply:
this class isn't used. delete it.

@samvera/hyrax-code-reviewers


elrayle (approved):
It seems fine to remove this at this time since it is not used.  There are multiple uses in Figgy, so it is possible we may want to reintroduce this at a later time.

PR #4636, opened 2020-12-01
wings: deprecate FileMetadataBuilder

no-reply:
you can now use `storage_adapter.upload` to do all the work of
`FileMetadataBuilder`. Fedora creates the metadata node automatically, so it
just needs to be queried.

We may still want a Hyrax-layer abstraction for handling the assumption that
there will always be a `FileMetadata` model for every stored (pcdm) File. This
would need special handling for non-Wings adapters, but not for Wings.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -47,7 +47,6 @@ module Hyrax
       def update_content(file, relation = :original_file)
         IngestJob.perform_later(wrapper!(file: file, relation: relation), notification: true)
       end
-
       # @!endgroup
 
       # Adds the appropriate metadata, visibility and relationships to file_set
@@ -138,7 +137,8 @@ module Hyrax
       end
 
       def build_file_actor(relation)
-        file_actor_class.new(file_set, relation, user, use_valkyrie: use_valkyrie)
+        fs = use_valkyrie ? file_set.valkyrie_resource : file_set
+        file_actor_class.new(fs, relation, user, use_valkyrie: use_valkyrie)
       end
 
       # uses create! because object must be persisted to serialize for jobs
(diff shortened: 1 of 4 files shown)


spec/actors/hyrax/actors/file_actor_spec.rb
```diff
@@ -131,112 +130,56 @@ class FileSetWithExtras < FileSet
...
-        # TODO: WINGS - When #ingest_file works, these should be uncommented.
-        # expect(Hyrax::VersioningService).to receive(:create).with(Hyrax::FileMetadata, user)
-        # expect(Hyrax::VersioningService).to receive(:create).with(Hyrax::FileMetadata, user2)
-        # expect(CharacterizeJob).to receive(:perform_later)
-        allow(CharacterizeJob).to receive(:perform_later)
-        actor.ingest_file(io)
-        actor2.ingest_file(io2)
-      end
 
-      it 'has two versions' do
```
no-reply, line 208: i'm having trouble maintaining this pending test (this is a ton of code!) so proposing removing it until we get versioning support on the valkyrie storage adapter.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -138,7 +137,8 @@ def ability
       end
 
       def build_file_actor(relation)
-        file_actor_class.new(file_set, relation, user, use_valkyrie: use_valkyrie)
+        fs = use_valkyrie ? file_set.valkyrie_resource : file_set
+        file_actor_class.new(fs, relation, user, use_valkyrie: use_valkyrie)
```
elrayle, line 141: Other than testing, do we ever go through the actor stack with valkyrie resources?  We did when this was written.  But now that transactions are called for updating when the controller sees a resource, I'm wondering about the code in the actor stack that checks use_valkyrie.  Should it be deprecated pointing to the transactions as the alternative?

no-reply: deprecating seems okay, but i'm not sure i see any need. it's getting reworked now, with the removal of this Wings special handling, but i think it's unlikely to need attention again until we're ready to yank Wings entirely. there's every reason to imagine we'd deprecate/remove the whole actor stack by then.

jeremyf (approved):

PR #4638, opened 2020-12-02
wings: remove *ValkyrieBehavior model extensions 

no-reply:
these modules try to reimplement Hydra::PCDM and Hydra::Works directly on valkyrie models, injecting them at conversion time. bundling this logic, which is mostly concerned with querying, creates undesirable coupling between data representation and business concerns.

removing it should help clarify what the application needs to support to handle valkyrie objects natively.

```
convert a Monograph object to ActiveFedora
                         95.083  (± 8.4%) i/s -    477.000  in   5.053864s
convert a derived Valkyrie object back to ActiveFedora
                         47.188  (± 8.5%) i/s -    236.000  in   5.039478s
save a Valkyrie object
                          6.966  (±14.4%) i/s -     35.000  in   5.094816s
```

@samvera/hyrax-code-reviewers


no-reply:
to keep scope from creeping and reviews from becoming more complex, i'm submitting fixes for the issues this PR reveals as separate patches: https://github.com/samvera/hyrax/pull/4639

lib/wings/orm_converter.rb
```diff
@@ -48,15 +40,9 @@ def self.base_for(klass:)
     #   results in long methods
     def self.to_valkyrie_resource_class(klass:)
       relationship_keys = klass.respond_to?(:reflections) ? relationship_keys_for(reflections: klass.reflections) : []
-      relationship_keys.delete('member_ids')
```
no-reply, line 51: handling these keys like any other, now.

lib/wings/orm_converter.rb
```diff
@@ -48,15 +40,9 @@ def self.base_for(klass:)
...
-      relationship_keys.delete('member_of_collection_ids')
       reflection_id_keys = klass.respond_to?(:reflections) ? klass.reflections.keys.select { |k| k.to_s.end_with? '_id' } : []
 
       Class.new(base_for(klass: klass)) do
-        include Wings::CollectionBehavior if klass.included_modules.include?(Hyrax::CollectionBehavior)
```
no-reply, line 56: this is the biggest change here. no longer injecting Hydra::Works-like behavior into valkyrie resources at runtime. now relying on the `Hyrax::FileSet`, `Hyrax::PcdmCollection` and `Hyrax::Work` classes to provide all the model layer behavior we want.

app/services/hyrax/thumbnail_path_service.rb
```diff
@@ -22,6 +22,11 @@ def call(object)
 
       private
 
+      def audio?(thumb)
+        service = thumb.respond_to?(:audio?) ? thumb : Hyrax::FileSetTypeService.new(file_set: thumb)
```
no-reply, line 26: no longer adding `Hydra::Works::MimeTypes` to generated classes. a partial re-implementation is provided as `Hyrax::FileSetTypeService`

jeremyf (approved):

PR #4639, opened 2020-12-02
query file_sets with custom query/navigator in WorksControllerBehavior

no-reply:
use native valkyrie approach for checking if the work has any file_sets.

@samvera/hyrax-code-reviewers


elrayle (approved):
This looks good.  Thanks for the update.  Much cleaner to use the custom query.

PR #4641, opened 2020-12-02
separate af and valkyrie tests for AttachMembersActor

no-reply:
these tests shouldn't overlap.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/actors/hyrax/actors/attach_members_actor_spec.rb
+++ b/spec/actors/hyrax/actors/attach_members_actor_spec.rb
@@ -17,77 +17,79 @@ RSpec.describe Hyrax::Actors::AttachMembersActor do
   end
 
   describe "#update" do
-    before { work.ordered_members << existing_child_work }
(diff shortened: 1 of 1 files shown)


spec/actors/hyrax/actors/attach_members_actor_spec.rb
```diff
@@ -17,77 +17,79 @@
   end
 
   describe "#update" do
-    before { work.ordered_members << existing_child_work }
+    context 'when using an active_fedora work' do
```
no-reply, line 20: i'm not sure why the smart diff rendering is doing so much work here. the diff is:

  - this `context` is added beginning on L20 and wrapping all the tests that aren't already wrapped by the valkyrie context.
  - everything within is indented 2 spaces, and otherwise unchanged.

orangewolf (approved):
thanks for the comment - hard to read the diff, but the side by side and your comment made it pretty clear

PR #4645, opened 2020-12-04
use AccessControlList#permissions= in InheritPermissionsJob

no-reply:
avoid complicated diffing logic in favor of a wipe and replace strategy.

ACL has an internal ChangeSet, so manipulating it inline has a low cost.

a slight change to `ACL#permissions=` is included to make the interface for this kind of permission copying a bit cleaner.

@samvera/hyrax-code-reviewers


jeremyf (approved):
I love this!

PR #4648, opened 2020-12-05
extract and minimize valkyrie fileset attachment behavior, add to change_set transaction(s)

no-reply:
makes uploads passed to the controller lead to files attached to the work.
 
this is the simplest version of valkyrie driven fileset attachment that i could come up with
that passes the current tests.

i fear that passing the current tests may not be enough. but i think the
preferred strategy at this point is to get an end-to-end implementation from the
`WorksControllerBehavior` tests so we can avoid duplicating the byzantine
processes involved in FileSetActor, FileActor, AttachFileToWorkJob, etc...

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -187,8 +187,11 @@ module Hyrax
       else
         form = build_form
 
-        @curation_concern = form.validate(params[hash_key_for_curation_concern]) &&
-                            transactions['change_set.create_work'].call(form).value!
+        @curation_concern =
+          form.validate(params[hash_key_for_curation_concern]) &&
+          transactions['change_set.create_work']
+          .with_step_args('work_resource.add_file_sets' => { uploaded_files: uploaded_files })
+          .call(form).value!
       end
     end
 
@@ -199,8 +202,11 @@ module Hyrax
       else
         form = build_form
 
-        @curation_concern = form.validate(params[hash_key_for_curation_concern]) &&
-                            transactions['change_set.update_work'].call(form).value!
+        @curation_concern =
+          form.validate(params[hash_key_for_curation_concern]) &&
+          transactions['change_set.update_work']
+          .with_step_args('work_resource.add_file_sets' => { uploaded_files: uploaded_files })
+          .call(form).value!
       end
     end
 
@@ -411,6 +417,10 @@ module Hyrax
       else
         Hyrax.custom_queries.find_child_fileset_ids(resource: curation_concern).any?
       end
+    end
+
+    def uploaded_files
+      UploadedFile.find(params.fetch(:uploaded_files, []))
     end
   end
 end
--- a/app/jobs/attach_files_to_work_job.rb
+++ b/app/jobs/attach_files_to_work_job.rb
@@ -10,7 +10,8 @@ class AttachFilesToWorkJob < Hyrax::ApplicationJob
     when ActiveFedora::Base
       perform_af(work, uploaded_files, work_attributes)
     else
-      Hyrax::WorkUploadsHandler.new(work: work).add(files: uploaded_files).attach
+      Hyrax::WorkUploadsHandler.new(work: work).add(files: uploaded_files).attach ||
+        raise("Could not complete AttachFilesToWorkJob. Some of these are probably in an undesirable state: #{uploaded_files}")
     end
   end
 
--- a/app/jobs/attach_files_to_work_job.rb
+++ b/app/jobs/attach_files_to_work_job.rb
@@ -9,8 +9,8 @@ class AttachFilesToWorkJob < Hyrax::ApplicationJob
     case work
     when ActiveFedora::Base
       perform_af(work, uploaded_files, work_attributes)
-    when Valkyrie::Resource
-      perform_valkyrie(work, uploaded_files, work_attributes)
+    else
+      Hyrax::WorkUploadsHandler.new(work: work).add(files: uploaded_files).attach
     end
   end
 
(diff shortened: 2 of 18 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -363,7 +369,8 @@ def after_create_response
     def after_update_response
       if concern_has_file_sets?
         return redirect_to hyrax.confirm_access_permission_path(curation_concern) if permissions_changed?
-        return redirect_to main_app.confirm_hyrax_permission_path(curation_concern) if curation_concern.visibility_changed?
+        # is this codepath unreachable? how can visibility have changed if permissions did not?!
```
no-reply, line 372: looking at this a bit closer, i'm pretty sure this is only reachable in two cases:

  - when the test suite has stubbed `visibility_changed?` (which seems to be the case whenever it's actually hit during testing, i'm running the full suite to double check this); and
  - when the `visibility_will_change!` flag has been set, but the permissions (and visibility has not actually changed)
    - e.g. if i set `visibility = 'public'; visibility = 'restricted'` on a work that started as "restricted", i'll have set the flag, but `premissions_changed?` will be false. i don't think we want to be redirecting in this case, and i strongly suspect this code, and the whole controller action/view that follow from it have been unreachable since the above line was added in 2017.

no-reply: a fix for this smaller issue is now in at #4650

jeremyf: I would imagine so; My recollection is of a one time quasi-separation of visibility and permission, but really they are conflated so this likely was added. I think your assessment is spot on and accurate.

app/jobs/attach_files_to_work_job.rb
```diff
@@ -9,8 +9,8 @@ def perform(work, uploaded_files, **work_attributes)
     case work
     when ActiveFedora::Base
       perform_af(work, uploaded_files, work_attributes)
-    when Valkyrie::Resource
-      perform_valkyrie(work, uploaded_files, work_attributes)
+    else
+      Hyrax::WorkUploadsHandler.new(work: work).add(files: uploaded_files).attach
```
jeremyf, line 13: I like extracting the class; It certainly helps separate the concerns.

app/jobs/inherit_permissions_job.rb
```diff
@@ -46,12 +46,11 @@ def af_perform(work)
...
   def valkyrie_perform(work)
-    work_permissions = Hyrax::AccessControlList.new(resource: work).permissions
+    work_acl = Hyrax::AccessControlList.new(resource: work)
 
     file_sets_for(work).each do |file_set|
-      acl = Hyrax::AccessControlList.new(resource: file_set)
-      acl.permissions = work_permissions
-      acl.save
+      Hyrax::AccessControlList
+        .copy_permissions(source: work_acl, target: file_set)
```
jeremyf, line 53: I like this idiom. It's clear in intention.

app/services/hyrax/access_control_list.rb
```diff
@@ -53,6 +70,19 @@ def initialize(resource:, persister: Hyrax.persister, query_service: Hyrax.query
...
     end
 
+    ##
+    # Copy and save permissions from source to target
+    #
+    # @param [Valkyrie::Resource, Hyrax::AccessControlList] source
+    # @param [Valkyrie::Resource, Hyrax::AccessControlList] target
+    #
+    # @return [Hyrax::AccessControlList] an acl for `target` with the updated permissions
+    def self.copy_permissions(source:, target:)
```
jeremyf, line 80: 👏

lib/hyrax/transactions/container.rb
```diff
@@ -88,7 +89,13 @@ class Container
...
       end
 
-      namespace 'work' do |ops|
+      namespace 'work_resource' do |ops| # valkyrie works
+        ops.register 'add_file_sets' do
+          Steps::AddFileSets.new
+        end
+      end
+
+      namespace 'work' do |ops| # legacy AF works
```
jeremyf, line 98: I love mentally expanding "legacy AF" into the "kids these days" vernacular.  Great comment!

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -361,10 +367,8 @@ def after_create_response
     end
 
     def after_update_response
-      if concern_has_file_sets?
-        return redirect_to hyrax.confirm_access_permission_path(curation_concern) if permissions_changed?
-        return redirect_to main_app.confirm_hyrax_permission_path(curation_concern) if curation_concern.visibility_changed?
-      end
+      return redirect_to hyrax.confirm_access_permission_path(curation_concern) if permissions_changed? && concern_has_file_sets?
```
jeremyf, line 370: 👍

spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -479,25 +471,20 @@
       end
 
       it "can update file membership" do
-        patch :update, params: { id: work, generic_work: { ordered_member_ids: ['foo_123'] } }
-        expect(actor).to have_received(:update).with(Hyrax::Actors::Environment) do |env|
-          expect(env.attributes).to eq("ordered_member_ids" => ['foo_123'],
-                                       "remote_files" => [],
-                                       "uploaded_files" => [])
-        end
+        patch :update, params: { id: work, generic_work: { ordered_member_ids: [file_set.id] } }
```
jeremyf, line 474: Thanks for removing a somewhat magic number!

spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -479,25 +471,20 @@
...
+        expect(work.reload.ordered_members.to_a).to contain_exactly(file_set)
       end
 
       describe 'changing rights' do
-        let(:visibility_changed) { true }
-        let(:actor) { double(update: true) }
-
         context 'when the work has file sets attached' do
           before do
+            allow(GenericWork).to receive(:find).and_return(work)
```
jeremyf, line 481: Good short circuit of a potentially expensive find.

spec/controllers/hyrax/generic_works_controller_spec.rb
```diff
@@ -512,8 +499,10 @@
       describe 'update failed' do
         let(:actor) { double(update: false) }
 
+        before { allow(Hyrax::CurationConcern).to receive(:actor).and_return(actor) }
+
         it 'renders the form' do
-          patch :update, params: { id: work, generic_work: {} }
+          patch :update, params: { id: work.id, generic_work: {} }
```
jeremyf, line 505: Good catch

spec/services/hyrax/work_uploads_handler_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      it 'assigns FileSet ids synchronously' do
+        expect { service.attach }
+          .to change { work.member_ids }
+          .to contain_exactly(an_instance_of(Valkyrie::ID),
+                              an_instance_of(Valkyrie::ID),
+                              an_instance_of(Valkyrie::ID))
+      end
+
+      it 'creates persisted filesets' do
```
jeremyf, line 27: It's a small ask, but could lines 27 and 32 be merged? I don't know if `:persisted?` is an attribute.  I think these two make sense as one singular expectation. (Not a blocker, but something to consider)

spec/services/hyrax/work_uploads_handler_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    context 'with existing file_sets' do
+      let(:work) { FactoryBot.valkyrie_create(:hyrax_work, :public, :with_member_file_sets) }
+
+      it 'appends the new file sets' do
+        first_id, second_id = work.member_ids
+
+        service.add(files: uploads).attach
+        expect(work).to have_file_set_members(have_attributes(id: first_id),
```
jeremyf, line 56: It looks like three is the magic number for `:with_member_file_sets`; I don't know if that's something to consider parameterizing?

jeremyf (approved):

spec/services/hyrax/work_uploads_handler_spec.rb
no-reply: i think i considered this fleetingly, wasn't sure if there was a pleasant way to get `have_attributes` to do this work, and decided i liked having separate tests/rspec messages for "did we save them" vs. "did we set the metadata". i also recall squashing visibility into L32, after initial writing it as yet another test case.

i guess our options for squashing are:

```ruby
have_attributes(title: ['image.jp2'], depositor: an_instance_of(String), visibility: 'open', "persisted?": true) # i'm not 100% sure this will work, but it seems plausible
# or
have_attributes(title: ['image.jp2'], depositor: an_instance_of(String), visibility: 'open', new_record: false)
```

if your vote is for one of the above over the current L27 test, i'll consider that decisive.

these tests _are_ a bit slow, as written, but we could add a `valkyrie_adapter: :test_adapter` flag and speed them up.

spec/services/hyrax/work_uploads_handler_spec.rb
no-reply: yeah. this could probably use some refinement... :thinking:

no-reply: the 3 are the ones we're adding here, while `first_id` and `second_id` are the `:with_member_file_sets` created members. but yeah, it's both confusing and fragile.

what we really want is probably `expect { service.add(files: uploads).attach }.to add_file_set_members.count(3)` or something like that?

spec/services/hyrax/work_uploads_handler_spec.rb
no-reply: i ended up making this use `:test_adapter` anyway. testing against generic Valkyrie is a benefit, so the speed factor is less of one now.

jeremyf (approved):

PR #4649, opened 2020-12-05
make build succeed on ci

no-reply:
benchmark introduced a problem with builds in CI, if we're in CI, we just don't need to load any of this.

@samvera/hyrax-code-reviewers


no-reply:
using admin privileges to merge this, since the issue is preventing all other builds from passing.

PR #4650, opened 2020-12-05
stop stubbing to reach otherwise unreachable code path; deprecate

no-reply:
since `#visibility` is a projection on `#permissions` we never need to confirm
propogation of visibility changes if permissions haven't changed. the stub
created an artificial scenario where the work's `visibility_will_change!` flag
is set, even though we've never changed the permissions. fix the test to edit
visibility through user parameters, then remove the unreachable/unreached code
path. deprecate the related controller action.

the only other way this artificial scenario is achievable is by editing the
visibility, and then editing it back. we don't want to confirm edits in this
case, since neither permissions nor vizibility have really changed.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -361,10 +361,8 @@ module Hyrax
     end
 
     def after_update_response
-      if concern_has_file_sets?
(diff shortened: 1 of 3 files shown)


jeremyf (dismissed):

no-reply:
sorry. had to kick this for silly syntax reasons.

PR #4654, opened 2020-12-07
support valkyrie in WorkflowFactory

no-reply:
generalizes Sipity workflow initialization to work with Valkyrie::Resource
models.

this doesn't pretend to solve workflows for Valkyrie in total, but does ensure
existing expectations for the `WorkflowFactory` initialization process are
portable.

the main obvious gap is that `Hyrax::Workflow::WorkflowActionService` has
a (undesirable?) dependency on `ActiveFedora::Base#update_index`. what is it
counting on being indexed? i'm not yet sure; but it's unlikely we've addressed
it in the valkyrie indexers. this will need some attention as we continue to
work on Valkyrie workflows.

replace `InitializeWorkflowActor` with a listener. compare: #4653.

*breaking change*: this DEPRECATES `InitializeWorkflowActor` and REMOVES it from
the default actor stack. while the replacement functionality should be an
improvement over the older implementation, and is designed to respect certain
customization (custom workflow factories set via the existing class
configuration are retained), application-side action is required if a custom
actor is used.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/initialize_workflow_actor.rb
+++ b/app/actors/hyrax/actors/initialize_workflow_actor.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 module Hyrax
   module Actors
+    ##
+    # @deprecated
+    #
     # Responsible for generating the workflow for the given curation_concern.
     # Done through direct collaboration with the configured Hyrax::Actors::InitializeWorkflowActor.workflow_factory
     #
@@ -13,6 +16,7 @@ module Hyrax
       # @param [Hyrax::Actors::Environment] env
       # @return [Boolean] true if create was successful
       def create(env)
+        Deprecation.warn('Use Hyrax::Listeners::WorkflowListener instead.')
         next_actor.create(env) && create_workflow(env)
       end
 
@@ -20,6 +24,10 @@ module Hyrax
 
       # @return [TrueClass]
       def create_workflow(env)
+        # if the entity exists, this is already initialized
+        Sipity::Entity(env.curation_concern)
+        true
+      rescue Sipity::ConversionError
         workflow_factory.create(env.curation_concern, env.attributes, env.user)
       end
     end
--- a/app/models/sipity.rb
+++ b/app/models/sipity.rb
@@ -83,6 +83,10 @@ module Sipity
   end
   module_function :WorkflowState
 
+  ##
+  # A parent error class for all workflow errors caused by bad state
+  class StateError < RuntimeError; end
+
   class ConversionError < PowerConverter::ConversionError
     def initialize(value, **options)
       options[:scope] ||= nil
--- a/app/models/sipity.rb
+++ b/app/models/sipity.rb
@@ -26,6 +26,8 @@ module Sipity
                Entity(input.model)
              when Sipity::Comment
                Entity(input.entity)
+             when Valkyrie::Resource
+               Entity(Hyrax::GlobalID(input))
              else
                Entity(input.to_global_id) if input.respond_to?(:to_global_id)
              end
--- a/app/models/sipity/workflow.rb
+++ b/app/models/sipity/workflow.rb
@@ -30,7 +30,7 @@ module Sipity
       workflows = Sipity::Workflow.arel_table
       Sipity::Workflow.where(active: true).where(
         workflows[:permission_template_id].in(
-          templates.project(templates[:id]).where(templates[:source_id].eq(admin_set_id))
+          templates.project(templates[:id]).where(templates[:source_id].eq(admin_set_id.to_s))
         )
       ).first!
     end
--- a/app/models/sipity/workflow.rb
+++ b/app/models/sipity/workflow.rb
@@ -33,6 +33,8 @@ module Sipity
           templates.project(templates[:id]).where(templates[:source_id].eq(admin_set_id.to_s))
         )
       ).first!
+    rescue ActiveRecord::RecordNotFound => err
+      raise NoActiveWorkflowError, err.message
     end
 
     # @api public
@@ -45,9 +47,10 @@ module Sipity
     # @param workflow_name [String] The name of the workflow within the given permission template that should be activated
     # @return [Sipity::Workflow] active workflow
     # @raise [ActiveRecord::RecordNotFound] When we have a mismatch on permission template and workflow id or workflow name
-    # @raise [RuntimeError] When you don't specify a workflow_id or workflow_name
+    # @raise [ArgumentError] When you don't specify a workflow_id or workflow_name
     def self.activate!(permission_template:, workflow_id: nil, workflow_name: nil)
-      raise "You must specify a workflow_id or workflow_name to activate!" if workflow_id.blank? && workflow_name.blank?
+      raise(ArgumentError, "You must specify a workflow_id or workflow_name to activate!") if
+        workflow_id.blank? && workflow_name.blank?
       workflow_to_activate = Sipity::Workflow.find_by!({ permission_template: permission_template, id: workflow_id, name: workflow_name }.compact)
       active_workflow = Sipity::Workflow.where(permission_template: permission_template, active: true)
       return workflow_to_activate if workflow_to_activate == active_workflow.first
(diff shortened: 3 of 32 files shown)


app/services/hyrax/workflow/workflow_action_service.rb
```diff
@@ -20,7 +20,7 @@ def run
         comment = create_sipity_comment
         handle_sipity_notifications(comment: comment)
         handle_additional_sipity_workflow_action_processing(comment: comment)
-        subject.work.update_index # So that the new actions and state are written into solr.
+        subject.work.try(:update_index) # So that the new actions and state are written into solr.
```
no-reply, line 23: this is a hard `ActiveFedora` dependency. i wasn't ready to chase it down, but i think we should extract it.

one issue for me taking this on is that, where i am in this process right now, i don't know what `WorkflowActionService` *does* or why it wants to reindex the model.

no-reply: cc: @jeremyf

jeremyf: Ugh, not well documented.  I'm at the end of my evening, but may have some insights.

Sipity has the concept of callbacks.  When you take an action it may fire functions.  Those functions can do _anything_; Which would be why we might need to update the index.

(That's the proverbial straw I have at this point)

jeremyf: At this point, I'm at 🍻 o'clock.  It would be reasonable to move the `subject.update_index` into the `ActionTakenService`.  After all, if the ActionTakenService does something, then _it_ should be the one to update the index.

I think from a responsibility stand-point if I generate a comment, I shouldn't need to update the index. (But are we storing comments in the index and rendering them there?)

no-reply: oh, that makes sense.

we need this index updated because we need this index to be real time, and `handle_additional_sipity_workflow_action_processing` might have changed metadata or other indexed (non-sipity) values?

is your proposal essentially that, here, we plead ignorance to all of that and expect actions to handle it?

no-reply: notably, https://github.com/samvera/hyrax/blob/130c4e600318a9194725b35dd0fd6e19e5108dd9/app/services/hyrax/workflow/action_taken_service.rb#L35 already saves the target, which needs to be refactored for valkyrie itself, but also should be reindexing.

wondering if this call is already unneeded?

jeremyf: That certainly looks like we need the call; The issue would be "Are we expecting comments in the SOLR document?"

no-reply: seems pretty weird. it seems like it would be preferable if the workflows didn't know anything about object persistence or indexing

jeremyf: It's an odd thing.  A state change of an object might impact persistence and indexing; However, I think this should be uni-directional.  That is to say, a state change can write to persistence via some semblance of a command.

It highlights that we have traditionally overloaded the index with TOO much data for the variety of purposes.

no-reply:
i'm tapping out of this one for now; marking it WIP. this latest round is very promising, but it looks like we have to make some deeper dives into Workflow to get it over the line from here.

no-reply:
the remaining test failures are intermittent. it looks a lot like some test somewhere is setting up a workflow and then leaving it behind for future tests to run.

in at least some of the tests failing for this reason the DB, repository, and index are all cleared. what else could be holding onto workflow state?

no-reply:
well, i feel silly about how many tries this took, but i think it's ready!

app/actors/hyrax/actors/initialize_workflow_actor.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 module Hyrax
   module Actors
+    ##
+    # @deprecated
```
jeremyf, line 5: Beautiful!

app/services/hyrax/default_middleware_stack.rb
```diff
@@ -46,9 +46,6 @@ def self.build_stack
 
         # Persist the metadata changes on the resource
         middleware.use Hyrax::Actors::ModelActor
-
-        # Start the workflow for this work
```
jeremyf, line 50: Yay!

lib/hyrax.rb
```diff
@@ -40,6 +40,18 @@ module Hyrax
...
 
+  ##
+  # @return [GlobalID]
+  # @see https://github.com/rails/globalid
+  def self.GlobalID(input) # rubocop:disable Naming/MethodName
+    case input
+    when Valkyrie::Resource
+      ValkyrieGlobalIdProxy.new(resource: input).to_global_id
+    else
+      input.to_global_id if input.respond_to?(:to_global_id)
```
jeremyf, line 51: Do we need an exception raised? We'll be returning `nil` if the input doesn't respond to `:to_global_id`.

jeremyf (approved):
I left a comment that's not a blocker (as I want this code shipped).  However, it may warrant another change.  I leave that to @no-reply's discretion.

lib/hyrax.rb
no-reply: i actually tried to remove this, and it was a quagmire involving `PowerConverter` backward compatibility. 

what i'd really like to do in the `else` case is return a `NullGlobalID` but i think that would involve touching pretty much every part of `Hyrax::Workflow` and `Sipity`.

jeremyf: Sounds like we ship it!

app/models/sipity.rb
```diff
@@ -81,6 +83,10 @@ def WorkflowState(input, workflow, &block) # rubocop:disable Naming/MethodName
   end
   module_function :WorkflowState
 
+  ##
+  # A parent error class for all workflow errors caused by bad state
+  class StateError < RuntimeError; end
```
no-reply, line 88: @jeremyf i was wondering if you had broader thoughts about sipity error handling/transactionality.

one thing that kept popping up as i was debugging proposed changes is that failed Sipity operations (admittedly, they normally failed because i broke something important) are rarely retry-able. a second attempt will fail as violating a uniqueness constraint for some sipity DB row. the errors raised are pretty much always some kind of ActiveRecord error.

the `StateError` changes included here were what i needed to get reasonably smooth handling in this PR, but it might be nice to do a once over on sipity error handling in general.

also, is it reasonable to try to wrap high-level operations (e.g. `::Hyrax::Workflow::WorkflowFactory.create`) in an ActiveRecord transaction?

app/models/sipity/workflow.rb
```diff
@@ -86,5 +104,7 @@ def remove_workflow_responsibilities(role, allowed_agents)
       wf_role = Sipity::WorkflowRole.find_by(workflow: self, role_id: role)
       wf_role.workflow_responsibilities.where.not(agent: allowed_agents).destroy_all
     end
+
+    class NoActiveWorkflowError < StateError; end
```
no-reply, line 108: oh, i guess i added this error type, too!

app/services/hyrax/workflow/grant_edit_to_depositor.rb
```diff
@@ -7,7 +7,9 @@ module GrantEditToDepositor
       # @param [#read_users=, #read_users] target (likely an ActiveRecord::Base) to which we are adding edit_users for the depositor
       # @return void
       def self.call(target:, **)
-        target.edit_users += [target.depositor]
+        return true unless target.try(:depositor)
+        target.edit_users = target.edit_users.to_a + Array.wrap(target.depositor) # += works in Ruby 2.6+
```
no-reply, line 11: i had to do this hack, because `Enumerator+=` isn't around until Ruby 2.6

app/services/hyrax/workflow/workflow_factory.rb
```diff
@@ -40,16 +40,17 @@ def initialize(work, attributes, user)
       # The Sipity::Entity acts as a proxy to a work within a workflow
       # @return [TrueClass]
       def create
+        action = find_deposit_action
```
no-reply, line 43: refactor to avoid querying this twice

lib/hyrax/transactions/steps/save.rb
```diff
@@ -37,6 +37,7 @@ def call(change_set, user: nil)
           saved.permission_manager.acl.permissions = unsaved.permission_manager.acl.permissions if
             unsaved.respond_to?(:permission_manager)
 
+          user ||= ::User.find_by_user_key(saved.depositor)
```
no-reply, line 40: it's increasingly important that we report some user as having been responsible for published events. this infers that it was the `depositor` unless we provided a user explicitly with `step_args`.

jeremyf:
For those curious, @no-reply spoke to some of the driving impetus behind https://wiki.lyrasis.org/display/samvera/Samvera+Tech+Call+2020-12-09

app/models/sipity.rb
jeremyf: > Is it reasonable to try to wrap high-level operations (e.g. ::Hyrax::Workflow::WorkflowFactory.create) in an ActiveRecord transaction?

Yes!

jeremyf: > one thing that kept popping up as i was debugging proposed changes is that failed Sipity operations (admittedly, they normally failed because i broke something important) are rarely retry-able. a second attempt will fail as violating a uniqueness constraint for some sipity DB row. the errors raised are pretty much always some kind of ActiveRecord error.

> the StateError changes included here were what i needed to get reasonably smooth handling in this PR, but it might be nice to do a once over on sipity error handling in general.

I think much of the errors come from a traditional 🤷 treatment of transactions.

A once over of error handling would be reasonable.  I'm at the over-promised stage of things.  I believe I still owe a read through/documentation of the Sipity workflows.

PR #4656, opened 2020-12-09
(Backport of 4634) make default translate_uri_to_id 2x faster

bkeese:
make default translate_uri_to_id 2x faster

this code is called very frequently. the old implementation was slow in two
places:

first, it calculated `baseparts` for every call. this inloved a memory
lookup (`Noid::Rails.config.template`) and several computations (most notably
`String#gsub`). since lambdas are closures, we can do this just once and have
the value (an Integer) passed around with the lambda. this maybe introduces a
cache invalidation problem (if `Noid::Rails.config` changes during execution,
we'll have the wrong number for `baseparts`). however, `Noid::Rails.config`
shouldn't change during execution; it seems safe for now to not worry about
invalidating this cache.

second, it ran `String#sub` on a long string match. `#split` on
`ActiveFedora.fedora.base_path` is more resilient (e.g. it's not protocol
dependent) and just a whole bunch faster.

in all, i saw a > 2x improvement on this code that's called, in some cases,
dozens of times for relatively simple AF operations.

Warming up --------------------------------------
       translate uri    35.111k i/100ms
       translate uri    356.748k (± 4.8%) i/s -      1.791M in   5.030930s

``` ruby
class PostsController
  def index
    respond_with Post.limit(10)
  end
end
```

*
*
*

*
*
*

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -479,10 +479,14 @@ module Hyrax
     attr_writer :translate_uri_to_id
 
     def translate_uri_to_id
-      @translate_uri_to_id ||= lambda do |uri|
-        baseparts = 2 + [(::Noid::Rails.config.template.gsub(/\.[rsz]/, '').length.to_f / 2).ceil, 4].min
-        uri.to_s.sub("#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}", '').split('/', baseparts).last
-      end
+      @translate_uri_to_id ||=
+        begin
+          baseparts = 2 + [(::Noid::Rails.config.template.gsub(/\.[rsz]/, '').length.to_f / 2).ceil, 4].min
+
+          lambda do |uri|
+            uri.to_s.split(ActiveFedora.fedora.base_path).last.split('/', baseparts).last
+          end
+        end
     end
 
     attr_writer :translate_id_to_uri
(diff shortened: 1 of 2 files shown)


jeremyf (approved):

no-reply:
the test that fails is aggressively stubbed. there was a commit somewhere in the mainline, just before the backported one, that patches it.

bkeese:
Sorry. I can't find that commit. Which failing test?

.regen
```diff
@@ -1 +1 @@
-3
\ No newline at end of file
+4
```
jeremyf, line 1: @bkeese Perhaps add a non-sensical suffix (`4blerg`).  Perhaps CircleCI has a regen 4 cached.  I know that in the main branch we're up to 32.

hyrax.gemspec
```diff
@@ -50,7 +50,7 @@ SUMMARY
   spec.add_dependency 'font-awesome-rails', '~> 4.2'
   spec.add_dependency 'hydra-derivatives', '~> 3.3'
   spec.add_dependency 'hydra-editor', '>= 3.3', '< 6.0'
-  spec.add_dependency 'hydra-head', '>= 10.6.1'
+  spec.add_dependency 'hydra-head', '>= 10.6.1', '< 12'
```
no-reply, line 53: should we do this separately

bkeese: Yeah. Good thought. I'll split this into two PRs.

cjcolvar (approved):

PR #4662, opened 2020-12-12
Adding @since docs for Hyrax::Publisher events

jeremyf:
This change adds documentation regarding the specific published events.
I spent a few hours poking around the yard `@!macro` and `@!attribute`
behavior to settle on this.  (Some of the documented behavior doesn't
quite work, and I wonder if its related to the two mixins).

Regardless the goal of this documentation is to help bring in line the
versions in which we started publishing each event.  These `@since`
declarations align with [the corresponding wiki entry][1]

Attached to this commit's associated pull request will be a
screen-capture of how this commit modifies the yard docs for `Hyrax::Publisher`.

Screen Shot 2020-12-12 at 9 22 04 AM

Note there are two sections: The "Registered Events" group and the Instance Attribute Details (which includes the @since declaration).

[1]:https://github.com/samvera/hyrax/wiki/Hyrax's-Event-Bus-(Hyrax::Publisher)

 In addition, I'm adding some emojies to the documentation, as I hope it draws the eye of future maintainers to say "Hmm, emojiis, that seems odd, I better check this out."

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/publisher.rb
+++ b/lib/hyrax/publisher.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # frozen_string_literal: true
 
 module Hyrax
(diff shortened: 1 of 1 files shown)


lib/hyrax/publisher.rb
```diff
@@ -79,18 +80,54 @@ class Publisher
     include Singleton
     include Dry::Events::Publisher[:hyrax]
 
+    # @!group Registered Events
+    #
+    # 🛑 👀 ❓
```
no-reply, line 85: :clap:

no-reply:
this is great!

no-reply (approved):

Issue #4572, opened 2020-11-06
Universal Viewer stops working after 2.8.0 --> 2.9.0 upgrade


gwiedeman:
### Descriptive summary

* Universal Viewer appears to work just fine in [Hyrax 2.8.0](https://archives.albany.edu/concern/images/6t053p39d).

* Changing the Hyrax version in the Gemfile to 2.9.0 and running `bundle update hyrax` causes Universal Viewer to 
![hyraxIssue](https://user-images.githubusercontent.com/10030353/98396542-de49e780-202b-11eb-9157-8f1f06aaecb7.png)

* In looking at [Troubleshooting RIIIF](https://samvera.github.io/troubleshooting_riiif.html), it seems like `iiif_info_url_builder` is returning the wrong file ID and Fedora is returning an empty text file instead of the file object.

```
2.6.5 :015 > Riiif::Engine.routes.url_helpers.info_url(fs.files.first.id, host: "http://localhost:3000")
 => "http://localhost:3000/images/2f75rg51j%2Ffiles%2Fd1420cd8-a605-4ecd-8197-9e0f0862df5f/info.json"
```
* This does return [a manifest](https://archives.albany.edu/images/2f75rg51j%2Ffiles%2Fd1420cd8-a605-4ecd-8197-9e0f0862df5f/info.json), but `iiif_image_url_builder` then returns an incorrect URL for the file:

```
2.6.5 :016 > Riiif::Engine.routes.url_helpers.image_url(fs.files.first.id, host: "http://localhost:3000", size: "100,")
 => "http://localhost:3000/images/2f75rg51j%2Ffiles%2Fd1420cd8-a605-4ecd-8197-9e0f0862df5f/full/100,/0/default.jpg"
```
* [This URL](https://archives.albany.edu/images/2f75rg51j%2Ffiles%2Fd1420cd8-a605-4ecd-8197-9e0f0862df5f/full/100,/0/default.jpg) returns a 500.

[...]

PR #4663, opened 2020-12-12
pass `base_url`, not just `hostname` through to manifest generation

no-reply:
this change realigns with older Hyrax behavior that used the protocol
from the current controller's request as part of the argument to the IIIF image
URI generator.

related to #4572.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -147,7 +147,7 @@ module Hyrax
 
     def iiif_manifest_presenter
       IiifManifestPresenter.new(search_result_document(id: params[:id])).tap do |p|
-        p.hostname = request.hostname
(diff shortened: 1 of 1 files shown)


jeremyf (dismissed):

no-reply:
this is even simpler than i originally thought. i think it will fix `nurax-dev`'s UV issues, but i'm uncertain whether nurax is still deploying, given Travis CI's issues. cc: @maxkadel @mark-dce

orangewolf:
This is likely a good change and moves you in the right direction. I would still recommend double checking the nginx or Apache headers or this may least to a forgery error (request.base_url does not match http origin or something like that). See https://serverok.in/nginx-rails-origin-header-didnt-match for the most concise of the thousand or so people running in to this ;-)

orangewolf (approved):

jeremyf (approved):

PR #4668, opened 2020-12-14
begin rework of Hyrax::Configuration to support readable YARD docs

no-reply:
these configuration options are historically under-documented. this begins the
process of organizing them so the YARD docs can be a usable source for folks
trying to configure Hyrax.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -3,6 +3,61 @@ require 'hyrax/role_registry'
 require 'samvera/nesting_indexer'
 
 module Hyrax
+  ##
+  # Handles configuration for the Hyrax engine.
+  #
+  # This class provides a series of accessors for setting and retrieving global
+  # engine options. For convenient reference, options are grouped into the
+  # following functional areas:
+  #
+  # - Groups
+  # - Identifiers
+  # - IIIF
+  # - Local Storage
+  # - System Dependencies
+  # - Theme
+  # - Valkyrie
+  #
+  # == Groups
+  #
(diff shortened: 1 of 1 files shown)


lib/hyrax/configuration.rb
```diff
@@ -83,41 +158,239 @@ def registered_user_group_name
     end
 
     # @!endgroup
+    # @!group Identifier Minting
+
+    attr_writer :enable_noids
+    def enable_noids?
+      return @enable_noids unless @enable_noids.nil?
+      @enable_noids = true
```
jeremyf, line 166: This seems to be an unexpected change in state.

lib/hyrax/configuration.rb
```diff
@@ -83,41 +158,239 @@ def registered_user_group_name
...
+    # @see Hyrax::Configuration#iiif_info_url_builder
+    # @see Hyrax::Configuration#iiif_image_compliance_level_uri
+    # @see Hyrax::Configuration#iiif_image_size_default
+    #
+    # @note Default is false
+    #
+    # @return [Boolean] true to enable, false to disable
+    def iiif_image_server?
+      return @iiif_image_server unless @iiif_image_server.nil?
+      @iiif_image_server = false
```
jeremyf, line 214: Again, an unexpected change in state.  I think this is okay.

lib/hyrax/configuration.rb
no-reply: this code is just moved from below. not changed.

jeremyf (approved):

PR #4671, opened 2020-12-15
more YARD fixups and expansions

no-reply:
trying to make the docs look nice

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/adapters/nesting_index_adapter.rb
+++ b/app/services/hyrax/adapters/nesting_index_adapter.rb
@@ -6,7 +6,10 @@ module Hyrax
       LIMITED_REINDEX = "limited"
 
       # @!group Providing interface for a Samvera::NestingIndexer::Adapter
+
+      ##
       # @api public
+      #
       # @param id [String]
       # @return Samvera::NestingIndexer::Document::PreservationDocument
       def self.find_preservation_document_by(id:)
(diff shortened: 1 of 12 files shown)


jeremyf (dismissed):

jeremyf:
I'm not certain about the broken test, but am doing a bit of sleuthing.

no-reply:
oh sorry. i thought i already pushed the fix for that

app/services/hyrax/adapters/nesting_index_adapter.rb
```diff
@@ -33,19 +38,22 @@ def self.find_preservation_parent_ids_for(id:)
...
         solr_document = find_solr_document_by(id: id)
         coerce_solr_document_to_index_document(original_solr_document: solr_document, id: id)
       end
 
+      ##
       # @api public
       # @yieldparam id [String]
       # @yieldparam parent_id [Array<String>]
-      # Samvera::NestingIndexer.reindex_all!(extent: FULL_REINDEX)
-      # rubocop:disable Lint/UnusedMethodArgument
```
jeremyf, line 48: You need to keep this method in place;

The interface specs test if this method takes a block.  The way to test that interface is through parameters (e.g. `&block`)  However it doesn't use `block.call` (The original author, likely me, must've chosen `yield`)

no-reply: sorry i pushed the patch for this some time ago, but forgot to readd the rubocop.

we should probably fix the upstream test `&block` has historically been much slower than omitting it. it should be fine to test the actual `yield` in the shared examples, right?

jeremyf: The shared examples lean on method signatures/parameters.  So yes, we could change that test, however, it'll be much more expensive.

jeremyf (approved):

PR #4672, opened 2020-12-15
remove dependency injection from WorkflowListener

no-reply:
listeners struggle with dependency injection because they need to be initialized
on app startup. Rails tries to reload their collaborator objects, which leads to
`ArgumentError (A copy of Hyrax::Workflow::WorkflowFactory has been removed
from the module tree but is still active!)`

pending a better fix, we just need to ban DI for listeners and hard code their
collaborators.

@samvera/hyrax-code-reviewers


jeremyf (approved):

jeremyf:
This looks like a reasonable path/approach to get things unstuck.

PR #4675, opened 2020-12-16
add SideKiq to dassie/docker-compose for development background jobs

no-reply:
having a realistic background job setup in development will help keep us honest
about how jobs work in practice.

adds a container build target for `hyrax-engine-dev-worker` which installs the
`fits.sh` Bash script and its dependencies (including `bash`). Keeping these off
the base image keeps image size down substantially. since workers need shared
drive space for uploads and derivatives, we make those configurable via ENV and
add entrypoint infrastructure.

the uploads path, it turns out, is set at *both* `working_path` and
`upload_path`, with different interfaces. we should clean this up ASAP.

SideKiq is `ENV` configurable out of the box with `REDIS_PROVIDER` (which needs
to be set to another ENV variable). it also can use `REDIS_URL` but this chooses
to avoid that because it seems more generic, making it less clear that it's a
sidekiq config. `REDIS_PROVIDER=SIDEKIQ_REDIS_URL` provides some hint that this
servicing sidekiq setup, rather than generic, app-wide redis configuration.

@samvera/hyrax-code-reviewers


no-reply:
we discussed running this in the same container as the application, but i couldn't bring myself to do it.

the single container approach would mean installing a Java runtime on the rails app container (which i don't love), but more importantly, is further away from [Dev/Prod Parity](https://12factor.net/dev-prod-parity). i think this last is a big deal, and we should take this seriously.

lib/hyrax/configuration.rb
```diff
@@ -282,7 +282,7 @@ def bagit_dir
     # Path on the local file system where derivatives will be stored
     attr_writer :derivatives_path
     def derivatives_path
-      @derivatives_path ||= ENV.fetch('HYRAX_DERIVATIVE_PATH', Rails.root.join('tmp', 'derivatives'))
+      @derivatives_path ||= ENV.fetch('HYRAX_DERIVATIVES_PATH', Rails.root.join('tmp', 'derivatives'))
```
cjcolvar, line 285: Is this ENV variable name change a breaking change?

no-reply: the old env was introduced earlier this week. :)

cjcolvar (approved):
This looks great!

PR #4679, opened 2020-12-18
fix fcrepo init-db to actually run database initialization

no-reply:
this might make the chart deployable again.

@samvera/hyrax-code-reviewers


orangewolf (approved):
I verified this by pulling the code, blowing away my cluster and creating a new one. database did come up correctly.

no-reply:
thanks @orangewolf. it can be such a pain to test these multi-chart changes properly.

PR #4680, opened 2020-12-18
don't try to load `hyrax-webapp` Rakefile unless in docker

no-reply:
this Rakefile is only available in docker, don't try to load it otherwise.

i'm honestly not sure why we need to do any of this(?), but this at least avoids
issues with running rake tasks when no `.internal_test_app` is present.

cc: @tpendragon do you recall why we needed to load `APP_RAKEFILE` here?

@samvera/hyrax-code-reviewers


jeremyf (approved):
Thanks for removing the `unless ||` and favor `if`.  I find that this improves legibility (e.g. I don't need to do as much mental logic mapping)

jeremyf:
Looking at the commit history and logic, the `APP_RAKEFILE` appears to be a memoizing strategy. And there's a similar strategy in play in the Rakefile (see https://github.com/samvera/hyrax/blob/c662190a5e37abe9e36ff5824ed8d799f196d291/Rakefile#L26-L29)

I suspect those two declarations of `APP_RAKEFILE` are interacting with each other.  @cjcolvar Any thoughts?

cjcolvar:
I don't understand why `hyrax-webapp`'s Rakefile needs to be loaded only in the wings benchmark/docker context.  
@jeremyf  Setting `APP_RAKEFILE` before loading `engine.rake` sets up the internal app's rake tasks under the `app` namespace making them available to be called from the engine's context.  E.g. From the root of your Hyrax checkout run the database migrations in the internal app with `rake app:db:migrate`.

PR #4682, opened 2020-12-23
replace complex controller/blacklight scope stubs with a Fake

no-reply:
using a 'fake' here instead of stubs targets tests on the direct
collaborator. we can have a realistic implementation of the
Blacklight::SearchBuilder's `scope` collaborator, and avoid complex and repeated
setup.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/filter_by_type.rb
+++ b/app/search_builders/hyrax/filter_by_type.rb
@@ -30,7 +30,9 @@ module Hyrax
     end
 
     def models_to_solr_clause
-      models.map { |model| model.try(:to_rdf_representation) || model.name }.join(',')
+      models.map do |model|
+        model.respond_to?(:to_rdf_representation) ? model.to_rdf_representation : model.name
+      end.join(',')
     end
 
     def generic_type_field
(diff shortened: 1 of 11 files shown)


app/search_builders/hyrax/filter_by_type.rb
```diff
@@ -30,7 +30,9 @@ def models
     end
 
     def models_to_solr_clause
-      models.map { |model| model.try(:to_rdf_representation) || model.name }.join(',')
+      models.map do |model|
+        model.respond_to?(:to_rdf_representation) ? model.to_rdf_representation : model.name
```
no-reply, line 34: some of our classes now have a `.try` method from `Dry::Struct` which behaves differently. since we're no longer stubbing this behavior out, we need to fix it.

cjcolvar (approved):
:+1:  This looks good to me.  Thanks for cleaning this up!

I'm curious if you're deliberately making a decision to always make explicit FactoryBot method calls (`FactoryBot.create` vs. `create`).  If so, does it make sense to enforce this in a rubocop cop?

no-reply:
> I'm curious if you're deliberately making a decision to always make explicit FactoryBot method calls (FactoryBot.create vs. create). If so, does it make sense to enforce this in a rubocop cop?

i'm trying to move us in this direction. we already do this pretty consistently with `RSpec` calls. explicit is a more friendly style, i think---less magic that's invisible to newcomers.

there are a lot of existing factorybot calls, so i haven't seriously looked at changing the style configuration. help on this front would be welcome.

PR #4683, opened 2020-12-24
test valkyrie native model deposit

no-reply:
@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/monograph_deposit_spec.rb
+++ b/spec/features/monograph_deposit_spec.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+RSpec.describe 'Creating a new Monograph (Valkyrie work)', :js, :workflow, :clean_repo do
+  let(:user) { FactoryBot.create(:user) }
+
+  before do
+    FactoryBot
+      .create(:permission_template_access,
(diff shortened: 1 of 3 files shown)


spec/features/monograph_deposit_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    choose('monograph_visibility_open')
+    check('agreement')
+
+    click_on('Save')
+
+    monograph_id = current_url.split('/').last.split('?').first
+    monograph = Hyrax.query_service.find_by(id: monograph_id)
+
+    expect(monograph)
```
cjcolvar, line 42: The other create work feature tests also test for content on the view page after save.  Do you expect to add that later in addition to the model persistence checks here?

cjcolvar (approved):
This is an amazing milestone! :tada: :confetti_ball:

spec/features/monograph_deposit_spec.rb
no-reply: yes.

as it happens, some of the view behavior needs some helping along. as i resolve any remaining show view issues i'll expand this test.

PR #4685, opened 2020-12-29
bootstrap mixin level ability tests

no-reply:
historically we've tested `Hyrax::Ability` only in the context of the generated
application `Ability`. this makes isolating dependencies very hard, which
results in a less clear interface. as we move toward supporting `Valkyrie` here,
we need to get some real unit level tests in place.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -114,6 +114,9 @@ module Hyrax
     # Add this to your {.ability_logic} if you want all logged in users to be able
     # to submit content.
     #
+    # @note this is normally injected into an application +::Ability+ by the
+    #   hyrax install generator.
+    #
     # @example
     #   self.ability_logic += [:everyone_can_create_curation_concerns]
     def everyone_can_create_curation_concerns
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -4,18 +4,21 @@ module Hyrax
   # Provides Hyrax's engine level user/group permissions.
   #
   # @note This is intended as a mixin layered over
-  # +Blacklight::AccessControls::Ability+ and +Hydra::AccessControls+. Its
-  # implementation may depend in part on behavioral details of either of those
-  # two mixins. As of Hyrax 3.0.0 there's an ongoing effort to clarify and
-  # document the specific dependencies.
+  #   +Blacklight::AccessControls::Ability+ and +Hydra::AccessControls+. Its
+  #   implementation may depend in part on behavioral details of either of those
+  #   two mixins. As of Hyrax 3.0.0 there's an ongoing effort to clarify and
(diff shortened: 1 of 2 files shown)


app/models/concerns/hyrax/ability.rb
```diff
@@ -124,15 +163,48 @@ def proxy_deposit_abilities
...
+    ##
+    # @api public
+    #
+    # Allow everyone to read {ContentBlock} data.
+    #
+    # Allow admins to read the admin dashboard, update {ContentBlock}, and edit
+    # {SolrDocument}(!?)
+    #
+    # @todo why does this allow so many different things? what is going on with
+    #   `can :edit, ::SolrDocument`.
```
cjcolvar, line 202: :joy:

spec/models/concerns/hyrax/ability_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    before { ability.can(:create, work_model) }
+  end
+
+  let(:ability_class) do
+    module Hyrax
+      module Test
+        module AbilityMixin
+          class Ability
+            include Blacklight::AccessControls::Ability
```
cjcolvar, line 35: Should this be `Hydra::Ability` to match the install generator and the example in Hyrax::Ability?

app/models/concerns/hyrax/ability.rb
```diff
@@ -93,13 +108,28 @@ def download_permissions
...
     end
 
-    # Add this to your ability_logic if you want all logged in users to be able
-    # to submit content
+    ##
+    # @api public
+    #
+    # Add this to your {.ability_logic} if you want all logged in users to be able
+    # to submit content.
+    #
```
cjcolvar, line 116: I think a note should maybe be added here that while it isn't included in ability_logic by default by this file it IS included by default in the application's ability by the install generator.

cjcolvar (commented):
This documentation is amazing! :1st_place_medal: 
I have a couple comments that I'd like responses to before I approve this PR.

app/models/concerns/hyrax/ability.rb
no-reply: good call.

spec/models/concerns/hyrax/ability_spec.rb
no-reply: i'm working these tests up from scratch, only adding functionality needed to pass them.

part of my goal is to figure out whether `Hydra::Ability` is a hard dependency for this mixin, or whether it's just most often included.

no-reply:
updated with an added note about `everyone_can_create_curation_concerns`.

app/models/concerns/hyrax/ability.rb
cjcolvar: Perfect.

cjcolvar (approved):

PR #4693, opened 2020-12-30
breaking: use `Hydra::AccessControls::Permissions` for AdminSet

no-reply:
`AdminSet` used `WithAccessRight` which included `:open_access?`,
`:authenticated_only_access?`, etc...

these methods aren't needed for `AdminSet`, so we can reduce its API footprint
in 3.0.0 by using a more targeted mixin.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
This seems good to me as it helps clean up the API which should help avoid confusion.

PR #4697, opened 2020-12-31
make collection gid index field configurable

no-reply:
this was partially configured at the `::Collection` class level, but not all
code respected it.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/my/collections_controller.rb
+++ b/app/controllers/hyrax/my/collections_controller.rb
@@ -9,9 +9,9 @@ module Hyrax
       def self.configure_facets
         configure_blacklight do |config|
           # Name of pivot facet must match field name that uses helper_method
-          config.add_facet_field ::Collection.collection_type_gid_document_field_name,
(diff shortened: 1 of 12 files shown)


cjcolvar (dismissed):
This looks good to me, but in the changes I'm seeing three different ways of using the config as a keyword argument to `#where`.  I think it would be nice to standardize around one.  (The first looks the cleanest to me.)
`Hyrax.config.collection_type_index_field =>`
`Hyrax.config.collection_type_index_field.to_sym =>`
`"#{Hyrax.config.collection_type_index_field}":`

no-reply:
@cjcolvar the first two (`Hyrax.config.collection_type_index_field =>` and `Hyrax.config.collection_type_index_field.to_sym =>`) are needed, since they are different hash keys, in line with previous behavior. `"#{Hyrax.config.collection_type_index_field}":` was an error on my part, and the reason the tests were failing.

i fixed it and will rebase now.

cjcolvar (approved):
Looks great!

PR #4698, opened 2021-01-04
remove sleep from `collection_type` feature specs

no-reply:
sleeping in tests is... :(

this one seems to pass just fine without the sleep, so let's not do it.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
:runner: :tada:

Issue #4696, opened 2020-12-31
Hyrax::CollectionType uses a custom GlobalId scheme


no-reply:
### Descriptive summary

`CollectionType#gid` has custom behavior that conflicts with `GlobalID`'s typical supported interface. We should bring it in line.

```ruby
ctype =  Hyrax::CollectionType.create(title: 'my type')

# the usual interface
ctype.to_global_id # => #<GlobalID:0x00005597abc31098 @uri=#<URI::GID gid://dassie/Hyrax::CollectionType/2>>
GlobalID::Locator.locate ctype.to_global_id # =>  #<Hyrax::CollectionType:0x...
GlobalID::Locator.locate ctype.to_global_id.to_s # =>  #<Hyrax::CollectionType:0x...

# Hyrax::CollectionType (as used by collection, etc...
ctype.gid # =>  ctype.gid=> "gid://dassie/hyrax-collectiontype/2"
GlobalID::Locator.locate ctype.gid #NameError: wrong constant name hyrax-collectiontype
Hyrax::CollectionType.find_by_gid(ctype.gid) # => #<Hyrax::CollectionType:0x...
Hyrax::CollectionType.find_by_gid('gid://dassie/just-making-some-stuff-up/2') # => #<Hyrax::CollectionType:0x0
```

we should realign so all users of  `#gid` and `.find_by_gid` get correct global id syntax.

no-reply:
the next step on this is to provide a data migration task to update existing references, then code to handle the custom format can be removed in the next major release.

PR #4701, opened 2021-01-05
make CollectionType#find_by_gid work with correct global id format

no-reply:
addresses #4696 by:

  - allowing `CollectionType#find_by_gid` to find "real" GlobalID objects and matching strings.
  - providing a rake task to convert existing references to the new syntax
  - deprecating `#gid` in favor of `#to_global_id`

to move away from the old custom format, we'll need to provide a data migration/rake task that can fix existing references.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -60,7 +60,7 @@ module Hyrax
         # Coming from the UI, a collection type id should always be present.  Coming from the API, if a collection type id is not specified,
         # use the default collection type (provides backward compatibility with versions < Hyrax 2.1.0)
         collection_type_id = params[:collection_type_id].presence || default_collection_type.id
-        @collection.collection_type_gid = CollectionType.find(collection_type_id).gid
+        @collection.collection_type_gid = CollectionType.find(collection_type_id).to_global_id
         add_breadcrumb t(:'hyrax.controls.home'), root_path
         add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
         add_breadcrumb t('.header', type_title: @collection.collection_type.title), request.path
@@ -111,7 +111,7 @@ module Hyrax
         authorize! :create, @collection
         # Coming from the UI, a collection type gid should always be present.  Coming from the API, if a collection type gid is not specified,
         # use the default collection type (provides backward compatibility with versions < Hyrax 2.1.0)
-        @collection.collection_type_gid = params[:collection_type_gid].presence || default_collection_type.gid
+        @collection.collection_type_gid = params[:collection_type_gid].presence || default_collection_type.to_global_id
         @collection.attributes = collection_params.except(:members, :parent_id, :collection_type_gid)
         @collection.apply_depositor_metadata(current_user.user_key)
         add_members_to_collection unless batch.empty?
--- a/app/services/hyrax/collections/migration_service.rb
+++ b/app/services/hyrax/collections/migration_service.rb
@@ -30,7 +30,7 @@ module Hyrax
       # @param collection [::Collection] collection object to be migrated
       def self.migrate_collection(collection)
         return if collection.collection_type_gid.present? # already migrated
-        collection.collection_type_gid = Hyrax::CollectionType.find_or_create_default_collection_type.gid
+        collection.collection_type_gid = Hyrax::CollectionType.find_or_create_default_collection_type.to_global_id
         create_permissions(collection)
         collection.save
       end
@@ -74,8 +74,8 @@ module Hyrax
       #
       # @param collection [::Collection] collection object to be migrated/repaired
       def self.repair_migrated_collection(collection)
-        return if collection.collection_type_gid.present? && collection.collection_type_gid != Hyrax::CollectionType.find_or_create_default_collection_type.gid
-        collection.collection_type_gid = Hyrax::CollectionType.find_or_create_default_collection_type.gid
+        return if collection.collection_type_gid.present? && collection.collection_type_gid != Hyrax::CollectionType.find_or_create_default_collection_type.to_global_id
+        collection.collection_type_gid = Hyrax::CollectionType.find_or_create_default_collection_type.to_global_id
         permission_template = Hyrax::PermissionTemplate.find_by(source_id: collection.id)
         if permission_template.present?
           collection.reset_access_controls!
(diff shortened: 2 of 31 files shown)


app/services/hyrax/collections/migration_service.rb
```diff
@@ -30,7 +30,7 @@ def self.migrate_all_collections
       # @param collection [::Collection] collection object to be migrated
       def self.migrate_collection(collection)
         return if collection.collection_type_gid.present? # already migrated
-        collection.collection_type_gid = Hyrax::CollectionType.find_or_create_default_collection_type.gid
+        collection.collection_type_gid = Hyrax::CollectionType.find_or_create_default_collection_type.to_global_id.to_s
```
cjcolvar, line 33: It looks like the `.to_s` here is unnecessary because it is being called in `#collection_type_gid=`

app/services/hyrax/collections/migration_service.rb
```diff
@@ -74,8 +74,8 @@ def self.repair_migrated_collections
       #
       # @param collection [::Collection] collection object to be migrated/repaired
       def self.repair_migrated_collection(collection)
-        return if collection.collection_type_gid.present? && collection.collection_type_gid != Hyrax::CollectionType.find_or_create_default_collection_type.gid
-        collection.collection_type_gid = Hyrax::CollectionType.find_or_create_default_collection_type.gid
+        return if collection.collection_type_gid.present? && collection.collection_type_gid != Hyrax::CollectionType.find_or_create_default_collection_type.to_global_id
```
cjcolvar, line 77: Do you need to use `.to_s` on the result of `.to_global_id` because `.collection_type_gid` is a string and not a `GlobalID`?

lib/generators/hyrax/templates/db/seeds.rb
```diff
@@ -62,7 +62,7 @@ def create_collection(user, type_gid, id, options)
   col = Collection.new(id: id)
   col.attributes = options.except(:visibility)
   col.apply_depositor_metadata(user.user_key)
-  col.collection_type_gid = type_gid
+  col.collection_type_gid = type_gid.to_s
```
cjcolvar, line 65: The `.to_s` appears to be unnecessary because it is called in `collection_type_gid=`.

spec/tasks/rake_spec.rb
```diff
@@ -25,4 +25,39 @@
...
+
+        before do
+          FactoryBot.create_list(:collection, 3, collection_type_gid: collection_type.to_global_id)
+          FactoryBot.create_list(:collection, 3, collection_type_gid: other_collection_type.to_global_id)
+        end
+
+        it 'updates collections to use standard GlobalId URI' do
+          expect { run_task 'hyrax:collections:update_collection_type_global_ids' }
+            .to change { collections_with_legacy_gids.map { |col| col.reload.collection_type_gid } }
+            .to eq [collection_type.to_global_id.to_s, other_collection_type.to_global_id.to_s]
```
cjcolvar, line 58: Nice!

cjcolvar (commented):
This looks really good!  I'm happy to fully approve once my question on MigrationService line 77 is resolved.

lib/generators/hyrax/templates/db/seeds.rb
no-reply: thanks. not sure how this snuck in on me

app/services/hyrax/collections/migration_service.rb
no-reply: ty!

app/services/hyrax/collections/migration_service.rb
no-reply: yeah, that's right.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -60,7 +60,7 @@ def new
         # Coming from the UI, a collection type id should always be present.  Coming from the API, if a collection type id is not specified,
         # use the default collection type (provides backward compatibility with versions < Hyrax 2.1.0)
         collection_type_id = params[:collection_type_id].presence || default_collection_type.id
-        @collection.collection_type_gid = CollectionType.find(collection_type_id).gid
+        @collection.collection_type_gid = CollectionType.find(collection_type_id).to_global_id
```
elrayle, line 63: I like the change to #to_global_id.  Makes the purpose of the method clearer.

elrayle (approved):
This looks really good.  I like seeing it move to use Rails provided Global ID processing.

app/services/hyrax/collections/migration_service.rb
elrayle: I gave my approval for this, although it looks like this change is still pending.

elrayle (changes requested):
@cjcolvar request needs to be resolved before merging.

app/services/hyrax/collections/migration_service.rb
elrayle: I wonder if there should be a test for this.  Seems like one of the tests should have failed when collection.collection_type_gid didn't equal Hyrax::CollectionType.find_or_create_default_collection_type.to_global_id because the first was a String and the later was a GlobalID.

elrayle: Looking at this more, the line after this sets `collection.collection_type_gid` using `to_global_id` which appears to imply that the gid in the collection is a GlobalID.

elrayle: And the line 63 in the collections_controller also sets it to GlobalID...
```
@collection.collection_type_gid = CollectionType.find(collection_type_id).to_global_id
```

Is this an issue for backward compatibility where collection_type_gid is a String?

elrayle (approved):
Per conversation with @no-reply in Slack, all comment by @cjcolvar were resolved.  This change looks good to me.

app/services/hyrax/collections/migration_service.rb
no-reply: > Is this an issue for backward compatibility where collection_type_gid is a String?

`#collection_type_gid=` casts to a string, right?

cjcolvar: I was concerned about this line because it looks like it is a String != GlobalID comparison.  Am I misreading this line or will a GlobalID automatically cast so it isn't an issue?

no-reply: let's just render the issue moot and revert the change in this line and the test for this service.

the whole thing is dead code that probably should have been deprecated for removal somewhere around Hyrax 2.2.0. https://github.com/samvera/hyrax/pull/4716

cjcolvar: Good point.  Sounds good to me.

jeremyf (approved):

PR #4711, opened 2021-01-08
backport: Update skip_to_content.js

### LINKED ISSUE #4705: ES6 breaks JS compression unless Uglifier is in harmony mode
### Descriptive summary

Version: `2.x.stable`, `2.9.1`

This line of ES6 syntax introduced in #4503 makes `assets:precompile` fail unless one also configures Uglifier to run in "harmony mode":
`let skipTo = '#' + $(this)[0].firstElementChild.hash.split('#')[1];`

Solution in `config/environments/production.rb`:
`config.assets.js_compressor = Uglifier.new(harmony: true)`

### Rationale
Introducing ES6 requires the above change to fix the asset pipeline in production environments. It also adds a requirement for implementers to add a transpiler (e.g. `babel`) for backward compatibility with older browsers like IE.

### Expected behavior
Assets precompile without issue.

### Actual behavior
During an `assets:precompile` task, the following error stops it from completing:
`Uglifier::Error: Unexpected token: name (skipTo). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).`

1. Run `rake assets:precompile`

### Related work
- #4503

bkeese:
Remove ES6 dependency from 2.x stable.

Fixes #4705 

This commit removes the offending ES6 `let` syntax.

This does not address any potential changes desired in CI build that would have triggered this bug.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/skip_to_content.js
+++ b/app/assets/javascripts/hyrax/skip_to_content.js
@@ -4,7 +4,7 @@ Blacklight.onLoad(function () {
   $(".skip-to-content").click(function(event) {
     event.preventDefault();
     // element to focus on
-    let skipTo = '#' + $(this)[0].firstElementChild.hash.split('#')[1];
+    var skipTo = '#' + $(this)[0].firstElementChild.hash.split('#')[1];
 
     // Setting 'tabindex' to -1 takes an element out of normal
     // tab flow but allows it to be focused via javascript


no-reply:
since this is such a simple change, do you think we should merge it to `master` first and backport from there?

bkeese:
I'll follow your lead on that one. If we are moving toward ES6 in 3.0, maybe it makes sense to leave it as-is there? On the other hand, less discrepancies are better, so... ? I don't have a preference.

no-reply:
> If we are moving toward ES6 in 3.0, maybe it makes sense to leave it as-is there? 

afaik, this is the only ES6 code in 3.0 as well. if we don't need to force apps to update, it's better not to.

no-reply (approved):

PR #4715, opened 2021-01-12
refactor migration_service_spec

no-reply:
cuts more than 30 seconds off typical test time by dropping `let!` calls in
favor of `let`.

there are probably more wins here, but might as well get this in.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/collections/migration_service_spec.rb
+++ b/spec/services/hyrax/collections/migration_service_spec.rb
@@ -1,25 +1,30 @@
 # frozen_string_literal: true
+
+# DON'T EVEN LOOK AT THESE TESTS!
+# the service tested here is deprecated for removal and is only called from a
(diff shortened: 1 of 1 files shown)


jeremyf (dismissed):

no-reply:
@jeremyf sorry. fixed a rubocop complaint.

spec/services/hyrax/collections/migration_service_spec.rb
```diff
@@ -1,25 +1,30 @@
 # frozen_string_literal: true
+
+# DON'T EVEN LOOK AT THESE TESTS!
```
no-reply, line 3: deprecation is proposed in https://github.com/samvera/hyrax/pull/4716

jeremyf (approved):

PR #4718, opened 2021-01-12
check FlipFlop for caching manifest configuration at runtime

no-reply:
checking FlipFlop at app startup can be problematic if the database isn't
present. it also causes issues if this configuration gets flipped while the app
is running (the normal use case for FlipFlop).

probably this whole configuration should be deprecated in favor of an
environment flag.

@samvera/hyrax-code-reviewers


no-reply:
@cjcolvar what do you think of this as an immediate fix to the issue you dropped in slack?

cjcolvar (dismissed):
This looks perfect! :heart:

cjcolvar (approved):

PR #4719, opened 2021-01-13
allow search filter to find `Hyrax::PcdmCollection` for searches

no-reply:
when searches should find a collection model, allow them to find
`Hyrax::PcdmCollection`.

@samvera/hyrax-code-reviewers


no-reply:
closing this as maybe unneeded.

PR #4724, opened 2021-01-15
Find ids model wings query

no-reply:
the inline implementation of this query logic was repeated in multiple places,and didn't scale (if there were many collections associated with the user via permission templates, the query would pass too many arguments to Solr).

the query is now supported in Hyrax as a Valkyrie custom query, so we provide a complete, Solr-based paging implementation for Wings and switch to using this for the query.

supersedes #4722  

@samvera/hyrax-code-reviewers


jeremyf (approved):
This is a fantastic change and improvement!

jeremyf (approved):

PR #4728, opened 2021-01-19
Switching blacklight http_method to :post

jeremyf:
Because too many times on Samvera tech people raise a problem regarding
a failed query to SOLR.  Often, it's because they inadvertantly exceeded
the character limit of a GET request.

@samvera/hyrax-code-reviewers


no-reply (approved):

no-reply:
should we backport to 2.x?

jeremyf:
Pushing the backport now-ish.

no-reply:
awesome! thank you!

PR #4730, opened 2021-01-20
extend permission template to work with valkyrie

no-reply:
add `PermissionTemplate#source` to act as the new source_id relationship
resolver. we don't deprecate the old `#source_model` here; it will hang around
until we start phasing out AF support.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/permission_template.rb
+++ b/app/models/hyrax/permission_template.rb
@@ -35,6 +35,14 @@ module Hyrax
     has_one :active_workflow, -> { where(active: true) }, class_name: 'Sipity::Workflow', foreign_key: :permission_template_id
 
     ##
+    # @note this is a convenience method for +Hyrax.query_service.find_by(id: template.source_id)+
+    #
+    # @return [Hyrax::Resource] the collection this template is associated with
+    def source
+      Hyrax.query_service.find_by(id: source_id)
+    end
+
+    ##
     #
     # A bit of an analogue for a `belongs_to :source_model` as it crosses from Fedora to the DB
     # @return [AdminSet, ::Collection]
--- a/app/models/hyrax/permission_template.rb
+++ b/app/models/hyrax/permission_template.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 module Hyrax
+  ##
   # Defines behavior that is applied to objects added as members of an AdminSet
   #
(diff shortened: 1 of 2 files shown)


app/models/hyrax/permission_template.rb
```diff
@@ -29,6 +34,16 @@ def agent_ids_for(agent_type:, access:)
...
+    ##
+    # @note this is a convenience method for +Hyrax.query_service.find_by(id: template.source_id)+
+    #
+    # @return [Hyrax::Resource] the collection this template is associated with
+    def source
+      Hyrax.query_service.find_by(id: source_id)
+    end
+
+    ##
+    #
```
no-reply, line 46:
```suggestion
```

spec/models/hyrax/permission_template_spec.rb
```diff
@@ -1,271 +1,249 @@
...
   let(:attributes) { { source_id: admin_set.id } }
-
-  subject { permission_template }
+  let(:admin_set) { FactoryBot.valkyrie_create(:hyrax_admin_set) }
 
   it { is_expected.to have_many(:available_workflows).dependent(:destroy) }
   it { is_expected.to have_one(:active_workflow).conditions(active: true).dependent(nil) }
   it { is_expected.to have_many(:access_grants).dependent(:destroy) }
 
-  describe 'factories' do
```
jeremyf, line 14: Always a code smell when you _must_ test a factory's setup; Usually indicative of a series of architectural challenges. I like that we had them to verify things, but am happy to see them go away, as it shows we're moving down a better path.

spec/models/hyrax/permission_template_spec.rb
```diff
@@ -1,271 +1,249 @@
...
   end
 
   describe "#release_before_date?" do
-    subject { permission_template.release_before_date? }
-
     context "with release_period='before'" do
       let(:attributes) { { source_id: admin_set.id, release_period: Hyrax::PermissionTemplate::RELEASE_TEXT_VALUE_BEFORE_DATE } }
 
-      it { is_expected.to be true }
+      it { is_expected.to be_release_before_date }
```
jeremyf, line 108: These are nice changes!

jeremyf (approved):

PR #4729, opened 2021-01-19
Switching blacklight http_method to :post

jeremyf:
Because too many times on Samvera tech people raise a problem regarding
a failed query to SOLR.  Often, it's because they inadvertantly exceeded
the character limit of a GET request.

Related to #4729 

@samvera/hyrax-code-reviewers

PR #4732, opened 2021-01-22
Switching blacklight http_method to :post

jeremyf:
Because too many times on Samvera tech people raise a problem regarding
a failed query to SOLR.  Often, it's because they inadvertantly exceeded
the character limit of a GET request.

Related to #4728 (supersedes #4729)

@samvera/hyrax-code-reviewers

### DIFF
--- a/.regen
+++ b/.regen
@@ -1,1 +1,1 @@
-3
+3.http_method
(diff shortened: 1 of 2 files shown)


cjcolvar (dismissed):

cjcolvar (approved):

jeremyf:
@cjcolvar Do you have any thoughts on why this might be breaking?

PR #4734, opened 2021-01-26
Allow registrar to return nil for the identifier

cjcolvar:
This is useful for when the registrar decides that the object shouldn't be registered but there hasn't been an error.

@samvera/hyrax-code-reviewers


no-reply (approved):
nice change!

does this need a documentation change over in the Registrar? should we add a test?

Issue #4753, opened 2021-02-23
View Your Works accessibility concerns in Hyrax.pre.rc3


rjkati:
Brought over from [https://github.com/samvera/hyrax/issues/3966](url)

Axe has flagged four potential accessibility issues in Firefox/Windows for Hyrax.pre.rc3. Two potential concerns are flagged in Chrome/Windows

![issue_1](https://user-images.githubusercontent.com/32644254/108897653-555bb980-75e4-11eb-9342-a44e1480223a.PNG)
![issue_2](https://user-images.githubusercontent.com/32644254/108897660-57be1380-75e4-11eb-8bff-d51edab79baf.PNG)
![issue_3](https://user-images.githubusercontent.com/32644254/108897668-5987d700-75e4-11eb-81ef-7bf997b81d68.PNG)
![issue_4](https://user-images.githubusercontent.com/32644254/108897677-5b519a80-75e4-11eb-92f8-72e17e2a3ee8.PNG)

### Steps to reproduce the behavior

1. Install the Axe plugin or addon from Deque (https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/ is the firefox version).
2. Open https://nurax-dev.curationexperts.com/dashboard/my/works?locale=en
3. Open the inspector (right-click, and select "Inspect element")
4. Click on the 'Axe' tab
5. Click 'Analyze'

jlhardes:
@rjkati Clarifying what still needs to be addressed here - is this to distinguish links in the Works list in the Dashboard? So some link title attributes are not including enough info (like the Work Title) to distinguish one link from another in the Works list?

rjkati:
@jlhardes Yes, Axe is flagging some link title attributes because they are not distinct enough from each other. 

The links in question seem to be the "Visibility of Work" links (e.g. "Private", "Public"), which I assume are intentionally similar and we probably don't want to change them. So the fix for this may be that we document a decision somewhere where it can be easily referenced by testers.

jlhardes:
Thanks! The text in the `<span>` ("Public" or "Private") shouldn't change because that is displayed as part of the row of information for the work but the `<a>` surrounding that text can have a @title attribute that includes that text and the title of the work for better screen reader interaction:

`<a id="permission_v118rd835" class="visibility-link" href="/concern/generic_works/v118rd835/edit?locale=en#share" title="[Work Title] is publicly visible"><span class="label label-success">Public</span></a>`

That seems like something to improve, if possible.

[...]

PR #4755, opened 2021-02-25
Adding title attribute for visibility link

jeremyf:
Given that we're linking to the "share" tab, I opted to include the I18n
translation of the [Share tab][#1].  I'm including it as a prefix for
the title.  To do much else would require another translation entry.  I
believe this is adequate to provide enough differentiation in the titles.

Related to #4753

[1]:https://github.com/samvera/hyrax/blob/001bb7ad73300a2183ed681c0b4e917f25199b67/app/views/hyrax/base/_guts4form.html.erb#L21

@samvera/hyrax-code-reviewers


no-reply (approved):
💮

PR #4761, opened 2021-02-26
Do not precompile assets in hyrax-engine-dev image

dlpierce:
Do not precompile assets in hyrax-engine-dev image

While updating a gem that includes breaking JS changes, the old JS mysteriously persisted being served up. It was discovered that the old code was living on as compiled assets in the hyrax-public volume which is mounted over the image's public directory. That outdated compiled JS is then used despite not matching the source.

To avoid this, we should not precompile assets in dev. However `yarn install` is being run as part of the precompile, and that behavior needs to be retained.

* Do not precompile assets in the hyrax-engine-dev image

* Should have no effect in production

@samvera/hyrax-code-reviewers


no-reply (approved):
seems okay to me!

PR #4770, opened 2021-03-03
Docker Build Improvements

orangewolf:
Lock base at alpine3.12 and update fits to a url that still resolves, skip chown of paths if vars are not set

FITS is no longer available at the Harvard URL. When I switched to the Github url, it would not resolve due to a bug in Alpine 3.13.  In order to debug, I wanted to run individual build steps, but could not due to the entry point vars.  These were all "on the way to fix something else" issues, but I wanted to capture them so others don't have to figure it out.

@samvera/hyrax-code-reviewers


no-reply:
looks like it needs a rebase?

otherwise 👍🏻. frustrating about the alpine image.

no-reply (approved):

PR #4773, opened 2021-03-05
Add/Update translations

dlpierce:
Add i18n for error messages in ProxyDepositRequest and then ran i18n-tasks

* Updated translations

* Check for correctness

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/models/proxy_deposit_request_spec.rb
+++ b/spec/models/proxy_deposit_request_spec.rb
@@ -149,7 +149,7 @@ RSpec.describe ProxyDepositRequest, type: :model do
       it 'raises an error' do
         subject.save!
         expect(subject2).not_to be_valid
-        expect(subject2.errors[:open_transfer]).to eq(['Must close open transfer on the work before creating a new one'])
+        expect(subject2.errors[:open_transfer]).to eq(['Must close the open transfer on the work before creating a new one'])
       end
 
       context 'when the first transfer is closed' do
--- a/spec/models/proxy_deposit_request_spec.rb
+++ b/spec/models/proxy_deposit_request_spec.rb
@@ -108,7 +108,7 @@ RSpec.describe ProxyDepositRequest, type: :model do
       it 'raises an error' do
         subject.transfer_to = 'dave'
         expect(subject).not_to be_valid
-        expect(subject.errors[:transfer_to]).to eq(['must be an existing user'])
+        expect(subject.errors[:transfer_to]).to eq(['Must be an existing user'])
       end
     end
 
@@ -139,7 +139,7 @@ RSpec.describe ProxyDepositRequest, type: :model do
       it 'raises an error' do
         subject.transfer_to = sender.user_key
         expect(subject).not_to be_valid
-        expect(subject.errors[:transfer_to]).to eq(['specify a different user to receive the work'])
+        expect(subject.errors[:transfer_to]).to eq(['Specify a different user to receive the work'])
       end
     end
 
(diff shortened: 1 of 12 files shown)


no-reply:
seems like there's still a string mismatch introduced by this.

spec/models/proxy_deposit_request_spec.rb
```diff
@@ -149,7 +149,7 @@
       it 'raises an error' do
         subject.save!
         expect(subject2).not_to be_valid
-        expect(subject2.errors[:open_transfer]).to eq(['must close open transfer on the work before creating a new one'])
+        expect(subject2.errors[:open_transfer]).to eq(['Must close open transfer on the work before creating a new one'])
```
no-reply, line 152:
```suggestion
        expect(subject2.errors[:open_transfer]).to eq(['Must close the open transfer on the work before creating a new one'])
```

dlpierce:
> seems like there's still a string mismatch introduced by this.

Yes, I forgot I added "the" to hopefully help the translation. Thanks!

no-reply (approved):
looks good! thanks for picking these off.

dlpierce:
This needs to be reviewed by proficient speakers before merging, correct?

no-reply:
cc: @jlhardes i'm loathe to block improvements on a time consuming human review process, but i also want to make sure we're moving toward a policy that improves the translations over time. what are your thoughts?

also, cc: @j-dornbusch (in case you have a second to review the `de` locale)

j-dornbusch:
There are some small errors in the French translation that I can fix. The problem is that the overall translation of "Work" by "travail" or "œuvre" in French Hyrax locales is inadequate which makes them almost unusable as is. But finding a word that will work in all contexts is not easy. Should I nevertheless leave "travail" or "œuvre" until an appropriate word is found? @no-reply

jlhardes:
Thank you for the review @j-dornbusch! That is a big mix of "travail" and "œuvre" for Work in the Dashboard but I also do not speak French so there is a lot here I probably don't understand. The Spanish translation is using "trabajo" in all places for Work and Italian is using "lavoro" except for referring to batches where it becomes "opera." "Trabajo" and "lavoro" are similar to "travail" in French, which is more like a job or assignment, correct? When I think about Works in Hyrax I don't think of jobs or assignments, I think of them as things or objects. I would probably think more along the lines of "œuvre" in French, "obra" in Spanish, and "opera" in Italian as the terms to use but that is through my limited language lens. Seems like Work is a messy concept to translate.

@no-reply I think there is more work (sigh) to do with all of the translations. The French translation for Work right now seems pretty confusing but Spanish and Italian, while more consistent, don't necessarily make sense either. Since we don't have a definite term to use in French right now, let's go with what we have now and know this is in need of improvement.

j-dornbusch:
Indeed @jlhardes, I'm not sure that there is a strict equivalent of "work" in all languages. We have to consider the wide variety of contexts in which Hyrax is used (publications, data, ...). In French, the word "œuvre" denotes a work of great value ("L’œuvre de Victor Hugo"). "Travail" is relevant in some contexts (scientific papers) but not in all. The most neutral word would be "ressource" which is an equivalent of "asset" as in the expression "digital asset management". But I don't have the competence to conclude this debate.
The risk, if there is no generic term equivalent to "work" in a language is that almost all locales would have to be rewritten for a non-English speaking implementation of Hyrax, which is our case, making it difficult to contribute to the improvement of the community's base of locales.

In the meantime,, Il will correct these few translations and keep "work" everywhere.

rjkati:
@jlhardes "travail" is used on the dashboard everywhere except for "vos oeuvers" ("my works"), which seems to be expected per the earlier messages in this ticket.

Issue #110, opened 2016-12-21
Users should be able to opt into GScholar integration by designating one FileSet within a Work as its citable full-text PDF


mjgiarlo:
<a href="https://github.com/mjgiarlo"><img src="https://avatars.githubusercontent.com/u/131982?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mjgiarlo](https://github.com/mjgiarlo)**
_Friday Dec 11, 2015 at 17:52 GMT_
_Originally opened as https://github.com/projecthydra/sufia/issues/1460_

----

As a result of decomposing the GenericFile from Sufia < 7 into separate FileSet and Work concerns, the embedded metadata that drives Google Scholar citations is now out of step with the implementation of Sufia 7's domain model.

To support indexing in Google Scholar, a work should have a citable full-text PDF document:

https://github.com/projecthydra/sufia/blob/master/app/views/shared/_citations.html.erb#L23

Perhaps the Work edit page could have a widget that prompts and allows a user to select a FileSet within a work as its citable full-text PDF. This work involves the UI so @mtribone should at least be consulted.

no-reply:
cc: @mark-dce

mjgiarlo:
This issue would benefit from a discussion with folks that have already successfully integrated GScholar with Sufia or Hyrax. There's significantly more work that needs done to satisfy this feature than designating a FS as the full-text one for a work.

julesies:
@no-reply I'm not clear what this is asking, our repo-manager group was thinking it might need review by others, so leaving open.

no-reply:
@jlhardes @jeremyf this reminds me of our current conversations, and of #4774

rjkati:
@jlhardes this ticket also seems to need discussion of expected behavior.

rjkati:
The value for "citation_pdf_url" is tied to the representative media selection for the work.

I tested this with https://nurax-dev.curationexperts.com/concern/generic_works/fb494864p?locale=en. Changing the thumbnail selection in the file manager did not change the value for "citation_pdf_url". However, changing the representative media selection in the file manager changed the value of "citation_pdf_url" to the fileset marked as the representative media.

jlhardes:
Great, thank you for testing! With the representative media selection feature in the file manager, I think that works to resolve this issue.

PR #4774, opened 2021-03-05
improve google scholar meta tag presentation

no-reply:
* move view logic for tag value presentation into a presenter
* create a hook for suppressing google scholar meta tags for non-scholarly content
* guarantee author order:
  * use `#ordered_authors` for author content, if present (i.e. add a hook for applicaitons)
  * use and sort `BasicMetadata#creator` otherwise

* make sure google scholar meta tags still appear in `<head>` for work show views

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/google_scholar_presenter.rb
+++ b/app/presenters/hyrax/google_scholar_presenter.rb
@@ -4,6 +4,12 @@ module Hyrax
   ##
   # Handles presentation for google scholar meta tags.
   #
+  # @example
+  #   my_book = Monograph.new(title: ['On Moomins'], creator: ['Tove', 'Lars'])
+  #   scholar = GoogleScholarPresenter.new(my_book)
+  #
+  #   scholar.title => 'On Moomins'
+  #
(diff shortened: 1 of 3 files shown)


jeremyf (dismissed):

no-reply:
@jeremyf after originally thinking the best thing to do was to have this presenter give authors as always-ordered, i thought better of it. adding the `.sort` call would have circumvented any existing users who have (probably painstakingly) added order support to their `#creator` field.

i think it's best to let the current unordered nature of authors stay that way until we can get order at the schema level.

i still think this goes a long way toward resolving the overall issue (it's the "make the change easy" part) and has value for #110, so i'm pushing a version with the tests fixed.

jeremyf (approved):

rjkati:
Google Scholar meta tags for "citation_title", "citation_author" and "citation_pdf_url" display for both generic and image works on the work show page. Per @no-reply's first comment it seems like this is all that needs to be tested for this PR.

@jlhardes

Issue #4191, opened 2020-01-22
Allow CreateDerivativesJob to receive either an AF object or a Valkyrie resource


randalldfloyd:
Allow CreateDerivativesJob to receive either an ActiveFedora object or a Valkyrie resource

### Descriptive summary
Convert CreateDerivativesJob to allow it to receive either an ActiveFedora based object or a Valkyrie::Resource equivalent object.

### Rationale
Allow jobs to execute correctly with existing AF implementation or the Wings Valkyrie implementation.

### Expected behavior
CreateDerivativesJob receives a Valkyrie::Resource and performs the action.

### Actual behavior
CreateDerivativesJob raises an exception when passed a Valkyrie::Resource.

### Approach for Conversion
[See Pattern for Jobs Receiving Parameter that may be an AF Object or a Valkyrie Resource](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727)

### Related work

These have all be modified to support AF object or Valkyrie resource using the [established pattern](https://docs.google.com/document/d/1f9yDjtkXapFf1DYLyPhprnB4CPGaZw9p_k8T_tcor-U/edit#heading=h.qt9rxydfd727). 

* Issue #4101 Allow GrantEditToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4102 Allow GrantReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4103 Allow RevokeReadToMembersJob to receive either an AF object or a Valkyrie resource as the Work (PR #4123)
* Issue #4099 Add GlobalID::Identification for resources passed to jobs (PR #4124)

mjgiarlo:
May be blocked by #4060.

hackartisan:
related work: https://github.com/samvera/hyrax/pull/5540

no-reply:
some discussion occurred in Slack https://samvera.slack.com/archives/CDRHAMUTV/p1692716666450449.

> I don't think we need to update the existing job

hackartisan:
Created https://github.com/samvera/hyrax/issues/6162 and closing this.

Issue #3722, opened 2019-03-28
CreateWithRemoteFilesOrderedMembersActor isn't thread safe


mbklein:
### Descriptive summary

Since `ActionDispatch::MiddlewareStack` reuses specific middlewares from a pool at runtime, individual middlewares (such as actors) should not count on having mutable instance variables. In this case, that manifested through the `@ordered_members` accumulator on the `CreateWithRemoteFilesOrderedMembersActor` actor, which ended up with a jumble of unrelated file sets under heavy ingest load (such as a batch ingest operation).

mbklein:
* Initially fixed by #3477
* Referenced by #3549

no-reply:
I'm tempted to fix this issue by dropping `ActionDispatch::MiddlewareStack`. We're really only using it to provide us a stack, and there's no reason we need this pooling, or other `MiddlewareStack` features.

jlhardes:
This is work that could be done but is not something the Hyrax Maintenance WG is going to undertake. Valkyrie will be offering improvements for Fedora users so if this is something that is needed in the meantime, it is an open issue that someone in the community can take.

PR #4780, opened 2021-03-09
Add indexing for Hyrax::FileSet through valkyrie indexing

### LINKED ISSUE #4766: Create a FileSet solr doc in Valkyrie solr when files are added to a Work
### Descriptive summary

When writing to Valkyrie's solr is enabled, FileSets are written to ActiveFedora managed solr core but not to the Valkyrie solr core.

### Rationale

FileSets are a core model type and should be saved in the Valkyrie solr core.

### Expected behavior

When `config.query_index_from_valkyrie = true` in `config/initializers/hyrax.rb` and a File is added to a work...

* should write a FileSet solr doc to ActiveFedora managed solr core (at least for now)
* should write a FileSet solr doc to Valkyrie's solr core 

### Actual behavior

When `config.query_index_from_valkyrie = true` in `config/initializers/hyrax.rb` and a File is added to a work...

* writes a FileSet solr doc to ActiveFedora managed solr core
* DOES NOT write a FileSet solr doc to Valkyrie's solr core 

#### Prerequisites

Edit `.internal_test_app/config/initializers/hyrax.rb` and set config

```
```

#### Get count of FileSets solr docs

Get counts through Solr UI for ActiveFedora Solr core:
1. navigate to: http://localhost:8983/solr/#/hydra-development/query
2. check facet and set `facet.field=has_model_ssim`
3. go to bottom of results and record counts for model types
    * FileSet

Get model type counts in Solr UI for Valkyrie Solr core:
1. navigate to: http://localhost:8987/solr/#/hyrax-valkyrie-test/query
2. check facet and set `facet.field=has_model_ssim`
3. go to bottom of results and record counts for model types
    * FileSet

#### Check for count changes when adding a file to a work

BEFORE get counts of FileSets for both ActiveFedora Solr core and Valkyrie Solr core.

1. Login
2. Dashboard -> Works -> Add new work -> type=Generic Work
3. Fill in required fields and add a file
4. Save

AFTER get counts of FileSets for both ActiveFedora Solr core and Valkyrie Solr core.

Verify that the counts for FileSets increased by one in both Solr cores.

### Related work

* Issue #4756 - UBER: Not all model types are saved to the Valkyrie Solr core 
* Issue #4765 - Create a solr doc in Valkyrie solr when creating/modifying an AdminSet
* Issue #4767 - Create AccessControl solr docs in Valkyrie solr when creating Works, Collections, and AdminSets

elrayle:
Fixes #4766

### Description

This creates indexing for [Hyrax::FileSet](https://github.com/samvera/hyrax/blob/master/app/models/hyrax/file_set.rb), which is a `Valkyrie::Resource`, using the Valkyrie indexing process called from the [MetadataIndexListener](https://github.com/samvera/hyrax/blob/master/app/services/hyrax/listeners/metadata_index_listener.rb#L18-L24).

### Questions
There are two questions posed in the comments in the file changes.
* Is the refactor of `Hyrax::FileSetIndexer` to `Hyrax::ActiveFedoraFileSetIndexer` acceptable?  It is necessary for Valkyrie indexing naming conventions, but does represent a public API change.
* Do we want to index all file metadata?

### Known remaining issues

Jobs that need to add support for Valkyrie
* CharacterizationJob (#4100)
* CreateDerivativesJob (#4191)
* possible others as well

Valkyrie works do not include FileSets as members in solr docs.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -11,7 +11,7 @@ module Hyrax
       # @param [FileSet] file_set the parent FileSet
       # @param [Symbol, #to_sym] relation the type/use for the file
       # @param [User] user the user to record as the Agent acting upon the file
-      def initialize(file_set, relation, user, use_valkyrie: false)
+      def initialize(file_set, relation, user, use_valkyrie: Hyrax.config.query_index_from_valkyrie)
         @use_valkyrie = use_valkyrie
         @file_set = file_set
         @relation = normalize_relation(relation)
@@ -75,7 +75,7 @@ module Hyrax
         CharacterizeJob.perform_later(file_set, repository_file.id, pathhint(io))
       end
 
-      def perform_ingest_file_through_valkyrie(io)
+      def perform_ingest_file_through_valkyrie(io) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         file =
           begin
             Hyrax.storage_adapter.upload(resource: file_set, file: io, original_filename: io.original_name, use: relation)
@@ -87,7 +87,9 @@ module Hyrax
         create_version(file_metadata, user)
         id = file_metadata.file_identifier
         file_set.file_ids << id
+        file_set.original_file_id = id
         Hyrax.persister.save(resource: file_set)
+        Hyrax.publisher.publish('object.metadata.updated', object: file_set, user: user)
         CharacterizeJob.perform_later(file_set, id.to_s, pathhint(io))
         file_metadata
       end
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -6,7 +6,7 @@ module Hyrax
       include Lockable
       attr_reader :file_set, :user, :attributes, :use_valkyrie
 
-      def initialize(file_set, user, use_valkyrie: false)
+      def initialize(file_set, user, use_valkyrie: Hyrax.config.query_index_from_valkyrie)
         @use_valkyrie = use_valkyrie
         @file_set = file_set
         @user = user
@@ -87,7 +87,7 @@ module Hyrax
       def attach_to_valkyrie_work(work, file_set_params)
         work = Hyrax.query_service.find_by(id: work.id) unless work.new_record
         file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
-        Hyrax.persister.save(resource: file_set)
+        @file_set = Hyrax.persister.save(resource: file_set)
         work.member_ids << file_set.id
         work.representative_id = file_set.id if work.representative_id.blank?
         work.thumbnail_id = file_set.id if work.thumbnail_id.blank?
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -68,7 +68,6 @@ module Hyrax
         yield(file_set) if block_given?
       end
 
-      # Adds a FileSet to the work using ore:Aggregations.
       # Locks to ensure that only one process is operating on the list at a time.
       def attach_to_work(work, file_set_params = {})
         acquire_lock_for(work.id) do
@@ -87,15 +86,18 @@ module Hyrax
       def attach_to_valkyrie_work(work, file_set_params)
         work = Hyrax.query_service.find_by(id: work.id) unless work.new_record
         file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
-        @file_set = Hyrax.persister.save(resource: file_set)
-        work.member_ids << file_set.id
-        work.representative_id = file_set.id if work.representative_id.blank?
-        work.thumbnail_id = file_set.id if work.thumbnail_id.blank?
+        fs = Hyrax.persister.save(resource: file_set)
+        Hyrax.publisher.publish('object.metadata.updated', object: fs, user: user)
+        work.member_ids << fs.id
+        work.representative_id = fs.id if work.representative_id.blank?
+        work.thumbnail_id = fs.id if work.thumbnail_id.blank?
         # Save the work so the association between the work and the file_set is persisted (head_id)
         # NOTE: the work may not be valid, in which case this save doesn't do anything.
         Hyrax.persister.save(resource: work)
+        Hyrax.publisher.publish('object.metadata.updated', object: work, user: user)
       end
 
+      # Adds a FileSet to the work using ore:Aggregations.
       def attach_to_af_work(work, file_set_params)
         work.reload unless work.new_record?
         file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
--- a/app/indexers/hyrax/valkyrie_file_set_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_file_set_indexer.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Indexes Hyrax::FileSet objects
+  class ValkyrieFileSetIndexer < Hyrax::ValkyrieIndexer
+    include Hyrax::ResourceIndexer
+    include Hyrax::PermissionIndexer
+    include Hyrax::VisibilityIndexer
+    include Hyrax::Indexer(:core_metadata)
+    include Hyrax::Indexer(:basic_metadata)
+
+    # include Hyrax::IndexesThumbnails # TODO: Is there a Valkyrie version of a thumbnail indexer?
+
+    def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+      super.tap do |solr_doc| # rubocop:disable Metrics/BlockLength
+        # Metadata from the FileSet
+        solr_doc['file_ids_ssim']         = resource.file_ids&.map(&:to_s)
+        solr_doc['original_file_id_ssi']  = resource.original_file_id.to_s
+        solr_doc['thumbnail_id_ssi']      = resource.thumbnail_id.to_s
+        solr_doc['extracted_text_id_ssi'] = resource.extracted_text_id.to_s
+
+        # Add in metadata from the original file.
+        file_metadata = original_file
+        return solr_doc unless file_metadata
+
+        # Label is the actual file name. It's not editable by the user.
+        solr_doc['original_file_alternate_ids_tesim'] = file_metadata.alternate_ids&.map(&:to_s) if file_metadata.alternate_ids.present?
+
+        solr_doc['original_filename_tesi']  = file_metadata.original_filename if file_metadata.original_filename.present?
+        solr_doc['original_filename_ssi']   = file_metadata.original_filename if file_metadata.original_filename.present?
+        solr_doc['mime_type_tesi']          = file_metadata.mime_type if file_metadata.mime_type.present?
+        solr_doc['mime_type_ssi']           = file_metadata.mime_type if file_metadata.mime_type.present?
+
+        solr_doc['file_format_tesim']       = file_format(file_metadata)
+        solr_doc['file_format_sim']         = file_format(file_metadata)
+        solr_doc['file_size_lts']           = file_metadata.size[0]
+        solr_doc['type_tesim']              = file_metadata.type if file_metadata.type.present?
+
+        # attributes set by fits
+        solr_doc['format_label_tesim']      = file_metadata.format_label if file_metadata.format_label.present?
+        solr_doc['size_tesim']              = file_metadata.size if file_metadata.size.present?
+        solr_doc['well_formed_tesim']       = file_metadata.well_formed if file_metadata.well_formed.present?
+        solr_doc['valid_tesim']             = file_metadata.valid if file_metadata.valid.present?
+        solr_doc['fits_version_tesim']      = file_metadata.fits_version if file_metadata.fits_version.present?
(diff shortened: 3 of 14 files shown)


app/actors/hyrax/actors/file_actor.rb
```diff
@@ -87,7 +87,9 @@ def perform_ingest_file_through_valkyrie(io)
         create_version(file_metadata, user)
         id = file_metadata.file_identifier
         file_set.file_ids << id
+        file_set.original_file_id = id
         Hyrax.persister.save(resource: file_set)
+        Hyrax.publisher.publish('object.metadata.updated', object: file_set, user: user)
```
elrayle, line 92: Not convinced we need to do the indexing here.  It will be done again after the ingest is complete.  This one creates a super small solr doc because most of the fields are populated from the file metadata.

app/indexers/hyrax/active_fedora_file_set_indexer.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+module Hyrax
+  class ActiveFedoraFileSetIndexer < ActiveFedora::IndexingService
+    include Hyrax::IndexesThumbnails
+    include Hyrax::IndexesBasicMetadata
+    STORED_LONG = ActiveFedora::Indexing::Descriptor.new(:long, :stored)
+
+    def generate_solr_document # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
elrayle, line 8: This is the most problematic change because it changes the public API.  This change moves `Hyrax::FileSetIndexer` to `Hyrax::ActiveFedoraFileSetIndexer`.  This is needed because following naming conventions, `Hyrax::FileSet` (valkyrie resource) expects its indexer to be named `Hyrax::FileSetIndexer`.

This is a straight move and rename of the class.  The code is unchanged.

app/indexers/hyrax/file_set_indexer.rb
```diff
@@ -1,57 +1,115 @@
...
-    STORED_LONG = ActiveFedora::Indexing::Descriptor.new(:long, :stored)
-
-    def generate_solr_document # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
-      super.tap do |solr_doc|
-        solr_doc['hasRelatedMediaFragment_ssim'] = object.representative_id
-        solr_doc['hasRelatedImage_ssim'] = object.thumbnail_id
+  ##
+  # Indexes Hyrax::FileSet objects
+  class FileSetIndexer < Hyrax::ValkyrieIndexer
+    include Hyrax::ResourceIndexer
```
elrayle, line 7: This is the refactor of `Hyrax::FileSetIndexer` to be the indexer for `Hyrax::FileSet` (valkyrie resource) instead of `::FileSet` (AF object).  This is needed because `Hyrax::ValkyrieIndexer #for` method uses naming convention to determine the indexer to use for a given Valkyrie resource.

app/indexers/hyrax/file_set_indexer.rb
```diff
@@ -1,57 +1,115 @@
...
+        solr_doc['mime_type_tesi']          = file_metadata.mime_type if file_metadata.mime_type.present?
+        solr_doc['mime_type_ssi']           = file_metadata.mime_type if file_metadata.mime_type.present?
+
+        solr_doc['file_format_tesim']       = file_format(file_metadata)
+        solr_doc['file_format_sim']         = file_format(file_metadata)
+        solr_doc['file_size_lts']           = file_metadata.size[0]
+        solr_doc['type_tesim']              = file_metadata.type if file_metadata.type.present?
+
+        # attributes set by fits
+        solr_doc['format_label_tesim']      = file_metadata.format_label if file_metadata.format_label.present?
```
elrayle, line 41: Do we want to include all file metadata in the solr_doc?  Right now, that is what this does.

app/indexers/hyrax/file_set_indexer.rb
```diff
@@ -1,57 +1,115 @@
...
+        solr_doc['size_tesim']              = file_metadata.size if file_metadata.size.present?
```
elrayle, line 42: Size is recorded in the solr doc twice.  At line 37, it is recorded as a long.  _(Consistent with how it is done in the AF generated solr doc.)_ Here is is recorded as a string. _(Consistent with how all the other FileMetadata fields are indexed.)_

app/models/concerns/hyrax/file_set/indexing.rb
```diff
@@ -6,7 +6,7 @@ module Indexing
 
       included do
         # the default indexing service
-        self.indexer = Hyrax::FileSetIndexer
+        self.indexer = Hyrax::ActiveFedoraFileSetIndexer
```
elrayle, line 9: Completes the refactor moving the AF indexing to `Hyrax::ActiveFedoraFileSetIndexer` _(for AF objects)_ out of `Hyrax::FileSetIndexer` _(now being used for Valkyrie resource `Hyrax::FileSet`)_.

app/models/hyrax/file_set.rb
```diff
@@ -7,6 +7,7 @@ module Hyrax
   # @see https://wiki.duraspace.org/display/samvera/Hydra%3A%3AWorks+Shared+Modeling
   class FileSet < Hyrax::Resource
     include Hyrax::Schema(:core_metadata)
+    include Hyrax::Schema(:basic_metadata)
```
elrayle, line 10: `Hyrax::FileSet` includes `basic_metadata`, so it should also index it.

app/models/job_io_wrapper.rb
```diff
@@ -63,7 +63,7 @@ def size
     nil # unable to determine
   end
 
-  def file_set(use_valkyrie: false)
+  def file_set(use_valkyrie: Hyrax.config.query_index_from_valkyrie)
```
elrayle, line 66: This allows setting `use_valkyrie` from a single point.  We may want to consider this for other places that have a `use_valkyrie` parameter.

spec/actors/hyrax/actors/create_with_remote_files_actor_spec.rb
```diff
@@ -134,4 +134,114 @@
       expect(actor.send(:validate_remote_url, URI('https://example.com/test.txt'))).to be true
     end
   end
+
+  context 'when work is a valkyrie resource' do
+    let(:work) { valkyrie_create(:monograph) }
```
elrayle, line 139: These are the same tests as above, but run with a Valkyrie resource instead of AF Object.

spec/indexers/hyrax/active_fedora_file_set_indexer_spec.rb
```diff
@@ -0,0 +1,151 @@
+# frozen_string_literal: true
+RSpec.describe Hyrax::ActiveFedoraFileSetIndexer do
+  include Hyrax::FactoryHelpers
+
+  let(:file_set) do
```
elrayle, line 5: This is a straight up move of the code that was in `file_set_indexer_spec.rb` _(which used to test the indexer for AF `::FileSet`, but now tests indexer for Valkyrie `Hyrax::FileSet`)_.

spec/indexers/hyrax/file_set_indexer_spec.rb
```diff
@@ -2,148 +2,249 @@
 RSpec.describe Hyrax::FileSetIndexer do
   include Hyrax::FactoryHelpers
 
+  let(:fileset_id) { 'fs1' }
```
elrayle, line 5: This spec used to test the indexer for the AF `::FileSet`, but now tests the indexer for the Valkyrie `Hyrax::FileSet`.

app/models/hyrax/work.rb
```diff
@@ -14,6 +14,8 @@ class Work < Hyrax::Resource
     attribute :on_behalf_of,             Valkyrie::Types::String
     attribute :proxy_depositor,          Valkyrie::Types::String
     attribute :state,                    Valkyrie::Types::URI.default(Hyrax::ResourceStatus::ACTIVE)
+    attribute :representative_id,        Valkyrie::Types::ID
+    attribute :thumbnail_id,             Valkyrie::Types::ID
```
elrayle, line 18: Required for processing in [FileSetActor #attach_to_valkyrie_work](https://github.com/samvera/hyrax/blob/master/app/actors/hyrax/actors/file_set_actor.rb#L92-L93).

app/indexers/hyrax/file_set_indexer.rb
```diff
@@ -1,57 +1,115 @@
 # frozen_string_literal: true
+
 module Hyrax
-  class FileSetIndexer < ActiveFedora::IndexingService
```
no-reply, line 3: i'm worried about this name swap. this would be a high-impact breaking change, right?

app/indexers/hyrax/active_fedora_file_set_indexer.rb
no-reply: ah yeah. my other comment. i think i'd rather leave the old class exactly as it was, and introduce special resolution logic for this case.

we could always realign the names later, after the old class has been deprecated and removed.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -87,7 +87,7 @@ def attach_to_work(work, file_set_params = {})
       def attach_to_valkyrie_work(work, file_set_params)
         work = Hyrax.query_service.find_by(id: work.id) unless work.new_record
         file_set.visibility = work.visibility unless assign_visibility?(file_set_params)
-        Hyrax.persister.save(resource: file_set)
+        @file_set = Hyrax.persister.save(resource: file_set)
```
no-reply, line 90: can you say more about this instance variable?

it will be cached with this actor in the `ActionDispatch::MiddlewareStack` pool, which will hold it in memory until the next time its set. this can lead to unexpected behavior. see #3722

app/actors/hyrax/actors/file_actor.rb
no-reply: if `file_set`'s metadata is updating, i think publishing the event is the right idea. we want listeners to be able to handle events as they prefer, and shouldn't try to regulate behavior by declining to publish.

is your concern about performance due to unneeded index calls/commits? or something else?

app/indexers/hyrax/valkyrie_file_set_indexer.rb
```diff
@@ -0,0 +1,116 @@
...
+        solr_doc['mime_type_tesi']          = file_metadata.mime_type if file_metadata.mime_type.present?
+        solr_doc['mime_type_ssi']           = file_metadata.mime_type if file_metadata.mime_type.present?
+
+        solr_doc['file_format_tesim']       = file_format(file_metadata)
+        solr_doc['file_format_sim']         = file_format(file_metadata)
+        solr_doc['file_size_lts']           = file_metadata.size[0]
+        solr_doc['type_tesim']              = file_metadata.type if file_metadata.type.present?
+
+        # attributes set by fits
+        solr_doc['format_label_tesim']      = file_metadata.format_label if file_metadata.format_label.present?
```
elrayle, line 41: Do we want to include all file metadata in the solr_doc? Right now, that is what this does.

app/indexers/hyrax/valkyrie_file_set_indexer.rb
```diff
@@ -0,0 +1,116 @@
...
+        solr_doc['size_tesim']              = file_metadata.size if file_metadata.size.present?
```
elrayle, line 42: Size is recorded in the solr doc twice.  
* At line 37, it is recorded as a long.  _(Consistent with how it is done in the AF generated solr doc.)_ 
* Here is is recorded as a string. _(Consistent with how all the other FileMetadata fields are indexed.)_

app/indexers/hyrax/valkyrie_indexer.rb
```diff
@@ -95,7 +95,12 @@ class << self
       # @example
       #     ValkyrieIndexer.for(resource: Book.new) # => #<BookIndexer ...>
       def for(resource:)
-        indexer_class_for(resource).new(resource: resource)
+        case resource
+        when Hyrax::FileSet
+          Hyrax::ValkyrieFileSetIndexer
```
elrayle, line 100: This is the 'special resolution logic' to allow `Hyrax::FileSet` to have an indexer that doesn't follow the naming convention.

app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -87,15 +86,18 @@ def attach_to_work(work, file_set_params = {})
...
-        work.thumbnail_id = file_set.id if work.thumbnail_id.blank?
+        fs = Hyrax.persister.save(resource: file_set)
+        Hyrax.publisher.publish('object.metadata.updated', object: fs, user: user)
+        work.member_ids << fs.id
+        work.representative_id = fs.id if work.representative_id.blank?
+        work.thumbnail_id = fs.id if work.thumbnail_id.blank?
         # Save the work so the association between the work and the file_set is persisted (head_id)
         # NOTE: the work may not be valid, in which case this save doesn't do anything.
         Hyrax.persister.save(resource: work)
+        Hyrax.publisher.publish('object.metadata.updated', object: work, user: user)
```
elrayle, line 97: Without this publish, member_ids for the work doesn't get updated and there is no connection between the work and the file set.

app/indexers/hyrax/valkyrie_file_set_indexer.rb
no-reply: 👍🏻  seems okay to add a lot now and try to tidy up another time.

app/models/hyrax/work.rb
no-reply: there was some discussion about other ways of doing this in slack, which seemed to land on sticking with this approach.

no-reply (approved):

Issue #4782, opened 2021-03-12
Only authorized users should be able to edit embargoes


maxkadel:
### Descriptive summary
Hyrax v2.9.3

Currently the embargoes_controller_behavior does not authorize the user on the edit method.

### Rationale

If a user does not have permissions (via CanCanCan) to edit an embargo, they can still go directly to the edit route and edit the embargo. 

### Expected behavior
* A user who is not authorized to edit `Hydra::AccessControls::Embargo` 
  * cannot access the `/embargoes/#{work.id}/edit` path
  * cannot update a work's embargo

### Actual behavior
* A user who is not authorized to edit `Hydra::AccessControls::Embargo` 
  * can access the `/embargoes/#{work.id}/edit` path
  * can update a work's embargo

### Steps to reproduce the behavior

1. WIP
2. WIP

no-reply:
@jlhardes @samvera/hyrax-repo-managers: there's some discussion about the scope of permissions that are needed here.

it seems to be true that the current embargo and lease controllers don't bother checking for authentication at all--which seems like a serious issue.

@maxkadel's proposed solution in #4789 is to limit embargo editing to admin users (it sounds like the client on DCE's side is advocating that approach). for other permissions, i think we allow users with edit access to the relevant object to edit the permission/ACL---that could be an alternative here?

who should be able to edit existing embargoes/leases in a default Hyrax?

maxkadel:
I agree that the fact that the embargo and lease controllers don't check permissions is the central issue here, and that probably my approach of making it admin-only is over-generalizing. 

I think I'm having trouble articulating the issue as a "rails engine" issue, and I've shared the particular use-case in which this has come up that highlights the issue. 

To me, since there are parts of the code that scope all the way down to `Hydra::AccessControls::Embargo` and `Hydra::AccessControls::Lease` it would make sense to have the controllers check against these qualities.

I'm not sure with CanCanCan what this would look like for the following examples:

* A user who can edit a specific work, and who can edit that object's Embargo
* A user who can edit a specific work, and who cannot edit any Embargoes
* A user who can edit any and all works, but not their Embargoes
* A user who can edit all works and their Embargoes

no-reply:
@maxkadel thanks, this is a helpful comment.

there's a part of this that seems straightforward: the embargo and lease controllers provided by hyrax should enforce application `::Ability` on controller actions. i think that's true whatever hyrax's default abilities are or should be.

[...]

PR #4789, opened 2021-03-15
Adds tests for Hydra::AccessControls::Embargo behavior

maxkadel:
refs #4782

Adds one passing and one failing test for Hydra::AccessControls::Embargo behavior. Currently, non-admin-users cannot view an embargo, but *can* edit that same embargo. 

This means that, while a user might not be able to navigate to an embargo via the UI, if they know the appropriate url, they can go directly to it and edit the embargo.

* Adds tests for the expected behavior
* [future] Gets those tests passing
*

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
@@ -49,6 +49,7 @@ module Hyrax
       add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
       add_breadcrumb t(:'hyrax.embargoes.index.manage_embargoes'), hyrax.embargoes_path
       add_breadcrumb t(:'hyrax.embargoes.edit.embargo_update'), '#'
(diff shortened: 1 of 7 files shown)


no-reply:
hi @maxkadel! thanks for this contribution!

we normally only merge to `2.x-stable` as a backport from the main branch (`master`). is this something you could work up against the current codebase if the Hyrax Maintenance WG were promising to backport?

maxkadel:
> hi @maxkadel! thanks for this contribution!
> 
> we normally only merge to `2.x-stable` as a backport from the main branch (`master`). is this something you could work up against the current codebase if the Hyrax Maintenance WG were promising to backport?

Ah, ok, I wasn't sure what the pattern was, thanks for the info. 

I'll try to apply the changes to the main branch, but I don't think I'll be able to shepherd it all the way through to completion, because of the complexity of permissions in Hyrax &co. and moving on to slightly different work projects.

maxkadel:
Opened a PR against master instead - see https://github.com/samvera/hyrax/pull/4799

PR #4792, opened 2021-03-16
Add features to support existing/external FCRepo and Postgres instances when deploying the Helm chart

no-reply:
when deploying the application into an environment with an existing FcRepo instance, you can now point the chart at that rather than deploying a new one.

since it may be desirable to deploy multiple hyrax instances on a single FCRepo, we use a name more unique that `/prod` for the Hyrax base path. using the Helm release name is useful, but will run into trouble if multiple releases with the same name are pushed to different namespaces or different clusters. for this case, we provide an `fcrepoBasePathOverride` to allow users to choose something suitably unique.

---

enables use of external/existing postgresql.

this will fail if the database doesn't already exist. when we use an internal postgresql subchart, we count on the chart arguments/initialization to create the database. we can't count on that here. so, in some sense, this is partial work. on the other hand, we can wave our hands a bit and say that if you're going to stand up your own postgres you can create the database while you're at it.

maybe some follow-up work to allow an `ensure-db` init pod is called for?

@samvera/hyrax-code-reviewers


chart/hyrax/README.md
```diff
@@ -53,22 +54,35 @@ The chart populates the following environment variables:
...
 
+## With an external Fedora Commons Repository
+
+By default, this chart deploys a local Fedora Repository, backed by the
+application's Postgresql system.
+
+To use an existing or external `fcrepo` instance, use the chart values:
+
+  - `fcrepo.enabled`: false
+  - `externalFcrepoHost`: "myfedora.hostname.example.com"
```
dunn, line 84: is the fcrepo api unauthenticated by default?

no-reply: that's right. or at least, the upstream chart we're using doesn't turn on authentication.

supporting auth would probably be a desirable feature for both charts, but i'd like to keep it out of scope here.

dunn (approved):

PR #4796, opened 2021-03-16
index Hyrax::Works after changes in metadata when adding files

### LINKED ISSUE #4794: Hyrax::Works do not index file sets
### Descriptive summary

Creating a Monograph, which is_a Hyrax::Work is_a Valkyrie::Resource, with files does not include member_ids_ssim in the solr_doc.

### Rationale

Works should record their members in the solr doc.

### Expected behavior

Creating a Monograph with files should include member_ids_ssim in the solr doc for the Monograph work.

### Actual behavior

Creating a Monograph with files does not include member_ids_ssim in the solr doc for the Monograph work.

#### Prerequisites

Edit `.internal_test_app/config/initializers/hyrax.rb` and set config

```
```

#### Test

1. Login
2. Dashboard -> Works -> Add new work -> select Monograph as the work type
3. Fill in required fields
4. Add a file
5. Save (forwards to the show page for the Monograph work)
6. Click Works in the breadcrumb to return to Dashboard -> Works

View the solr doc in the Valkyrie index core.

### Related Issues

PR #4780 Add indexing for Hyrax::FileSet through valkyrie indexing 

Issue #4795 - Hyrax::Works do not show Date Added in Dashboard -> Works
Issue #4788 - Hyrax::Works do not show thumbnail in Dashboard -> Works
Issue #4487 - Solr docs generated by ActiveFedora and Hyrax::ValkyrieWorkIndexer are not compatible

elrayle:
Fixes #4794

WorkUploadsHandler was not persisting or reindexing works after adding files as members.  Now it does.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/work_uploads_handler.rb
+++ b/app/services/hyrax/work_uploads_handler.rb
@@ -88,7 +88,8 @@ module Hyrax
 
       acquire_lock_for(work.id) do
         event_payloads = files.each_with_object([]) { |file, arry| arry << make_file_set_and_ingest(file) }
-        Hyrax.persister.save(resource: work)
+        @persister.save(resource: work)
+        Hyrax.publisher.publish('object.metadata.updated', object: work, user: files.first.user)
         event_payloads.each { |payload| Hyrax.publisher.publish('file.set.attached', payload) }
       end
     end
@@ -105,6 +106,7 @@ module Hyrax
       # copy ACLs; should we also be propogating embargo/lease?
       Hyrax::AccessControlList.copy_permissions(source: target_permissions, target: file_set)
       append_to_work(file_set)
+
       IngestJob.perform_later(wrap_file(file, file_set))
       Hyrax.publisher.publish('object.metadata.updated', object: file_set, user: file.user)
       { file_set: file_set, user: file.user }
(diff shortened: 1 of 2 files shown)


app/services/hyrax/work_uploads_handler.rb
```diff
@@ -114,10 +115,12 @@ def make_file_set_and_ingest(file)
     # @api private
     #
     # @todo figure out how to know less about Work's ideas about FileSet use here. Maybe post-Wings, work.
-    def append_to_work(file_set)
+    def append_to_work(file_set, user)
       work.member_ids << file_set.id
       work.representative_id = file_set.id if work.respond_to?(:representative_id) && work.representative_id.blank?
       work.thumbnail_id = file_set.id if work.respond_to?(:thumbnail_id) && work.thumbnail_id.blank?
+      @persister.save(resource: work)
+      Hyrax.publisher.publish('object.metadata.updated', object: work, user: user)
```
no-reply, line 123: should it be a `file.set.attached` event instead?

no-reply:
i think i might be confused.

shouldn't `#attach` handle the behavior you're looking for?

no-reply (approved):
thanks for this!

app/services/hyrax/work_uploads_handler.rb
elrayle: `#on_file_set_attached` initiates `FileSetAttachedEventJob` for a FileSet.  That job creates an event notification that the `FileSet` was attached to the work.  It doesn't reindex the work.

So, I think `file.set.attached` should be published for each file.  And the `object.metadata.updated` should be done once for the work.

elrayle: The file.set.attached event is already being initiated.  The only change needed was to move `object.metadata.updated` for the work outside the files loop block so the indexing happens only once.

PR #4797, opened 2021-03-16
add a db:create step to the helm deployment process

no-reply:
ensures that the configured database exists by adding a `rake db:create`
invocation to the `db-setup` initContainer.

in the event we're trying to deploy hyrax against a database engine that it has
permissions for, but where the configured database doesn't exist, this will
ensure it gets added before we try to run the migrations.

see also #4792

@samvera/hyrax-code-reviewers

### DIFF
--- a/bin/db-migrate-seed.sh
+++ b/bin/db-migrate-seed.sh
@@ -2,8 +2,10 @@
 set -e
 
 db-wait.sh "$DB_HOST:$DB_PORT"
+bundle exec rails db:create
+bundle exec rails db:migrate
+
 db-wait.sh "$FCREPO_HOST:$FCREPO_PORT"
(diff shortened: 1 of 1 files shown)


dunn:
should it be part of the `db-seed` script?

no-reply:
> should it be part of the `db-seed` script?

i kept it out because that script is used by the docker-compose environment (which doesn't need this), and waits for Solr and Fcrepo (which isn't necessary). possibly the whole thing should be restructured?

no-reply:
i considered that maybe my hesitation about adding it to the script is that it would have required a whole docker build, while the original change only changed the chart. since that's a bad reason to avoid a change...

@dunn what do you think about this version.

dunn (approved):

PR #4799, opened 2021-03-17
Only admins should be able to edit embargoes

maxkadel:
refs #4782

Currently, non-admin-users cannot view an embargo, but *can* edit that same embargo. This is counter-intuitive - "can :index" is usually the most conservative permission *granted*; however, if you are *removing* permissions, which the "cannot" code does, the ability does not flow to restrict more-permissive restrictions from `:index`, so removing the ability to `:index` only removes that ability, not any others.

This means that, while a user might not be able to navigate to an embargo via the UI, if they know the appropriate url, they can go directly to it and edit the embargo.

Previously opened this as a PR against 2.x, and was asked to do it against master for future back porting instead. (see https://github.com/samvera/hyrax/pull/4789)

* Adds tests for the expected behavior
* [future] Gets those tests passing
  * First commit adds temporary "puts" to explain confusion with test - it seems that somehow the user groups are not available or have been deleted in the form context, and I don't know why

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -309,9 +309,8 @@ module Hyrax
       # user can version if they can edit
       alias_action :versions, to: :update
       alias_action :file_manager, to: :update
-
       return if admin?
-      cannot :index, Hydra::AccessControls::Embargo
+      cannot :manage, Hydra::AccessControls::Embargo
       cannot :index, Hydra::AccessControls::Lease
     end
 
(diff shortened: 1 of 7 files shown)


app/models/concerns/hyrax/ability.rb
```diff
@@ -309,9 +309,8 @@ def curation_concerns_permissions
       # user can version if they can edit
       alias_action :versions, to: :update
       alias_action :file_manager, to: :update
-
       return if admin?
-      cannot :index, Hydra::AccessControls::Embargo
+      cannot :manage, Hydra::AccessControls::Embargo
```
maxkadel, line 313: Need deeper conversation on who exactly should be able to edit an embargo.

no-reply: I've tried to launch the discussion in the originating issue at #4782. 

i think #4801 might be a fix for the test suite issue this PR ran into.

PR #4805, opened 2021-03-18
refactor embargo and lease expiry jobs to use valkyrie

no-reply:
these jobs were recently written using ActiveFedora. these versions retain the existing API and behavior, but use Valkyrie for the implementation.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/hyrax/transactions/steps/save_access_control_spec.rb
+++ b/spec/hyrax/transactions/steps/save_access_control_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe Hyrax::Transactions::Steps::SaveAccessControl, valkyrie_adapter: 
 
     it 'persists the new permissions' do
       expect { step.call(work) }
-        .to change { Hyrax::AccessControlList(work).permissions }
+        .to change { Hyrax::AccessControlList.new(resource: work).permissions }
         .to contain_exactly(have_attributes(mode: :read, agent: user.user_key))
     end
   end
--- a/app/jobs/embargo_expiry_job.rb
+++ b/app/jobs/embargo_expiry_job.rb
@@ -1,15 +1,17 @@
 # frozen_string_literal: true
 class EmbargoExpiryJob < Hyrax::ApplicationJob
   def perform
-    records_with_expired_embargos.each do |id|
-      work = ActiveFedora::Base.find(id)
-      Hyrax::Actors::EmbargoActor.new(work).destroy
+    records_with_expired_embargos.each do |resource|
+      Hyrax::EmbargoManager.release_embargo_for(resource: resource) &&
+        Hyrax::AccessControlList(resource).save
     end
   end
 
   ##
-  # @return [Enumerator<String>] ids for all the objects that have expired active embargoes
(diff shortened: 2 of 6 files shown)


spec/hyrax/transactions/steps/save_access_control_spec.rb
```diff
@@ -17,7 +17,7 @@
 
     it 'persists the new permissions' do
       expect { step.call(work) }
-        .to change { Hyrax::AccessControlList(work).permissions }
+        .to change { Hyrax::AccessControlList.new(resource: work).permissions }
```
cjcolvar, line 20: Why do you switch to using `new` here when you use the class level static method in the jobs?  What is the difference?

no-reply: it's confusing to me too, and i'm glad you asked.

here's what's going on:

each `Hyrax::Resource` instance carries around a `permission_manager` that holds it's own in-memory acl state. this is is mainly to support on-object permission editing via, e.g. `work.visibility = 'open'` and `work.visibility.read_groups = ['public']`.

in general we're (i'm) gently recommending that people try to avoid relying on this state being passed around with the model object, instead querying for up to date permissions or using change sets, but making this a clean break internally hasn't been simple. particularly, wings really needs the on-model permissions data because AF tries ensure persisted ACLs are updated on every object save. it's also awfully convenient for forms. so we have `Hyrax::Resource#permission_manager`.

that leaves us with the decision, everywhere, between using the on-object cached permissions data or querying the persistence layer. which is best depends on what else the code has been up to. basically a cache invalidation problem. this change involved me changing `Hyrax::AccessControlList(resource)` to use the resource's cached data preferentially (at https://github.com/samvera/hyrax/pull/4805/files#diff-67f2f818ab37b14eeb09c92a031dddc0d0b5f8c6094ff259a3ba128701bc526aR17). possibly/arguably, that isn't the best. the alternative is to use `resource.permission_manager.acl.save` for the persistence at https://github.com/samvera/hyrax/pull/4805/files#diff-7fd8fb93f0112844e5bd8668541e77fd258017e38c8eb67694e9b31a5ccb2d66R6 and in the same place in the other job.

basically i felt like telling folks to use *that* magic invocation wasn't very kind, and it was better to let the `Hyrax::AccessControlList()` type-caster take care of it. the message i want to be able to put out there is "use `Hyrax::AccessControlList(resource)` to get the ACLs for the resource object. this work flipped me on which of the disk or cached version "the ACLs for the resource object" ought to refer to. the only thing that broke was this one test, which now needs to look in the persistence layer for the pre-cache data, since `step.call(work)` (what we're testing here is whether the code under test saves the data from the cache).

jeremyf: 👍 though I want to make sure that @cjcolvar can review.

cjcolvar: @no-reply Thanks for this thorough and clear explanation!  This makes perfect sense now.

cjcolvar (approved):

Issue #2657, opened 2018-02-16
Embargoes and Leases do not automatically expire


chrisdaaz:
### Descriptive summary

Embargoes and leases do not auto-expire, as currently implemented. You have to expire them in the UI. (And I don't believe the UI is fully baked yet.)

### Expected behavior

Embargoes and leases should auto-expire according to the deposit settings.

### Related work

This issue is moved from: https://github.com/curationexperts/nurax/issues/90

mjgiarlo:
I'm not sure they are currently designed to auto-expire. AFAIK, they've always required human intervention. This may not be a bug but a change in feature behavior. We should get folks' input on this change, particularly the @samvera/hyrax-repo-managers

newmanld:
My input is that most repo-managers would like notifications sent to users of embargoes/leases about to expire (frequency to be configurable?) combined with an auto-expiration function.  I suppose some might want to turn auto-expiration off, if they have a repo manager with the time to do this manually, but I think most will not.

vantuyls:
i agree with @newmanld's assessment. though i would strongly encourage expiration (auto or manual) to be configurable/switchable so that people can do with this what they will. I've heard from a few folks (maybe @mark-dce) who have indicated that their use cases want manual removal of embargoes.

mark-dce:
If you have really sensitive stuff embargoed, the most conservative approach is to notify staff that an embargo has expired and then have the embargo manually lifted once the expiration has been verified.  This is definitely the less frequent case, but we've seen it at for at least two institutions and I think it's pretty important to be able to support this use case without too much extra work.  A configuration option would work.  At present, we've used a cron job running a rake task for those institutions wanting auto-expirations - would documenting that solution be a reasonable response to this ticket?

no-reply:
I'd like to break this ticket down to two separate tickets:

 - [ ] Implement embargo/lease expiry notifications.
   - [ ] Notify (who???) when an embargo is set to expire N days from now.
   - [ ] Make notifications configurable.
  - [ ] Ship an embargo/lease management rake task and document its use in the Hyrax Management Guide.
    - [ ] Task sends notifications, and;
    - [ ] optionally (by argument): Lifts expired embargos and leases.

As far as I can see, embargo and lease is completely undocumented in the Management guide to date. The embargo APIs are less than ergonomic, and under documented. I've seen at least one custom task in the wild that uses `deactivate!` to try to lift embargoes, to no effect.

[...]

PR #4806, opened 2021-03-18
add embargo and lease expiry rake tasks; schedule  for helm deploys

no-reply:
ensure embargo and lease expire daily (by default ) in Helm. adds a task to clear the current expired embargos and leases daily.
@samvera/hyrax-code-reviewers


no-reply:
related to #2657.

cjcolvar (approved):
:+1: Looks good to me.

Issue #3570, opened 2019-03-11
Enable valkyrie shared specs on wings query service


hackartisan:
Initial specs, run on resources created from ActiveFedora objects:
- [x] find_all #3581
- [x] find_all_of_model #3615
- [x] find_by
- [x] find_by_alternate_identifier #3620
- [x] find_many_by_ids #3614
- [x] find_members #3622 
- [x] find_references_by #3625
- [x] find_inverse_references_by #3694
- [x] find_parents #3705

Then:
- [ ] analyze the ones that are still pending; are there some use cases we can't accommodate?
- if not
  - [ ] copy in the full shared spec in a new context, run it on native valkyrie resources.
  - [ ] turn on the `behaves_like` and delete the shared specs context
- [ ] If so, replicate shared specs as closely as we can for some subset?
- [ ] weed out local tests that don't test extra assumptions or requirements

laritakr:
it { is_expected.to respond_to(:find_all).with(0).arguments }
  it { is_expected.to respond_to(:find_all_of_model).with_keywords(:model) }
  it { is_expected.to respond_to(:find_by).with_keywords(:id) }
  it { is_expected.to respond_to(:find_by_alternate_identifier).with_keywords(:alternate_identifier) }
  it { is_expected.to respond_to(:find_many_by_ids).with_keywords(:ids) }
  it { is_expected.to respond_to(:find_members).with_keywords(:resource, :model) }
  it { is_expected.to respond_to(:find_references_by).with_keywords(:resource, :property) }
  it { is_expected.to respond_to(:find_inverse_references_by).with_keywords(:resource, :property) }
  it { is_expected.to respond_to(:find_inverse_references_by).with_keywords(:id, :property) }
  it { is_expected.to respond_to(:find_parents).with_keywords(:resource) }

hackmastera:
In general, do we want to try to pull from fedora (via ActiveFedora) or solr (via ActiveFedora)?

hackartisan:
The remainder of this issue is about simplifying the wings test suite. Unless @no-reply wants to fold this into #6137 I propose we close it, since Wings is bridge code and the specs are slated for future deletion.

no-reply:
agreed that we should close this.

the requirement for the Wings query service is that it enables Hyrax to pass its test suite and QA process. if the Valkyrie shared specs help us with that, great, but if they haven't so far i don't see any value in jumping through hoops.

PR #4810, opened 2021-03-18
Wings: add support for 'count all of model` 

no-reply:
reduce failing tests for Valkyrie queries to 11.

related to #3570

some refactors are included in a second commit. maybe some small-ish performance optimizations there, but mostly a lot of changes for code readability. please review with that in mind.

@samvera/hyrax-code-reviewers


no-reply:
@elrayle i wonder if you have thoughts about the issue of `Hyrax::ObjectNotFoundError` vs `Valkryie::Persistence::ObjectNotFoundError`.

PR #4819, opened 2021-03-19
support Valkyrie in `WorksControllerBehavior#destroy`

no-reply:
adds transactional support for the destroy action. publishes the `object.deleted` event when destroying the object. this replaces
the older `:after_destroy` callback. the callback has been merely a wrapper to the publisher/listner event for some time.

to support this, add edit abilities for `Hyrax::Resource`. the edit rules for `Hyrax::Resource` should be the same as those used for
`ActiveFedora::Base`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/valkyrie/persister.rb
+++ b/lib/wings/valkyrie/persister.rb
@@ -53,6 +53,7 @@ module Wings
         af_object = ActiveFedora::Base.new
         af_object.id = resource.alternate_ids.first.to_s
         af_object.delete
+        resource
       end
 
       # Deletes all resources from Fedora and Solr
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -113,10 +113,21 @@ module Hyrax
     end
 
     def destroy
-      title = curation_concern.to_s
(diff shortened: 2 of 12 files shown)


lib/wings/valkyrie/persister.rb
```diff
@@ -53,6 +53,7 @@ def delete(resource:)
         af_object = ActiveFedora::Base.new
         af_object.id = resource.alternate_ids.first.to_s
         af_object.delete
+        resource
```
jeremyf, line 56: I _think_ this is okay, but it changes the expected returning value.  I'm thinking of the AR pattern (e.g. `#delete` returns a boolean).  However, this is not at the same place.

I also suspect that nothing upstream is checking on the return value of this method.

Thoughts?

no-reply: yes, this brings the return value in line with Valkyrie's interface. see https://github.com/samvera/valkyrie/blob/master/lib/valkyrie/persistence/postgres/persister.rb#L57

i think the missing test coverage is due to the valkyrie shared examples not being run (since order and several days types are unsupported). I'm running those tests now regularly as i work on order support. this maybe risks regression, but nothing in the hyrax code base seems to count on this return type, so it feels like a small risk(?).

the old return value was the AF object.

jeremyf (approved):

PR #4822, opened 2021-03-22
Make collection-related search builders inherit from the base hyrax collection search builder

bkiahstroud:
This changes all Collection-related `SearchBuilder`s to inherit from the "base" `Hyrax::CollectionSearchBuilder`. This allows for consistent behavior across Collection queries as well as less repeated code. 

@samvera/hyrax-code-reviewers


no-reply (approved):
thanks for your patience @bkiahstroud.

this looks great! confirming that the test coverage for each of these modules looks pretty robust.

these will all still inherit `::SearchBuilder`.

i'm holding merge for a moment to get #4846 out the door.

PR #4823, opened 2021-03-22
install ffmpeg in `hyrax-engine-dev-worker`

no-reply:
`ffmpeg` for media transcoding in derivatives jobs.

@samvera/hyrax-code-reviewers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,9 +63,10 @@ ENV MALLOC_ARENA_MAX=2
 
 USER root
 RUN apk --no-cache add bash \
+  ffmpeg \
+  mediainfo \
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
FWIW Avalon is installing ffmpeg from a static build (https://github.com/avalonmediasystem/avalon/blob/master/Dockerfile#L32-L34) and mediainfo from mediaarea directly (https://github.com/avalonmediasystem/avalon/blob/master/Dockerfile#L41).  I think the upstream repo versions were too old at the time it got setup or for ffmpeg were missing support for certain media types. @phuongdh might remember more.

PR #4825, opened 2021-03-23
wings: handle ActiveFedora associations/reflections deliberately

no-reply:
the older approach was a little bit too trial-and-error for real use cases. this attempts to handle the reflection types used in typical Hyrax usage specifically. there are probably performance gains to be had by optimizing handling for the individual branches.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/attribute_transformer.rb
+++ b/lib/wings/attribute_transformer.rb
@@ -15,14 +15,17 @@ module Wings
       end
     end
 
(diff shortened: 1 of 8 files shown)


no-reply:
loose benchmarking suggests this is a wash, performance wise, but it's a much more organized approach to mapping reflection/association attributes.

jeremyf (approved):

PR #4826, opened 2021-03-24
wings: fix serveral shared example specs for Valkyrie persister

no-reply:
this reduces the space of shared example failures to type issues, order
issues, and nesting issues. 22 tests fail in total.

RDF.rb and ActiveFedora cast many RDF types to Ruby objects, which is mostly
helpful, and we don't have plans to circumvent this to pass Valkyrie tests.

ordered attribute support is forthcoming, and nesting is to be explored.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/converter_value_mapper.rb
+++ b/lib/wings/converter_value_mapper.rb
@@ -73,7 +73,7 @@ module Wings
     end
 
     def result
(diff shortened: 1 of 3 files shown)


elrayle (approved):
Nice to see more tests passing.  Thanks!

PR #4829, opened 2021-03-25
cleanup trophies for valkyrie works

no-reply:
add trophy cleanup when valkyrie works are deleted.

this listener code will likely also cleanly replace the existing trophy cleanup
actor stack behavior, with less risk of unwanted/unimportant errors.

we should consider deprecating the actor stack version in favor of this.

@samvera/hyrax-code-reviewers


elrayle (approved):
Looks good.  Glad to see more Listener events!

PR #4834, opened 2021-03-29
chart: support a worker image

no-reply:
support a worker image and persistent volumes in the helm chart

this setup requires a RWX storage class. i'd like to provide an alternative using S3/minio for object storage for all of these paths ASAP.

@samvera/hyrax-code-reviewers


dlpierce (approved):
I'm not so familiar with charts, but this looks alright to me.

PR #4836, opened 2021-03-29
use valkyrie to handle branding storage

no-reply:
drop direct file and fileutils calls for storage of collection branding
images. use valkyrie instead.

because valkyrie's disk storage adapter defaults to pair tree style paths, we
need a custom path handler to ensure branding files are stored in the same place
as before.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/configuration.rb
+++ b/lib/hyrax/configuration.rb
@@ -386,6 +386,18 @@ module Hyrax
     # @!group Valkyrie
 
     ##
+    # @return [Valkyrie::StorageAdapter]
+    def branding_storage_adapter
+      @branding_storage_adapter ||= Valkyrie::StorageAdapter.find(:branding_disk)
+    end
+
+    ##
+    # @param [#to_sym] adapter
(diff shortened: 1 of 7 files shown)


lib/hyrax/configuration.rb
```diff
@@ -377,6 +377,26 @@ def display_media_download_link?
     # @!endgroup
     # @!group Valkyrie
 
+    ##
+    # @return [Valkyrie::StorageAdapter]
+    def branding_storage_adapter
+      @branding_storage_adapter ||=
+        begin
+          Valkyrie::StorageAdapter.register(
```
no-reply, line 385: i guess this should probably move to an initailizer?

jeremyf: Yeah, that looks like the better pattern.

no-reply: how does this version look?

dlpierce (approved):

PR #4842, opened 2021-03-30
drop shx from .dassie

no-reply:
`shx` wasn't working in the alpine images; so much for portability(?). since it's not needed in that context, it's fine to just exclude it from `.dassie`.

should we also do this in the generator? or is `shx` providing a lot of value there?

@samvera/hyrax-code-reviewers


no-reply:
cc: @dlpierce this might help move forward your current ticket?

jeremyf (approved):

PR #4843, opened 2021-03-30
Wrap riiif initializer in a Reloader

dlpierce:
The configuration contained in an application's `config/initializers/riiif.rb` will (sometimes?) be lost at runtime in development mode causing an incorrect file_resolver to be used when requesting an image via riiif. This will explicitly reload that initializer when the application is reloading code.

Suggested in https://github.com/curationexperts/riiif/issues/119

* Working IIIF image requests in dassie.
* Should have no effect in production.

@samvera/hyrax-code-reviewers


no-reply:
seems good. is this something we should recommend to folks with existing apps with initializers generated by Hyrax?

PR #4849, opened 2021-04-02
Fix solr_home dir in docker-compose

dlpierce:
Fixes the solr_home docker volume mount location that changed in solr 8.

* Dassie's solr index is not lost when restarting the docker-compose stack.

@samvera/hyrax-code-reviewers


no-reply (approved):
we think this mount hasn't been working for some time. possibly due to changes in the upstream image.

PR #4851, opened 2021-04-02
Use samvera/fcrepo4 image

dlpierce:
Change the image used in docker-compose to match the image used in fcrepo-charts.

Fedora version changes from 4.7.2 to 4.7.5 and the base url changes from `/fcrepo/rest` to `/rest`.

@samvera/hyrax-code-reviewers


no-reply (approved):
this brings the dev environment in line with the helm chart

PR #4852, opened 2021-04-02
chart: support extra volumes/mounts evaluated as a template

no-reply:
in the case a particular application needs to add/mount extra volumes, these
values provide a hook for it to do so both on the rails application deployments,
and on the worker pods.

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 0.7.1
+version: 0.8.0
 appVersion: 3.0.1
 dependencies:
   - name: fcrepo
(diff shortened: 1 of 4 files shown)


chart/hyrax/min.yaml
```diff
@@ -0,0 +1,19 @@
+ingress:
```
mcritchlow, line 1: Would it be worth referencing the use of this file in the chart's README?

no-reply: oh, we probably should just drop it.

this was for my local testing and got caught up in an overeager git add

no-reply: squashed out the extra `.yaml` files after porting the example to the inline docs.

dunn (approved):

chart/hyrax/min.yaml
mcritchlow: LGTM!

PR #4856, opened 2021-04-06
docker: fix entrypoint script for sh

dunn:
see
```
In POSIX sh, unary -v (in place of [ -n "${var+x}" ]) is undefined. [SC2039]
```

@samvera/hyrax-code-reviewers


dunn:
in local deployments we currently see:
```
│ hyrax sh: HYRAX_DERIVITIVES_PATH: unknown operand                                                                                    │
│ hyrax sh: HYRAX_UPLOAD_PATH: unknown operand                                                                                         │
```

mcritchlow (approved):

no-reply:
great. we'll need to publish the image after this is merged.

PR #4857, opened 2021-04-06
Allow Hyrax Helm chart to disable fcrepo

mcritchlow:
This adds support in the Hyrax helm chart to disable an `fcrepo` chart dependency installation.

* Add gate around looking for fedora in the `db-seed-migrate.sh` script used in
	the chart's `initContainer`
* Add a new environment variable, disabled by default, to determine whether to
	load the Hyrax engine seed file.
* Add gate around defining fedora env vars in the Chart ConfigMap for
	environment variables.

* `docker build --target hyrax-engine-dev --tag k3d-registry.localhost:41906/dassie:nofedora .`
* `docker push k3d-registry.localhost:41906/dassie:nofedora`
* `helm upgrade --timeout 30m0s --atomic --install --namespace=hyrax-nofedora
--values=hyrax.yaml dassie chart/hyrax`

Values of `hyrax.yaml` were:

```yaml
image:
  repository: k3d-registry.localhost:41906/dassie
  tag: "nofedora"
ingress:
  enabled: false
brandingVolume:
  enabled: false
derivativesVolume:
  enabled: false
uploadsVolume:
  enabled: false
fcrepo:
  enabled: false
loadSolrConfigSet: false
solr:
  enabled: true
livenessProbe:
  enabled: false
readinessProbe:
  enabled: false
worker:
  enabled: false
```

Screenshot of deployment with the values/settings as described above:

image

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/db/seeds.rb
+++ b/.dassie/db/seeds.rb
@@ -6,7 +6,7 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-Hyrax::Engine.load_seed
+Hyrax::Engine.load_seed unless ActiveModel::Type::Boolean.new.cast(ENV["SKIP_HYRAX_ENGINE_SEED"])
 
 puts "\n== Loading users"
 User.where(email: 'admin@example.com').first_or_create do |f|
--- a/chart/hyrax/templates/configmap-env.yaml
+++ b/chart/hyrax/templates/configmap-env.yaml
@@ -24,11 +24,14 @@ data:
   {{- if .Values.redis.enabled }}
   REDIS_HOST: {{ template "hyrax.redis.host" . }}
   {{- end }}
+  {{- if .Values.fcrepo.enabled }}
   FCREPO_BASE_PATH: {{ .Values.fcrepoBasePathOverride | default (printf "/%s" (include "hyrax.fullname" .)) | quote }}
   FCREPO_HOST: {{ template "hyrax.fcrepo.host" . }}
   FCREPO_PORT: {{ .Values.fcrepo.servicePort | default 8080 | quote }}
   FCREPO_REST_PATH: {{ .Values.fcrepo.restPath | default "rest" }}
+  {{- end }}
   REDIS_PROVIDER: SIDEKIQ_REDIS_URL
+  SKIP_HYRAX_ENGINE_SEED: {{  .Values.skipHyraxEngineSeed | default 0 }}
   SOLR_ADMIN_USER: {{ template "hyrax.solr.username" . }}
   SOLR_ADMIN_PASSWORD: {{ template "hyrax.solr.password" . }}
   SOLR_COLLECTION_NAME: {{ template "hyrax.solr.collectionName" . }}
--- a/bin/db-migrate-seed.sh
+++ b/bin/db-migrate-seed.sh
@@ -5,7 +5,9 @@ db-wait.sh "$DB_HOST:$DB_PORT"
 bundle exec rails db:create
 bundle exec rails db:migrate
 
-db-wait.sh "$FCREPO_HOST:$FCREPO_PORT"
+if [ "$FCREPO_HOST" ]; then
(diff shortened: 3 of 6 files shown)


db/seeds.rb
```diff
@@ -1,4 +1,6 @@
-ActiveFedora.fedora.connection.send(:init_base_path)
+if ENV["FCREPO_HOST"].present?
```
no-reply, line 1: we talked about moving this conditional to the application seeds.

avoiding putting it here will help these engine seeds support a wider variety of applications (i.e. those that don't configure with `FCREPO_HOST` should still expect `:init_base_path` to succeed)

.dassie/db/seeds.rb
```diff
@@ -6,7 +6,7 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-Hyrax::Engine.load_seed
+Hyrax::Engine.load_seed if ENV["LOAD_HYRAX_ENGINE_SEED"].present?
```
no-reply, line 9: 👍🏻

chart/hyrax/templates/configmap-env.yaml
```diff
@@ -19,15 +19,20 @@ data:
   HYRAX_CACHE_PATH: "/app/samvera/file_cache"
   HYRAX_UPLOAD_PATH: "/app/samvera/uploads"
   {{- end }}
+  {{- if .Values.loadHyraxEngineSeed }}
+  LOAD_HYRAX_ENGINE_SEED: "true"
```
no-reply, line 23: the extra layer of boolean interpretation added by Go makes me slightly nervous. maybe:

```suggestion
  LOAD_HYRAX_ENGINE_SEED: {{  .Values.loadHyraxEngineSeed | default 0 }}
```

mcritchlow: Yeah, I'm not crazy about it either. This seems like a good alternative, thanks!

.dassie/db/seeds.rb
no-reply: oh, maybe this should be `unless ActiveModel::Type::Boolean.new.cast(ENV['SKIP_HYRAX_ENGINE_SEED'])`?

requiring the variable for the preferred default behavior is probably too much. otherwise we need to add this to the docker-compose's environment variables, right?

chart/hyrax/templates/configmap-env.yaml
```diff
@@ -19,15 +19,20 @@ data:
   HYRAX_CACHE_PATH: "/app/samvera/file_cache"
   HYRAX_UPLOAD_PATH: "/app/samvera/uploads"
   {{- end }}
+  {{- if .Values.loadHyraxEngineSeed }}
```
no-reply, line 22: GitHub wouldn't let me do a multi-line suggestion, but now this is unneeded i think(?)

mcritchlow: I think it is, because the check in the dassie seed file does a `ENV[].present?`. So we don't want any
variable defined in that scenario, right?

no-reply: i think we should change that line to use `ActiveModel::Types::Boolean`, which is how we interpret boolean ENV variables elsewhere in the codebase.

that way users can do an explicit `false` or `0`

mcritchlow: Ok, I'm not familiar with that pattern but I'll look and update once I've got something working using that.

no-reply: see https://api.rubyonrails.org/v6.1.0/classes/ActiveModel/Type/Boolean.html

.dassie/db/seeds.rb
mcritchlow: Sorry, I just saw this and pushed a commit with the `if` in place still. I
admittedly wasn't aware this would have implications for a docker-compose
environment. So I can flip the env var value and use `unless`. I'll do that now.

no-reply (dismissed):

no-reply (approved):

PR #4858, opened 2021-04-06
bump ruby version to 2.7.3

no-reply:
2.7.3 has security fixes. let's test it.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,3 +192,21 @@ workflows:
           bundler_version: "2.1.4"
           requires:
             - build
+  ruby2-7-3:
+    jobs:
+      - bundle:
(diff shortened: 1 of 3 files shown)


no-reply:
in the past, we've done a hard swap over on circleci builds, but i think a more stepwise approach would suit better:

let's merge the new build as optional, give a couple of days to gauge stability and for rebases to happen, then change the GitHub config and remove the 2.7.2 build.

no-reply:
looks like a genuine Ruby 2.7.3 support issue, to me.

cjcolvar:
That's really surprising!  I wouldn't have expected a ruby patch release to cause any test failures.  I guess it shows the value of running tests on _every_ change!

PR #4862, opened 2021-04-08
Use db-worker-wait.sh in chart worker template

mcritchlow:
This allows us to optionally skip Fedora waiting.

* Rely on a new `db-worker-wait.sh` script in worker template initContainer rather than individual call to the `db-wait.sh` script

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/templates/deployment-worker.yaml
+++ b/chart/hyrax/templates/deployment-worker.yaml
@@ -33,9 +33,7 @@ spec:
           command:
             - sh
             - -c
-            - db-wait.sh "$DB_HOST:$DB_PORT"
-            - db-wait.sh "$FCREPO_HOST:$FCREPO_PORT"
-            - db-wait.sh "$SOLR_HOST:$SOLR_PORT"
+            - db-worker-wait.sh
       serviceAccountName: {{ include "hyrax.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
--- a/bin/db-worker-wait.sh
+++ b/bin/db-worker-wait.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+set -e
+
+db-wait.sh "$DB_HOST:$DB_PORT"
+if [ "$FCREPO_HOST" ]; then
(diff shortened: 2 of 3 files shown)


chart/hyrax/templates/deployment-worker.yaml
```diff
@@ -33,9 +33,7 @@ spec:
           command:
             - sh
             - -c
-            - db-wait.sh "$DB_HOST:$DB_PORT"
-            - db-wait.sh "$FCREPO_HOST:$FCREPO_PORT"
-            - db-wait.sh "$SOLR_HOST:$SOLR_PORT"
+            - db-migrate-seed.sh
```
no-reply, line 36: i see the motivation for this, but am wondering whether this is the right solution?

are there downsides to the workers trying to run migrations and seeds? i think in a normal deploy on a responsive cluster, this is going to lead to many pods trying to run the migrations all at once, is that okay?

mcritchlow: I see your point. 


I suppose alternatively we update this to something more along the lines of:

```
- db-wait.sh "$DB_HOST:$DB_PORT" bundle exec rake db:abort_if_pending_migrations
```

no-reply: that seems reasonable.

what happens if we just drop the fcrepo and/or solr checks here completely and do nothing else?

mcritchlow: I took another pass at this with a dedicated script that solely does waiting
now.

no-reply: do you think it would be viable to remove the whole container? i'm noticing that the only DB backend that really ought to matter to `sidekiq` (rather than the individual jobs) isn't actually in this check: `redis`.

no-reply: @mcritchlow and i chatted about this in a back channel and decided that it's *likely* `sidekiq` is going to try to run Rails initializers eagerly, so we'll need these checks.

i'm going to merge, and Matt is going to try a deploy with the whole init container removed to confirm that it is needed.

no-reply (approved):

PR #4863, opened 2021-04-08
Wait for redis in Chart worker initContainer

mcritchlow:
The only database the sidekiq worker really needs to wait on is redis.

Note: I did not change the Chart version since this ultimately does a better job
of doing what my previous PR implemented. So we can publish that chart version
if/when we're happy with this alternative.

Screenshot:
db-wait container successfully waiting on redis:
image

worker pod with a successful db-wait initContainer and a running sidekiq worker container:
image

* Remove the script to check solr and the database before starting the worker container
* Replace with a specific check against redis

@samvera/hyrax-code-reviewers

### DIFF
--- a/bin/db-worker-wait.sh
+++ b/bin/db-worker-wait.sh
@@ -1,8 +1,0 @@
-#!/bin/sh
-set -e
-
-db-wait.sh "$DB_HOST:$DB_PORT"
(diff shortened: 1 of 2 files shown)


dunn (approved):

no-reply (approved):
fantastic! thank you!

PR #4864, opened 2021-04-09
Wait until Solr auth is on before uploading config

mcritchlow:
This is a race/timing issue that can happen intermittently. Adding the
check for a `401` response has proved to address this in local projects.

* Add an additional check before trying to upload the solr configset to ensure
	that auth is enabled/available

@samvera/hyrax-code-reviewers


no-reply:
ah. this is the same timing issue we hit in Surfliner? probably we're hitting it for the first time with this stack because we're running deploys that don't wait on fedora? or...?

mcritchlow:
Yeah, that's my guess. I'm almost done with a trial run locally and will mark
this as Ready unless I messed something up :)

no-reply (approved):

PR #4865, opened 2021-04-09
chart: add environment variables/secret for the db username and pass

no-reply:
in the case that an application wants to connect to the postgresql instance
outside the database configured with `DATABASE_URL` it's handy to have these
values. all this information is already exposed to the application via the
`DATABASE_URL` secret entry, so passing this along shouldn't have any security
implications.

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/templates/secrets.yaml
+++ b/chart/hyrax/templates/secrets.yaml
@@ -7,7 +7,7 @@ metadata:
 type: Opaque
 data:
   SECRET_KEY_BASE: {{ randAlphaNum 20 | b64enc | quote }}
-  DATABASE_PASSWORD: {{ include "hyrax.postgresql.password" . | b64enc }}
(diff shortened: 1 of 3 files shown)


chart/hyrax/templates/secrets.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
 type: Opaque
 data:
   SECRET_KEY_BASE: {{ randAlphaNum 20 | b64enc | quote }}
+  DATABASE_PASSWORD: {{ include "hyrax.postgresql.password" . | b64enc }}
```
no-reply, line 10:
```suggestion
  DB_PASSWORD: {{ include "hyrax.postgresql.password" . | b64enc }}
```

dunn (approved):

PR #4872, opened 2021-04-15
Use Ingress apiVersion 1 for Helm Chart

mcritchlow:
* Update chart version to `0.11.0`
* Upgrade the `ingress.yaml` template to use the `apiVersion` `1`. And update
	configuration accordingly.

Potential Concerns:
* Deployments on clusters with older versions of k8s could break on this change. We could be more conservative and wrap some conditionals into the `ingress.yaml` template around k8s versions. I'm starting with this proposal, as I don't _think_ there are many deployments of the chart in the wild (yet).

I tested with the following configuration:
```yaml
ingress:
  enabled: true
  hosts:
    - host: hyrax.k3d.localhost
      paths:
       - path: /
```

image

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 0.10.2
+version: 0.11.0
 appVersion: 3.0.1
 dependencies:
   - name: fcrepo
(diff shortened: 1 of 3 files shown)


no-reply:
cc: @orangewolf

this drops support for kubernetes versions prior to 1.16 (i think), is that okay with you?

orangewolf:
👍 I means I have to upgrade our staging cluster, but that was on my todo list anyway, so yes, I'm good with it.

orangewolf (approved):

PR #4884, opened 2021-04-22
Make dassie sidekiq depend on services

dlpierce:
When starting dassie with `docker-compose up` with empty volumes (perhaps after using `docker-compose down -v`) the sidekiq worker service would fail to come up with permission problems. This seems to be a race condition because starting the sidekiq service again manually works.

Adding some dependent services to the sidekiq service to match what is used by the app service should help address this by delaying its startup. It also makes bringing just the sidekiq service up (without the web server) easier.

This change will only affect dev environments using docker-compose.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -97,6 +97,13 @@ services:
     env_file:
       - .env
       - .dassie/.env
+    depends_on:
+      - db_migrate
+      - fcrepo
+      - memcached
+      - postgres
+      - redis
+      - solr
     volumes:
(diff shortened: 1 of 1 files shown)


no-reply:
looks good!

i'm not 100% sure this will fix the issue, since this hook only waits for the service container to start, not for the service to be available. if the issue is a startup timing one, the success of this strategy will depend on how quickly the services run their startup processes after the container comes up.

this is very sensible either way and will certainly reduce the footprint of any racy behavior.

no-reply (approved):

PR #4885, opened 2021-04-23
Gh 4833 depositing a work with past embargo gives no id error

### LINKED ISSUE #4833: Depositing a work with past embargo gives no id error
### Descriptive summary

Depositing a work with an embargo date in the past gives an "No ID provided for ActiveFedora::Base" error.
noID

https://nurax-dev.curationexperts.com/concern/generic_works?locale=en

### Expected behavior
Depositing a work with a past embargo should fail to be created and entered information should be preserved (per #3978 )

### Actual behavior
The user receives an error message.

1. Log in
2. Create a work
3. Set an embargo or lease date to before today
4. Press the save button

### Related work

jeremyf:
## Release Notes:

Fixes #4833: Ensures that no exception is raised when giving an embargo date set in the past;  This relates to introducing the improved permission handling from #3166 and how form data is passed in.

## Renaming method to better reflect return value

76c68fb1fdb7950c22985ff7ec02066464f1f9e0

Given that we're working with an array of IDs, I want the private method
name to better reflect the expected behavior.  The previous name ain't
wrong, this just nudges the comprehendability in a slightly better
direction.

## Compacting array building logic

c97c31fc6cbc5393565f03597ef9c848cd65b7d9

This commit serves two purposes:

1) First, and most important, it fixes #4833
2) Second, it unifies the logic for determining the ids

More on 4833 (in case you don't have access to Github)

The observed error is as follows:

> Depositing a work with past embargo gives no id error
>
> Depositing a work with an embargo date in the past gives an "No ID
> provided for ActiveFedora::Base" error.
>
> Expected Behavior: Depositing a work with a past embargo should fail
> to be created and entered information should be preserved

In debugging the application, the `@object.member_of_collection_ids` was
the following: `["", "admin_set/default"]`

That appears to come from the form submitting a select box that has no
selected value (e.g., the `""` value).  I verified this logic change by
UI interaction.

## Removing redundant `.any?` check

696da5dbb1afb128f2d97151e83b91f48e0cb935

The guard of `.any?` prevented us from iterating via the `.each`.  In
the case of an empty Array, this new implementation is logically
equivalent.  And ini the case of a non-empty Array, we're saving a
method invocation.

## Documenting spot to remove ActiveFedora::Base call

657ef7c83ba3b08cc9fc0e9446268876409ee3df

This documentation reflects some of what I worked through when testing
a bug fix for #4833.  My intention is to continue working on this code,
but only after a consultation.

## Renaming local variable for clarity

d2ddfd793cedd7cc6d1254fff802376193ff112c

The variable name is of minor importance, but highlighting that its an
array of IDs helps set expectations (which of course map to the @return
documentation).

## Factoring from ActiveFedora::Base to query_service

22c888138a783f0976ee5fe3356a38aff765e78a

This commit builds on a TODO item and implements the minimum viable
change.  It highlights that there's considerable branching logic, and
this may not be the best place to encode that branching logic.

## Refactoring to extract inner block logic

35af65db1c1370d734c053f91526a69a0fde43d6

This commit appeases two Rubocop violations.  More importantly it
isolates the logic related to a single resource and whether or not it is
something the given user can or cannot manage.

### DIFF
--- a/app/services/hyrax/edit_permissions_service.rb
+++ b/app/services/hyrax/edit_permissions_service.rb
@@ -161,16 +161,37 @@ module Hyrax
       @object_managed_collection_ids ||= object_member_of_ids & managed_collection_ids
     end
 
-    # find all of the work's collections a user cannot manage
-    # note: if the collection type doesn't include "sharing_applies_to_new_works", we don't limit access
+    # find all of the work's collections a user cannot manage note: if
+    # the collection type doesn't include
+    # "sharing_applies_to_new_works", we don't limit access
+    #
     # @return [Array] of collection ids with limited access
+    #
+    # @todo Refactor code to remove explicit ActiveFedora::Base call (see method implementation for details)
     def object_unauthorized_collection_ids
       @object_unauthorized_collection_ids ||= begin
                                                 limited_access = []
                                                 unauthorized_collection_ids = object_member_of_ids - object_managed_collection_ids
                                                 unauthorized_collection_ids.each do |id|
-                                                  # TODO: Can we instead use a SOLR query?  This seems to be somewhat expensive.  However, as this is
-                                                  # used in administration instead of user front-end displays, I'm not as concerned.
+                                                  # TODO: Refactor to remove ActiveFedora::Base
+                                                  #
+                                                  # Option 1: use `Hyrax.query_service.find_many_by_ids`.  This requires
+                                                  #           some further adjustments.
+                                                  #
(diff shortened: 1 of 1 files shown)


jeremyf:
Note, I submitted https://github.com/samvera/hyrax/pull/4886 that is a PR into this branch.  #4886 builds on the work of this branch with a refactor towards Valkyrie but not so far as to be a full Valkyrization.

dlpierce (approved):

PR #4888, opened 2021-04-26
Ensuring non-parseable date bubbles up to UI

### LINKED ISSUE #4757: Lease date field error when non-date text entered
### Descriptive summary

In Hyrax 3.x, entering non-date text in the Lease date field ("asdfas", for example) on a Work produces a stack trace error when saving the Work.

### Rationale

Stack trace errors shouldn't happen for this action.

### Expected behavior

Error message should show on Work form (like it does when doing this same action on Embargo) to indicate:

Visibility When setting visibility to "lease" you must also specify lease reduced access date.
Lease reduced access date Must be a future date.

Error checking applied to Embargo seems to not be applied to Lease.

### Actual behavior

See attached stack trace error screenshot

1. Login to nurax-dev
2. Create a new Work
3. Fill in required text fields (title, creator, rights statement)
4. Select Lease for Visibility
5. Enter "adfasdg" (or text that is not a date) in date field
6. Check box for deposit agreement
7. Click Save button

Screen Shot 2021-02-26 at 10 50 15 AM

jeremyf:
## Release Notes

Fixes #4757, when creating a work and you enter a non-date string (e.g., "This is Not a Date") into the lease expiration date field, this fix now bubbles up a meaningful error message to the UI.

## Ensuring non-parseable date bubbles up to UI

c689eda6a6804a8c5447f7613589d71aca4fd50e

Prior to this commit, we were not verifying that a user's input for a
release_date was in fact a parseable date.  Chrome and Firefox apply a
date picker "widget" that enforces entries for a `type="date"`.
However, Safari does not.

Closes #4757

Note, to test this in the UI requires a fix for #4833.  That fix, at the
time of crafting this commit message, can be seen at PR #4885.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/visibility_intention.rb
+++ b/app/services/hyrax/visibility_intention.rb
@@ -54,7 +54,7 @@ module Hyrax
     ##
     # @return [Boolean]
     def valid_embargo?
-      wants_embargo? && release_date.present?
+      wants_embargo? && release_date.present? && a_valid_date?(release_date)
     end
 
     ##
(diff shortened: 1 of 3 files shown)


app/services/hyrax/visibility_intention.rb
```diff
@@ -66,13 +66,28 @@ def wants_embargo?
...
+    ##
+    # @param date [Object]
+    # @return [Boolean]
+    # @note If we don't have a valid date, we really can't have a
+    # valid release_date
+    def a_valid_date?(date)
+      return true if date.is_a?(Date)
+      return true if date.is_a?(Time)
+      Date.parse(date)
+    rescue
```
no-reply, line 89: should this be more carefully targeted? what do we want to `rescue`?

if it should catch `StandardError`, our practice is to include that explicitly for clarity.

jeremyf: I tried catching a more specific exception, and was getting a `NameError` and parse time.  I can give it another attempt.

jeremyf:
Ah, now I'm remembering the error issue. `Date::Error` is a new constant in Ruby 2.7.2.  So I'll do some sleuthing to figure out what errors to catch

no-reply (approved):
thank you!

PR #4889, opened 2021-04-26
Gh 4845 settinig lease date in past

### LINKED ISSUE #4845: Setting lease with date in the past has error messaging problems
### Descriptive summary

Setting the Visibility of a Work with a Lease that has a date in the past does not work, which is correct. However, the error messaging that shows is not helpful and ends up leaving you somewhat stuck in the Work edit form.

### Rationale

If it does not work to set a lease date in the past (which is correct, it should not work that way), the other visibility options (Public, Embargo, Private) should still be available while the error message inside Lease shows that the date must be a future date. It should be possible to either correct the Lease date and then save to apply the lease or select another visibility option and continue editing the work.

### Expected behavior

If a Lease date is in the past, the lease is not applied and the other Visibility options are still available while the error message inside Lease shows that the date must be a future date. The flash error message at the top contains text to explain that the Lease date needs to be a date in the future (or the date value is not valid).

### Actual behavior

Empty flash error message appears at the top and the Visibility looks like a Lease has been applied (no other visibility options show) and there is an error message that the date must be a future date. User can either correct the date and continue applying the Lease or they have to exit the Work edit form (using menu options or something similar to abandon the edit form without saving) and come back into the Work edit form, where the visibility options are all available, set as they were before editing, and the invalid Lease has not been applied.

Screen Shot 2021-03-31 at 4 34 56 PM

1. Edit a Work
2. Under Visibility, select Lease
3. Change the date on the lease to a date in the past
4. Save changes
5. See empty flash error message and no visibility options except for Lease (which has to have a corrected date to be applied).

jeremyf:
## Release Notes

Fixes #4845.  This fix adds both a meaningful error message and renders the correct form controls when you enter an invalid lease date.

## Fixing rendering logic for visibility component

e1fd2244d8747deb0365a69a6f34a92fda87fe3a

Fixes #4845

Prior to this fix, if we had a new record that either set an embargo or
a lease, we would narrow our rendering of the visibility component to
only a partial related to the embargo or lease.

```gherkin
Given a form that set an embargo (or release date)
And that form had invalid data
When I submitted the form
Then the response form would inidicate there was invalid data
And would render only the portion of the visibility related to
  embargo (or release date if set)
```

With this change, the form renders the entire visibility component,
letting the user change their visibility options (e.g., go from with
Lease to Public or what have you).  However, the error messaging is not
providing insight into the problem; That is to say there's a red flash
error, but the content is empty; and there is no indicator on the lease
field that there existed an error.  This warrants further work, and will
be a commit that builds on this change.

## Adding rendering of lease related error messages

9511dce400382a9d17c92465d71e2645c40af0a5

Prior to this commit, the `form_visibility_error` partial did not
include the errors for the `:lease_expiration_date` nor the
`:visibility_after_lease`.  Per convention of Hyrax, we place the error
messages on the form and not in close DOM proximity to the HTML field
in which we encountered the error.

This relates to #4845

## Adjusting lease/embargo_enforced helper logic

adb14ecf08747d3a16d4fb9cacb7c70f0db30128

This reflects moving the change from e1fd2244d — "Fixing rendering logic
for visibility component" into the helper method.  This suggestion comes
from [no_reply's review comment][1], and makes sense.  I don't think we
want to push the logic further down the stack, as there's a case
statement that would add complexity, as well as it may have unintended
consequences.  The change at this level (e.g., a helper method) makes it
more clear that this is for rendering UI components.

Related to #4845.

[1]:https://github.com/samvera/hyrax/pull/4889/files#r625466486

@samvera/hyrax-repo-managers

### DIFF
--- a/app/views/hyrax/base/_form_visibility_component.html.erb
+++ b/app/views/hyrax/base/_form_visibility_component.html.erb
@@ -1,6 +1,6 @@
-<% if embargo_enforced?(f.object) %>
+<% if embargo_enforced?(f.object) && !f.object.new_record? %>
   <%= render 'form_permission_under_embargo', f: f %>
-<% elsif lease_enforced?(f.object) %>
+<% elsif lease_enforced?(f.object) && !f.object.new_record? %>
   <%= render 'form_permission_under_lease', f: f %>
 <% else %>
     <fieldset>
--- a/app/views/hyrax/base/_form_visibility_component.html.erb
+++ b/app/views/hyrax/base/_form_visibility_component.html.erb
@@ -1,6 +1,6 @@
-<% if embargo_enforced?(f.object) && !f.object.new_record? %>
+<% if embargo_enforced?(f.object) %>
   <%= render 'form_permission_under_embargo', f: f %>
-<% elsif lease_enforced?(f.object) && !f.object.new_record? %>
+<% elsif lease_enforced?(f.object) %>
   <%= render 'form_permission_under_lease', f: f %>
 <% else %>
     <fieldset>
--- a/app/helpers/hyrax/embargo_helper.rb
+++ b/app/helpers/hyrax/embargo_helper.rb
@@ -25,6 +25,10 @@ module Hyrax
     #   holder until we switch to Valkyrie::ChangeSet instead of Form
     #   objects
     def embargo_enforced?(resource)
+      # This is a guard; from the UI rendering perspective, there's no
+      # active embargo enforcement until the object is saved.
+      return false unless resource.persisted?
+
       case resource
       when Hydra::AccessControls::Embargoable
         !resource.embargo_release_date.nil?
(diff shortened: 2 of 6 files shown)


app/views/hyrax/base/_form_visibility_component.html.erb
```diff
@@ -1,6 +1,6 @@
-<% if embargo_enforced?(f.object) %>
+<% if embargo_enforced?(f.object) && !f.object.new_record? %>
```
no-reply, line 1: i'm wondering whether the `embargo_enforced?` helper shouldn't take the `new_record?` case into account? then this `&&` could be inlined there and kept out of the view?

the other place this is used is `_form_permission.html.erb` and, at first glance, i'm guessing that partial will have similar issues in the same cases as this one. it might be fair to say that `embargo/lease_enforced?` isn't true when the object isn't yet saved.

jeremyf: This makes sense, and a new commit is incoming!

no-reply (approved):
fantastic! thanks!

Issue #2259, opened 2017-11-22
valid_delete? for permission_template_access should allow delete for :view and :deposit access


elrayle:
Branch: master

### Descriptive summary

Hyrax::Admin::PermissionTemplateAccessesController #valid_delete? method prevents users from deleting the required admin_group managers.  It also prevents removing admin_group for viewers and depositors, but it shouldn't.

### Rationale

Admins have viewer and depositor access as managers.  A user can still add the admin_group as a viewer and/or depositor.  After adding, they should be able to remove the admin_group from viewer and depositor roles.

### Expected behavior

After adding the admin_group as a viewer/depositor, the user should be able to remove the admin_group from that role.  The Repository Administrators should be removed from the list of Viewers/Depositors.  And a flash message should display stating: 'The administrative set's participant rights have been updated'.

### Actual behavior

After adding the admin_group as a viewer/depositor, the user can NOT remove the admin_group from that role.  Clicking Remove causes flash message: 'The repository administrators group cannot be removed.' and the Repository Administrators remain listed under Viewers/Depositors.

### Steps to reproduce the behavior

1. Login as admin
1. Dashboard -> Administrative Sets -> click title of an admin set -> Edit -> Participants tab
1. group: admin  role: Depositor --> Add
1. click Remove beside Repository Administrators in Depositors list

- [ ] verify that Repository Administrators was removed form the list of Depositors
- [ ] verify a flash message is displayed: 'The administrative set's participant rights have been updated'

Repeat test for role: Viewer

### Related work

#2214 Don't allow the admin permission to be revoked
#2077 remove permission_template and permission_template_access entries when permission_template is deleted

mjgiarlo:
@jcoyne inclined to have another go at this?

rjkati:
This issue is still present in Hyrax 3.0.0 rc3

jeremyf:
I have to wonder if #4497 might fix this.  I have a branch up and running for that PR, I'll be spending some time trying to reproduce.

jeremyf:
The primary logic in play is:  https://github.com/samvera/hyrax/blob/ab6c75616e3e87c0e9dfa3bb2ac483abada31edf/app/controllers/hyrax/admin/permission_template_accesses_controller.rb#L22-L31

[...]

PR #4894, opened 2021-04-29
Removing unused method

jeremyf:
In analyzing #2259, I unearthed an unused/orphaned method.

@samvera/hyrax-code-reviewers


no-reply (approved):
looks good.

this is a private method in an internal controller that i don't think applications should be subclassing, so deprecation shouldn't be needed.

PR #4906, opened 2021-05-03
initial move to config to swap between AF and Valkyrie

### LINKED ISSUE #4895: Use config to determine if valkyrie should be used.
### Descriptive summary

Methods that have separate processing for ActiveFedora and Valkyrie commonly have a parameter named `use_valkryie` that defaults to `false`.  Instead of defaulting to false, default to the configuration ``.

### Rationale

Allow all methods capable of using valkyrie to be set to use it with one configuration change.

### Expected behavior

```
use_valkyrie: Hyrax.config.use_valkryie?
```

Locate methods to update by searching for `use_valkyrie: false`.

!!! CAUTION !!!  **DO NOT** change for any call to Hyrax.query_service!

### Actual behavior

```
use_valkyrie: false
```

straleyb:
Fixes #4895

This PR sets up the application to draw from a config variable to switch between ActiveFedora default and Valkyrie. The config variable already exists, but this just transitions from the hardcoded truthy statements to the config variable. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/services/custom_queries/find_many_by_alternate_ids.rb
+++ b/lib/wings/services/custom_queries/find_many_by_alternate_ids.rb
@@ -22,7 +22,7 @@ module Wings
       # @param alternate_ids [Enumerator<#to_s>] list of ids
       # @param use_valkyrie [boolean] defaults to true; optionally return ActiveFedora::Base objects if false
       # @return [Array<Valkyrie::Resource>, Array<ActiveFedora::Base>]
-      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: true)
+      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: Hyrax.config.use_valkryie?)
         af_objects = ActiveFedora::Base.find(alternate_ids.map(&:to_s))
         return af_objects unless use_valkyrie == true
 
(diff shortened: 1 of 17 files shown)


no-reply (dismissed):

lib/wings/services/custom_queries/find_many_by_alternate_ids.rb
```diff
@@ -22,7 +22,7 @@ def initialize(query_service:)
       # @param alternate_ids [Enumerator<#to_s>] list of ids
       # @param use_valkyrie [boolean] defaults to true; optionally return ActiveFedora::Base objects if false
       # @return [Array<Valkyrie::Resource>, Array<ActiveFedora::Base>]
-      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: true)
+      def find_many_by_alternate_ids(alternate_ids:, use_valkyrie: Hyrax.config.use_valkryie?)
```
no-reply, line 25: i think these checks in `wings` should keep their existing defaults.

in truth, these should be deprecated. AFAIK, there's not a good use case for not using valkyrie within wings.

straleyb: Yep, I talked with @elrayle about it. I reverted them back to the hardcoded defaults.

PR #4907, opened 2021-05-03
add test suite hook for using the valkyrie solr index adapter

no-reply:
allows individual tests to set the valkyrie index adapter for integration tests.

@samvera/hyrax-code-reviewers


elrayle (approved):
Looks good.  Thanks for adding this.

Issue #3820, opened 2019-06-17
Replace calls to ActiveFedora::Base.where (removes 7 refs to ActiveFedora)


elrayle:
### Prerequisites

This issue cannot be completed until the following issues are completed first...

* #3800  setup solr indexers and solr metadata adapter

----
### Descriptive summary

The `ActiveFedora::Base.where` is used to apply a solr query and return instances of Active Fedora objects.  

It is used to accomplish a few tasks in Hyrax...
* get a count of objects searching solr and counting results
* get a count of objects that the current user can see by searching solr, limiting results based on `current_ability`, and counting the filtered results
* get a set of objects

This issue involves removing references to `ActiveFedora::Base.where`.

### ActiveFedora code

Active Fedora class being replaced...

Example from app/models/hyrax/collection_type.rb
```
      ActiveFedora::Base.where(collection_type_gid_ssim: gid.to_s)
```

To locate, search code for `ActiveFedora::Base.where` in `app`.  You will also need to search for `.where`, but some of these will be for non `ActiveFedora::Base` objects (e.g. `Users`) and should not be updated as they are using `ActiveRecord #where`.  The ones for `ActiveFedora::Base` objects (e.g. `AdminSet`, `Collection`, `Work`, `FileSet`) will need to be updated.

### Replacement code

Pseudocode replacement code...
* query = build a solr query based on what was being passed to `ActiveFedora::Base.where` - Can existing [Hyrax::SolrQueryServiceBuilder](https://github.com/samvera/hyrax/blob/main/app/services/hyrax/solr_query_builder_service.rb) be used or expanded as a replacement for building the solr query?
* ids = submit query to [Hyrax::SolrService](https://github.com/samvera/hyrax/blob/main/app/services/hyrax/solr_service.rb) 
* resources = use [Wings::Valkyrie::QueryService #find_many_by_ids](https://github.com/samvera/hyrax/blob/main/lib/wings/valkyrie/query_service.rb#L72-L96) passing ids from solr search if resources are needed

Perhaps create a service that performs all these steps that can be used as a direct replacement for #where. 
 
### Considerations

* Can some uses of `where` to find related objects be replaced with calls to one of the [custom navigation queries](https://github.com/samvera/hyrax/tree/main/app/services/hyrax/custom_queries/navigators) or the standard [find_references_by](https://github.com/samvera/hyrax/blob/main/lib/wings/valkyrie/query_service.rb#L134-L150) query method?
[...]

PR #4910, opened 2021-05-04
replace ActiveFedora where with Valkyrized where

elrayle:
Partially addresses Issue #3820

Establishes the pattern for how to replace calls to ActiveFedora::Base #where method.

Establishes a new service class that searches solr to get object IDs and then uses Hyrax.query_service get the objects.  The method name that is the equivalent of the ActiveFedora where method is `.find_for_model_by_field_pairs`.

This allows for other types of queries for solr to be constructed as needed at a later time.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/find_objects_via_solr_service.rb
+++ b/app/services/hyrax/find_objects_via_solr_service.rb
@@ -9,18 +9,18 @@ module Hyrax
 
     class << self
       # Find objects matching search criteria.
-      # @param model [Class]
+      # @param model [Class] if not using Valkyrie, this is expected to be an ActiveFedora::Base object that supports #where
       # @param field_pairs [Hash] a list of pairs of property name and values
       # @param join_with [String] the value we're joining the clauses with (default: ' OR ' for backward compatibility with ActiveFedora where)
       # @param type [String] The type of query to run. Either 'raw' or 'field' (default: 'field')
       # @param use_valkyrie [Boolean] if true, return Valkyrie resource(s); otherwise, return ActiveFedora object(s)
-      # @return [ActiveFedora | Valkyrie::Resource] objects matching the query
+      # @return [Array<ActiveFedora|Valkyrie::Resource>] objects matching the query
       def find_for_model_by_field_pairs(model:, field_pairs:, join_with: ' OR ', type: 'field', use_valkyrie: Hyrax.config.use_valkyrie?)
+        return model.where(field_pairs).to_a unless use_valkyrie
         query = solr_query_builder.construct_query_for_model(model, field_pairs, join_with, type)
         results = solr_service.query(query)
         ids = results.map(&:id)
-        return query_service.find_many_by_ids(ids: ids) if use_valkyrie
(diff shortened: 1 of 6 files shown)


app/services/hyrax/multiple_membership_checker.rb
```diff
@@ -53,11 +53,15 @@ def check(collection_ids:, include_current_members: false)
...
+      field_pairs = {
+        :id => collection_ids,
+        Hyrax.config.collection_type_index_field.to_sym => collection_type_gids_that_disallow_multiple_membership
+      }
+      Hyrax::FindObjectsViaSolrService.find_for_model_by_field_pairs(model: ::Collection, field_pairs: field_pairs, use_valkyrie: true)
     end
 
     def collection_type_gids_that_disallow_multiple_membership
-      Hyrax::CollectionType.gids_that_do_not_allow_multiple_membership
+      Hyrax::CollectionType.gids_that_do_not_allow_multiple_membership.map(&:to_s)
```
no-reply, line 64: what's going on here?

app/services/hyrax/find_objects_via_solr_service.rb
```diff
@@ -0,0 +1,27 @@
...
+      # @param join_with [String] the value we're joining the clauses with (default: ' OR ' for backward compatibility with ActiveFedora where)
+      # @param type [String] The type of query to run. Either 'raw' or 'field' (default: 'field')
+      # @param use_valkyrie [Boolean] if true, return Valkyrie resource(s); otherwise, return ActiveFedora object(s)
+      # @return [ActiveFedora | Valkyrie::Resource] objects matching the query
+      def find_for_model_by_field_pairs(model:, field_pairs:, join_with: ' OR ', type: 'field', use_valkyrie: Hyrax.config.use_valkyrie?)
+        query = solr_query_builder.construct_query_for_model(model, field_pairs, join_with, type)
+        results = solr_service.query(query)
+        ids = results.map(&:id)
+        return query_service.find_many_by_ids(ids: ids) if use_valkyrie
+        ids.map { |id| query_service.find_by_alternate_identifier(alternate_identifier: id.to_str, use_valkyrie: false) }
```
no-reply, line 23: what's the advantage of using the valkyrie query service?

would there be fewer risks if we stuck with the the existing `ActiveFedora::Base#where` implementation in the case where what we want is to query AF objects?

no-reply (approved):

app/services/hyrax/find_objects_via_solr_service.rb
elrayle: I updated to use AF where unless use_valkyrie is true.

jeremyf (commented):
I think this is a great pattern and helps create the inflection point.  I "approve" this contingent on pending `.map(&:to_s)` question

app/services/hyrax/multiple_membership_checker.rb
elrayle: Tests initially failed when refactoring for the new approach which required this change.  But the final refactor really doesn't need this.  So I set it back to what it was.

jeremyf (approved):

PR #4915, opened 2021-05-05
Verifying round trip AF conversion validation

### LINKED ISSUE #3917: when valkyrie resource returns false for valid?, the converted AF FileSet returns true for valid?
### Descriptive summary

If a Valkyrie::Resource which returns false for valid? is converted to an Active Fedora FileSet (and likely any AF object), the new AF FileSet return true for valid?

### Rationale

If the Valkyrie::Resource is invalid, then any object created from it should also be invalid.

### Background

This was discovered while writing an equivalent test for https://github.com/samvera/hydra-works/blob/master/spec/hydra/works/services/add_file_to_file_set_spec.rb#L22-L30.  This test was to be applied to [Wings::Works::AddFileNodeToFileSet](https://github.com/samvera/hyrax/blob/master/lib/wings/hydra/works/services/add_file_node_to_file_set.rb).

The wings' version of the test causes the Valkyrie::Resource for the FileSet to return false for `valid?`  The first thing that happens in the service is a conversion from the Valkyrie::Resource to an Active Fedora FileSet.  When the conversion is complete, the AF FileSet does not return false for `valid?`.

NOTE: In the test, the Valkyrie::Resource is not really invalid based on the contents of the resource.  But the `valid?` method is stubbed to return false.

### Unknowns

Should the first step in the conversion process from resource to AF object be a check for whether the resource is valid?  The same question applies when converting from AF object to resource.

Should the conversion process try to do the conversion even if the resource says it is invalid and somehow set a flag in the AF object that it is also invalid?

### Expected behavior

OPTION 1:  Throw exception

The conversion process should fail with an exception stating that the resource is invalid and cannot be converted.

OPTION 2: Try to convert and have new AF object.valid? return false
 
An invalid resource that is converted to an AF object should return the same results for `valid?`

```
resource.valid?  # returns false
af_object = Wings::ActiveFedoraConverter.new(resource: resource).convert
af_object.valid?  # returns false
```

### Actual behavior

```
resource.valid?  # returns false
af_object = Wings::ActiveFedoraConverter.new(resource: resource).convert
af_object.valid?  # returns true
```

1. create a valkyrie resource and make it return false for `valid?`
2. use `Wings::ActiveFedoraConverter.new(resource: resource).convert` to try to convert it

Make sure the result is as described in the Expected behavior section.

### Related work

Issue #3813 Valkyrize FileActor including the ability to save the file using valkyrie wings
PR #3814 use active_fedora storage adapter to save file and connect it to the file_set

jeremyf:
The desired behavior is as follows:

Given `x` is an ActiveFedora object which responds to `x.valid?` and returns `false`
And `f(x)` converts `x` to a `Valkyrie::Resource` (which doesn't respond to `valid?`)
And `g(f(x))` converts `f(x)` resource back to an ActiveFedora object that responds to `valid?`
Then `g(f(x))` responds to `valid?` and returns `false`

This spec tests that behavior.  However, I'm not able to duplicate the
error.  Perhaps it's a matter of the specific parameters?  I'm adding
this spec as a placeholder for what could be the specific
attributes/values that trigger a failing scenario.

Closes #3917 as I'm unable to duplicate the error

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/wings/active_fedora_converter_spec.rb
+++ b/spec/wings/active_fedora_converter_spec.rb
@@ -24,6 +24,25 @@ RSpec.describe Wings::ActiveFedoraConverter, :clean_repo do
   end
 
   describe '#convert' do
+    context 'with an invalid GenericWork object' do
+      it 'round trip converts to an GenericWork object this is also invalid' do
+        work = GenericWork.new(title: nil)
+        expect(work).not_to be_valid
+        converted_work = described_class.new(resource: work.valkyrie_resource).convert
+        expect(converted_work).not_to be_valid
+        expect(work.errors.full_messages).to eq(converted_work.errors.full_messages)
+      end
+    end
+    context 'with an invalid FileSet object' do
+      it 'round trip converts to an FileSet object this is also invalid' do
+        file_set = ::FileSet.new(lease_expiration_date: 1.day.ago)
+        expect(file_set).not_to be_valid
+        converted_file_set = described_class.new(resource: file_set.valkyrie_resource).convert
+        expect(converted_file_set).not_to be_valid
+        expect(file_set.errors.full_messages).to eq(converted_file_set.errors.full_messages)
+      end
+    end
+
     it 'returns the ActiveFedora model' do
       expect(converter.convert).to eq work
     end


elrayle (approved):
Nice to have this confirmed that it is no longer a problem.

PR #4922, opened 2021-05-05
don't force `webdrivers` on downstream apps

no-reply:
if apps are using an external browser server for testing, don't make them
install `webdrivers`.

@samvera/hyrax-code-reviewers


dunn (approved):
:pray:

PR #4923, opened 2021-05-06
use the latest `selenium/standalone-chrome`

no-reply:
selenium has changed its terminology to `allowed-ips`. let's take advantage of
that, and upgrade besides.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/specs/capybara.rb
+++ b/lib/hyrax/specs/capybara.rb
@@ -18,7 +18,7 @@ require 'selenium-webdriver'
 require 'webdrivers' unless ENV['IN_DOCKER'].present? || ENV['HUB_URL'].present?
 
 if ENV['IN_DOCKER'].present? || ENV['HUB_URL'].present?
-  args = %w[disable-gpu no-sandbox whitelisted-ips window-size=1400,1400]
+  args = %w[disable-gpu no-sandbox allowed-ips window-size=1400,1400]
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
       - rails-tmp:/app/samvera/hyrax-webapp/tmp
 
   chrome:
-    image: selenium/standalone-chrome:3.141
```
dunn, line 38: this version scheme is still in use, and we might want to keep using it to make it easier to match the version on the ruby end: https://hub.docker.com/r/selenium/standalone-chrome/tags?page=1&ordering=last_updated

no-reply: good call. in the meanwhile, i'm able to confirm that i could run the feature specs in docker-compose with this PR.

dunn (approved):

no-reply:
well, with the rolled back version, no luck. are we sure we want `3.141`.

the ruby gem is at `3.142.7`, and chromium itself is on `90`. i think we might just want 90.

jeremyf (approved):

no-reply:
@jeremyf were you able to get this working?

my last attempt at end of day yesterday was broken on my end, per: https://github.com/samvera/hyrax/pull/4923#issuecomment-833173148

PR #4926, opened 2021-05-06
WIP - Review Requested

jeremyf:
## Refactoring to make an easier change

d5dec6eac99fe656c33fdef7d95dc650d4bd175e

This is code change is a simple restructuring to enable an easier review
for future work on #4196.  _I know I struggle reviewing a code change
that both wraps a chunk of code with a containing block AND changes the
logic of that block._

Related to #4196

## Renaming variable for better clarity

28408b1e42a06570e16e1a99f87c6c1ba0ec0204

Prior to this commit, when I was reading the code I saw "f" and
"file_set" in the same context.  That "f" is then passed to another
method which clarifies what it's type.

This commit looks to separate the "f" (which implies "file") from the
"FileSet" which is the container of a "file".

## Favor case instead of `use_valkyrie` switch

3f4618f99ac42e84f9cb809dae805a363a8fd50e

Aside from the `file_set.reload`, all other collaborators with the
ImportUrlJob can handle a Valkyrie::Resource or ActiveFedora::Base
object.

That change will be a future commit.

## Refactoring to provide switch for parent

30f41515dbfc9ff2cbff11eb2e4cd51f17da4402

A FileSet responds to "#parent", but a Hyrax::FileSet does not.  This
refactor helps with a future change.

## WIP

d44fcd552ca8da660a0c306084965a96043db666

Adding a possible approach to resolving #4196; I want to put this up for
comment before I proceed with further tests.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/file_set_actor.rb
+++ b/app/actors/hyrax/actors/file_set_actor.rb
@@ -140,11 +140,16 @@ module Hyrax
       # @param file_set [FileSet]
       # @return [ActiveFedora::Base]
       def parent_for(file_set:)
-        file_set.parent
+        case file_set
+        when ActiveFedora::Base
+          file_set.parent
+        else
+          Hyrax.query_service.find_parents(resource: file_set).first
+        end
       end
 
       def build_file_actor(relation)
-        fs = use_valkyrie ? file_set.valkyrie_resource : file_set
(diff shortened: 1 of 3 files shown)


jeremyf:
I'm going to close this as I don't think it's relevant.

PR #4928, opened 2021-05-06
Move Hydra:AccessControl conversion to a valkyrie resource into ModelTransformer

cjcolvar:
wings.rb was defining #valkyrie_resource directly on
Hydra::AccessControl but the query service is attempting to convert
using ModelTransformer which doesn't know how to handle AccessControl
objects.  Getting rid of the direct method declaration in favor of
Valkyrizable standardizes and solves this problem.

Fixes #4901

@samvera/hyrax-code-reviewers

PR #4930, opened 2021-05-06
Default to ValkyrieWorkIndexer when resource is a Hyrax::Work

cjcolvar:
When working on #4928 I noticed that `Hyrax::Test::SimpleWork` wasn't having it's permissions indexed because it doesn't have an indexer following the naming conventions.  Instead it was getting assigned `ValkyrieIndexer` which doesn't have the work specific includes.  This PR changes the default for `Hyrax::Work` resources to be `Hyrax::ValkyrieWorkIndexer` as a more helpful default.

### DIFF
--- a/app/indexers/hyrax/valkyrie_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_indexer.rb
@@ -114,7 +114,7 @@ module Hyrax
 
         return indexer_class if indexer_class.is_a?(Class) &&
                                 indexer_class.instance_methods.include?(:to_solr)
-        ValkyrieIndexer
+        resource.try(:work?) ? Hyrax::ValkyrieWorkIndexer : ValkyrieIndexer
       end
     end
   end
--- a/app/models/concerns/hyrax/solr_document_behavior.rb
+++ b/app/models/concerns/hyrax/solr_document_behavior.rb
@@ -117,6 +117,12 @@ module Hyrax
       hydra_model == ::AdminSet
     end
 
+    ##
+    # @return [Boolean]
+    def work?
+      Hyrax.config.curation_concerns.include? hydra_model
+    end
+
     # Method to return the model
     def hydra_model(classifier: ActiveFedora.model_mapper)
       "::#{first('has_model_ssim')}".safe_constantize ||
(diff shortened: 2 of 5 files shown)


app/indexers/hyrax/valkyrie_indexer.rb
```diff
@@ -114,7 +114,7 @@ def indexer_class_for(resource)
 
         return indexer_class if indexer_class.is_a?(Class) &&
                                 indexer_class.instance_methods.include?(:to_solr)
-        ValkyrieIndexer
+        resource.is_a?(Hyrax::Work) ? Hyrax::ValkyrieWorkIndexer : ValkyrieIndexer
```
no-reply, line 117: do we prefer the type check or `resource.work?`

cjcolvar: Oooh, that sounds better.

app/models/concerns/hyrax/solr_document_behavior.rb
```diff
@@ -117,6 +117,12 @@ def admin_set?
       hydra_model == ::AdminSet
     end
 
+    ##
+    # @return [Boolean]
+    def work?
+      Hyrax.config.curation_concerns.include? hydra_model
```
cjcolvar, line 123: I've taken this from FilterByType: https://github.com/samvera/hyrax/blob/main/app/search_builders/hyrax/filter_by_type.rb#L42-L52
I'm not sure if we need to worry about documents in solr with `generic_type_sim: Work` which aren't registered curation_concerns.

no-reply: what are the tradeoffs with using the `generic_type` field instead? registered concerns is a bit awkward and used in inconsistent ways in other parts of the app, so i've been avoiding using it in new code. this looks like a pretty straightforward use, so i'm 👍🏻 on it, but wondering whether `generic_type` isn't better all-around.

cjcolvar: This is the document returned from solr so it doesn't include the `generic_type_si[m]` field because it is only indexed.  Maybe it is worth switching this to `ssi`?

no-reply (approved):

Issue #4911, opened 2021-05-05
Embargoes #update method in embargoes_controller_behavior.rb needs to be able to work with Valkyrie 


straleyb:
### Descriptive summary

EmbargoesControllerBehavior needs to be transitioned to work with Valkyrie. The configuration is on line 31 and is used by `Hyrax.custom_queries.find_many_by_alternate_ids` This method isn't just returning an af_object anymore. When the configuration is swapped to true for the custom queries, it can either return an af_object or valkyrie_object. The #update method in this behavior needs to know how to handle that properly.

# Current behavior
The #update method in the EmbargoesControllerBehavior doesn't handle valkyrie objects if `Hyrax.custom_queries.find_many_by_alternate_ids` is swapped to `use_valkyrie: true`. If the tests are run for this method, they fail due to not supporting the valkyrie object returned by the query.

# Expected behaior
The #update method in the EmbargoesControllerBehavior handles both the af_object and valkyrie_object when `use_valkyrie: true` is set. Once this is done, then we can swap that to use the actual configuration `Hyrax.config.use_valkyrie`

elrayle:
Fixed by #4934.  Confirmed by @straleyb in Slack #developer-congress channel on 2021-07-13 14:48ET.

PR #4934, opened 2021-05-06
enable expiration of embargoes when valkyrie is enabled

straleyb:
This is a proposed fix for #4911

This allows for the expiration of embargoes when `use_valkyrie: true`. The main issue here is that when the persister is used, it tries to save the expired embargo. We really dont NEED to save the valkyrie object. The embargo is updated properly using the `EmbargoManager` and all we need to do is tell the AF Object that its embargo is expired too.

```
manager = EmbargoManager.new(resource: valkyrie_object)
manager.release!
af = Wings::Valkyrie::ResourceFactory.new(adapter: Wings::Valkyrie::MetadataAdapter).from_resource(resource: valkyrie_object)
Hyrax::Actors::EmbargoActor.new(af).destroy
```

This is the main bulk of the change. This transforms the valkyrie object back to an af object and then uses the embargo actor to expire it.

This is only a proposed change, and if this functionality needs to live elsewhere, then we can move this around to where it needs to go.

### DIFF
--- a/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
@@ -28,7 +28,24 @@ module Hyrax
       filter_docs_with_edit_access!
       copy_visibility = []
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] } if params[:embargoes]
-      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
+      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: Hyrax.config.use_valkyrie?)
+      if Hyrax.config.use_valkyrie? 
+        af_objects.each do |valkyrie_object|
+          manager = EmbargoManager.new(resource: valkyrie_object)
+          manager.release!
+          af = Wings::Valkyrie::ResourceFactory.new(adapter: Wings::Valkyrie::MetadataAdapter).from_resource(resource: valkyrie_object)
+          Hyrax::Actors::EmbargoActor.new(af).destroy
+        end
+      else
+        af_objects.each do |curation_concern|
+          Hyrax::Actors::EmbargoActor.new(curation_concern).destroy
+          # if the concern is a FileSet, set its visibility and visibility propagation
+          if curation_concern.file_set?
+            curation_concern.visibility = curation_concern.to_solr["visibility_after_embargo_ssim"]
+            curation_concern.save!
+          elsif copy_visibility.include?(curation_concern.id)
+            Hyrax::VisibilityPropagator.for(source: curation_concern).propagate
+          end
       af_objects.each do |curation_concern|
         Hyrax::Actors::EmbargoActor.new(curation_concern).destroy
         # if the concern is a FileSet, set its visibility and visibility propagation


app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
```diff
@@ -28,7 +28,24 @@ def update
       filter_docs_with_edit_access!
       copy_visibility = []
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] } if params[:embargoes]
-      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
+      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: Hyrax.config.use_valkyrie?)
```
straleyb, line 31: Im going to change this name to `resources`

app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
```diff
@@ -24,23 +24,29 @@ def destroy
...
       filter_docs_with_edit_access!
       copy_visibility = []
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] } if params[:embargoes]
-      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
-      af_objects.each do |curation_concern|
-        Hyrax::Actors::EmbargoActor.new(curation_concern).destroy
+      resources = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: Hyrax.config.use_valkyrie?)
+      resources.each do |resource|
+        EmbargoManager.new(resource: resource).release! if Hyrax.config.use_valkyrie?
+        resource = Wings::Valkyrie::ResourceFactory.new(adapter: Wings::Valkyrie::MetadataAdapter).from_resource(resource: resource) if Hyrax.config.use_valkyrie?
```
no-reply, line 36: i'm not sure i'm following this. what's gonig on here?

straleyb: This transforms the object back into an activefedora object. Then we can use that to expire the embargo on the af side of things. 

When an embargo is expired and you try to persist the valkyrie object, it rejects saving the object because the `Embargo date needs to be set in the future`. So since we aren't really trying to save, just expire the embargo, we can just use the old code and operate on a valkyrie object if use_valkyrie is set to true, and then expire the embargo for the af object as per usual.

no-reply: >  Then we can use that to expire the embargo on the af side of things.

ah, i think this is where i got lost. i don't think we want to do two transformations (which are time-wise costly) to do this work.

>  it rejects saving the object because the Embargo date needs to be set in the future

let's focus in on this as undesirable behavior. can you work up a failing test demonstrating this case?

app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
```diff
@@ -24,23 +24,29 @@ def destroy
...
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength
     def update
       filter_docs_with_edit_access!
       copy_visibility = []
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] } if params[:embargoes]
-      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: false)
-      af_objects.each do |curation_concern|
-        Hyrax::Actors::EmbargoActor.new(curation_concern).destroy
+      resources = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: Hyrax.config.use_valkyrie?)
```
no-reply, line 33: i think we should consider just skipping all the `use_valkyrie?` stuff here. if this controller works with valkyrie and the impact is all internal, why not just use it and not present any other option?

PR #4935, opened 2021-05-06
Restructuring spec to ease review for Valkyrie

jeremyf:
This is code change is a simple restructuring to enable an easier review
for future work on #4196.  _I know I struggle reviewing a code change
that both wraps a chunk of code with a containing block AND changes the
logic of that block._

Related to #4196

@samvera/hyrax-code-reviewers


no-reply (changes requested):
could i ask for a less generic top-line in the commit message?

no-reply (approved):

PR #4938, opened 2021-05-07
WIP: Refactor hyrax search query builder to build and execute queries

### LINKED ISSUE #4936: Refactor Hyrax::SolrQueryBuilderService to build up the query and allow execution
### Descriptive summary

The solr query builder service will be more functional if it can build up queries across several methods (e.g. include list of ids, limit to model, and attr-value pairs) instead of creating a single query complete query for each method.  This will facilitate creation of more complex queries.

### Rationale

There is a need to build up queries instead of executing a simple query and then filter the results.

#### Rename Class

To recognize that the service does more than construct queries, it is being renamed from `Hyrax::SolrQueryBuilderService` to `Hyrax::SolrQueryService`.  `Hyrax::SolrQueryBuilderService` and its methods are being deprecated.

_NOTE: SolrQueryBuilderService uses class methods.  The refactored SolrQueryService uses instance methods.  This allows the query to be built up within the instance._

| | |
| -- | -- |
| Name: | initialize |
| Parameters: | query: <Hash> _(Default: {})_, solr_service: <Class> _(Default: Hyrax::SolrService)_ | 
| Returns: | self |
| Description: | Sets @query and @solr_service |

attr_reader :query, :solr_service

#### Rename Methods that Generate Query Snippets

Methods are being renamed as follows...

| | |
| -- | -- |
| Old Name: | construct_query_for_ids |
| New Name: | with_ids |
| Parameters: | ids: <Array> |
| Returns: | self |
| Generates: | `'{!terms f=id}id1,id2'` |

| | |
| -- | -- |
| Old Name: | construct_query_for_model |
| New Name: | with_model |
| Parameters: | model: <Class> |
| Returns: | self |
| Generates: | `'_query_:"{!field f=has_model_ssim}Monograph"'` |

| | |
| -- | -- |
| Old Name: | _N/A_ |
| New Name: | with_generic_type |
| Parameters: | generic_type: <String> _(Default: 'Work')_ |
| Returns: | self |
| Generates: | `'(_query_:"{!field f=generic_type_sim}Work" OR _query_:"{!field f=generic_type_si}Work")'` |

| | |
| -- | -- |
| Old Name: | construct_query_for_pairs |
| New Name: | with_field_pairs |
| Parameters: |  field_pairs: <Hash>, join_with: <String> _(Default: ' AND ')_, type: <String> _(Default: 'field')_ |
| Returns: | self |
| Generates: | `'(_query_:"{!field f=library_id_ssim}123" AND _query_:"{!field f=owner_ssim}Fred")'` |

#### Methods that Submit Queries

| | |
| -- | -- |
| Name: | build |
| Returns: | a solr query string |
| Description: |  builds a query string from @query.  See description of Building the Query below. |

| | |
| -- | -- |
| Name: | count |
| Returns: | count of results |
| Description: | calls SolrService.count with the built query |

| | |
| -- | -- |
| Name: | get |
| Returns: | solr documents matching the query |
| Description: | calls SolrService.get with the built query |

| | |
| -- | -- |
| Name: | reset |
| Returns: | self |
| Description: | resets @query to {} |

#### Building the Query

Example Query hash:

```
{
  '{!terms f=id}id1,id2',
  '_query_:"{!field f=has_model_ssim}Monograph"',
  '(_query_:"{!field f=generic_type_sim}Work" OR _query_:"{!field f=generic_type_si}Work")',
  '(_query_:"{!field f=library_id_ssim}123" AND _query_:"{!field f=owner_ssim}Fred")'
  '(_query_:"{!field f=subject_id_ssim}Science" OR _query_:"{!field f=subject_ssim}Philosophy")'
}
```

Building process:  each component is joined using `' AND '`

Results:

```
  "{!terms f=id}id1,id2 AND " \
  "_query_:\"{!field f=has_model_ssim}Monograph\" AND " \
  "(_query_:\"{!field f=generic_type_sim}Work\" OR _query_:\"{!field f=generic_type_si}Work\") AND " \
  "(_query_:\"{!field f=library_id_ssim}123\" AND _query_:\"{!field f=owner_ssim}Fred\") AND " \
  "(_query_:\"{!field f=subject_id_ssim}Science\" OR _query_:\"{!field f=subject_ssim}Philosophy\")"
```

### Related work

PR #4929 replace AF.where in dashboard_helper_behavior
PR #4910 replace ActiveFedora where with Valkyrized where

elrayle:
----
**Work in Progress -- DO NOT MERGE*

Remaining work
* write tests for the new service
* replace uses of the deprecated service methods with the new service

NOTE: Tests for the deprecated class being replacing continues to pass.
----
Fixes #4936

This implements the refactor design described in the Issue #4936.

@samvera/hyrax-code-reviewers


app/services/hyrax/solr_query_builder_service.rb
```diff
@@ -1,16 +1,24 @@
 # frozen_string_literal: true
 module Hyrax
+  ##
   # Methods in this class are from/based on ActiveFedora::SolrQueryBuilder
   class SolrQueryBuilderService
```
no-reply, line 5: is the whole class deprecated?

app/services/hyrax/solr_query_service.rb
```diff
@@ -0,0 +1,162 @@
...
+    def reset
+      @query = []
+      self
+    end
+
+    ##
+    # @param ids [Array] id_array the ids that you want included in the query
+    # @return [Hyrax::SolrQueryService] the existing service with id query appended
+    def with_ids(ids: [])
+      raise ArgumentError, "Expected [Array] for ids:, but received <#{ids.class}>" unless ids.is_a? Array
```
no-reply, line 44: it looks like this will reject a lot of arguments that would otherwise work fine? is there any harm in dropping the type check?

app/services/hyrax/solr_query_service.rb
```diff
@@ -0,0 +1,162 @@
...
+      id_query = construct_query_for_ids(ids)
+      @query += [id_query]
+      self
+    end
+
+    ##
+    # @param model [Class|String] a class from the model (e.g. Hyrax::Work, Hyrax::FileSet, etc.)
+    # @return [SolrQueryService] the existing service with model query appended
+    def with_model(model:)
+      raise ArgumentError, "Expected <Class> or <String> for model:, but received <#{model.class}>" unless model.is_a?(Class) || model.respond_to?(String)
```
no-reply, line 56: i think `model.respond_to?(String)` might be an error?

it seems like the argument type for `model` is `#to_s`? could the whole type check be dropped?

app/services/hyrax/solr_query_service.rb
```diff
@@ -0,0 +1,162 @@
...
+    # @param [String] join_with the value we're joining the clauses with (default: ' AND ')
+    # @param [String] type of query to run. Either 'raw' or 'field' (default: 'field')
+    # @return [String] a solr query
+    # @example
+    #   construct_query([['library_id_ssim', '123'], ['owner_ssim', 'Fred']])
+    #   # => "_query_:\"{!field f=library_id_ssim}123\" AND _query_:\"{!field f=owner_ssim}Fred\""
+    def construct_query_for_pairs(field_pairs, join_with = default_join_with, type = 'field')
+      clauses = pairs_to_clauses(field_pairs, type)
+      return "" if clauses.count.zero?
+      return clauses.first if clauses.count == 1
```
no-reply, line 109: is this line to keep the parentheses wrap out? i think the internal string is the same, either way, so i was wondering if this conditional is necessary.

app/services/hyrax/solr_query_builder_service.rb
```diff
@@ -1,16 +1,30 @@
 # frozen_string_literal: true
 module Hyrax
+  ##
+  # @deprecated
+  # This class is being replaced by Hyrax::SolrQueryService.
+  #
   # Methods in this class are from/based on ActiveFedora::SolrQueryBuilder
+  #
+  # @see Hyrax::SolrQueryService
+  ##
```
no-reply, line 10:
```suggestion
```

app/services/hyrax/solr_query_service.rb
elrayle: I copied this over from Hyrax::SolrQueryBuilderService without examining it to determine if it should be changed.  I'm not sure of the thoughts of the original author.

elrayle: I looked at the original ActiveFedora version of `construct_query` and it is quite different.  So that isn't a help.

But looking at the code, I agree with you assessment that the purpose of line 109 is to return the clause without parentheses if there is only one field pair.

PR #4939, opened 2021-05-07
limit embargo/lease date validation to create

no-reply:
introduce a wrapper for `Hydra::AccessControls::Embargoable` with less agressive
validation behavior.

@samvera/hyrax-code-reviewers


lib/wings/active_fedora_converter/default_work.rb
```diff
@@ -100,6 +100,9 @@ class DefaultWork < ActiveFedora::Base
       include Hydra::AccessControls::Embargoable
```
cjcolvar, line 100: Why not change this to `Hyrax::Embargoable` here as well instead of copying in the methods?

cjcolvar (dismissed):
Looks good!

no-reply:
hmmmm... i'm confused about how this got merged. i just finished writing the tests.

PR #4940, opened 2021-05-07
docker: add network

dunn:
without this I'm seeing 
```
db_migrate_1  | waiting for solr:8983
db_migrate_1  | nc: bad address 'solr'
```
during startup

https://docs.docker.com/network/bridge/

@samvera/hyrax-code-reviewers


dunn:
@no-reply could be, but part of the benefit of user networks is better isolation from the environment

no-reply (approved):

no-reply:
i'm pulling and testing locally. i'll merge after that.

PR #4941, opened 2021-05-07
Feature/expire leases for valkyrie

### LINKED ISSUE #4912: Leases #update method in leases_controller_behavior.rb needs to be able to work with Valkyrie
### Descriptive summary

LeasesControllerBehavior needs to be transitioned to work with Valkyrie. The configuration is on line 30 and is used by `Hyrax.custom_queries.find_many_by_alternate_ids` This method isn't just returning an af_object anymore. When the configuration is swapped to true for the custom queries, it can either return an af_object or valkyrie_object. The #update method in this behavior needs to know how to handle that properly.

# Current behavior
The #update method in the LeasesControllerBehavior doesn't handle valkyrie objects if `Hyrax.custom_queries.find_many_by_alternate_ids` is swapped to `use_valkyrie: true`. If the tests are run for this method, they fail due to not supporting the valkyrie object returned by the query.

# Expected behaior
The #update method in the LeasesControllerBehavior handles both the af_object and valkyrie_object when `use_valkyrie: true` is set. Once this is done, then we can swap that to use the actual configuration `Hyrax.config.use_valkyrie`

straleyb:
Fixes #4912 

This code is currently based off the fixes for the embargoes. This is subject to change depending on how we want the embargo code to look.

This code is also dependent on the changes to hydra-head found here: https://github.com/samvera/hydra-head/pull/533 and an update found here: https://github.com/samvera/hyrax/pull/4939

What this does is allows for the setting of the hyrax config to enable valkyrie. Expiring leases looks different for valkyrie objects as they can use the LeaseManager to do so. This enables the code to make the choice between Valkyrie and AF.

This code is reliant on the changes to the Persister getting pushed through.

### DIFF
--- a/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/embargoes_controller_behavior.rb
@@ -28,24 +28,17 @@ module Hyrax
       filter_docs_with_edit_access!
       copy_visibility = []
       copy_visibility = params[:embargoes].values.map { |h| h[:copy_visibility] } if params[:embargoes]
-      af_objects = Hyrax.custom_queries.find_many_by_alternate_ids(alternate_ids: batch, use_valkyrie: Hyrax.config.use_valkyrie?)
-      if Hyrax.config.use_valkyrie? 
-        af_objects.each do |valkyrie_object|
-          manager = EmbargoManager.new(resource: valkyrie_object)
-          manager.release!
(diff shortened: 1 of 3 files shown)


no-reply:
does this replace #4934?

straleyb:
> does this replace #4934?

No, I just have this up here for when the other one gets merged in. Its branched off #4934 . Like I said, this is here just as a place holder until that one moves forward. I wanted to get my work up so that it could be looked at in the mean time.

PR #4942, opened 2021-05-07
limit embargo/lease date validation to create

no-reply:
introduce a wrapper for `Hydra::AccessControls::Embargoable` with less agressive
validation behavior.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/models/generic_work_spec.rb
+++ b/spec/models/generic_work_spec.rb
@@ -59,6 +59,39 @@ RSpec.describe GenericWork do
     end
   end
 
+  describe "embargo" do
+    subject(:work) { described_class.new(title: ['a title'], embargo_release_date: embargo_release_date) }
+    let(:embargo_release_date) { Time.zone.today + 10 }
+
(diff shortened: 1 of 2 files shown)


spec/models/generic_work_spec.rb
```diff
@@ -59,6 +59,39 @@
...
+    context 'with a saved embargo' do
+      let(:past) { Time.zone.today - 10 }
+
+      before { work.save! }
+
+      it 'can update the embargo with any date' do
+        work.embargo_release_date = past
+
+        expect(work).to be_valid
+        expect{ work.save! }.not_to raise_error
```
no-reply, line 90:
```suggestion
        expect { work.save! }.not_to raise_error
```

straleyb: Is this requiring a space? Syntax wise, the space makes sense because { = do and if you write it out, it would need spaces. Im on board with this suggestion.

PR #4958, opened 2021-05-11
document and test Hyrax::ContextualPath

no-reply:
add tests and documentation for Hyrax::ContextualPath.

add a `Hyrax::FileSet.model_name` implementation for compatibility with
`FileSet`. the change in model name/namespaceing shouldn't change our paths.

@samvera/hyrax-code-reviewers


jeremyf (approved):

no-reply:
bump with a change to the i18n keys!

PR #4959, opened 2021-05-17
Allow all ActiveFedora 12.x releases

cjcolvar:
There are a number of improvements in the 12.2.x line of ActiveFedora that it would be great to be able to use in Hyrax 2.x.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,7 @@ jobs:
       - attach_workspace:
           at: ~/
       - samvera/install_solr_core:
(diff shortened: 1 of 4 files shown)


cjcolvar:
These 5 tests pass for me locally and it doesn't seem like there has been any substantial change to the UserStat code so I'm not really sure where the failure is coming from.  I'm out of time at the moment so ll have to come back to debug more later.

no-reply (approved):

PR #4960, opened 2021-05-17
[WIP] Transition to true for test failures

straleyb:


app/models/sipity.rb
```diff
@@ -105,4 +105,9 @@ def handle_conversion(input, result, method_name)
     raise ConversionError.new(input) # rubocop:disable Style/RaiseArgs
   end
   module_function :handle_conversion
+
+  def hyrax_or_valkyrie_global_id(input)
+    Entity.find_by(proxy_for_global_id: Hyrax::GlobalID(input).to_s).nil? ? Hyrax::GlobalID(ActiveFedora::Base.find(input.id.id)) : Hyrax::GlobalID(input)
```
no-reply, line 110: what's going on here?

i think the `ActiveFedora::Base.find(input.id.id)` won't be okay. this is a whole database round trip to get (what looks like the wrong) `GlobalID`

straleyb: Yeah, this is just the remedy for the problem. Im sure you all would like to really discuss how you would like to truly fix this problem. We can chat at standup if needed. This is a systemic problem with how the AF::Base and Valkyrie mechanism works when transforming a AF::Base object to a Valkyrie object. 

Basically, if an AF::Base object exists in the DB and gets transformed to a Valkyrie object by way of `Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id], use_valkyrie: true)`, the Valkyrie object doesn't have a Sipity::Entity that exists in the DB for the Valkyrie object. So when trying to process the workflow, it fails because the GlobalID for the Valkyrie object isn't the same as the AF::Base Global::ID

app/controllers/hyrax/workflow_actions_controller.rb
```diff
@@ -36,8 +36,8 @@ def workflow_action_params
 
     def after_update_response
       respond_to do |wants|
-        wants.html { redirect_to [main_app, curation_concern], notice: "The #{curation_concern.human_readable_type} has been updated." }
-        wants.json { render 'hyrax/base/show', status: :ok, location: polymorphic_path([main_app, curation_concern]) }
+        wants.html { redirect_to main_app.hyrax_generic_work_url(curation_concern, locale: 'en'), notice: "The #{curation_concern.human_readable_type} has been updated." }
```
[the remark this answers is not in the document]
straleyb: Yeah, this is due to the path being generated for a Hyrax::Work doesn't work. It pops out as `hyrax_hyrax_work_path` which doesn't exist. So there is another mechanism here that is busted at its root.

straleyb:
Fixed by way of #4968 and #4967

PR #4961, opened 2021-05-18
Allow for overriding the way form tab labels are created

cjcolvar:
This allows for downstream apps and engines to pass variables into translations based upon the form and tab.

The use case we have is to have a tab for collecting notes and to have the tab label include the number of notes present.

image

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/helpers/hyrax/work_form_helper_spec.rb
+++ b/spec/helpers/hyrax/work_form_helper_spec.rb
@@ -32,6 +32,15 @@ RSpec.describe Hyrax::WorkFormHelper do
     end
   end
 
+  describe '.form_tab_label_for' do
+    let(:form) { double('form') }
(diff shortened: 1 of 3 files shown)


spec/helpers/hyrax/work_form_helper_spec.rb
```diff
@@ -32,6 +32,40 @@
     end
   end
 
+  describe '.form_tab_label_for' do
+    context 'with a change set style form' do
+      let(:work) { build(:hyrax_work) }
```
no-reply, line 37: is all this stubbing necessary?  since the form isn't used in the default implementation, could we  do `let(:form) { double('form') }` and drop the rest?

no-reply (approved):

PR #4964, opened 2021-05-18
support ::Ability (can? style) checks for WorkShowPresenter and FileSetPresenter

no-reply:
this is to deprecate presenter `#editor?` methods.

we're in a bit of a bind, re: deprecation. our views use these methods, which
means application-side presenters will have been reasonable to use this as a
hook to drive view behavior.

the proposed deprecation path is:

now: mark these methods deprecated now, encourage calling `can?(:edit)`
directly.

4.0.0: remove existing calls to #editor? from the Hyrax codebase. add
deprecation warnings to the methods.

5.0.0: remove #editor? methods.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -417,8 +417,10 @@ module Hyrax
 
     def extract_subjects(subject)
       case subject
+      when Hyrax::WorkShowPresenter, FileSetPresenter
+        extract_subjects(subject.solr_document)
       when Draper::Decorator
-        super(subject.model)
+        extract_subjects(subject.model)
       else
         super
       end
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -132,8 +132,13 @@ module Hyrax
       graph.dump(:ttl)
     end
 
+    ##
+    # @deprecated use `::Ability.can?(:edit, presenter)`. Hyrax views calling
+    #   presenter {#editor} methods will continue to call them until Hyrax
+    #   4.0.0. The deprecation time horizon for the presenter methods themselves
+    #   is 5.0.0.
     def editor?
-      current_ability.can?(:edit, solr_document)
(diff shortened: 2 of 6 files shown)


jeremyf (approved):

app/models/concerns/hyrax/ability.rb
```diff
@@ -417,8 +417,10 @@ def can_review_submissions?
 
     def extract_subjects(subject)
       case subject
+      when Hyrax::WorkShowPresenter, FileSetPresenter
+        extract_subjects(subject.solr_document)
       when Draper::Decorator
-        super(subject.model)
+        extract_subjects(subject.model)
```
no-reply, line 423: skipping straight to `super` fails for nested decorators.

jeremyf: I suspected that to be the case.

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -132,8 +132,13 @@ def export_as_ttl
...
     end
 
+    ##
+    # @deprecated use `::Ability.can?(:edit, presenter)`. Hyrax views calling
+    #   presenter {#editor} methods will continue to call them until Hyrax
+    #   4.0.0. The deprecation time horizon for the presenter methods themselves
+    #   is 5.0.0.
     def editor?
-      current_ability.can?(:edit, solr_document)
+      current_ability.can?(:edit, self)
```
jeremyf, line 141: Do you want to add a deprecation warning?  It might make it easier to refactor the existing application.

jeremyf (approved):
I'm approving, and it's good to go.  But if you want to add the warning, I'll let you.

app/presenters/hyrax/work_show_presenter.rb
no-reply: i think the deprecation warning would just be annoying. Hyrax will need to continue to call these methods until 4.0.0, so we'd get a lot of chatter in the logs from local calls.

PR #4965, opened 2021-05-19
helperize download link display flag

no-reply:
move some repeated (and incorrect in some places!) view logic into a helper.

this reduces duplication, aligning the behavior and hopefully preventing future drift.

## Guidance for Testing
  - it would be good to confirm that the download links are displaying on the file set show pages

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/file_set_helper.rb
+++ b/app/helpers/hyrax/file_set_helper.rb
@@ -1,5 +1,18 @@
 # frozen_string_literal: true
 module Hyrax::FileSetHelper
+  ##
+  # @todo inline the "workflow restriction" into the `can?(:download)` check.
+  #
+  # @param file_set [#id]
+  #
+  # @return [Boolean] whether to display the download link for the given file
+  #   set
+  def display_media_download_link?(file_set:)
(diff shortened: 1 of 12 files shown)


jeremyf (approved):

jlhardes:
Not sure if I am testing this as it needs to be tested but I think everything is working as expected on nurax-dev. Let me know if some other conditions should be tested.

Note: none of the users below are repo-admin users

user1 deposited the following work as private https://nurax-dev.curationexperts.com/concern/generic_works/bk128b11v?locale=en
which has fileset
https://nurax-dev.curationexperts.com/concern/parent/bk128b11v/file_sets/kh04dp97g
that can be downloaded via
https://nurax-dev.curationexperts.com/downloads/kh04dp97g?locale=en

non-logged in users cannot get to any of these links

user2 is not on this work at all and also cannot get to any of these links

user2 is added to work with access to view/download as user and can get to all of the links

user3 is not on this work at all and also cannot get to any of these links

user3 is added with access to view/download as user on fileset only
user3 can not access the work link
user3 can access the fileset link and can download the associated file

PR #4967, opened 2021-05-19
update wings to support generating proper GID for non-valkyrie native objects

straleyb:


lib/wings/orm_converter.rb
```diff
@@ -45,6 +45,11 @@ def name
           def _canonical_valkyrie_model
             ancestors[1..-1].find { |parent| parent < ::Valkyrie::Resource }
           end
+
```
straleyb, line 48:
```suggestion
```

no-reply (approved):

PR #4968, opened 2021-05-19
update to use to_model on converted object

straleyb:


lib/wings/orm_converter.rb
```diff
@@ -55,6 +55,14 @@ def self.to_s
           internal_resource
         end
 
+        def to_model
+          Wings::ActiveFedoraConverter.convert(resource: self)
+        end
+
+        def errors
```
straleyb, line 62: This is for when #errors is called on the JSON response. Allows the use of the #add method for the errors to be appended.

straleyb:
@no-reply This is ready. Changing the #name method was too much integration. A lot of stuff was relying on it to return that information. Also, the ModelWrapper was still returning the model name and not the internal_resource. So as we discussed, this is the easiest way for now.

lib/wings/orm_converter.rb
no-reply: i'm wondering how this works? does it return the correct error object?

PR #4972, opened 2021-05-21
use valkyrie to power collection member removal from Dashboard

### LINKED ISSUE #4914: CollectionsController needs to support valkyrie_objects
### Descriptive summary

CollectionsController needs to be transitioned to work with Valkyrie. The configuration is used by `Hyrax.custom_queries.find_many_by_alternate_ids` This method isn't just returning an af_object anymore. When the configuration is swapped to true for the custom queries, it can either return an af_object or valkyrie_object. This controller needs to know how to handle that properly.

# Current behavior
The CollectionsController doesn't handle valkyrie objects if `Hyrax.custom_queries.find_many_by_alternate_ids` is swapped to `use_valkyrie: true`. If the tests are run for this method, they fail due to not supporting the valkyrie object returned by the query.

# Expected behaior
The CollectionsController handles both the af_object and valkyrie_object when `use_valkyrie: true` is set. Once this is done, then we can swap that to use the actual configuration `Hyrax.config.use_valkyrie`

# Code Snippet
```
batch.each do |pid|
         -work = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: pid, use_valkyrie: false)
         +work = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: pid, use_valkyrie: Hyrax.config.use_valkryie?)
          work.member_of_collections.delete @collection
          work.save!
end
```

straleyb:
Fixes #4914 

This object has been updated to use Valkyrie when deleting collections. Instead of responding to #member_of_collections, it responds to #member_of_collection_ids. Then you can call delete using the id of the collection and then persist the valkyrie object.

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -384,7 +384,7 @@ module Hyrax
 
       def remove_members_from_collection
         batch.each do |pid|
-          work = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: pid, use_valkyrie: false)
+          work = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: pid, use_valkyrie: Hyrax.config.use_valkyrie?)
           work.member_of_collections.delete @collection
           work.save!
         end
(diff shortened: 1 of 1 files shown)


jeremyf (commented):
I don't know if we need a test case for use valkyrie?

https://github.com/samvera/hyrax/blob/5faf4034537a3d56fd59019df848d6b82d9458ef/spec/controllers/hyrax/dashboard/collections_controller_spec.rb#L179-L187

no-reply:
@jeremyf i think the `use_valkyrie:` case is tested in the CI matrix.

no-reply (approved):
@straleyb since this seems to work with valkyrie, i'd like to just squash it down to the one branch.

we don't return these objects, so doing this behavior with valkyrie behind the scenes is a no-impact refactor. i'm going to approve and merge this, then submit a follow-up to remove the argument and case block.

PR #4975, opened 2021-05-24
make VersionPresenterList an Enumerable; extract creation to factory

no-reply:
`VersionPresenterList` is effectively an `Enumerable` (responds only to
`#each`), extending it with `include Enumerable` makes it more
versitile (perhaps we should consider a rename?). the current implementation
relies on loading each member up-front and then sorting, but this might not be
necessary for all backends. `Enumerable` also provides some guidance about how
to do this kind of loading lazily.

move the generation to a factory so the details of how to resolve an object to a
list of versions are encapsulated.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/version_list_presenter.rb
+++ b/app/presenters/hyrax/version_list_presenter.rb
@@ -1,8 +1,27 @@
 # frozen_string_literal: true
 module Hyrax
+  ##
+  # @api public
   class VersionListPresenter
+    include Enumerable
+
+    ##
+    # @param version_list [Array<#created>]
     def initialize(version_list)
       @raw_list = version_list
+    end
+
+    ##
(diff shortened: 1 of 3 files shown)


app/presenters/hyrax/version_list_presenter.rb
```diff
@@ -1,10 +1,29 @@
...
       @raw_list = version_list
     end
 
+    ##
+    # @param [Object] an object representing the File Set
+    #
+    # @return [Enumerable<Hyrax::VersionPresenter>] an enumerable of presenters
+    #   for the relevant file versions.
+    #
+    # @raise [ArugumentError] if we can't build an enu
```
no-reply, line 20:
```suggestion
    # @raise [ArgumentError] if we can't build an enu
```

PR #4977, opened 2021-05-25
Allow active-fedora 12.2 

cjcolvar:
This unblocks pulling in newer hydra-derivatives.

@samvera/hyrax-code-reviewers


no-reply:
is this the same as #4959?

cjcolvar:
:joy: Yes.  I obviously don't remember making that just a week ago.

no-reply:
i was thinking maybe you had a weekly reminder in your calendar! 😂

PR #4978, opened 2021-05-25
wings: generated Valkyrie::Resource classes route back to AF

no-reply:
generated Valkyrie::Resource models should support ActiveModel-like behavior for
routing back to the ActiveFedora::Base objects they are based on.

we had an existing implementation of exactly this requirement for
`SolrDocument`, so that is extracted and reused in `Wings`.

@samvera/hyrax-code-reviewers


no-reply:
supersedes a large part of #4968

dlpierce (approved):

PR #4981, opened 2021-05-27
add basic view test for the file_sets/edit template

no-reply:
this key view was completely untested, and requires quite a bit of careful setup
to work correctly!

this is a very basic test that stubs most of the later partials. getting this in
should help us scale down the setup required for the view and later partials.

@samvera/hyrax-code-reviewers


no-reply:
oops, thought this didn't depend on other work, but it does. closing.

PR #4986, opened 2021-05-28
don't reload the collection to populate subcollection modal

no-reply:
this view used to call `#find` to retirieve a collection from an id. we already
have a presenter and a solr document, so there's no need for this.

instead, allow `Hyrax::Ability` to resolve `CollectionPresenter` to its
underlying `SolrDocument`. Add `CollectionPresenter#nestable?` as an alias to
the existing `#collection_type_is_nestable?` to match the `Collection`
interface, and pass the presenter through to the modal. we continue to pass the
id into the modal for backward compatibility.

Guidance for testing:
  * add a subcollection to an existing collection from the dashboard

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/ability.rb
+++ b/app/models/concerns/hyrax/ability.rb
@@ -417,7 +417,7 @@ module Hyrax
 
     def extract_subjects(subject)
       case subject
-      when Hyrax::WorkShowPresenter, FileSetPresenter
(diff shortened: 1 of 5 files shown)


dlpierce (approved):

jlhardes:
I can add a subcollection to an existing collection from the dashboard on both nurax-dev and my local build so this is done.

PR #4987, opened 2021-05-28
don't show the "Add to collection" action when type isn't nestable

no-reply:
the previous guard skipped showing this button only if no collection types were
nestable (checking this involved a solr query). it would then base the actual
behavior on the current collection type's nestability. in the case of admin
sets, it was even hard coded to false!

instead, check whether the current type is nestable. this avoids the most common
case where the user might push this button only to be given a deny popup. since
none of the data is loaded in dynamically in response to the button push,
there's no reason we can't remove that modal implementation altogether.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/my/_admin_set_action_menu.html.erb
+++ b/app/views/hyrax/my/_admin_set_action_menu.html.erb
@@ -41,16 +41,5 @@
         <%= t("hyrax.dashboard.my.action.delete_admin_set") %>
       <% end %>
     </li>
-    <% if Hyrax::CollectionType.any_nestable? %>
-        <li role="menuitem" tabindex="-1">
-          <%= link_to "#",
-                      class: 'itemicon add-to-collection',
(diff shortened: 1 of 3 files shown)


app/views/hyrax/my/_admin_set_action_menu.html.erb
```diff
@@ -41,16 +41,5 @@
         <%= t("hyrax.dashboard.my.action.delete_admin_set") %>
       <% end %>
     </li>
-    <% if Hyrax::CollectionType.any_nestable? %>
-        <li role="menuitem" tabindex="-1">
-          <%= link_to "#",
-                      class: 'itemicon add-to-collection',
-                      title: t("hyrax.dashboard.my.action.add_to_collection"),
-                      data: { nestable: false,
```
no-reply, line 49: this line guaranteed that the `deny` modal will be used here.

dlpierce (approved):

PR #4989, opened 2021-05-28
move parent collection JSON generator into helper 

no-reply:
the implementation of query for eligible parent collections had a handful of
problems.

first, it was awkwardly located in the presentation layer. since this code is
really about a business rule ("which parent collections are eligible for this
object?"), hiding it in presenters makes it hard to understand how to approach
changes. it was also duplicated across `CollectionForm` and
`CollectionPresenter`, compounding this problem. moving it to a helper doesn't
completely solve this problem, but it does give us a single point of focus for
continued refactors. as it stands, the business rule is tightly coupled to both
the JSON data generation and the underlying query. we'll probably want to loosen
that up before changing any of these three separate concerns.

second, the implementations retrieved a full object from Fedora. this is
particularly worrying because of how many collections are likely to be included
in a page. in general, we should avoid database access initiated by a
view (controllers should do this!). in the particular case, the called code
doesn't actually need the full object so we can just skip the query. there's
still a view-initiated solr query here, but baby steps.

lastly, there were some bugs in the specific existing implementations: they
cached and returned the wrong data type for repeated calls (Array<Hash>, vs JSON
String). i think this caching is completely unused, so i dropped it from the
helper's implementation.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/collection_form.rb
+++ b/app/forms/hyrax/forms/collection_form.rb
@@ -123,6 +123,11 @@ module Hyrax
         collection_member_service.available_member_subcollections.documents
       end
 
+      ##
+      # @deprecated this implementation requires an extra db round trip, had a
+      #   buggy cacheing mechanism, and was largely duplicative of other code.
+      #   all versions of this code are replaced by
+      #   {CollectionsHelper#available_parent_collections_data}.
       def available_parent_collections(scope:)
         return @available_parents if @available_parents.present?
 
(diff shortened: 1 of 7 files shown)


no-reply:
not sure why github thinks this has failed, when circle says it succeeded.

in the meanwhile, i'm trying to figure out how to setup and test the actual requirements of these methods. what are the correct criteria for eligibility to be a parent/child collection?

it looks something like

parent:
  - provided the collection type is nestable, all collections of the same type excluding:
    - those that are already a parent or child of the current collection, recursively (?);
    - those the user cannot deposit to.

child:
  - provided the collection type is nestable, all collections of the same type excluding:
    - those that are already a parent or child of the current collection, recursively (?);
    - those the user cannot read.

no-reply:
still no test here, but i'm increasingly thinking that's fine. at least pending some further discussion about the questions here.

elrayle:
@no-reply Your interpretation of what makes a collection able to be a parent or child of a given collection matches my expectations.

jeremyf (approved):

PR #4992, opened 2021-06-04
Allow for setting of file visibility independent of work visibility

cjcolvar:
We have many cases where a work is open but a file within it is private or embargoed.  It is tedious and potentially problematic to have to attach a file then edit that file to set the correct visibility.  In the case of bulkrax it would have to be done post import.   This PR is a first step to allowing editing of a file's permissions as part of adding files in the deposit form and correctly setting file permissions in a bulkrax import.

The create/update request to the works controller needs an array of file set param hashes for this to work.  Each file set param hash is identified by the uploaded file id.  For example:
```
{ generic_work: { file_set: [{ uploaded_file_id: 1, visibility: 'restricted' },
                             { uploaded_file_id: 2, visibility: 'institution' }],
                  visibility: 'open' } }
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -203,7 +203,7 @@ module Hyrax
         @curation_concern =
           form.validate(params[hash_key_for_curation_concern]) &&
           transactions['change_set.create_work']
-          .with_step_args('work_resource.add_file_sets' => { uploaded_files: uploaded_files },
+          .with_step_args('work_resource.add_file_sets' => { uploaded_files: uploaded_files, file_set_params: params[hash_key_for_curation_concern][:file_set] },
                           'change_set.set_user_as_depositor' => { user: current_user })
           .call(form).value!
       end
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -203,7 +203,7 @@ def create_work
         @curation_concern =
           form.validate(params[hash_key_for_curation_concern]) &&
           transactions['change_set.create_work']
-          .with_step_args('work_resource.add_file_sets' => { uploaded_files: uploaded_files },
+          .with_step_args('work_resource.add_file_sets' => { uploaded_files: uploaded_files, file_set_params: params[hash_key_for_curation_concern][:file_set] },
```
no-reply, line 206: i'm slightly put off by the way `file_set_params` has to be passed through multiple layers, and then reintegrated with `uploaded_files`. do you have any thoughts about this? it feels very britle, off-hand.

i'm struggling a bit to think about better patterns. the best i've got at the moment is that maybe `desired_visibility` (or `input_parameters`/`user_parameters`?) should be added as a field on the uploaded file objects, so this data can be handled down the stack without the need to pass it around as an argument?

cjcolvar: That's an interesting idea.  I think that it would make sense to keep those parameters in the DB so it could be passed around outside of just request/job parameters which are more ephemeral.

no-reply: i'm going to approve and merge this as is, since all the added arguments are optional.

i still think a big refactor here is desirable, but i won't block the feature on it.

no-reply (approved):

PR #4996, opened 2021-06-15
remove duplicate `okcomputer` gem requires

no-reply:
`okcomputer` is is the hyrax gemspec as a development dependency. it shouldn't
be necessary here.

@samvera/hyrax-code-reviewers


jeremyf (approved):

jeremyf (approved):

no-reply:
something something enginecart...

PR #4997, opened 2021-06-17
Split off hyrax-work-base image and bundle install from dassie

dlpierce:
Fixes issues encountered when running `docker-compose build` without a Gemfile.lock in .dassie.

The bundle install command for the hyrax-engine-dev and hyrax-engine-dev-worker images was trying to resolve gem dependencies from the hyrax-engine directory instead of hyrax-webapp. This appeared to cause bundler to get into a circular dependency loop and eventually run out of stack space, or would cause bundler to not find some gems. ​

Problems also stemmed from the use of a scoped source in the hyrax engine Gemfile (from #4700).  

* Bundle install gems using .dassie/Gemfile for dev docker images.
* Create hyrax-worker-base image that is the base for the other workers to avoid the onbuild bundle install in the dev worker.
* Have dassie load hyrax using `gemspec` instead of `gem` which allows the error catching hack to be removed.
* Undo the scoped source in hyrax's Gemfile and make dassie detect if a global source has already been loaded. This resolves bundler not being able to find compatible versions of gems.
* Update bundler in more places to ensure we have the latest version. Despite the hyrax-base image updating bundler, an old version could still be used if a cached image is used during the build.
  * The hyrax and hyrax-worker images still depend on hyrax-base to update bundler correctly. Should this be changed?
* Add the hyrax engine and dassie Gemfile.lock files to .dockerignore to avoid building with the host's gem environment.

* `docker-compose build` is successful
* The built images start without error using `docker-compose up`
* Able to run `docker-compose exec -w /app/samvera/hyrax-engine app sh -c "rubocop"` 
* Able to run `docker-compose exec -w /app/samvera/hyrax-engine app sh -c "rspec"` 

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/Gemfile
+++ b/.dassie/Gemfile
@@ -4,7 +4,7 @@ if @sources.global_rubygems_source == Bundler::SourceList.new.global_rubygems_so
   Bundler.ui.info '[Dassie] Adding global rubygems source.'
   source 'https://rubygems.org'
 else
-  Bundler.ui.warn "[Dassie] Global rubygems source already set: #{@sources.global_rubygems_source.inspect}"
+  Bundler.ui.info "[Dassie] Global rubygems source already set: #{@sources.global_rubygems_source.inspect}"
 end
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
--- a/.dassie/Gemfile
+++ b/.dassie/Gemfile
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 # Attempts to determine if a global gem source has ready been added by another Gemfile
 if @sources.global_rubygems_source == Bundler::SourceList.new.global_rubygems_source
-  Bundler.ui.warn '[Dassie] Adding global rubygems source.'
+  Bundler.ui.info '[Dassie] Adding global rubygems source.'
   source 'https://rubygems.org'
 else
   Bundler.ui.warn "[Dassie] Global rubygems source already set: #{@sources.global_rubygems_source.inspect}"
--- a/.dassie/Gemfile
+++ b/.dassie/Gemfile
@@ -1,4 +1,11 @@
-source 'https://rubygems.org'
+# frozen_string_literal: true
+# Attempts to determine if a global gem source has ready been added by another Gemfile
+if @sources.global_rubygems_source == Bundler::SourceList.new.global_rubygems_source
+  Bundler.ui.warn '[Dassie] Adding global rubygems source.'
+  source 'https://rubygems.org'
+else
+  Bundler.ui.warn "[Dassie] Global rubygems source already set: #{@sources.global_rubygems_source.inspect}"
+end
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby '2.7.2'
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -83,11 +86,11 @@ ENV HYRAX_ENGINE_PATH /app/samvera/hyrax-engine
 COPY --chown=1001:101 $APP_PATH /app/samvera/hyrax-webapp
 COPY --chown=1001:101 . /app/samvera/hyrax-engine
 
-RUN cd /app/samvera/hyrax-engine && bundle install --jobs "$(nproc)"
```
no-reply, line 86: sorry that it's taken me so long to come around to this. in my testing, this leaves the Hyrax dev environment broken/unready.

after pulling this and building `docker-compose exec -w /app/samvera/hyrax-engine app sh -c "rubocop"` gives `sh: rubocop: not found`, and similar results for `rspec` and other common development commands.

i think we really do want to bundle for the engine, in this context.

dlpierce: Okay, I've edited the PR description and will work on this more.

.dassie/Gemfile
```diff
@@ -1,4 +1,11 @@
-source 'https://rubygems.org'
+# frozen_string_literal: true
+# Attempts to determine if a global gem source has ready been added by another Gemfile
+if @sources.global_rubygems_source == Bundler::SourceList.new.global_rubygems_source
+  Bundler.ui.warn '[Dassie] Adding global rubygems source.'
```
no-reply, line 4:
```suggestion
  Bundler.ui.info '[Dassie] Adding global rubygems source.'
```

.dassie/Gemfile
```diff
@@ -1,4 +1,11 @@
...
+  source 'https://rubygems.org'
+else
+  Bundler.ui.warn "[Dassie] Global rubygems source already set: #{@sources.global_rubygems_source.inspect}"
```
no-reply, line 7:
```suggestion
  Bundler.ui.info "[Dassie] Global rubygems source already set: #{@sources.global_rubygems_source.inspect}"
```

.dassie/Gemfile
```diff
@@ -43,9 +50,7 @@ end
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
-begin # this is a hack to allow `eval_gemfile` to work correctly
-  gem 'hyrax', path: ENV.fetch('HYRAX_ENGINE_PATH', '..')
-rescue; end
+gemspec name: 'hyrax', path: ENV.fetch('HYRAX_ENGINE_PATH', '..')
```
no-reply, line 53: 👏🏻

Gemfile
```diff
@@ -22,6 +21,7 @@ test_app_gemfile = File.expand_path('Gemfile', test_app_path)
 # rubocop:disable Bundler/DuplicatedGem
 if File.exist?(test_app_gemfile)
   begin
+    Bundler.ui.warn "[Hyrax] Including test application dependencies from #{test_app_gemfile}"
```
no-reply, line 24:
```suggestion
    Bundler.ui.info "[Hyrax] Including test application dependencies from #{test_app_gemfile}"
```

Gemfile
```diff
@@ -32,5 +32,7 @@ elsif ENV['RAILS_VERSION'] == 'edge'
   ENV['ENGINE_CART_RAILS_OPTIONS'] = '--edge --skip-turbolinks'
 elsif ENV['RAILS_VERSION']
   gem 'rails', ENV['RAILS_VERSION'], source: 'https://rubygems.org'
+else
+  Bundler.ui.warn '[Hyrax] Skipping all Rails dependency injection'
```
no-reply, line 36: i think this one can stay `warn`. it seems like we shouldn't reach this case normally?

no-reply (changes requested):
this is great! i was able to get a clean build from here. i did need to delete my local `Gemfile.lock` files on my existing setup, but i think that's appropriate: what it means is that i can use my local copy of the app to control dependencies. it seems right to leave this control in the hands of the developer, and reasonable to trust them.

i think `info` is a better log level for the `Bundler.ui` output. otherwise this looks good to go.

no-reply (approved):

PR #4998, opened 2021-06-21
Updating kubernetes README

jeremyf:
These were the instructions that I followed to stand up a kubernetes
cluster.

Note, in the case of the ImagePullBackOff error, I tried clearing my
image cache and all kinds of "start with a clean slate."  However, the
errors continued to be somewhat erratic.  I was running through an 8MB
up/1MB down DSL connection and kept encountering problems.  I moved to a
15 MB up/15 MB down fiber connection at another location and the
instructions worked without a hitch.

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/README.md
+++ b/chart/hyrax/README.md
@@ -90,11 +90,57 @@ For those interested in trying out or contributing to this Chart, it's helpful
 to setup a simple cluster locally. Various projects exist to make this easy; we
 recommend [`k3d`][k3d] or [minikube][minikube].
 
-For example, with `k3d`:
+### Getting Started with K3D
+
+Make sure you have `docker`, `k3d`, and `helm` installed (e.g., for OSX run `brew install k3d` and `brew install helm`; and install `Docker Desktop`).
+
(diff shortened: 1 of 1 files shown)


chart/hyrax/README.md
```diff
@@ -90,12 +90,58 @@ For those interested in trying out or contributing to this Chart, it's helpful
...
 recommend [`k3d`][k3d] or [minikube][minikube].
 
-For example, with `k3d`:
+### Getting Started with K3D
+
+Make sure you have `docker`, `k3d`, and `helm` installed (e.g., for OSX run `brew install k3d` and `brew install helm`; and install `Docker Desktop`).
+
+Make sure you have Docker up and running.
+
+In the Hyrax directory, run the following command; it will pull images and when finished you'll have a Docker container runninig.
```
no-reply, line 99: the directory for this doesn't matter, so this can be:

```suggestion
Run the following command; it will pull images and when finished you'll have a Docker container runninig.
```

no-reply (approved):

Issue #3171, opened 2018-07-26
Review submissions dashboard not scoped as expected


bess:
### Descriptive summary

In Hyrax 2.1.0, as an admin, I expect to be able to go to the "Review Submissions" dashboard and see things that I need to review, and things that have completed the review process. However, the list of objects displayed in this view does not match my expectations when there are large numbers of Admin Sets and workflows in play. 

### Expected behavior

There are 138 `published` objects (where `published` is our final workflow state) in this repository:

![all_published_works](https://user-images.githubusercontent.com/65608/43285989-ac7c66fc-90ee-11e8-8fe7-71b7c397063b.png)

I would expect that if I go to the "Published" tab in the Review Submissions dashboard, I should see 138 objects.

### Actual behavior

However, logged in as an admin user who should have a role in all workflows, when I go to the Review Submissions dashboard to look at Published works, I only see 40:

![review_submissions_wrong](https://user-images.githubusercontent.com/65608/43286049-d922a658-90ee-11e8-9789-c00be3403703.png)
Note that we had to override `app/controllers/hyrax/admin/workflows_controller.rb` in order to change the name of the final workflow state:
```
    self.deposited_workflow_state_name = 'published'
```
AFAICT, the Review Submissions screen works as expected for small numbers of objects spread across small numbers of admin sets and workflows. However, this repository has over twenty admin sets, which use several different workflows. I believe the scoping of the query for Review submissions isn't returning everything it should.

### Steps to reproduce the behavior

Unfortunately I don't know how to reproduce this behavior easily without creating lots and lots of admin sets with different workflows. 
I'd be very interested to know if others are observing similar behavior.

rjkati:
In Hyrax 3.0.0 rc3, there are 1,644 published objects on nurax-dev
![published_numbers](https://user-images.githubusercontent.com/32644254/106193108-11a79880-617b-11eb-8a97-5c9d88a5766b.PNG)

The "Published" tab on the "Review Submissions" page shows 1,000 results.
![published_numbers_review_submissions](https://user-images.githubusercontent.com/32644254/106193501-98f50c00-617b-11eb-9f74-084224174e53.PNG)

dlpierce:
I made a new issue about the 1000 result limit seen by rjkati. The original issue appears to be a separate problem.

Looking at the referenced issues where this was originally seen, @mark-dce was the original reporter. Do you still encounter this problem in your system, and if so, would it be possible to take a closer look at the solr documents for the affected works?

[...]

PR #4999, opened 2021-06-21
Remove row limit

dlpierce:
This limit prevented the review submission dashboard from showing accurate result totals.

 refs #3171

This fixes the issue currently seen on nurax and other systems with large amounts of works. There may be further inaccuracies to be discovered after this limit is removed.

* The total amount of works reported in the review submissions dashboard agrees with the amount shown on the works dashboard.

@samvera/hyrax-code-reviewers


jeremyf (approved):
What's the default number of rows?

PR #5000, opened 2021-06-22
Fixed stray ul tag

anusharanganathan:
A simple PR to fix a stray ul tag in a search form.

This was noticed during a security audit and we were requested to fix it. We have done it locally, by overriding the file. It would be good to have it fixed upstream (in the 2.x branch though)

@samvera/hyrax-code-reviewers


no-reply (approved):

no-reply:
thank you @anusharanganathan!

PR #5003, opened 2021-06-24
chart: re-instate external fcrepo hostname support

no-reply:
when the fcrepo chart is disabled, still set the `FCREPO_HOST` and
`FCREPO_BASE_PATH` and other environment variables. the templates and variable
logic should (and already do?) handle all cases.

i think this regressed because of UC work to make it possible to simply skip
fcrepo. setting the environment variables is harmless in this case though. i
think we just got a bit overzealous.

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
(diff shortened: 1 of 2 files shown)


jeremyf (dismissed):

no-reply:
sorry, @jeremyf. i forgot to bump the version number

jeremyf (approved):

PR #5006, opened 2021-06-25
docker: run assets precompile on worker images

no-reply:
workers need precompiled assets too, even though they won't serve them, because
use of `ActionController::Base.helpers` in ThumbnailPathService (and perhaps
elsewhere).

@samvera/hyrax-code-reviewers


no-reply:
this provides a work around for https://github.com/samvera/hyrax/issues/5007

jeremyf (approved):

Issue #395, opened 2017-02-13
Alphasort field entries


hackartisan:
### Descriptive summary

With rdf2 values may come back in different orders. It could be disorienting for users if things jump around on page reload. We should alphasort here https://github.com/projecthydra-labs/hyrax/blob/master/app/renderers/hyrax/renderers/attribute_renderer.rb#L29. and maybe there are other places?

Should this be an option to attribute_to_html?

agoslen:
SMIG has been discussing open issues labeled metadata. We discussed this issue on 6/22:

- SMIG has recently made the recommendation (for #4781) that Creator, Contributor, and Title are the multi-value fields where ordering would matter most; but this order would be context-dependent, rather than purely alphabetical.
- Would it be desirable for alphasort to be a fallback for field ordering, where user-supplied order is not possible? 
  - No; group consensus is that alphasort is not necessarily better than the out-of-the-box chaos of RDF nonsorting. 
  - Alphasort implies an order and context that is artificial. If this is desirable to a local instance, it could be done locally.

no-reply:
thanks @agoslen. i think on this basis we should close this ticket out and subsume further discussion into #4781.

speaking to the original suggestion: https://github.com/samvera/hyrax/pull/5010 provides a very simple implementation of a `:sort` option for `attributes_to_html`

PR #5010, opened 2021-06-30
handle optional `:sort` option for attribute renderers

no-reply:
following a suggestion in #395, introduce a `:sort` option for renderers and a startlingly naive implementation.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/renderers/hyrax/renderers/attribute_renderer.rb
+++ b/app/renderers/hyrax/renderers/attribute_renderer.rb
@@ -11,12 +11,14 @@ module Hyrax
 
       attr_reader :field, :values, :options
 
+      ##
       # @param [Symbol] field
       # @param [Array] values
       # @param [Hash] options
       # @option options [String] :label The field label to render
       # @option options [String] :include_empty Do we render if if the values are empty?
       # @option options [String] :work_type Used for some I18n logic
+      # @option options [Boolean] :sort sort the values with +Array#sort+ if truthy
       def initialize(field, values, options = {})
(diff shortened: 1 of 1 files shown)


app/renderers/hyrax/renderers/attribute_renderer.rb
```diff
@@ -31,7 +31,10 @@ def render
 
         attributes = microdata_object_attributes(field).merge(class: "attribute attribute-#{field}")
 
-        markup += Array(values).map do |value|
+        values_array = Array(values)
+        values_array.sort! if options[:sort]
+
+        markup += values_array.map do |value|
```
cjcolvar, line 37: Do we need to worry at all about modifying the order of the passed in `values`?  Would `sort` be safer?

cjcolvar (commented):
This is cool and simple.  It would be helpful to include the option in the yard docs on #initialize.

app/renderers/hyrax/renderers/attribute_renderer.rb
no-reply: probably. good call. 👍🏻

no-reply:
> This is cool and simple. It would be helpful to include the option in the yard docs on #initialize.

yeah, you're right. i'll take a second pass through to do some quick docs and a test in the AM.

no-reply:
removed array mutation and added a yard annotation for the new option

jeremyf (approved):

PR #5013, opened 2021-07-06
File set visibility fixes

ashikajith:
Small fixes and improvements building on #4992 to allow using this code from a controller by permitting the new parameters in the form and casting uploaded_file_id to an integer since it will be a string when posted by the form.

The job code was also refactored to appease rubocop's length requirements.

Future work is submitting these parameters from the deposit form.

TODO: Add tests


cjcolvar:
Replaced by #5022

PR #5020, opened 2021-07-12
Enable feature specs to pass with HUB_URL/Selenium

mcritchlow:
* Remove `CAPYBARA_SERVER` env var
* Replace `CAPYBARA_SERVER` usage with `IPSocket.getaddress`
* Replace hard-coded `example.com` reference in twitter feature spec assertion
	with `Capybara.app_host`

Locally, in the `docker-compose` environment, if I run all the feature specs
via:

```sh
% docker-compose exec -w /app/samvera/hyrax-engine app sh -c "bundle exec rspec spec/features"
```

I can get all test to pass:

```sh
Finished in 10 minutes 50 seconds (files took 5.67 seconds to load)
157 examples, 0 failures, 4 pending
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,3 @@
-CAPYBARA_SERVER=http://app:3010
 CHROME_HEADLESS_MODE=false
 DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL=true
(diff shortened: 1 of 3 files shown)


no-reply:
@mcritchlow looks like this is those tests we were talking about.

no-reply (approved):
fantastic!

PR #5021, opened 2021-07-13
Extracting method to ease future switching logic

jeremyf:
﻿## Extracting method to ease future switching logic

bf9e13b6e84b73696a64b212a30ef20369e888d1

In discussion with @no_reply, in addressing #4195, we want to explore
the following path:

1) As early as possible in the request cycle, create a
   Hyrax::UploadedFile for both a CarrierWave file or a file selected
   for upload by BrowseEverything.
2) Building on step 1, we would then hope to reduce branching logic in the
   request cycle, and preserve existing class interfaces.

By making this non-breaking change, I'm orienting the code to deliver on
this aspirational goal.

Next steps would be:

1) Update the Hyrax::WorksControllerBehavior to create
   Hyrax::UploadedFile records for the files selected from
   BrowseEverything (note CarrierWave handles the "inline" uploads).
2) Adjust the extracted `Hyrax::UploadedFile#perform_ingest_later`
   method to handle files that were created from BrowseEverything.

The plan is to add a field to the UploadedFile (perhaps
`url_of_file_to_ingest`) that the Hyrax::WorksControllerBehavior would
populate based on user input parameters.

With that in place, we could look at backporting the solution into the
ActorStack.

## Handling CarrierWave/BrowseEverything controller logic

b2f2af41313f04e757ba37912a2cfaa1f6f11eb1

Prior to this commit, the Hyrax::WoksControllerBehavior's Valkyrie
branching logic ignored files selected via BrowseEverything.  With
this change the controller normalizes(-ish) the two methods of file
upload:

1) CarrierWave
2) BrowseEverything

An immediate follow-up step is to update
`Hyrax::UploadedFile.perform_ingest_later`.  In it's present
implementation, that method assumes a CarrierWave object.  Adding a
switching case for UploadedFiles that have a
`url_of_remote_file_for_ingest` (e.g., a file selected via
BrowseEverything) instead of the assumptive `file` attribute (e.g., a
file uploaded via CarrierWave).

Relates to #4195


jeremyf:
Writing some notes for future work.  Based on the work of #5036 and of the current refactoring, there's continues to be some ghastly branching logic.  When I started this, my expectation was that I could add a step between "upload file" and attach file to file set.  That added step would optionally go and fetch the remote file.

For now, I'm going to ignore the actor stack (even though I refactored [Hyrax::Actors::CreateWithRemoteFilesActor](https://github.com/samvera/hyrax/blob/04f257ad5b63f99688d44c77cbda7d2e3f90aa8b/app/actors/hyrax/actors/create_with_remote_files_actor.rb#L12).  The [JobIoWrapper.create_with_varied_file_handling!](https://github.com/samvera/hyrax/blob/04f257ad5b63f99688d44c77cbda7d2e3f90aa8b/app/models/job_io_wrapper.rb#L37-L50) appears to be a reasonable place to explore.  Both the JobIoWrapper and the CreateWithRemoteFilesActor make use of the implemented (though misnamed) Hyrax::Actors::FileActor.

PR #5023, opened 2021-07-13
deprecate FindObjectsViaSolrService replaced by SolrQueryService#get_objects

elrayle:
Use `SolrQueryService#get_objects` instead of `FindObjectsViaSolrService` when objects are required.  If solr documents are sufficient, then just use `SolrQueryService#get`.

By using `SolrQueryService`, this allows for the construction of more complex queries even when objects are being returned.

Also updated `MultipleMembershipChecker` to use `SolrQueryService` instead of `FindObjectViaSolrService`.  This was the only use of `FindObjectViaSolrService`.

Related to: Issue #3820 which is replacing ActiveFedora::Base.where calls.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/find_objects_via_solr_service.rb
+++ b/app/services/hyrax/find_objects_via_solr_service.rb
@@ -24,7 +24,7 @@ module Hyrax
         solr_query_builder.new
                           .with_model(model: model)
                           .with_field_pairs(field_pairs: field_pairs, join_with: join_with, type: type)
-                          .get_objects(use_valkyrie: use_valkyrie)
+                          .get_objects(use_valkyrie: use_valkyrie).to_a
       end
     end
   end
--- a/app/services/hyrax/find_objects_via_solr_service.rb
+++ b/app/services/hyrax/find_objects_via_solr_service.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 module Hyrax
-  # This class is being replaced by Hyrax::SolrQueryService #get_objects method.
+  # @deprecated This class is being replaced by Hyrax::SolrQueryService #get_objects method.
   #
   # Methods in this class search solr to get the ids and then use the query service to find the objects.
   class FindObjectsViaSolrService
--- a/app/services/hyrax/find_objects_via_solr_service.rb
+++ b/app/services/hyrax/find_objects_via_solr_service.rb
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 module Hyrax
+  # This class is being replaced by Hyrax::SolrQueryService #get_objects method.
+  #
   # Methods in this class search solr to get the ids and then use the query service to find the objects.
   class FindObjectsViaSolrService
     class_attribute :solr_query_builder, :solr_service, :query_service
-    self.solr_query_builder = Hyrax::SolrQueryBuilderService
+    self.solr_query_builder = Hyrax::SolrQueryService
     self.solr_service = Hyrax::SolrService
     self.query_service = Hyrax.query_service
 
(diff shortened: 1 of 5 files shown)


app/services/hyrax/find_objects_via_solr_service.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 module Hyrax
+  # This class is being replaced by Hyrax::SolrQueryService #get_objects method.
```
no-reply, line 3:
```suggestion
  # @deprecated This class is being replaced by Hyrax::SolrQueryService #get_objects method.
```

app/services/hyrax/solr_query_service.rb
```diff
@@ -17,6 +20,21 @@ def get
...
+      results = get
+      results['response']['docs'].map { |doc| doc['id'] }
+    end
+
+    ##
+    # @return [Array<Valkyrie::Resource|ActiveFedora::Base>] objects matching the current query
+    def get_objects(use_valkyrie: Hyrax.config.use_valkyrie?)
+      ids = get_ids
+      return ids.map { |id| ActiveFedora::Base.find(id) }.to_a unless use_valkyrie
+      query_service.find_many_by_ids(ids: ids).to_a
```
no-reply, line 35: is the `#to_a` desirable, or could this return an `Enumerator`?

it seems like avoiding loading data that won't be used is attractive.

no-reply (commented):
looks good!

one small doc suggestion, and a question are in inline comments

app/services/hyrax/solr_query_service.rb
elrayle: That makes sense. I'll remove it and see if everything continues to work.  I used `to_a` because the deprecated method used it.  I can't remember the original reason for doing using `to_a`.

app/services/hyrax/find_objects_via_solr_service.rb
```diff
@@ -14,13 +16,15 @@ class << self
...
-        results = solr_service.query(query)
-        ids = results.map(&:id)
-        query_service.find_many_by_ids(ids: ids).to_a
+        Deprecation.warn("'##{__method__}' will be removed in Hyrax 4.0.  " \
+                         "Instead, use 'Hyrax::SolrQueryService.new.with_model(...).with_field_pairs(...).get_objects'.")
+
+        solr_query_builder.new
+                          .with_model(model: model)
+                          .with_field_pairs(field_pairs: field_pairs, join_with: join_with, type: type)
+                          .get_objects(use_valkyrie: use_valkyrie).to_a
```
elrayle, line 27: FindObjectViaSolrService #find_for_model_by_field_pairs needs to continue to return an Array for backward compatibility.

This method is no longer used in any Hyrax code.

app/services/hyrax/multiple_membership_checker.rb
```diff
@@ -55,9 +55,12 @@ def single_membership_collections(collection_ids)
 
       field_pairs = {
         :id => collection_ids,
-        Hyrax.config.collection_type_index_field.to_sym => collection_type_gids_that_disallow_multiple_membership
+        Hyrax.config.collection_type_index_field.to_sym => collection_type_gids_that_disallow_multiple_membership&.map(&:to_s)
```
elrayle, line 58: The map to strings is required to turn GlobalIDs into string IDs.  I'm not sure why this worked before.  It definitely does not work with the SolrQueryService.

app/services/hyrax/solr_query_service.rb
```diff
@@ -17,6 +20,21 @@ def get
...
+      results = get
+      results['response']['docs'].map { |doc| doc['id'] }
+    end
+
+    ##
+    # @return [Array<Valkyrie::Resource|ActiveFedora::Base>] objects matching the current query
+    def get_objects(use_valkyrie: Hyrax.config.use_valkyrie?)
+      ids = get_ids
+      return ids.map { |id| ActiveFedora::Base.find(id) } unless use_valkyrie
+      query_service.find_many_by_ids(ids: ids)
```
elrayle, line 35: Removed `.to_a` on the returned result to maintain the Enumerable.

app/services/hyrax/multiple_membership_checker.rb
```diff
@@ -55,9 +55,12 @@ def single_membership_collections(collection_ids)
...
       }
-      Hyrax::FindObjectsViaSolrService.find_for_model_by_field_pairs(model: ::Collection, field_pairs: field_pairs, use_valkyrie: true)
+      Hyrax::SolrQueryService.new
+                             .with_model(model: ::Collection)
+                             .with_field_pairs(field_pairs: field_pairs, join_with: ' OR ')
+                             .get_objects(use_valkyrie: true).to_a
```
elrayle, line 63: The result is required to be an Array because later it is used in...

```
collections_to_check |= single_membership_collections(item.member_of_collection_ids)
```
where the result is `collections_to_check`.

no-reply (approved):

PR #5025, opened 2021-07-13
add accessible_by to SolrQueryService

elrayle:
Partially addresses Issue #3820 by removing use of `ActiveFedora::Base #where` method when using `#accessible_by` method.

See original [`#accessible_by`](https://github.com/samvera/hydra-head/blob/main/hydra-access-controls/lib/active_fedora/accessible_by.rb) method defined in hydra-access-controls.

See Hyrax override of [`#gated_discovery_filters`](https://github.com/samvera/hyrax/blob/b034218b89dde7df534e32d1e5ade9161e129a1d/app/search_builders/hyrax/search_filters.rb#L16-L21).  Note: To have access to this method, the SolrQueryService needs to inherit from `::SearchBuilder`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/collection_presenter.rb
+++ b/app/presenters/hyrax/collection_presenter.rb
@@ -78,19 +78,36 @@ module Hyrax
     end
 
     def total_items
-      Hyrax::SolrService.new.count("member_of_collection_ids_ssim:#{id}")
+      field_pairs = { "member_of_collection_ids_ssim" => id.to_s }
+      SolrQueryService.new
+                      .with_field_pairs(field_pairs: field_pairs)
+                      .count
     end
 
     def total_viewable_items
-      ActiveFedora::Base.where("member_of_collection_ids_ssim:#{id}").accessible_by(current_ability).count
+      field_pairs = { "member_of_collection_ids_ssim" => id.to_s }
+      SolrQueryService.new
+                      .with_field_pairs(field_pairs: field_pairs)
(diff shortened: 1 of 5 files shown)


app/presenters/hyrax/collection_presenter.rb
```diff
@@ -78,19 +78,36 @@ def size
     end
 
     def total_items
-      Hyrax::SolrService.new.count("member_of_collection_ids_ssim:#{id}")
+      field_pairs = { "member_of_collection_ids_ssim" => id.to_s }
+      SolrQueryService.new
+                      .with_field_pairs(field_pairs: field_pairs)
+                      .count
```
elrayle, line 84: This change is not strictly necessary, but it makes the code more consistent.

app/services/hyrax/solr_query_service.rb
```diff
@@ -3,7 +3,7 @@ module Hyrax
   ##
   # Methods in this class are providing functionality previously supported by ActiveFedora::SolrQueryBuilder.
   # It includes methods to build and execute a query.
-  class SolrQueryService
+  class SolrQueryService < ::SearchBuilder
```
elrayle, line 6: Inheriting from `SearchBuilder` is required to have access to the private method `gated_discovery_filters`.

no-reply:
seems like this needs a manual rebase now

no-reply (approved):

PR #5027, opened 2021-07-14
Reset acl refactor

no-reply:
extract and provide one implementation of `AdminSet/Collection#reset_access_controls!`

these methods had a lot in common. extracting them to a single implementation sets us up to add a valkyrie ready version of this behavior.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/permission_template_form.rb
+++ b/app/forms/hyrax/forms/permission_template_form.rb
@@ -12,7 +12,13 @@ module Hyrax
       # @return [#to_s] the primary key of the associated admin_set or collection
       # def source_id (because you might come looking for this method)
       delegate :id, to: :source_model, prefix: :source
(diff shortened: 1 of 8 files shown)


app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -12,7 +12,13 @@ class PermissionTemplateForm
       # @return [#to_s] the primary key of the associated admin_set or collection
       # def source_id (because you might come looking for this method)
       delegate :id, to: :source_model, prefix: :source
-      delegate :reset_access_controls!, to: :source_model
+
+      ##
+      # @deprecated
```
no-reply, line 17:
```suggestion
      # @deprecated use PermissionTemplate#reset_access_controls instead
```

Issue #4765, opened 2021-03-02
Create a solr doc in Valkyrie solr when creating/modifying an AdminSet


elrayle:
### Descriptive summary

When writing to Valkyrie's solr is enabled, AdminSets are written to ActiveFedora managed solr core but not to the Valkyrie solr core.

### Rationale

AdminSets are a core model type and should be saved in the Valkyrie solr core.

### Expected behavior

When `config.query_index_from_valkyrie = true` in `config/initializers/hyrax.rb` and an AdminSet is created or edited...

* should write an AdminSet solr doc to ActiveFedora managed solr core (at least for now)
* should write an AdminSet solr doc to Valkyrie's solr core 

### Actual behavior

When `config.query_index_from_valkyrie = true` in `config/initializers/hyrax.rb` and an AdminSet is created or edited...

* writes an AdminSet solr doc to ActiveFedora managed solr core
* DOES NOT write an AdminSet solr doc to Valkyrie's solr core 

#### Prerequisites

Edit `.internal_test_app/config/initializers/hyrax.rb` and set config

```
  config.query_index_from_valkyrie = true
```

#### Get count of AdminSet solr docs

Get counts through Solr UI for ActiveFedora Solr core:
1. navigate to: http://localhost:8983/solr/#/hydra-development/query
2. check facet and set `facet.field=has_model_ssim`
3. go to bottom of results and record counts for model types
    * AdminSet

Get model type counts in Solr UI for Valkyrie Solr core:
1. navigate to: http://localhost:8987/solr/#/hyrax-valkyrie-test/query
2. check facet and set `facet.field=has_model_ssim`
3. go to bottom of results and record counts for model types
    * AdminSet

#### Check for count changes when adding an Admin Set

BEFORE get counts of AdminSets for both ActiveFedora Solr core and Valkyrie Solr core.

1. Login as an admin
2. Dashboard -> Collections -> New Collection -> type=Admin Set
3. Fill in required fields and Save

AFTER get counts of AdminSets for both ActiveFedora Solr core and Valkyrie Solr core.

Verify that the counts for AdminSets increased by one in both Solr cores.

### Related work

* Issue #4756 - UBER: Not all model types are saved to the Valkyrie Solr core 
* Issue #4766 - Create a FileSet solr doc in Valkyrie solr when files are added to a Work
* Issue #4767 - Create AccessControl solr docs in Valkyrie solr when creating Works, Collections, and AdminSets

elrayle:
[...]

PR #5032, opened 2021-07-15
Add valkyrie_administrative_set_indexer

mcritchlow:
I believe this is related to #4765

* Add an administrative_set indexer
* Add an administrative_set spec

@samvera/hyrax-code-reviewers


no-reply:
this looks good. i'm pulling it down to pick up work from here.

Issue #5030, opened 2021-07-15
Move CollectionMemberService to CollectionMemberSearchService


elrayle:
### Descriptive summary

It is confusing to have search related methods in a CollectionMemberService.  We want to use this class name for non-search related member services.  It is undesirable to have a mix of search and non-search code in the same class.

For this reason, the current search code in CollectionMemberService will be moved to CollectionMemberSearchService.  Then the initializer in CollectionMemberService will create an instance of CollectionMemberSearchService and delegate all the existing methods to that class.  The methods in CollectionMemberService will be deprecated.

### Related work

Link to related tickets or prior related work here.

PR #5033, opened 2021-07-16
create service to add/remove collection members

### LINKED ISSUE #5031: Create a service to handle adding/removing collection members
### Prerequisite

Deprecate existing search related methods in CollectionMemberService.  See Issue #5030

### Descriptive summary

Due to legacy code, Valkyrie collections and their children have a dual-direction set of attributes for the parent-child relationship. For performance reasons, it is desirable to have the relationship stored only on the child.  This avoids the need to reindex the parent when a new child is added.

To avoid confusion and potential mistakes, a service will be created to add and remove collection members.  The service will be implemented in `services/hyrax/collections/collection_member_service.rb`.

NOTE: The Valkyrie find custom queries for collections already uses the inverse relationship (i.e. search the children for a matching parent).  So no work needs to be done for those.

### Rationale

For data integrity, the same data (in this case the relationship between the parent and child) should not be stored in multiple places.

### Expected behavior

There is one and only one place to add and remove collection membership.

### Actual behavior

There are multiple places and approaches for adding/removing collection membership.

### Related work

Issue #5030

elrayle:
Fixes #5031

Centralizes the code to add and remove collection members to avoid confusion about which relationship to use (i.e. child's `member_of_collection_ids` and collection's `member_ids`) and to handle pre-processing (e.g. multiple collection membership check).

Also updates the Valkyrie collection factory with the following changes:

* Updated the relationship being used for collection membership to `member_of_collection_ids`.  Since the collection id must exist, the processing of members was moved to `after :create`

* Remove `collection_type` as an attribute.  Before this change you could pass in `collection_type` and/or `collection_type_gid`.  The change removes the confusion of what should happen if both are passed in and represent different types.  Since only `collection_type_gid` was in use, `collection_type` was removed.

NOTE: Issue #5031 stated that Issue #5030 was a prerequisite, but the identified deprecations are happening after addressing Issue #5031.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/collections/collection_member_service.rb
+++ b/app/services/hyrax/collections/collection_member_service.rb
@@ -108,14 +108,11 @@ module Hyrax
         # @param new_member [Hyrax::Resource] the new child collection or child work
         # @return [Hyrax::Resource] updated member resource
         def add_member(collection:, new_member:, user:)
-          message = Hyrax::MultipleMembershipChecker.new(item: new_member).check(collection_ids: collection.id, include_current_members: true)
-          if message
-            new_member.errors.add(:collections, message)
-          else
-            new_member.member_of_collection_ids += [collection.id] # only populate this direction
-            new_member = Hyrax.persister.save(resource: new_member)
-            Hyrax.publisher.publish('object.metadata.updated', object: new_member, user: user)
-          end
+          message = Hyrax::MultipleMembershipChecker.new(item: new_member).check(collection_ids: [collection.id], include_current_members: true)
+          raise Hyrax::SingleMembershipError, message if message
+          new_member.member_of_collection_ids += [collection.id] # only populate this direction
+          new_member = Hyrax.persister.save(resource: new_member)
+          Hyrax.publisher.publish('object.metadata.updated', object: new_member, user: user)
           new_member
         end
 
--- a/app/services/hyrax/collections/collection_member_service.rb
+++ b/app/services/hyrax/collections/collection_member_service.rb
@@ -67,6 +67,96 @@ module Hyrax
         end
         response
       end
+
+      class << self
+        # Check if a work or collection is already a member of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param member [Hyrax::Resource] the child collection and/or child work to check
+        # @return [Boolean] true if already in the member set; otherwise, false
+        def member?(collection:, member:)
+          member.member_of_collection_ids.include? collection.id
+        end
+
+        # Add works and/or collections as members of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_member_ids [Enumerable<Valkyrie::ID>] the ids of the new child collections and/or child works
+        # @return [Enumerable<Hyrax::Resource>] updated member resources
+        def add_members_by_ids(collection:, new_member_ids:, user:)
+          new_members = Hyrax.query_service.find_many_by_ids(ids: new_member_ids)
+          add_members(collection: collection, new_members: new_members, user: user)
+        end
+
+        # Add works and/or collections as members of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_members [Enumerable<Hyrax::Resource>] the new child collections and/or child works
+        # @return [Enumerable<Hyrax::Resource>] updated member resources
+        def add_members(collection:, new_members:, user:)
+          new_members.map { |new_member| add_member(collection: collection, new_member: new_member, user: user) }
+        end
+
+        # Add a work or collection as a member of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_member_id [Valkyrie::ID] the id of the new child collection or child work
+        # @return [Hyrax::Resource] updated member resource
+        def add_member_by_id(collection:, new_member_id:, user:)
+          new_member = Hyrax.query_service.find_by(id: new_member_id)
+          add_member(collection: collection, new_member: new_member, user: user)
+        end
+
+        # Add a work or collection as a member of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_member [Hyrax::Resource] the new child collection or child work
+        # @return [Hyrax::Resource] updated member resource
+        def add_member(collection:, new_member:, user:)
+          message = Hyrax::MultipleMembershipChecker.new(item: new_member).check(collection_ids: collection.id, include_current_members: true)
+          if message
+            new_member.errors.add(:collections, message)
+          else
+            new_member.member_of_collection_ids += [collection.id] # only populate this direction
+            new_member = Hyrax.persister.save(resource: new_member)
+            Hyrax.publisher.publish('object.metadata.updated', object: new_member, user: user)
+          end
+          new_member
+        end
+
+        # Remove collections and/or works from the members set of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param member_ids [Enumerable<Valkyrie::ID>] the ids of the child collections and/or child works to be removed
+        # @return [Enumerable<Hyrax::Resource>] updated member resources
+        def remove_members_by_ids(collection:, member_ids:, user:)
+          members = Hyrax.query_service.find_many_by_ids(ids: member_ids)
+          remove_members(collection: collection, members: members, user: user)
+        end
+
+        # Remove collections and/or works from the members set of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param members [Enumerable<Valkyrie::Resource>] the child collections and/or child works to be removed
+        # @return [Enumerable<Hyrax::Resource>] updated member resources
+        def remove_members(collection:, members:, user:)
+          members.map { |member| remove_member(collection: collection, member: member, user: user) }
+        end
+
+        # Remove collections and/or works from the members set of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param member_id [Valkyrie::ID] the id of the child collection or child work to be removed
+        # @return [Hyrax::Resource] updated member resource
+        def remove_member_by_id(collection:, member_id:, user:)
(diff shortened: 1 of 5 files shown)


app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
         end
         response
       end
+
+      class << self
+        # Check if a work or collection is already a member of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param member [Hyrax::Resource] the child collection and/or child work to check
+        # @returns [Boolean] true if already in the member set; otherwise, false
```
no-reply, line 75:
```suggestion
        # @return [Boolean] true if already in the member set; otherwise, false
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def member?(collection:, member:)
+          member.member_of_collection_ids.include? collection.id
+        end
+
+        # Add works and/or collections as members of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_member_ids [Enumerable<Valkyrie::ID>] the ids of the new child collections and/or child works
+        # @returns [Enumerable<Hyrax::Resource>] updated member resources
```
no-reply, line 83:
```suggestion
        # @return [Enumerable<Hyrax::Resource>] updated member resources
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def add_members_by_ids(collection:, new_member_ids:, user:)
+          new_members = Hyrax.query_service.find_many_by_ids(ids: new_member_ids)
+          add_members(collection: collection, new_members: new_members, user: user)
+        end
+
+        # Add works and/or collections as members of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_members [Enumerable<Hyrax::Resource>] the new child collections and/or child works
+        # @returns [Enumerable<Hyrax::Resource>] updated member resources
```
no-reply, line 92:
```suggestion
        # @return [Enumerable<Hyrax::Resource>] updated member resources
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def add_members(collection:, new_members:, user:)
+          new_members.map { |new_member| add_member(collection: collection, new_member: new_member, user: user) }
+        end
+
+        # Add a work or collection as a member of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_member_id [Valkyrie::ID] the id of the new child collection or child work
+        # @returns [Hyrax::Resource] updated member resource
```
no-reply, line 100:
```suggestion
        # @return [Hyrax::Resource] updated member resource
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def add_member_by_id(collection:, new_member_id:, user:)
+          new_member = Hyrax.query_service.find_by(id: new_member_id)
+          add_member(collection: collection, new_member: new_member, user: user)
+        end
+
+        # Add a work or collection as a member of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_member [Hyrax::Resource] the new child collection or child work
+        # @returns [Hyrax::Resource] updated member resource
```
no-reply, line 109:
```suggestion
        # @return [Hyrax::Resource] updated member resource
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+            new_member = Hyrax.persister.save(resource: new_member)
+            Hyrax.publisher.publish('object.metadata.updated', object: new_member, user: user)
+          end
+          new_member
+        end
+
+        # Remove collections and/or works from the members set of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param member_ids [Enumerable<Valkyrie::ID>] the ids of the child collections and/or child works to be removed
+        # @returns [Enumerable<Hyrax::Resource>] updated member resources
```
no-reply, line 125:
```suggestion
        # @return [Enumerable<Hyrax::Resource>] updated member resources
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def remove_members_by_ids(collection:, member_ids:, user:)
+          members = Hyrax.query_service.find_many_by_ids(ids: member_ids)
+          remove_members(collection: collection, members: members, user: user)
+        end
+
+        # Remove collections and/or works from the members set of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param members [Enumerable<Valkyrie::Resource>] the child collections and/or child works to be removed
+        # @returns [Enumerable<Hyrax::Resource>] updated member resources
```
no-reply, line 134:
```suggestion
        # @return [Enumerable<Hyrax::Resource>] updated member resources
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def remove_members(collection:, members:, user:)
+          members.map { |member| remove_member(collection: collection, member: member, user: user) }
+        end
+
+        # Remove collections and/or works from the members set of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param member_id [Valkyrie::ID] the id of the child collection or child work to be removed
+        # @returns [Hyrax::Resource] updated member resource
```
no-reply, line 142:
```suggestion
        # @return [Hyrax::Resource] updated member resource
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def remove_member_by_id(collection:, member_id:, user:)
+          member = Hyrax.query_service.find_by(id: member_id)
+          remove_member(collection: collection, member: member, user: user)
+        end
+
+        # Remove a collection or work from the members set of a collection, also removing the inverse relationship
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param member [Hyrax::Resource] the child collection or child work to be removed
+        # @returns [Hyrax::Resource] updated member resource
```
no-reply, line 151:
```suggestion
        # @return [Hyrax::Resource] updated member resource
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        def remove_member(collection:, member:, user:)
+          return collection unless member?(collection: collection, member: member)
```
no-reply, line 153: we return the `member` in the method's primary code path, so i think we should do the same for this guard condition.

```suggestion
          return member unless member?(collection: collection, member: member)
```

app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -67,6 +67,96 @@ def available_member_work_ids
...
+        end
+
+        # Add a work or collection as a member of a collection
+        # @param collection [Hyrax::PcdmCollection] the collection
+        # @param new_member [Hyrax::Resource] the new child collection or child work
+        # @returns [Hyrax::Resource] updated member resource
+        def add_member(collection:, new_member:, user:)
+          message = Hyrax::MultipleMembershipChecker.new(item: new_member).check(collection_ids: collection.id, include_current_members: true)
+          if message
+            new_member.errors.add(:collections, message)
```
no-reply, line 113: where does the `#errors` method come from. this looks like an `ActiveModel` thing, and i'm surprised it's working here.

my intuition is that this method should `raise` something in the case that it refuses to add the member. i'm not sure a valkyrie user will know to check `errors` to determine if this method succeeded or not (would they need to check it every time?).

no-reply (changes requested):
this looks great.

i think we can't use `#errors` (is this getting added by Wings? why?), so we'll need to find another solution for error handling.

a few docs suggestions and a fix for a variable return type are also in the review.

app/services/hyrax/collections/collection_member_service.rb
elrayle: This is part of the MultipleMembershipChecker.  I can change it to raise the error and see if anything breaks.

app/services/hyrax/collections/collection_member_service.rb
elrayle: Good catch.  When we were setting the relationship on the collection, it returned collection.  It should have been updated to member when the relationship now that the relationship is only set on the member.  I'll double check all the returns.

elrayle:
FYI... I am working on adding the rspec check for the metadata updated listener.

app/services/hyrax/collections/collection_member_service.rb
elrayle: FYI... This code came from [models/concerns/hyrax/collection_behavior.rb #add_member_objects](https://github.com/samvera/hyrax/blob/v3.0.2/app/models/concerns/hyrax/collection_behavior.rb#L61-L75).

no-reply: i think that code is handling only ActiveFedora objects.

no-reply: i'm still pretty sure this will fail if the member passed in isn't generated by Wings. e.g. if it's a `Monograph`.

elrayle: DONE: Change to raise an exception.

elrayle:
@no-reply Requested changes are in and tests are green.  Ready for your review.

no-reply (approved):

PR #5034, opened 2021-07-16
deprecate methods in CollectionMemberService; moved to CollectionMemberSearchService

### LINKED ISSUE #5030: Move CollectionMemberService to CollectionMemberSearchService

elrayle:
Fixes #5030

This sets the stage to use CollectionMemberService for non-search related member services.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/collections_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/collections_controller_behavior.rb
@@ -22,7 +22,7 @@ module Hyrax
       # The search builder to find the collection
       self.single_item_search_builder_class = SingleCollectionSearchBuilder
       # The search builder to find the collections' members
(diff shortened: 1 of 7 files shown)


no-reply (dismissed):

no-reply:
ack. looks like it needs a rebase

no-reply (approved):

PR #5035, opened 2021-07-16
Add valkyrie_administrative_set_indexer

no-reply:
supersedes #5032.

renames the proposed class to ensure it's resolved correctly by the indexer factory.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/indexers/hyrax/administrative_set_indexer.rb
+++ b/app/indexers/hyrax/administrative_set_indexer.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Indexes Hyrax::AdministrativeSet objects
+  class AdministrativeSetIndexer < Hyrax::ValkyrieIndexer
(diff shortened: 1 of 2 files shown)


dlpierce:
This looks pretty similar to where I got to when I was working on #4765. I had started to add in some changes to the AdministrativeSet model, but it was still in progress.

dunn:
so [AdminSet](https://github.com/samvera/hyrax/blob/main/app/models/admin_set.rb) is the AF version and [AdministrativeSet](https://github.com/samvera/hyrax/blob/main/app/models/hyrax/administrative_set.rb) is the Valkyrized one?

no-reply:
> so [AdminSet](https://github.com/samvera/hyrax/blob/main/app/models/admin_set.rb) is the AF version and [AdministrativeSet](https://github.com/samvera/hyrax/blob/main/app/models/hyrax/administrative_set.rb) is the Valkyrized one?

that's right.

dunn (approved):

PR #5036, opened 2021-07-16
Extracting class as interstitial refactor

jeremyf:
I'm continuing to work on #4195.  This change extracts a class that can
be called outside the ActorStack.  The goal being to leverage this
class's logic to ingest remote URLs.

This change looks to remove lots of repeated logic and instead provide a
singular point of entry for associating a remote file with a work.

@samvera/hyrax-code-reviewers


no-reply:
this looks fantastic. thanks!

no-reply (approved):

PR #5038, opened 2021-07-16
add some documentation for `Hyrax::PermissionTemplate`

no-reply:
the existing docs for this code were broadly unhelpful. this is what i could
figure out about how this works in brief.

it would be useful to write up some code examples and use cases here, too.

@samvera/hyrax-code-reviewers


no-reply:
trying to mitigate this:
pt

jeremyf (approved):

PR #5039, opened 2021-07-19
expand documentation for Sipity::Entity

no-reply:
try to expand and improve {Sipity::Entity} documentation.

@samvera/hyrax-code-reviewers


lsitu (approved):
@no-reply Nice! I think this will help us dev understand how the Sipity::Entity link to works/objects and think about how user comments can be added.

PR #5040, opened 2021-07-19
add top level module documentation for Sipity with some use cases

no-reply:
this is an attempt to document Sipity at the top level. the focus here is on
describing the domain model and explaining some of the use cases. i think a lot
more is required to make this crystal clear; e.g. example code, state diagrams,
and a deep description of Roles, Notifications and Methods is probably called
for.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/sipity.rb
+++ b/app/models/sipity.rb
@@ -2,6 +2,36 @@
 
 ##
 # {Sipity} is a workflow/state engine.
+#
+# This module and the classes namespaced within it provide a domain model and
+# implementation for managing the movement of repository objects through a
+# defined workflow. The workflows themselves are configured using JSON documents
+# and loaded into the database/domain model as {Sipity::Workflow}. Each workflow
+# can be understood as finite sets of {Sipity::WorkflowState},
+# {Sipity::WorkflowAction} (transitions from state to state), and {Sipity::Role}
+# authorized to carry out each action.
+#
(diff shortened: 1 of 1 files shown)


app/models/sipity.rb
```diff
@@ -2,6 +2,36 @@
...
+# authorized to carry out each action.
+#
+# Any uniquely identifiable object can be managed by a {Sipity::Workflow}.
+# Normally Hyrax uses workflows to handle the deposit process and maintenance
+# lifecycle of repository objects at the level of the Work (within the Hydra
+# Works model). Objects are represented within the Sipity engine's domain model
+# by a {Sipity::Entity}. Each object has at most one {Sipity::Entity}, is
+# governed by one {Sipity::Workflow}, and in one {Sipity::WorkflowState} at any
+# given time.
+#
```
lsitu, line 21: Do we want to mention Sipity::EntitySpecificResponsibility and workflow comments feature at this top level?

lsitu (approved):
@no-reply Looks great! Thanks. I just added a little question.

app/models/sipity.rb
no-reply: yeah. i think so. i'm just starting to understand how these fit in, and haven't yet figured out how to document them. do you have ideas?

lsitu: I don't know what's the best way to address it either but it's a feature that we may interest in for Comet workflow. Maybe we can add a use case for it?

PR #5041, opened 2021-07-19
refactor view dependency on AdminSetOptionsPresenter into a helper

no-reply:
this is a straight extraction of the `_form_relationships` partial's dependency
on `AdminSetOptionsPresenter` and `AdminSetService` into a helper. none of the
actual behavior here is changed at all, but the extraction presents an idiomatic
seam for further reworking, and for downstream overrides.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/work_form_helper.rb
+++ b/app/helpers/hyrax/work_form_helper.rb
@@ -2,6 +2,16 @@
 module Hyrax
   module WorkFormHelper
     ##
+    # @todo this implementation hits database backends (solr) and is invoked
+    #   from views. refactor to avoid
+    # @return  [???]
+    def admin_set_options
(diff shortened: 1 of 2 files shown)


app/helpers/hyrax/work_form_helper.rb
```diff
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 module Hyrax
   module WorkFormHelper
+    ##
+    # @todo this implementation hits database backends (solr) and is invoked
+    #   from views. refactor to avoid
+    # @return  [???]
```
no-reply, line 7:
```suggestion
    # @return  [Array<Array<String, String, Hash>] options for the admin set drop down.
```

dunn (approved):

PR #5042, opened 2021-07-21
allow Valkyrie::Resource collections to be members of other collections

elrayle:
Adds the member_of_collections_ids attribute to Hyrax::PcdmCollections.

Also, updates the tests for Hyrax::CollectionsController spec to test using Valkyrie resources only.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/controllers/hyrax/collections_controller_spec.rb
+++ b/spec/controllers/hyrax/collections_controller_spec.rb
@@ -5,17 +5,19 @@ RSpec.describe Hyrax::CollectionsController, clean_repo: true do
   let(:other) { build(:user) }
 
   let(:collection) do
-    create(:public_collection_lw, title: ["My collection"],
-                                  description: ["My incredibly detailed description of the collection"],
-                                  user: user)
+    FactoryBot.valkyrie_create(:hyrax_collection,
+                               :public,
+                               title: ["My collection"],
+                               description: ["My incredibly detailed description of the collection"],
+                               edit_users: [user.user_key], read_users: [user.user_key])
   end
 
-  let(:asset1)         { create(:work, title: ["First of the Assets"], user: user) }
-  let(:asset2)         { create(:work, title: ["Second of the Assets"], user: user) }
-  let(:asset3)         { create(:work, title: ["Third of the Assets"], user: user) }
-  let(:asset4)         { build(:collection_lw, title: ["First subcollection"], user: user) }
-  let(:asset5)         { build(:collection_lw, title: ["Second subcollection"], user: user) }
-  let(:unowned_asset)  { create(:work, user: other) }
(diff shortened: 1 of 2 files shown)


spec/controllers/hyrax/collections_controller_spec.rb
```diff
@@ -1,130 +1,162 @@
...
 
-      context 'when the page parameter is passed' do
-        it 'loads the collection (paying no attention to the page param)' do
-          get :show, params: { id: collection, page: '2' }
-          expect(response).to be_successful
-          expect(assigns[:presenter]).to be_kind_of Hyrax::CollectionPresenter
-          expect(assigns[:presenter].to_s).to eq 'My collection'
-        end
-      end
+      describe "#show" do # public landing page
```
elrayle, line 48: The tests are unchanged except for indentation.  All changes were in the setup.

spec/controllers/hyrax/collections_controller_spec.rb
```diff
@@ -1,130 +1,162 @@
...
-  routes { Hyrax::Engine.routes }
-  let(:user)  { create(:user) }
-  let(:other) { build(:user) }
-
-  let(:collection) do
-    create(:public_collection_lw, title: ["My collection"],
-                                  description: ["My incredibly detailed description of the collection"],
-                                  user: user)
-  end
+  [false, true].each do |use_valkyrie|
```
no-reply, line 3: i'm a little uncertain about the need for this. much of the behavior in this test setup is already using valkyrie directly, and the internal behavior of the mixin should be equivalent however the test is setup, right?

what's the motivation for the duplicate test runs?

elrayle: The error with the missing attribute `member_of_collection_ids` on `Hyrax::PcdmCollection` was not uncovered by these tests.  There wasn't a failing test until `use_valkyrie=true`.

This is part of what prompted me to ask if we want to start migrating tests to only run with `Valkyrie` objects and `use_valkyrie=true` when possible.  In that case, the change here would be to only run the `use_valkryie=true` set of tests.

no-reply: i think that's right. this controller should function the same way however the objects were created. migrating the tests to setup with valkyrie seems like something we can just do, without any branching or iteration.

elrayle: Ok, I will drop the AF setup and only test with Valkyrie resources.

elrayle: @no-reply This change has been made.  The tests only use Valkyrie resources.

no-reply (approved):

PR #5047, opened 2021-07-22
move resolution of available admin sets to controller

no-reply:
reduces the queries needed to build the work form by retrieving the data needed
to determine admin set behavior directly from the controller and all at once.

the older implementation looped over the admin sets and pulled the
`PermissionTemplate` for each. this version uses Solr to retrieve the admin set
solr documents, then gets the permission templates matching them all at once
before grouping the two sets together in-memory.

giving the controller direct responsibility for the query behavior should be
more flexible and open the door to other optimizations.

---

i don't love how unclear the controller behavior becomes, so maybe there are some refactors still to take on here. there definitely should be controller specs.

picking this up again tomorrow.

cc: @mcritchlow @dunn


app/helpers/hyrax/work_form_helper.rb
```diff
@@ -6,8 +6,9 @@ module WorkFormHelper
     #   from views. refactor to avoid
     # @return  [Array<Array<String, String, Hash>] options for the admin set drop down.
     def admin_set_options
-      service = Hyrax::AdminSetService.new(controller)
+      return @admin_set_options.select_options if @admin_set_options
```
mcritchlow, line 9: With this PR, it seems like we're making this the default. So how would one fall
through this guard `return` and hit the older Service/Presenter setup below? Is it in the case where folks have overridden the controller downstream?

spec/presenters/hyrax/admin_set_selection_presenter_spec.rb
```diff
@@ -28,10 +28,40 @@
     its(:id) { is_expected.to eq admin_set.id }
 
     describe '#data' do
-      it 'is a hash' do
+      it 'is a hash with no releose delay and restricted visibility' do
```
mcritchlow, line 31: s/releose/release

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -437,5 +439,21 @@ def concern_has_file_sets?
     def uploaded_files
       UploadedFile.find(params.fetch(:uploaded_files, []))
     end
+
+    def available_admin_sets
```
mcritchlow, line 443: It seems like the pattern in the controllers is to store methods that do this sort of work inline. Should we consider a controller helper?

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -437,5 +439,21 @@ def concern_has_file_sets?
...
+      admin_set_results = Hyrax::AdminSetService.new(self).search_results(:deposit)
+      # get all the templates at once, reducing query load
+      templates = PermissionTemplate.where(id: admin_set_results.map(&:id)).to_a
```
mcritchlow, line 446: Not a useful comment but this is a great optimization

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -437,5 +439,21 @@ def concern_has_file_sets?
     def uploaded_files
       UploadedFile.find(params.fetch(:uploaded_files, []))
     end
+
+    def available_admin_sets
+      admin_set_results = Hyrax::AdminSetService.new(self).search_results(:deposit)
```
mcritchlow, line 444: Do we need a branch on this initial AdminSet/AdministrativeSet query to support different query backends?

app/helpers/hyrax/work_form_helper.rb
no-reply: any controller override would hit it, but i also think there's at least one other controller (DashboardCollections ?) that routes to this helper.

no-reply: i had been thinking to explicitly deprecate the fall-through path here once all the relevant internal controllers are populating `@admin_set_options`

app/controllers/concerns/hyrax/works_controller_behavior.rb
no-reply: can you say more? i'm not sure i'm understanding the suggestion.

app/controllers/concerns/hyrax/works_controller_behavior.rb
no-reply: i'm glad you commented, because i think we probably want to get a controller test in that ensures this is working right.

we want this working reasonably for the following cases:

  - no admin sets exist;
  - one or more admin sets exist, and they all have permission templates;
  - one or more admin sets exist, and at least one is missing a permission template;
  - one or more admin sets exist, and all are missing permission templates.

app/controllers/concerns/hyrax/works_controller_behavior.rb
no-reply: i don't think so. i think this is all Solr based, and using blacklight infrastructure.

unless/until Fedora supports a fast lookup for all admin sets filtered on deposit access, i think we probably want it to stay that way.

app/controllers/concerns/hyrax/works_controller_behavior.rb
mcritchlow: Sorry about that. At best, it's probably suggesting a premature optimization, and initially I was reading the wrong file location..

I guess in short, in this controller concern in particular, there are quite a few methods doing various things.

So I'm wondering if it makes sense to eventually extract this to its own concern/class. Something like that.

app/helpers/hyrax/work_form_helper.rb
mcritchlow: Ah, ok that's great. Thanks!

no-reply:
superseded by #5050

PR #5048, opened 2021-07-22
don't lookup `Sipity::Workflow` instances from their own ids

no-reply:
`PermissionTemplate#active_workflow` returns a `Sipity::Workflow`. the older
code used its id to retrieve it from the database again. instead we can simply
use it as-is.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/admin_set_options_presenter.rb
+++ b/app/presenters/hyrax/admin_set_options_presenter.rb
@@ -59,16 +59,9 @@ module Hyrax
       # > remove files from a work, and add new works to the set.
       return true if @current_ability.can?(:manage, permission_template)
 
-      # Otherwise, we check if the workflow was setup, active, and
-      # allows_access_grants.
-      wf = workflow(permission_template: permission_template)
-      return false unless wf
-      wf.allows_access_grant?
-    end
-
-    def workflow(permission_template:)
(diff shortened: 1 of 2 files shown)


app/presenters/hyrax/admin_set_options_presenter.rb
```diff
@@ -59,16 +59,9 @@ def sharing?(permission_template:)
...
-      # allows_access_grants.
-      wf = workflow(permission_template: permission_template)
-      return false unless wf
-      wf.allows_access_grant?
-    end
-
-    def workflow(permission_template:)
-      return unless permission_template.active_workflow
-      Sipity::Workflow.find_by!(id: permission_template.active_workflow.id)
+      # Otherwise, we check if the workflow active workflow allows access
```
no-reply, line 62:
```suggestion
      # Otherwise, we check if the active workflow allows access grants
```

app/presenters/hyrax/admin_set_options_presenter.rb
```diff
@@ -59,16 +59,9 @@ def sharing?(permission_template:)
...
+      # grants.
```
no-reply, line 63:
```suggestion
```

dunn (approved):

PR #5053, opened 2021-07-23
open AdminSetSearchBuilder to post-Wings model scope

no-reply:
this search builder should find objects of class `AdminSet` and
`Hyrax::AdministrativeSet`. with Wings enabled, `AdministrativeSet` already
resolves its `internal_resource` as `AdminSet`. for apps without wings, its
desirable for this service to find objects indexed with `has_model_ssim` of
`Hyrax::AdministrativeSet`.

maybe we also want to move this builder toward using the "generic type" index?
that would require a broad reindex from downstream apps, because the legacy
admin sets don't seem to include any generic type data.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/admin_set_search_builder.rb
+++ b/app/search_builders/hyrax/admin_set_search_builder.rb
@@ -14,7 +14,7 @@ module Hyrax
 
     # This overrides the models in FilterByType
     def models
-      [::AdminSet]
+      [::AdminSet, Hyrax::AdministrativeSet]
     end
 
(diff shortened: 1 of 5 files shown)


dunn (approved):

dunn (approved):

dunn:
not sure what to make of this error, only appearing on 2.6 https://app.circleci.com/pipelines/github/samvera/hyrax/5748/workflows/0d47312b-b196-4a42-8afe-a9252b0f6125/jobs/43308/parallel-runs/0/steps/0-111

PR #5055, opened 2021-07-27
fix solr query in MultipleMembershipChecker  and refactor to improve readability

elrayle:
The solr query being generated was causing the multiple membership #check to fail if there were two collections of the same single membership type even if the item was not in both collections.  The tests failed to catch this very common scenario.  

This led to a refactor of the code which was complex and difficult to follow.  Now each method has one and only one task and is clearly named for that task.

Also the reason the tests failed to catch the common scenario is that the access to solr was heavily mocked and returned what was expected to be returned from solr and not what was actually being returned from solr.  This was hiding errors in the generated solr query for multiple tests.  The tests are rewritten to organize them into clear scenarios based on the options for passed in parameters and the state of membership for the item using real solr access.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/multiple_membership_checker.rb
+++ b/app/services/hyrax/multiple_membership_checker.rb
@@ -6,8 +6,7 @@ module Hyrax
   class MultipleMembershipChecker
     attr_reader :item
 
-    # @param [#member_of_collection_ids] item an object that belongs to
-    #   collections
+    # @param [#member_of_collection_ids] item an object that belongs to collections
     def initialize(item:)
       @item = item
     end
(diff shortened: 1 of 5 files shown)


app/services/hyrax/multiple_membership_checker.rb
```diff
@@ -20,53 +19,63 @@ def initialize(item:)
...
-    # wholesale replacement and when making an incremental change, such as
-    # adding a single collection membership to an object. In the former case,
-    # `#check` only scans the passed-in collection identifiers. In the latter,
-    # `#check` must also scan the collections to which an object currently
-    # belongs for potential conflicts.
-    #
     # @param collection_ids [Array<String>] a list of collection identifiers
-    # @param include_current_members [Boolean] a flag to also scan an object's
-    #   current collection memberships
+    # @param include_current_members [Boolean] if true, include item's existing collections in check; else if false, check passed in collections only
```
no-reply, line 23: can we shorten the line lengths here? i try to set a soft cap at 80 chars especially in consideration of https://github.com/samvera-labs/bixby/issues/48.

since these lines can be wrapped without syntax considerations, it seems best to do so.

spec/spec_helper.rb
```diff
@@ -108,6 +108,7 @@ def clean_active_fedora_repository
   # The JS is executed in a different thread, so that other thread
   # may think the root path has already been created:
   ActiveFedora.fedora.connection.send(:init_base_path)
+  Hyrax.persister.wipe! if Hyrax.config.query_index_from_valkyrie
```
no-reply, line 111: i'm not following this. the `Hyrax.persister` is independent of the index, right?

elrayle: You would think that.  And perhaps this should be somewhere else, but I'm just using the existing `wipe!` method defined in wings persister.

https://github.com/samvera/hyrax/blob/4efa6b8f3133fbcb282cda7e791b5bbd7a07d3d7/lib/wings/valkyrie/persister.rb#L60-L65

elrayle: An alternative...

```
# define in Hyrax::SolrService
def wipe!
   Hyrax::SolrService.delete_by_query("*:*") 
   Hyrax::SolrService.commit 
end
```

```
# update in Wings::Valkyrie::Persister
def wipe!
  Hyrax::SolrService.wipe!
  ActiveFedora::Cleaner.clean!
end
```

elrayle:
FYI... I kicked off the failing tests again.  The two that failed on the last round fail locally whether on my branch or main.  And they passed for some ruby versions.  It looks like they may be random failures.

spec/spec_helper.rb
no-reply: i don't think it's safe to assume that `Hyrax.persister` is a the wings persister (it's not for many tests).

i think i'm not understanding the use case in general, either. this method is named `#clean_active_fedora_repository`; when/why does that need to clear the custom, non-AF solr index?

app/services/hyrax/multiple_membership_checker.rb
elrayle: I shortened all long lines.

spec/spec_helper.rb
elrayle: I added a `wipe!` method to `Hyrax::SolrService` and moved that call to the `:clean_repo` section of `spec_helper.rb` since that is what triggers the call from the spec.  I think the assumption here is that the writer of the test does want all repos (e.g. Fedora, Redis, Solr) cleared to start the tests with a clean slate, including Valkyrie solr if it is in use.  

Long term, this should perhaps call `Hyrax.persister.wipe!` with the theory that if `:clean_repo` is being called by a spec, then that persister was used.  Or we may want to make this more granular (e.g. `:wipe_hyrax_persister`, `:wipe_solr`, etc.).

no-reply:
yeah. the two failing specs are new in the last week or so, and appear to be a test order issue. i'll try to clear it later today.

no-reply (approved):

Issue #4788, opened 2021-03-15
Hyrax::Works do not show thumbnail in Dashboard -> Works


elrayle:
### Prerequisite to Resolving this Issue

Need PR #4787 (Resources need to include the depositor to be included in search results) to be merged first.  Without this PR, the monograph won't be in the list of works on Dashboard -> Works.

### Descriptive summary

Any work that is_a Hyrax::Work (e.g. Monograph), which is a Valkyrie::Resource, does not include its thumbnail in the list of works in Dashboard -> Works.

### Rationale

All works with thumbnails should show the thumbnail in the list of works on Dashboard -> Works -> Your Works (tab).

### Expected behavior

Monographs that have thumbnails should have the thumbnail included in the list of works in Dashboard -> Works -> Your Works (tab).

`gw03-15-01 prime` is_a `GenericWork` is_a `ActiveFedora::Base` and should show a thumbnail.
`m03-15-01` is_a `Monograph` is_a `Valkyrie::Resource` and should show a thumbnail.

### Actual behavior

Monographs that have thumbnails show the title but not the thumbnail in the list of works in Dashboard -> Works -> Your Works (tab).

`gw03-15-01 prime` is_a `GenericWork` is_a `ActiveFedora::Base` and shows a thumbnail.
`m03-15-01` is_a `Monograph` is_a `Valkyrie::Resource`` and does NOT show a thumbnail.

![image](https://user-images.githubusercontent.com/6855473/111330479-5b264700-8646-11eb-801f-f44c69e0d830.png)

### Steps to reproduce the behavior

1. Login
2. Dashboard -> Works -> Add new work -> select Monograph as the work type
3. Fill in required fields
4. Add an image file
5. Save (forwards to the show page for the Monograph work)
6. Click Works in the breadcrumb to return to Dashboard -> Works

The new Monograph is listed, but the thumbnail is not displayed.

### Related Issues

Issue #4795 - Hyrax::Works do not show Date Added in Dashboard -> Works
Issue #4794 - Hyrax::Works do not index file sets
Issue #4487 - Solr docs generated by ActiveFedora and Hyrax::ValkyrieWorkIndexer are not compatible

hackartisan:
related to https://github.com/samvera/hyrax/issues/4191?

abelemlih:
❌.  @jlhardes This ticket did not pass QA. I created a Monograph, and added an image file, and was unable to see the image as a thumbnail of the monograph under the "Works" list, as shown below:

[...]

PR #5056, opened 2021-07-28
Data seeds improvements

apertome:
Reimplement database seeds using Valkyrie rather than ActiveFedora. Add option for multiple seed files and an environment variable to control whether they are loaded

* Create db seeds using Valkyrie
* Split seeds into 2 files, 1 for users, 1 for collections/works
* Create collections/works as described in this sheet: https://docs.google.com/spreadsheets/d/1b93me1Gh614eQEWzeqzsSw32PDr_ymIfJ-PvrR7zxDM/edit#gid=291627674

A few caveats:
* First attempt at implementing this used Monograph but had some issues in dassie with that. So the work is created as a GenericWork and the `valkyrie_resource` is retrieved.
* Not all the categories from the sheet are implemented yet
* File attachment is not implemented yet
* The script attempts to set the collections/works as public, but this doesn't seem to actually work as implemented

@samvera/hyrax-code-reviewers


jlhardes:
Work on #4788 Hyrax::Works do not show thumbnail in Dashboard -> Works also connected to this PR work.

jlhardes:
Pretty sure this PR is still something we want to do, even if it isn't complete yet. I have a zip of files to provide to cover minimum needs for seeding different types of works. I can contribute that if we want to finish this up.

apertome:
Agree that we still want to do this.

apertome:
Pushed some updates to this, but I'm getting some flakiness still. 

Current behavior looks like:
- Some seed collections and works are created
- Currently works are being created as GenericWorks. I think this is mostly working but no image attachments yet -- still  need to figure out how to do that.
- The seeds script is able to create Works and automatically assicate with Default Admin Set

generic_work_default_admin_set

- It can also associate Works with Collections:

generic_work_in_a_collection


- The seed script is not creating Monograph works, but when I try to add some manually via the GUI, and upload files, it seems like it recognized that there are supposed to be files but any time you try to access them (either to view, download, or edit) there are errors like this:

hyrax_objectnotfounderror

dlpierce:
Closing in favor of #5472

PR #5057, opened 2021-07-28
deprecate CollectionBehavior #add_member_objects and replace with CollectionMemberService.add_members

elrayle:
Partially addresses Issue #4920 

_NOTE: The issue says to update to support Valkyrie objects, but instead methods are being deprecated in favor of services in order to keep the Valkyrie resource object light weight._

* deprecates `Hyrax::CollectionBehavior#add_member_objects`
* updates all places that call it to `Hyrax::Collections::CollectionMemberService.add_members`

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -112,10 +112,10 @@ module Hyrax
         @collection.collection_type_gid = params[:collection_type_gid].presence || default_collection_type.to_global_id
         @collection.attributes = collection_params.except(:members, :parent_id, :collection_type_gid)
         @collection.apply_depositor_metadata(current_user.user_key)
-        add_members_to_collection unless batch.empty?
         @collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE unless @collection.discoverable?
         if @collection.save
           after_create
+          add_members_to_collection unless batch.empty?
         else
           after_create_error
         end
@@ -379,7 +379,9 @@ module Hyrax
 
       def add_members_to_collection(collection = nil)
         collection ||= @collection
-        collection.add_member_objects batch
+        Hyrax::Collections::CollectionMemberService.add_members_by_ids(collection: collection.valkyrie_resource,
+                                                                       new_member_ids: batch,
+                                                                       user: current_user)
       end
 
       def remove_members_from_collection
--- a/app/services/hyrax/collections/collection_member_service.rb
+++ b/app/services/hyrax/collections/collection_member_service.rb
@@ -81,7 +81,15 @@ module Hyrax
         # @param new_members [Enumerable<Hyrax::Resource>] the new child collections and/or child works
         # @return [Enumerable<Hyrax::Resource>] updated member resources
         def add_members(collection:, new_members:, user:)
-          new_members.map { |new_member| add_member(collection: collection, new_member: new_member, user: user) }
+          messages = []
+          new_members.map do |new_member|
+            begin
+              add_member(collection: collection, new_member: new_member, user: user)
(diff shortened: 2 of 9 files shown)


app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -81,7 +81,15 @@ def add_members_by_ids(collection:, new_member_ids:, user:)
...
         # @return [Enumerable<Hyrax::Resource>] updated member resources
         def add_members(collection:, new_members:, user:)
-          new_members.map { |new_member| add_member(collection: collection, new_member: new_member, user: user) }
+          messages = []
+          new_members.map do |new_member|
+            begin
+              add_member(collection: collection, new_member: new_member, user: user)
+            rescue Hyrax::SingleMembershipError => err
+              messages += [err.message]
+            end
```
elrayle, line 90: It was passing through the exception at the first member that violated single membership.  But the original code for adding members built up each violation message and passed back the unique messages.  This is updating adding multiple members at a time to do the same thing.

spec/models/collection_spec.rb
```diff
@@ -55,14 +55,15 @@
...
-      let(:work2) { create(:work) }
-      let(:work3) { create(:work) }
-
-      it "allows multiple files to be added" do
-        collection.add_member_objects [work1.id, work2.id]
-        collection.save!
-        expect(collection.reload.member_objects).to match_array [work1, work2]
+      let(:work1) { valkyrie_create(:hyrax_work, title: 'Work 1') }
+      let(:work2) { valkyrie_create(:hyrax_work, title: 'Work 2') }
+      let(:work3) { valkyrie_create(:hyrax_work, title: 'Work 3') }
```
elrayle, line 60: Everywhere CollectionMemberService adding members is called in the code, it calls `add_members_by_ids`.  Since this turns around and instantiates the members before calling `add_members`, the tests save a step and just pass in members.  The code is actually adds the members expects a `Hyrax::Resource`.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -112,10 +112,10 @@ def create
         @collection.collection_type_gid = params[:collection_type_gid].presence || default_collection_type.to_global_id
         @collection.attributes = collection_params.except(:members, :parent_id, :collection_type_gid)
         @collection.apply_depositor_metadata(current_user.user_key)
-        add_members_to_collection unless batch.empty?
         @collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE unless @collection.discoverable?
         if @collection.save
           after_create
+          add_members_to_collection unless batch.empty?
```
elrayle, line 118: The collection has to be saved before adding members, otherwise there isn't an id for the new collection to add to the member's `member_of_collection_ids`.

spec/models/collection_spec.rb
```diff
@@ -79,26 +80,30 @@
...
 
         it 'fails to add the member' do
-          collection.add_member_objects [work1.id, work2.id, work3.id]
-          collection.save!
-          expect(collection.reload.member_objects).to match_array [work1, work3]
+          begin
+            Hyrax::Collections::CollectionMemberService.add_members(collection: collection.valkyrie_resource,
+                                                                    new_members: [work1, work2, work3],
+                                                                    user: nil)
+          rescue; end # rubocop:disable Lint/SuppressedException
```
elrayle, line 87: Have to capture the exception or the test will stop here.

app/controllers/hyrax/dashboard/collections_controller.rb
no-reply: is this a new issue, or a fix for a standing bug?

no-reply (approved):

PR #5058, opened 2021-07-29
add a rake task to regenerate derivatives

no-reply:
it's a common pitfall in deploying Hyrax apps to misconfigure
thumbnail/derivatives path. when the configuration changes, the thumbnails will
be missing. depending on whether the storage medium they were originally
generated to still exists, it can be impossible to recover the old thumbnails.

adding a task to regen the derivatives completely seems of high utility.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/tasks/regenerate_derivatives.rake
+++ b/lib/tasks/regenerate_derivatives.rake
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+namespace :hyrax do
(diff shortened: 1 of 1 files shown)


orangewolf (approved):

orangewolf:
I have looked this up and found an old slack message from Julie A at least 4 times. Thank you for taking the extra few minutes to get it committed

PR #5060, opened 2021-08-02
Make initContainers more flexible for downstream chart users

mcritchlow:
Downstream `extraInitContainers` may be dependent, in a way, on the
upstream `initContainers` running first.

For example, if one wants to add an additional Solr collection (say for
running tests against). If a local `extraInitContainer` is created to do
this, it may be implicitly dependent on the `load-solr-config`
`initContainer` having already ran (to reference the `configset`).

It may also be desirable for downstream users of the Hyrax chart to run their
own database setup `initContainers`, or potentially external to the Helm
process entirely.

This introduces the `skipDbMigrateSeed` value, which defaults to `false`
to preserve existing behavior. But provides a mechanism for downstream
users to opt-out.

* Move `extraInitContainers` to the end of the `initContainers` section, which
	follows the general pattern of extras going at the end of definitions.
* Add `skipDbMigrateSeed` Helm chart value, defaulting to `false`, to allow
	downstream users to choose to opt-out of running the `db-setup`
	`initContainer` 

* A `helm deploy` with the defaults should continue to work as-is
* A `helm deploy` setting `skipDbMigrateSeed: true` should skip the `db-setup`
	`initContainer`

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 0.21.1
+version: 0.21.2
 appVersion: 3.0.2
 dependencies:
   - name: fcrepo
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 0.21.2
+version: 0.22.0
 appVersion: 3.0.2
 dependencies:
   - name: fcrepo
--- a/chart/hyrax/templates/deployment-worker.yaml
+++ b/chart/hyrax/templates/deployment-worker.yaml
@@ -20,9 +20,6 @@ spec:
         {{- include "hyrax.selectorLabels" . | nindent 8 }}
     spec:
       initContainers:
-        {{- if .Values.worker.extraInitContainers }}
-        {{- toYaml .Values.worker.extraInitContainers | nindent 8 }}
-        {{- end }}
         - name: db-wait
           image: "{{ .Values.worker.image.repository }}:{{ .Values.worker.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.worker.image.pullPolicy }}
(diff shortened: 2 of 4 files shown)


dunn (approved):

no-reply:
@mcritchlow the original stated use case for `extraInitContainers` was

> A use case for this might be waiting on an additional database not natively monitored by the chart, but is required to be up and running for the application to start.

in this case, it seems like we'd need the init containers to run early, before the seed container runs (is this also why we need to add the skip seed gate here?).

i'm okay with moving these to be late running tasks, if it turns out that works better for most use cases, but i felt like we made a principled decision at the start and i want to be sure we're not just shuffling this for short term convenience. do you think that's where we're at? what other solutions are available (could we add `skipSolrSetup` and add a custom init container for our entire solr setup?).

mcritchlow:
Great, thanks for the feedback. The more I looked at this and thought about it, I guess where I landed was that downstream order dependency of `initContainers` of any kind is kind of problematic. So part of what I was trying to go for here was a bit more flexibility for downstream users to setup `initContainers` as they'd like, while the default `initContainer` setup from the Chart still "just works".

> in this case, it seems like we'd need the init containers to run early, before the seed container runs (is this also why we need to add the skip seed gate here?).

Yeah, that's right.

> but i felt like we made a principled decision at the start

Ok, if that's true I suppose I forget the context or wasn't aware of it :) I guess the pattern I've seen, for whatever that's worth, is to put `extra`-anything at the end of the template definition. 

> could we add skipSolrSetup and add a custom init container for our entire solr setup?

Yeah, we could definitely go that route as well. And might be more consistent with doing the same with the database setup.

In short, I'm happy to make whatever adjustments you think would be more preferable. The intention isn't a short-term fix, so if there's a better long-term strategy lets do that.

chart/hyrax/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 0.21.1
+version: 0.21.2
```
no-reply, line 5:
```suggestion
version: 0.22.0
```

no-reply:
i'm convinced by the argument that `extraEnvVars` (and secrets) follow an append pattern.

no-reply (approved):

no-reply:
@mcritchlow i bumped the minor version (since the order dependency change is breaking, a minor version bump seems called for). if this looks good to you, it's approved. if you merge, i can push the new chart version manually.

i may try to get that CI change for chart pushes up today.

mcritchlow:
@no-reply - yep that sounds great to me. And thanks for catching the versioning issue.

PR #5067, opened 2021-08-12
test Valkyrie object creation and edit

no-reply:
do an end-to-end browser test of object creation for valkyrie native works.

this test seemed to be missing. it's important to have this in as an example for
adopters and as a baseline for related tests.

also add an edit test; this depends on `hydra-editor` 5.0.5

@samvera/hyrax-code-reviewers


no-reply:
@rotated8 i wonder if you have any ideas what is going on with these builds.

i'm explictly asking for `bundler` 2.2.25, but the error i'm getting (only on some bulids) is

> You must use Bundler 2 or greater with this lockfile.

https://app.circleci.com/pipelines/github/samvera/hyrax/5805/workflows/da5d97e0-387d-4d6e-8cbe-7185e842cb71/jobs/43893?invite=true#step-103-4

PR #5073, opened 2021-08-16
collection membership service uses collection id instead of collection resource

elrayle:
Refactor: Since none of the methods in this service need the collection resource, there is no reason to require a collection resource.  It is more efficient to pass and use the collection’s id when that is all that is needed.

@samvera/hyrax-code-reviewers


no-reply (approved):
thanks. it seems good to avoid requiring callers to have a whole object when the internals only need the id.

PR #5074, opened 2021-08-16
add parent collections navigator

elrayle:
This follows the established navigator pattern.  It will be used as part of the process for replacing uses of ActiveFedora relationship `member_of_collections`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/custom_queries/navigators/parent_collections_navigator.rb
+++ b/app/services/hyrax/custom_queries/navigators/parent_collections_navigator.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module CustomQueries
+    module Navigators
+      ##
+      # Navigate from a resource to the parent collections of the resource.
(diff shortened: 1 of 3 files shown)


app/services/hyrax/custom_queries/navigators/parent_collections_navigator.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+
+        ##
+        # Find the ids of parent collections of a given resource, and map to Valkyrie Resources IDs
+        #
+        # @param [Valkyrie::Resource] resource
+        #
+        # @return [Array<Valkyrie::ID>]
+        def find_parent_collection_ids(resource:)
+          resource.member_of_collection_ids
```
elrayle, line 41: This is super simple.  It is included to follow the pattern of having navigators that return either the resource objects or the ids.

no-reply (approved):

Issue #5071, opened 2021-08-16
WINGS: remove uses of `member_of_collections`


elrayle:
### Descriptive summary

Hydra::Pcdm defines `member_of_collections` on objects and collections using...

https://github.com/samvera/hydra-pcdm/blob/0b1aeaea7ff96bd16ad26b0cf21fd0341331c92f/lib/hydra/pcdm/models/concerns/pcdm_behavior.rb#L50-L56

As this uses ActiveFedora to define the relationship and the Valkyrie::Resource version of collections defines the relationship using...

https://github.com/samvera/hyrax/blob/5752a4acf6f6779b4c601636367075607486ae95/app/models/hyrax/pcdm_collection.rb#L14

Need to replace all references to `member_of_collections` with the Valkyrie equivalent.

### Rationale

As part of the hryax-valkyrization process, all references to ActiveFedora are being removed.

### ActiveFedora usage

Examples of setting the relationship...
```
child.member_of_collections.push(parent)
child.member_of_collections << collection
child.member_of_collections += [collection]
```

Example of removing the relationship...
```
child.member_of_collections.delete(parent)
```

Examples of getting the objects from the relationship...
```
parents = child.member_of_collections
```

### Replacement Valkryie usage

Examples of setting the relationship...
```
child.member_of_collection_ids << parent.id
```

Examples of removing the relationship...
```
child.member_of_collection_ids.delete(parent.id)
```

Examples of getting the objects from the relationship...
```
parents = Hyrax.query_service.find_many_by_idsHyrax(ids: child.member_of_collection_ids)

child_collections = Hyrax.custom_queries.find_child_collections(resource: parent)
child_works =  Hyrax.custom_queries.find_child_works(resource: parent)
```

### To find

Search for `member_of_collections`.

PR #5076, opened 2021-08-17
update NestedCollectionPersistenceService to use CollectionMemberService

elrayle:
Since CollectionMemberService publishes `'object.metadata.updated'` and message published wants to attribute the action to a user, the persistence service methods added a `user:` parameter which defaults to nil for backwards compatibility.  Callers of these methods within Hyrax code were updated to pass a user or updated to call CollectionmemberService directly to avoid having to create an instance of the parent collection.

Partially addresses #5071

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/collections/nested_collection_persistence_service_spec.rb
+++ b/spec/services/hyrax/collections/nested_collection_persistence_service_spec.rb
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
 RSpec.describe Hyrax::Collections::NestedCollectionPersistenceService, with_nested_reindexing: true do
-  let(:parent) { build(:collection_lw) }
-  let(:child) { create(:collection) }
+  let(:parent) { FactoryBot.valkyrie_create(:hyrax_collection) }
+  let(:child) { FactoryBot.valkyrie_create(:hyrax_collection) }
 
   describe '.persist_nested_collection_for' do
     subject { described_class.persist_nested_collection_for(parent: parent, child: child) }
 
     it 'creates the relationship between parent and child' do
       subject
-      expect(parent.member_objects).to eq([child])
-      expect(child.member_of_collections).to eq([parent])
+      expect(Hyrax.custom_queries.find_parent_collection_ids(resource: child)).to eq [parent.id]
+      expect(Hyrax.custom_queries.find_child_collection_ids(resource: parent)).to eq [child.id]
     end
   end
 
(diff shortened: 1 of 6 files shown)


spec/services/hyrax/collections/nested_collection_persistence_service_spec.rb
```diff
@@ -22,8 +22,8 @@
 
     it 'removes the relationship between parent and child' do
       subject
-      expect(parent.member_objects).to eq([])
-      expect(child.member_of_collections).to eq([])
+      expect(Hyrax.custom_queries.find_parent_collection_ids(resource: child)).to eq []
```
no-reply, line 25:
```suggestion
      expect(Hyrax.custom_queries.find_parent_collection_ids(resource: child)).to be_empty
```

spec/services/hyrax/collections/nested_collection_persistence_service_spec.rb
```diff
@@ -22,8 +22,8 @@
...
+      expect(Hyrax.custom_queries.find_child_collection_ids(resource: parent)).to eq []
```
no-reply, line 26:
```suggestion
      expect(Hyrax.custom_queries.find_child_collection_ids(resource: parent)).to be_empty
```

spec/services/hyrax/collections/nested_collection_persistence_service_spec.rb
```diff
@@ -1,15 +1,15 @@
...
+  let(:child) { FactoryBot.valkyrie_create(:hyrax_collection) }
 
   describe '.persist_nested_collection_for' do
     subject { described_class.persist_nested_collection_for(parent: parent, child: child) }
 
     it 'creates the relationship between parent and child' do
       subject
-      expect(parent.member_objects).to eq([child])
-      expect(child.member_of_collections).to eq([parent])
+      expect(Hyrax.custom_queries.find_parent_collection_ids(resource: child)).to eq [parent.id]
```
no-reply, line 11:
```suggestion
      expect(Hyrax.custom_queries.find_parent_collection_ids(resource: child)).to contain_exactly(parent.id)
```

no-reply:
```suggestion
      expect(Hyrax.custom_queries.find_parent_collection_ids(resource: child)).to contain_exactly(parent.id)
```

no-reply (changes requested):
i think the hard test requirements on `eq []` will come back to bite us (many of our valkyrie query implementations are optimized to be more lazy than this, and return other `Enumerable` types.

i've added some suggestions. otherwise, this looks great to me.

no-reply (approved):

PR #5077, opened 2021-08-17
provide NullUser for use when real user is not available

elrayle:
Primary usage is to provide backward compatibility for methods adding a `user:` parameter.  Whenever possible, a real user should be passed to methods with `user: NullUser.new` parameter.

Updates usage of `user: nil` added in PR #5076 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/null_user.rb
+++ b/app/models/hyrax/null_user.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+module Hyrax
+  #
+  # This class supports passing a user or providing a default user when the currently
+  # logged in user is not available.  When adding the `user:` parameter to an existing
+  # method, the `Hyrax::NullUser` can be used as a default value to maintain backward
+  # compatibility. Whenever possible, calling methods should be updated to pass the
+  # real user through from a place where the user was available.
+  #
+  # Example:
+  # `Hyrax::Collections::NestedCollectionPersistenceService` methods use `Hyrax::NullUser.new`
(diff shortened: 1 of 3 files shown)


app/models/hyrax/null_user.rb
```diff
@@ -0,0 +1,30 @@
...
+  # method passes the current user to the service's `#persist_nested_collection_for` method.
+  # All callers of all methods in this service in Hyrax code pass in the real user.  Use
+  # of the `Hyrax::NullUser.new` provides backward compatibility for customizations.
+  #
+  class NullUser < ::User
+    ##
+    # @return [nil]
+    def id
+      '_NULL_USER_ID_'
+    end
```
elrayle, line 22: Any concerns having a fake id?  The tests do not pass if `#id` returns `nil`.

ERROR:
`Unable to serialize Hyrax::NullUser without an id.` raised in `ContentUpdateEventJob.perform_later` when running the specs for `NestedCollectionPersistenceService`.

no-reply: `link_to_profile depositor` the eventual caller? i think including a broken id is going to (best case) cause a broken link. worse, i think with a real adapter ActiveJob is going to serialize/deserialize a GlobalID, which will fail on the deserialization, preventing the job from running.

i think we might need to consider passing a real user. there are various system users by default. we could use one of those or make a new one.

elrayle: Do you have a suggestion for the 'real' user to pass?  This impacts PR #5083 as well.  I temporarily used `::User.audit_user` for that PR.  I am happy to drop the `NullUser` if there is a viable alternative.

PR #5078, opened 2021-08-17
use CollectionMemberService instead of member_of_collections relationship

elrayle:
Update collection and work forms to use Valkyrie approach to managing relationships instead of AF approach.

Partially addresses #5071

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/collection_form.rb
+++ b/app/forms/hyrax/forms/collection_form.rb
@@ -116,10 +116,14 @@ module Hyrax
       end
 
       def list_parent_collections
+        Deprecation.warn("'##{__method__}' will be removed in Hyrax 4.0.  " \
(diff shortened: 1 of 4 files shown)


no-reply:
can you comment on the need for changes to `WorkForm`?

all else being equal, i think it would be okay to let `WorkForm` continue to use ActiveFedora internally, since it won't be in use in a Valkyrie native app. but i'm wondering if there was something driving your changes here.

elrayle:
What is the replacement for WorkForm?  Is there an equivalent replacement for CollectionForm?

no-reply:
`ResourceForm` is what's used by `WorksControllerBehavior`. there's not a Collection equivalent yet.

i recall you trying something out with ResourceForm a few months back and deciding it was too `Work` specific (which rings true to me).

Issue #4530, opened 2020-10-01
Analytics Graph is not responsive on mobile devices


njaffer:
### Descriptive summary
Analytics Graph is not responsive on mobile devices. Additionally, it is not accessible by screen readers.

### Rationale
When testing Hyrax, nurax on a mobile device, the analytics page has two graphs that do not scale according to the screen size.

### Expected behavior
The graphs should be responsive to the screen size

### Actual behavior
The graphs extends the screen size on a mobile device

### Steps to reproduce the behavior

1. View a work
2. Click on Analytics button

no-reply:
> Additionally, it is not accessible by screen readers.

can we break this out into a separate ticket and tag `accessibility-concern`? i think this should be described and prioritized separately from the responsiveness issue.

njaffer:
yeah sure, I will create another ticket and tag accessibility-concern

Issue #3551, opened 2019-02-21
Work stats page discrepancy in the dates on mouseover for pageviews chart


tahirpoduska:
### Descriptive summary
When you mouse over the dates on the work stats or download stats page page graph, the date displayed is a day older than the date on the x-axis.

### Rationale
If a user is collecting data from the chart and using the mouse over to get the page views, the information provided by the mouse over is wrong since it associates the previous day with the page views.

### Expected behavior
The date on mouse over should be the same as date on the x-axis

### Actual behavior
The date on mouse over is a day older than the date on the x-axis

### Steps to reproduce the behavior

1. Go to nurax work stats page: https://nurax-dev.curationexperts.com/works/qr46r0831/stats?locale=en. Or any other work stats page. Zoom in so you can see the dates on x-axis, if the date range is too long.
For download stats, go to https://nurax-dev.curationexperts.com/files/rn301153s/stats?locale=en

2. Then mouse over on a data point (page view). Check the x-axis date and the date on the mouse-over display, they are different but should be the same.

no-reply:
I can't reproduce on Firefox 68.0.1.

Anyone have ideas about this one?

no-reply:
![2019-09-20-201257_3840x2160_scrot](https://user-images.githubusercontent.com/1361776/65367250-57e20480-dbe3-11e9-8d5b-227bb28b5acf.png)

rjkati:
I've observed this on https://nurax-dev.curationexperts.com/works/sb397849b/stats?locale=en in Firefox. The date on the x-axis is "March 23, 2021"
![analytics](https://user-images.githubusercontent.com/32644254/112334912-2d6a7f00-8c92-11eb-87a9-a95cd46e417f.PNG)

However on mouseover, "March 22, 2021" is displayed:
![analytics2](https://user-images.githubusercontent.com/32644254/112334986-3bb89b00-8c92-11eb-9bc2-6973b38e2c4a.PNG)

jlhardes:
Analytics work from Oregon Digital grant will be coming in before end of August. Not likely to address this particular issue but would be good to wait on this issue until after that code is merged.

jlhardes:
Analytics updates changed Work stats view to bar chart format and this date discrepancy no longer exists.

Issue #2566, opened 2018-01-24
Single Work Analytics Page


nestorw:
### Descriptive summary
Analytics for a single work

<img width="1322" alt="c7-analytics-work" src="https://user-images.githubusercontent.com/1766876/35357540-dcbf7884-0121-11e8-95c2-09231d5cea21.png">

vantuyls:
Needs:
- [ ]  metadata summary (see design mockup)
- [ ] button to View Work (takes one to the work showpage)
- [ ] total file size table
- [ ] activity summary with table
- [ ] monthly activity with table
- [ ] download buttons for the above tables

vantuyls:
link at the top right of the screen "return to work reports" does not need to be surfaced until work reports are surfaced (likely the next milestone).

jlhardes:
Testing on Nurax-dev with Analytics update and no analytics service connected shows the Works Report for a single work has all of these updates. There's no download option in a single work report but there is an export option for the Top Works table in the Works Report for all works. That includes both page views and downloads info (and actually seems to include all Works so there is no cut-off for Top Works). The metadata showing per work in the single work report seems to somewhat match the mock-up although I haven't seen any "Approved by", "Published on", or "Last modified on" info for any single work reports I viewed (Nurax-dev data can be light on activities like those but every single work report is showing deposited date and user info). The requests in this issue seem satisfied to me. @nestorw it has been a while, but do you have any concerns or questions?

jlhardes:
Correction: Nurax-dev is connected to Google Analytics and that is why the single work analytics views and reports are available. When there is no analytics service connected there are no analytics graphs available from the Work view and the Analytics reports on the Dashboard (for Works and Collections) are also not available.

jlhardes:
Closing issue as checklist is resolved.

Issue #2542, opened 2018-01-22
Clarity on what a pageview is and what a download is and how they are generated


nestorw:
### Descriptive summary

Needs Clarification. We should probably look at Google analytics' and Piwik's documentation to see what they consider a page view. 

https://wiki.duraspace.org/pages/viewpage.action?pageId=87467336

mcritchlow:
@vantuyls - I have a question related to this ticket as I'm working on trying to setup queries for GA to pull down (and filter) pageview data. What do we want counting as a pageview?

Just the show/public URL, or also the edit/update/etc. routes?

Example:
http://localhost:3000/concern/images/ -- show/public/read-only view
vs
http://localhost:3000/concern/images/edit -- edit view

vantuyls:
chatted about this. we only want to count views on the landing page for a work or fileset. We do not want to count things like edit views.

jlhardes:
Analytics update has these defined but if there isn't an analytics service in place, is there anything that shows for Work stats regarding Page Views and Downloads? Trying views and downloads on Nurax-dev is not changing the Analytics graphs for any Works so far but Nurax-dev also doesn't have any analytics service connected at this point. This might require more testing via a Hyrax instance with a connected analytics service (Google Analytics or Matomo) to know for sure it is resolved. Does that seem right to you, @orangewolf, @sara-g or others involved in that Analytics update work?

jlhardes:
An update to the update: Nurax-dev does have an analytics service connected - it is using Google Analytics. If there is no analytics service connected, there are no analytics graphs available from the Work view and the Analytics reports on the Dashboard are also not available. This issue is still addressed and can stay closed.

PR #5089, opened 2021-09-01
Analytics

### LINKED ISSUE #4564: thinking about analytics
### Descriptive summary

consider the analytics dashboard. count. count the things. [rattle, rattle]

### Rationale

I wear the chain I forged in life,” replied the Ghost. “I made it link by link, and yard by yard; I girded it on of my own free will, and of my own free will I wore it. Is its pattern strange to you?”

### Expected behavior

comprehensive analytics. [rattle]

### Actual behavior

empty. so...cold...[rattle, rattle]

orangewolf:
Fixes #4564 #4530 #3551 #2542 #2566 Maybe others?

This work was sponsored by Oregon Digital as part of the IMLS National Leadership Grant to support its project, Open Impact: Developing Robust Analytics for Open Source Solution Bundle Hyrax.

This code overhauls the built in display of reports and metrics for Hyrax. It makes it compatible with both GA v3 and Matomo and lays ground work for other adapters in the future. 

* Turn analytics on/off, choose your analytics provider of choice (google or matomo), and set api details in an ENV file
* Wire up graphs on dashboard to work based off analytics data (number of total visitors, returning visitors, new user signups)
* New tab in dashboard sidebar for "analytics reports".  Dropdown includes options for "Work Reports" or "Collection Reports"
* Design and develop Work and Collection reports to show views and download counts for individual files. Data is set by default to last 30 days, by month by month, as well as custom date range.  This data is pulled in directly from Google or Matomo API's. 
* View and Download counts are available to be exported to .csv format


.dassie/config/environments/development.rb
```diff
@@ -29,6 +29,8 @@
     config.cache_store = :null_store
   end
 
+  config.web_console.whitelisted_ips = ['172.18.0.0/16', '172.27.0.0/16', '0.0.0.0/0']
+
```
orangewolf, line 33: this line makes the webconsole work in docker

app/views/hyrax/stats/_downloads.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  Morris.Bar({
+      element: 'downloads',
+      data: [
+        <% @downloads.list.take(30).reverse.each do |result| %>
+          { date: "<%= result[0] %>", metric: <%= result[1] %>},
+        <% end %>
+      ],  
+      xkey: 'date',
+      ykeys: ['metric'],
+      labels: ['Downloads'],
```
orangewolf, line 11: since we are in erb anyway, should we translate here?

app/views/hyrax/stats/_pageviews.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  Morris.Bar({
+      element: 'pageviews',
+      data: [
+        <% @pageviews.list.take(30).reverse.each do |result| %>
+          { date: "<%= result[0] %>", metric: <%= result[1] %>},
+        <% end %>
+      ],  
+      xkey: 'date',
+      ykeys: ['metric'],
+      labels: ['Page Views'],
```
orangewolf, line 11: translations

app/views/hyrax/stats/work.html.erb
```diff
@@ -1,26 +1,34 @@
...
+      <h3 class="text-center">Pageviews</h3>
+      <div id="pageviews">
+        <%= render 'pageviews' %>
+      </div>
     </div>
-    <div class="stats-container" style="height:150px;">
-      <div id="overview" class="stats-placeholder"></div>
+
+    <div class="stats-container">
+      <h3 class="text-center">Downloads</h3>
```
orangewolf, line 27: translations

bin/decrypt-secrets
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env ruby
```
orangewolf, line 1: n8 structure,  remove from PR

bin/encrypt-secrets
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env ruby
```
orangewolf, line 1: n8 structure, remove from PR

bin/helm_delete
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
orangewolf, line 1: n8 structure, remove from PR

bin/helm_deploy
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
```
orangewolf, line 1: n8 structure, remove from PR

config/locales/hyrax.en.yml
```diff
@@ -274,7 +274,8 @@ en:
         profile: Profile
```
orangewolf, line 274: once we are done adding translations to the en file - we should generate the other files

orangewolf: there is an automated generation process for them.

docker-compose.yml
```diff
@@ -8,11 +8,14 @@ services:
       args:
         - EXTRA_APK_PACKAGES=git less
     image: ghcr.io/samvera/dassie
+    # TODO(alishaevn): remove this command when the bug that causes errors while building is fixed
```
orangewolf, line 11: n8 structure, remove from PR

docker-compose.yml
```diff
@@ -8,11 +8,14 @@ services:
...
+    # command: sh -l -c "bundle && bundle exec puma -v -b tcp://0.0.0.0:3000"
     stdin_open: true
     tty: true
     user: root
     env_file:
       - .env
+      - .env.development
```
orangewolf, line 18: n8 structure, remove from PR

docker-compose.yml
```diff
@@ -109,6 +112,8 @@ services:
       context: .
       target: hyrax-engine-dev-worker
     image: ghcr.io/samvera/dassie-worker
+    # TODO(alishaevn): remove this command when the bug that causes errors while building is fixed
```
orangewolf, line 115: n8 structure, remove from PR

lib/generators/hyrax/templates/config/analytics.yml
```diff
@@ -2,8 +2,11 @@
...
-#   privkey_secret: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_SECRET']
-#   client_email: <%= ENV['GOOGLE_OAUTH_CLIENT_EMAIL']
+#   google:
+#     app_name: <%= ENV['GOOGLE_OAUTH_APP_NAME']
+#     app_version: <%= ENV['GOOGLE_OAUTH_APP_VERSION']
+#     privkey_path: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_PATH']
+#     privkey_secret: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_SECRET']
+#     client_email: <%= ENV['GOOGLE_OAUTH_CLIENT_EMAIL']
+#   matomo:
+#     key:
```
orangewolf, line 12: this needs to be finished

lib/generators/hyrax/templates/config/initializers/hyrax.rb
```diff
@@ -43,15 +43,23 @@
...
   # config.analytics = false
 
-  # Google Analytics tracking ID to gather usage statistics
-  # config.google_analytics_id = 'UA-99999999-1'
+  # Analytics provider type - select one of google or matomo
+  # config.analytics_provider = 'google'
+  # config.analytics_provider = 'matomo'
 
-  # Date you wish to start collecting Google Analytic statistics for
+  # Analytics tracking ID to gather usage statistics
```
orangewolf, line 54: I don't think these changes are correct any more?

lib/hyrax/configuration.rb
```diff
@@ -112,11 +112,19 @@ def analytics?
         ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_ANALYTICS', false))
     end
 
-    attr_writer :google_analytics_id
```
orangewolf, line 115: I believe we need to deprecate these and not just remove them.

ops/ga-deploy.tmpl.yaml
```diff
@@ -0,0 +1,102 @@
+imagePullSecrets:
```
orangewolf, line 1: n8 structure, remove from PR

ops/matomo-deploy.tmpl.yaml
```diff
@@ -0,0 +1,80 @@
+imagePullSecrets:
+  - name: gitlab-registry
```
orangewolf, line 2: n8 structure, remove from PR

spec/lib/hyrax/configuration_spec.rb
```diff
@@ -16,55 +16,56 @@
...
+  it { is_expected.to respond_to(:active_deposit_agreement_acceptance?) }
   it { is_expected.to respond_to(:activity_to_show_default_seconds_since_now) }
   it { is_expected.to respond_to(:admin_set_predicate) }
   it { is_expected.to respond_to(:admin_set_predicate=) }
   it { is_expected.to respond_to(:analytic_start_date) }
   it { is_expected.to respond_to(:analytics?) }
-  it { is_expected.to respond_to(:analytics) }
-  it { is_expected.to respond_to(:arkivo_api?) }
+  it { is_expected.to respond_to(:analytics_provider) }
+  it { is_expected.to respond_to(:analytics_provider=) }
```
orangewolf, line 27: this is the important part.  The rest of the changes in this file are just 1 man's love of the alphabetical order.

spec/presenters/hyrax/admin/user_activity_presenter_spec.rb
```diff
@@ -1,21 +1,23 @@
 # frozen_string_literal: true
 RSpec.describe Hyrax::Admin::UserActivityPresenter do
-  let(:instance) { described_class.new }
+  start_date = Time.zone.today - 6.days
```
orangewolf, line 3: these need to be in lets or they will break in _very_ subtle ways. For example here - start_date is set when the rspec command is first run whether or not it ever runs this spec file.

spec/services/hyrax/statistics/works/over_time_spec.rb
```diff
@@ -1,15 +1,22 @@
...
-  describe "#points", :clean_repo do
+  describe '#points', :clean_repo do
     before do
       create(:generic_work)
     end
 
     subject { service.points }
 
-    it "is a list of points" do
+    xit 'is a list of points' do
```
orangewolf, line 19: ?

.dassie/app/models/collection.rb
```diff
@@ -4,5 +4,5 @@ class Collection < ActiveFedora::Base
   include ::Hyrax::CollectionBehavior
   # You can replace these metadata if they're not suitable
   include Hyrax::BasicMetadata
-  self.indexer = Hyrax::CollectionWithBasicMetadataIndexer
+  self.indexer = Hyrax::CollectionWithBasicMetadataIndexer
```
elrayle, line 7: I wonder if you ran rubocop.  I would have expected this change to fail with the space at the end of the line.  Maybe rubocop doesn't run over the .dassie directory.

orangewolf: @elrayle rubocop is set up in main to specifically not run on .dassie

app/views/layouts/_head_tag_content.html.erb
```diff
@@ -25,8 +25,13 @@ signed in %>
 
 <%= render 'shared/appearance_styles' %>
 
-<!-- Google Analytics -->
-<%= render partial: '/ga', formats: [:html] %>
+<% unless Hyrax.config.analytics == 'false' %>
+  <% if Hyrax.config.analytics_provider == 'google' %>
```
no-reply, line 29: it might be nice to helperize this:

```ruby
def hyrax_analytics_partial(provider: Hyrax.config.analytics_provider)
  case provider
  when 'google'
    'shared/ga'
  when 'matomo'
    'shared/matomo'
  end
end
```

then the view code becomes

```ruby
<% if Hyrax.config.analytics %>
  <%= render partial: hyrax_analytics_partial, formats: [:html] %>
<% end %>
```

app/views/layouts/_head_tag_content.html.erb
```diff
@@ -25,8 +25,13 @@ signed in %>
 
 <%= render 'shared/appearance_styles' %>
 
-<!-- Google Analytics -->
-<%= render partial: '/ga', formats: [:html] %>
+<% unless Hyrax.config.analytics == 'false' %>
```
no-reply, line 28: i think this `unless` isn't functional. `Hyrax.config.analytics` is casting to a boolean, so it should never `== 'false'`? am i missing anything here?

.dassie/config/initializers/hyrax.rb
```diff
@@ -5,6 +5,8 @@
   # Defaults to false.  See README for more info
   config.arkivo_api = true
 
+  config.analytics = ENV.fetch('HYRAX_ANALYTICS', 'false')
```
no-reply, line 8: i think this just removes the casting behavior from the default `Hyrax::Configuration` tooling.

```suggestion
```

app/controllers/hyrax/admin/analytics/collection_reports_controller.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+module Hyrax
+  module Admin
+    module Analytics
+      class CollectionReportsController < AnalyticsController
+        include Hyrax::BreadcrumbsForCollectionAnalytics
+        def index
+          return unless Hyrax.config.analytics == 'true'
```
no-reply, line 8:
```suggestion
          return unless Hyrax.config.analytics?
```

app/controllers/hyrax/admin/analytics/collection_reports_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+          @top_collection_pages = Hyrax::Analytics.top_events('collection-page-view', date_range)
+          respond_to do |format|
+            format.html
+            format.csv { export_data }
+          end
+        end
+
+        def show
+          @document = ::SolrDocument.find(params[:id])
+          return unless Hyrax.config.analytics == 'true'
```
no-reply, line 25:
```suggestion
          return unless Hyrax.config.analytics?
```

app/controllers/hyrax/admin/analytics/collection_reports_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+          @top_downloads = Hyrax::Analytics.top_events('work-in-collection-download', date_range)
+          @top_collection_pages = Hyrax::Analytics.top_events('collection-page-view', date_range)
+          respond_to do |format|
+            format.html
+            format.csv { export_data }
+          end
+        end
+
+        def show
+          @document = ::SolrDocument.find(params[:id])
```
no-reply, line 24: do we want to populate `@document` even if analytics is turned off?

app/controllers/hyrax/admin/analytics/work_reports_controller.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
+module Hyrax
+  module Admin
+    module Analytics
+      class WorkReportsController < AnalyticsController
+        include Hyrax::BreadcrumbsForWorksAnalytics
+
+        def index
+          return unless Hyrax.config.analytics == 'true'
```
no-reply, line 9:
```suggestion
          return unless Hyrax.config.analytics?
```

app/views/hyrax/admin/analytics/collection_reports/index.html.erb
```diff
@@ -0,0 +1,70 @@
...
+
+<% provide :page_header do %>
+  <h1><span class="fa fa-bar-chart" aria-hidden="true"></span> <%= t('.title') %></h1>
+<% end %>
+
+<div class="row">
+  <div class="col-sm-12">
+    <div class="collection-reports">
+
+<% if Hyrax.config.analytics == 'true' %>
```
no-reply, line 11:
```suggestion
<% if Hyrax.config.analytics? %>
```

app/views/hyrax/admin/analytics/work_reports/index.html.erb
```diff
@@ -0,0 +1,77 @@
...
+<% provide :page_header do %>
+  <h1><span class="fa fa-bar-chart" aria-hidden="true"></span><%= t('.title') %></h1>
+<% end %>
+
+<div class="row">
+  <div class="col-sm-12">
+    <div class="work-reports">
+
+    <div class="panel panel-default">
+    <% if Hyrax.config.analytics == 'true' %>
```
no-reply, line 12:
```suggestion
    <% if Hyrax.config.analytics? %>
```

app/views/hyrax/dashboard/_user_activity.html.erb
```diff
@@ -1,28 +1,22 @@
...
-    <tbody>
-      <% JSON.parse(user_activity).each do |row| %>
-        <tr>
-          <td><%= row["y"] %></td>
-          <td><%= row["a"] %></td>
-        </tr>
-      <% end %>
-    </tbody>
-  </table>
+  <% if Hyrax.config.analytics == 'true' %>
```
no-reply, line 5:
```suggestion
  <% if Hyrax.config.analytics? %>
```

.dassie/config/environments/development.rb
no-reply: let's get this in now as a separate PR?

.dassie/config/routes.rb
```diff
@@ -31,5 +31,6 @@
       delete 'clear'
     end
   end
+
```
no-reply, line 34:
```suggestion
```

.dassie/config/role_map.yml
```diff
@@ -24,3 +24,4 @@ test:
 production:
   admin:
     - admin@example.com
+    - another_admin@example.com
```
no-reply, line 27: this and the related seed change could go in separately and right away, if it helps.

app/controllers/hyrax/user_pinned_collections_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+module Hyrax
+  class UserPinnedCollectionsController < ApplicationController
```
no-reply, line 3: how does the `UserPinnedCollections` feature relate to the analytics one?

it seems nice, but just seems like a very different feature set.

Dockerfile
```diff
@@ -103,3 +103,4 @@ COPY --chown=1001:101 $APP_PATH /app/samvera/hyrax-webapp
 COPY --chown=1001:101 . /app/samvera/hyrax-engine
 
 RUN bundle install --jobs "$(nproc)"
+RUN RAILS_ENV=production SECRET_KEY_BASE='fakesecret1234' DB_ADAPTER=nulldb DATABASE_URL='postgresql://fake' bundle exec rake assets:precompile
```
no-reply, line 106: is this needed? it looks like it duplicates asset precompilation in the worker, which should have just run due to the `ONBUILD` at https://github.com/samvera/hyrax/pull/5089/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R76.

app/views/hyrax/dashboard/_collections.html.erb
```diff
@@ -0,0 +1,46 @@
...
+                <th><%= t('.works') %></th>
+                <th><%= t('.files') %></th>
+                <th></th>
+              </tr>
+            </thead>
+            <tbody>
+             <% @collections.each do |collection| %>
+                <tr>
+                  <td><%= collection %></td>
+                  <td><%= collection.modified_date.strftime("%Y-%m-%d") %></td>
```
no-reply, line 26: i'm worried about these additions, which look like they are computationally complex and likely to be too slow to be tolerable for view code.

especially the implementation of `collection.total_file_sets` (i'll comment further there). if this isn't closely related to the analytics work, i'd suggest we consider it separately so it doesn't hold up other features here.

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -85,6 +86,16 @@ def to_s
       title.present? ? title.join(' | ') : 'No Title'
     end
 
+    def total_file_sets
```
no-reply, line 89: i'm worried about this implementation on two fronts:

first, i think it's likely to be intolerably slow for collections of any size. it looks to me like it pulls each member work from Fedora, and then pulls each file set from each of those. given the relatively slow turn around on Fedora requests, this could get pathological really quickly! there are definitely optimizations to be had (we can easily count the file_sets in each work from the ids, without retrieving the file set metadata); but i'm still not sure it's a good solution to try to compute this for each request.

second, i don't think the existing implementation will succeed at counting the file sets. it looks like it will miss any file sets that are members transitively (i.e. because they in works that are members of a subcollection of this collection, or because they are in works that are in works that are in this collection).

i think it would be best to discuss this separately from the analytics features bundled here.

app/views/hyrax/dashboard/_collections.html.erb
```diff
@@ -0,0 +1,46 @@
...
+                  <td><%= collection.member_works.count %></td>
+                  <td><%= collection.total_file_sets %></td>
+                  <td>
+                    <% pinned = collection.pinned?(current_user.id) %>
```
no-reply, line 30: i'd prefer this as a helper instead of a collection method.

we're trying in general to avoid introducing new dependencies on model layer specifics to the views.

```ruby
def pinned_collection?(collection:, user:)
  UserPinnedCollection.where(user_id: user_id, collection_id: id).exists?
end
```

otherwise i'd be okay making this a class method on `UserPinnedCollection`, like `UserPinnedCollection.pinned?(collection_id:, user_id:)` ?

app/models/concerns/hyrax/collection_behavior.rb
```diff
@@ -85,6 +86,16 @@ def to_s
...
+      files = []
+      member_works.map { |work| files.push(work.file_sets.count) }
+      files.sum
+    end
+
+    def pinned?(user_id)
+      UserPinnedCollection.find_by(user_id: user_id, collection_id: id)
```
no-reply, line 96: if we keep this here use `exists?` instead of `find_by` to avoid pulling data from the DB unnecessarily.

```suggestion
      UserPinnedCollection.exists?(user_id: user_id, collection_id: id)
```

app/presenters/hyrax/admin/user_activity_presenter.rb
```diff
@@ -2,28 +2,24 @@
 module Hyrax
   module Admin
     class UserActivityPresenter
-      def initialize
-        @x_min = 90.days.ago.beginning_of_day
-        @date_format = ->(x) { x.strftime('%b %-d') }
+      def initialize(start_date, end_date)
```
no-reply, line 5: this method signature change is breaking, but could be made compatible fairly easily by providing default parameters.

```suggestion
      def initialize(start_date: 90.days.ago.beginning_of_day, end_date: Date.today.end_of_day)
```

i think i prefer the named parameters over order dependent parameters in general, but the following is fine too:

```ruby
      def initialize(start_date = 90.days.ago.beginning_of_day, end_date = Date.today.end_of_day)
````

File: app/presenters/hyrax/work_usage.rb [hunk 41]
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module Hyrax
+  # TODO: - Analytics do we still need this?
```
no-reply, line 4: do we? 🍡

app/search_builders/hyrax/README.md
```diff
@@ -66,4 +66,4 @@ module Hyrax
 end
 ```
 
-There is no point having the other `filter_models` methods apply `:fq`s that we then try to undo or overwrite.  In general, directly overwriting the whole `default_processor_chain` or solr parameters like `:fq` is less flexible than appending constraints sufficient for your use case.  In particular, you might find that you have overwritten components that implement access controls, thereby making your SearchBuilder less useful and less secure.  When in doubt, examine the actual solr queries produced.
+There is no point having the other `filter_models` methods apply `:fq`s that we then try to undo or overwrite.  In general, directly overwriting the whole `default_processor_chain` or solr parameters like `:fq` is less flexible than appending constraints sufficient for your use case.  In particular, you might find that you have overwritten components that implement access controls, thereby making your SearchBuilder less useful and less secure.  When in doubt, examine the actual solr queries produced.
```
no-reply, line 69:
```suggestion
There is no point having the other `filter_models` methods apply `:fq`s that we then try to undo or overwrite.  In general, directly overwriting the whole `default_processor_chain` or solr parameters like `:fq` is less flexible than appending constraints sufficient for your use case.  In particular, you might find that you have overwritten components that implement access controls, thereby making your SearchBuilder less useful and less secure.  When in doubt, examine the actual solr queries produced.  
```
```suggestion
There is no point having the other `filter_models` methods apply `:fq`s that we then try to undo or overwrite.  In general, directly overwriting the whole `default_processor_chain` or solr parameters like `:fq` is less flexible than appending constraints sufficient for your use case.  In particular, you might find that you have overwritten components that implement access controls, thereby making your SearchBuilder less useful and less secure.  When in doubt, examine the actual solr queries produced.
```

no-reply:
really excited to see this one!

i provided a partial review, really with just the things that i spotted in some spare time. i have a few structural questions about the scope of the PR. and a few substantive suggestions. i suspect review on this one might take a village, but i'd also be keen to start picking off parts that are ready to merge ASAP, if that sounds attractive.

.dassie/config/initializers/hyrax.rb
sara-g: If this is removed then Hyrax.config.analytics just returns nil.

app/presenters/hyrax/admin/user_activity_presenter.rb
sara-g: what does it break?   If I change to the suggested the dashboard does not load.

app/presenters/hyrax/work_usage.rb
sara-g: There is some existing code that I wasn't sure what to do with.

app/assets/javascripts/hyrax/reports-buttons.js
```diff
@@ -0,0 +1,33 @@
+Blacklight.onLoad(function() {
+  var summaryButton = $('#summary-btn');
+  var monthlyButton = $('#monthly-btn');
+  var rangeButton = $('#range-btn');
+
+  $('#monthly-btn').click(function() {
```
lfarrell, line 6: I'd recommend switching this to `$('#monthly-btn').on('click', function() {` so any dynamically added elements get the event listener too.

.dassie/config/role_map.yml
sara-g: i took this out.

app/controllers/hyrax/admin/analytics/collection_reports_controller.rb
sara-g: probably not.  removed!

app/controllers/hyrax/user_pinned_collections_controller.rb
sara-g: Agreed. I removed this feature and I will put this in a separate PR.

app/views/hyrax/dashboard/_collections.html.erb
sara-g: Removed for now (part of pinned collections feature).

app/models/concerns/hyrax/collection_behavior.rb
sara-g: Removed for now (part of pinned collections feature).

app/views/hyrax/dashboard/_collections.html.erb
sara-g: Removed for now (part of pinned collections feature).

app/assets/javascripts/hyrax/reports-buttons.js
sara-g: updated.  thank you!

jlhardes:
I have reviewed how this is functioning on the two Notch8 demo sites with Google Analytics and Matomo providing data and I think the visualization of this data with the adjustable date ranges is definitely an improvement over the analytics view provided right now out of the box in Hyrax. Both of those sites appear to be running Hyrax 3.0.2 so it is not the latest release (3.1.0) or code from the main branch. I haven't been able to test this running Hyrax locally since I don't have a GA account or Matomo account to connect for testing.

lib/hyrax/configuration.rb
```diff
@@ -109,19 +109,38 @@ def register_roles
...
-      @google_analytics_id ||= nil
+    # Currently supports 'google' or 'matomo'
+    # google is default for backward compatability
+    attr_writer :analytics_provider
+    def analytics_provider
+      @analytics_provider ||=
+        ENV.fetch('HYRAX_ANALYTICS_PROVIDER', 'google')
+    end
+
+    attr_writer :analytics_start_date
```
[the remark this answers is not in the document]
sara-g: added!  thank you!

app/presenters/hyrax/admin/user_activity_presenter.rb
no-reply: @sara-g it breaks the method signature for `Hyrax::Admin::UserActivityPresenter.new`. any client code calling `UserActivityPresenter.new` without arguments will get an error about the now required arguments.

we generally try to avoid breaking method signatures like this in Hyrax, because our classes and methods can be used downstream in ways that aren't predictable to us. when we do make changes like this, we bump the major version; so this change would put us into 4.0.0 territory. since i didn't see many other changes that would require that, it seemed worth considering reinstating compatibility.

>  If I change to the suggested the dashboard does not load.

is this just because all the existing callers to `.new` have changed in this PR? i think you'd need to either use the unnamed parameters in the second suggestion, or update the callers to be inline with the proposed named parameter changes.

or is there something bigger going on?

app/presenters/hyrax/work_usage.rb
no-reply: maybe we could open a ticket to do whatever work this is pointing to? this kind of comment is likely to get ignored indefinitely, not least because it doesn't provide whatever context the author had at time of writing.

app/models/concerns/hyrax/collection_behavior.rb
no-reply: i think this might still need to be extracted?

.dassie/config/initializers/hyrax.rb
no-reply: if `Hyrax.config.analytics` returns `nil` then `Hyrax.config.analytics?` (which is the existing preferred usage) should return `false`.

the casting behavior should be something we do once, so we can avoid the string equality checks throughout.

.rubocop.yml
```diff
@@ -14,6 +14,7 @@ AllCops:
     - 'vendor/**/*'
     - 'lib/hyrax/specs/**/*'
 
+
```
no-reply, line 17:
```suggestion
```

Dockerfile
```diff
@@ -102,4 +102,4 @@ ENV HYRAX_ENGINE_PATH /app/samvera/hyrax-engine
 COPY --chown=1001:101 $APP_PATH /app/samvera/hyrax-webapp
 COPY --chown=1001:101 . /app/samvera/hyrax-engine
 
-RUN bundle install --jobs "$(nproc)"
+RUN bundle install --jobs "$(nproc)"
```
no-reply, line 105:
```suggestion
RUN bundle install --jobs "$(nproc)"

```

README.md
```diff
@@ -77,6 +77,43 @@ This is where you work on the code-base that will be used by yours and other Hyr
     <p>By moving to Docker, we are encoding the documentation steps for standing up a Hyrax-engine development environment.</p>
 </aside>
 
+### Installing Analytics
+
+Hyrax supports your choice of either Google Analytics or Matomo.  To enable analytics tracking and reporting features, follow the directions below.
+
+Enable Analytics Features
+
+In your .env file, set HYRAX_ANALYTICS to true, set either 'google' or 'matomo' for  HYRAX_ANALYTICS_PROVIDER, and set the date you would like reporting to start (ANALYTICS_START_DATE).
```
no-reply, line 86: use of `.env` isn't prescribed by Hyrax; reference to it may confuse folks deploying into environments where it's not used. it's probably better to say something like "analytics can be configured using environment variables..." instead.

lib/hyrax/configuration.rb
```diff
@@ -109,19 +109,41 @@ def register_roles
...
-      @google_analytics_id ||= nil
+    # Currently supports 'google' or 'matomo'
+    # google is default for backward compatability
+    attr_writer :analytics_provider
+    def analytics_provider
+      @analytics_provider ||=
+        ENV.fetch('HYRAX_ANALYTICS_PROVIDER', 'google')
+    end
+
+    # Date you wish to start collecting analytics for
```
no-reply, line 123: this is pretty minor, but it would be better if these were formatted with [yard](https://rubydoc.info/gems/yard/file/docs/GettingStarted.md). that would make them readable with other documentation at https://www.rubydoc.info/gems/hyrax/Hyrax/Configuration

something like:

```suggestion
    ##
    # @!attribute [w] analytics_start_date
    #   @note this can be set using the +ANALITICS_START_DATE+ environment variable (format is YYYY-MM-DD)
    #   @return [String] date you wish to start collecting analytics for. used to compute the 
    #     "all-time" metrics.
```

no-reply: is it true that this is the date to start collecting analytics? or is it more like a filter?

lib/hyrax/configuration.rb
```diff
@@ -109,19 +109,41 @@ def register_roles
...
+    # This is used to compute the "all-time" metrics
+    # Set this in your .env file (format is YYYY-MM-DD)
```
no-reply, line 125: i'd drop the `.env` reference here, too.

Dockerfile
no-reply: this suggestion might be a little wonky. what i'm getting at is that there aren't any real changes left in this file, so we should be able to drop it from the diff.

lib/hyrax/configuration.rb
```diff
@@ -109,19 +109,41 @@ def register_roles
...
+    attr_writer :analytics_start_date
+    def analytics_start_date
+      @analytics_start_date ||=
+        ENV.fetch('ANALYTICS_START_DATE', Time.zone.today - 1.year)
```
no-reply, line 129: this looks like it will return a `Date` in the fallback case, but a `String`(?) when it's configured via ENV. is that okay?

app/controllers/hyrax/admin/analytics/analytics_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+          @start_date = params[:start_date] || Hyrax.config.analytics_start_date
+          @end_date = params[:end_date] || Time.zone.today + 1.day
+        end
+
+        def date_range
+          "#{@start_date},#{@end_date}"
+        end
+
+        def paginate(results_array, rows: 10)
+          return if results_array.nil?
```
no-reply, line 30: does `results_array` sometimes return `nil`? could we make it give an empty array in this case instead?

lib/hyrax/configuration.rb
sara-g: I guess its more like a filter.  Its the start_date part of the calculation for the all-time metrics.

lib/hyrax/configuration.rb
```diff
@@ -109,19 +109,41 @@ def register_roles
     attr_reader :analytics
     def analytics?
       @analytics ||=
-        ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_ANALYTICS', false))
+        ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_ANALYTICS', 'false'))
```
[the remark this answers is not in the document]
sara-g: I added this to be able to change to your suggested `Hyrax.config.analytics?  ` instead of `if Hyrax.config.analytics == 'true'`

app/controllers/hyrax/admin/analytics/collection_reports_controller.rb
no-reply: on a closer look, i'm wondering if this is a repeated query due to `AnalyticsController#set_document`? it looks like that is running as a `before_action` for `#show`?

lib/hyrax/configuration.rb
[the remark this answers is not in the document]
no-reply: what happens if this line is unchanged? it doesn't seem like this should change how this line evaluates:

```ruby
ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_ANALYTICS', 'false')) # => true
ActiveModel::Type::Boolean.new.cast(ENV.fetch('HYRAX_ANALYTICS', false)) # => true
ActiveModel::Type::Boolean.new.cast(ENV.fetch('MISSING_ENV_VAR', 'false')) # => false
ActiveModel::Type::Boolean.new.cast(ENV.fetch('MISSING_ENV_VAR', false)) # => false
```

lib/hyrax/configuration.rb
sara-g: yes

lib/hyrax/configuration.rb
```diff
@@ -109,19 +109,41 @@ def register_roles
...
+    # Set this in your .env file (format is YYYY-MM-DD)
+    attr_writer :analytics_start_date
+    def analytics_start_date
+      @analytics_start_date ||=
+        ENV.fetch('ANALYTICS_START_DATE', Time.zone.today - 1.year)
     end
-    alias google_analytics_id? google_analytics_id
 
     # Defaulting analytic start date to whenever the file was uploaded by leaving it blank
     attr_writer :analytic_start_date
```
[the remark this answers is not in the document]
sara-g: I didn't want to remove any existing code because I was afraid it would break for people with google analytics already installed and working in their applications.  I am not totally clear on what is needed for that to work, and what is not.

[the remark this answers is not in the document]
scherztc: We discussed keeping the existing statistics code side by side with the new analytics code because although the new analytics code has a way better dashboard experience and fixes the views and downloads on works and filetypes, it does not do anything for the user statistics found in the user profile and their is still some unique value in the statistics dashboard option.

Since they both essential run off of the same database entries (populated by provider) they shouldn't be in conflict with one another.  I am testing this theory this week, since we have statistics working here at UC.

So it is is the recommendation of the tech call attendees to merge in the analytics PR and create a separate issue for removing the statistics, updating the user profile stats, and consolidating the dashboard views.

scherztc (approved):
After reading through the code and discussing it with the Samvera Tech call, I think we can merge this Analytics PR without removing the Statistics code.  Although they both work from the same data, they are both doing different jobs.

PR #5090, opened 2021-09-02
Add :environment to rake tasks that are missing it

elrayle:
Rake tasks should include `:environment` in the line naming the task.  Otherwise, `Undefined constant` errors occur when referencing classes defined under `app`.

This was reported in Slack as a bug in rake task `collection_type_global_id.rake` in a release 3.0.2 install.  I checked all other tasks and found it missing in only one other.  Rake task `regenerate_derivatirves.rake` is also repaired in this PR.

@samvera/hyrax-code-reviewers


svanschuyver (approved):
Looks good. Can confirm that this fixes the issue.

no-reply (approved):

no-reply:
thank you!

PR #5092, opened 2021-09-04
bump alpine linux to 3.14 to fix imagemagick pyramidal tiffs

orangewolf:
Some tiffs thumbnails are broken in the version of ImageMagick that Alpine 3.12 ships with.  Updating here and picking up the latest Ruby patch version seems like a win.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,21 +189,21 @@ workflows:
           hyrax_valkyrie: "true"
           requires:
             - build
-  ruby2-7-2:
+  ruby2-7-4:
     jobs:
       - bundle:
-          ruby_version: "2.7.2"
+          ruby_version: "2.7.4"
           rails_version: "5.2.6"
           bundler_version: "2.1.4"
       - build:
-          ruby_version: "2.7.2"
+          ruby_version: "2.7.4"
           rails_version: "5.2.6"
           bundler_version: "2.1.4"
           requires:
             - bundle
       - test:
-          name: "ruby2-7-2"
-          ruby_version: "2.7.2"
+          name: "ruby2-7-4"
+          ruby_version: "2.7.4"
           bundler_version: "2.1.4"
           requires:
             - build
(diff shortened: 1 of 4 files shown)


no-reply:
last i checked, ruby support was stuck on `2.7.2`. can we bump CI here if we think Ruby 2.7.4 will work? https://github.com/samvera/hyrax/blob/main/.circleci/config.yml#L192

orangewolf:
@no-reply these seem like 2 odd failures to me https://app.circleci.com/pipelines/github/samvera/hyrax/5857/workflows/9bd6767a-7bf7-422b-a0fd-c2a67bc73cc1/jobs/44353/tests do you have any insites in to them?

no-reply:
@orangewolf: @dlpierce is going to take a peek.

dlpierce:
It looks like the actor is using the tempfile's filename instead of the original filename for the label. Still tracking down why.

dlpierce:
It appears the list of allowed characters for tempfiles [changed in ruby 2.7.3](https://github.com/ruby/ruby/commit/4b6fa03a72b23fa03a8f37cbfdcad3f0251832cf#diff-a37a7e2119bc66938a1069d15a7cd0be7ed991c6675779b4d15ba920012af29aL118-R118)

This change exposed that the failing specs were actually setup to cause the FileSetActor to use the tempfile basename for the label because the Tempfile object didn't respond to original_name. It had passed prior to 2.7.3 because the original basename is prefixed to the tempfile basename and only a small set of separator characters were filtered out. With the 2.7.3 change to only safe ascii characters being allowed, so the tempfile no longer contained the odd characters the spec is (mistakenly!) concerned with.

To resolve this I've added original_filename as a potential file label source right before it defaults to using basename. It seems unlikely this will affect file labels outside of tests, but it's not impossible if an application is using odd file objects.

no-reply:
thank you @dlpierce ! nice work!

orangewolf:
@dlim87 i actually think this may have been a production issue for NIMs (a Hyrax that is used by the Japanese Gov). they'll be excited to know its fixed.  Thank you for legging it down!

PR #5094, opened 2021-09-09
Backport "check FlipFlop for caching manifest configuration at runtime"

maxkadel:
See https://github.com/samvera/hyrax/commit/88aebb3321ac905d02fb838d191dcf733f2c67cb 

Backporting to v2.9.4 version for patch version release

Quote from that commit:
### check FlipFlop for caching manifest configuration at runtime
checking FlipFlop at app startup can be problematic if the database isn't
present. it also causes issues if this configuration gets flipped while the app
is running (the normal use case for FlipFlop).

probably this whole configuration should be deprecated in favor of an
environment flag.


mlooney (approved):

maxkadel:
Going to backport against 2.x branch instead

PR #5101, opened 2021-09-13
Backport to fix #2273 in 2.x stable

### LINKED ISSUE #2273: TinyMCE editor is not loading menu based on config yml

anusharanganathan:
refs https://github.com/samvera/hyrax/issues/2273

Updates tinymce-rails gem and fixes broken config loading.

This is a backport of PR https://github.com/samvera/hyrax/pull/4759 for Hyrax 2.x stable

This updates the TinyMCE gem that has been pinned at an older version for several years. It also was not loading the application's tinymce config file, instead using defaults. This PR addresses both of those issues.

* Remove version pin on tinymce-rails gem
* Modify how its JS is loaded according to the gem's install docs.
* Update the provided config file.

* The TinyMCE editor loads without error and functions when editing Pages and Content Blocks in the dashboard.

@samvera/hyrax-code-reviewers


no-reply:
this seems okay.

i was worried about backward compatibility, but it seems like there's no impact to existing apps? 

looks like we could release as 2.9.7(?)

cc: @dlpierce

.dassie/config/tinymce.yml
```diff
@@ -0,0 +1,13 @@
...
+  <<: *default
+  menubar: false
+  toolbar1: styleselect | bold italic | undo redo
+  toolbar2: table | fullscreen | image
+  plugins:
+    - table
+    - fullscreen
+    - image
+custom:
+  <<: *default
```
dlpierce, line 13: This is for dassie which is not in 2.x, so this file can be removed from this PR.

dlpierce (changes requested):
Aside from the dassie config file, looks good and works locally for me. Thanks for the backport!

.dassie/config/tinymce.yml
no-reply: dassie was backported to 2.x sometime in the last few months!

no-reply (approved):

dlpierce:
Oh, I was not aware and the .dassie directory looked empty.

PR #5102, opened 2021-09-14
remove call to deprecated method

no-reply:
`#available_parent_collections` was deprecated in 0c0b08e0, but remained in this
view. use the helper implementation of this method.

@samvera/hyrax-code-reviewers


lsitu (approved):
Good catch!

PR #5104, opened 2021-09-15
do not adjust routing if name is already ‘Collection’

elrayle:
Without the check for `@name == 'Collection'`, a resource named `Collection` will enter an infinite loop as `Hyrax::CollectionName` attempts to update routes..

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/collection_name.rb
+++ b/lib/hyrax/collection_name.rb
@@ -7,7 +7,7 @@ module Hyrax
   class CollectionName < Name
     def initialize(klass, namespace = nil, name = nil)
       super
-
+      return if @name == 'Collection'
       @route_key          = Collection.model_name.route_key
       @singular_route_key = Collection.model_name.singular_route_key
     end


no-reply:
i'm not sure if i'm following. by "a resource named Collection", you mean a model class called `Collection`?

elrayle:
The infinite loop doesn't happen for `Collection < ActiveFedora::Base`.  It only happens with `Collection < Hyrax::PcdmCollection`.

no-reply:
@elrayle can you say more about the need for a `PcdmCollection` subclass?

by design, applications are meant to use `Hyrax::PcdmCollection` and `Hyrax::FileSet` directly. i can't see any reason a subclass wouldn't work (i hope we're consistently using `#collection?` and `#file_set?` rather than inheritance driven type checks, for this reason), but i'd like to understand what needs are pushing you in this direction.

i doubt this is the only place where `::Collection` as a class name is going to cause you trouble, since the assumption of a generated AF `::Collection` is endemic to the legacy code, and is probably assumed in some Wings-related compatibility code, too. i might suggest a different name for your subclass.

with that in mind, do you still think this extra check is helpful? or should we focus on warning other apps off of this pitfall?

if we're moving forward with this, let's switch it to `return if klass == ::Collection`, which should be quite a bit faster to resolve than string comparison. since this code will run frequently, it seems good to optimize the extra equality check.

elrayle:
Closing... Fixed in PR #5110

PR #5107, opened 2021-09-15
sipity: add debug-level logging to #Entity

dunn:
Because it's a recursive method, debugging entity creation can be difficult.
This helps trace the method flow when it's not clear how the input is being
transformed during the iterations.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/sipity.rb
+++ b/app/models/sipity.rb
@@ -57,23 +57,32 @@ module Sipity
   # @return [Sipity::Entity]
   # rubocop:disable Naming/MethodName, Metrics/CyclomaticComplexity, Metrics/MethodLength
   def Entity(input, &block)
+    Rails.logger.debug("Trying to make an Entity for #{input.inspect}")
+
     result = case input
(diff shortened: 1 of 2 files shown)


app/models/sipity.rb
```diff
@@ -57,23 +57,32 @@ def Agent(input, &block) # rubocop:disable Naming/MethodName
   # @return [Sipity::Entity]
   # rubocop:disable Naming/MethodName, Metrics/CyclomaticComplexity, Metrics/MethodLength
   def Entity(input, &block)
```
no-reply, line 59: let's prefer this inline, rather than in `fixme` (i doubt we will fix it)
```suggestion
  def Entity(input, &block) # rubocop:disable  Metrics/AbcSize
```

dunn: makes sense

out of curiosity, why are we using a handwritten fixme file instead of rubocop's autogenerated?

no-reply (approved):

PR #5111, opened 2021-09-20
remove dependency on collection.collection_type in dashboard/collections_controller

elrayle:
Hyrax::PcdmCollection does not respond to #collection_type.  Both Hydra::Pcdm::Collection and Hyrax::PcdmCollection respond to #collection_type_gid.  Updating `controllers/hyrax/dashboard/collections_controller.rb` to add private method `#collection_type` that gets the gid from the collection instance variable and returns the instance of collection_type.

This is part of the preparation for this controller to work with a Valkyrie resource.

@samvera/hyrax-code-reviewers


no-reply:
looks great!

no-reply (approved):

PR #5117, opened 2021-09-20
workflow: fix approval process for Valkyrie objects

dunn:
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/workflow_actions_controller.rb
+++ b/app/controllers/hyrax/workflow_actions_controller.rb
@@ -20,7 +20,7 @@ module Hyrax
 
     def curation_concern
       @curation_concern ||=
-        Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id])
+        Hyrax::ValkyrieCanCanAdapter.find(nil, params[:id])
     end
 
     def workflow_action_form
--- a/app/services/hyrax/workflow/deposited_notification.rb
+++ b/app/services/hyrax/workflow/deposited_notification.rb
@@ -14,7 +14,7 @@ module Hyrax
       end
 
       def users_to_notify
-        user_key = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: work_id).depositor
+        user_key = Hyrax::ValkyrieCanCanAdapter.find(nil, work_id).depositor
         super << ::User.find_by(email: user_key)
       end
     end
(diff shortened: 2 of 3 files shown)


app/controllers/hyrax/workflow_actions_controller.rb
```diff
@@ -20,7 +20,7 @@ def update
 
     def curation_concern
       @curation_concern ||=
-        Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: params[:id])
+        Hyrax::ValkyrieCanCanAdapter.find(nil, params[:id])
```
no-reply, line 23: should it be `load_resource` instead? or what's the right way to do this with CanCanCan directly, instead of hard coding the adapter?

app/services/hyrax/workflow/revoke_edit_from_depositor.rb
```diff
@@ -5,7 +5,7 @@ module Workflow
     # removes the creators the ability to alter it.
     module RevokeEditFromDepositor
       def self.call(target:, **)
-        target.edit_users -= [target.depositor]
+        target.edit_users = (target.edit_users.to_a - [target.depositor])
```
no-reply, line 8: does this work? i'd think we'd need to save the ACLs?

app/services/hyrax/workflow/deposited_notification.rb
```diff
@@ -14,7 +14,7 @@ def message
       end
 
       def users_to_notify
-        user_key = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: work_id).depositor
+        user_key = Hyrax::ValkyrieCanCanAdapter.find(nil, work_id).depositor
```
no-reply, line 17: does this need to be the alt id at all? or should it just be `Hyrax.query_service.find_by(id: work_id)`?

app/services/hyrax/workflow/revoke_edit_from_depositor.rb
no-reply: @dunn #5119 adds tests and handles this more thoroughly

app/services/hyrax/workflow/deposited_notification.rb
no-reply: turns out there's already a model resolver available to us via `@entity`. https://github.com/samvera/hyrax/pull/5120 uses that.

PR #5118, opened 2021-09-20
Add valkyrie support to Hyrax::Workflow::GrantReadToDepositor

no-reply:
following the pattern in `GrantEditToDepositor`, extend `GrantReadToDepositor` to support valkyrie works

@samvera/hyrax-code-reviewers


elrayle (approved):
Looks good.  Thanks!

PR #5120, opened 2021-09-21
use correct/configured user_key in Workflow::DepositedNotification

no-reply:
we should look up the depositor based on the user key, since apps can configure
their keys.

also some refactors on specs and the abstract class.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/workflow/deposited_notification.rb
+++ b/app/services/hyrax/workflow/deposited_notification.rb
@@ -10,11 +10,11 @@ module Hyrax
 
       def message
         I18n.t('hyrax.notifications.workflow.deposited.message', title: title, link: (link_to work_id, document_path),
-                                                                 user: user.user_key, comment: comment)
+               user: user.user_key, comment: comment)
       end
 
       def users_to_notify
-        user_key = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: work_id).depositor
+        user_key = Hyrax.query_service.find_by(id: work_id).depositor
         super << ::User.find_by(email: user_key)
       end
     end
--- a/app/services/hyrax/workflow/deposited_notification.rb
+++ b/app/services/hyrax/workflow/deposited_notification.rb
@@ -9,13 +9,17 @@ module Hyrax
       end
 
       def message
-        I18n.t('hyrax.notifications.workflow.deposited.message', title: title, link: (link_to work_id, document_path),
-               user: user.user_key, comment: comment)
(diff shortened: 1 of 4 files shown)


spec/services/hyrax/workflow/deposited_notification_spec.rb
```diff
@@ -1,35 +1,49 @@
...
-  let(:depositor) { create(:user) }
-  let(:to_user) { create(:user) }
-  let(:cc_user) { create(:user) }
-  let(:work) { create(:generic_work, user: depositor) }
-  let(:entity) { create(:sipity_entity, proxy_for_global_id: work.to_global_id.to_s) }
+  let(:approver) { FactoryBot.create(:user) }
+  let(:depositor) { FactoryBot.create(:user) }
+  let(:to_user) { FactoryBot.create(:user) }
+  let(:cc_user) { FactoryBot.create(:user) }
+  let(:work) { FactoryBot.create(:work, user: depositor) }
```
elrayle, line 7: Would this spec pass for a Valkyrie resource work?

app/services/hyrax/workflow/deposited_notification.rb
```diff
@@ -9,13 +9,17 @@ def subject
...
+               title: title,
+               link: (link_to work_id, document_path),
+               user: user.user_key,
+               comment: comment)
       end
 
       def users_to_notify
-        user_key = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: work_id).depositor
-        super << ::User.find_by(email: user_key)
+        user_key = @entity.proxy_for.depositor
```
elrayle, line 20: In the recent PRs for granting/revoking permissions, you checked that depositor method exists before using it.  Would the same apply here?

Example:
```
        return true unless target.try(:depositor)
```

elrayle (commented):
A couple of questions, but otherwise, looks good.

app/services/hyrax/workflow/deposited_notification.rb
no-reply: i think in those cases we want the grant to be a no-op if there's no depositor. here an error seems better than failing to notify silently.

spec/services/hyrax/workflow/deposited_notification_spec.rb
no-reply: no, it fails on other test setup issues. this PR isn't intended to change the tests to valkyrie.

elrayle (approved):
Answers to my questions make sense.  This looks good.  Thanks!

Issue #5085, opened 2021-08-25
Fix routes for non-work objects


elrayle:
### Descriptive summary

Non-work models fail to conform to expected rails routing causing `polymorphic_path` to produce non-existing paths causing callers of this method to fail.  This was seen while working with AdminSets but appears to impact Collection and FileSets as well.

### Rationale

Provide the rationale or user story that describes "why" this issue should be addressed. Especially if this is a new feature or significant change to the existing implementation.

### Expected behavior

```
>> Rails.application.routes.url_helpers.polymorphic_path(AdminSet.new(id: 3))
/admin/admin_sets/3
```
It should return a valid path.  Currently in Hyrax, the path to show an admin set is as seen in the example. Not sure this is following an expected Rails pattern.

### Actual behavior

```
>> Rails.application.routes.url_helpers.polymorphic_path(AdminSet.new)
NoMethodError (undefined method `admin_sets_path' ...)
```

### Steps to reproduce the behavior

1. In Rails console... `Rails.application.routes.url_helpers.polymorphic_path(AdminSet.new(id: 3))`

### Related work

PR #5083 update AdminSetCreateService to create valkyrie resources

no-reply:
is this still relevant?

PR #5122, opened 2021-09-21
limit lifecycle deposit and metadata updated events

elrayle:
Issue #5085 describes a problem with polymorphic paths causing some event jobs to fail for anything but works and filesets.  For now, these jobs will be skipped for other object types.  Having this allows these messages to be published for all object types.  There are other listeners that perform other actions that do work properly for all object types and that need to be run when these messages are published.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/listeners/object_lifecycle_listener.rb
+++ b/app/services/hyrax/listeners/object_lifecycle_listener.rb
@@ -14,13 +14,23 @@ module Hyrax
       ##
       # @param event [Dry::Event]
       def on_object_deposited(event)
+        return unless run_jobs(event)
         ContentDepositEventJob.perform_later(event[:object], event[:user])
       end
 
       ##
       # @param event [Dry::Event]
       def on_object_metadata_updated(event)
+        return unless run_jobs(event)
         ContentUpdateEventJob.perform_later(event[:object], event[:user])
+      end
+
+      private
+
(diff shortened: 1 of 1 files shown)


app/services/hyrax/listeners/object_lifecycle_listener.rb
```diff
@@ -14,14 +14,24 @@ def on_object_deleted(event)
...
       # @param event [Dry::Event]
       def on_object_metadata_updated(event)
+        return unless run_jobs(event)
         ContentUpdateEventJob.perform_later(event[:object], event[:user])
       end
+
+      private
+
+      def run_jobs(event)
+        # TODO: Jobs should run for all model types, but currently fails for
```
no-reply, line 31: i think the scope of this listener isn't all model types, but "PCDM Objects". in Hydra Works, we only account for "Work and "FileSet" as types of Objects.

what am i missing?

elrayle:
This PR is not needed.  Any messages for Collections and AdminSets will be published with the message starting with `collection` (e.g. `'collection.metadata.updated'`).  If there needs to be a lifecycle event job kicked off for collections, then a method will be added to `Hyrax::Listeners::ObjectLifecycleListener #on_collection_metadata_updated`.  _Note: If this is the first listener processes this message, the message will need to be registered._

PR #5123, opened 2021-09-21
use correct/configured user_key in Workflow::ChangesRequiredNotification

no-reply:
we should look up the depositor based on the user key, since apps can configure
their keys.

see also #5120 
@samvera/hyrax-code-reviewers


elrayle (approved):
Looks good. Thanks!

PR #5125, opened 2021-09-21
use changesets for ActionTakenService

no-reply:
@samvera/hyrax-code-reviewers


spec/services/hyrax/workflow/action_taken_service_spec.rb
```diff
@@ -23,55 +23,83 @@
...
     context "when the method exists" do
       around do |example|
-        class FooBar
-          def self.call; end
+        class SpyAction
+          class_attribute :no_op
+          class_attribute :fail
+          class_attribute :comment
+          class_attribute :target
+          class_attribute :user
```
elrayle, line 31: This is cool.

elrayle (approved):
I do like the use of ChangeSets.  It is an improvement.  Thanks for getting this one working.

PR #5143, opened 2021-09-24
add a listener for collection.metadata.updated to reindex collection

elrayle:
Adds listener for collection.metadata.updated that is separate from object.metadata.updated.  When it is published, the collection metadata will be reindexed.  

Members of collections can be works or collections.  The collection membership service was updated to publish the correct message depending on the type of object updated.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/collections/collection_member_service.rb
+++ b/app/services/hyrax/collections/collection_member_service.rb
@@ -155,8 +155,7 @@ module Hyrax
         private
 
         def publish_metadata_updated(member, user)
-          case member
-          when Hyrax::PcdmCollection
+          if member.collection?
             Hyrax.publisher.publish('collection.metadata.updated', collection: member, user: user)
           else
             Hyrax.publisher.publish('object.metadata.updated', object: member, user: user)
--- a/app/services/hyrax/collections/collection_member_service.rb
+++ b/app/services/hyrax/collections/collection_member_service.rb
@@ -110,7 +110,7 @@ module Hyrax
           raise Hyrax::SingleMembershipError, message if message.present?
           new_member.member_of_collection_ids << collection_id # only populate this direction
           new_member = Hyrax.persister.save(resource: new_member)
-          Hyrax.publisher.publish('object.metadata.updated', object: new_member, user: user)
+          publish_metadata_updated(new_member, user)
           new_member
         end
 
@@ -148,8 +148,19 @@ module Hyrax
           return member unless member?(collection_id: collection_id, member: member)
           member.member_of_collection_ids.delete(collection_id)
           member = Hyrax.persister.save(resource: member)
-          Hyrax.publisher.publish('object.metadata.updated', object: member, user: user)
+          publish_metadata_updated(member, user)
           member
+        end
+
+        private
+
+        def publish_metadata_updated(member, user)
+          case member
+          when Hyrax::PcdmCollection
+            Hyrax.publisher.publish('collection.metadata.updated', collection: member, user: user)
+          else
+            Hyrax.publisher.publish('object.metadata.updated', object: member, user: user)
+          end
         end
       end
     end
--- a/app/services/hyrax/listeners/metadata_index_listener.rb
+++ b/app/services/hyrax/listeners/metadata_index_listener.rb
@@ -15,11 +15,16 @@ module Hyrax
       # Re-index the resource.
       #
       # @param event [Dry::Event]
-      def on_object_metadata_updated(event)
-        log_non_resource(event) && return unless
-          event[:object].is_a?(Valkyrie::Resource)
+      def on_collection_metadata_updated(event)
+        metadata_updated(event, :collection)
+      end
 
-        Hyrax.index_adapter.save(resource: event[:object])
+      ##
+      # Re-index the resource.
+      #
+      # @param event [Dry::Event]
+      def on_object_metadata_updated(event)
+        metadata_updated(event, :object)
       end
 
       ##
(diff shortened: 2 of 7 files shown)


app/services/hyrax/collections/collection_member_service.rb
```diff
@@ -148,9 +148,20 @@ def remove_member(collection_id:, member:, user:)
...
           member = Hyrax.persister.save(resource: member)
-          Hyrax.publisher.publish('object.metadata.updated', object: member, user: user)
+          publish_metadata_updated(member, user)
           member
         end
+
+        private
+
+        def publish_metadata_updated(member, user)
+          case member
```
no-reply, line 158: i think this should be based on `member.collection?` to conform to the engine's commitments about typing for the PCDM classes?

app/services/hyrax/listeners/metadata_index_listener.rb
```diff
@@ -35,9 +40,16 @@ def on_object_deleted(event)
...
       private
 
-      def log_non_resource(event)
+      def log_non_resource(event, idx = :object)
         Hyrax.logger.info('Skipping object reindex because the object ' \
-                          "#{event[:object]} was not a Valkyrie::Resource.")
+                          "#{event[idx]} was not a Valkyrie::Resource.")
+      end
+
+      def metadata_updated(event, idx)
```
no-reply, line 48: would it be cleaner to pass the resource through from the payload directly, instead of the `event` and `idx`?

so the signatures for the private methods become:

`def log_non_resource(obj)` and `def metadata_updated(resource)`

app/services/hyrax/listeners/metadata_index_listener.rb
```diff
@@ -35,9 +40,16 @@ def on_object_deleted(event)
...
+        log_non_resource(event, idx) && return unless
```
no-reply, line 49: does this evaluate right? it looks like it will `log_non_resource` in all cases, and then `return unless...`?

elrayle: I just copied this down, so I assumed it is correct.  It may get re-evaluated as part of the requested method signature changes.

app/services/hyrax/collections/collection_member_service.rb
elrayle: That's much better.  Thanks for the suggestion.

app/services/hyrax/listeners/metadata_index_listener.rb
elrayle: This is a normal unless guard.  The log and return will not happen `unless` the resource isn't a Valkyrie::Resource.  It doesn't scan well.  I could put parentheses around the log && return part for readability.

elrayle: I did a refactor to make this more readable.  The check is simplified down to...
```
return unless resource? {resource}
```

The logging now occurs in the method checking if the payload has a Valkyrie::Resource.

spec/services/hyrax/collections/collection_member_service_spec.rb
```diff
@@ -144,7 +144,8 @@
 
     context 'when no members' do
       it "updates the collection member set to contain only the new members" do
-        expect(custom_query_service.find_members_of(collection: collection).map(&:id)).to match_array [new_member.id]
+        expect(custom_query_service.find_members_of(collection: collection).map(&:id))
+          .to match_array [new_member.id]
```
elrayle, line 148: noop - moved `.to` to next line for readability

spec/services/hyrax/collections/collection_member_service_spec.rb
```diff
@@ -153,19 +154,38 @@
       context 'and the new member already exists in the member set' do
         let(:new_member) { existing_work }
         it "the collection member set remains unchanged" do
-          expect(custom_query_service.find_members_of(collection: collection).map(&:id)).to match_array existing_member_ids
+          expect(custom_query_service.find_members_of(collection: collection).map(&:id))
+            .to match_array existing_member_ids
```
elrayle, line 158: no change - moved `.to` to next line for readability

spec/services/hyrax/collections/collection_member_service_spec.rb
```diff
@@ -153,19 +154,38 @@
...
         end
       end
       context 'and the new member does not exist in the member set' do
-        it "updates the collection member set adding the new member" do
-          expect(custom_query_service.find_members_of(collection: collection).map(&:id)).to match_array existing_member_ids + [new_member.id]
+        it "updates the collection member set adding the new work member" do
+          expect(custom_query_service.find_members_of(collection: collection)
+                   .map(&:id)).to match_array existing_member_ids + [new_member.id]
```
elrayle, line 164: no change - moved `.map` to next line for readability

spec/services/hyrax/collections/collection_member_service_spec.rb
```diff
@@ -153,19 +154,38 @@
...
         end
-        it "publishes metadata updated event for member" do
-          updated_work = described_class.add_member(collection_id: collection.id, new_member: new_member, user: user)
+        it "publishes object metadata updated event for work member" do
+          updated_work = described_class.add_member(collection_id: collection.id,
+                                                    new_member: new_member,
+                                                    user: user)
```
elrayle, line 169: no change - moved parameters to next line for readability

elrayle:
@no-reply Requested changes are all in the second commit.  Also uses the collection.metadata.updated message in save transaction step.

lib/hyrax/transactions/steps/save.rb
```diff
@@ -38,13 +38,23 @@ def call(change_set, user: nil)
...
 
+          publish_changes(unsaved, saved, user)
           Success(saved)
         rescue StandardError => err
           Failure([err.message, change_set.resource])
         end
+
+        private
+
+        def publish_changes(unsaved, saved, user)
```
no-reply, line 50: seems good. i take it this means you plan to re-use this transaction step for the collection save?

at some point in that process, we should add unit tests that run it against a collection.

no-reply (approved):

lib/hyrax/transactions/steps/save.rb
elrayle: This will be used in the controller to create a collection.  I have this working in my local branch and hope it will be ready to commit soon.

PR #5144, opened 2021-09-24
allow collection_type_ability to pass ability to PermissionsService

elrayle:
Calls to the collection_type and collection permission services should pass ability instead of current_user to allow for check of user_groups the user may be in.  The collection_type_ability to see if a user can create a collection of a certain type was passing the user.   Updating this to ability revealed an error in the permission service that was returning false if the user was not passed in.  This was fixed and tests were added to be sure that ability and user are both allowed as parameters.

This can probably be refactored at some point to remove the ability to pass either or both.  You can get the user from the ability (i.e. `ability.current_user`) and vice_versa (i.e. `user.ability`).  Would need to verify that this is always the case before refactoring.  This would change the method signature of the public methods requiring deprecation of the `ability` parameter.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/collection_types/permissions_service.rb
+++ b/app/services/hyrax/collection_types/permissions_service.rb
@@ -159,7 +159,7 @@ module Hyrax
       #   If calling from Abilities, pass the ability.  If you try to get the ability from the user, you end up in an infinit loop.
       def self.access_to_collection_type?(collection_type:, access:, user: nil, ability: nil) # rubocop:disable Metrics/CyclomaticComplexity
         return false unless user.present? || ability.present?
-        return false unless user && collection_type
+        return false unless collection_type
         return true if ([user_id(user, ability)] & agent_ids_for(collection_type: collection_type, agent_type: 'user', access: access)).present?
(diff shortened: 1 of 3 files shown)


app/services/hyrax/collection_types/permissions_service.rb
```diff
@@ -159,7 +159,7 @@ def self.create_access_for_collection_type?(collection_type:, user: nil, ability
       #   If calling from Abilities, pass the ability.  If you try to get the ability from the user, you end up in an infinit loop.
       def self.access_to_collection_type?(collection_type:, access:, user: nil, ability: nil) # rubocop:disable Metrics/CyclomaticComplexity
         return false unless user.present? || ability.present?
-        return false unless user && collection_type
+        return false unless collection_type
```
no-reply, line 162: was this just a bug?

what will happen on `user` references below? should we consider `user ||= ability&.current_user`?

i realize that whatever complex logic is happening here predates these changes, but i want to understand at least a little, if possible. :)

no-reply (approved):

Issue #5132, opened 2021-09-23
Val MVP: Create PcdmCollection as a valkyrie resource through UI


elrayle:
### Descriptive summary

As part of the Valkyrization MVP, a user should be able to create a collection through the UI with the underlying code processing the collection as a `Hyrax::PcdmCollection`.

### Rationale

Part of the basic Valkyrization MVP for testing the progress of the valkyrization effort.

### Steps to reproduce the behavior

_**Prerequisite**:  Assumes that collection type User Collection exists._

1. Login to [nurax-pg](https://nurax-pg.curationexperts.com)
1. Dashboard -> [Collections](https://nurax-pg.curationexperts.com/dashboard/my/collections?locale=en) -> New Collection
2. Select User Collection
2. Fill in required metadata
3. Save

The collection should be created and listed in the Dashboard -> Collections -> Your Collections.  Viewing the collection's show page should display with the correct metadata.

_NOTE: If you aren't an admin, there won't be All Collections and Your Collections tabs.  You will just see the list of collections and it should include the new collection._

### Related work

See all work tagged with Milestone [Valkyrization MVP](https://github.com/samvera/hyrax/milestone/19)

MPLSFedResearchTZ:
No confirmation after clicking save in creating a new collection, but actually the collection has been added to the db.

MPLSFedResearchTZ:
Got this error when editing an existing collection:
![image](https://user-images.githubusercontent.com/93392603/140778482-d63bb704-4860-42e8-a38d-e39c903ebda4.png)

rjkati:
When I create a new collection and click "save" the "save" button becomes grayed out and I do not see a confirmation message. When I click on the "Collections" tab in the dashboard, I can see the new collection. As a user, I expect nurax-pg to confirm that the collection has been created successfully.

rjkati:
Collections can be created on nurax-pg in hyrax 3.4.1

jlhardes:
Closing this issue since collections can be created.

PR #5147, opened 2021-09-25
add transactions related to creating collections

elrayle:
This is in preparation for using `Hyrax::PcdmCollection` for new/create in the collections controller.  Partially addresses Issue #5132 Val MVP: Create PcdmCollection as a valkyrie resource through UI.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/collection_create.rb
+++ b/lib/hyrax/transactions/collection_create.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+require 'hyrax/transactions/transaction'
+
+module Hyrax
+  module Transactions
+    ##
+    # Creates a Collection from a ChangeSet
+    #
+    # @since 3.0.0
+    class CollectionCreate < Transaction
+      DEFAULT_STEPS = ['change_set.set_user_as_depositor',
+                       'change_set.set_collection_type_gid',
+                       'change_set.add_to_collections',
+                       'change_set.apply',
+                       'collection_resource.apply_collection_type_permissions',
+                       'collection_resource.save_acl'].freeze
+
+      ##
+      # @see Hyrax::Transactions::Transaction
+      def initialize(container: Container, steps: DEFAULT_STEPS)
+        super
+      end
+    end
+  end
+end
--- a/lib/hyrax/transactions/collection_create.rb
+++ b/lib/hyrax/transactions/collection_create.rb
@@ -6,7 +6,7 @@ module Hyrax
     ##
     # Creates a Collection from a ChangeSet
     #
-    # @since 3.0.0
+    # @since 3.2.0
     class CollectionCreate < Transaction
       DEFAULT_STEPS = ['change_set.set_user_as_depositor',
                        'change_set.set_collection_type_gid',
--- a/lib/hyrax/transactions/container.rb
+++ b/lib/hyrax/transactions/container.rb
@@ -19,6 +19,7 @@ module Hyrax
     # @see https://dry-rb.org/gems/dry-container/
     class Container # rubocop:disable Metrics/ClassLength
       require 'hyrax/transactions/apply_change_set'
+      require 'hyrax/transactions/collection_create'
       require 'hyrax/transactions/create_work'
       require 'hyrax/transactions/destroy_work'
       require 'hyrax/transactions/file_set_destroy'
@@ -29,11 +30,13 @@ module Hyrax
       require 'hyrax/transactions/steps/add_to_collections'
       require 'hyrax/transactions/steps/add_to_parent'
       require 'hyrax/transactions/steps/apply_collection_permission_template'
+      require 'hyrax/transactions/steps/apply_collection_type_permissions'
       require 'hyrax/transactions/steps/apply_permission_template'
       require 'hyrax/transactions/steps/apply_visibility'
       require 'hyrax/transactions/steps/delete_resource'
       require 'hyrax/transactions/steps/destroy_work'
       require 'hyrax/transactions/steps/ensure_admin_set'
+      require 'hyrax/transactions/steps/set_collection_type_gid'
       require 'hyrax/transactions/steps/ensure_permission_template'
       require 'hyrax/transactions/steps/remove_file_set_from_work'
       require 'hyrax/transactions/steps/save'
@@ -58,6 +61,10 @@ module Hyrax
           ApplyChangeSet.new
         end
 
+        ops.register 'create_collection' do
+          CollectionCreate.new
+        end
+
         ops.register 'create_work' do
           WorkCreate.new
         end
@@ -68,6 +75,10 @@ module Hyrax
 
         ops.register 'save' do
           Steps::Save.new
+        end
+
+        ops.register 'set_collection_type_gid' do
+          Steps::SetCollectionTypeGid.new
         end
 
         ops.register 'set_default_admin_set' do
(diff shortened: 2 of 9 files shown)


lib/hyrax/transactions/steps/set_collection_type_gid.rb
```diff
@@ -0,0 +1,34 @@
...
+      # @since 2.4.0
+      class SetCollectionTypeGid
+        include Dry::Monads[:result]
+
+        ##
+        # @param [Hyrax::ChangeSet] change_set
+        # @param collection_type_gid [URI::GID] global id for the collection type
+        # @return [Dry::Monads::Result] `Failure` if there is no `collection_type_gid` or
+        #   it can't be set to the default for the input; `Success(input)`, otherwise.
+        def call(change_set, collection_type_gid: default_collection_type_gid)
```
no-reply, line 17: i'm wondering about the utility of this over treating `collection_type_gid` as a regular attribute? that is, doesn't `change_set.validate(collection_type_gid: some_gid)` work just as well?

i'm aware that at least some of the view forms and controllers don't include the `collection_type_gid` value in the collection params. the solution we came up with in Comet was to do `permitted = permitted.merge(collection_type_gid: params.require(:collection_type_gid))` in the controller. we felt like this user input interpretation was best handled directly in the controller. it doesn't seem any more complicated than adding step args, and removes a whole transaction step from the code base.

thoughts?

lib/hyrax/transactions/steps/apply_collection_type_permissions.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # A `dry-transcation` step that applies permission templates from a
+      # collection type on a given collection.
+      #
+      # @since 3.0.0
```
no-reply, line 9:
```suggestion
      # @since 3.2.0
```

lib/hyrax/transactions/collection_create.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+require 'hyrax/transactions/transaction'
+
+module Hyrax
+  module Transactions
+    ##
+    # Creates a Collection from a ChangeSet
+    #
+    # @since 3.0.0
```
no-reply, line 9:
```suggestion
      # @since 3.2.0
```

lib/hyrax/transactions/steps/set_collection_type_gid.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # A step that sets the `#collection_type_gid` in the change set.
+      #
+      # @since 2.4.0
```
no-reply, line 8:
```suggestion
      # @since 3.2.0
```

lib/hyrax/transactions/container.rb
```diff
@@ -109,6 +120,16 @@ class Container # rubocop:disable Metrics/ClassLength
         end
       end
 
+      namespace 'collection_resource' do |ops| # valkyrie collection
+        ops.register 'apply_collection_type_permissions' do
+          Steps::ApplyCollectionTypePermissions.new
+        end
+
+        ops.register 'save_acl' do
+          Steps::SaveAccessControl.new
```
no-reply, line 129: is this a real step (i couldn't find it here, or in the existing codebase)? if so, is it needed or could this be handled by the pub/sub setup?

elrayle: This uses the same [step](https://github.com/samvera/hyrax/blob/main/lib/hyrax/transactions/steps/save_access_control.rb) as used by [WorkCreate](https://github.com/samvera/hyrax/blob/main/lib/hyrax/transactions/work_create.rb#L16) ([registered in container](https://github.com/samvera/hyrax/blob/main/lib/hyrax/transactions/container.rb#L129-L131)).  This code registers the SaveAccessControl step for use with collections.

The CollectionCreate process follows the pattern established by WorkCreate.

no-reply: ah thanks, this is just me being confused.

elrayle:
NOTE: I accepted the proposed comment changes for `@since` and they are causing rubocop errors.  I will bring this back down local and fix rubocop.  The code has not changed.

lib/hyrax/transactions/steps/set_collection_type_gid.rb
no-reply: i guess what i'm asking is: what is different about `collection_type_gid` vs any other attribute on this model/form? what's the reason for not setting it in the same way all the other attributes are set? likewise, why not handle defaults and validations in the same way defaults and validations are handled for other attributes (using `Valkyrie::ChangeSet`)?

i'd rather not confuse things with [comet's specifics](https://gitlab.com/surfliner/surfliner/-/blob/trunk/comet/app/controllers/hyrax/dashboard/collections_controller.rb#L19-41) (there's no collection transaction on comet's side, for no particularly good reason). the relevant code samples are included in my original comment.

i get the benefits of transactions in general, but in this case it's 35 extra lines of code and the effects i see are:

  - apply an attribute that would otherwise get applied with all the others in the `validate` step;
  - setup a default that will silently override any defaults set on the form object.

it just seems to complicate things to break this out into a separate step.

lib/hyrax/transactions/steps/apply_collection_type_permissions.rb
```diff
@@ -0,0 +1,29 @@
...
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # A `dry-transcation` step that applies permission templates from a
+      # collection type on a given collection.
+      #
+      # @since 3.2.0
+      class ApplyCollectionTypePermissions
+        include Dry::Transaction::Operation
```
no-reply, line 11: this seems to be the deprecated transaction infrastructure

should it be:
```suggestion
        include Dry::Monads[:result]
```

lib/hyrax/transactions/steps/set_collection_type_gid.rb
elrayle: `CollectionCreate` is modeled after the pattern in [WorkCreate](https://github.com/samvera/hyrax/blob/main/lib/hyrax/transactions/work_create.rb).  `WorkCreate` has steps [change_set.set_default_admin_set](https://github.com/samvera/hyrax/blob/main/lib/hyrax/transactions/steps/set_default_admin_set.rb), which sets the admin set to the default if one isn't assigned, and a second step [change_set.ensure_admin_set](https://github.com/samvera/hyrax/blob/main/lib/hyrax/transactions/steps/ensure_admin_set.rb), which verifies that the admin_set is set.  The transaction steps in `CollectionCreate` are consistent with those in `WorkCreate` in level of granularity.

Are you inclined to refactor the `WorkCreate` transactions?

no-reply: i think `change_set.set_default_admin_set` calls a service which has to create permission templates as well as assigning a variable.

if there's a way to neatly refactor those transaction steps to use form features, then yes, that seems very attractive to me. i'm not trying to have an argument about this though. i'm trying to provide a helpful review. if you don't want to change this, you can just say so.

no-reply (approved):

PR #5148, opened 2021-09-27
update collection ability to work with Hyrax::PcdmCollection

elrayle:
This is in preparation for using Hyrax::PcdmCollection for new/create in the collections controller. Partially addresses Issue #5132 Val MVP: Create PcdmCollection as a valkyrie resource through UI.

Updates collection_ability to allow/disallow the same abilities when the object is an `ActiveFedora::Base` `::Collection` or a `Valkyrie::Resource` `Hyrax::PcdmCollection`.  The same tests are run for both object classes.

Notable changes:
* The ActiveFedora tests were refactored to have each ability check run separately to allow all failing abilities to be identified.  Otherwise, only the first failure is reported.  I believe it was original written as a single test for performance reasons.  With CircleCI running tests in parallel, this seems less critical.
* There was an error in the Depositor tests that denied read access to the public show page, but allowed access to the more restricted admin show page.  This was an error in the test and not in the code.  A slight rewrite of the test shows the abilities for Depositor behaving as expected.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/abilities/collection_ability_spec.rb
+++ b/spec/abilities/collection_ability_spec.rb
@@ -1,164 +1,392 @@
 # frozen_string_literal: true
 require 'cancan/matchers'
 
-RSpec.describe Hyrax::Ability do
+RSpec.describe Hyrax::Ability, :clean_repo do
   subject { ability }
 
   let(:ability) { Ability.new(current_user) }
-  let(:user) { create(:user) }
+  let(:user) { create(:user, email: 'user@example.com') }
   let(:current_user) { user }
   let(:collection_type) { FactoryBot.create(:collection_type) }
+  let(:collection_type_gid) { collection_type.to_global_id }
 
   context 'when admin user' do
-    let(:user) { FactoryBot.create(:admin) }
-    let!(:collection) { FactoryBot.build(:collection_lw, id: 'col_au', with_permission_template: true, collection_type: collection_type) }
-    let!(:solr_document) { SolrDocument.new(collection.to_solr) }
+    let(:current_user) { admin }
+    let(:admin) { FactoryBot.create(:admin, email: 'admin@example.com') }
 
-    it 'allows all abilities' do # rubocop:disable RSpec/ExampleLength
-      is_expected.to be_able_to(:manage, Collection)
-      is_expected.to be_able_to(:manage_any, Collection)
-      is_expected.to be_able_to(:create_any, Collection)
-      is_expected.to be_able_to(:view_admin_show_any, Collection)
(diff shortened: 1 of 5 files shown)


spec/abilities/collection_ability_spec.rb
```diff
@@ -1,164 +1,392 @@
...
+      end
+
+      context 'for abilities open to depositor' do
+        it { is_expected.to be_able_to(:view_admin_show_any, Collection) }
+        it { is_expected.to be_able_to(:deposit, collection) }
+        it { is_expected.to be_able_to(:deposit, solr_document) }
+        it { is_expected.to be_able_to(:view_admin_show, collection) }
+        it { is_expected.to be_able_to(:view_admin_show, solr_document) }
+        it { is_expected.to be_able_to(:read, collection) }
+        it { is_expected.to be_able_to(:read, solr_document) } # defined in solr_document_ability.rb
```
elrayle, line 190: These 2 read statements were denied in the original test, but shouldn't have been.  The update to the test that revealed this error was having `user@example.com` create the collection and granting `depositor@example.com` DEPOSIT access.  Both users should have read access.

See [PermissionTemplate #read_users](https://github.com/samvera/hyrax/blob/main/app/models/hyrax/permission_template.rb#L203-L208) which sets `read_users` to be users with `view` and `deposit` access.

spec/factories/hyrax_collection.rb
```diff
@@ -8,18 +8,13 @@
...
+      user { create(:user) }
       edit_groups { [] }
       edit_users { [] }
       read_groups { [] }
       read_users { [] }
       members { nil }
-    end
-
-    after(:build) do |collection, evaluator|
-      collection.permission_manager.edit_groups = evaluator.edit_groups
```
elrayle, line 19: Permission templates should not be created prior to saving the resource.

spec/hyrax/transactions/steps/apply_collection_type_permissions_spec.rb
```diff
@@ -7,7 +7,8 @@
   let(:collection) do
     FactoryBot.valkyrie_create(:hyrax_collection,
                                title: "My Resource",
-                               collection_type_gid: collection_type_gid)
+                               collection_type_gid: collection_type_gid,
+                               with_permission_template: false)
```
elrayle, line 11: This is testing the ability of this step to add the permission template and associated initial state permissions.  It will fail if the permission template already exists.

no-reply (approved):

PR #5149, opened 2021-09-29
use CanCan tooling to load resource for `WorkflowActionsController`

no-reply:
there doesn't seem to be any reason to avoid using the configurable cancan
resource resoultion. switiching to this setup allows us to respect the noid
configuration, as handled by `Hyrax::ValkyrieCanCanAdapter`.

cc: @dunn i think this replaces #5117

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/workflow_actions_controller.rb
+++ b/app/controllers/hyrax/workflow_actions_controller.rb
@@ -3,6 +3,13 @@ module Hyrax
   class WorkflowActionsController < ApplicationController
     DEFAULT_FORM_CLASS = Hyrax::Forms::WorkflowActionForm
 
+    ##
+    # @!attribute [r] curation_concern
+    #   @api private
+    #   @return [Hyrax::Resource]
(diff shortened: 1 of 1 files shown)


app/controllers/hyrax/workflow_actions_controller.rb
```diff
@@ -3,6 +3,13 @@ module Hyrax
...
     DEFAULT_FORM_CLASS = Hyrax::Forms::WorkflowActionForm
 
+    ##
+    # @!attribute [r] curation_concern
+    #   @api private
+    #   @return [Hyrax::Resource]
+    attr_reader :curation_concern
+
+    load_resource class: Hyrax::Resource, instance_name: :curation_concern
     before_action :authenticate_user!
```
no-reply, line 13: we should probably be able to do `load_and_authorize_resource` but i'm wanting to only change one thing at a time, here.

dunn (approved):

PR #5152, opened 2021-09-30
update collection controller to process create with Hyrax::PcdmCollection

### LINKED ISSUE #5132: Val MVP: Create PcdmCollection as a valkyrie resource through UI

elrayle:
Fixes #5132

Includes changes to Hyrax::Forms::PcdmCollectionForm that support creating collections through the UI.

There are a few changes in the presentation to users with the change to Hyrax::PcdmCollection based on the way metadata is defined in that class.  See comments in changed files for more information.

NOTE: This was confirmed in the UI that a collection can be created with `Hyrax:PcdmCollection` and then edited using `::Collection`.  Edit will also be updated in a later PR.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -43,7 +43,7 @@ module Hyrax
       # The search builder to find the collections' members
       self.membership_service_class = Collections::CollectionMemberSearchService
 
-      load_and_authorize_resource except: [:index, :create], instance_name: :collection
+      load_and_authorize_resource except: [:index, :create, :new], instance_name: :collection
 
       def deny_collection_access(exception)
         if exception.action == :edit
@@ -57,14 +57,10 @@ module Hyrax
       end
 
       def new
-        # Coming from the UI, a collection type id should always be present.  Coming from the API, if a collection type id is not specified,
-        # use the default collection type (provides backward compatibility with versions < Hyrax 2.1.0)
-        collection_type_id = params[:collection_type_id].presence || default_collection_type.id
-        @collection.collection_type_gid = CollectionType.find(collection_type_id).to_global_id
+        authorize! :create_collection_of_type, collection_type
         add_breadcrumb t(:'hyrax.controls.home'), root_path
         add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
         add_breadcrumb t('.header', type_title: collection_type.title), request.path
-        @collection.apply_depositor_metadata(current_user.user_key)
         form
       end
 
@@ -82,42 +78,33 @@ module Hyrax
 
       def after_create
         form
-        set_default_permissions
         # if we are creating the new collection as a subcollection (via the nested collections controller),
         # we pass the parent_id through a hidden field in the form and link the two after the create.
-        link_parent_collection(params[:parent_id]) unless params[:parent_id].nil?
         respond_to do |format|
-          Hyrax::SolrService.commit
           format.html { redirect_to edit_dashboard_collection_path(@collection), notice: t('hyrax.dashboard.my.action.collection_create_success') }
           format.json { render json: @collection, status: :created, location: dashboard_collection_path(@collection) }
         end
       end
 
-      def after_create_error
-        form
+      def after_create_error(err_msg:)
+        flash[:error] = err_msg
         respond_to do |format|
           format.html { render action: 'new' }
-          format.json { render json: @collection.errors, status: :unprocessable_entity }
+          format.json { render json: err_msg, status: :unprocessable_entity }
         end
       end
 
       def create
-        # Manual load and authorize necessary because Cancan will pass in all
-        # form attributes. When `permissions_attributes` are present the
-        # collection is saved without a value for `has_model.`
-        @collection = ::Collection.new
-        authorize! :create, @collection
-        # Coming from the UI, a collection type gid should always be present.  Coming from the API, if a collection type gid is not specified,
-        # use the default collection type (provides backward compatibility with versions < Hyrax 2.1.0)
-        @collection.collection_type_gid = params[:collection_type_gid].presence || default_collection_type.to_global_id
-        @collection.attributes = collection_params.except(:members, :parent_id, :collection_type_gid)
-        @collection.apply_depositor_metadata(current_user.user_key)
-        @collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE unless @collection.discoverable?
-        if @collection.save
+        authorize! :create_collection_of_type, collection_type
+        form
+        result = create_collection(form: @form,
+                                   user: current_user,
+                                   collection_type_gid: collection_type.to_global_id,
+                                   parent_id: params[:parent_id])
+        if result == true
           after_create
-          add_members_to_collection unless batch.empty?
         else
-          after_create_error
+          after_create_error(err_msg: result)
         end
       end
 
@@ -146,7 +133,7 @@ module Hyrax
         @collection.visibility = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE unless @collection.discoverable?
         # we don't have to reindex the full graph when updating collection
         @collection.reindex_extent = Hyrax::Adapters::NestingIndexAdapter::LIMITED_REINDEX
-        if @collection.update(collection_params.except(:members))
+        if @collection.update(collection_params_without_permissions.except(:members))
           after_update
         else
           after_update_error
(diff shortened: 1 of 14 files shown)


app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -43,7 +43,7 @@ class CollectionsController < Hyrax::My::CollectionsController
       # The search builder to find the collections' members
       self.membership_service_class = Collections::CollectionMemberSearchService
 
-      load_and_authorize_resource except: [:index, :create], instance_name: :collection
+      load_and_authorize_resource except: [:index, :create, :new], instance_name: :collection
```
elrayle, line 46: Requited to accomodate Hyrax::PcdmCollection instead of ::Collection.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -57,14 +57,10 @@ def deny_collection_access(exception)
...
       def new
-        # Coming from the UI, a collection type id should always be present.  Coming from the API, if a collection type id is not specified,
-        # use the default collection type (provides backward compatibility with versions < Hyrax 2.1.0)
-        collection_type_id = params[:collection_type_id].presence || default_collection_type.id
-        @collection.collection_type_gid = CollectionType.find(collection_type_id).to_global_id
+        authorize! :create_collection_of_type, collection_type
         add_breadcrumb t(:'hyrax.controls.home'), root_path
         add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
         add_breadcrumb t('.header', type_title: collection_type.title), request.path
-        @collection.apply_depositor_metadata(current_user.user_key)
```
elrayle, line 67: This happens in transactions.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -82,42 +78,30 @@ def edit
 
       def after_create
         form
-        set_default_permissions
```
elrayle, line 85: This happens in transactions.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -82,42 +78,30 @@ def edit
...
         # if we are creating the new collection as a subcollection (via the nested collections controller),
         # we pass the parent_id through a hidden field in the form and link the two after the create.
-        link_parent_collection(params[:parent_id]) unless params[:parent_id].nil?
```
elrayle, line 88: This happens in transactions.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -82,42 +78,30 @@ def edit
...
         respond_to do |format|
-          Hyrax::SolrService.commit
```
elrayle, line 90: This happens when metadata change is published.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -82,42 +78,30 @@ def edit
...
           format.html { redirect_to edit_dashboard_collection_path(@collection), notice: t('hyrax.dashboard.my.action.collection_create_success') }
           format.json { render json: @collection, status: :created, location: dashboard_collection_path(@collection) }
         end
       end
 
-      def after_create_error
-        form
```
elrayle, line 97: form isn't used here, so it is not needed.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -82,42 +78,30 @@ def edit
...
+        authorize! :create_collection_of_type, collection_type
+        begin
+          create_form_value(user: current_user,
+                            collection_type_gid: params[:collection_type_gid],
+                            parent_id: params[:parent_id])
           after_create
-          add_members_to_collection unless batch.empty?
-        else
-          after_create_error
+        rescue Wings::Valkyrie::Persister::FailedSaveError => ex
```
elrayle, line 103: Is there a generic Valkyrie error that all persisters raise when save fails?  Having a Wings error here will cause problems when switching to Postgres or other adatper.

app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -345,6 +362,12 @@ def single_item_search_builder
       deprecation_deprecate :single_item_search_builder
 
       def collection_params
+        # For backward compatibility, need to support params[:collection] for existing APIs
+        @collection_params = params[:collection]
+        @collection_params ||= params[:pcdm_collection]
+      end
+
+      def collection_params_without_permissions
         @participants = extract_old_style_permission_attributes(params[:collection])
```
elrayle, line 371: Do we still need to support passing in permissions in the params?  This was marked for deprecation starting at 3.0, but didn't get removed.  See related methods...  [controllers/hyrax/dashboard/collections_controller.rb](https://github.com/samvera/hyrax/blob/main/app/controllers/hyrax/dashboard/collections_controller.rb#L352-L376)

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -21,97 +21,133 @@
...
+      expect(assigns(:collection)).to be_kind_of(Hyrax::PcdmCollection)
     end
   end
 
   describe '#create' do
     before { sign_in user }
 
-    # rubocop:disable RSpec/ExampleLength
-    it "creates a Collection" do
+    it 'creates a Collection with metadata' do # rubocop:disable RSpec/ExampleLength
```
elrayle, line 48: New Test to test basic metadata is assigned.

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -21,97 +21,133 @@
...
-      expect(assigns[:collection].edit_users).to contain_exactly "archivist1", user.email
+      expect(assigns(:collection)).to be_kind_of(Hyrax::PcdmCollection)
+      expect(assigns[:collection].visibility).to eq 'restricted'
+      expect(assigns[:collection].depositor).to eq user.user_key
+      expect(assigns[:collection].permission_manager.edit_users).to contain_exactly user.user_key
       expect(flash[:notice]).to eq "Collection was successfully created."
     end
 
-    it "removes blank strings from params before creating Collection" do
+    it "publishes collection.metadata.updated" do
```
elrayle, line 62: New Test: tests that collection metadata updated is published which will index the collection.

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -21,97 +21,133 @@
...
-      expect(assigns(:collection)).to be_kind_of(Collection)
+      expect(assigns(:collection)).to be_kind_of(Hyrax::PcdmCollection)
     end
   end
 
   describe '#create' do
     before { sign_in user }
 
-    # rubocop:disable RSpec/ExampleLength
-    it "creates a Collection" do
```
elrayle, line 40: This test is currently commented out starting around line 73.  This is related to the question about deprecated old style parameters and whether or not we can remove support for them.

Additional question about `:create` supporting passing in parameters not defined on the form (e.g. visibility).  Should this be allowed?  What is the use case for this?  For the visibility params specifically, you can set it in the edit form, just not the new form.  The question applies in general.

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -21,97 +21,133 @@
...
       expect do
         post :create, params: {
-          collection: collection_attrs.merge(creator: [''])
-        }
-      end.to change { Collection.count }.by(1)
-      expect(assigns[:collection].title).to eq ["My First Collection"]
-      expect(assigns[:collection].creator).to eq []
-    end
-
-    context "with files I can access" do
```
elrayle, line 69: Note: These assets are works and collections that are being added as children of the collection being created.  Is there a use case for this?  You can do this through the edit form or batch process (e.g., I believe this is used on the Dashboard -> Works -> select works -> Add works to collection).  I thought perhaps you could add selected works to a New Collection, but I don't see that as an option through the UI.

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
```diff
@@ -147,14 +184,7 @@
   end
 
   describe "#update" do
-    let(:listener) { Hyrax::Specs::SpyListener.new }
```
elrayle, line 150: `listener` is used in multiple places so it was moved outside of the `#update` tests.

app/controllers/hyrax/dashboard/collections_controller.rb
no-reply: can you say more about this? why not load `Hyrax::PcdmCollection` with CanCan's tooling as well?

app/controllers/hyrax/dashboard/collections_controller.rb
no-reply: what are you trying to rescue from? is there any reason we wouldn't want to run `after_create_error` on all uncaught `StandardError`s raised during execution of the main block?

app/controllers/hyrax/dashboard/collections_controller.rb
jlhardes: I am not sure if this is connected, but Bulkrax might include the ability to create collections when a work is imported (created). Would that require passing in parameters here that aren't defined or visible on the form?

no-reply: seems like if we drop them we need to bump the major version.

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
jlhardes: Again, the only use case that I can think of that might require passing in parameters that aren't part of the form is if a collection creation action doesn't use the form - an import or something along those lines.

spec/controllers/hyrax/dashboard/collections_controller_spec.rb
jlhardes: I think you are correct that you can't select works and create a collection to add them to via the UI (the collection has to already exist) but if Bulkrax is expecting this functionality when importing new works, does that effect this test?

app/controllers/hyrax/dashboard/collections_controller.rb
elrayle: This went away because the exception is rescued in the `save` transaction and it rescues `StandardError`.

PR #5153, opened 2021-09-30
apply more targeted handling for transactional save failures

no-reply:
narrow the scope for failure message handling in the transactional save step.

this avoids blanket catching errors that arise due to listeners for published
events. treating all the possible errors from this method as exactly the same,
and squashing them to their `#message`, makes debugging hard (e.g. it's hard to
tell what is raising, exactly).

this change will cause errors from published events to raise eagerly, skipping
the `Success|Failure` `Result` monad pipeline. in theory, this should be okay,
since pub/sub listeners should generally not be raising uncaught exceptions. it
appears that some ARE at this point, at least in some cases. the hope is that
this change will help us get to the bottom of that.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/steps/save.rb
+++ b/lib/hyrax/transactions/steps/save.rb
@@ -43,18 +43,18 @@ module Hyrax
 
           user ||= ::User.find_by_user_key(saved.depositor)
 
-          publish_changes(unsaved, saved, user)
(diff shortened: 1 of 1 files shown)


no-reply:
this change was integral in diagnosing the issue that motivated https://github.com/samvera/hyrax/pull/5154

i hope we can merge it!

dunn (approved):

PR #5157, opened 2021-10-01
Update PcdmCollection metadata to match existing ::Collection metadata

elrayle:
This is in preparation for using `Hyrax::PcdmCollection` for new/create in the collections controller. Partially addresses Issue #5132 Val MVP: Create PcdmCollection as a valkyrie resource through UI.

This issue update the metadata for `Hyrax::PcdmCollection` to match that currently provided through `::Collection`.

Possible follow-on work:
* Currently, core metadata is included in `collection_basic_metadata` to allow title to have multiple values specified in the form.  If there was support for overriding metadata with the last reference to an attribute taking precedence, then `core_metadata` can be included and an override for just `title` can happen in `collection_basic_metadata`.
* `collection_metadata` is an empty metadata file that allows sites to add additional metadata for collections by adding this file to their app and adding custom metadata definitions.  This may need additional testing and may be dependent on override functionality being in place first.
* `collection_basic_metadata` does not bring in thumbnail selection into the form.
* Locations, driven by the `based_near` attribute, is not connecting to `GeoNames` for auto-complete.  It is possible that there is some outside configuration that has to happen in the app for this to work and does not represent additional work in Hyrax.  Testing needs to happen to confirm this.
* There is a larger question about what metadata should be included as collection metadata.  This will require community input to determine if we want to make a change.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/pcdm_collection_form.rb
+++ b/app/forms/hyrax/forms/pcdm_collection_form.rb
@@ -6,6 +6,9 @@ module Hyrax
     # @api public
     # @see https://github.com/samvera/valkyrie/wiki/ChangeSets-and-Dirty-Tracking
     class PcdmCollectionForm < Valkyrie::ChangeSet # rubocop:disable Metrics/ClassLength
+      include Hyrax::FormFields(:collection_basic_metadata)
+      include Hyrax::FormFields(:collection_metadata)
+
       property :title, required: true, primary: true
 
       property :human_readable_type, writable: false
--- a/app/forms/hyrax/forms/pcdm_collection_form.rb
+++ b/app/forms/hyrax/forms/pcdm_collection_form.rb
@@ -6,6 +6,7 @@ module Hyrax
     # @api public
     # @see https://github.com/samvera/valkyrie/wiki/ChangeSets-and-Dirty-Tracking
     class PcdmCollectionForm < Valkyrie::ChangeSet # rubocop:disable Metrics/ClassLength
+      include Hyrax::FormFields(:core_metadata)
       include Hyrax::FormFields(:collection_basic_metadata)
       include Hyrax::FormFields(:collection_metadata)
 
@@ -21,7 +22,6 @@ module Hyrax
 
       property :member_of_collection_ids, default: [], type: Valkyrie::Types::Array
 
-      validates :title, presence: true
       validates :collection_type_gid, presence: true
 
       class << self
--- a/app/services/hyrax/simple_schema_loader.rb
+++ b/app/services/hyrax/simple_schema_loader.rb
@@ -76,7 +76,11 @@ module Hyrax
       ##
       # @return [Dry::Types::Type]
       def type
-        collection_type = config['multiple'] ? Valkyrie::Types::Array : Identity
+        collection_type = if config['multiple']
+                            Valkyrie::Types::Array.constructor { |v| Array(v).select(&:present?) }
+                          else
+                            Identity
+                          end
         collection_type.of(type_for(config['type']))
       end
 
--- a/lib/hyrax/specs/shared_specs/metadata.rb
+++ b/lib/hyrax/specs/shared_specs/metadata.rb
@@ -96,28 +96,7 @@ end
 RSpec.shared_examples 'a model with collection basic metadata' do
   subject(:resource) { described_class.new }
   let(:date)         { Time.zone.today }
-
-  # from core metadata included in collection basic metadata
-  it 'has a date_modified' do
-    expect { resource.date_modified = date }
-      .to change { resource.date_modified }
-            .to eq date
-  end
-
-  # from core metadata included in collection basic metadata
-  it 'has a date_uploaded' do
-    expect { resource.date_uploaded = date }
-      .to change { resource.date_uploaded }
-            .to eq date
-  end
-
-  # from core metadata included in collection basic metadata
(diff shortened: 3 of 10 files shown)


config/locales/hyrax.en.yml
```diff
@@ -1566,6 +1566,21 @@ en:
         resource_type: Pre-defined categories to describe the type of content being uploaded, such as &quot;article&quot; or &quot;dataset.&quot;  More than one type may be selected.
         subject: Headings or index terms describing what the collection is about; these do need to conform to an existing vocabulary.
         title: A name to aid in identifying a collection.
+      pcdm_collection:
```
no-reply, line 1569: what's the cause of the need for new i18n keys?

elrayle: I don't know where the generation happens, but something generates the part of the i18n key to be `pcdm_collection` instead of `collection`.

elrayle: Without the additional i18n keys, the default keys are used which has `work` in place of `collection` for many of the fields.

no-reply: let's patch it to use the existing `collection` keys, instead of introducing new, duplicate ones

jlhardes:
Hopefully this is a helpful comment here: The descriptive fields on Collection are something that is part of Hyrax now (3.3.0) and we should keep those descriptive fields in place in Valkyrized Hyrax. I don't want to remove features from Hyrax moving forward. I can understand that collection level metadata might vary more between institutional implementations so if it is possible to provide an option to change/remove/override the collection descriptive fields, that seems reasonable, but since they are already part of Collections in current Hyrax, those fields need to remain part of the code base.

marrus-sh:
We’re currently looking at using [samvera-labs/houndstooth](https://github.com/samvera-labs/houndstooth) to provide metadata information for [surfliner/comet](https://gitlab.com/surfliner/surfliner/-/tree/trunk/comet) (which is Hyrax‐ and Valkyrie‐based) and would like to define all non‐mission metadata fields in that way.

Ideally, I think we would be able to define our own collection as a subclass of `Hyrax::PcdmCollection`, setting it as `Hyrax.config.collection_class` (is that right?), and use the subclass definition to provide our own metadata. This is how we are currently handling objects (as a subclass of `Hyrax::Work`).

But that method won’t work if `Hyrax::PcdmCollection` necessarily loads a bunch of metadata already. A means to disable that metadata loading would be appreciated!

config/locales/hyrax.en.yml
elrayle: The PR was updated to remove the changes to translations.

elrayle:
@marrus-sh The last commit added documentation for how to override and/or extend metadata for collections.  To remove all metadata defined in the basic metadata yaml file, all you need to do is edit the file and replace the contents with...

```
attributes = {}
```

Collections will continue to include core_metadata.yaml, which they did prior to this PR.

Any other concerns for the surfliner/comet use case?

app/services/hyrax/simple_schema_loader.rb
```diff
@@ -76,7 +76,11 @@ def index_keys
       ##
       # @return [Dry::Types::Type]
       def type
-        collection_type = config['multiple'] ? Valkyrie::Types::Array : Identity
+        collection_type = if config['multiple']
+                            Valkyrie::Types::Array.constructor { |v| Array(v).select(&:present?) }
+                          else
+                            Identity
+                          end
```
elrayle, line 83: This is required to remove empty field values passed in from the form.  For example, this changes `creator: [""]` to `creator: []`.

app/forms/hyrax/forms/pcdm_collection_form.rb
```diff
@@ -6,6 +6,10 @@ module Forms
     # @api public
     # @see https://github.com/samvera/valkyrie/wiki/ChangeSets-and-Dirty-Tracking
     class PcdmCollectionForm < Valkyrie::ChangeSet # rubocop:disable Metrics/ClassLength
+      include Hyrax::FormFields(:core_metadata)
+      include Hyrax::FormFields(:collection_basic_metadata)
+      include Hyrax::FormFields(:collection_metadata)
+
       property :title, required: true, primary: true
```
elrayle, line 13: I'm wondering if title should be defined as a property here since core_metadata is included.

lib/hyrax/specs/shared_specs/metadata.rb
```diff
@@ -92,3 +92,70 @@
       .to contain_exactly('title')
   end
 end
+
+RSpec.shared_examples 'a model with collection basic metadata' do
+  subject(:resource) { described_class.new }
+  let(:date)         { Time.zone.today }
```
elrayle, line 98: Including this in shared_specs can cause apps that override basic metadata to fail this set of specs.  I'll look at how works handle basic metadata.  I would think they would have the same issue.

elrayle:
@marrus-sh I am exploring using a generator to create the valkyrie version of the collection class which would allow pushing the inclusion of default basic metadata out to that class.  Then if you want to remove the basic metadata, you would simply comment out that line.  This would be closer to what happens now with Works.

marrus-sh:
@elrayle this all sounds workable to me—thanks for your efforts here :)

tpendragon:
Collections are being created. I think it's possible all the metadata isn't there, but we're not seeing that in QA. Closing.

PR #5158, opened 2021-10-01
add validations for required fields

elrayle:
Required fields were not getting validated.  The tests in this PR both return true for `valid?` without the addition to set `validates`.

See this gotcha described in [Dive into Valkyrie](https://github.com/samvera/valkyrie/wiki/Validating-change-sets#resource-attribute-with-required-true-passes-validation-when-missing).

@samvera/hyrax-code-reviewers


no-reply (approved):
awesome! thanks!

PR #5159, opened 2021-10-01
WIP: Add read only mode feature

### LINKED ISSUE #2877: Add a read-only mode flipper
### Descriptive summary
Add a FlipFlop feature flipper to allow the system to be put into read-only mode. While in read-only mode, no changes can be made. No new works can be deposited, no workflow changes can be made, etc. 

### Rationale

It is difficult and dangerous to perform large scale data migrations or imports live on a production repository. It would be much better if we could put the production system into read-only mode, take a snapshot, migrate on the snapshot, and then replace the production system. 

### Related work

Penn State has a feature like this which they rely on heavily for production migrations. Here is an example of a feature like this in a Hyrax 1 application: https://github.com/curationexperts/laevigata/pull/945/files

lfarrell:
Fixes #2877

Adds read only mode feature. It disables all write functionality when turned on. The feature can be turned on/off by going to dashboard->Features

When in read only mode it removes all create/edit buttons/links from view. There are a lot buttons/links in different places. I think I found them all. However, even if a user tries to add/edit a record they should immediately be redirected to the page they came from.

@samvera/hyrax-code-reviewers


dlpierce (commented):
I'm concerned that adding all of this flipflop logic to the various views creates complexity and extra work when modifying or adding features in the future. 

A lighter approach is to keep `check_read_only` and have the read only alert be displayed on all pages.

PR #5161, opened 2021-10-04
make the Collection model class to use configurable

no-reply:
introduce `Hyrax.config.collection_class` to allow applications to choose what
model class to use for collections. many existing hard-coded refernces to
`::Collection`, are kept in this commit.

this is targeted at adding a hook for turning use of the Valkyrie-based
collection class, but it could also be useful for using an ActiveFedora
Collection class with another name (e.g. because your application has some other
model/class/module named `::Collection`).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/collections_controller.rb
+++ b/app/controllers/hyrax/collections_controller.rb
@@ -4,7 +4,9 @@ module Hyrax
     include CollectionsControllerBehavior
     include BreadcrumbsForCollections
     with_themed_layout :decide_layout
-    load_and_authorize_resource except: [:index, :show, :create], instance_name: :collection
+    load_and_authorize_resource except: [:index, :show, :create],
+                                instance_name: :collection,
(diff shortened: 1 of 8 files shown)


mcritchlow (approved):
This could probably be done in `CollectionsMembersController` too at some point,
but limiting the scope to this for now makes sense to me.

no-reply:
> This could probably be done in `CollectionsMembersController` too at some point, but limiting the scope to this for now makes sense to me.

sounds good. it looks like `CollectionMembersController` uses `::Collection.find`. i was think to follow up with a global find/replace of `::Collection` with the variable, followed by some work to make sure all cases where the variable is used can support Valkyrie as appropriate.

how does that sound?

mcritchlow:
That sounds great, and that will then let us drop some of the overrides we currently have in Comet.

PR #5164, opened 2021-10-05
Use blank? instead of nil? for parent_collections in CollectionPresenter

mcritchlow:
The `CollectionPresenter` determines the `total_parent_collections` and
`parent_collection_count` by doing a conditional check on the response
from a call to `parent_collections`

The `nil?` check is problematic because if the `child` is not `nestable?` then
[the response from `NestedCollectionQueryService.parent_collections` is an empty `[]`](https://github.com/samvera/hyrax/blob/main/app/services/hyrax/collections/nested_collection_query_service.rb#L72), not `nil`.

This causes the `total_parent_collections` and `parent_collection_count` methods to try and call `#response` and `#documents` , respectively, against an empty `[]`, throwing errors.

So, change the check to use `blank?`

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/collection_presenter.rb
+++ b/app/presenters/hyrax/collection_presenter.rb
@@ -116,13 +116,13 @@ module Hyrax
 
     # The total number of parents that this collection belongs to, visible or not.
     def total_parent_collections
-      parent_collections.nil? ? 0 : parent_collections.response['numFound']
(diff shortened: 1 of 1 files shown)


no-reply:
seems clearly better, thanks!

no-reply (approved):

PR #5179, opened 2021-10-08
switch call to deprecated presenter method with a helper call

no-reply:
this presenter method was previously deprecated in favor of a helper. it looks
like this view got missed in that update.

@samvera/hyrax-code-reviewers


no-reply:
> I think the keyword argument is misnamed.

oops. thanks!

no-reply:
@dunn @cjcolvar i think i got this one updated to tolerate the test stubbing, and actually improved the coverage a bit in the process.

PR #5207, opened 2021-10-19
centralize the list of collection classes

elrayle:
This centralizes the list of collection classes to `Hyrax.config.collection_classes`.

Places where any collection class could be input or processing occurs across all collection classes were updated to use `Hyrax.config.collection_classes`.  Remaining single references to `::Collection` were updated to `Hyrax.config.collection_class`.  These changes allow for almost all remaining uses of `::Collection` to be removed.

Exceptions:
* work_form which is only used with ActiveFedora
* `PermissionTemplate #collection` which is deprecated and only works with `::Collection`
* `Hyrax::Collections::MigrationService` which performs a migration dependent on reading in the collection as a `::Collection`

This PR does not update `CollectionAbiility` as it is being updated in PR #5206.  Whichever PR is merged first, there will be a small fix required to get abilities to reference the `collection_classes` config instead of creating the list in the `CollectionAbility` class.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/collections_helper.rb
+++ b/app/helpers/hyrax/collections_helper.rb
@@ -25,6 +25,13 @@ module Hyrax
     end
 
     ##
+    # @since 3.1.0
+    # @return [String] the css class for the collection model icon
+    def collection_model_icon
+      Hyrax::ModelIcon.css_class_for(Hyrax.config.collection_class)
+    end
+
+    ##
     # @since 3.0.0
     # @return [#to_s]
     def collection_metadata_label(collection, field)
--- a/app/views/hyrax/collections/_list_collections.html.erb
+++ b/app/views/hyrax/collections/_list_collections.html.erb
@@ -3,7 +3,7 @@
   <td></td>
   <td>
     <div class="media">
-      <span class="<%= Hyrax::ModelIcon.css_class_for(::Collection) %> collection-icon-small pull-left"></span>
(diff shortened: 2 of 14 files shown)


app/views/hyrax/collections/_list_collections.html.erb
```diff
@@ -3,7 +3,7 @@
   <td></td>
   <td>
     <div class="media">
-      <span class="<%= Hyrax::ModelIcon.css_class_for(::Collection) %> collection-icon-small pull-left"></span>
+      <span class="<%= Hyrax::ModelIcon.css_class_for(Hyrax.config.collection_class) %> collection-icon-small pull-left"></span>
```
cjcolvar, line 6: Should this be changed like the change in app/views/hyrax/collections/_media_display.html.erb?
```suggestion
      <span class="<%= collection_model_icon %> collection-icon-small pull-left"></span>
```

app/helpers/hyrax/collections_helper.rb
```diff
@@ -24,6 +24,13 @@ def available_parent_collections_data(collection:)
       end.to_json
     end
 
+    ##
+    # @since 3.1.0
```
cjcolvar, line 28: This should be updated to 3.4.0?
```suggestion
    # @since 3.4.0
```

cjcolvar (commented):
This looks like a good change to me.  Is it still relevant?  How bad of a rebase will it be to bring this up to date?

PR #5222, opened 2021-10-23
refactor NestedCollectionQueryService specs

no-reply:
reduces indirection and repeated setup in tests for NestedCollectionQueryService.

some of the more complicated setup for search builders (as used here) has been
extracted to `FakeSearchBuilderScope` since this was originally written.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/collections/nested_collection_query_service_spec.rb
+++ b/spec/services/hyrax/collections/nested_collection_query_service_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe Hyrax::Collections::NestedCollectionQueryService, clean_repo: tru
   let(:coll_a) do
     FactoryBot.build(:public_collection,
                      id: 'Collection_A',
-                    collection_type: collection_type,
+                     collection_type: collection_type,
                      with_nesting_attributes:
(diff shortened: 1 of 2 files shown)


spec/services/hyrax/collections/nested_collection_query_service_spec.rb
```diff
@@ -214,142 +210,168 @@
...
     end
 
-    describe 'given parent and child are nestable' do
+    describe 'when parent and child are nestable' do
       describe 'and are the same object' do
-        let(:child) { parent }
-
-        it { is_expected.to eq(false) }
+        it 'is false' do
+          expect(described_class.parent_and_child_can_nest?(parent: coll_c, child: coll_c, scope: scope))
```
cjcolvar, line 237: I think this is a typo:
```suggestion
          expect(described_class.parent_and_child_can_nest?(parent: coll_c, child: another, scope: scope))
```

cjcolvar (commented):
@no-reply This looks good to me as far as I can tell.  That's a lot of test refactoring!
I skimmed through it and tried to double check visually and only had one question.

PR #5233, opened 2021-10-29
make ChangeListWriter work when valkyrie collections are setup

no-reply:
@samvera/hyrax-code-reviewers


no-reply:
@elrayle did my answer above address your questions about this change, or is there more to discuss/do here?

PR #5234, opened 2021-10-29
move default admin set id to Hyrax module

elrayle:
This is the first step in moving the default admin set id to a place where it won't have to be regenerated every time we want to check it.

@samvera/hyrax-code-reviewers


elrayle:

PR #5236, opened 2021-10-29
set alternate_identifier instead of id during default adminset creation

elrayle:
For at least some non-wings adapters, you cannot set the id directly.  So need to set the alternate id instead.

@samvera/hyrax-code-reviewers


elrayle:

PR #5237, opened 2021-10-31
set alternate_identifier instead of id during default adminset creation

elrayle:
For at least some non-wings adapters, you cannot set the id directly.  So need to set the alternate id instead.

Also, update methods checking default admin set to use create service methods.

This makes the same changes as PR #5236, but without the dependency of caching the default admin set in the Hyrax module.

@samvera/hyrax-code-reviewers


elrayle:

PR #5240, opened 2021-11-02
Fix admin set valkyrie create

elrayle:
The valkyrie_create! method was passing the original admin_set to`Hyrax::Collections::PermissionsCreateService.create_default`.  This works for ActiveFedora because it updates the original admin_set during save.  For valkyrie, the original admin_set does not have a value for id after it is persisted.  It needs to pass the results of `Hyrax.persister.save` which is the updated admin_set that does have a value for id.

### Expected

Create the new admin set and associated permission template without errors.

### Actual

Creates the new admin set, but gets an error when creating the permission template.

```
     RSolr::Error::Http:
       RSolr::Error::Http - 400 Bad Request
       Error: {
         "responseHeader":{
           "status":400,
           "QTime":1},
         "error":{
           "metadata":[
             "error-class","org.apache.solr.common.SolrException",
             "root-error-class","org.apache.solr.common.SolrException"],
           "msg":"Document is missing mandatory uniqueKey field: id",
           "code":400}}
 ```

### To reproduce:

Run the following test with the first commit that adds the test, but not the fix.

```
bundle exec rspec spec/services/hyrax/admin_set_create_service_spec.rb:157
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/admin_set_create_service.rb
+++ b/app/services/hyrax/admin_set_create_service.rb
@@ -141,7 +141,7 @@ module Hyrax
       updated_admin_set = Hyrax.persister.save(resource: admin_set).tap do |result|
         if result
           ActiveRecord::Base.transaction do
-            permission_template = permissions_create_service.create_default(collection: admin_set,
+            permission_template = permissions_create_service.create_default(collection: result,
                                                                             creating_user: creating_user)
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Good catch!

PR #5242, opened 2021-11-03
Allow work form to display when wings is disabled

elrayle:
### Description

When `Hyrax.config.disable_wings # true` the `Wings` constant is not defined.  This is the correct behavior, but means that any reference to `Wings` in `app` needs to check if `disable_wings #true` before using the `Wings` constant.

This PR addresses a specific instance in `Hyrax::ResourceForm` where there was a check to see if the metadata adapter is wings.

This PR includes
* update existing non-wings test to remove the `Wings` constant so that it provides a failing test for the scenario described in this PR
* code fix to prevent the exception

### Expected

Dashboard -> Works -> Add new work -> displays the work form

### Actual

Dashboard -> Works -> Add new work -> raises an exception

image

### Related Work

There are other places where `Wings` is used without checking `disable_wings #true`.  A cursory review indicates that these locations are places not expected to be encountered when using a non-wings adapter for Valkyrie (e.g. actor stack code, AF forms).  It is possible that other locations may be uncovered later that would need a similar approach as applied in this PR.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -53,16 +53,15 @@ module Hyrax
       #   we want to move away from application side lock validation and rely
       #   on the adapter/database features instead.
       LockKeyPopulator = lambda do |_options|
-        self.version =
-          case Hyrax.metadata_adapter
-          when Wings::Valkyrie::MetadataAdapter
+        if Hyrax.config.disable_wings || !Hyrax.metadata_adapter.is_a?(Wings::Valkyrie::MetadataAdapter)
+          Hyrax.logger.info "trying to prepopulate a lock token for " \
+                            "#{self.class.inspect}, but optimistic locking isn't " \
+                            "supported for the configured adapter: #{Hyrax.metadata_adapter.class}"
+          self.version = ''
+        else
+          self.version =
             model.persisted? ? Wings::ActiveFedoraConverter.convert(resource: model).etag : ''
-          else
-            Hyrax.logger.info 'trying to prepopulate a lock token for ' \
-                              "#{self.class.inspect}, but optimistic locking isn't " \
-                              "supported for the configured adapter: #{Hyrax.metadata_adapter.class}"
-            ''
-          end
+        end
       end
 
       class_attribute :model_class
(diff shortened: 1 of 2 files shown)


spec/forms/hyrax/forms/resource_form_spec.rb
```diff
@@ -384,7 +384,13 @@
     end
 
     context 'when using a generic valkyrie adapter', valkyrie_adapter: :test_adapter do
+      before do
+        allow(Hyrax).to receive_message_chain(:config, :disable_wings).and_return(true) # rubocop:disable RSpec/MessageChain
+        hide_const("Wings") # disable_wings=true removes the Wings constant
+      end
```
elrayle, line 390: It already has a test using a non-wings adapter, but it doesn't fully simulate disabled wings.  When it is disabled, the `Wings` constant doesn't exist.  The new before block simulates `disable_wings` returning `true` and the `Wings` constant not existing.

spec/forms/hyrax/forms/resource_form_spec.rb
```diff
@@ -384,7 +384,13 @@
...
       it 'prepopulates as empty before save' do
+        expect(Hyrax.logger).to receive(:info)
+          .with(starting_with("trying to prepopulate a lock token for"))
```
elrayle, line 393: It isn't a change that the logger adds this info message.  This expect line just confirms the message was logged.

app/forms/hyrax/forms/resource_form.rb
```diff
@@ -53,16 +53,15 @@ class ResourceForm < Hyrax::ChangeSet # rubocop:disable Metrics/ClassLength
       #   we want to move away from application side lock validation and rely
       #   on the adapter/database features instead.
       LockKeyPopulator = lambda do |_options|
-        self.version =
-          case Hyrax.metadata_adapter
-          when Wings::Valkyrie::MetadataAdapter
+        if Hyrax.config.disable_wings || !Hyrax.metadata_adapter.is_a?(Wings::Valkyrie::MetadataAdapter)
```
elrayle, line 56: The `case` statement didn't allow for checking `disable_wings` before trying to use the `Wings` constant.  The change is a simple rewrite of the `case` into an `if` to allow for the more complex check.  The code that is executed for each case is the same.

dlpierce (approved):

PR #5243, opened 2021-11-04
Use file_set.id instead of file_set in can?

### LINKED ISSUE #5241: can?(:download, file_set) always returns false (display_media_download_link? in Hyrax::FileSetHelper)
https://github.com/samvera/hyrax/blob/97f45609c208dab8d93ce0597f90e39c519a41e2/app/helpers/hyrax/file_set_helper.rb#L12

This call to `can?(:download, file_set)` always returned 'false' in our testing, which was preventing the download link from appearing below thumbnails on work pages.

I believe the call should be rewritten to `can?(:download, file_set.id)`, like in `app/views/hyrax/file_sets/_actions.html.erb`.

cudevmaxwell:
Fixes #5241;

I think the call to `can?` in `display_media_download_link?` should use `file_set.id`, not `file_set`. This PR makes that change.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/file_set_helper.rb
+++ b/app/helpers/hyrax/file_set_helper.rb
@@ -9,7 +9,7 @@ module Hyrax::FileSetHelper
   #   set
   def display_media_download_link?(file_set:)
     Hyrax.config.display_media_download_link? &&
-      can?(:download, file_set) &&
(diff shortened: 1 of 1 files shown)


cudevmaxwell:
Failing tests, closing.

PR #5249, opened 2021-11-05
generated indexer test requires resource variable

elrayle:
`it_behaves_like 'a Hyrax::Resource indexer'` requires `:indexer_class` and `:resource` to be defined for tests to run without error.  The work_resource generator did not set `:resource`.  This PR adds the definition so that generates work resource tests pass out of the box.

@samvera/hyrax-code-reviewers


no-reply (approved):

no-reply:
thanks for this addition!

PR #5253, opened 2021-11-09
Adds workflow to trigger nurax deploy on push

mlooney:
@samvera/hyrax-code-reviewers

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,17 @@
+name: Trigger Nurax build
+on:
+  workflow_dispatch:
+  push:
+
+jobs:
+  trigger:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: peter-evans/repository-dispatch@v1
+      with:
+         token: ${{ secrets.NURAX_ACCESS_TOKEN }}
+         event-type: push
+         repository: curationexperts/nurax
+         client-payload: '{"ref": "${{ github.ref }}", "sha": "${{ github.sha }}"}'
+
+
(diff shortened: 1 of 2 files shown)


no-reply:
note that this adds a new use of GitHub Actions.

is this something we're open to?

dlpierce:
What is the concern? If GH Actions were to become non-free for public repos in the future?

no-reply:
> What is the concern? If GH Actions were to become non-free for public repos in the future?

i think it's mainly just that it's one more pipeline to configure and monitor, and one more configuration format to manage.

i don't have any other github actions projects, so personally i would probably avoid doing maintenance on this.

elrayle:
For context, the purpose of this is to provide an auto-deploy to the DCE supported `nurax-dev` and `nurax-stable` test applications.  It connects into github [actions](https://github.com/curationexperts/nurax/actions) in [curationexperts/nurax](https://github.com/curationexperts/nurax).  Without these, someone would have to manually deploy [nurax-dev](https://nurax-dev.curationexperts.com/) and [nurax-stable](https://nurax-dev.curationexperts.com/). 

I am in favor of merging this PR.

elrayle:
@no-reply Are you open to having this merged?

elrayle:
Discussed today in Tech Call and there are no blockers to this being merged.  There was a question about who would maintain this pipeline.  There are several folks, including DCE staff, that work with Github Actions.  So it seems like there are members of the community that can help maintain this pipeline.  

It is also worth noting that this is not a critical pipeline.  If it fails, it only impacts nurax related deploys.  Additionally there are manual steps to deploy that would address the failure.

elrayle (approved):
Thanks for adding this.  It will be nice to have the auto-deploys working.

PR #5255, opened 2021-11-11
mark feature test for AF work types pending

### LINKED ISSUE #5248: Remove generated feature tests for AF based work types
### Current Challenge

The feature spec generated when a new work type is generated using `rails generate hyrax:work GenericWork` fails out of the box.  I've tried several things to get the test to work by comparing the Hyrax feature spec `spec/features/create_work_spec.rb` to the generated spec for the new work type without success.

### Proposed Modification

I would like to propose that we not generate the feature tests.  These seem well tested by Hyrax's feature tests.  It seems like applications should be able to trust that the Hyrax feature testing process is sufficient unless they make extensive customizations.  In that case, the generated feature test would also need extensive modifications to test the customizations. 

Looking for others opinions on this.

elrayle:
Issue #5248 recommended removal of the failing feature test generated for work types.  Further discussion recommended marking it pending.

Minimally, users need to uncomment some lines if they generate more than one work type.  To prevent this test from failing out-of-the-box, it is marked pending.

Fixes #5248

_NOTE: This PR and description originally were removing the tests.  Initial comments on this PR are related to the possibility of removing the tests.  As recommended in those comments, this PR was refactored to mark the tests pending instead._

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/work/templates/feature_spec.rb.erb
+++ b/lib/generators/hyrax/work/templates/feature_spec.rb.erb
@@ -31,11 +31,13 @@ RSpec.feature 'Create a <%= class_name %>', js: false do
     end
 
     scenario do
+      pending 'Changes may be required for this test to pass.  See TODO in test.'
+
       visit '/dashboard'
       click_link "Works"
       click_link "Add new work"
 
-      # If you generate more than one work uncomment these lines
+      # TODO: If you generate more than one work uncomment these lines
       # choose "payload_concern", option: "<%= class_name %>"
       # click_button "Create work"
 


no-reply:
in other cases, we generate a pending test even if no actual test bodies are provided.

i'm agnostic about whether that's the best approach here, but i wanted to make sure it was considered as a possibility before pushing this through.

dlpierce:
I'm in favor of having skipped stub feature specs so application developers are reminded they exist.

cjcolvar (approved):
Looks good!

PR #5256, opened 2021-11-12
Prepare release 3.2.0

dlpierce:
@samvera/hyrax-code-reviewers


no-reply (approved):
i think we should aim to adjust the release notes to look a bit more like the other recent ones: emphasizing changes that are most likely to be meaningful to adopters, and moving straight refactors and dev env/ci/chart changes toward the bottom.

let's do get the release out the door, though.

PR #5263, opened 2021-12-01
persist the default admin set id in the database

### LINKED ISSUE #5261: create default admin set for non-wings adapters
### Descriptive summary

There needs to be a process for generating the default admin set and remembering it.  Historically this was done by assigning `"DEFAULT_ADMIN_SET"` as the `id` of the admin set.  But other adapters, like postgres, do not allow directly setting of `id`.

### Rejected approaches

Several approaches have been tried:
* set ID to predetermined value (current approach) - doesn't work for some adapters (e.g. postgres)
* set alternate_ids to be the predetermined value - doesn't work when NOIDs are turned on because they are saved in as the alternate id
* set title to be a predetermined value - doesn't work because query by title isn't supported

### Proposed approach

Allow the adapter to save the admin set and then store the generated ID in permanent storage.

Basic functioning of this approach:
* first admin set created will be assigned as the default
* store id of default permanently-- where? perhaps in database
* read default from from permanent store and cache in Hyrax.config.default_admin_set_id
* set default using Hyrax.config.default_admin_set_id= and update permanent store
* change which admin set is default (any admin set can be the default) - via UI?  

### Alternate simplified approach
* create an admin set
* edit `configs/initializers/hyrax.rb` and set `default_admin_set_id`
* restart app
* Hyrax.config.default_admin_set_id will be read from hyrax initializer

elrayle:
Fixes #5261 

This also makes the default admin set and id accessible through `Hyrax.config`.  The first time it is accessed, the default admin set will be created if it doesn’t already exist.  If the default admin set id saved in the database changes, it will reset the instance variable in `Hyrax.config` that caches the default admin set.

### Final Design

The final design is somewhat different than the one proposed in #5261.  

It addresses challenges:

* for existing apps, it will use the existing default admin set identified by DEFAULT_ID
* allow the default admin set to be created with an assigned id (required for postgres adapter)
* cache the default admin set and id so it doesn't have to be retrieved from the persistent store every time it is referenced
* saves the id in the database so it is remembered between restarts of the system and other events that will wipe the cached version
* the plumbing to allow for the identified default admin set to be changed is in place, but not surfaced in the UI.  This is prep for having this functionality in the future.

Supporting changes:

* persistance of the default admin set id is handled by model `Hyrax::DefaultAdministrativeSet` which saves the default admin set id in the database
* runtime caching is handled by `Hyrax.config.default_admin_set` which caches the default admin set in @default_admin_set instance variable the first time it is accessed.  It can be forced to update by calling `Hyrax.config.reset_default_admin_set.  `Hyrax.config.default_admin_set_id` provides a shortcut to the default admin set's id.
* `Hyrax::AdminSetCreateService.find_or_create_default_admin_set` uses `Hyrax::DefaultAdministrativeSet` to make sure the id is persisted.  This maintains this method as "the" way to create the default admin set.  Using this method, the persisted and cached versions of the default admin set will be refreshed if needed.

@samvera/hyrax-code-reviewers


elrayle:
I tested this with nurax-pg with a freshly migrated database for both checks.

### Check 1: creating with the rake task 

This is the historical way to create default admin set for Hyrax apps.

```
$ bin/rails hyrax:default_admin_set:create
Successfully created default admin set
```

Confirm in rails console...

```
>> Hyrax::DefaultAdministrativeSet.first.default_admin_set_id
=> "5418df25-4cbd-4e2a-8bc7-c85e135bd5fc"
>> Hyrax::AdminSetCreateService.find_or_create_default_admin_set.id.to_s
=> "5418df25-4cbd-4e2a-8bc7-c85e135bd5fc"
>> Hyrax.config.default_admin_set_id
=> "5418df25-4cbd-4e2a-8bc7-c85e135bd5fc"
```

### Check 2: creating first time accessed when one doesn't exist

_Cleared db before running second check._

Create and confirm in rails console...

```
>> Hyrax::DefaultAdministrativeSet.first
=> nil
>> Hyrax::AdminSetCreateService.find_or_create_default_admin_set.id
=> "7f0e2948-a90b-47ca-8871-424b81b07cb8"
>> Hyrax::DefaultAdministrativeSet.first.default_admin_set_id.to_s
=> "7f0e2948-a90b-47ca-8871-424b81b07cb8"
>> Hyrax.config.default_admin_set_id
=> "7f0e2948-a90b-47ca-8871-424b81b07cb8"
```

dunn (approved):

PR #5265, opened 2021-12-03
make the member presenter factory for WorkShowPresenter configurable

no-reply:
the default implementation of `WorkShowPresenter#member_presenter_factory` had a
hard coded dependency on `MemberPresenterFactory`, which depended on
`AF::Base#list_source`. an alteranative implementation at
`PcdmMemberPresenterFactory` already exists and is in use in the
`form_file_set_select_for` view helper, as well as a drop-in replacement for
`WorkShowPresenter#member_presenter_factory` in Surfliner's `comet`. this
override can now be set by doing `WorkShowPresenter.presenter_factory_class =
PcdmMemberPresenterFactory` (e.g.) in an initializer.

the factory for member presenters was already settable on the instance level,
but the more common configuration need is to have a site-wide override.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/pcdm_member_presenter_factory.rb
+++ b/app/presenters/hyrax/pcdm_member_presenter_factory.rb
@@ -19,7 +19,7 @@ module Hyrax
     ##
     # @param [#member_ids] object
     # @param [::Ability] ability
-    def initialize(object, ability)
+    def initialize(object, ability, _request = nil)
       @object = object
       @ability = ability
     end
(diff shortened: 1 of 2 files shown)


cjcolvar:
This looks good.  Is there any reason why `Hyrax::PcdmMemberPresenterFactory` shouldn't be the default implementation?

dunn (approved):

no-reply:
> Is there any reason why `Hyrax::PcdmMemberPresenterFactory` shouldn't be the default implementation?

i'm not totally sure it's feature complete. the existing implementation uses the request context, and passes it down to the built presenters. i haven't yet followed up on whether that's necessary and what implications of removing it are.

getting this in as configuration, and then a new setting into `nurax-pg` should help identify any issues that reach down to users.

PR #5266, opened 2021-12-03
fix syntax in .env; no empty assignments

no-reply:
@samvera/hyrax-code-reviewers

### DIFF
--- a/.env
+++ b/.env
@@ -24,5 +24,5 @@ SOLR_URL=http://solr:8983/solr/hyrax
 VALKYRIE_SOLR_PORT=8983
 VALKYRIE_SOLR_HOST=solr
 HYRAX_ANALYTICS=false # change to true to enable analytics
-HYRAX_ANALYTICS_PROVIDER=  #options are 'matomo' or 'google'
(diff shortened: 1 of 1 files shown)


dunn (dismissed):

.env
```diff
@@ -24,5 +24,4 @@ SOLR_URL=http://solr:8983/solr/hyrax
 VALKYRIE_SOLR_PORT=8983
 VALKYRIE_SOLR_HOST=solr
 HYRAX_ANALYTICS=false # change to true to enable analytics
```
orangewolf, line 26:
```suggestion
HYRAX_ANALYTICS=false # change to true to enable analytics
HYRAX_ANALYTICS_PROVIDER=google  # options are 'matomo' or 'google', 'google' is the default if not specified.
```

dlpierce:
@orangewolf The last commit seems to have a bunch of changes that are not bumping the circleci cache. What's going on there?

no-reply:
@dlpierce i force pushed with a rebase on current `main`. hopefully that clears things up?

dunn (approved):

Issue #5130, opened 2021-09-23
Val MVP: Create AdministrativeSet as a valkyrie resource through UI


elrayle:
### Descriptive summary

As part of the Valkyrization MVP, a user should be able to create an admin set through the UI with the underlying code processing the admin set as a `Hyrax::AdministrativeSet`.

### Rationale

Part of the basic Valkyrization MVP for testing the progress of the valkyrization effort.

### Steps to reproduce the behavior

1. Login to [nurax-pg](https://nurax-pg.curationexperts.com)
1. Dashboard -> [Collections](https://nurax-pg.curationexperts.com/dashboard/my/collections?locale=en) -> New Collection
2. Select Admin Set
2. Fill in required metadata
3. Save

The admin set should be created and listed in the Dashboard -> Collections -> Your Collections.

### Related work

See all work tagged with Milestone [Valkyrization MVP](https://github.com/samvera/hyrax/milestone/19)

rjkati:
I am able to complete the steps above to create an admin set on nurax-pg. However, when I click "Save" in the final step, I get a "NoMethodError in Hyrax::Admin::AdminSetsController#create " message.

URL: https://nurax-pg.curationexperts.com/admin/admin_sets?locale=en
![nomethoderror](https://user-images.githubusercontent.com/32644254/140406321-82c24fba-79f6-4324-a745-1acf66cbbe8b.PNG)

rjkati:
Admin sets can be created on nurax-pg in Hyrax 3.4.1

jlhardes:
Closing this issue since admin sets can now be created.

PR #5268, opened 2021-12-03
add transactions for creating an admin set

elrayle:
Partially addresses Issue #5130 - Val MVP: Create AdministrativeSet as a valkyrie resource through UI

----

This adds transactions used when creating a `Hyrax::AdministrativeSet` that is a valkyrie resource.  This will be used for the `#create` action in the `Hyrax::Admin::AdminSetsController`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/admin_set_create.rb
+++ b/lib/hyrax/transactions/admin_set_create.rb
@@ -4,7 +4,7 @@ require 'hyrax/transactions/transaction'
 module Hyrax
   module Transactions
     ##
-    # Creates a Collection from a ChangeSet
+    # Creates a Hyrax::AdministrativeSet from a ChangeSet
     #
     # @since 3.2.0
     class AdminSetCreate < Transaction
--- a/lib/hyrax/transactions/admin_set_create.rb
+++ b/lib/hyrax/transactions/admin_set_create.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+require 'hyrax/transactions/transaction'
+
+module Hyrax
+  module Transactions
+    ##
+    # Creates a Collection from a ChangeSet
+    #
+    # @since 3.2.0
+    class AdminSetCreate < Transaction
+      DEFAULT_STEPS = ['change_set.apply',
+                       'admin_set_resource.apply_collection_type_permissions',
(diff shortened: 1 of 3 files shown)


lib/hyrax/transactions/admin_set_create.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+require 'hyrax/transactions/transaction'
+
+module Hyrax
+  module Transactions
+    ##
+    # Creates a Collection from a ChangeSet
```
cjcolvar, line 7:
```suggestion
    # Creates an AdminSet from a ChangeSet
```

cjcolvar (dismissed):
This looks good!  I just had one suggestion to clarify what the transaction is creating.

I started that suggestion with `AdminSet` then changed it to `AdministrativeSet` which got me wondering about the potential disconnect in naming between AdminSet and AdministrativeSet.  Is the approach to keep naming classes using "admin set" to match the existing names (e.g. `AdminSetCreate` and the `admin_set_resource` transaction container namespace) except for the `AdministrativeSet` model iteself?

lib/hyrax/transactions/admin_set_create.rb
elrayle: Nice catch!

elrayle:
RE: admin set vs administrative set - There are a lot of places in the code that will not be reproduced for `Hyrax::AdministrativeSet`.  The code is blending handling of both in the same class.  As a result, admin set has been used as a generic short hand for either.  Not ideal as it can create some confusion.  But it would be a challenging refactor to do everywhere.  That said, I am open to using the term administrative set when ever possible to at least minimize the confusion.  

We see the same thing happening for collections where the term collection is used as a generic term for both `::Collection` and `Hyrax::PcdmCollection` and any other class that may get created for collections.  I think it would be more difficult to use the full name for collections because the class isn't restricted to just those two classes.

elrayle:
While looking at a refactor to use `administrative_set` instead of `admin_set`, I noticed that other transactions use the abbreviated `admin_set` (i.e. ensure_admin_set, set_default_admin_set and their associated steps).

cjcolvar (approved):

PR #5277, opened 2021-12-07
Fix working_directory_spec for docker-compose

marrus-sh:
Previously this test hardcoded the working path as `/tmp/uploads/`, which caused it to fail in docker-compose situations. This commit fixes it to simply use `Hyrax.config.working_path`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/working_directory_spec.rb
+++ b/spec/services/hyrax/working_directory_spec.rb
@@ -6,8 +6,8 @@ RSpec.describe Hyrax::WorkingDirectory do
   describe "#full_filename" do
     it "generates unique filenames for different files" do
       expect(path1).not_to eq(path2)
-      expect(path1).to match(/\/tmp\/uploads\/ab\/cd\/ef\/gh\/abcdefghijklmnop1\/foo.tif$/)
-      expect(path2).to match(/\/tmp\/uploads\/ab\/cd\/ef\/gh\/abcdefghijklmnop2\/foo.tif$/)
+      expect(path1).to eq("#{ Hyrax.config.working_path }ab/cd/ef/gh/abcdefghijklmnop1/foo.tif")
+      expect(path2).to eq("#{ Hyrax.config.working_path }ab/cd/ef/gh/abcdefghijklmnop2/foo.tif")
     end
   end
 end


spec/services/hyrax/working_directory_spec.rb
```diff
@@ -6,8 +6,8 @@
   describe "#full_filename" do
     it "generates unique filenames for different files" do
       expect(path1).not_to eq(path2)
-      expect(path1).to match(/\/tmp\/uploads\/ab\/cd\/ef\/gh\/abcdefghijklmnop1\/foo.tif$/)
-      expect(path2).to match(/\/tmp\/uploads\/ab\/cd\/ef\/gh\/abcdefghijklmnop2\/foo.tif$/)
+      expect(path1).to eq("#{ Hyrax.config.working_path }ab/cd/ef/gh/abcdefghijklmnop1/foo.tif")
```
no-reply, line 9:
```suggestion
      expect(path1).to eq("#{Hyrax.config.working_path}ab/cd/ef/gh/abcdefghijklmnop1/foo.tif")
```

spec/services/hyrax/working_directory_spec.rb
```diff
@@ -6,8 +6,8 @@
...
+      expect(path2).to eq("#{ Hyrax.config.working_path }ab/cd/ef/gh/abcdefghijklmnop2/foo.tif")
```
no-reply, line 10:
```suggestion
      expect(path2).to eq("#{Hyrax.config.working_path}ab/cd/ef/gh/abcdefghijklmnop2/foo.tif")
```

no-reply:
rubocop is mad about spaces. i made some suggestions that i think will fix the build.

marrus-sh:
Right; bad habits on my part from other languages. Should have linted first 😅.

I’ve pushed the changes; hopefully they can be squashed on merge?

marrus-sh:
Alas! It looks like `Hyrax.config.working_path` has a trailing slash in `docker-compose` but not in CircleCI. Should be able to get around this using `File.join` (which is what the relevant `Hyrax::WorkingDirectory` method uses); done in the latest commit.

no-reply:
cla is cleared.

no-reply (approved):

PR #5278, opened 2021-12-08
characterize valkyrie file uploads

dunn:
this adds characterization metadata to `Hyrax::FileMetadata` for whatever setters it has defined that match the terms extracted by fits.

this is accomplished by
- adding a new `ValkyrieUploadsHandler` to circumvent the current pipeline that makes many AF assumptions (this means that the Valkyrie logic in `JobIOWrapper` is dead code now)
- adding a new event `object.file.uploaded` (distinct from `object.metadata.updated` because characterization should not run every time metadata is chaged)
- adding a new listener that runs  a new `ValkyrieCharacterizationService` to execute fits and attach the metadata (currently runs synchronously, but making it run as a job may be added as acceptance criteria)

this does not yet teach Hyrax how to display characterization information in the UI.

@samvera/hyrax-code-reviewers


dunn:
closing while i rethink this a bit

PR #5279, opened 2021-12-08
handle the case when table for saving default admin set doesn’t exist

elrayle:
This removes the requirement to install the migration CreateDefaultAdministrativeSet.  This allows a release at the minor level.

@samvera/hyrax-code-reviewers


dunn (approved):

no-reply (approved):
looks great!

it might be nice to follow up with a `warn` notice that the migrations aren't installed, so folks are more likely to get to it prior to 4.0.0. that could be an initializer?

either way, this looks ready!

PR #5280, opened 2021-12-08
WIP: try a minimal upgrade to Blacklight 7

no-reply:
gratefully, this is working well enough to load the hyrax homepage. forcing a
build via CI.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/Gemfile
+++ b/.dassie/Gemfile
@@ -53,7 +53,7 @@ gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 gemspec name: 'hyrax', path: ENV.fetch('HYRAX_ENGINE_PATH', '..')
 
 gem 'rsolr', '>= 1.0', '< 3'
-gem 'bootstrap-sass', '~> 3.0'
+gem 'bootstrap', '~> 4.0'
 gem 'twitter-typeahead-rails', '0.11.1.pre.corejavascript'
 gem 'jquery-rails'
 gem 'devise'
--- a/.dassie/app/assets/stylesheets/hyrax.scss
+++ b/.dassie/app/assets/stylesheets/hyrax.scss
@@ -4,14 +4,7 @@
  *= require_self
 */
 
-@import "bootstrap-sprockets";
 @import "bootstrap-default-overrides";
-@import 'bootstrap';
-@import 'blacklight/blacklight';
 @import "font-awesome";
-@import "blacklight_gallery/gallery";
-@import "blacklight_gallery/masonry";
-@import "blacklight_gallery/slideshow";
-@import "blacklight_gallery/osd_viewer";
-@import 'hyrax/hyrax';
-@import 'hyrax/login_signup';
+// @import 'hyrax/hyrax';
+// @import 'hyrax/login_signup';
--- a/.dassie/app/assets/stylesheets/hyrax.scss
+++ b/.dassie/app/assets/stylesheets/hyrax.scss
@@ -5,6 +5,11 @@
 */
 
 @import "bootstrap-default-overrides";
+@import 'bootstrap';
 @import "font-awesome";
-// @import 'hyrax/hyrax';
-// @import 'hyrax/login_signup';
+@import "blacklight_gallery/gallery";
+@import "blacklight_gallery/masonry";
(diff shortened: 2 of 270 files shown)


marrus-sh:
Here’s how things currently look :—

### Fixable prior to Blacklight 7

- [x] One test used the deprecated name `InvalidSolrID` instead of the current name `RecordNotFound`. This was fixed in #5289.
- [ ] Blacklight 7 moved `SearchHelper` things off of `Blacklight::Controller` and introduced a `Searchable` mixin to replace it. However, `Searchable` doesn’t work as well for our purposes as they currently stand, and we depend on old Blacklight 6 methods being present in our controllers. #5288 compensates for this and should work in both 6 and 7.
- [ ] Use a fake instead of doubles in some Blacklight tests: d68b5c9, ed14386, 7697185 (no PR yet).
- [ ] One test hardcoded the list of Blacklight filters, which will change in Blacklight 7. This is fixed in 39fc6b3 (no PR yet).
- [ ] One test (mistakenly?) used `*_tesim` instead of `*_sim`, which worked in Blacklight 6 but won’t in Blacklight 7. This is fixed in ef519cb (no PR yet).
- [ ] The Catalog `_index_test_default` test used unrealistic values, which causes problems for Blacklight 7. This is fixed in c24897a (no PR yet).
- [ ] Blacklight 7 `SearchBuilder`s expect to be constructed with an object with a `blacklight_config`, so this needs to be added to `RestrictionService`. This is fixed in 0944a7e (no PR yet).

### Known fixes for Blacklight 7

- [ ] Basic setup: 5591fa7, 449b8b5, 0c7d7e4, 4427c8e, b99c4f7, 659740b.
- [ ] Blacklight 7 moved some things off of controllers and into `blacklight_config`. A temporary fix for this was added in 497eeec and 923bb1a; see the `bl-controller-compatibility-pass-through` branch.
    - Commits 40f6ea7 and 2e0577e attempt to solve the same thing and can probably be dropped.
- [ ] Blacklight 7 no longer initializes Solr parameters to `[]`, so they may be `nil`. There are also dedicated methods (I think also present in Blacklight 6?) for interacting with Solr parameters which are probably better than the approach we are currently using. Fixes for this are in a0fa897, 147bdb4 (& fa86228), and 610fa7c, and there are a few in ab8d245 as well.

### Outstanding problems not related to the upgrade

A number of tests fail in Docker for reasons unrelated to Blacklight 7. Consequently, it’s impossible to say whether there are also Blacklight 7 issues here or not.

- [ ] `characterize_job_spec` hardcodes a `tmp/` directory which `docker-compose` doesn’t use (issue filed at #5290).
- [ ] The tests for `Hyrax::UserStatImporter` fail for an unknown reason (no issue/PR yet).
- [ ] Selenium fails on a number of tests in `spec/features` with an error of `Illegal key values seen in w3c capabilities: [chromeOptions]` (no issue/PR yet).
- [ ] Jasmine tests fail with an error of `sh: phantomjs: not found` (no issue/PR yet).

### Outstanding issues with Blacklight 7

- [ ] Authorities are messed up? Work for this was begun in 44ee0c5 but I’m still getting authority tests failing.
- [ ] The tests in `spec/views/hyrax/base/_member.html.erb_spec` and `spec/views/hyrax/base/file_manager.html.erb_spec.rb` give the following error: ``Unable to find track_test_path route helper. Did you add `concerns :searchable` routing mixin to your `config/routes.rb`?``. The source of this seems to be projectblacklight/blacklight@b89b8cc but I don’t know what to do about it.
- [ ] I don’t know how to approach the following test failures :—
    - [ ] `spec/helpers/blacklight_helper_spec.rb`
    - [ ] `spec/services/hyrax/collections/nested_collection_query_service_spec.rb`
    - [ ] `spec/views/hyrax/admin/admin_sets/show.html.erb_spec.rb`
    - [ ] `spec/views/hyrax/dashboard/collections/show.html.erb_spec.rb`

### Open questions

- Anything frontend‐related 😅.

marrus-sh:
There are a *lot* of variables to rename in the SCSS (working through them to see if I can get something to load); aside from that, I think #5292 is presently the biggest blocker on the frontend side of things.

marrus-sh:
Future work is probably going to take place on the `bl7bootstrap4` branch instead of here :)

PR #5285, opened 2021-12-10
dassie: update db schema

dunn:
looks like the new migrations haven't been incorporated by the dassie schema; this is what i get after a fresh `docker-compose up`

@samvera/hyrax-code-reviewers


no-reply:
@elrayle this looks right to me, but i wanted to check in with you since these migrations originated with your work.

no-reply (approved):

PR #5288, opened 2021-12-11
add Blacklight::Searchable-like behavior to `Hyrax::Controller`

no-reply:
set the search service, and provide compatibility between Hyrax's use of
`search_results` and Blacklight 7's `Searchable` mixin.

provides a measure of forward compatibility with Blacklight 7 controller style.

@samvera/hyrax-code-reviewers


cjcolvar (commented):
This looks good to me but there is a conflict that needs to be resolved.

marrus-sh:
^ rebased on top of main

cjcolvar (approved):

PR #5294, opened 2021-12-15
wings: hard-code AF::File.default_sort_params

dunn:
fixes https://github.com/samvera/active_fedora/issues/1459

@samvera/hyrax-code-reviewers


no-reply:
this is related to https://github.com/samvera/active_fedora/pull/1460

no-reply (approved):

PR #5303, opened 2021-12-17
add basic characterization of Valkyrie file uploads

dunn:
this adds `ValkyrieCharacterizationService`, based on the hydra-works service class, to allow characterizing file uploads of Valkyrie objects.

the attributes/setters haven't been modified, so currently this just assigns values for those which match the attributes in the fits response, like `height` and `width`:
```
[1] pry(Hyrax::Characterization::ValkyrieCharacterizationService)> object
=> #<Hyrax::FileMetadata id=#<Valkyrie::ID:0x00007f4683d82680 @id="test_id"> internal_resource="Hyrax::FileMetadata" created_at=nil updated_at=nil new_record=true file_identifier=nil alternate_ids=[] file_set_id=nil label=["world.png"] original_filename="world.png" mime_type="image/png" type=[#<RDF::Vocabulary::Term:0x4dff8 ID:http://pcdm.org/use#OriginalFile>] format_label=["test_format_label", "Portable Network Graphics"] size=[] well_formed=[] valid=[] date_created=[] fits_version=[] exif_version=[] checksum=[] frame_rate=[] bit_rate=[] duration=[] sample_rate=[] height=["50"] width=["50"] bit_depth=[] channels=[] data_format=[] offset=[] file_title=[] creator=[] page_count=[] language=[] word_count=[] character_count=[] line_count=[] character_set=[] markup_basis=[] markup_language=[] paragraph_count=[] table_count=[] graphics_count=[] byte_order=[] compression=["Deflate/Inflate"] color_space=[] profile_name=[] profile_version=[] orientation=[] color_map=[] image_producer=[] capture_device=[] scanning_software=[] gps_timestamp=[] latitude=[] longitude=[] aspect_ratio=[]>
```

the characterization status is also not yet reflected in the ui, since the valkyrie presenter does not use `Hyrax::CharacterizationBehavior`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/valkyrie_ingest_job.rb
+++ b/app/jobs/valkyrie_ingest_job.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+class ValkyrieIngestJob < Hyrax::ApplicationJob
+  queue_as Hyrax.config.ingest_queue_name
+
+  # after_perform do |job|
+  #   # We want the lastmost Hash, if any.
+  #   opts = job.arguments.reverse.detect { |x| x.is_a? Hash } || {}
+  #   wrapper = job.arguments.first
+  #   ContentNewVersionEventJob.perform_later(wrapper.file_set, wrapper.user) if opts[:notification]
+  # end
+
+  # @param [Valkyrie::StorageAdapter::StreamFile] file
+  # @param [Boolean] notification send the user a notification, used in after_perform callback
+  # @see 'config/initializers/hyrax_callbacks.rb'
+  def perform(file, _notification: false)
+    ingest(file: file)
+  end
+
+  # @param [Valkyrie::StorageAdapter::StreamFile] file
+  # @return [void]
+  def ingest(file:)
+    file_set = Hyrax.query_service.find_by(id: file.file_set_uri)
+    file_metadata = Hyrax::FileMetadata.for(file: file.uploader.file)
+
+    updated_metadata = upload_file(file: file, file_metadata: file_metadata, file_set: file_set)
+    add_file_to_file_set(file_set: file_set, file_metadata: updated_metadata)
+  end
+
+  # @return FileSet updated file set
+  def add_file_to_file_set(file_set:, file_metadata:)
+    file_set.file_ids << file_metadata.id
+    Hyrax.persister.save(resource: file_set)
+  end
+
+  # @param [Hyrax::UploadedFile] file
+  # @param [Hyrax::FileMetadata] file_metadata
+  # @param [Hyrax::FileSet] file_set
+  # @return Hyrax::FileMetadata uploaded file
+  def upload_file(file:, file_metadata:, file_set:)
+    uploader = file.uploader
+    file_metadata.file_set_id = file.file_set_uri
+    uploaded = Hyrax.storage_adapter
+                    .upload(resource: file_set,
+                            file: File.open(uploader.file.file),
+                            original_filename: file_metadata.original_filename)
+    file_metadata.file_identifier = uploaded.id
+    file_metadata.size = uploaded.size
+
+    Hyrax.publisher.publish(
+      "object.file.uploaded",
+      metadata: file_metadata
+    )
+
+    file_metadata
+  end
+end
--- a/app/services/hyrax/characterization/valkyrie_characterization_service.rb
+++ b/app/services/hyrax/characterization/valkyrie_characterization_service.rb
@@ -3,10 +3,9 @@ require 'hydra-file_characterization'
 require 'nokogiri'
 
 class Hyrax::Characterization::ValkyrieCharacterizationService
-  # @param [Hyrax::FileMetadata] object which has properties to recieve characterization values.
+  # @param [Hyrax::FileMetadata] object which has properties to recieve characterization values
   # @param [Valkyrie::StorageAdapter::StreamFile] source for characterization to
-  # be run on.  File object or path on disk.  If none is provided, it will
-  # assume the binary content already present on the object.
+  # be run on
   # @param [Hash] options to be passed to characterization.  parser_mapping:, parser_class:, tools:
   #
   # @return [Hash]
@@ -40,11 +39,7 @@ class Hyrax::Characterization::ValkyrieCharacterizationService
 
   protected
 
-  # @return content of object if source is nil; otherwise, return a File or the source
   def source_to_content
-    return object.file if source.nil?
-    # do not read the file into memory It could be huge...
-    return File.open(source) if source.is_a? String
     source.rewind
     source.read
   end
--- a/app/services/hyrax/characterization/valkyrie_characterization_service.rb
+++ b/app/services/hyrax/characterization/valkyrie_characterization_service.rb
@@ -1,0 +1,113 @@
+# frozen_string_literal: true
+require 'hydra-file_characterization'
+require 'nokogiri'
+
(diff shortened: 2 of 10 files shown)


dunn:
past the rubocop issues this is the spec failure i'm seeing
```
  1) Hyrax::Characterization::ValkyrieCharacterizationService run successfully sets the property values
     Failure/Error: raise FailedSaveError.new(err.message, obj: af_object)

     Wings::Valkyrie::Persister::FailedSaveError:
       Save the file first
     # ./lib/wings/valkyrie/persister.rb:37:in `rescue in save'
     # ./lib/wings/valkyrie/persister.rb:24:in `save'
     # ./app/services/hyrax/characterization/valkyrie_characterization_service.rb:14:in `run'
     # ./spec/services/hyrax/characterization/valkyrie_characterization_service_spec.rb:14:in `block (3 levels) in <top (required)>'
     # /usr/local/bundle/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # RuntimeError:
     #   Save the file first
     #   /usr/local/bundle/gems/active-fedora-13.2.4/lib/active_fedora/with_metadata/metadata_node.rb:49:in `save'
```
i'm not sure exactly what wings/af wants here

app/jobs/valkyrie_ingest_job.rb
```diff
@@ -0,0 +1,59 @@
...
+    file_metadata.file_set_id = file.file_set_uri
+    uploaded = Hyrax.storage_adapter
+                    .upload(resource: file_set,
+                            file: File.open(uploader.file.file),
+                            original_filename: file_metadata.original_filename)
+    file_metadata.file_identifier = uploaded.id
+    file_metadata.size = uploaded.size
+
+    # characterization is run on the Hyrax::FileMetadata, but re-indexing is
+    # performed on the FileSet, so we need to pass that through as well
```
dunn, line 50: actually this isn't necessary; the `metadata.updated` event emitted by the WorkUploadsHandler triggers a reindex

dunn:
as far as i can tell the wings problem starts here: https://github.com/samvera/active_fedora/blob/main/lib/active_fedora/with_metadata.rb#L21-L26

the object we're saving is a `PCDM::File`, and `super` on line 22 leads us up to: https://github.com/samvera/active_fedora/blob/main/lib/active_fedora/file.rb#L160-L164

where it tries to save the associated metadata:
```
From: /usr/local/bundle/gems/active-fedora-13.2.4/lib/active_fedora/file.rb:162 ActiveFedora::File#create_or_update:

    160: def create_or_update(*options)
    161:   super.tap do
 => 162:     metadata.save if metadata.changed?
    163:   end
    164: end

[1] pry(#<Hydra::PCDM::File>)> metadata
=> #<ActiveFedora::WithMetadata::MetadataNode:0x67958 ID:<http://fcrepo:8080/rest/test/_i/d/_id>>
[2] pry(#<Hydra::PCDM::File>)> metadata.changed?
=> true
```

finally bottoming out with the error:
```
From: /usr/local/bundle/gems/active-fedora-13.2.4/lib/active_fedora/with_metadata/metadata_node.rb:49 ActiveFedora::WithMetadata::MetadataNode#save:

    48: def save
 => 49:   raise "Save the file first" if file.new_record?
    50:   SparqlInsert.new(changes_for_update, file.uri).execute(metadata_uri)
    51:   @ldp_source = nil
    52:   @metadata_uri = nil
    53:   true
    54: end
```
would it work to call `metadata_node.save` first, before the super invocation? https://github.com/samvera/active_fedora/blob/main/lib/active_fedora/with_metadata.rb#L25

no-reply:
i think the issue is simply on the test setup side. FCRepo won't let us persist a `FileMetadata` without first persisting the File as an [LDP Non-RDF Source](https://www.w3.org/TR/ldp/#ldpnr). in fact, fcrepo proactively creates the resource we map to FileMetadata when the file is uploaded. (following LDP 5.2.3.12).

the test setup is avoiding creating a file at: https://github.com/samvera/hyrax/pull/5303/files#diff-b31fcf0b1604a76b7bfd892b2b62fe4957e88b8b777081054994930077ad1d88R10, and i think the `FileMetadata` just won't be persistable via `ActiveFedora` until one is created. i don't think having Wings make the file anyway is likely to be a helpful solution; it seems like callers would end up accidentally creating extra (mostly empty) files.

one of the tests that fails under the proposed Wings change tests this behavior explicitly `it 'raises an error because AF declines to save files that do not have content'`

app/services/hyrax/characterization/valkyrie_characterization_service.rb
```diff
@@ -0,0 +1,133 @@
...
+  # @api public
+  # @param [Hyrax::FileMetadata] metadata  which has properties to recieve characterization values
+  # @param [Valkyrie::StorageAdapter::StreamFile] source to run characterization against
+  # @param [Hash] options the options pass to characterization
+  # @option options [Hash{Symbol => Symbol}] parser_mapping
+  # @option options [Class] parser_class
+  # @option options [Symbol] ch12n_tool
+  #
+  # @return [Hash]
+  def self.run(metadata, source = nil, **options)
```
no-reply, line 16: i took the liberty of removing positional arguments from the initializer, but wasn't sure if `#run` has an interface defined upstream somewhere.

does this class have potential callers outside the current diff, or should we use named parameters here too?

app/services/hyrax/characterization/valkyrie_characterization_service.rb
```diff
@@ -0,0 +1,133 @@
...
+
+  ##
+  # @!attribute [rw] source
+  #   @return [Valkyrie::StorageAdapter::StreamFile]
+  # @!attribute [rw] metadata
+  #   @return [Hyrax::FileMetadata]
+  # @!attribute [rw] source
+  #   @return [Valkyrie::StorageAdapter::StreamFile]
+  # @!attribute [rw] tools
+  #   @todo clarify what this is for
```
no-reply, line 29: i'm not sure i understand `@tools` or its relationship to  `:ch12n_tool`. can you help me document it?

app/services/hyrax/characterization/valkyrie_characterization_service.rb
```diff
@@ -0,0 +1,133 @@
...
+    @characterizer.characterize(content, file_name, tools) do |cfg|
+      cfg[:fits] = Hydra::Derivatives.fits_path
+    end
+  end
+
+  def file_name
+    metadata.original_filename
+  end
+
+  # Use OM to parse metadata
```
no-reply, line 82: i don't understand how OM is involved.

it's deprecated since 2019, and isn't anywhere in the dependency tree. i think the reference is just confusing and can probably be removed?

app/services/hyrax/characterization/valkyrie_characterization_service.rb
```diff
@@ -0,0 +1,133 @@
...
+  def parse_metadata(metadata)
+    omdoc = parser_class.new
+    omdoc.ng_xml = Nokogiri::XML(metadata) if metadata.present?
+    omdoc.__cleanup__ if omdoc.respond_to? :__cleanup__
```
no-reply, line 86: in our case, this calls a method that does nothing and returns `nil`. can we just remove the whole line?

PR #5315, opened 2021-12-31
Support previous yaml config structure for Analytics

orangewolf:
Allowing time for folks to transition their applications to the new format.  This is meant to allow analytics to be released as a minor instead of a major release.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/analytics/google.rb
+++ b/app/services/hyrax/analytics/google.rb
@@ -32,7 +32,7 @@ module Hyrax
             end
             config = yaml.fetch('analytics')&.fetch('google', nil)
             unless config
-              Rails.logger.info("Deprecated analytics format found. Please update your yaml file.")
+              Deprecation.warn("Deprecated analytics configuration format found. Please update config/analytics.yml.")
               config = yaml.fetch('analytics')
               # this has to exist here with a placeholder so it can be set in the Hyrax initializer
               # it is only for backward compatibility
@@ -58,7 +58,7 @@ module Hyrax
           end
 
           # This method allows setting the analytics id in the initializer
-          # which is deprecated
+          # @deprecated set the analytics id in either ENV['GOOGLE_ANALYTICS_ID'] or config/analytics.yaml
           def analytics_id=(value)
             @config['analytics_id'] = value
           end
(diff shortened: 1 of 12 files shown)


app/services/hyrax/analytics/google.rb
```diff
@@ -30,7 +30,15 @@ def self.load_from_yaml
               Rails.logger.error("Unable to fetch any keys from #{filename}.")
               return new({})
             end
-            new yaml.fetch('analytics')&.fetch('google')
+            config = yaml.fetch('analytics')&.fetch('google', nil)
+            unless config
+              Rails.logger.info("Deprecated analytics format found. Please update your yaml file.")
```
cjcolvar, line 35: Elsewhere Hyrax is using `Deprecation.warn` instead of using the rails logger for deprecation messages.  I think it would be nice to give more guidance in the message.  Is there a wiki page or release notes that explain the expected configuration format?

Start of a suggestion:
```suggestion
              Deprecation.warn("Deprecated analytics configuration format found. Please update config/analytics.yml.")
```

app/services/hyrax/analytics/google.rb
```diff
@@ -48,6 +56,12 @@ def valid?
           REQUIRED_KEYS.each do |key|
             class_eval %{ def #{key};  @config.fetch('#{key}'); end }
           end
+
+          # This method allows setting the analytics id in the initializer
+          # which is deprecated
```
cjcolvar, line 61: Should this method have a deprecation warning as well?  I've seen `@deprecation` yard doc tags used elsewhere in Hyrax so maybe this method should at least have that.

spec/lib/hyrax/analytics_spec.rb
```diff
@@ -29,6 +29,32 @@
       end
     end
 
+    context "When the yaml file has no values" do
```
cjcolvar, line 32:
```suggestion
    context "When the yaml file has a deprecated format" do
```

cjcolvar (approved):
This looks good to me.  My comments are small cleanup suggestions and shouldn't block merging since getting this change in is important.

app/services/hyrax/analytics/google.rb
orangewolf: great point. updated yarn tag and what one should do instead.

orangewolf:
@cjcolvar completely agree with those suggestions. thank you!

cjcolvar (approved):

dunn (approved):

cjcolvar (approved):

orangewolf:
yay! thank you to all who helped me get this through

PR #5319, opened 2022-01-07
ci: use new orb bundle command

dunn:
@samvera/hyrax-code-reviewers


cjcolvar (approved):
Thanks for cleaning this up!

orangewolf (approved):

PR #5320, opened 2022-01-11
don't attempt to save AF::WithMetadata::MetadataNode#file_hash

no-reply:
`ActiveFedora::WithMetadata::MetadataNode` provides a `#file_hash` attribute,
but writing to it, even with the existing data, prevents all saves.

Wings needs to map data in and out of this class freely, and this attribute
functions as a honeypot. it baits Wings to preserve the data, and then fails
save when the converters try to do keep the existing data. since any attempt to
write this data will break save, ActiveFedora should treat it as read only.

never include `MetadataNode#file_hash` in changed attributes lists when Wings is
loaded.

i think this is the issue presently blocking #5278.

### DIFF
--- a/lib/wings/setup.rb
+++ b/lib/wings/setup.rb
@@ -48,6 +48,17 @@ module ActiveFedora
     end
   end
 
+  module WithMetadata
+    class MetadataNode
+      ##
+      # @note fcrepo rejects `:file_hash` updates. the attribute is managed by
+      #   the data store. always drop it from changed attributes.
+      def changed_attributes
+        super.except(:file_hash)
+      end
+    end
+  end
+
   module Associations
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Brilliant! :sparkles:

no-reply:
i'm hoping that the rebase will fix the build.

no-reply:
@cjcolvar do you think it would be a good idea to open an issue on ActiveFedora?

the error behind this is easily repeatable there. it might be appropriate to port this monkeypatch upstream, or come up with a more principled fix.

dunn (approved):

cjcolvar:
@no-reply An upstream fix would probably be a good idea.

PR #5321, opened 2022-01-11
Valkyrie Characterization and File attachment

no-reply:
adds valkyrie upload handling and characterization, based on the hydra-works service class, to allow characterizing file uploads of Valkyrie objects.

replaces #5303.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/wings/valkyrie/persister.rb
+++ b/lib/wings/valkyrie/persister.rb
@@ -34,6 +34,9 @@ module Wings
 
         resource_factory.to_resource(object: af_object)
       rescue ActiveFedora::RecordInvalid, RuntimeError => err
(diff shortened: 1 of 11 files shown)


lib/wings/valkyrie/persister.rb
```diff
@@ -34,6 +34,9 @@ def save(resource:)
 
         resource_factory.to_resource(object: af_object)
       rescue ActiveFedora::RecordInvalid, RuntimeError => err
+        raise "more serious and helpful error message" if
```
dunn, line 37: were you going to replace this message?

dunn (approved):

PR #5324, opened 2022-01-12
test deposit of files via WorkUploadsHandser 

no-reply:
`Hyrax::WorkUploadsHandler` should always ingest files matching its
uploads. prior to this, its tests skipped running the background job that
handles ingest. that job has (still) no unit tests of its own.

this makes inroads toward comprehensive ingest tests by wireing in one spec from
`WorkUploadsHandler` down. we stub out the characterizer since that behavior has
strong tests of its own, and we don't want to include the system call to
`fits.sh` here---the background job is run by the ActiveJob test adapter
directly in the app container, so the fits dependency isn't expected to be
installed.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/valkyrie_ingest_job.rb
+++ b/app/jobs/valkyrie_ingest_job.rb
@@ -14,9 +14,8 @@ class ValkyrieIngestJob < Hyrax::ApplicationJob
   # @return [void]
   def ingest(file:)
     file_set = Hyrax.query_service.find_by(id: file.file_set_uri)
-    file_metadata = Hyrax::FileMetadata.for(file: file.uploader.file)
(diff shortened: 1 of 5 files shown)


app/jobs/valkyrie_ingest_job.rb
```diff
@@ -37,20 +36,19 @@ def add_file_to_file_set(file_set:, file_metadata:)
   # @param [Hyrax::FileMetadata] file_metadata
```
dunn, line 36: this parameter has been removed

no-reply: thanks fixed!

dunn (approved):

PR #5328, opened 2022-01-20
fix dashboard/my/collections when default admin set table doesn’t exist

elrayle:
### Description

A bug caused the missing table to be checked even after determining that saving the default admin set id wasn’t supported.  It is supposed to be ok for the table to be missing for backward compatibility.

image

### Solution

The primary cause of the bug was that a return was missing in the method that determines the default admin set id.  It was supposed to return the `DEFAULT_ID` after determining that saving the id was not supported because the table was missing.  

Adding the return of the `DEFAULT_ID` when saving is not supported fixed the first problem and revealed a second problem where after not finding an admin set with the `DEFAULT_ID`, it was throwing an exception.  It is not ok for an admin set not to exist with the saved admin set id, but it is ok for an admin set with the `DEFAULT_ID` not to exist.  When this happens, the default admin set will be created with `id=DEFAULT_ID`.  This missing object exception is now ignored when the `id==DEFAULT_ID`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/admin_set_create_service.rb
+++ b/app/services/hyrax/admin_set_create_service.rb
@@ -118,7 +118,12 @@ module Hyrax
         return if id.blank?
         Hyrax.query_service.find_by(id: id)
       rescue Valkyrie::Persistence::ObjectNotFoundError
-        # id is saved but doesn't exist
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Looks good to me.

PR #5355, opened 2022-01-24
ValkyrieIngestJob: ensure FileMetadata exists

dunn:
Some adapters (fcrepo) will automatically create a metadata object, while others
won't.  So let's do a little `find_or_create`-style check here to make sure we
don't try to query a non-existent object.

@samvera/hyrax-code-reviewers


no-reply (approved):
it looks like this job doesn't yet have any tests.

i won't hold this on getting some in, so approving, but i do think it needs specs if it's churning already.

PR #5358, opened 2022-01-25
fix #original_file? status for FileMetadata created by ValkyrieIngestJob

no-reply:
after the ingest job runs:

  - the metadata should be saved, even if the characterizer didn't save it.
  - the uploaded file should be set as an `#original_file? # => true`

since the file set is created specifically for this uploaded file, we always
want a file uploaded through this process to be THE `#original_file` for the
FileSet.

it's possible we'll want to reuse this ingest job to attach files for other
"uses"/of other types. in this case, we should consider parameterizing the use.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/valkyrie_ingest_job.rb
+++ b/app/jobs/valkyrie_ingest_job.rb
@@ -3,13 +3,15 @@ class ValkyrieIngestJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
 
   ##
-  # @param [Valkyrie::StorageAdapter::StreamFile] file
+  # @param [Hyrax::UploadedFile] file
   def perform(file)
     ingest(file: file)
   end
 
   ##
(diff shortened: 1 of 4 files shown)


dunn (approved):

elrayle (approved):
Looks good to me.  Really glad to see the addition of tests to help clarify expectations.

elrayle:
How are characterization files with other uses (e.g. thumbnail) uploaded currently?  I like your suggestion of passing in use as a parameter if this ingest job is used for those uploads.

no-reply:
> How are characterization files with other uses (e.g. thumbnail) uploaded currently? I like your suggestion of passing in use as a parameter if this ingest job is used for those uploads.

at least right now, all web uploads in a basic Hyrax app are the original in their respective FileSet. that is, the upload process creates a file set per upload, and attaches the related upload as the original file. derivatives aren't generated through the `UploadedFile` process, so thumbnails are unrelated.

Issue #5292, opened 2021-12-15
Migrate away from Glyphicons


marrus-sh:
### Descriptive summary

Hyrax will likely need to transition to a different icon font, like [Fork Awesome](https://forkaweso.me/Fork-Awesome/) or [Octicons](https://octicons.github.com/), as Glyphicons is a non‐free icon set and dropped from Bootstrap 4.

### Rationale

Bootstrap 4 support.

### Related work

See #5276, projectblacklight/blacklight#1585, geoblacklight/geoblacklight#840.

marrus-sh:
https://github.com/samvera/hyrax/blob/e3baf5ea219b98a69e9921e20612ab858348da94/lib/hyrax/engine.rb#L121 ??!?!

PR #5359, opened 2022-01-25
Update navbar UI for Bootstrap 4

### LINKED ISSUE #5335: Navbar items alignment
### Descriptive summary

Hyrax Bootstrap 7 branch.

Language selector, alerts and login/profile nav items should be right-aligned in header

image

### Expected behavior

Items are right aligned in desktop viewport size

image

### Actual behavior

Items are left aligned in desktop viewport size

adamjarling:
Fixes #5335 

Updates nav bar styling and layout.  Now looks like this:

image

* Update to Bootstrap 4 convention and some cleanup

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/_logo.html.erb
+++ b/app/views/_logo.html.erb
@@ -1,4 +1,3 @@
 <a id="logo" class="navbar-brand" href="<%= hyrax.root_path %>" data-no-turbolink="true">
-  <span class="glyphicon glyphicon-globe" role="img" aria-label="<%= application_name %>" aria-hidden="true"></span>
   <span class="institution_name"><%= application_name %></span>
 </a>
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -5,23 +5,22 @@
       <%= render_notifications(user: current_user) %>
     </li>
     <li class="nav-item dropdown">
-      <%= link_to hyrax.dashboard_profile_path(current_user), role: 'button', class: 'nav-link dropdown-toggle', data: { toggle: 'dropdown' }, aria: { haspopup: true, expanded: false, controls: 'user-util-links' } do %>
+      <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
         <span class="sr-only"><%= t("hyrax.toolbar.profile.sr_action") %></span>
-        <span class="d-none d-md-block">&nbsp;<%= current_user.name %></span>
+        <span><%= current_user.name %></span>
         <span class="sr-only"> <%= t("hyrax.toolbar.profile.sr_target") %></span>
-        <span class="fa fa-user" aria-hidden="true"></span>
-      <% end %>
-      <ul id="user-util-links" class="dropdown-menu" role="menu">
-        <li class="dropdown-item"><%= link_to t("hyrax.toolbar.dashboard.menu"), hyrax.dashboard_path %></li>
-
-        <li class="dropdown-divider"></li>
-        <li class="dropdown-item"><%= link_to t("hyrax.toolbar.profile.logout"), main_app.destroy_user_session_path %></li>
-      </ul>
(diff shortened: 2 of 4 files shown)


app/views/_logo.html.erb
```diff
@@ -1,4 +1,3 @@
 <a id="logo" class="navbar-brand" href="<%= hyrax.root_path %>" data-no-turbolink="true">
-  <span class="glyphicon glyphicon-globe" role="img" aria-label="<%= application_name %>" aria-hidden="true"></span>
```
marrus-sh, line 2: Glyphicons have been dropped (they used to be a part of Bootstrap) but we do also support FontAwesome (4, i assume) if you want to preserve something similar; see #5292.

**Edit:** I see further down that you did in fact do that 😅.

app/views/_user_util_links.html.erb
```diff
@@ -1,27 +1,26 @@
 <ul id="user_utility_links" class="navbar-nav">
-  <%= render 'shared/locale_picker' if available_translations.size > 1 %>
+  <!-- <%= render 'shared/locale_picker' if available_translations.size > 1 %>-->
   <% if user_signed_in? %>
     <li class="nav-item">
       <%= render_notifications(user: current_user) %>
     </li>
     <li class="nav-item dropdown">
-      <%= link_to hyrax.dashboard_profile_path(current_user), role: 'button', class: 'nav-link dropdown-toggle', data: { toggle: 'dropdown' }, aria: { haspopup: true, expanded: false, controls: 'user-util-links' } do %>
+      <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
marrus-sh, line 8: Can you share the reason behind this change?

app/views/_logo.html.erb
adamjarling: Yea, maybe I'm making a style decision but seems the Globe icon just gets in the way (to me).  Do we want to keep it?

app/views/_user_util_links.html.erb
adamjarling: Sure.  I think the `a.dropdown-toggle` element (from looking at Bootstrap 4 documentation), might serve as a toggle element and not necessarily a navigation element.  Clicking it opens/closes the dropdown menu.  If it's navigable, then clicking on it would take the user to the link instead of triggering the dropdown UI state.

https://getbootstrap.com/docs/4.0/components/navbar/

app/views/_logo.html.erb
marrus-sh: I think the “intent” of the globe is to be a sort of standin for organization logos; see for example:

+ http://libarchive.linnbenton.edu
+ https://scholarspace.library.gwu.edu

But I’ve also found ones which just deleted it:

+ https://archive.kingsfund.org.uk

Personally, I don’t find having a hardcoded globe that institutions would have to manually replace to be particularly helpful, and am fine with just getting rid of it.

app/views/_user_util_links.html.erb
marrus-sh: My only question with changes like this is whether it is conceivable a user might want to right-click here and click “open in new tab” instead of toggling the state. If that’s not very likely (for example because there’s a better / more obvious way of getting to that page) then I’m fine with it.

adamjarling: Ah, yea however you want to do it is fine with me.  Doesn't look like the v4 documentation provides any examples of the toggle being navigable, and it maybe gives an example to follow Bootstrap implementation as closely as possible?

marrus-sh: Considering that things like command-click and middle-click already don’t seem to work to open this link in a new tab, I think replacing this with `#` is probably fine. It might be nice to have a “profile” link in this dropdown, but I’m not sure translations exist for it and don’t really want to worry about that right now 😅.

marrus-sh (approved):

PR #5364, opened 2022-01-25
update admin set create process to convert to valkyrie resource when receiving AF adminset

### LINKED ISSUE #5334: Creating an Admin Set does not save to Valkyrie solr core
### Descriptive summary

Saving an Admin Set should index the new set in the Active Fedora solr core and the Valkyrie solr core (when configured).

### Rationale

Indexing of core models should be indexed in the Valkyrie solr core when it is configured.

### Expected behavior

When an Admin Set is created, the ActiveFedora and Valkyrie solr cores should both have a document created describing the new Admin Set.

### Actual behavior

When an Admin Set is created, only the ActiveFedora solr core has a document created describing the new Admin Set.

#### Prerequisites:

Configure Hyrax:
```
```
2. restart rails app
3. Make sure regular `hyrax-development` solr is running and `hyrax-valkyrie-dev`.  _Your core names may vary depending on how you start them._

#### To reproduce:

Create Admin Set
2. select type: Admin Set
3. set Title: MY ADMIN SET (or other meaningful name)
4. click Save

In solr cores for the Hyrax app:
* run `q`:  `title_tesim:MY ADMIN SET` for hyrax-development core and verify the admin set solr doc was created
* run `q`:  `title_tesim:MY ADMIN SET` for hyrax-valkyrie-dev core and note that the admin set solr doc was NOT created

### Related work

elrayle:
Fixes #5334

To move closer to full Valkyrization, this refactors the AdminSetControllerto always convert `@admin_set` to a `Hyrax::AdministrativeSet` which is a `Valkyrie::Resource` before performing actions or calling other classes (e.g. `Hyrax::AdminSetCreateService`).

The create service was refactored to internally perform a conversion to `Hyrax::AdministrativeSet` if it received an ActiveFedora `AdminSet`.  This allowed removal of special processing depending on the type.  It already only returned a `Hyrax::AdministrativeSet`.  And the deprecated method `.call` still continues to allow `AdminSet` as a parameter.  These allow for backward compatibility.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -106,7 +106,8 @@ module Hyrax
       updated_admin_set = admin_set_create_service.call!(admin_set: admin_set_resource, creating_user: current_user)
       update_admin_set(updated_admin_set)
       true
-    rescue RuntimeError
+    rescue RuntimeError => err
+      Rails.logger.error("Failed to create admin set through valkyrie: #{err.message}")
       false
     end
 
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -71,7 +71,7 @@ module Hyrax
 
     def create
       if create_admin_set
-        redirect_to hyrax.edit_admin_admin_set_path(@admin_set), notice: I18n.t('new_admin_set', scope: 'hyrax.admin.admin_sets.form.permission_update_notices', name: @admin_set.title.first)
+        redirect_to hyrax.edit_admin_admin_set_path(admin_set_id), notice: I18n.t('new_admin_set', scope: 'hyrax.admin.admin_sets.form.permission_update_notices', name: @admin_set.title.first)
       else
         setup_form
         render :new
@@ -82,7 +82,7 @@ module Hyrax
       if @admin_set.destroy
         after_delete_success
       else
-        redirect_to hyrax.admin_admin_set_path(@admin_set), alert: @admin_set.errors.full_messages.to_sentence
+        redirect_to hyrax.admin_admin_set_path(admin_set_id), alert: @admin_set.errors.full_messages.to_sentence
       end
     end
 
@@ -99,11 +99,15 @@ module Hyrax
     private
 
     def update_referer
-      hyrax.edit_admin_admin_set_path(@admin_set) + (params[:referer_anchor] || '')
+      hyrax.edit_admin_admin_set_path(admin_set_id) + (params[:referer_anchor] || '')
     end
 
     def create_admin_set
-      admin_set_create_service.call(admin_set: @admin_set, creating_user: current_user)
+      updated_admin_set = admin_set_create_service.call!(admin_set: admin_set_resource, creating_user: current_user)
+      update_admin_set(updated_admin_set)
+      true
+    rescue RuntimeError
+      false
     end
 
     def setup_form
@@ -144,6 +148,28 @@ module Hyrax
       else
         redirect_to hyrax.my_collections_path, notice: t(:'hyrax.admin.admin_sets.delete.notification')
       end
+    end
+
+    def admin_set_id
+      @admin_set&.id&.to_s
+    end
+
+    def admin_set_resource
+      case @admin_set
+      when Valkyrie::Resource
+        @admin_set
+      else
+        @admin_set.valkyrie_resource
+      end
+    end
+
+    def update_admin_set(updated_admin_set)
+      @admin_set = case @admin_set
+                   when Valkyrie::Resource
+                     updated_admin_set
+                   else
+                     Wings::ActiveFedoraConverter.convert(resource: updated_admin_set)
+                   end
     end
   end
 end
--- a/app/services/hyrax/admin_set_create_service.rb
+++ b/app/services/hyrax/admin_set_create_service.rb
@@ -64,7 +64,7 @@ module Hyrax
         Deprecation.warn("'##{__method__}' will be removed in Hyrax 4.0.  " \
                          "Warning: This method may hide runtime errors.  " \
                          "Instead, use 'Hyrax::AdminSetCreateService.call!'.  ")
-        call!(admin_set: admin_set, creating_user: creating_user, **kwargs).present?
+        call!(admin_set: admin_set_resource(admin_set), creating_user: creating_user, **kwargs).present?
       rescue RuntimeError => err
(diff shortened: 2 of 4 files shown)


app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -71,7 +71,7 @@ def update
 
     def create
       if create_admin_set
-        redirect_to hyrax.edit_admin_admin_set_path(@admin_set), notice: I18n.t('new_admin_set', scope: 'hyrax.admin.admin_sets.form.permission_update_notices', name: @admin_set.title.first)
+        redirect_to hyrax.edit_admin_admin_set_path(admin_set_id), notice: I18n.t('new_admin_set', scope: 'hyrax.admin.admin_sets.form.permission_update_notices', name: @admin_set.title.first)
```
elrayle, line 74: This works for both `AdminSet` (active fedora) and `Hyrax::AdministrativeSet` (valkyrie).  The method gets the id and forces it to a `String`.  The redirect fails when id is a `Valkyrie::ID`

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -99,11 +99,15 @@ def self.local_prefixes
...
 
     def update_referer
-      hyrax.edit_admin_admin_set_path(@admin_set) + (params[:referer_anchor] || '')
+      hyrax.edit_admin_admin_set_path(admin_set_id) + (params[:referer_anchor] || '')
     end
 
     def create_admin_set
-      admin_set_create_service.call(admin_set: @admin_set, creating_user: current_user)
+      updated_admin_set = admin_set_create_service.call!(admin_set: admin_set_resource, creating_user: current_user)
+      update_admin_set(updated_admin_set)
```
elrayle, line 107: This is required because @admin_set does not get updated by the create service when it is a `Hyrax::AdministrativeSet`.  The create service returns the updated admin set, so the instance variable update has to happen in the controller.

app/controllers/hyrax/admin/admin_sets_controller.rb
```diff
@@ -99,11 +99,15 @@ def self.local_prefixes
...
+      true
+    rescue RuntimeError
+      false
```
elrayle, line 110: Returning Boolean maintains the current behavior based on the `create_admin_set` methods results.

app/services/hyrax/admin_set_create_service.rb
```diff
@@ -182,23 +191,6 @@ def create
     # @return [Hyrax::AdministrativeSet] The fully created admin set.
     # @raise [RuntimeError] if admin set cannot be persisted
     def create!
-      admin_set.respond_to?(:valkyrie_resource) ? active_fedora_create! : valkyrie_create!
-    end
-
-    private
-
-    def default_admin_set?
```
elrayle, line 190: This diff shows as a delete, but most of the private methods actually still exist and show as an add farther down.  This is because the guts of the `valkyrie_create!` method moved up to the `create!` method.  The only private methods removed are `valkyrie_create!` and `fedora_create!` as all creates are now done using valkyrie processing.

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
```diff
@@ -174,31 +174,32 @@
       end
 
       describe "#create" do
-        before do
-          controller.admin_set_create_service = service
-        end
-
         context "when it's successful" do
-          let(:service) do
-            lambda do |admin_set:, **_kargs|
```
elrayle, line 183: The controller now calls the `.call!` method instead of the deprecated `.call` method.  The lambda service expects to receive `call`, and as such won't work anymore.

app/controllers/hyrax/admin/admin_sets_controller.rb
no-reply: if we're going to swallow all errors, let's at least log them?

spec/services/hyrax/admin_set_create_service_spec.rb
```diff
@@ -203,12 +205,16 @@
...
+        let(:listener) { Hyrax::Specs::SpyListener.new }
         let(:permission_template) { Hyrax::PermissionTemplate.find_by(source_id: admin_set.id) }
         let(:grants) { permission_template.access_grants }
         let(:available_workflows) { [create(:workflow), create(:workflow)] }
 
         # rubocop:disable RSpec/AnyInstance
         before do
+          Hyrax.publisher.subscribe(listener)
+          allow_any_instance_of(Hyrax::Listeners::MetadataIndexListener)
+            .to receive(:on_collection_metadata_updated).and_return(available_workflows)
```
no-reply, line 217: what's going on here? the index listener is returning the workflows?

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
no-reply: what motivated the switch to `#call!`?

this seems like a downside of that move. i'm not necessarily wanting to revisit if it's already done, but am curious.

spec/services/hyrax/admin_set_create_service_spec.rb
elrayle: Got to be a copy-paste error.  I'll take a closer look.

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
elrayle: The create service saves the admin set.  

For ActiveFedora:  `admin_set.save` updates `admin_set`.  

For Valkyrie:  `Hyrax.persister.save(resource: admin_set` does not, so you have to return the results of the save to have access to the updated.

For backward compatibility, `.call` continues to return Boolean.  The `.call` method is deprecated. `.call!` is the replacement method and it returns the updated admin set.

app/controllers/hyrax/admin/admin_sets_controller.rb
elrayle: I really don't like swallowing error.  I was doing a minimal change here.  One click up is to log the error.  If we show it to the user, we'd probably have to pretty it up.  Things like `Valkyrie::ObjectNotFound` is an ugly error to show an end user.

spec/controllers/hyrax/admin/admin_sets_controller_spec.rb
no-reply: thanks! this refreshed my memory.

no-reply (approved):

Issue #5276, opened 2021-12-07
Support Bootstrap 4


jlhardes:
### Descriptive summary

Bootstrap 4 is a needed upgrade to Hyrax UI capabilities

### Rationale

We should keep our core dependencies up to date. Also, this upgrade will allow Hyrax to upgrade to support Rails 6 and Blacklight 7.

### Related work

There might be a start to this work on this branch but this work is also from a little while back in time: https://github.com/samvera/hyrax/tree/3351-blacklight-7-bootstrap-4

Also seems connected to this closed issue to start this work: https://github.com/samvera/hyrax/issues/3351

jlhardes:
This work helps support work for these issues:

Add support for Blacklight 7 - https://github.com/samvera/hyrax/issues/3271
Support Rails 6.0 - https://github.com/samvera/hyrax/issues/4016

marrus-sh:
I’m hopefully going to be poking at this a bit in #5280; I think the goal right now is to try to resolve any failing tests in the backend after the upgrade to Blacklight 7, and then look at the frontend and see where that leaves us.

This is largely exploratory, and I imagine that at least some of the needed frontend work will be able to be applied right away (i.e. without the full Blacklight upgrade; but it’s important to do the upgrade at least minimally just to see what work actually needs to be done), but there might be a discussion to be had (or maybe which has been had) regarding stability guarantees for frontend work—if people are expecting Hyrax to have such‐and‐such a UI for their own downstream theming purposes, and the upgrade to Bootstrap 4 is going to be significantly changing the internal structure of that UI (i.e. breaking a lot of downstream themes), then it likely needs to be a major version change, in which case I imagine we’ll want to do the full Blacklight upgrade (to 7 or maybe even 8?!) at the same time?

marrus-sh:
98a30d7de925cd59b90ca8281fc08caa766fe99a is far from a complete migration, but it at least identifies all of the variables which are no longer defined (and thus will break asset compilation) in Bootstrap 4.

marrus-sh:
A significant amount of work has been done in this dev congress. Remaining work to be done includes:

+ Updating the feature tests to accommodate the recent changes to views #5420.
+ Finding a substitute for jQuery-File-Upload, which doesn’t support Bootstrap 4 #5301.
+ Works and collections can’t currently be added, presumably due to some jQuery selectors being out-of-date? (There may be similar buttons which fail to work elsewhere in the UI.) These need to be tracked down and updated #5345.

[...]

PR #5366, opened 2022-01-25
Add placeholder stylesheet for adjusting font-size globally

### LINKED ISSUE #5340: Bootstrap 4: Default font-size has shifted
### Descriptive summary

Default font-size has shifted from `14px` in Bootstrap 3 to `1rem` (computed by modern browsers as `16px`). This change gives greater control over font-size globally but may impact installs of Hyrax. 

### Rationale

To avoid cosmetic issues, we may want to set a fixed font-size of `0.875em` computed as 14px (the previous default for bootstrap 3)

### Expected behavior

Should we reset the computation of the default font-size for the Hyrax front-end to be `0.875em` in the `<html>` element? All bootstrap elements would still calculate at 1rem, but that rem would be computed as 14px.

Alternatively, it may be okay to ignore this and let everything inflate. This might be the most forward facing solution but could lead to issues for anyone who has set `px` based font-sizing in a custom way.

### Actual behavior
Bootstrap 4 will inflate everything from a default of 14px to 16px (1rem)

mathewjordan:
Fixes #5340  ; refs #5276 

## Summary

Bootstrap 4 has increased global font-size from 14px to 1rem, computed  by modern browsers as 16px. This pull request provides minimal guidance to Hyrax developers on how to reset font-size. 

See https://getbootstrap.com/docs/4.0/migration/ for more information.

### DIFF
--- a/app/assets/stylesheets/hyrax/_html.scss
+++ b/app/assets/stylesheets/hyrax/_html.scss
@@ -1,0 +1,11 @@
+/*
+ * Bootstrap 4 has increased global font-size from 14px to 1rem, computed 
+ * by modern browsers as 16px. This can be reset by overriding the html
+ * selector and adding the property/value of `font-size: 14px;`.
+ *
+ * See https://getbootstrap.com/docs/4.0/migration/ for more information.
+ */
+
+html {
+  font-size: inherit;
+}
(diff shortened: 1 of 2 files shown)


marrus-sh (approved):
This looks good to me; I’ll try to remember to call this out when it comes time for actually merging this branch.

PR #5368, opened 2022-01-25
Test returning of pcdm_collections

cjcolvar:
Changes proposed in this PR:
- Default use_valkyrie to Hyrax.config.use_valkyrie? to avoid having to explicitly pass in a value
- Use non-deprecated form of collections? in collection_type index view
- Add test for returning of PcdmCollections from #collections
- Force globalid to string when passing to custom query

I came across this when working on nurax_pg.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/collection_type.rb
+++ b/app/models/hyrax/collection_type.rb
@@ -109,9 +109,9 @@ module Hyrax
 
     ##
     # @return [Enumerable<Collection, PcdmCollection>]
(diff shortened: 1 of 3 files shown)


elrayle (approved):
Looks good.  Thanks for cleaning this up.

elrayle (approved):
Still looks good.  Thanks for the work.

Issue #5329, opened 2022-01-20
Hyrax::PcdmCollectionForm does not support #permission_template


elrayle:
### Descriptive summary

When editing a collection, the Sharing tab allows the user to set default permissions for works created in that collection and the Discoverable tab allows the user to set the visibility of the collection.  The view `hyrax/dashbaord/collections/_form_share.html.erb` uses @form.permission_template to construct the form allowing the user to add access.  Similarly, the view `hyrax/dashboard/collections/_form_discovery.html.erb` uses @form.permission_template to determine the current visibility of the collection.  

When working with Hyrax::PcdmCollections (Valkyrie::Resource), the Hyrax::PcdmCollectionForm (Valkyrie::ChangeSet) is used.  At this time, it does not support `#permission_template`.  This error impacts Sharing and Discovery tabs of the collection edit form.

### Rationale

Working with AF Collections or Valkyrie Collections, the Sharing and Discovery tabs should allow users to set default permissions and visibility, respectively, for the collection.

### Expected behavior

Editing a Hyrax::PcdmCollection should load the edit form, including the Sharing and Discovery tabs.

### Actual behavior

Editing a Hyrax::PcdmCollection fails to load the edit form for multiple reasons.  One includes the inability to load the Sharing and Discovery tabs because the `@form` object does not support the `#permission_template` method.

![image](https://user-images.githubusercontent.com/6855473/150575395-a2227d32-861c-47df-afc8-a28e878518d6.png)

#### Prerequisites:

Edit `/config/initializers/hyrax.rb` and set:
```
  config.colleccollectiontion_model = "Hyrax::PcdmCollection"
```

To isolate issues in other tabs, create a collection_type with all options off except Discovery.
1. navigate to Dashboard -> Settings -> Collection Types 
2. click button: Create new collection type
3. set Type name: Discovery only (or other meaningful name)
4. click button: Save
5. click tab: Settings
6. set to only have Discovery turned on (as seen below)
7. click button: Save changes

![image](https://user-images.githubusercontent.com/6855473/150574407-db2ad7e9-9ff6-49be-9392-e3ebbd4c465e.png)

#### To reproduce:

1. navigate to Dashboard -> Collections
2. click button: New Collection
3. select type: Discovery only (or the name you used when creating the collection type)
3. click button: Create Collection
3. set Title: My Discoverable Collection (or other meaningful name)
4. click Save (this loads the editor for the collection including the Discovery tab)

The error will occur as the edit form tries to load.

[...]

PR #5373, opened 2022-01-25
Only call render if tabs should be shown

cjcolvar:
When working on https://github.com/samvera/hyrax/issues/5329 and following the steps to reproduce I found that even if a collection is of a collection type that doesn't allow branding the branding partial was still being rendered.  These changes now mirror the logic at the top of the partial for displaying the tab headers.

@samvera/hyrax-code-reviewers


cjcolvar:
I'm not sure why the formatting changed...I'll have to see what vim did and fix it.

elrayle (approved):
Nice catch.

PR #5374, opened 2022-01-25
WorkUploadsHandler: don't pre-publish events that depend on bg jobs

dunn:
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/work_uploads_handler.rb
+++ b/app/services/hyrax/work_uploads_handler.rb
@@ -116,9 +116,6 @@ module Hyrax
 
       ValkyrieIngestJob.perform_later(file)
 
-      # this triggers the re-index
-      Hyrax.publisher.publish('object.metadata.updated', object: file_set, user: file.user)
-
       { file_set: file_set, user: file.user }
     end
 
--- a/spec/services/hyrax/work_uploads_handler_spec.rb
+++ b/spec/services/hyrax/work_uploads_handler_spec.rb
@@ -74,10 +74,7 @@ RSpec.describe Hyrax::WorkUploadsHandler, valkyrie_adapter: :test_adapter do
       it 'publishes object.metadata.updated event for work' do
         expect { service.attach }
           .to change { listener.object_metadata_updated }
-          .to contain_exactly(have_attributes(payload: include(object: be_work)),
-                              have_attributes(payload: include(object: be_file_set)),
-                              have_attributes(payload: include(object: be_file_set)),
-                              have_attributes(payload: include(object: be_file_set)))
+          .to contain_exactly(have_attributes(payload: include(object: be_work)))
       end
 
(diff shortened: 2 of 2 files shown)


app/services/hyrax/work_uploads_handler.rb
```diff
@@ -92,7 +92,6 @@ def attach
       acquire_lock_for(work.id) do
         event_payloads = files.each_with_object([]) { |file, arry| arry << make_file_set_and_ingest(file) }
         @persister.save(resource: work)
-        Hyrax.publisher.publish('object.metadata.updated', object: work, user: files.first.user)
```
no-reply, line 95: this one should stay!

it's necessary since we save immediately above it.

app/services/hyrax/work_uploads_handler.rb
```diff
@@ -116,9 +115,6 @@ def make_file_set_and_ingest(file)
 
       ValkyrieIngestJob.perform_later(file)
 
-      # this triggers the re-index
-      Hyrax.publisher.publish('object.metadata.updated', object: file_set, user: file.user)
```
no-reply, line 120: there mighh be unit tests checking this that should be removed?

app/jobs/valkyrie_ingest_job.rb
```diff
@@ -36,6 +34,8 @@ def ingest(file:)
   def add_file_to_file_set(file_set:, file_metadata:)
     file_set.file_ids << file_metadata.id
     Hyrax.persister.save(resource: file_set)
+    # this triggers the re-index
+    Hyrax.publisher.publish('object.metadata.updated', object: file_set)
```
no-reply, line 38: i think we want to introduce `object.membership.updated`.

i'm working on this in this line: https://github.com/samvera/hyrax/pull/5375

app/services/hyrax/work_uploads_handler.rb
dunn: oh right!

app/jobs/valkyrie_ingest_job.rb
dunn: yeah that's why i marked this as draft, i wasn't sure if you wanted to tackle that first

spec/services/hyrax/work_uploads_handler_spec.rb
```diff
@@ -71,15 +71,6 @@
                               have_attributes(payload: include(file_set: be_file_set)))
       end
 
-      it 'publishes object.metadata.updated event for work' do
-        expect { service.attach }
-          .to change { listener.object_metadata_updated }
-          .to contain_exactly(have_attributes(payload: include(object: be_work)),
```
no-reply, line 77: i think we still want to test this `publish` for the work, just not the file sets?

cjcolvar (approved):
Looks good!

PR #5377, opened 2022-01-25
Use FacetItemPresenter not render_facet_item

### LINKED ISSUE #5369: Cannot visit Collections page
### Descriptive summary

When there are existing Collections in the repository, cannot go to the Collections page.

### Expected behavior
collection-main

### Actual behavior
collection-bl7

1. Create a Collection while on `main` branch
2. Switch to `bl7bootstrap4` branch, `rm .dassie/Gemfile.lock`, `bundle update`, `docker-compose up`
3. Visit Collections page from the sidebar on the dashboard

marrus-sh:
Fixes #5369.

`render_facet_item` is deprecated in Blacklight 7 and seems to be broken. Poking through the code, it looks like `Blacklight::FacetItemPresenter` is the new way of doing this, so I’ve taken an initial stab at that here.

I don’t love how messy this solution is, but without a deeper understanding of all the involved parts this is the simplest way I could get this page to work.

### DIFF
--- a/app/views/hyrax/my/_facet_pivot.html.erb
+++ b/app/views/hyrax/my/_facet_pivot.html.erb
@@ -9,7 +9,8 @@
         <% item.fq= {} if subfacet %>
 
         <% # The unless prevents Collection from being included in the select list for the Collection Type filter. %>
-        <%= render_facet_item(item.field, item) unless subfacet != true && item.value == "Collection" %>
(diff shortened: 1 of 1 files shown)


adamjarling:
@marrus-sh Can you include a screen shot of what this fixes?  I'm not sure how to test this.

adamjarling (approved):

PR #5379, opened 2022-01-26
index a FileSet when its pcdm:OriginalFile's metadata is updated

no-reply:
when a `Hyrax::FileMetadata` updates, we want to index the FileSet it is
contained within ONLY if the PCDM Use is `OriginalFile`.

check this in response to messages on the new `file.metadata.updated` event.

@samvera/hyrax-code-reviewers


dunn (dismissed):

spec/services/hyrax/listeners/file_metadata_listener_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        expect { listener.on_file_metadata_updated(event) }
+          .not_to change { fake_adapter.saved_resources }
+          .from(be_empty)
+      end
+
+      it 'logs the unexpected message' do
+        expect(Hyrax.logger)
+          .to receive(:warn)
+          .with(/#{metadata.id}/)
+      end
```
cjcolvar, line 42:
```suggestion
      it 'logs the unexpected message' do
        expect(Hyrax.logger)
          .to receive(:warn)
          .with(/#{metadata.id}/)
        listener.on_file_metadata_updated(event)
      end
```

cjcolvar (commented):
These two suggested changes should clear up the failing tests.

cjcolvar (approved):

PR #5383, opened 2022-01-26
add member ids for AdministrativeSets

elrayle:
Adds members to AdministrativeSets allowing for works to be added to an admin set.  Uses the shared testing to confirm that the admin set behaves like it has members.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/administrative_set_form.rb
+++ b/app/forms/hyrax/forms/administrative_set_form.rb
@@ -9,6 +9,8 @@ module Hyrax
       property :title, required: true, primary: true
       property :description, primary: true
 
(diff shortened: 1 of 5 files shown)


no-reply:
i'm uncertain whether this addition is useful or just confusing.

in my understanding, membership in admin sets is managed on the member objects, via `#admin_set_id`. i believe that uploading an object via the UI already sets the appropriate admin set via that property on `ResourceForm`, and `SetDefaultAdminSet`.

i think there are fairly extensive tests for the admin set behavior in the `WorksControllerBehavior` specs?

can you clarify the motivation for this change?

elrayle:
This is not needed.

PR #5386, opened 2022-01-26
ValkyrieIngestJob: don't publish stale metadata

dunn:
Publishing the un-persisted version of the FileMetadata meant failing to preserve characterization information.

@samvera/hyrax-code-reviewers


no-reply (approved):

no-reply:
i think this will work if rebased on #5391

no-reply (approved):

PR #5389, opened 2022-01-26
prepopulate `AdministrativeSetForm#member_ids` for display in UI

no-reply:
when loading the admin set form, we want to populate the member ids.

these are tracked by an inverse reference property `admin_set_id` on the
members, so we define a prepopulator to run a query on that.

i found that wasn't quite enough, because `Wings` wasn't supporting inverse
lookup for admin sets. since the admin set RDF property is configurable, and the
index terms on ActiveFedora's side are derived from the configured property's
qname, we need to do the same on query. the solution for this problem is a
little hacky, but gets the job done; maybe it can be refactored later?

cc: @elrayle 
@samvera/hyrax-code-reviewers


elrayle (approved):
Looks good.

PR #5392, opened 2022-01-27
a first hack at editing admin sets using Valkyrie

no-reply:
working on a transactional approach
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/admin/admin_sets_controller.rb
+++ b/app/controllers/hyrax/admin/admin_sets_controller.rb
@@ -64,19 +64,9 @@ module Hyrax
     def update
       case @admin_set
       when Valkyrie::Resource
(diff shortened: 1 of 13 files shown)


elrayle (commented):
I fixed the rubocop errors.  Tests pass.  Is there anything else making this a draft?  It all looks good to me.

no-reply:
i think this could still use a little more attention, but i don't see any reason not to merge it until then.

no-reply:
@elrayle it seems like you are also doing this, but i've been testing this by switching the controller's `load_and_authorize_resource` arguments to include `class: 'Hyrax::AdministrativeSet'`. at that point, the existing tests seem solid for testing Valkyrie handling.

elrayle (approved):
Looks great.  This really moves admin sets along.

PR #5410, opened 2022-01-28
deprecate #source_model in permission template; use #source instead

### LINKED ISSUE #5406: Valkyrie: Hyrax::PermissionTemplate #source_model uses ActiveFedora::Base.find
### Descriptive summary

Hyrax::PermissionTemplate should use the Hyrax.query_service to find and load the object.  This will work for both ActiveFedora wings adapter and for other Valkyrie adapters.

The error is seen when editing an admin set.  The view `hyrax/admin/admin_sets/_form_visibility.html.erb` gets access to `Hyrax::PermissionTemplate` thru the `Hyrax::PermissionTemplateForm`.  It tries to retrieve the model from `Hyrax::PermissionTemplate #source_model`.

### Rationale

PermissionTemplate methods should work for both ActiveFedora and Valkyrie

### Expected behavior

Editing a Hyrax::AdministrativeSet and clicking the Participants tab should load the tab in the edit form.

### Actual behavior

Editing a Hyrax::AdministrativeSet and clicking the Participants tab gets an `Hyrax::ObjectNotFoundError` while loading the tab in the edit form.

image

#### Prerequisites:

Edit `/config/initializers/hyrax.rb` and set:
1. set the admin set model
```
```
2. restart rails server

#### To reproduce:

3. select type: Admin Set
3. click button: Create Collection
3. set Title: My Admin Set (or other meaningful name)
4. click Save (this loads the editor for the admin set)
5. click the Participants tab

The error will occur as the tab tries to load.

NOTE: There may be other errors once this error is resolved.

### Related work

elrayle:
Fixes #5406

The issue is addressed by changes in `PermissionTemplateForm`
* changing to use `PermissionTemplate#source` instead of `PermissionTemplate#source_method`
* calling `PermissionTemplate#reset_access_controls_for(collection: source)` instead of `source_model.reset_access_controls!` in `#update_access`.

### Add note suggesting to use `PermissionTemplate#source` instead of `PermissionTemplate#source_method`

`PermissionTemplate#source_method` uses `ActiveFedora::Base.find` to get an instance based on `source_id`.  It always returns an `ActiveFedora::Base`.

`PermissionTemplate#source` uses `Hyrax.query_service.find_by` to get an instance based on `source_id`.  It always returns a `Valkyrie::Resource`.

The only place `#source_method` is called within Hyrax code is `Hyrax::Forms::PermissionTemplateForm`.  Usages include:
* #source_model - makes this method available from PermissionTemplateForm through delegation.  This was moved to a method to allow for note recommending use of #source
* #reset_access_controls! - This method calls `source_model.reset_access_controls!` which requires an ActiveFedora::Base object.  This is a deprecated method.  It was not changed other than to update the delegation warning to recommend use of `PermissionTemplate#reset_access_controls_for` instead of `PermissionTempate#reset_access_controls.
* #update_access - This called `source_model.reset_access_controls!` which requires an ActiveFedora::Base object provided by `source_model`.  It was updated to call `PermissionTemplate #reset_access_controls_for` which can receive both ActiveFedora::Base or Valkyrie::Resource objects.  So it can be (and is) passed `source`.

### Deprecate `PermissionTemplate#reset_access_controls`, use `PermissionTemplate#reset_access_controls_for` instead

Similarly, `Hyrax::PermissionTemplate#reset_access_controls` was only called from `Hyrax::Forms::PermissionTemplateForm`.  It is worth noting that the replacement `PermissionTemplate#reset_access_controls_for` is already in use in all other classes that want to reset access.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/permission_template_form.rb
+++ b/app/forms/hyrax/forms/permission_template_form.rb
@@ -7,16 +7,12 @@ module Hyrax
       self.model_class = PermissionTemplate
       self.terms = []
       delegate :access_grants, :access_grants_attributes=, :release_date, :release_period, :visibility, to: :model
-      delegate :available_workflows, :active_workflow, :source_model, to: :model
-
-      # @return [#to_s] the primary key of the associated admin_set or collection
-      # def source_id (because you might come looking for this method)
-      delegate :id, to: :source_model, prefix: :source
+      delegate :available_workflows, :active_workflow, :source, :source_id, to: :model
 
       ##
-      # @deprecated  use PermissionTemplate#reset_access_controls instead.
+      # @deprecated  use PermissionTemplate#reset_access_controls_for instead.
       def reset_access_controls!
-        Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls instead.")
+        Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls_for instead.")
         source_model.reset_access_controls!
       end
 
@@ -73,8 +69,8 @@ module Hyrax
       # Copy this access to the permissions of the Admin Set or Collection and to
       # the WorkflowResponsibilities of the active workflow if this is an Admin Set
       def update_access(remove_agent: false)
-        source_model.reset_access_controls!
-        update_workflow_responsibilities(remove_agent: remove_agent) if source_model.is_a?(AdminSet)
+        model.reset_access_controls_for(collection: source)
+        update_workflow_responsibilities(remove_agent: remove_agent) if source.is_a?(Hyrax::AdministrativeSet)
       end
 
       # This method is used to revoke access to a Collection or Admin Set and its workflows
@@ -83,6 +79,18 @@ module Hyrax
       def remove_access!(permission_template_access)
         construct_attributes_from_template_access!(permission_template_access)
         update_access(remove_agent: true)
+      end
+
+      ##
+      # A bit of an analogue for a `belongs_to :source_model` as it crosses from Fedora to the DB
+      # @return [AdminSet, ::Collection]
+      # @raise [Hyrax::ObjectNotFoundError] when neither an AdminSet or Collection is found
+      # @note This method will eventually be replaced by #source which returns a Hyrax::Resource
+      #   object.  Many methods are equally able to process both Hyrax::Resource and
+      #   ActiveFedora::Base.  Only call this method if you need the ActiveFedora::Base object.
+      # @see #source
+      def source_model # rubocop:disable Rails/Delegate
+        model.source_model
       end
 
       private
--- a/app/models/admin_set.rb
+++ b/app/models/admin_set.rb
@@ -87,12 +87,12 @@ class AdminSet < ActiveFedora::Base
   end
 
   ##
-  # @deprecated use PermissionTemplate#reset_access_controls instead
+  # @deprecated use PermissionTemplate#reset_access_controls_for instead
   #
   # Calculate and update who should have edit access based on who
   # has "manage" access in the PermissionTemplateAccess
   def reset_access_controls!
-    Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls instead.")
+    Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls_for instead.")
 
     permission_template.reset_access_controls_for(collection: self)
   end
--- a/app/models/hyrax/permission_template.rb
+++ b/app/models/hyrax/permission_template.rb
@@ -4,7 +4,7 @@ module Hyrax
   # Holds policy data about the workflow and permissions applied objects when
   # they are deposited through an Administrative Set or a Collection. Each
   # template record has a {#source} (through {#source_id}); the template's
-  # rules inform the behavior of objects deposited through that {#source_model}.
+  # rules inform the behavior of objects deposited through that {#source}.
   #
   # The {PermissionTemplate} specifies:
   #
@@ -81,6 +81,10 @@ module Hyrax
     # A bit of an analogue for a `belongs_to :source_model` as it crosses from Fedora to the DB
     # @return [AdminSet, ::Collection]
     # @raise [Hyrax::ObjectNotFoundError] when neither an AdminSet or Collection is found
+    # @note This method will eventually be replaced by #source which returns a Hyrax::Resource
+    #   object.  Many methods are equally able to process both Hyrax::Resource and
+    #   ActiveFedora::Base.  Only call this method if you need the ActiveFedora::Base object.
+    # @see #source
     def source_model
       ActiveFedora::Base.find(source_id)
     rescue ActiveFedora::ObjectNotFoundError
@@ -88,11 +92,11 @@ module Hyrax
     end
 
     # A bit of an analogue for a `belongs_to :admin_set` as it crosses from Fedora to the DB
-    # @deprecated Use #source_model instead
+    # @deprecated Use #source instead
     # @return [AdminSet]
     # @raise [Hyrax::ObjectNotFoundError] when the we cannot find the AdminSet
     def admin_set
-      Deprecation.warn('Use #source_model instead')
+      Deprecation.warn("#admin_set is deprecated; use #source instead.")
       return AdminSet.find(source_id) if AdminSet.exists?(source_id)
       raise Hyrax::ObjectNotFoundError
     rescue ActiveFedora::ActiveFedoraError # TODO: remove the rescue when active_fedora issue #1276 is fixed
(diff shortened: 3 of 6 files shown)


app/models/hyrax/permission_template.rb
```diff
@@ -79,11 +79,16 @@ def source
...
     # A bit of an analogue for a `belongs_to :source_model` as it crosses from Fedora to the DB
-    # @return [AdminSet, ::Collection]
-    # @raise [Hyrax::ObjectNotFoundError] when neither an AdminSet or Collection is found
+    # @return [AdminSet, ::Collection, Hyrax::AdministrativeSet, Hyrax::PcmdCollection]
+    # @raise [Hyrax::ObjectNotFoundError] when neither an admin set or collection is found
     def source_model
-      ActiveFedora::Base.find(source_id)
-    rescue ActiveFedora::ObjectNotFoundError
+      if Hyrax.config.disable_wings
+        source
```
elrayle, line 86: The `#source` method just above uses `query_service.find_by(:id)`.  So I didn't repeat that code here.  It seems likely that once fully Valkyrized, `#source_model` will be deprecated and replaced by `#source`.

app/models/hyrax/permission_template.rb
```diff
@@ -79,11 +79,16 @@ def source
...
+      else
+        Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: source_id,
+                                                         use_valkyrie: Hyrax.config.use_valkyrie?)
+      end
```
elrayle, line 90: When `use_valkyrie?` is `false`, `find_by_alternate_identifier` will return an `AdminSet` or `Collection` (ActiveFedora).  Otherwise, this returns a `Hyrax::AdministrativeSet` or `Hyrax::PcdmCollection` (Valkyrie).

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -73,8 +73,13 @@ def update(attributes)
       # Copy this access to the permissions of the Admin Set or Collection and to
       # the WorkflowResponsibilities of the active workflow if this is an Admin Set
       def update_access(remove_agent: false)
-        source_model.reset_access_controls!
-        update_workflow_responsibilities(remove_agent: remove_agent) if source_model.is_a?(AdminSet)
+        case source_model
+        when Valkyrie::Resource
+          source_model.permission_manager.acl.save
+        else
+          source_model.reset_access_controls!
```
no-reply, line 80: isn't there an existing pattern for replacing `reset_access_controls!`?

i think it might be `reset_access_controls_for(collection: source)`, established in db58e27e787f87dd930d8ac7b04e28741712d90c

spec/controllers/hyrax/admin/permission_template_accesses_controller_spec.rb
```diff
@@ -4,6 +4,7 @@
   let(:hyrax) { Hyrax::Engine.routes.url_helpers }
   let(:permission_template_access) { FactoryBot.create(:permission_template_access) }
   let(:source_id) { permission_template_access.permission_template.source_id }
+  let(:user_liz) { FactoryBot.create(:user, email: 'liz@example.com') }
```
elrayle, line 7: Without a FactoryBot created user, some of the actions did not have their methods executed because they failed `load_and_authorize_resource` in the class.

app/forms/hyrax/forms/permission_template_form.rb
elrayle: Switched to this pattern.

app/models/hyrax/permission_template.rb
no-reply: if this is to be deprecated (which seems right to me) why is it changing? would it be okay to switch to using `source` uniformly in `PermissionTemplateForm` instead?

app/models/hyrax/permission_template.rb
```diff
@@ -225,7 +222,7 @@ def read_groups
     #   visibility when applying permission template ACLs
     # @return [Boolean]
     def reset_access_controls(interpret_visibility: false)
-      reset_access_controls_for(collection: source_model,
+      reset_access_controls_for(collection: source,
```
no-reply, line 225: this shouldn't change, i think. it was implemented to ensure the implementation for `reset_access_controls` stayed the same for AF models, after the refactor.

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -7,16 +7,16 @@ class PermissionTemplateForm
       self.model_class = PermissionTemplate
       self.terms = []
       delegate :access_grants, :access_grants_attributes=, :release_date, :release_period, :visibility, to: :model
-      delegate :available_workflows, :active_workflow, :source_model, to: :model
+      delegate :available_workflows, :active_workflow, :source, to: :model
```
elrayle, line 10: `:source_model` is still available.  It is defined as a method to allow for a deprecation message.

app/models/hyrax/permission_template.rb
elrayle: I can't see any reason not to deprecate `#source_model` now.  The 3rd commit does that.  The description in this PR identifies the uses of `#source_model` and how they were handled as part of the deprecation.

app/models/hyrax/permission_template.rb
elrayle: I changed this back to use `source_model`.

app/models/admin_set.rb
```diff
@@ -87,12 +87,12 @@ def active_workflow
...
 
   ##
-  # @deprecated use PermissionTemplate#reset_access_controls instead
+  # @deprecated use PermissionTemplate#reset_access_controls_for instead
   #
   # Calculate and update who should have edit access based on who
   # has "manage" access in the PermissionTemplateAccess
   def reset_access_controls!
-    Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls instead.")
+    Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls_for instead.")
```
elrayle, line 95: There are a number of deprecated methods that were updated to use `reset_access_controls_for` but the deprecation message said to use `reset_access_controls`.  These were updated.

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -7,16 +7,12 @@ class PermissionTemplateForm
       self.model_class = PermissionTemplate
       self.terms = []
       delegate :access_grants, :access_grants_attributes=, :release_date, :release_period, :visibility, to: :model
-      delegate :available_workflows, :active_workflow, :source_model, to: :model
-
-      # @return [#to_s] the primary key of the associated admin_set or collection
-      # def source_id (because you might come looking for this method)
-      delegate :id, to: :source_model, prefix: :source
+      delegate :available_workflows, :active_workflow, :source, :source_id, to: :model
```
elrayle, line 10: The previous delegate to get the `id` from `source_model` resulted in an unnecessary population of the `source_model` just to get the `id` which is already available from the template using `source_id`.  So this removes the original delation and adds a direct delegate to `:source_id`.

NOTE: There is still access to the `source_model` in the template even though the delegation was removed.  There is a method named `#source_model` defined later in this class that include a note recommending use of `#source` unless the ActiveFedora model is really needed.

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -7,16 +7,12 @@ class PermissionTemplateForm
...
 
       ##
-      # @deprecated  use PermissionTemplate#reset_access_controls instead.
+      # @deprecated  use PermissionTemplate#reset_access_controls_for instead.
       def reset_access_controls!
-        Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls instead.")
+        Deprecation.warn("reset_access_controls! is deprecated; use PermissionTemplate#reset_access_controls_for instead.")
```
elrayle, line 15: The long term method that should be used is `#reset_access_controls_for`, so the deprecation notice was updated.

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -73,8 +69,8 @@ def update(attributes)
       # Copy this access to the permissions of the Admin Set or Collection and to
       # the WorkflowResponsibilities of the active workflow if this is an Admin Set
       def update_access(remove_agent: false)
-        source_model.reset_access_controls!
-        update_workflow_responsibilities(remove_agent: remove_agent) if source_model.is_a?(AdminSet)
+        model.reset_access_controls_for(collection: source)
```
elrayle, line 72: This is part of the fix of the issue.  `source_model` uses `ActiveFedora::Base.find` which is not available when using Valkyrie postgres adapter.  `source` uses `Hyrax.query_service.find_by`.

app/forms/hyrax/forms/permission_template_form.rb
```diff
@@ -85,6 +81,18 @@ def remove_access!(permission_template_access)
...
 
+      ##
+      # A bit of an analogue for a `belongs_to :source_model` as it crosses from Fedora to the DB
+      # @return [AdminSet, ::Collection]
+      # @raise [Hyrax::ObjectNotFoundError] when neither an AdminSet or Collection is found
+      # @note This method will eventually be replaced by #source which returns a Hyrax::Resource
+      #   object.  Many methods are equally able to process both Hyrax::Resource and
+      #   ActiveFedora::Base.  Only call this method if you need the ActiveFedora::Base object.
+      # @see #source
+      def source_model # rubocop:disable Rails/Delegate
```
elrayle, line 92: This is defined here instead of as a delegate to allow the note encouraging use of `#source` instead of `#source_model`.

no-reply (approved):

no-reply:
thank you for working through this one! i really like how it shaped up and hope you do too!

PR #5414, opened 2022-01-28
Fix nokogiri on arm64 alpine

dlpierce:
The nokogori gem when used in alpine linux on the arm64 platform does not provide compatible precompiled binaries. The gcompat package allows the binaries compoiled for glibc to work.

* Add gcompat package to the Dockerfile in the hyrax-base image

* `docker-compose build` is successful on an arm64 computer such as an M1 Mac

@samvera/hyrax-code-reviewers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ ARG EXTRA_APK_PACKAGES="git"
 RUN apk --no-cache upgrade && \
   apk --no-cache add build-base \
   curl \
+  gcompat \
   imagemagick \
   tzdata \
   nodejs \


marrus-sh:
when i tried to get glibc to work in alpine linux (as part of #5314) i found that the gcompat package was not sufficient, and had to go with the more involved solution here: https://github.com/samvera/hyrax/blob/3d462760fe35e45e4ecff8f34fce43616f4e3922/Dockerfile#L29-L101

if `gcompat` works for nokogiri, good! but i figured i’d mention it here in case it doesn’t…

i know in conversations during the tech call @no-reply mentioned possibly revisiting the decision to use Alpine Linux for everything, and if it is posing a problem for M1 development, that might be a compelling reason for doing so?

dlpierce:
Using gcompat is nokogiri's [suggested solution](https://nokogiri.org/tutorials/installing_nokogiri.html#linux-musl-error-loading-shared-library), but I would be happy switching away from alpine too.

no-reply:
let's move forward with `gcompat` for the moment. if it's not sufficient, we can always consider a bigger change later.

no-reply (approved):

PR #5417, opened 2022-01-28
Clean collection page css

### LINKED ISSUE #5348: Collection page cleanup
### Descriptive summary

General clean up on the Collection page

### Actual behavior

image

1. Do this
2. Then do this...

### Related work

Dananji:
Fixes #5348

* Override default `flex-wrap: wrap` injected from `.row` for the horizontal layout of the `.search-form` when screen width > 768px (there might be a better way to achieve this)
* Override `word-wrap: break-word` leaking from `div.card.collections-panel-wrapper` in `.search-form`

  Before: 
  wordbreak-bl7
  After:
  wordbreak-fix

* Replace `input-group-btn` class with `input-group-append`

  Before: 
  searchform-bl7
  After:
  searchform-fix

* Mainly look at collections and works page
* Check other pages where `search_form` is being used

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/catalog/_search_form.html.erb
+++ b/app/views/catalog/_search_form.html.erb
@@ -7,10 +7,10 @@
       <%= t("hyrax.search.form.q.label", application_name: application_name) %>
     </label>
 
-    <div class="input-group col-sm-9">
+    <div class="input-group mb-2 col-sm-9">
       <%= text_field_tag :q, current_search_parameters , class: "q form-control", id: "search-field-header", placeholder: t("hyrax.search.form.q.placeholder") %>
 
-      <div class="input-group-btn">
+      <div class="input-group-append">
         <button type="submit" class="btn btn-primary" id="search-submit-header">
           <%= t('hyrax.search.button.html') %>
         </button>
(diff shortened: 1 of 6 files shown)


app/views/catalog/_search_form.html.erb
```diff
@@ -10,7 +10,7 @@
     <div class="input-group col-sm-9">
       <%= text_field_tag :q, current_search_parameters , class: "q form-control", id: "search-field-header", placeholder: t("hyrax.search.form.q.placeholder") %>
 
-      <div class="input-group-btn">
+      <div class="input-group-append">
```
adamjarling, line 13: @Dananji Looks like this is a nested `input-group`, which could be messing with things?  I wonder if we should approach this by looking at the markup surrounding the search forms themselves to see if they're aligned with BS4?

adamjarling: Maybe this route?  https://getbootstrap.com/docs/4.0/components/forms/#inline-forms

Dananji:
Fixed the classes for the forms, references;
- https://getbootstrap.com/docs/4.0/components/forms/
- https://getbootstrap.com/docs/4.0/components/input-group/

adamjarling (approved):
Hey @Dananji This does look cleaner to me.  Thanks for looking into it!

Issue #5331, opened 2022-01-21
Hyrax::PcdmCollection fails to delete


elrayle:
### Descriptive summary

When a collection is created as a Hyrax::PcdmCollection and later deleted, it goes through the process without error, but fails to fully delete.  Viewing Dashboard -> Collections after the delete, the deleted collection is still in the list.

### Rationale

Deleting collections should work whether the collection is an Active Fedora collection or a Valkyrie collection.

### Expected behavior

Deleting a collection deletes the collection and it is no longer listed in Dashboard -> Collections.

### Actual behavior

Deleting a collection goes through all steps and no error was displayed in the UI, but it continues to be listed in Dashboard -> Collections.

_NOTE: Initial debugging appears to show that the collection was deleted from the data store, but not solr.  Need to confirm this._

#### Prerequisites:

Configure Hyrax:
1. Edit `/config/initializers/hyrax.rb` and set:
```
  config.collection_model = "Hyrax::PcdmCollection"
  config.query_index_from_valkyrie = true
  config.index_adapter = :solr_index
```
2. restart rails app
3. Make sure regular `hyrax-development` and `hyrax-valkyrie-dev` solr cores are running .  _Your core names may vary depending on how you start them and how they are configured in the app._
4. Make sure you have access to the datastore (i.e. postgres)

Create Collection
1. click button: New Collection
2. select type: any collection type (except Admin Set)
3. click button: Create Collection
3. set Title: DELETE ME (or other meaningful name)
4. click Save

Verify that the collection was created in both solr cores with `q`:  `title_tesim:DELETE ME` (or similar query).
Verify that the collection and its access control were created in the datastore.

#### To reproduce:

1. navigate to Dashboard -> Collections
2. click drop menu in Actions column beside the collection you created
3. choose: Delete collection
4. click button: Delete
5. navigate to Dashboard -> Collections 

Verify that the collection was deleted from both solr cores with `q`:  `title_tesim:DELETE ME` (or similar query).
Verify that the collection and its access control were created in the datastore.

### Related work

See related issues in Issue #5326 UBER: Hyrax-Valkyrization MVP
See Issue #5407 describing the same issue for Hyrax::AdministrativeSets

no-reply:
is this effectively a duplicate of #5407?

hackartisan:
Cannot reproduce in nurax-pg or nurax-dev. This appears to be resolved.

PR #5418, opened 2022-01-28
Hyrax::PcdmCollection delete

### LINKED ISSUE #5407: Valkyrie: Hyrax::AdministrativeSet fails to delete fully
### Descriptive summary

When an admin set is created as a Hyrax::AdminstrativeSet and later deleted, it goes through the process without error, but fails to fully delete. 

### Rationale

Deleting admin sets should work whether the admin set is an Active Fedora admin set or a Valkyrie admin set.

### Expected behavior

Deleting an admin set fully deletes the admin set and its access control from the data store and solr.

### Actual behavior

Deleting an admin set deletes the admin set from the datastore, but fails to delete it from solr and fails to delete its access control from the datastore.

#### Prerequisites:

Configure Hyrax:
```
```
2. restart rails app
3. Make sure regular `hyrax-development` and `hyrax-valkyrie-dev` solr cores are running .  _Your core names may vary depending on how you start them and how they are configured in the app._
4. Make sure you have access to the datastore (i.e. postgres)

NOTE: For testing of nurax-pg, the solr core is typically named nurax-pg.

Create Admin Set
2. select type: Admin Set
3. click button: Create Collection
3. set Title: DELETE ME (or other meaningful name)
4. click Save
5. Copy the id from the URL

Verify that the admin set was created in both solr cores with `q`:  `title_tesim:DELETE ME` (or similar query).
Verify that the admin set and it access control were created in the datastore.

#### To reproduce:

1. Navigate to: Dashboard -> Collections
1. Click on the title of the admin set you want to delete (e.g. DELETE ME)
2. click button: Delete

Verify that the admin set was deleted from both solr cores with `q`:  `title_tesim:DELETE ME` (or similar query).
Verify that the admin set and it's access control were created in the datastore.

### Related work

See Issue #5331 describing the same issue for Hyrax::PcdmCollection

cjcolvar:
Fixes #5407 and #5331

This PR got rather large and the pieces intertwined.  I can split this apart into smaller PRs if necessary.

Proposed changes in this PR:
- Destroy collection by using new transaction `collection_resource.destroy`
- Add listener for `collection.deleted` event which removes references to the collection from member resources
- Add listener for `collection.deleted` event which removes collection resource from the index
- Destroy access control resource in new `DeleteAccessControl` transaction step
- Add new `DeleteAccessControl` transaction step to `AdminSetDestroy`, `CollectionDestroy`, and `WorkDestroy` transactions
- Add `#destroy` method to AccessControlList which deletes the acl resource and publishes a `object.acl.deleted` event
- Add listener for new `object.acl.deleted` event which removes the acl resource from the index

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -168,10 +168,10 @@ module Hyrax
         # leaving id to avoid changing the method's parameters prior to release
         respond_to do |format|
           format.html do
-            redirect_to my_collections_path,
+            redirect_to hyrax.my_collections_path,
                         notice: t('hyrax.dashboard.my.action.collection_delete_success')
           end
-          format.json { head :no_content, location: my_collections_path }
+          format.json { head :no_content, location: hyrax.my_collections_path }
         end
       end
 
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -188,8 +188,7 @@ module Hyrax
       def destroy
         case @collection
         when Valkyrie::Resource
-          Hyrax.persister.delete(resource: @collection)
-          after_destroy(params[:id])
+          valkyrie_destroy
         else
           if @collection.destroy
             after_destroy(params[:id])
@@ -239,6 +238,14 @@ module Hyrax
                         .call(form)
                         .value_or { return after_update_error }
         after_update
+      end
+
+      def valkyrie_destroy
+        if transactions['collection_resource.destroy'].call(@collection).success?
+          after_destroy(params[:id])
+        else
+          after_destroy_error(params[:id])
+        end
       end
 
       def default_collection_type
--- a/app/services/hyrax/listeners/member_cleanup_listener.rb
+++ b/app/services/hyrax/listeners/member_cleanup_listener.rb
@@ -16,7 +16,7 @@ module Hyrax
           begin
             Hyrax.persister.delete(resource: file_set)
             Hyrax.publisher
-                 .publish('object.deleted', object: file_set, id: file_set.id, user: user)
+                 .publish('object.deleted', object: file_set, id: file_set.id, user: event[:user])
           rescue StandardError # we don't uncaught errors looping filesets
             Hyrax.logger.warn "Failed to delete #{file_set.class}:#{file_set.id} " \
                               "during cleanup for resource: #{event[:object]}. " \
--- a/app/services/hyrax/listeners/member_cleanup_listener.rb
+++ b/app/services/hyrax/listeners/member_cleanup_listener.rb
@@ -3,7 +3,7 @@
 module Hyrax
   module Listeners
     ##
-    # Listens for object deleted events and cleans up associated members
+    # Listens for resource deleted events and cleans up associated members
     class MemberCleanupListener
       # Called when 'object.deleted' event is published
       # @param [Dry::Events::Event] event
@@ -21,6 +21,26 @@ module Hyrax
             Hyrax.logger.warn "Failed to delete #{file_set.class}:#{file_set.id} " \
                               "during cleanup for resource: #{event[:object]}. " \
                               'This member may now be orphaned.'
+          end
+        end
+      end
+
+      # Called when 'collection.deleted' event is published
+      # @param [Dry::Events::Event] event
+      # @return [void]
+      def on_collection_deleted(event)
+        return unless event.payload.key?(:collection) # legacy callback
+        return if event[:collection].is_a?(ActiveFedora::Base) # handled by legacy code
+
+        Hyrax.custom_queries.find_members_of(collection: event[:collection]).each do |resource|
+          begin
+            resource.member_of_collection_ids -= [event[:collection].id]
+            Hyrax.persister.save(resource: resource)
+            Hyrax.publisher
+                 .publish('collection.membership.updated', collection: event[:collection], user: event[:user])
+          rescue StandardError
+            Hyrax.logger.warn "Failed to remove collection reference from #{work.class}:#{work.id} " \
+                              "during cleanup for collection: #{event[:collection]}. "
           end
         end
       end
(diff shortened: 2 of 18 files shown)


app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -168,10 +168,10 @@ def after_destroy(_id)
         # leaving id to avoid changing the method's parameters prior to release
         respond_to do |format|
           format.html do
-            redirect_to my_collections_path,
+            redirect_to hyrax.my_collections_path,
```
cjcolvar, line 171: This was needed to fix the controller test that was checking paths being redirected to.

lib/hyrax/transactions/steps/save_access_control.rb
```diff
@@ -18,8 +18,8 @@ class SaveAccessControl
         # @return [Dry::Monads::Result]
         def call(obj)
           return Success(obj) unless obj.respond_to?(:permission_manager)
-          obj.permission_manager&.acl&.save ||
-            (return Failure[:failed_to_save_acl, acl])
+          acl = obj.permission_manager&.acl
+          acl&.save || (return Failure[:failed_to_save_acl, acl])
```
cjcolvar, line 22: I noticed this missing definition of `acl` on line 22 when writing the `delete_access_control` step so I refactored it to match.

app/services/hyrax/listeners/member_cleanup_listener.rb
```diff
@@ -16,14 +16,34 @@ def on_object_deleted(event)
           begin
             Hyrax.persister.delete(resource: file_set)
             Hyrax.publisher
-                 .publish('object.deleted', object: file_set, id: file_set.id, user: user)
+                 .publish('object.deleted', object: file_set, id: file_set.id, user: event[:user])
```
cjcolvar, line 19: Writing the missing test for this listener revealed this bug.

app/controllers/hyrax/dashboard/collections_controller.rb
no-reply: any idea why this popped up now?

lib/hyrax/transactions/steps/save_access_control.rb
no-reply: 👍🏻

lib/hyrax/publisher.rb
```diff
@@ -155,6 +155,10 @@ class Publisher
     # @macro a_registered_event
     register_event('object.acl.updated')
 
+    # @since 3.4.0
+    # @macro a_registered_event
+    register_event('object.acl.deleted')
```
no-reply, line 160: is it worth documenting that this is a delete of the whole ACL resource, not just of a single permission statement?

lib/hyrax/transactions/steps/delete_access_control.rb
```diff
@@ -0,0 +1,32 @@
...
+      # @see https://dry-rb.org/gems/dry-monads/1.0/result/
+      class DeleteAccessControl
+        include Dry::Monads[:result]
+
+        ##
+        # @param [Valkyrie::Resource] obj
+        #
+        # @return [Dry::Monads::Result]
+        def call(obj)
+          return Success(obj) unless obj.respond_to?(:permission_manager)
```
no-reply, line 20: i think it's at least theoretically possible for an object to have an ACL, but not a permission manager. the permission manager is mainly a tool for holding permission changes and passing them around with an object.

it might be worth refactoring to use `AccessControlList(obj)`? (though i'm unsure of how well that interface handles the possibility of the ACL not yet existing...)

lib/hyrax/transactions/steps/delete_access_control.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # Deletes the Hyrax::AccessControlList for any resource with a `#permission_manager`.
+      # If `#permission_manager` is undefined, succeeds.
+      #
+      # @see https://dry-rb.org/gems/dry-monads/1.0/result/
+      class DeleteAccessControl
```
no-reply, line 12: this looks great.

i'm curious whether this logic wouldn't be more effective as a listener on `object/collection.deleted`? the considerations i have in mind are:
  - are there common cases where you'd want to delete an object or collection using the persister, but NOT cleanup this access control? a listener is closer to universal than a transaction step, so this could be a good reason to switch.
  - if the destroy step succeeds, but the object acl deletion fails, will we want to report an error back to the controller? if so, i'd think that's a point in favor of the transaction step; but maybe we'd rather log and continue processing, leaving the ACL orphaned?

i don't want to hold this up on the answer either way. it's easy enough to get this in and rework the implementation if we decide a listener is better.

no-reply (commented):
all my suggestions are additive, and can be addressed in follow-up if that's the way it shakes out.

i would like some clarification about why the redirect target is changing here, but if you can provide some context there, i'm happy to approve.

lib/hyrax/transactions/steps/delete_access_control.rb
cjcolvar: Good point.  I was following the pattern of the save_access_control step.  So I wonder if it should happen the same way too.

lib/hyrax/transactions/steps/delete_access_control.rb
cjcolvar: This code mirrors what is in the save_access_control step so maybe this change could be explored and tested as a change across the two steps in a follow up PR?

lib/hyrax/transactions/steps/delete_access_control.rb
cjcolvar: But I guess if the save ACL step fails then we would want to provide feedback to the user in the controller.

no-reply: my first intuition is that save is different on both counts. for save:
  - yes, we do commonly want to save only the metadata and not the access controls---the need to check access control save is contextual to the transaction.
  - yes, we do want to fail visibly if we've saved an object but not been able to apply requested ACLs.

Issue #5385, opened 2022-01-26
AdministrativeSetForm populates Description in new form with []


elrayle:
### Descriptive summary

When the Create New Administrative Set form first populates, the Description field has `[]`.

### Rationale

Form fields, for the most part, populate as empty for new forms.

### Expected behavior

When a new admin set is being created via the `AdminstrativeSetForm`, the `Description` field should be empty.

### Actual behavior

When a new admin set is being created via the `AdminstrativeSetForm`, the `Description` field is populated with `[]`.

![image](https://user-images.githubusercontent.com/6855473/151238225-a456b881-ee97-4a7b-ae26-36efe54aed26.png)

### Steps to reproduce the behavior

Prerequisites:

Configure Hyrax:
1. Edit `/config/initializers/hyrax.rb` and set:
```
  config.admin_set_model = "Hyrax::AdministrativeSet"
```
2. restart rails app

#### To reproduce:

1. navigate to Dashboard -> Collections
2. click button: New Collection
3. select type: Admin Set
4. click button: Create Collection

Notice that Description field is populated with `[]`.

### Related work

See related issues in Issue #5326 UBER: Hyrax-Valkyrization MVP

elrayle:
NOTE: When the edit form is loaded, it also displays the description as an array (e.g. `["My description"]`

elrayle:
@no-reply PR #5419 looks like it fixes this issue.  Do you see remaining work before closing this?

PR #5419, opened 2022-01-28
AdministrativeSet#description should be single-valued

no-reply:
related to #5385.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
:+1:

PR #5423, opened 2022-01-31
use configured adminset model in search builders

### LINKED ISSUE #5408: Hyrax::AdministrativeSets fail to be listed on Dashboard->Collections
### Descriptive summary

Dashboard -> Collections correctly lists Hyrax::PcdmCollections, but fails to list Hyrax::AdministrativeSets.  Investigation as part of Issue #5334 determined that the AdministrativeSet is created as expected.  It is just failing to be listed.

### Rationale

All collections and admin sets should be listed for Dashboard -> Collections regardless of the model used to create them.

### Expected behavior

After creating a `Hyrax::AdministrativeSet`, it should be listed in Dashboard -> Collections.

### Actual behavior

After creating a `Hyrax::AdministrativeSet`, it is NOT listed in Dashboard -> Collections.

#### Prerequisites:

Configure Hyrax:
```
```
2. restart rails app

#### To reproduce:

Create Admin Set
2. select type: Admin Set
3. set Title: MY ADMIN SET (or other meaningful name)
4. click Save
5. navigate to Dashboard -> Collections

The new admin set should be in the list of collections, but it is not.

### Related work

See Issue #5334 for an investigation of this issue

elrayle:
Fixes #5408

Instead of hardcoding the admin set class, use the configured `Hyrax.config.admin_set_model`.

----

FEEDBACK REQUESTED:

@no-reply @cjcolvar 

This works for pure Valkyrie and pure ActiveFedora apps.  It will not work for any apps that may have some solr docs with model `AdminSet` and some with model `Hyrax::AdministrativeSet`.  

OPTIONS:
* OPTION 1:  Use the change in this PR.  If this is the long term solution, then once apps start to migrate to Valkyrie, this would lead to a need to add a data migration for solr docs to update all docs with model `AdminSet` to have model `Hyrax::AdministrativeSet` instead.
* OPTION 2: Update solr query to allow either model.  Not a simple query.  See notes below.  Seems like the best to prevent the need to migrate solr docs once Valkyrie migrations start in the community.  (Maybe migrations would be expected anyway?)
* OPTION 3: Accept the PR as it is now and create an issue addressing OPTION 2's OR query or creating a migration script for solr docs.  Fixes the functionality required for the MVP while acknowledging the need for a long term plan for how to deal with or prevent the need for solr doc migrations.
* Other options?
 
Seems like there should be a way to include an OR that allows for either model.  I have not been able to determine the way to write that query.  See Issue #5408 for more information on the current query and attempts to make the query work with both models.

I confirmed that with this fix, the admin sets are in the list of collections on Dashboard -> My (and All) Collections for nurax-pg.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/dashboard/collections_search_builder.rb
+++ b/app/search_builders/hyrax/dashboard/collections_search_builder.rb
@@ -19,7 +19,7 @@ module Hyrax
         return if current_ability.admin?
         clauses = [
           '-' + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(depositor: current_user_key),
-          '-' + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::AdminSet.to_s, creator: current_user_key)
+          '-' + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: Hyrax.config.admin_set_model, creator: current_user_key)
         ]
         solr_parameters[:fq] ||= []
         solr_parameters[:fq] += ["(#{clauses.join(' OR ')})"]
--- a/app/search_builders/hyrax/my/collections_search_builder.rb
+++ b/app/search_builders/hyrax/my/collections_search_builder.rb
@@ -12,7 +12,7 @@ class Hyrax::My::CollectionsSearchBuilder < ::Hyrax::CollectionSearchBuilder
   def show_only_collections_deposited_by_current_user(solr_parameters)
     clauses = [
       ActiveFedora::SolrQueryBuilder.construct_query_for_rel(depositor: current_user_key),
-      ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::AdminSet.to_s, creator: current_user_key)
+      ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: Hyrax.config.admin_set_model, creator: current_user_key)
     ]
     solr_parameters[:fq] ||= []
     solr_parameters[:fq] += ["(#{clauses.join(' OR ')})"]


elrayle:
After extensive discussion in Slack, it was decided to go with Option 3, accepting the PR as is and revisiting the question once a migration path is defined for moving from ActiveFedora to non-ActiveFedora Valkyrie persisters (e.g. postgres).

For documentation, there was a suggestion to try the following query, which did return both AdminSets and Hyrax::AdministrativeSets.  But there seem to be items returned for the My -> Collections tab that included admin sets and collections created by other users.  The query may be on the right track, but needs additional work to return the correct results for the current user.

```
        clauses = [
          '-' + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(depositor: current_user_key),
          '-' + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: Hyrax.config.admin_set_model, creator: current_user_key),
          '-' + ActiveFedora::SolrQueryBuilder.construct_query_for_rel(has_model: ::AdminSet.to_s, creator: current_user_key)
        ]
```

elrayle:
More information from the Slack conversation in case we come back to this later:

From @cjcolvar 
> When looking at this with @elrayle last week, I think one thing complicating things is that Hyrax.config.admin_set_model for the valkyrie case is Hyrax::AdministrativeSet which doesn't seem to play well with the raw query that construct_query_for_rel generates.  An alternative is construct_query which defaults to a field query but could be configured to use terms instead which does seem to work.
https://github.com/samvera/active_fedora/blob/511ee837cd9a461021e53d5e20f362b634de39a8/lib/active_fedora/solr_query_builder.rb#L35

no-reply (approved):

PR #5424, opened 2022-02-02
use Hyrax::SolrQueryService for My/Collections

elrayle:
refs #5408

When deployed to nurax-pg, the query to  list my collections in Dashboard -> Collections still did not list admin sets I created.  This reworks the query to use `terms` instead of `raw`.

As an aside, it also removed references to `ActiveFedora::SolrQueryBuilder`.  The calls are replaced with calls to `Hyrax::SolrQueryService` which is a replacement for `ActiveFedora::SolrQueryBuilder`.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Looks good!

PR #5421, opened 2022-01-28
set and index permissions for admin sets

elrayle:
Partially Addresses #5408

There are several issue preventing admin sets from being listed on Dashboard -> Collections.  
* the `valkyrie_create` method in the controller did not pass the user to the transaction creating permissions
* the indexer didn't index permissions
* the search builder didn't include Hyrax::AdministrativeSet in the list of models (selects all of the model set)
* the search builder didn't include Hyrax::AdministrativeSet in the permissions limiter (filters out all without valid permissions)

The first 3 are fixed in this PR.  The last part of the search build issue is not complete.  See Remaining Work below.

----
### Remaining Work

There is still an outstanding issue with the query to retrieve collections and admin sets.

#### Old Query:
```
{"qt"=>"search", "facet.field"=>["visibility_ssi", "suppressed_bsi", "resource_type_sim", "has_model_ssim"], "facet.query"=>[], "facet.pivot"=>["has_model_ssim,collection_type_gid_ssim"], "fq"=>["", "{!terms f=has_model_ssim}AdminSet,Hyrax::AdministrativeSet,Hyrax::PcdmCollection", "-suppressed_bsi:true", "(_query_:\"{!raw f=depositor_ssim}****@hotmail.com\" OR (_query_:\"{!raw f=has_model_ssim}AdminSet\" AND _query_:\"{!raw f=creator_ssim}****@hotmail.com\"))"], "hl.fl"=>[], "rows"=>10, "qf"=>"title_tesim description_tesim creator_tesim keyword_tesim", "facet"=>true, "f.visibility_ssi.facet.limit"=>6, "f.resource_type_sim.facet.limit"=>6, "f.collection_type_gid_ssim.facet.limit"=>6, "f.has_model_ssim.facet.limit"=>6, "sort"=>"title_si asc"}
```

This includes adding the `Hyrax::AdministrativeSet` to the set of models `"{!terms f=has_model_ssim}AdminSet,Hyrax::AdministrativeSet,Hyrax::PcdmCollection"`, but the part that limits admin sets to the current user being the creator does not include `Hyrax::AdministrativeSet`.  I wasn't able to come to a syntax for the query that would allow the model for this part of the query to be `AdminSet` OR `Hyrax::AdministrativeSet`.  `(_query_:\"{!raw f=has_model_ssim}AdminSet\" AND _query_:\"{!raw f=creator_ssim}****@hotmail.com\")`

#### Attempt at a New Query (doesn't work):
```
[...]

PR #5425, opened 2022-02-02
use configured admin_set_class to test admin_set? in solr doc

elrayle:
refs #5408

PR #5421, PR #5422, PR #5423, PR #5424 were all related to producing and querying a solr doc so that admin sets appeared in the list of collections on Dashboard -> Collections.

Once they were in the list, their links were not correct.  The links were created using the collection routes instead of admin set routes.  This was happening because the `SolrDocument` for admin sets was always return `false` for `#admin_set?`.  And this was happening because the method `#admin_set?` was checking for class `::AdminSet`.  This PR updates the method to check for class `Hyrax.config.admin_set_class`.  This config defaults to `AdminSet` allowing for backward compatibility.  And allows for valkyrie apps (e.g. nurax-pg) which are configured to `Hyrax::AdministrativeSets` to correctly identify admin sets.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/solr_document_behavior.rb
+++ b/app/models/concerns/hyrax/solr_document_behavior.rb
@@ -65,7 +65,7 @@ module Hyrax
     ##
     # @return [Boolean]
     def admin_set?
-      hydra_model == ::AdminSet
+      hydra_model == Hyrax.config.admin_set_class
     end
 
     ##


no-reply (approved):

no-reply:
thank you!

PR #5427, opened 2022-02-03
Delete access control object when destroying a resource

cjcolvar:
Split out part of #5418.

Changes proposed in this PR:
- Destroy access control resource in new DeleteAccessControl transaction step
- Add new DeleteAccessControl transaction step to AdminSetDestroy, CollectionDestroy, and WorkDestroy transactions
- Add #destroy method to AccessControlList which deletes the acl resource ~and publishes a object.acl.deleted event~
- ~Register new object.acl.deleted event in Hyrax::Publisher~

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/access_control_list_spec.rb
+++ b/spec/services/hyrax/access_control_list_spec.rb
@@ -164,4 +164,25 @@ RSpec.describe Hyrax::AccessControlList do
       end
     end
   end
+
+  describe "#destroy" do
+    let(:listener) { Hyrax::Specs::SpyListener.new }
+
+    before do
+      acl << permission
+      acl.save
+
+      # Subscribe to events after acl has been persisted
+      Hyrax.publisher.subscribe(listener)
+    end
+
+    after { Hyrax.publisher.unsubscribe(listener) }
+
+    it 'deletes the acl resource' do
+      expect { acl.destroy }
+        .to change { Hyrax::AccessControl.for(resource: resource, query_service: acl.query_service).persisted? }
+        .from(true)
+        .to(false)
+    end
+  end
 end
--- a/.rubocop_fixme.yml
+++ b/.rubocop_fixme.yml
@@ -114,6 +114,7 @@ RSpec/AnyInstance:
     - 'spec/controllers/hyrax/batch_edits_controller_spec.rb'
     - 'spec/controllers/hyrax/stats_controller_spec.rb'
     - 'spec/controllers/hyrax/users_controller_spec.rb'
+    - 'spec/hyrax/transactions/steps/delete_access_control_spec.rb'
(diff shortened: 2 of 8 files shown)


lib/hyrax/publisher.rb
```diff
@@ -155,6 +155,10 @@ class Publisher
     # @macro a_registered_event
     register_event('object.acl.updated')
 
+    # @since 3.4.0
+    # @macro a_registered_event
+    register_event('object.acl.deleted')
```
no-reply, line 160: seeing `object.acl` in this context is making me wonder if this is the right event namespace. we're going to publish this for `AdminSet` resources, which are not Objects. i was just encountering a similar issue that motivated to #5429, so i'm curious.

i know this is borrowing from the `object.acl.updated` event above, but i'm wondering whether both shouldn't change...

or i guess this part might be moot in light of https://github.com/samvera/hyrax/pull/5428#issuecomment-1028547693

cjcolvar: I was thinking the same thing, so I think `object.acl.updated` should get explanded.  But outside the scope of this PR.  I'll remove this event since I closed #5428.

no-reply:
sorry i missed the overlap between this and #5418.

a couple of comments over there are relavant here.

https://github.com/samvera/hyrax/pull/5418/files/8840c4e208d48939dbfab1f886ec689625a81679#diff-e2e9a7e504ecdd51e9b0a9892efa1715dc2c9864bb28221b4936cd52188fa3f1
https://github.com/samvera/hyrax/pull/5418/files/8840c4e208d48939dbfab1f886ec689625a81679#r798749816

no-reply (approved):
both the linked comments are additive, and i'm keen to get this in. 👍🏻

cjcolvar:
I'm going to merge this now since there are other PRs (continuing the breakdown of #5418 into smaller PRs) I have coming which will build on this.  So I could address those comments later.

spec/services/hyrax/access_control_list_spec.rb
```diff
@@ -164,4 +164,25 @@
...
+
+      # Subscribe to events after acl has been persisted
+      Hyrax.publisher.subscribe(listener)
+    end
+
+    after { Hyrax.publisher.unsubscribe(listener) }
+
+    it 'deletes the acl resource' do
+      expect { acl.destroy }
+        .to change { Hyrax::AccessControl.for(resource: resource, query_service: acl.query_service).persisted? }
```
elrayle, line 183: I don't see the listener message defined here, so I'm guessing that the check for the message will happen in another PR?

cjcolvar: The event and the publishing of it was removed from this PR per discussion on https://github.com/samvera/hyrax/pull/5428

PR #5428, opened 2022-02-03
Handle object.acl.deleted event by deleting ACL from index

cjcolvar:
Split out part of https://github.com/samvera/hyrax/pull/5418.

Changes proposed in this PR:
- Add listener for new object.acl.deleted event which removes the acl resource from the index

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/listeners/acl_index_listener.rb
+++ b/app/services/hyrax/listeners/acl_index_listener.rb
@@ -18,6 +18,16 @@ module Hyrax
         return unless event[:result] == :success # do nothing on failure
         Hyrax.index_adapter.save(resource: event[:acl].resource)
       end
+
+      ##
+      # Deletes the resource for the ACL from the index.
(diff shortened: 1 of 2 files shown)


no-reply:
do we index ACLs separately from the Object/Collection/File they relate to?

cjcolvar:
I thought maybe we did, but I'm not seeing it now.  I'm confused as to why I did this before.
If this isn't needed then I can remove the event being registered and published in #5427.

no-reply:
> I thought maybe we did, but I'm not seeing it now. I'm confused as to why I did this before. If this isn't needed then I can remove the event being registered and published in #5427.

i think ActiveFedora does (to power the query DSL), but should clean up after itself. i'm 96% sure we don't index valkyrie ACL objects, and that the relevant queries are handled are against the index documents for the relevant pcdm resource.

PR #5430, opened 2022-02-03
Destroy PcdmCollection using a transaction

cjcolvar:
Split out part of #5418

Proposed changes in this PR:
- Destroy collection by using new transaction collection_resource.destroy

This builds on top of #5427 so will need to be rebased once that gets merged.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -168,10 +168,10 @@ module Hyrax
         # leaving id to avoid changing the method's parameters prior to release
         respond_to do |format|
           format.html do
-            redirect_to my_collections_path,
+            redirect_to hyrax.my_collections_path,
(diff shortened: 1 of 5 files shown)


app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -168,10 +168,10 @@ def after_destroy(_id)
         # leaving id to avoid changing the method's parameters prior to release
         respond_to do |format|
           format.html do
-            redirect_to my_collections_path,
+            redirect_to hyrax.my_collections_path,
```
cjcolvar, line 171: Without this change the controller test that I turned back on would redirect to: `http://test.host/assets?action=index&controller=hyrax%2Fmy%2Fcollections&locale=en`
I'm not sure why the analogous controller test for the non-valkyrie based controller doesn't have the same problem, but this does fix the problem.

elrayle (approved):
This is looking good.

PR #5432, opened 2022-02-03
Remove collection from index when deleted event published

cjcolvar:
Part of break out of #5418 

This PR addresses removing the collection resource from the index when a collection.deleted event is published.

@samvera/hyrax-code-reviewers


elrayle (approved):
Looking good.

PR #5434, opened 2022-02-03
Ensure Failure doesn't fail to return

cjcolvar:
Part of break out of #5418 

Bugfix found when working on the delete access control transaction step.

@samvera/hyrax-code-reviewers


elrayle (approved):
Thanks for all the delete PRs.

PR #5437, opened 2022-02-08
add documentation of the relationship between admin sets and works

elrayle:
Adds documentation of the relationship between admin sets and works.  Also establishes a pattern for documenting other relationships in the model.

Basic guidelines:
* define the relationship both directions in both classes involved in the relationship
* include cardinality of each direction of the relationship (e.g. (1..m): An admin set can have many works.)
* if a class includes an attribute storing the relationship
  * include code for how to set in an instance of this class
  * include code for how to find the related resource(s) given an instance of this class
* if a class does not include an attribute storing the relationship
  * include a note to see the other class for how to set the relationship
  * include code for how to find the related resource(s) given an instance of this class
* include information on where the relationship is tested
* include @see for 
  * related class 
  * query service method(s) used to find related resource(s)
  * path to tests of the relationship

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/work.rb
+++ b/app/models/hyrax/work.rb
@@ -4,6 +4,29 @@ module Hyrax
   ##
   # Valkyrie model for `Work` domain objects in the Hydra Works model.
   #
+  # *Relationships:*
+  # * Administrative Set and Work
+  #   * Administrative Set to Work (1..m): An admin set can have many works.  This relationship
+  #     is defined by the inverse relationship stored in the work's attribute `:admin_set_id`.
+  #     * See Hyrax::AdministrativeSet for code to get works in an admin set.
+  #   * Work to Administrative Set (1..1):  A work must be in one and only one admin set.  The
+  #     relationship to the admin set is defined in the work's attribute `:admin_set_id`.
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
This is really clear and understandable!  I like how it gives code tips in the model class.

app/models/hyrax/work.rb
```diff
@@ -4,6 +4,29 @@ module Hyrax
...
+  #       for tests of this relationship.
+  # * Collection and Work (TBD)
+  # * Work and Work (TBD)
+  # * Work and File Set (TBD)
+  #
+  # @see Hyrax::AdministrativeSet
+  # @see Valkyrie query adapter's #find_by
+  # @see /lib/hyrax/specs/shared_specs/hydra_works.rb
+  #
+  # TODO: The description in Hydra::Works Shared Modeling is out of date and uses
```
no-reply, line 26:
```suggestion
  # @todo The description in Hydra::Works Shared Modeling is out of date and uses
```

no-reply:
looks great. i think we could work on the YARD formatting, but i'm not keen to hold it up on that. 💯

PR #5439, opened 2022-02-09
add ACL key handling for Hyrax::Group

no-reply:
knowledge about how the ACL keys are handled for Groups is spread throughout the
codebase. treating this information as an attribute on the group seems cleaner.

`Hyrax::Group#user_key` is provided for easy compatibility with `::User`.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/group.rb
+++ b/app/models/hyrax/group.rb
@@ -7,11 +7,25 @@ module Hyrax
       DEFAULT_NAME_PREFIX
     end
 
+    ##
+    # @return [Hyrax::Group]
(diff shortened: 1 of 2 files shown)


cjcolvar (commented):
This looks good to me.  I think it would be nice to have some simple tests that illustrate and test the functionality of `Hyrax::Group.from_key` and `#group_key`.

no-reply:
good call. to date there aren't ANY tests for Hyrax::Group.

i'll find a bit of time to write them ASAP.

PR #5440, opened 2022-02-11
add navigator to find parent work for child works and filesets

elrayle:
This provides a consistent API for navigating from a child work or fileset to the parent work without the caller needing to know the attribute storing the relationship or the directionality of the relationship.

This is consistent with other navigator custom queries found in [app/services/hyrax/custom_queries/navigators](https://github.com/samvera/hyrax/tree/main/app/services/hyrax/custom_queries/navigators).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/custom_queries/navigators/parent_work_navigator.rb
+++ b/app/services/hyrax/custom_queries/navigators/parent_work_navigator.rb
@@ -7,7 +7,7 @@ module Hyrax
       # Navigate from a resource to it's parent work.
       #
       # @see https://github.com/samvera/valkyrie/wiki/Queries#custom-queries
-      # @since 3.0.0
+      # @since 3.4.0
       class ParentWorkNavigator
         # Define the queries that can be fulfilled by this navigator.
         def self.queries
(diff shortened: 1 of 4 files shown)


spec/services/hyrax/custom_queries/navigators/parent_work_navigator_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    context 'when no parents' do
+      let(:member_ids) { [] }
+      it 'returns nil' do
+        expect(custom_query_service.find_parent_work(resource: child_work1)).to be_nil
+        expect(custom_query_service.find_parent_work(resource: fileset1)).to be_nil
+      end
+    end
+  end
+
+  describe '#find_parent_work_id' do
```
cjcolvar, line 67: I think it would be useful to add another small test here for when there are no parents.

app/services/hyrax/custom_queries/navigators/parent_work_navigator.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module CustomQueries
+    module Navigators
+      ##
+      # Navigate from a resource to it's parent work.
+      #
+      # @see https://github.com/samvera/valkyrie/wiki/Queries#custom-queries
+      # @since 3.0.0
```
cjcolvar, line 10: Should this be 3.4.0?
```suggestion
      # @since 3.4.0
```

cjcolvar (approved):
Looks good!

PR #5441, opened 2022-02-11
add Work-to-Work relationship documentation

elrayle:
Adds documentation for the Work-to-Work relationship.

Also refactors the layout of each relationship to avoid repeating statements and code and to make each characteristic (where defined, where tested, each direction of the relationship) easier to see at a glance.  Code is separated out into code blocks instead of being inline.

### In the work

image

### In the admin set

image

### DIFF
--- a/app/models/hyrax/work.rb
+++ b/app/models/hyrax/work.rb
@@ -4,52 +4,59 @@ module Hyrax
   ##
   # Valkyrie model for `Work` domain objects in the Hydra Works model.
   #
-  # *Relationships:*
+  # ## Relationships
   #
-  # Administrative Set and Work
+  # ### Administrative Set and Work
   #
-  # * <b>Defined:</b> The relationship is defined by the work's `:admin_set_id` attribute.
-  # * <b>Tested:</b> The relationship is tested in shared spec 'a Hyrax::Work' by testing
+  # * Defined: The relationship is defined by the work's `:admin_set_id` attribute.
+  # * Tested: The relationship is tested in shared spec `'a Hyrax::Work'` by testing
   #   `#admin_set_id`.  Shared specs are defined in /lib/hyrax/specs/shared_specs/hydra_works.rb.
-  # * <b>Administrative Set to Work:</b> (1..m) An admin set can have many works.
+  # * Administrative Set to Work: (1..m) An admin set can have many works.
   #   * See Hyrax::AdministrativeSet for code to get works in an admin set.
-  # * <b>Work to Administrative Set:</b> (1..1)  A work must be in one and only one admin set.
-  #   * Set admin set for a work using:
+  # * Work to Administrative Set: (1..1)  A work must be in one and only one admin set.
+  #
+  # @example Set admin set for a work:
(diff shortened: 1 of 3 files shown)


cjcolvar (commented):
I think this is looking good overall!

I'm wondering if it is possible to use `*Something*` instead of `<b>Something</b>` and still retrain the same appearance in the generated yard docs.  I think the former is easier to read when reading the yard docs inline in the file.

elrayle:
I don't know why, but yard doesn't always like `*something*`.  It appears that:
* `*something*` will correctly become **something**
* but `*something with multiple words*` becomes \*something with multiple words\*

app/models/hyrax/work.rb
```diff
@@ -7,38 +7,61 @@ module Hyrax
...
-  # Work and Work (TBD)
+  # Work and Work
+  #
+  # * <b>Defined:</b> The relationship is defined in the parent work's `:member_ids` attribute.
+  # * <b>Tested:</b> The relationship is tested in shared spec 'a Hyrax::Work' by testing
+  #   `it_behaves_like 'has_members'`.  Shared specs are defined in /lib/hyrax/specs/shared_specs/hydra_works.rb.
+  # * <b>Work to child Work:</b> (0..m)  A work can have many child works.
+  #   * Add a child work to a work using:
+  #       Hyrax::Transactions::Container['work_resource.add_to_parent']
+  #         .call(child_work, parent_id: parent_work.id, user: current_user)
```
elrayle, line 44: I'm not sure about recommending calling a transaction step outside the context of the full transaction.

### Potential Refactor Considerations

On the plus side, use of a transaction step, even outside of a full transaction, allows for the code to define what it means to add a child work to a work to be in only one place in the code.

Alternatives:  Collections have a service that adds/removes members of the collection.  Something similar could be added for works members.

Right now, the transaction to add a collection does not call the collection service.  And the collection service does not call the transaction.  It would be better if only one of those did the work and the other called to it.  I would tend to lean toward having the transaction call the service, but this is not a strong preference.

Interested in what others think.

If it's not a straight forward decision, I can break this out into an issue for further discussion.

cjcolvar: I was thinking the same and thought that this is what works now, but can be updated once there is a better pattern.
+1 to a separate issue for figuring out a better pattern.

elrayle:
### Switched to github style markdown.

I looked at existing yard doc in other classes and modules (e.g. lib/hyrax/publish.rb) and the ones with significant yard doc seem to already be using github style markdown.

To declutter for reading in code, I removed the bold text.  I also changed code blocks to use @example.

To run locally with github markdown:

```
gem install yard
yard doc -m markdown
```

If it complains about not understanding `-m markdown`, you may need to install it.

```
gem install redcarpet
```

The resulting generated doc is a little different.  Primarily, it moves examples under all the descriptive test.  Since that seems to be what would be expected for yard doc, that seems ok.

image

cjcolvar (approved):
This looks great!  Switching to github markdown to be consistent is a good catch.

PR #5444, opened 2022-02-14
add documentation for the Work and File Set relationship

elrayle:
Adds documentation for the Work-to-FileSet relationship.

Also performs a refactor of the find child file sets navigator.

### Description of Refactor

`fileset` and `file_set` are used interchangably in the code.  It is better to be consistent, so as code is touched, references are being refactored to be `file_set` whenever possible.  This commit refactors the child file set navigator methods to use `file_set` in method names.

### Rationale

The model name is File Set.  Classes and methods that act on file sets should use camel case and snake case, respectively,  of File Set (i.e. FileSet for classes and modules, and file_set for methods).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -432,7 +432,7 @@ module Hyrax
       when ActiveFedora::Common
         curation_concern.file_sets.present?
       else
-        Hyrax.custom_queries.find_child_fileset_ids(resource: curation_concern).any?
+        Hyrax.custom_queries.find_child_file_set_ids(resource: curation_concern).any?
       end
     end
 
--- a/app/models/hyrax/work.rb
+++ b/app/models/hyrax/work.rb
@@ -67,7 +67,7 @@ module Hyrax
   #
   # @see Valkyrie query adapter's #find_by
   # @see Hyrax::CustomQueries::Navigators::CollectionMembers#find_collections_for
-  # @see Hyrax::CustomQueries::Navigators::CollectionMembers#find_parent_work
+  # @see Hyrax::CustomQueries::Navigators::ParentWorkNavigator#find_parent_work
   #
   # @see /lib/hyrax/specs/shared_specs/hydra_works.rb
   #
--- a/app/models/hyrax/work.rb
+++ b/app/models/hyrax/work.rb
@@ -56,7 +56,24 @@ module Hyrax
   #
   # @note `:member_ids` holds ids of child works and file sets.
   #
-  # ### Work and File Set (TBD)
+  # ### Work and File Set
+  #
+  # * Defined: The relationship is defined in the parent work's `:member_ids` attribute.
+  # * Tested: The relationship is tested in shared spec `'a Hyrax::Work'` by testing
+  #   `it_behaves_like 'has_members'`.  Shared specs are defined in /lib/hyrax/specs/shared_specs/hydra_works.rb.
+  # * Work to File Set: (0..m)  A work can have many file sets.
+  #   * Add a file set to a work -- This is a multi-step process.  A good place to
+  #     start exploring the code is the calling of the transaction that creates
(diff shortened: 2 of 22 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -432,7 +432,7 @@ def concern_has_file_sets?
       when ActiveFedora::Common
         curation_concern.file_sets.present?
       else
-        Hyrax.custom_queries.find_child_fileset_ids(resource: curation_concern).any?
+        Hyrax.custom_queries.find_child_file_set_ids(resource: curation_concern).any?
```
elrayle, line 435: See PR description for rationale related to renaming the `find_child_fileset*` methods to `find_child_file_set*`.

app/services/hyrax/custom_queries/navigators/child_filesets_navigator.rb
```diff
@@ -11,7 +11,8 @@ module Navigators
       class ChildFilesetsNavigator
```
elrayle, line 11: I did not refactor the class name, but am open to doing that.

spec/services/hyrax/custom_queries/navigators/child_filesets_navigator_spec.rb
```diff
@@ -10,7 +10,7 @@
   let(:fileset1) { build(:file_set, id: 'fs1', title: ['Child File Set 1']) }
```
elrayle, line 10: I did not refactor variable names in the test, but am open to doing that.

app/services/hyrax/custom_queries/navigators/child_filesets_navigator.rb
cjcolvar: It looks like this navigator is the only class named that way.  I think it would be good to cleanup.
How hard would it be to deprecate it and change it?

app/models/hyrax/work.rb
```diff
@@ -56,18 +56,38 @@ module Hyrax
...
   #
-  # ### Work and File Set (TBD)
+  # ### Work and File Set
+  #
+  # * Defined: The relationship is defined in the parent work's `:member_ids` attribute.
+  # * Tested: The relationship is tested in shared spec `'a Hyrax::Work'` by testing
+  #   `it_behaves_like 'has_members'`.  Shared specs are defined in /lib/hyrax/specs/shared_specs/hydra_works.rb.
+  # * Work to File Set: (0..m)  A work can have many file sets.
+  #   * Add a file set to a work -- This is a multi-step process.  A good place to
+  #     start exploring the code is the calling of the transaction that creates
```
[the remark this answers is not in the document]
cjcolvar: I think the attaching of a file set to a work is the only part of the process that is relevant to this documentation.  Is there a single line that encapsulates that action?

app/services/hyrax/custom_queries/navigators/child_file_sets_navigator.rb
```diff
@@ -0,0 +1,49 @@
...
+        # Find child filesets of a given resource, and map to Valkyrie Resources
+        #
+        # @param [Valkyrie::Resource] resource
+        #
+        # @return [Array<Valkyrie::Resource>]
+        def find_child_file_sets(resource:)
+          query_service.find_members(resource: resource).select(&:file_set?)
+        end
+        alias find_child_filesets find_child_file_sets
+        deprecation_deprecate find_child_filesets: "use find_child_file_sets instead"
```
cjcolvar, line 33: Do you think the alias and its deprecation could be dropped since this is a new class?  

I'm thinking if someone has been calling the old class and deprecated method then they would fix the method call first since it has a deprecation warning but they would have to deliberately change the class by seeing it in the code and would see the old method was gone.  Also it feels like a smell to add a new class with deprecations in it.

cjcolvar (approved):
This is a great cleanup!  Thanks for doing this!

PR #5451, opened 2022-02-22
bump to ruby 2.7.5 and alpine 3.15 for base image

orangewolf:
Update Ruby and Alpine version for the Docker images

There is a change in Ruby 2.7.5 that makes the very obsolete `require 'enumerator'` call fail. This require hasn't been necessary since ruby 1.8.7 was released, but was common practice for some time. Ruby gems like `oai` and `edtf` have not been updated to support Ruby 2.7.5 yet, so that may be a factor in merging this code.  We use the following patch to make those libraries work:

```
sed -i '/require .enumerator./d' /usr/local/bundle/gems/oai-1.1.0/lib/oai/provider/resumption_token.rb && \
sed -i '/require .enumerator./d' /usr/local/bundle/gems/edtf-3.0.5/lib/edtf.rb
```

Advantages to this change are current versions of Alpine and Ruby that are fully patched.

* Build the dassie app and see that Hyrax still functions properly

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/Gemfile
+++ b/.dassie/Gemfile
@@ -8,7 +8,7 @@ else
 end
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.7.4'
+ruby '2.7.5'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.2.4', '>= 5.2.4.4'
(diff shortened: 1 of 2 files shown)


dunn (approved):

orangewolf:
closing this in because of #5629 which is the same code on top of main

PR #5454, opened 2022-02-22
add adminset title to work solrdoc

### LINKED ISSUE #5450: Cannot change admin set when using AdministrativeSets
### Descriptive summary

Setting the admin set during create or editing a work and changing the admin set, does not move the work from the Default Admin Set to the newly selected admin set.

### Rationale

Changing admin sets is a core behavior and should work for ActiveFedora AdminSets and Valkyrie AdministrativeSets.

### Expected behavior

Edit work and change the Administrative Set for the work should update which admin set the work is in.

### Actual behavior

Edit work and change the Administrative Set for the work has no impact on the admin set.

Configure Hyrax:
```
```
2. restart rails app
3. Make sure regular `hyrax-development` and `hyrax-valkyrie-dev` solr cores are running .  _Your core names may vary depending on how you start them and how they are configured in the app._
4. Make sure you have access to the datastore (i.e. postgres)

NOTE: For testing of nurax-pg, the solr core is typically named nurax-pg.

Create an admin set
1. navigate to: Dashboard -> Collections
2. select type: Admin Set
4. set Title: My Admin Set (or other meaningful name)
5. click Save

#### To reproduce:

1. navigate to: Dashboard -> Works
2. click button: Add new work
3. select type: Monograph (needs to be a Valkyrie resource)
4. fill in required metadata and upload file if needed
5. click tab: Relationships
6. under Administrative Set, select the admin set created in the prerequisites
7. click button: Save

This forwards to the new work's show page.  The admin set relationship does not show in the Relationships section.

8. click button: Edit
9. click tab: Relationships

The admin set is set to the Default Admin Set instead of the one you selected.

### Related work

elrayle:
Fixes #5450 

Two changes:
* adding adminset title to work’s solrdoc is required for the relationship to appear on the work’s show page
* resource form’s prepopulator for `admin_set_id` was switched to use the `AdminSetCreateService` which is the preferred way to get the default admin set and is required for non-Wings implementations which cannot use `AdminSet::DEFAULT_ID` as the default admin set id.  This is required for the default admin set to be the default for new works.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -176,13 +176,7 @@ module Hyrax
     end
 
     def admin_set_id_for_new
-      # admin_set_id is required on the client, otherwise simple_form renders a blank option.
-      # however it isn't a required field for someone to submit via json.
-      # Set the default admin set if it exists; otherwise, set to first admin_set they have access to.
-      admin_sets = Hyrax::AdminSetService.new(self).search_results(:deposit)
-      return nil if admin_sets.blank? # shouldn't happen
-      return AdminSet::DEFAULT_ID if admin_sets.map(&:id).include?(AdminSet::DEFAULT_ID)
-      admin_sets.first.id
+      Hyrax::AdminSetCreateService.find_or_create_default_admin_set.id.to_s
     end
 
     def build_form
--- a/app/indexers/hyrax/valkyrie_work_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_work_indexer.rb
@@ -14,6 +14,9 @@ module Hyrax
         solr_doc['generic_type_si'] = 'Work'
         solr_doc['suppressed_bsi'] = suppressed?(resource)
         solr_doc['admin_set_id_ssim'] = [resource.admin_set_id.to_s]
+        admin_set_label = admin_set_label(resource)
+        solr_doc['admin_set_sim']   = admin_set_label
+        solr_doc['admin_set_tesim'] = admin_set_label
         solr_doc['member_of_collection_ids_ssim'] = resource.member_of_collection_ids.map(&:to_s)
         solr_doc['member_ids_ssim'] = resource.member_ids.map(&:to_s)
         solr_doc['depositor_ssim'] = [resource.depositor]
@@ -25,6 +28,12 @@ module Hyrax
 
     def suppressed?(resource)
       Hyrax::ResourceStatus.new(resource: resource).inactive?
+    end
+
+    def admin_set_label(resource)
+      return "" if resource.admin_set_id.blank?
+      admin_set = Hyrax.query_service.find_by(id: resource.admin_set_id)
+      admin_set.title
     end
   end
 end
--- a/app/indexers/hyrax/valkyrie_work_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_work_indexer.rb
@@ -31,7 +31,7 @@ module Hyrax
     end
 
     def admin_set_label(resource)
-      return "" if resource.admin_set_id.blank?
+      return if resource.admin_set_id.blank?
       admin_set = Hyrax.query_service.find_by(id: resource.admin_set_id)
       admin_set.title
     end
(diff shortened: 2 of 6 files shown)


app/indexers/hyrax/valkyrie_work_indexer.rb
```diff
@@ -26,5 +29,11 @@ def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Met
     def suppressed?(resource)
       Hyrax::ResourceStatus.new(resource: resource).inactive?
     end
+
+    def admin_set_label(resource)
+      return "" if resource.admin_set_id.blank?
```
cjcolvar, line 34: If it is blank should this return nil instead?  If it is nil then the field won't get stored in solr, but I think it will if it is empty string.

app/models/concerns/hyrax/solr_document/metadata.rb
```diff
@@ -52,6 +52,7 @@ def self.coerce(input)
         attribute :read_groups, Solr::Array, ::Ability.read_group_field
         attribute :collection_ids, Solr::Array, 'collection_ids_tesim'
         attribute :admin_set, Solr::Array, "admin_set_tesim"
```
cjcolvar, line 54: Should there be an attribute added for `admin_set_sim`?

app/indexers/hyrax/valkyrie_work_indexer.rb
```diff
@@ -14,6 +14,9 @@ def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Met
         solr_doc['generic_type_si'] = 'Work'
         solr_doc['suppressed_bsi'] = suppressed?(resource)
         solr_doc['admin_set_id_ssim'] = [resource.admin_set_id.to_s]
+        admin_set_label = admin_set_label(resource)
+        solr_doc['admin_set_sim']   = admin_set_label
```
cjcolvar, line 18: What is `admin_set_sim` used for?

app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -176,13 +176,7 @@ def collections_service
...
 
     def admin_set_id_for_new
-      # admin_set_id is required on the client, otherwise simple_form renders a blank option.
-      # however it isn't a required field for someone to submit via json.
-      # Set the default admin set if it exists; otherwise, set to first admin_set they have access to.
-      admin_sets = Hyrax::AdminSetService.new(self).search_results(:deposit)
-      return nil if admin_sets.blank? # shouldn't happen
-      return AdminSet::DEFAULT_ID if admin_sets.map(&:id).include?(AdminSet::DEFAULT_ID)
-      admin_sets.first.id
+      Hyrax::AdminSetCreateService.find_or_create_default_admin_set.id.to_s
```
cjcolvar, line 179: This is big improvement!

app/indexers/hyrax/valkyrie_work_indexer.rb
elrayle: I do not know if `admin_set_sim` is used.  It was [indexed for ActiveFedora works](https://github.com/samvera/hyrax/blob/main/app/indexers/hyrax/work_indexer.rb#L22-L23), so I added it here too.

app/models/concerns/hyrax/solr_document/metadata.rb
elrayle: Not completely sure.  Title has `_tesim` and `_sim`.  It only has [`title_tesim` in this file](https://github.com/samvera/hyrax/blob/main/app/models/concerns/hyrax/solr_document/metadata.rb#L60).  There are other `_sim` properties indexed, but none of them appear in this file.

app/indexers/hyrax/valkyrie_work_indexer.rb
elrayle: Here is how I believe this is used...
* _tesim - te-tokenized text; s-stored; i-indexed; m-multi-value
* _sim - s-string; i-indexed; m-multi-value

By being indexed, they are both searchable.  By having tokenized and string, they can be searched by token or by the entire string.  Stored is indicating the field is returned as part of the solr document in search results.  It probably doesn't matter which you return, but you only want to return one of them.

app/indexers/hyrax/valkyrie_work_indexer.rb
elrayle: DONE

app/models/concerns/hyrax/solr_document/metadata.rb
elrayle: I'm assuming that the set of attributes are those that can be in the solrdoc for search results.  Since only _tesim is returned in the solrdoc, the _sim version does not need to be an attribute.

cjcolvar (approved):
Looks good!

Issue #5351, opened 2022-01-24
Seed app with data


adamjarling:
### Descriptive summary

Does an existing script exist to seed data in a dev environment?

### Rationale

Need to see real data in the app to identify all UI issues.

Maybe we can reach out to community Slack channels for help / advice.

apertome:
@adamjarling There is a draft PR relating to this here, loose ends from last dev congres -- I hope to wrap this up this dev congress, if I can get the API pieces working correctly. 

At least, I think that is what you're asking for (we call it "data seeds" in that case). Definitely feel free to provide feedback/suggestions either there. Looking at the label ... is this really related to Bootstrap 4 though? I am not sure I see the connection (though admittedly I have not been following the Bootstrap work).

adamjarling:
Hi @apertome  great news about the work being in progress.  I think you're right, it doesn't really have to do with the upgrade if it's something which will be a new feature add to `main` as well.  I'll remove the label.  Thanks.

PR #5472, opened 2022-02-28
add infrastructure for clearing repository and running seeds

elrayle:
This puts in place a series of classes that support generation of seed data.

### Pattern

The pattern handles 3 scenarios.  _NOTE: all paths to files start from `app/utils`._
* clear existing repository metadata and data that is tightly coupled with the repository model (see `data_maintenance.rb` and files under `data_destroyers`)
* generate required repository objects (e.g. default collection types, default admin set) (see `required_data_seeder.rb` and files under `required_data_seeders`)
* generate repository metadata for release and development testing  (see `test_data_seeder.rb` and files under `test_data_seeders`)

### Benefits

Benefits of this approach:
* each scenario can be run together or separately
* allows applications to easily override the  details of the seeding process using normal customization approaches
* establishes a pattern that applications can follow to create their own seed data 
* provides an easy means of adding additional scenarios (e.g. current scenarios are for testing; a site can also have seeds that drive static production data, like collection types and collections)

### To run:

_NOTE: These examples show setting environment variables from the command line.  They can be set in the regular way that an app sets environment variables.  But since the running of seeds is generally not something you want to run all the time, it seems more like that setting the environment variables from the command line is a more common approach for seeds._

To list options without running anything...
```
$ bundle exec rails db:seed
```

To wipe data and generate required data...
```
$ bundle exec rails db:seed WIPE_DATA=true
####################################################################################

WARNING: You are about to clear all repository metadata from the datastore and solr.
Are you sure? [YES|n]
YES
```

To generate release data...
```
$ bundle exec rails db:seed SEED_RELEASE_TESTING=true
```

To run both...
```
$ bundle exec rails db:seed WIPE_AND_SEED_RELEASE_TESTING=true
```

### Related Work

Issue #5351
PR #5056

@samvera/hyrax-code-reviewers


dlpierce:
I like the approach and that it allows seed specs to be written.

Could the combined option be replaced with running `bundle exec rails db:seed WIPE_DATA=true SEED_RELEASE_TESTING=true`?

dlpierce (approved):

Issue #5436, opened 2022-02-07
New work under review not showing up at "Review Submissions" on dashboard


MPLSFedResearchTZ:
### Descriptive summary

In Hyrax 3.3.0, as a manager of a workflow, after a work is created by a user and is under review, I expect to see it in two sections on dashboard: 1) In the list of the "Managed Works" in "Works" section; 2) In the list of the "Under Review" in "Review Submissions" section. 

### Expected behavior

The manager can see the newly created work under review at the section of "Managed Works":

![image](https://user-images.githubusercontent.com/93392603/152865081-9efa652f-6294-4f30-8b56-c79e56a0de45.png)

It should be in the list of the section of "Review Submissions".

### Actual behavior

But the newly created work is missing in the list of "Under Review" in "Review Submissions" section:
![image](https://user-images.githubusercontent.com/93392603/152868523-514b8146-1a8a-4dc0-9015-15780e11e3d0.png)

### Steps to reproduce the behavior

1. Log in as an admin user,  select an admin set and then attach to it a submission-review workflow, e.g., the "mediated deposit workflow".  After this, assign a non-admin user as the manager of the admin set and define the user/user roles who will use this admin set with the submission workflow to submit works.
2. Log in as a non-admin user that has the authority to submit work to the submission-review workflow, create a new work.
3. Log in as the non-admin user who was assigned as the manager to the submission-review workflow, check if the newly created work is in the list of "Managed Works" at dashboard.
4. Then go to the list of the "Under Review" under dashboard section of "Review Submissions", and check if work is also there.

### Related work

[Link to related tickets or prior related work here.](https://github.com/samvera/hyrax/issues/3171)

fritzfreiheit:
I could not reproduce this on my local host (built from scratch). But I did find an inverted version of this problem where the works in an admin set with single-step mediated deposit were showing up in the "Under Review" section of "Review Submissions", but not in the "Managed Works" tab under Dashboard Works page for non-admin users with managing role for that admin set. I've open a ticket to address this alternate issue: #5471

rjkati:
I can replicate this issue in my local build of Hyrax 4.0.0.beta1

rjkati:
This is working now on nurax-dev (Hyrax 4.0.0.beta2)

PR #5473, opened 2022-02-28
Hyrax-5471 - Works in mediated workflow not listed as managed works f…

### LINKED ISSUE #5471: Works in mediated workflow not listed as managed works for user with managing role
### Descriptive summary

Works in mediated workflow not listed as managed works on on a user's Dashboard Works page for users with managing role in an admin set with one-step mediated deposit workflow. These works are listed in the Review Submissions page. This occurs in HEAD.

### Rationale

This is an expected behavior.

### Expected behavior

Users with managing role for an admin set with a one-step mediated workflow should see a list of managed works on their Managed Works tab on their dashboard Managed Works page.

### Actual behavior

The Dashboard Managed Works tab does not display for users with a managing role in an admin set with one-step mediated deposit that have one or more member works.

1. Ensure that an admin set with a one-step mediated workflow exists.
2. As a normal user, create a work and add the admin set with a one-step mediated workflow identified or created in step 1.
3. Grant a non-admin user managing status to the admin set with the one-step mediated workflow used in step 2.
4. Go to the Dashboard, then Works page of a non-admin user with managing status from step 3.
5. The Managed User tab should be visible. Go to that tab.
6. The work created in step 2 should be listed in this tab. (It should also be listed in the Review Submissions tab.)

### Related work

New work under review not showing up at "Review Submissions" on dashboard #5436

fritzfreiheit:
Works in mediated workflow not listed as managed works for user with managing role

Fixes #5471

This issue stemmed from investigation into issue: #5436 

Add code to search builders to filter for admin sets with user's managing role.

Update [app/search_builders/hyrax/dashboard/managed_search_filters.rb] to add a method that performs the following steps:
* Find the managing sippity role
* convert the current user into a sippity agent
* from the agent's list of workflow responsibilities, find all responsibilities that are managing
* with this list of workflow responsibilities, find the corresponding admin set ids through the corresponding permission templates
* add a list of search filters that include "isPartOf" admin set ids identified in prior step

* (1) Ensure that an admin set with a one-step mediated workflow exists.
* (2) As a normal user, create a work and add the admin set with a one-step mediated workflow identified or created in step 1.
* (3) Grant a non-admin user managing status to the admin set with the one-step mediated workflow used in step 2.
* (4) Go to the Dashboard, then Works page of a non-admin user with managing status from step 3.
* (5) The Managed User tab should be visible. Go to that tab.
* (6) The work created in step 2 should be listed in this tab. (It should also be listed in the Review Submissions tab.)

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/dashboard/managed_search_filters.rb
+++ b/app/search_builders/hyrax/dashboard/managed_search_filters.rb
@@ -25,6 +25,12 @@ module Hyrax
         search_terms
       end
 
+      # Look for a user's managing role and add filters for all admin sets that have permission
+      # templates that include managing roles.
+      #
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/PerceivedComplexity
+      # rubocop:disable Metrics/CyclomaticComplexity
       def add_managing_role_search_filter(ability:, search_terms: [])
         search_terms ||= []
         # Look for managing role assignement
@@ -51,6 +57,9 @@ module Hyrax
         end
         search_terms
       end
+      # rubocop:enable Metrics/MethodLength
+      # rubocop:enable Metrics/PerceivedComplexity
+      # rubocop:enable Metrics/CyclomaticComplexity
     end
   end
 end
--- a/app/search_builders/hyrax/dashboard/managed_search_filters.rb
+++ b/app/search_builders/hyrax/dashboard/managed_search_filters.rb
@@ -16,22 +16,22 @@ module Hyrax
         return search_terms if groups.empty?
         permission_types.each do |type|
           field = solr_field_for(type, 'group')
-          user_groups = type == 'read' ? groups - [::Ability.public_group_name,
-                                                   ::Ability.registered_group_name] : groups
+          delete_groups = [::Ability.public_group_name, ::Ability.registered_group_name]
+          user_groups = type == 'read' ? groups - delete_groups : groups
           next if user_groups.empty?
           # parens required to properly OR the clauses together:
           search_terms << "({!terms f=#{field}}#{user_groups.join(',')})"
         end
-        return search_terms
+        search_terms
       end
 
-      def add_managing_role_search_filter( ability:, search_terms: [] )
+      def add_managing_role_search_filter(ability:, search_terms: [])
         search_terms ||= []
         # Look for managing role assignement
         managing_role = Sipity::Role.find_by(name: Hyrax::RoleRegistry::MANAGING)
-        return search_terms unless managing_role.present?
+        return search_terms if managing_role.blank?
         agent = ability.current_user.to_sipity_agent
-        return search_terms unless agent.workflow_responsibilities.present?
+        return search_terms if agent.workflow_responsibilities.blank?
         managing_workflow_roles = []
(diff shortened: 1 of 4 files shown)


app/search_builders/hyrax/dashboard/managed_search_filters.rb
```diff
@@ -11,15 +11,55 @@ def discovery_permissions
...
       end
+
+      # Look for a user's managing role and add filters for all admin sets that have permission
+      # templates that include managing roles.
+      #
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/PerceivedComplexity
+      # rubocop:disable Metrics/CyclomaticComplexity
+      def add_managing_role_search_filter(ability:, search_terms: [])
+        search_terms ||= []
```
dlpierce, line 35: Is setting `search_terms` again here needed? I think `search_terms` has a default value of [] from the arguments list a line above.

spec/search_builders/hyrax/dashboard/collections_search_builder_spec.rb
```diff
@@ -55,6 +55,11 @@
 
     subject { builder.gated_discovery_filters }
 
+    before do
```
dlpierce, line 58: What problem is this addition addressing? This spec passes locally for me without manually saving the users.

app/search_builders/hyrax/dashboard/managed_search_filters.rb
fritzfreiheit: I was covering the situation that nil was deliberately passed in as the value of :search_term. Does the default value block that possibility? I haven't actually tried that, but I'll test it locally.

spec/search_builders/hyrax/dashboard/collections_search_builder_spec.rb
fritzfreiheit: I was getting the following error at the .to_sippity_agent line: Failure/Error: create_sipity_agent! ActiveRecord::RecordNotSaved: You cannot call create unless the parent is saved
I can take those saves out and retry it locally.

app/search_builders/hyrax/dashboard/managed_search_filters.rb
fritzfreiheit: So ruby passes nil when a named parameter with a non-nil default is called:
`[1] pry(main)> def test_named_parm( parm: [] )
  puts parm.class.name
[1] pry(main)> def test_named_parm( parm: [] )
  puts parm.class.name
end
=> :test_named_parm
[2] pry(main)> test_named_parm
Array
=> nil
[3] pry(main)> test_named_parm( parm: nil )
NilClass
=> nil
`

dlpierce (approved):

jlhardes:
This seems to be better for non-admin users with managing roles on mediated workflow admin sets. The admin user I can see on nurax-dev is still not seeing the full list of works in the Review Submissions section in the Under Review tab but the All Works tab shows a long list of works under review so this overall does seem to be an improvement. I am calling this a win and we'll see if there are other issues to figure out around the admin user view of mediated works.

PR #5476, opened 2022-03-01
force collection for sharing to load before running the feature test

elrayle:
Addresses 2 flaky feature tests for dashboard/collections.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/dashboard/collection_spec.rb
+++ b/spec/features/dashboard/collection_spec.rb
@@ -762,10 +762,6 @@ RSpec.describe 'collection', type: :feature, clean_repo: true do
   end
 
   describe 'edit collection' do
-    let(:collection) { build(:named_collection_lw, user: user, with_permission_template: true) }
-    let!(:work1) { create(:work, title: ["King Louie"], member_of_collections: [collection], user: user) }
-    let!(:work2) { create(:work, title: ["King Kong"], member_of_collections: [collection], user: user) }
-
     context 'from dashboard -> collections action menu' do
       before do
         create(:permission_template_access,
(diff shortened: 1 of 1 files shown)


spec/features/dashboard/collection_spec.rb
```diff
@@ -936,15 +939,16 @@ def get_url_fragment(type)
...
-            click_link('Sharing')
+            find_link('Sharing').click
             expect(page).to have_selector(".form-inline.add-users .select2-container")
             select_user(user2, 'Depositor')
-            click_button('Save')
-            click_link('Sharing')
-            expect(page).to have_selector('td', text: user2.user_key)
+            expect(page).to have_content "The collection's sharing options have been updated."
+            within('section.section-collection-sharing') do
+              expect(page).to have_selector('td', text: user2.user_key)
```
elrayle, line 950: Adding a user to a role saves immediately, so there is no need to Save and reclick Sharing tab.  This does check for the appropriate flash message to confirm the save happened and then confirms that the user was assigned the role.

spec/features/dashboard/collection_spec.rb
```diff
@@ -792,7 +788,14 @@ def get_url_fragment(type)
 
     context 'from dashboard -> collections action menu' do
       context 'for a collection' do
+        let(:collection) { build(:named_collection_lw, user: user, with_permission_template: true) }
+        let(:work1) { create(:work, title: ["King Louie"], member_of_collections: [collection], user: user) }
+        let(:work2) { create(:work, title: ["King Kong"], member_of_collections: [collection], user: user) }
```
elrayle, line 793: Moving these into the one test that uses them and removing the force load with let!, preferring a load in the before block, fixes LDP::Gone failures in tests that don't use those object. This is far less common than the other flaky test fixed by this PR.

conorom (approved):

PR #5478, opened 2022-03-01
Upgrade to Bixby 4.0.0

cjcolvar:
~This is a work in progress waiting on https://github.com/samvera-labs/bixby/pull/65 and a release of bixby.~

This PR upgrades hyrax to the 4.0.0 release of bixby which uses rubocop 1.x.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ inherit_gem:
   bixby: bixby_default.yml
 
 AllCops:
-  TargetRubyVersion: 2.4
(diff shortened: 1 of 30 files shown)


cjcolvar:
@elrayle I'm not sure how to do that since rubocop would fail without both pieces present.  I could make an upgrade PR which adds exceptions for all of the auto-correctable violations then I could fix them in a separate PR.  Do you think that makes sense to do?

no-reply (approved):

PR #5479, opened 2022-03-01
Nginx for Assets, Cantaloup proxy and Bot Blocking

orangewolf:
Right now Puma processes service the assets directly, which is not ideal from a performance stand point. We've also had significant issues with bots and other known bad actors racking our servers. Lastly, we'd like to be able to safely put Cantaloup in place behind the application level auth, but without incurring the overhead of proxying through the Rails stack. We'll do the Cantaloup as a separate PR. This PR adds bitnami/nginx to the stack and leaves an example of possible values to accomplish all of these tasks.

- [x] set up nginx
- [x] modify ingress if nginx is enabled
- [x] example of serving assets through nginx
- [x] bot blocking example

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/templates/_helpers.tpl
+++ b/chart/hyrax/templates/_helpers.tpl
@@ -179,6 +179,14 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- printf "redis://:%s@%s:%s" .Values.redis.password (include "hyrax.redis.host" .) "6379/0" -}}
 {{- end -}}
 
+{{- define "hyrax.hyrax.host" -}}
+  {{- include "hyrax.fullname" . }}
+{{- end -}}
+
+{{- define "hyrax.nginx.host" -}}
+{{- printf "%s-%s" .Release.Name "nginx" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
 {{- define "hyrax.sharedPvcAccessModes" -}}
 {{- if .Values.worker.enabled }}
 accessModes:
--- a/chart/hyrax/templates/_helpers.tpl
+++ b/chart/hyrax/templates/_helpers.tpl
@@ -179,10 +179,6 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- printf "redis://:%s@%s:%s" .Values.redis.password (include "hyrax.redis.host" .) "6379/0" -}}
 {{- end -}}
 
-{{- define "hyrax.hyrax.host" -}}
-  {{- include "hyrax.fullname" . }}
-{{- end -}}
-
 {{- define "hyrax.nginx.host" -}}
 {{- printf "%s-%s" .Release.Name "nginx" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
--- a/chart/hyrax/values.yaml
+++ b/chart/hyrax/values.yaml
@@ -268,7 +268,7 @@ nginx:
   #     subPath: public-assets
   # serverBlock: |-
   #   upstream rails_app {
-  #     server {{ .Values.global.hyraxName }};
+  #     server {{ .Values.global.hyraxHostName }};
   #   }
 
   #   map $status $loggable {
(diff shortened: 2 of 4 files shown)


chart/hyrax/values.yaml
```diff
@@ -273,6 +411,7 @@ global:
   postgresql:
     postgresqlUsername: hyrax
     postgresqlPassword: hyrax_pass
+  hyraxName: hyrax
```
mcritchlow, line 414: Would it be worth giving this a name that more directly conveys it being used for `nginx` configuration? Or, might it be used elsewhere as well? On initial read I didn't understand why this new value was here, but then found it in the `nginx.serverBlock` setup.

mcritchlow:
This seems like a great idea. I'll try and pull this down locally soon and kick the tires with it.

chart/hyrax/values.yaml
orangewolf: I went back and forth on it. on the one hand we only use it once right now. on the other hand, its global and potentially useful in other places. Would it feel better as `.Values.global.hyraxHostName` or `hyraxHostUrl`?

orangewolf: @mcritchlow have you had a chance to think more about this? I've been using this code in prod for a week now and it is working well.  We've had this bot blocker set up on another Rails app for over a year and I highly recommend it.

orangewolf: no problem on the delay - I 100% understand. I already updated the description to say Cantaloup comes later. I agree with you on the name, I'll push an update.

chart/hyrax/templates/_helpers.tpl
```diff
@@ -179,6 +179,14 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- printf "redis://:%s@%s:%s" .Values.redis.password (include "hyrax.redis.host" .) "6379/0" -}}
 {{- end -}}
 
+{{- define "hyrax.hyrax.host" -}}
```
mcritchlow, line 182: This template doesn't seem to be used/needed?

mcritchlow:
This seems great to me. I have a general question about the uploads PVC for nginx as it relates to the hyrax chart uploads pvc, and I _think_ that the `hyrax.hyrax.host` isn't needed.

Otherwise, I think we should bump the version in `Chart.yaml` and I'll approve and merge this. This is an awesome contribution :clap: 

Once we have this in I'm happy to push a new chart release as well.

chart/hyrax/values.yaml
```diff
@@ -238,6 +238,144 @@ postgresql:
...
+  # The set up below does malicious bot / ip blocking and mounts
+  # vaolumes to allow nginx to server assets and other public directory items
+  # image:
+  #   registry: registry.gitlab.com
+  #   repository: notch8/scripts/bitnami-nginx
+  #   tag: 1.21.5-debian-10-r4
+  # extraVolumes:
+  #   - name: "uploads"
+  #     persistentVolumeClaim:
+  #       claimName: dev-nnp-uploads
```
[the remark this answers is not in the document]
orangewolf: good point. it needs to match the uploadsVolume name. I wonder if for example purposes the sub out of `{{ .Values.global.hyraxHostName }}-uploads` is a reasonable thing to use - you have to have something from global that's specific otherwise.

chart/hyrax/templates/_helpers.tpl
orangewolf: you are correct - I really tied my self in knots a bit trying to keep from having to have a global. I'll remove this.

mcritchlow:
@orangewolf - sorry i should have caught this yesterday.. can you also update the `Chart.lock` file. I think that's it and this is good to go from my end

no-reply:
i had occasion to catch up to this for the first time this afternoon, and it looks fantastic.

i'm going to try a deploy on my side in the AM and if there are still Chart.lock changes or whatever else to do at that time, i'll try to close the loop so we can get this merged.

this is a great contribution!

orangewolf:
@mcritchlow finally bumped the chart version. this should be gtg now

no-reply (approved):

PR #5491, opened 2022-03-03
check if logo file id is a string, not just public.

blancoj:
This could potentially fix issue #3475

The code presently checks for the string "public" in the id list of logo files to determine if the file was previously uploaded.  Because configurations can change, it is better to check if the id is a string.  Basically:

If the user has chosen a new logo, the file id will be an integer
If the logo was previously chosen, the file id will be a path (string)
If it is a path, update the branding record, else create a new branding record.

I do not S3 configured on my local machine, but if someone does, perhaps they can check if this fixes issue #3475

* Create a collection.
* Upload a banner and logo, and save the changes.
* Save the changes again.
* Upload a new logo, and Save the changes.
* You should be able to do this without any errors occurring.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -340,7 +340,10 @@ module Hyrax
       def process_logo_records(uploaded_file_ids)
         public_files = []
         uploaded_file_ids.each_with_index do |ufi, i|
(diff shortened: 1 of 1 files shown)


elrayle (approved):
Thanks for updating this check.  It should allow this to work with alternate data stores (e.g. S3, path outside of app's public dir).

Issue #5461, opened 2022-02-23
Editing batch of works generates "ActiveFedora::ObjectNotFoundError in Hyrax::BatchEditsController#edit" error


rjkati:
### Descriptive summary

On nurax-pg, editing a batch of works displays a "ActiveFedora::ObjectNotFoundError in Hyrax::BatchEditsController#edit" error

### Expected behavior

Selecting multiple works on the Works page and clicking "Edit Selected" should bring me to an edit page

### Actual behavior

Selecting multiple works on the Works page and clicking "Edit Selected" generates an error message

### Steps to reproduce the behavior

1. On nurax-pg, navigate to Dashboard -> Works
2. Select two or more works on the Works page
3. Click the "Edit Selected" button
4. Observe error message
![batcheditscontroller](https://user-images.githubusercontent.com/32644254/155389130-67da3d9d-4bc9-449f-8a67-13c8696d5338.PNG)

elrayle:
`Hyrax::Forms::BatchEditForm` extends
`Hyrax::Forms::WorkForm` which is a `Hyrax::Editor` form and isn't setup for Valkyrie.  Ideally, it would be converted to extend `Hyrax::Forms::ResourceForm` which is a `Valkyrie::ChangeSet`.

jlhardes:
Closing issue since different error message showing now when trying to batch edit works.

PR #5498, opened 2022-03-07
Allow batch form to load valkyrie objects

hackartisan:
I think we'll still need a form object here even with valkyrie.

refs #5461


elrayle (approved):
Glad it turned out to be a change to a couple of #find calls.

PR #5503, opened 2022-03-07
set use_valkyrie to true when wings is disabled

elrayle:
When an application is using an adapter other than wings, the app is able to disable wings using an environment variable.

```
ENV["HYRAX_SKIP_WINGS"] = true
```
This is tied to Hyrax.config.disable_wings.

There is a second environment variable, `"HYRAX_VALKYRIE"`.  It is tied to `Hyrax.config.use_valkyrie?`, which should always be true when using an adapter other than wings.

This PR updates `Hyrax.config.use_valkyrie?` to always return `true` when `Hyrax.config.disable_wings  == true`; otherwise, it returns the value based on `ENV["HYRAX_VALKRIE"]`.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Looks good to me!

Issue #5490, opened 2022-03-03
Inventory UI for consistent language


hackartisan:
### Descriptive summary

Throughout the UI there are inconsistencies in buttons, links, and action labels. For example, in the Dashboard Collections area the button says "New Collection" and in the Dashboard Works area the button says "Add new work"

Inventory the UI to find inconsistencies and bring them into alignment with one another.

elrayle:
Example:
* Dashboard -> Works uses `Add new work`
* Dashboard -> Collections uses `New Collection`

elrayle:
Listed from high to low priority.
* Limit to Dashboard -> Works and Collections (highest priority)
* Work and Collection show pages.  NOTE: Collections has an admin show page and a public show page.  Work only has one show page that serves both.  
* Could explore FileSet show page as well.

elrayle:
Inconsistent dates in Dashboard listings:
* Dashboard -> Collections - has column Last Modified
* Dashboard -> Works - has column Date Added

This may impact work on Issue #5486 which proposes sorting dashboard tables with most recently updated resources first.

elrayle:
@jlhardes and I discussed order of columns.  Seems like this order would be best:
* Work -- Title, Last Date Modified, Highlight, Visibility
* Column -- Title, Collection Type, Last Date Modified, Items, Visibility

There is a question about impact of sorting on future Blacklight/Bootstrap .

sdellis:
> Inconsistent dates in Dashboard listings:
> 
> * Dashboard -> Collections - has column Last Modified
> * Dashboard -> Works - has column Date Added
> 

@elrayle Are these inconsistent? I tested Last Modified for collections, and that is accurate. Are we sure that this should be "Last Modified" for Works?

hackartisan:
> > Inconsistent dates in Dashboard listings:
> > 
> > * Dashboard -> Collections - has column Last Modified
> > * Dashboard -> Works - has column Date Added
> 
> @elrayle Are these inconsistent? I tested Last Modified for collections, and that is accurate. Are we sure that this should be "Last Modified" for Works?

@sdellis I just edited a work i had added last week and it is definitely showing the date added, not the date modified. If we want these to be consistent we'll need to change both the label and the behavior.

sdellis:
Thanks for the helpful info, @hackartisan

sdellis:
Refs #5500 
Refs #5508 
Refs #5510

PR #5508, opened 2022-03-08
Removes icons from Works dashboard actions dropdown menu

### LINKED ISSUE #5506: Remove icons from the Works dashboard "actions" dropdown menu
### Descriptive summary

The icons in the Works dashboard "actions" dropdown menu are inconsistent with the "actions" dropdown in the Collection dashboard. Given that the icons are not enhancing the accessibility in any way, they should be removed. This is only a UI change as it does not affect backend models.

### Related work

Ref #5490

sdellis:
Fixes #5506 ; refs #5490

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/my/_work_action_menu.html.erb
+++ b/app/views/hyrax/my/_work_action_menu.html.erb
@@ -12,7 +12,8 @@
 
     <% if can? :edit, document.id %>
       <li role="menuitem" tabindex="-1">
-        <%= link_to [main_app, :edit, document] do %>
+        <%= link_to [main_app, :edit, document],
+                    id: 'action-edit-work' do %>
           <%= t("hyrax.dashboard.my.action.edit_work") %>
         <% end %>
       </li>
@@ -20,6 +21,7 @@
       <li role="menuitem" tabindex="-1">
         <%= link_to [main_app, document],
                     method: :delete,
+                    id: 'action-delete-work',
                     data: {
                       confirm: t("hyrax.dashboard.my.action.work_confirmation", application_name: application_name) } do %>
           <%= t("hyrax.dashboard.my.action.delete_work") %>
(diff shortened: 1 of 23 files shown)


app/views/hyrax/my/_work_action_menu.html.erb
```diff
@@ -13,7 +13,6 @@
     <% if can? :edit, document.id %>
       <li role="menuitem" tabindex="-1">
         <%= link_to [main_app, :edit, document] do %>
-          <i class="glyphicon glyphicon-pencil" aria-hidden="true"></i>
           <span> <%= t("hyrax.dashboard.my.action.edit_work") %> </span>
```
hackartisan, line 16: Do we still need the spans once the icons are gone?

hackartisan (changes requested):

elrayle (dismissed):
Looks good to me. 

I noticed that Works use "Edit Work" and Collections use "Edit collection".  A follow-on PR to make those consistent would be good.

hackartisan (approved):

sdellis:
Screen Shot 2022-03-11 at 11 44 44 AM
Screen Shot 2022-03-11 at 11 41 11 AM

app/views/hyrax/my/works/_list_works.html.erb
```diff
@@ -28,7 +28,15 @@
       </div>
     </div>
   </td>
-  <td class="date text-center"><%= document.date_modified %></td>
+  <td class="date text-center">
+    <% if document.date_modified.nil? %>
```
elrayle, line 32: I think the reason something like this wasn't done before is that date_modified and date_uploaded are core fields and are set by the system, not the user.  I wonder if you can put just the change to this file in a separate PR.  Everything else looks good to go.  I'd like to think about this change.

elrayle (changes requested):

sdellis:
@elrayle I extracted the date formatting from this PR, so this is ready for final review. Will create a new PR for that one.

elrayle (approved):
Thanks for getting this in place.  These consistency updates to the UI will really improve the user experience.

PR #5511, opened 2022-03-09
make change set forms allow multiple values for title

elrayle:
Per conversation with @jlhardes, work forms should allow for multiple titles to maintain backward compatibility with metadata that already has multiple fields.  Otherwise, this presents problems for migrated data.

Prior to this PR
* work titles could have multiple values
  * this allows multi-valued titles to migrate as multi-valued
* work form only displays a single field for titles
  * this restricts new works to a single title
  * for existing works with multi-valued title, only one title would be selected for display and there isn't a way for the multiple values to be edited

With this PR
* work titles can have multiple values (no change)
* work form also allows for multiple values

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/metadata/core_metadata.yaml
+++ b/config/metadata/core_metadata.yaml
@@ -8,7 +8,7 @@ attributes:
     form:
       required: true
       primary: true
-      multiple: false
+      multiple: true
   date_modified:
     type: date_time
   date_uploaded:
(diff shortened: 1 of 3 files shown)


jlhardes:
This is setting up Valkyrized Works and Collections to allow for multiple Titles which is how metadata is structured for titles in non-Valkyrized Works and Collections. This keeps things moving forward smoothly to allow for upgrading Hyrax versions. Ideally, a migration path might be to select one of the multiple titles as THE title and the other values are moved to Alternative Title but it makes sense for this work to continue to support the capabilities that are currently available, and that includes multiple Titles.

dlpierce (approved):

Issue #5467, opened 2022-02-24
Sharing tab does not display on works


rjkati:
### Descriptive summary

on nurax-pg, the sharing tab does not display on the new work or edit works pages. The default admin set allows sharing.

### Expected behavior

The new work and edit work pages should display the sharing tab if allowed by the admin set

### Actual behavior

The new work and edit work pages do not display the sharing tab, even if allowed by the admin set

#### New works
1. On nurax-pg, navigate to Dashboard -> Works
2. Click "Add new work" and select Generic Work
3. Observe three tabs: "Descriptions", "Files" and "Relationships". Note that "Sharing" does not appear

#### Edit works
1. On nurax-pg, navigate to Dashboard -> Works
2. Click on a work for which you have edit permissions
3. Click "Edit"
4. Observe three tabs: "Descriptions", "Files" and "Relationships". Note that "Sharing" does not appear

![nosharing](https://user-images.githubusercontent.com/32644254/155599861-15261e07-df28-4257-a5dd-761e7abc9d34.PNG)

Issue #5462, opened 2022-02-23
Editing a parent work displays a "NoMethodError in Hyrax::GenericWorks#edit" error message


rjkati:
### Descriptive summary

On nurax-pg, when I attempt to edit a parent work I get a "NoMethodError in Hyrax::GenericWorks#edit" message.

### Expected behavior

Clicking "edit" on a parent work should display the work edit page

### Actual behavior

Clicking "edit" on a parent work displays a "NoMethodError in Hyrax::GenericWorks#edit" error message.

### Steps to reproduce the behavior

1. On nurax-pg, navigate to Dashboard -> Works
2. Select a work
3. Click on the "Edit" button
4. Click on the "Relationships" tab
5. Locate the "Other Works in this Work" heading
6. Click on the "Deposit new work as child of this work" button
7. Create a new Generic Work
8. Save the new Generic Work
9. Navigate to the first Work
10. Click on "Edit" 
11. Observe error message

![imagesedit](https://user-images.githubusercontent.com/32644254/155398670-907e3da4-2a9e-4822-a113-dee13dac7dda.PNG)

PR #5514, opened 2022-03-09
Find permission templates via their admin set relation

### LINKED ISSUE #5460: Embargoes and leases cannot be selected on new work form
### Descriptive summary

On nurax-pg, I cannot select "Embargo" or "Lease" from the visibility menu. The default admin set allows embargoes and leases to be selected.

### Expected behavior

Embargoes and leases should be selectable if allowed by the admin set.

### Actual behavior

Embargoes and leases cannot be selected, even if the admin set permits.

1. On nurax-pg, navigate to Dashboard -> Works
2. Click "Add new work" and select "Generic Work"
3. Click the "Relationships" tab and verify that "Default Admin Set" is selected
4. Observe that "Embargo" and "Lease" options are grayed out in the Visibility menu and cannot be selected

embargo_lease_show_page
default_admin_set_visibility

hackartisan:
Without this fix the form just always uses the default permission template behavior

closes #5460

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -456,7 +456,7 @@ module Hyrax
     def available_admin_sets
       admin_set_results = Hyrax::AdminSetService.new(self).search_results(:deposit)
       # get all the templates at once, reducing query load
-      templates = PermissionTemplate.where(id: admin_set_results.map(&:id)).to_a
+      templates = PermissionTemplate.where(source_id: admin_set_results.map(&:id)).to_a
 
       admin_sets = admin_set_results.map do |admin_set_doc|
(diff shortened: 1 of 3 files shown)


hackartisan:
I'm honestly not sure what the expectation means on the controller test I changed. But this is what I observe in a working development environment. The bug and the test values were introduced at the same time (in https://github.com/samvera/hyrax/pull/5050) so I think it's fair to assume the test was checking for the fallthrough (no permission template found) defaults rather than the default adminset defaults. Review from someone more familiar with the meaning of these permissions would be very helpful.

elrayle (changes requested):
This is a good fix.  I have one request for the works controller spec.

elrayle:
Does this fix Issue #5462 editing a parent work error OR #5460 can't select embargo on work?

elrayle (approved):

elrayle:
Also fixes #5467 Sharing tab does not display on works

PR #5515, opened 2022-03-09
Pcdm collection search results

### LINKED ISSUE #5463: Searches with collections in results are failing
### Descriptive summary

Conducting a search for a term that brings back any collection in the result set is failing with "ActionView::MissingTemplate in Catalog#index" error:
Screen Shot 2022-02-23 at 4 07 39 PM

1. Search nurax-pg for term that only appears in works (like "cat" - https://nurax-pg.curationexperts.com/catalog?utf8=%E2%9C%93&locale=en&search_field=all_fields&q=cat) - this search works correctly and brings back results
2. Search nurax-pg for term that includes collections in the results (like "test" - https://nurax-pg.curationexperts.com/catalog?utf8=%E2%9C%93&locale=en&search_field=all_fields&q=test) - this search fails with error message

cjcolvar:
Fixes #5463

Journey to getting Hyrax::PcdmCollection resources to show in search results:
- Added additional values to `Hyrax::CollectionName` to force Rails to resolve view partials to `hyrax/collections/collection`.  Without this change Rails was looking in `hyrax/hyrax/pcdm_collections/pcdm_collection` when attempting to render a SolrDocument with `has_model_ssim: Hyrax::PcdmCollection` (See https://github.com/samvera/hyrax/blob/main/lib/hyrax/active_fedora_dummy_model.rb#L53)
- Copied `catalog/_index_headers_list_collection.html.erb` to `catalog/_index_headers_list_hyrax_pcdm_collection.html.erb` so Blacklight could find it.  This seemed simpler than overriding Blacklight's helper which looks for partials to render (See https://github.com/projectblacklight/blacklight/blob/v6.24.0/app/helpers/blacklight/render_partials_helper.rb#L128-L146)
- Cloned a simple feature test that shows work and collection search results to reconfigure it to work with valkyrie resources and the valkyrie solr index.  
- To get the tests to pass I added an around hook which changes the blacklight config for the CatalogController to point to the valkyrie solr core instead of the normal `hyrax-test` core. 
- I also had to modify the `basic_metadata.yaml` to have the description field be indexed like it is in ActiveFedora work and collection models.

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/metadata/basic_metadata.yaml
+++ b/config/metadata/basic_metadata.yaml
@@ -50,6 +50,8 @@ attributes:
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "description_tesim"
   identifier:
     type: string
     multiple: true
--- a/spec/features/search_spec_valkyrie.rb
+++ b/spec/features/search_spec_valkyrie.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+RSpec.describe 'searching', index_adapter: :solr_index do
+  let(:user) { create :user }
+  let(:subject_value) { 'mustache' }
+  let!(:work) do
+    FactoryBot.valkyrie_create(:monograph, :public,
+           title: ["Toothbrush"],
+           keyword: [subject_value, 'taco'])
+  end
+
+  let!(:collection) do
+    FactoryBot.valkyrie_create(:hyrax_collection, :public, title: ['collection title abc'], description: [subject_value], members: [work])
+  end
+
+  before { allow(Hyrax.config).to receive(:collection_model).and_return('Hyrax::PcdmCollection') }
+
+  context "as a public user", :clean_repo do
+    it "using the gallery view" do
+      visit '/'
+      fill_in "search-field-header", with: "Toothbrush"
+      click_button "search-submit-header"
+      expect(page).to have_content "1 entry found"
+      within "#search-results" do
+        expect(page).to have_content "Toothbrush"
+      end
+
+      click_link "Gallery"
+      expect(page).to have_content "Filtering by: Toothbrush"
+      within "#documents" do
+        expect(page).to have_content "Toothbrush"
+      end
+    end
+
+    it "only searches all and does not display search options for dashboard files" do
+      visit '/'
+
+      # it "does not display search options for dashboard files" do
+      # This section was tested on its own, and required a full setup.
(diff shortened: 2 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -303,6 +303,14 @@ def clean_active_fedora_repository
     Valkyrie::IndexingAdapter.find(adapter_name).wipe!
   end
 
+  # Configure blacklight to use the valkyrie solr index
+  config.around(:example, index_adapter: :solr_index) do |example|
```
elrayle, line 307: Nice to have this in place.  It will help with future tests that need the valkyrie core to be primary.

spec/features/search_spec_valkyrie.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+RSpec.describe 'searching', index_adapter: :solr_index do
+  let(:user) { create :user }
```
elrayle, line 3: First fully valkyrie feature test.  💯

config/metadata/basic_metadata.yaml
```diff
@@ -50,6 +50,8 @@ attributes:
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "description_tesim"
```
elrayle, line 54:
```suggestion
      - "description_sim"
      - "description_tesim"
```

Should this have the string version as well?  It is pretty common to include both, but the string stores as the exact string typically used for exact matching.  I don't imagine that someone would be looking for an exact match for the entire description.

> String stores a word/sentence as an exact string without performing tokenization etc. Commonly useful for storing exact matches, e.g, for facetting.
> 
> Text typically performs tokenization, and secondary processing (such as lower-casing etc.). Useful for all scenarios when we want to match part of a sentence.

elrayle (approved):
This looks really good.  I just have one question.  I'm going to go ahead and  approve because I am good with your decision on the question either way, and even lean a bit toward the way it is now.

config/metadata/basic_metadata.yaml
cjcolvar: From what I can tell description was only every indexed as `_tesim`.  (https://github.com/samvera/hyrax/blob/main/app/indexers/hyrax/basic_metadata_indexer.rb#L7)

PR #5521, opened 2022-03-11
remove values that are blank from resources

elrayle:
In other work under development, collections are including basic metadata.  There was already a test similar to the new work test in this PR that checked for blank values.  That test started failing with the addition of basic metadata.  

The work test in this PR was added as a failing test for the current code, and which is now passing with the change in the way type is generated through the schema loader.  Thanks to @tpendragon for the proper code to remove blank values.

Co-authored-by: @tpendragon 

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/models/hyrax/work_spec.rb
+++ b/spec/models/hyrax/work_spec.rb
@@ -20,6 +20,14 @@ RSpec.describe Hyrax::Work do
       .to be_empty
   end
 
+  context 'when fields generated from schema' do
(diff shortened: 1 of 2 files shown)


spec/models/hyrax/work_spec.rb
```diff
@@ -20,6 +20,14 @@
       .to be_empty
   end
 
+  context 'when fields generated from schema' do
+    let(:monograph) { FactoryBot.valkyrie_create(:monograph, title: 'A Monograph', creator: [""]) }
+    it "removes blank strings from params before updating work metadata" do
```
elrayle, line 25: Test fails without the fix.

PR #5493, opened 2022-03-04
Geonames labels

dlpierce:
Fixes #1065, #3269

Retrieves, indexes and displays the full location label

Previously when indexing the Based Near/Location field on works, only the city name was available as it was all that is available in the rdf data that was retrieved. This PR add a service that retrieves a full label from another Geonames endpoint via questioning authority, caching the results locally. That service is then used when creating the solr doc and rendering the edit form.

Geonames wants a username to use this service. This is defined in `config/initializers/hyrax.rb` as `config.geonames_username = Qa::Authorities::Geonames.username = '{username}'` to set the user in Qa at the same time.

Changes proposed in this pull request:
* Add service based on Qa::Authorities::Geonames to supply a full label
* Index that full label
* Display that full label on the work edit form

Guidance for testing, such as acceptance criteria or new user interface behaviors:
* Creating a work with a location
* Verify full label (e.g. City, State, Country) is shown on work details and edit form

@samvera/hyrax-code-reviewers

PR #5533, opened 2022-03-14
call canonicalize instead of canonicalize! for URIs

elrayle:
Calling `canonicalize!` updates the URI object.  When the URI is frozen, this results in error `<FrozenError: can’t modify frozen RDF::URI…>`

The `canonicalize`, defined in `RDF::URI`, makes a copy of the URI object first and then calls `canonicalize!`

refs #5254; refs #5493

This issue was introduced in PR #5254.  And it is a blocker for follow-on work in PR #5493.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Good catch!

cjcolvar (approved):

PR #5534, opened 2022-03-14
Use the link text to activate the link

jcoyne:
This better helps test the accessibility of the link, as it looks for text that would be used by a screen reader

### DIFF
--- a/spec/features/create_work_spec.rb
+++ b/spec/features/create_work_spec.rb
@@ -19,7 +19,7 @@ RSpec.describe 'Creating a new Work', :js, :workflow, :clean_repo do
     before do
       sign_in user
       click_link 'Works'
-      find('#add-new-work-button').click
+      click_link "Add New Work"
       choose "payload_concern", option: "GenericWork"
       click_button 'Create work'
     end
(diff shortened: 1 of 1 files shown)


elrayle:
@jcoyne Is there precedence that you can point us to?  We just switched to preferring ids because it helps with some flakiness in the tests.  I'm not sure how testing for text vs. id helps with assuring things work for screen readers other than possible identifying if text disappears and only an id is available.

jcoyne:
@elrayle it's possible to have a link that only has a visual representation:
```html
<a href="/foo" id="my-link"><img src="close.png"></a>
```
This would be poor accessability.  Contrast that with: 
```html
<a href="/foo" id="my-link"><span class="visually-hidden">Close"</span> <img src="close.png"></a>
```

The id selector would have no problem with either of these, but the text based selector would only work on the second.  This is good, because the second one is better for accessability.  There is text that can be read.  The test, shows that this is accessible.

sdellis:
@jcoyne while your approach works to make sure that there is a string value label for the button, and hence can be read by a screenreader, it wasn't clear from code/comments that this is an Accessibility test hidden within an unrelated test. Furthermore, each button only needs to be tested once, but the same test is repeated across many unrelated tests, making it redundant and causing all of them to fail upon a single character change in the label. 

Automated accessibility tools can catch these types of issues (and more) and would separate them out of the unrelated Capybara tests. It's also easy to add button labels in the examples you used [using `aria-labelledby` and the input ID](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby), which has the added bonus of not being dependent on CSS for implementation. 

If the `aria-labelledby` approach works for you, I can create a ticket and add them this week. I can also create a ticket for adding the Axe automated accessibility testing tool to the pipeline for a more comprehensive and "best practices" approach.

PR #5550, opened 2022-03-17
Provide deprecation on ProxyDepositListener

hackartisan:
followup to https://github.com/samvera/hyrax/pull/5529

related documentation: https://dry-rb.org/gems/dry-events/0.2/

### DIFF
--- a/app/services/hyrax/listeners/proxy_deposit_listener.rb
+++ b/app/services/hyrax/listeners/proxy_deposit_listener.rb
@@ -19,9 +19,8 @@ module Hyrax
           response to a difficult-to-diagnose race condition bug. This listener \
           is now a no-op. To retain functionality ensure that \
           DefaultMiddlewareStack is configured to use \
-          Hyrax::Actors::TransferRequestActor and unregister this listener \
-          in config/initializers/listeners.rb by adding the line: \n
-          Hyrax.publisher.unsubscribe(Hyrax::Listeners::ProxyDepositListener.new)"
+          Hyrax::Actors::TransferRequestActor. To quiet this deprecation remove any \
+          local initializer configuration that subscribes this listener."
         )
       end
     end
--- a/app/services/hyrax/listeners/proxy_deposit_listener.rb
+++ b/app/services/hyrax/listeners/proxy_deposit_listener.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Listeners
+    ##
+    # @deprecated transfer requests are now carried out synchronously during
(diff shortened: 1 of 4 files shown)


app/services/hyrax/listeners/proxy_deposit_listener.rb
```diff
@@ -0,0 +1,29 @@
...
+      # @return [void]
+      def on_object_deposited(_event)
+        Deprecation.warn(
+          "The ProxyDepositListener was deprecated, effective immediately, in \
+          response to a difficult-to-diagnose race condition bug. This listener \
+          is now a no-op. To retain functionality ensure that \
+          DefaultMiddlewareStack is configured to use \
+          Hyrax::Actors::TransferRequestActor and unregister this listener \
+          in config/initializers/listeners.rb by adding the line: \n
+          Hyrax.publisher.unsubscribe(Hyrax::Listeners::ProxyDepositListener.new)"
```
elrayle, line 24: Does this require that they copy listeners.rb to their /config/initializers?  Or will the engine version of this file run first and then the local one of the same name will run second?

elrayle: If having a file of the same name causes an override of Hyrax's listeners.rb initializer, would it work to recommend they create a local file called `config/initializers/listeners_ext.rb` and put the unsubscribe there?

elrayle (commented):
See question about deprecation message.

app/services/hyrax/listeners/proxy_deposit_listener.rb
hackartisan: These are great questions and I spent some time trying this recommendation.

The engine initializers run first. The existence of an initializer by the same name in an app is fine; both load. However, it seems like you maybe have to unsubscribe the same instance of the listener that you subscribed? For example, if I put an unsubscribe in the local initializer it doesn't affect the listener list. If I put another subscribe of the same listener then it's in there twice (once for each instance).

So the options I see are
- Don't subscribe the listener in our initializer. If someone has for some reason subscribed it in their own, they will see the deprecation. Probably no one has done this and this means no one will ever see the deprecation.
- Have a noisy deprecation no one can do anything about. I guess make it a lot shorter.

elrayle: The first option has a great deal of appeal.  I wouldn't like have a noise deprecation with no way to quiet it.

elrayle (approved):
Looks good.  Thanks!

eliotjordan (approved):

PR #5554, opened 2022-03-17
Remove SingularSubresourceController concern from CitationsController to enable citations for Valkyrie resources

### LINKED ISSUE #5452: citation button on work show page displays "NoMethodError in Hyrax::CitationsController#work" error
### Descriptive summary

When I click on the "Citation" button on the work show page on nurax-pg, a "NoMethodError in Hyrax::CitationsController#work" error message displays.

1. On nurax-pg, navigate to a work's show page (https://nurax-pg.curationexperts.com/concern/generic_works/b471ea02-bf36-4047-b4f4-0f0c41272541?locale=en)
2. Click the "Citations" button under the thumbnail image
3. Observe error message

citationscontrollerwork

eliotjordan:
Closes #5452 

- Enables citations button in dassie test app so it can be tested
- Remove SingularSubresourceController concern from CitationsController to enable citations for Valkyrie resources

SingularSubresourceController is bound tightly to ActiveFedorda and removing the controller mix-in enables citations for Valkyrie and non-Valkyrie works.

### DIFF
--- a/app/controllers/hyrax/citations_controller.rb
+++ b/app/controllers/hyrax/citations_controller.rb
@@ -3,7 +3,6 @@ module Hyrax
   class CitationsController < ApplicationController
     include WorksControllerBehavior
     include Breadcrumbs
-    include SingularSubresourceController
 
     # Overrides decide_layout from WorksControllerBehavior
     with_themed_layout '1_column'
--- a/app/controllers/hyrax/citations_controller.rb
+++ b/app/controllers/hyrax/citations_controller.rb
@@ -2,6 +2,7 @@
 module Hyrax
   class CitationsController < ApplicationController
     include WorksControllerBehavior
+    include DenyAccessOverrideBehavior
     include Breadcrumbs
 
     # Overrides decide_layout from WorksControllerBehavior
--- a/.dassie/config/initializers/hyrax.rb
+++ b/.dassie/config/initializers/hyrax.rb
@@ -16,6 +16,7 @@ Hyrax.config do |config|
 
   config.iiif_image_server = true
   config.work_requires_files = false
+  config.citations = true
 
   # Returns a URL that resolves to an image provided by a IIIF image server
   config.iiif_image_url_builder = lambda do |file_id, base_url, size, format|
--- a/spec/controllers/hyrax/citations_controller_spec.rb
+++ b/spec/controllers/hyrax/citations_controller_spec.rb
@@ -29,6 +29,22 @@ RSpec.describe Hyrax::CitationsController do
         expect(session['user_return_to']).to eq request.url
       end
     end
+
+    context "with a Valkyrie resource" do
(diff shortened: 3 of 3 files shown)


app/controllers/hyrax/citations_controller.rb
```diff
@@ -3,7 +3,6 @@ module Hyrax
   class CitationsController < ApplicationController
     include WorksControllerBehavior
     include Breadcrumbs
-    include SingularSubresourceController
```
elrayle, line 6: @eliotjordan and I looked at this together to determine why it is ok to remove this.

From Eliot...
> SingularSubresourceController is bound tightly to ActiveFedorda and removing the controller mix-in enables citations for Valkyrie and non-Valkyrie works.
>
> SingularSubresourceController finds the work using the Hyrax::WorkRelation class: https://github.com/samvera/hyrax/blob/main/app/controllers/concerns/hyrax/singular_subresource_controller.rb#L14
> 
> WorkRelation is a subclass of ActiveFedora::Relation:
> [5:33](https://samvera.slack.com/archives/CDRHAMUTV/p1647552796997939)
> https://github.com/samvera/hyrax/blob/f050d8f7f644321f1524af26c1adc18cc0eff0f8/app/search_builders/hyrax/abstract_type_relation.rb#L3

Tracking in GitHub, this line came over with Sufia.  It was used to find the work through an ActiveFedora relationship.  This no longer happens.  Only `@presenter` is referenced in the [view](https://github.com/samvera/hyrax/blob/1e504c200fd9c39120f514ac33cd42cd843de9fa/app/views/hyrax/citations/work.html.erb).

elrayle: Digging a little deeper, the part of SingularSubresourceController that is used is `load_and_authorize_resource` for both work and file.  

### With SingularSubresourceController
If the user doesn't have access to the work or file...
* they are forwarded to Home and shown a message about not having access

### Without SingularSubresourceController
If the user doesn't have access to the work or file...
* they stay on the citations page and are shown a message about the work being private

eliotjordan: @elrayle For an user that is not logged in, the behavior is the same as what is is tested in the controller spec:
https://github.com/samvera/hyrax/blob/main/spec/controllers/hyrax/citations_controller_spec.rb#L27-L28

A logged-in user without permission has the same experience you mention above, but this is not tested in the specs. Is being re-directed to Home the preferred behavior?

eliotjordan: Pushed an update to redirect to the home page.

eliotjordan:
There continues to be a flaky test failure in ruby2-5-8 that is unrelated to any of the changes here, despite repeated re-running of the failed tests.

elrayle (approved):
Thanks for the update to redirect to home when user doesn't have permission.  Good for users to have consistent behavior.

Issue #5557, opened 2022-03-18
Move depositor change propagation to a new background job


hackartisan:
### Descriptive summary

Follow-up needed for work done in #5553

Logic was moved from a background job to synchronous code that propagates depositor changes from a resource to its file sets. the updates to file sets should run in a new background job.

### Rationale

There could be thousands of file sets on a resource, and modifying the file sets should not conflict with modifying the resource itself in such a way as to create race conditions.

### Related work

 #5553

PR #5559, opened 2022-03-18
Move code propagating proxy depositor to file sets to new job

hackartisan:
It would probably be better to have this job spin off one job per file set.

it would also be good to stop checking for active fedora. But there were time constraints and this should at least ensure that nothing runs too long inline.

closes https://github.com/samvera/hyrax/issues/5557


elrayle (changes requested):
Thanks for getting this in.  There is one small change that will make this more efficient.  And one other change that will help with the long term effort to move to Valkyrie.  The second is optional if you don't have time.

PR #5560, opened 2022-03-19
Remove hard dependency on ActiveFedora/Wings from SolrDocument

cjcolvar:
None of the calls to #hydra_model pass the classifier kwarg so Wings/ActiveFedora is invoked as the default value for the kwarg even though it is not used.  This refactor changes the kwarg default value to nil and passes it the kwarg to a private method to dynamically determine if the default Wings implementation should be used.  In most cases this method should never be called and thus the hard dependency on Wings has been removed.

### DIFF
--- a/spec/models/concerns/hyrax/solr_document_behavior_spec.rb
+++ b/spec/models/concerns/hyrax/solr_document_behavior_spec.rb
@@ -64,6 +64,17 @@ RSpec.describe Hyrax::SolrDocumentBehavior do
       it 'resolves the correct model name' do
         expect(solr_document.hydra_model).to eq Monograph
       end
+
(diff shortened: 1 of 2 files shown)


spec/models/concerns/hyrax/solr_document_behavior_spec.rb
```diff
@@ -64,6 +64,17 @@
       it 'resolves the correct model name' do
         expect(solr_document.hydra_model).to eq Monograph
       end
+
+      context 'using non-wings adapter', valkyrie_adapter: :test_adapter do
```
cjcolvar, line 68: I'm not sure if this test is really necessary but it does prove that the method still returns when Wings is disabled.

PR #5561, opened 2022-03-19
Use cancan override in tests that uses Hyrax.config to find resources

cjcolvar:
Remove anonymous controller to test that adapter works

This resolves the issue where the before action set on the controller by `load_and_authorize_resource` does not dynamically evaluate the class used so looks for the wrong model when mocking Hyrax.config.collection_model in specific tests.


elrayle (approved):
This will be helpful.  Thanks.

PR #5562, opened 2022-03-19
Disable wings when testing with a different valkyrie adapter

cjcolvar:


cjcolvar:
Looks like this should pass when rebased on top of #5560.

spec/spec_helper.rb
```diff
@@ -317,6 +320,11 @@ def clean_active_fedora_repository
     allow(Hyrax)
       .to receive(:metadata_adapter)
       .and_return(Valkyrie::MetadataAdapter.find(adapter_name))
+
+    if adapter_name != :wings_adapter
+      allow(Hyrax.config).to receive(:disable_wings).and_return(true)
+      hide_const("Wings") # disable_wings=true removes the Wings constant
```
elrayle, line 326: Better location for this instead of in individual tests.

PR #5563, opened 2022-03-21
Index admin set relationship using the same field as ActiveFedora-based works

### LINKED ISSUE #5497: Hyrax::AdministrativeSet shows 0 Total Items and No entries found under Works in This Set when it has works
### Descriptive summary

When works have been added in a Hyrax::AdministrativeSet, the show page for the admin set says there are 0 works and doesn't list any works under Works in This Set.

### Rationale

AF and Valkyrie admin sets should show the works that are in the admin set.

### Expected behavior

After adding a work to an admin set, the admin set show page should list the work.

### Actual behavior

After adding a work to an admin set, the admin set show page shows 0 works.

image

#### Prerequisite

Configure Hyrax:

```
```
2. restart rails app

Create an admin set.
1. navigate to: Dashboard -> Collections
3. select type: Admin Set
4. set title: MY ADMIN SET (or other meaningful name)
5. Save

#### To Reproduce:

1. navigate to: Dashboard -> Works
2. click button: Add a work
3. select type: Monograph (on Hyrax) or any type on nurax-pg
4. file in required metadata and file if required
5. click tab: Relationships
6. under Administrative Set, select: MY ADMIN SET (or the name you set in prerequisites)
7. Save

This should forward to the work show page and you will see MY ADMIN SET in the Relations section under In Administrative Set.

1. navigate to: Dashboard -> Works
2. click title: MY ADMIN SET (or the name you set in prerequisites)

This should forward to the admin set show page.  You will see that the work is not shown in the list of works and the total items is 0.

### Related work

cjcolvar:
Fixes #5497 

The `isPartOf_ssim` field is used when searching for works belonging to an admin set. 
https://github.com/samvera/hyrax/blob/d628b58ca15fe0614547c3ad3fd7e6d900070660/app/presenters/hyrax/admin_set_presenter.rb#L11

Future work:
- write a feature test
- change the hardcoded `isPartOf_ssim` in the line referenced above and the other occurrence in that presenter to use the configuration `Hyrax.config.admin_set_predicate.qname.last` which is used in the search builder
https://github.com/samvera/hyrax/blob/b034218b89dde7df534e32d1e5ade9161e129a1d/app/search_builders/hyrax/admin_admin_set_member_search_builder.rb#L27

### DIFF
--- a/app/indexers/hyrax/valkyrie_work_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_work_indexer.rb
@@ -18,7 +18,7 @@ module Hyrax
         admin_set_label = admin_set_label(resource)
         solr_doc['admin_set_sim']   = admin_set_label
         solr_doc['admin_set_tesim'] = admin_set_label
-        solr_doc['isPartOf_ssim'] = [resource.admin_set_id.to_s]
+        solr_doc["#{Hyrax.config.admin_set_predicate.qname.last}_ssim"] = [resource.admin_set_id.to_s]
         solr_doc['member_of_collection_ids_ssim'] = resource.member_of_collection_ids.map(&:to_s)
         solr_doc['member_ids_ssim'] = resource.member_ids.map(&:to_s)
         solr_doc['depositor_ssim'] = [resource.depositor]
--- a/app/indexers/hyrax/valkyrie_work_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_work_indexer.rb
@@ -18,6 +18,7 @@ module Hyrax
         admin_set_label = admin_set_label(resource)
         solr_doc['admin_set_sim']   = admin_set_label
         solr_doc['admin_set_tesim'] = admin_set_label
+        solr_doc['isPartOf_ssim'] = [resource.admin_set_id.to_s]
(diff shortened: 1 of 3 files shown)


elrayle:
Is there an issue for the suggested future work, especially the change of hardcoded `isPartOf_ssim` to use the config instead?

cjcolvar:
I think I was hesitant before because I don't really like the configuration but this change at least makes it consistent and it can be cleaned up in the future.

app/indexers/hyrax/valkyrie_work_indexer.rb
```diff
@@ -18,6 +18,7 @@ def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Met
         admin_set_label = admin_set_label(resource)
         solr_doc['admin_set_sim']   = admin_set_label
         solr_doc['admin_set_tesim'] = admin_set_label
+        solr_doc['isPartOf_ssim'] = [resource.admin_set_id.to_s]
```
elrayle, line 21: Since you are adding this, can you update the hardcoded use of 'isPartOf_ssim' to use the configured relationship predicate?

PR #5565, opened 2022-03-21
Index resources created from test factries by default

cjcolvar:
This follows the pattern already set in the `hyrax_collection` factory.
https://github.com/samvera/hyrax/blob/d628b58ca15fe0614547c3ad3fd7e6d900070660/spec/factories/hyrax_collection.rb#L44

A couple tests needed slight changes as a result of this change in indexing behavior.

### DIFF
--- a/spec/factories/hyrax_work.rb
+++ b/spec/factories/hyrax_work.rb
@@ -18,6 +18,7 @@ FactoryBot.define do
       read_users         { [] }
       members            { nil }
       visibility_setting { nil }
+      with_index         { true }
     end
 
     after(:build) do |work, evaluator|
@@ -46,6 +47,8 @@ FactoryBot.define do
       work.permission_manager.read_users  = evaluator.read_users
 
       work.permission_manager.acl.save
+
+      Hyrax.index_adapter.save(resource: work) if evaluator.with_index
     end
 
     trait :public do
--- a/spec/forms/hyrax/forms/resource_form_spec.rb
+++ b/spec/forms/hyrax/forms/resource_form_spec.rb
@@ -438,7 +438,7 @@ RSpec.describe Hyrax::Forms::ResourceForm do
 
     context 'when using a generic valkyrie adapter', valkyrie_adapter: :test_adapter do
       before do
-        allow(Hyrax).to receive_message_chain(:config, :disable_wings).and_return(true) # rubocop:disable RSpec/MessageChain
+        allow(Hyrax.config).to receive(:disable_wings).and_return(true)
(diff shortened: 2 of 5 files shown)


spec/forms/hyrax/forms/resource_form_spec.rb
```diff
@@ -438,7 +438,7 @@
 
     context 'when using a generic valkyrie adapter', valkyrie_adapter: :test_adapter do
       before do
-        allow(Hyrax).to receive_message_chain(:config, :disable_wings).and_return(true) # rubocop:disable RSpec/MessageChain
+        allow(Hyrax.config).to receive(:disable_wings).and_return(true)
```
cjcolvar, line 441: This was needed because the indexing done in the factory uses Hyrax.config which was not allowed in this stubbing.

spec/services/hyrax/listeners/file_metadata_listener_spec.rb
```diff
@@ -5,7 +5,7 @@
   let(:data)         { { metadata: metadata, user: user } }
   let(:event)        { Dry::Events::Event.new(event_type, data) }
   let(:fake_adapter) { FakeIndexingAdapter.new }
-  let(:file_set)     { FactoryBot.valkyrie_create(:hyrax_file_set) }
+  let(:file_set)     { FactoryBot.valkyrie_create(:hyrax_file_set, with_index: false) }
```
cjcolvar, line 8: The listener does the indexing and the tests expect the file set isn't already indexed so don't index the file set.

spec/factories/hyrax_work.rb
```diff
@@ -46,6 +47,8 @@
       work.permission_manager.read_users  = evaluator.read_users
 
       work.permission_manager.acl.save
+
+      Hyrax.index_adapter.save(resource: work) if evaluator.with_index
```
elrayle, line 51: _The following comment is to document differences in processing of membership for collections and works..._

In `:hyrax_collection`, if `members` are passed in, the `after(:create)` loops through members to save and index each.  This is because the relationship is stored on the members.  

This is not required for works because the relationship is stored on the parent work, meaning there are no changes to the members when setting the relationship.

spec/forms/hyrax/forms/resource_form_spec.rb
elrayle: This is a cleaner stub.

elrayle (approved):
Good to have the `with_index` attribute supported by all the Valkyrie factories.

PR #5570, opened 2022-03-23
Fix flaky file ordering in spec for ValkyrieIngestJob

dlpierce:
The order that files are returned can vary causing the test to fail sometimes. This makes the spec not care about the ordering.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Thanks so much for fixing this!  The past couple days I've had to restart the tests many times to get them to pass.  This is a huge help!

I had one small suggestion that isn't a blocker.

dlpierce:
Thanks, that's much cleaner.

cjcolvar (approved):

PR #5571, opened 2022-03-23
Fix flaky test caused by querying solr for admin set with a nil id

dlpierce:
`presenter.total_items` queries solr using the admin set id, but a built (not created) admin set has a nil id. This leads to flaky spec failures depending on the random order specs are run. Using create should avoid this.


cjcolvar (approved):
:1st_place_medal: This one was tripping me up too!  Thanks for fixing it!

PR #5589, opened 2022-03-30
Existing valkyrie feature tests should work without Wings

cjcolvar:
Adjusts the existing feature tests which use valkyrie to run cleanly on a non-wings adapter without making any calling fedora.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/factories/permission_templates.rb
+++ b/spec/factories/permission_templates.rb
@@ -16,16 +16,7 @@ FactoryBot.define do
     before(:create) do |permission_template, evaluator|
       if evaluator.with_admin_set
         source_id = permission_template.source_id
-        admin_set =
-          if source_id.present?
-            begin
-              AdminSet.find(source_id)
-            rescue ActiveFedora::ObjectNotFoundError
-              create(:admin_set, id: source_id)
-            end
-          else
-            create(:admin_set)
-          end
+        admin_set = SourceFinder.find_or_create_admin_set(source_id)
         permission_template.source_id = admin_set.id
       elsif evaluator.with_collection
         source_id = permission_template.source_id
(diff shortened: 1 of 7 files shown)


spec/features/create_work_spec.rb
```diff
@@ -6,7 +6,7 @@
   let(:file2) { File.open(fixture_path + '/image.jp2') }
   let!(:uploaded_file1) { Hyrax::UploadedFile.create(file: file1, user: user) }
   let!(:uploaded_file2) { Hyrax::UploadedFile.create(file: file2, user: user) }
-  let(:permission_template) { create(:permission_template, source_id: 'admin_set/default', with_admin_set: true, with_active_workflow: true) }
+  let(:permission_template) { create(:permission_template, with_admin_set: true, with_active_workflow: true) }
```
cjcolvar, line 9: The source_id was never referenced explicitly so removing it allows it to be any admin set.

spec/features/create_work_spec.rb
```diff
@@ -135,7 +135,7 @@
     end
   end
 
-  context "with valkyrie resources", metadata_adapter: :postgres_adapter do
+  context "with valkyrie resources", valkyrie_adapter: :postgres_adapter do
```
cjcolvar, line 138: `metadata_adapter` was the wrong metadata tag and wasn't actually being handled.

spec/features/valkyrie_search_spec.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-RSpec.describe 'searching', index_adapter: :solr_index do
+RSpec.describe 'searching', index_adapter: :solr_index, valkyrie_adapter: :test_adapter do
```
cjcolvar, line 2: Specifying a `valkyrie_adapter` other than wings ensures that this test works without ActiveFedora.

spec/factories/permission_templates.rb
```diff
@@ -85,4 +76,36 @@ def self.create_access(permission_template_id, agent_type, access, agent_ids)
       end
     end
   end
+
+  class SourceFinder
+    def self.find_or_create_admin_set(source_id)
+      Hyrax.config.use_valkyrie? ? find_or_create_admin_set_valkyrie(source_id) : find_or_create_admin_set_active_fedora(source_id)
```
dunn, line 82: is `use_valkyrie?` deprecated?

cjcolvar: I'm still seeing it used in a number of places so I don't think so.  I'm open to using a different approach than `use_valkyrie?` here to determine which model to create.  Do you have any suggestions?

dunn: oh my mistake, i was remembering seeing https://github.com/samvera/hyrax/blob/main/lib/wings/services/custom_queries/find_file_metadata.rb#L110 but that's specifically for wings queries.

dunn (approved):

PR #5598, opened 2022-04-11
Featured Works MIA After Reindex

orangewolf:
Stop removing featured records if the item is missing from the index. If an app is being reindexed OR solr is having a bad day, we shouldn't destroy database records, just skip them.

* Create a work
* Feature that work
* Clear or drop your solr index
* Load the home page
* Reindex
* Previously the work would no longer be featured, with this code change the featuring comes back

@samvera/hyrax-code-reviewers


no-reply:
!!

thanks for this. looks like a very long lived issue, but i can't see any good reason why listing featured works should ever destroy them.

no-reply (approved):

PR #5599, opened 2022-04-12
use core metadata in forms for pcdm collections

elrayle:
### Description

This is a simple change to make the pcdm collection form use the schema loader to load core metadata which generates the property and validation for the title field instead of defining the property directly in the the form.  The model and indexer already include core metadata.  This doesn't change any behaviors of Hyrax.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Looks great!

Issue #5494, opened 2022-03-04
Edit collection only shows title field for Hyrax::PcdmCollection


elrayle:
### Descriptive summary

Creating and editing a Hyrax::PcdmCollection only includes a Title field on the Description tab.   Should there be more fields?

### Rationale

A reasonable default set of collection fields should be available to the end user out of the box.

### Expected behavior

When editing a collection, expect to be able to set more than Title.  Exact fields TBD.  <--  @jlhardes 

### Actual behavior

When editing a collection, only presented with a Title field and no other metadata.

![image](https://user-images.githubusercontent.com/6855473/156822362-7720be12-757f-4dd9-8ca8-3e2ce08c50a7.png)

#### Prerequisite:
Create a collection first if none exist.

#### To Reproduce:
1. Dashboard -> Collections
2. beside any existing collection, Select Action: Edit collection

NOTE: Only title field is available on the Description tab.

### Related work

PR #5157 - Update PcdmCollection metadata to match existing ::Collection metadata

jlhardes:
All descriptive metadata fields are now available to fill in when creating or editing a collection but they do not all show on the collection show page. Closing this issue and will create another issue or issues referencing PR #5157 to have collection descriptive metadata show and be configurable.

PR #5600, opened 2022-04-12
do not include basic metadata in pcdm collections

elrayle:
### Description

The change in this PR is being made to make the processing of PcdmCollection's consistent across the components (i.e., model, indexer, form).

### Expected

Model, indexer, and form will all handle core and basic metadata, if included, through the schema loader consistently across all components.  All will be tested.

### Actual

Model, indexer, and form do not handle core and basic metadata consistently through the schema loader across all components.  Some are tested.

* model included core and basic metadata loaded through the schema, and tested both
* indexer included core and basic metadata loaded through the schema, but only tested core
* form didn't include either.  There don't appear to be tests specific to parts of the form loaded through the schema loader.  For resource form, title is tested directly in specs.  The same is done in the PcdmCollectionForm tests.

#### Prerequisites:
Configure to collection model:
```
```
2. Restart rails server.

#### Test form in browser

Before and after this change, the following steps take you to a form that has only one field, the Title field.

3. select type: any type
4. click button: Create Collection

NOTE: Title is the only available metadata field before and after the change in this PR.

### Related Work

Working toward resolution of Issue #5494.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/controllers/hyrax/collections_controller_spec.rb
+++ b/spec/controllers/hyrax/collections_controller_spec.rb
@@ -8,7 +8,6 @@ RSpec.describe Hyrax::CollectionsController, clean_repo: true do
     FactoryBot.valkyrie_create(:hyrax_collection,
                                :public,
                                title: ["My collection"],
-                               description: ["My incredibly detailed description of the collection"],
                                edit_users: [user.user_key], read_users: [user.user_key])
   end
 
@@ -20,7 +19,7 @@ RSpec.describe Hyrax::CollectionsController, clean_repo: true do
   let(:unowned_asset) { FactoryBot.valkyrie_create(:work, user: other) }
 
   let(:collection_attrs) do
-    { title: ['My First Collection'], description: ["The Description\r\n\r\nand more"] }
+    { title: ['My First Collection'] }
   end
 
   describe "#show" do # public landing page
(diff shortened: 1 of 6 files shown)


spec/controllers/hyrax/collections_controller_spec.rb
```diff
@@ -8,7 +8,6 @@
     FactoryBot.valkyrie_create(:hyrax_collection,
                                :public,
                                title: ["My collection"],
-                               description: ["My incredibly detailed description of the collection"],
```
elrayle, line 11: The description attribute is not tested in any tests, so it was simply removed from the setup process.

spec/presenters/hyrax/collection_presenter_spec.rb
```diff
@@ -3,15 +3,26 @@
   subject(:presenter) { described_class.new(solr_doc, ability) }
   let(:collection) do
     build(:hyrax_collection,
+          id: 'adc12v',
+          title: ['A clever title'])
+  end
+  let(:active_fedora_collection) do
+    build(:collection_lw,
```
elrayle, line 10: With basic metadata removed from PcdmCollection, tests in the presenter that try to access basic metadata by delegating to the solr_doc fail.  Basic metadata is not required for collections, so it is unclear when these methods would be called.  Perhaps a refactor could be explored to see if there is another pattern for supporting additional metadata methods.

For now, use an ActiveFedora collection to maintain the tests of metadata methods.  A follow-on PR will add a generated collection resource class that will include basic metadata for testing.

cjcolvar (approved):
This looks good to me.

PR #5601, opened 2022-04-13
Fix translations in german locale file

jzimpel:
As a native german speaker I found some literal translations in the german locale file that do not fit the context.

It' s not a complete review of the locale, just a few that caught my attention when updating hyrax to v3.4.1

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/locales/hyrax.de.yml
+++ b/config/locales/hyrax.de.yml
@@ -220,7 +220,7 @@ de:
             summary: Zusammenfassung
             title: Sammlungsbericht
             view_collection_btn: Kollektion ansehen
-            visibility: Sichtweite
+            visibility: Sichtbarkeit
           summary:
             all_time: Alle Zeit
             collection_home_page_views: Aufrufe der Sammlungs-Startseite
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Thanks for fixing these!

cjcolvar (changes requested):
Thanks for your contribution!  I need to check on CLAs before I'm able to merge this.

cjcolvar:
@jzimpel I don't see CLAs on file for you.  Unfortunately I believe the current Samvera policy requires CLAs even for this kind of contribution.  Fortunately though, the Samvera board is considering removing this requirement at its meeting later this month.  If you don't want to wait until then to get this PR merged then you'll have to submit a [cCLA and iCLA](https://samvera.atlassian.net/wiki/x/AwonG).  Sorry!

jzimpel:
Ok, I'll have a look at it. Thanks so far!

cjcolvar (approved):

cjcolvar:
The CLA requirement has been dropped so nothing is blocking this now.  Thanks for the contribution!

PR #5448, opened 2022-02-21
update collection membership for works during edit

elrayle:
Fixes #5446, #5465, #5466

### For 5446...

Collection membership is set by the form in the `member_of_collections_attributes` parameter.  It includes a hash structure that lists collections to add and to remove. 

```
      # member_of_collections_attributes"=>
        {"0"=>{"id"=>"r207tp33p", "_destroy"=>"false"},
         "1"=>{"id"=>"1234asef", "_destroy"=>"true"}}
```
 This PR adds a virtual property to the form changeset to provide access to the `member_of_collections_attributes` parameter and uses a validator to determine that the membership changes are allowed and updates the `member_of_collection_ids` property which syncs with the work resource attribute of the same name.

### For 5465...

The work instance was being set to the results of the validation failure when it failed and to the updated work when the validation passed...

WAS:
```
@curation_concern =
          form.validate(params[hash_key_for_curation_concern]) &&
          transactions['change_set.create_work']
```

Now, the form validation is processed before setting `@curation_concern`.  If a failure occurs, the process is aborted and calls `after_update_error` to process the error and redisplay the edit form.

### For 5466...

Assuming validation passed, but a transaction failure occurred, there was an attempt to call #value on the transaction Failure instance which does not respond to #value.  set the work instance was being set to 

```
.call(form).value!
```

The solution  is to use `.call(form).value_or { #err handling }` which will return the value if there isn't a failure and process the code block if there is.  Similar to the validations failure processing, the err handling block for transactions aborts the process and calls `after_update_error`.

@samvera/hyrax-code-reviewers

PR #5604, opened 2022-04-15
Improve error handling for creating valkyrie collections

elrayle:
### Description

The new collection instance was being set to the results of the validation failure when it failed and to false when the transactions failed.

WAS:

```
@collection =
          form.validate(params[collection_params]) &&
          transactions['change_set.create_collection']
          ...
         .call(form).value_or { return after_create_error }
```

Now, the form validation is processed before setting @collection. If a failure occurs, the process is aborted and calls after_create_errors.  Likewise, if there is a failure during the transaction process, after_create_errors is called to process the error and redisplay the new form.

_NOTE: This PR also moves after and before methods for create to private, deprecating the public versions.  Again, this is consistent with the pattern used in works._

### Follow-on work

Need to do the same for updates.  This will be done in a separate PR to keep this one easier to review.

### Related work

The approach is based on the same change that was made for works in [PR #5448](https://github.com/samvera/hyrax/pull/5448/files#diff-843d5f344066ba9de62a999cf88a87b29479b10988676c0bc474f4853e2adaf4).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -99,18 +99,12 @@ module Hyrax
         # collection is saved without a value for `has_model.`
         @collection = Hyrax.config.collection_class.new
         authorize! :create, @collection
-        return valkyrie_create if @collection.is_a?(Valkyrie::Resource)
 
-        # Coming from the UI, a collection type gid should always be present.  Coming from the API, if a collection type gid is not specified,
-        # use the default collection type (provides backward compatibility with versions < Hyrax 2.1.0)
-        @collection.collection_type_gid = params[:collection_type_gid].presence || default_collection_type.to_global_id
-        @collection.attributes = collection_params.except(:members, :parent_id, :collection_type_gid)
-        @collection.apply_depositor_metadata(current_user.user_key)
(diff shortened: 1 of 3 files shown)


app/controllers/hyrax/dashboard/collections_controller.rb
```diff
@@ -84,26 +84,13 @@ def edit
...
-          # if we are creating the new collection as a subcollection (via the nested collections controller),
-          # we pass the parent_id through a hidden field in the form and link the two after the create.
-          link_parent_collection(params[:parent_id]) unless params[:parent_id].nil?
-        end
-        respond_to do |format|
-          Hyrax::SolrService.commit
-          format.html { redirect_to edit_dashboard_collection_path(@collection), notice: t('hyrax.dashboard.my.action.collection_create_success') }
-          format.json { render json: @collection, status: :created, location: dashboard_collection_path(@collection) }
-        end
+        Deprecation.warn("Method `#after_create` will be removed in the next major release.")
```
cjcolvar, line 87: Is it possible to explicitly name the release?  I think it would be most useful to anyone who gets this warning to know in which version it will be removed.
```suggestion
        Deprecation.warn("Method `#after_create` will be removed in Hyrax 4.0.")
```

elrayle: DONE

cjcolvar (approved):
Looks great!

Issue #5301, opened 2021-12-17
Remove/replace jQuery-File-Upload


marrus-sh:
### Descriptive summary

[blueimp/jQuery-File-Upload](https://github.com/blueimp/jQuery-File-Upload) does [not support Bootstrap 4](https://github.com/blueimp/jQuery-File-Upload/issues/3662#issuecomment-664232619). The repository has been archived by its owner, so it is unlikely support will manifest anytime soon.

### Rationale

Bootstap 4 support is required as part of the Blacklight 7 upgrade and general upgrade to Rails 6.

### Related work

See #5276; #5280.

jcoyne:
I recommend dropzone.js This is what we used for our self deposit app at Stanford.

orangewolf:
I propose splitting this ticket in to 2 tickets - one that does Bootstrap 4 support work and a separate one for replacing this entire JS widget with an updated library and maybe even an updated flow. The patch for Bootstrap 4 is simple and scope for BL7/B4 is already pretty big. I'd like to see the Blueimp widget replaced, but it seems wise to me to resolve other UI issues with the file uploader at the same time.

orangewolf:
@jlhardes I believe this is no longer a bootstrap 4 issue, the b4 part is resolved.  It's still worth resolving doing, but as a separate epic about refactoring the file uploader feature.

jlhardes:
From @orangewolf: "I fixed the rails 6 / blacklight 7 / bootstrap 4 issue. We should still someday replace that widget but I feel like deferring it to a broader “overhaul js” that we will want to do for Rails 7 is probably wise. As far as I can tell from my testing it works as is now on the bootstrap branch." I am closing this issue and will create a new issue for replacing the entire file upload JS widget.

PR #5605, opened 2022-04-18
Bl7 fix fileupload

orangewolf:
Refs #5301 

This patch fixes Bootstrap 4 support for the file uploads.  We should still replace the file upload widget at some point, but this patch is simple and it reduces scope on the already large Blacklight 7 / Bootstrap 4 effort.

* Navigate to the new work page -> files tab. Upload a file and see that the file upload UI works. Save the work
* Navigate file version interface and upload a new files. See that the upload UI works

@samvera/hyrax-code-reviewers


dlpierce:
I cherry picked the commit into the bl7bootstrap4 branch. Thanks!

PR #5608, opened 2022-04-27
Corrects several deprecations in test suite.

bwatson78:
All but last one: fixes several deprecation warnings showing up in test suite.
spec/features/search_spec.rb: updates the selector class.

### DIFF
--- a/app/search_builders/hyrax/nested_collections_parent_search_builder.rb
+++ b/app/search_builders/hyrax/nested_collections_parent_search_builder.rb
@@ -22,7 +22,7 @@ module Hyrax
     # @return [void]
     def parent_collections_only(solr_parameters)
       solr_parameters[:fq] ||= []
-      solr_parameters[:fq] += [Hyrax::SolrQueryBuilderService.construct_query_for_ids(child.member_of_collection_ids)]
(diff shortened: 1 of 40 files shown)


app/search_builders/hyrax/nested_collections_parent_search_builder.rb
```diff
@@ -22,7 +22,7 @@ def initialize(scope:, child:, page:)
     # @return [void]
     def parent_collections_only(solr_parameters)
       solr_parameters[:fq] ||= []
-      solr_parameters[:fq] += [Hyrax::SolrQueryBuilderService.construct_query_for_ids(child.member_of_collection_ids)]
+      solr_parameters[:fq] += [Hyrax::SolrQueryService.new.with_ids(ids: child.member_of_collection_ids).build]
```
dlpierce, line 25: I think this one is not quite right. Many specs are failing here.

bwatson78: Looking at it now.

bwatson78: Just added to it with [a9d4d7f](https://github.com/samvera/hyrax/pull/5608/commits/a9d4d7f30c55525104625de0e60546031b43f2db) to address that issue.

PR #5613, opened 2022-05-03
Add generator for collection resource optionally including basic metadata

elrayle:
### Description

This PR...
* adds a generator for creating a collection resource class
* adds to dassie: CollectionResource, a collection resource class with basic metadata created by the generator
* adds to .internal_test app generation process: generate CollectionResource with basic metadata
* adds a factory to create an instance of CollectionResource
* updates collection_controller_with_resource_spec.rb to use the generated CollectionResource class

### Collection Resource Generator

Generates:
1. Creates a collection model and spec.
2. Creates a collection form and spec.
3. Creates a collection indexer and spec.
4. Creates a collection metadata config.
5. Sets this to be the collection class.

The generated collection resource classes optionally include `basic_metadata`.

### Add CollectionResource in test app

The test generator was updated to create a collection resource class named "CollectionResource".  The config that is generated to set this as 'the' collection class is commented out in `/config/initializers/hyrax.rb` by the test generator.  

For now, this allows the ActiveFedora `::Collection` class to continue to be 'the' collection class.

_NOTE: At this time, Hyrax supports a single class that is 'the' collection class.  It is configured in `config/initializers/hyrax.rb` by setting `config.collection_model =`._

### Potential Future Work

This pattern allows for future work that would update Hyrax to allow multiple collection resource classes each with their own metadata.

### Updates CollectionController test for resources

Updates `spec/controllers/hyrax/dashboard/collection_controller_with_resource_spec.rb` to use CollectionResource class for testing.  This verifies the generated class and allows removal of pending for specs that require basic metadata.

@samvera/hyrax-code-reviewers

PR #5614, opened 2022-05-04
handle valkyrie collection models in search results

### LINKED ISSUE #5567: Clicking "view all collections" on nurax-pg homepage returns no results
### Descriptive summary

On the [nurax-pg homepage](https://nurax-pg.curationexperts.com/?locale=en), clicking the "View All Collections" button returns no results, even though the "Explore Collections" list displays multiple collections.

### Expected behavior
When collections appear in the "Explore Collections" list, they should appear in a "View All Collections" search.

### Actual behavior
No collections appear when clicking "View All Collections"

1. Navigate to https://nurax-pg.curationexperts.com/?locale=en
2. Click the "View All Collections" button in the "Explore Collections" tab.
3. Observe that no results have been returned.

elrayle:
Fixes #5567 

This refactor handles all valkyrie collection models regardless of the model name.

This is required because blacklight generates partials based on the model name.  This is controlled in the catalog controller by setting the `display_type_field`.  For Hyrax, this is configured to…

```
    config.index.display_type_field = "has_model_ssim"
```
Then partials are created for each model that will be handled.

Pre-valkyrie, there was one and only one collection.  Everything else was a work.  So there were two partials, one to handle collections, and the default which handles all works regardless of model name.

Now there are at least two collection models, the AF `::Collection` and the Valkyrie `Hyrax::PcdmCollection`.  This was initiallly addressed by adding another partial to support hyrax_pcdmcollection.

With the ability to generate a collection using any name, added in PR https://github.com/samvera/hyrax/pull/5613, the name of the collection is not known in advance and a partial cannot be created specific to that collection model.  To address this, the PR adds processing to the default to check the model.  If it is a collection model, then it is processed as a collection.  Otherwise, it is processed as a work.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/catalog/_index_header_list_default.html.erb
+++ b/app/views/catalog/_index_header_list_default.html.erb
@@ -1,3 +1,10 @@
+<% model = document["has_model_ssim"].first.safe_constantize %>
 <div class="search-results-title-row">
-  <h3 class="search-result-title"><%= link_to document.title_or_label, document %></h3>
+  <% if model == Hyrax::PcdmCollection || model < Hyrax::PcdmCollection %>
+    <h3 class="search-result-title"><%= link_to document.title_or_label, [hyrax, document] %></h3>
+    <%= Hyrax::CollectionPresenter.new(document, current_ability).collection_type_badge %>
+  <% else %>
+    <h3 class="search-result-title"><%= link_to document.title_or_label, document %></h3>
+  <% end %>
 </div>
+
--- a/app/views/catalog/_index_header_list_default.html.erb
+++ b/app/views/catalog/_index_header_list_default.html.erb
@@ -1,4 +1,4 @@
-<% model = document["has_model_ssim"].first.safe_constantize %>
+<% model = document.hydra_model %>
 <div class="search-results-title-row">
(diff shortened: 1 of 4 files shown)


app/views/catalog/_index_header_list_default.html.erb
```diff
@@ -1,3 +1,10 @@
+<% model = document["has_model_ssim"].first.safe_constantize %>
```
cjcolvar, line 1: There is a [method for this on SolrDocument](https://github.com/samvera/hyrax/blob/main/app/models/concerns/hyrax/solr_document_behavior.rb#L77-L81) which has more handling of edge cases (primarily wings I believe).
```suggestion
<% model = document.hydra_model %>
```

app/views/catalog/_thumbnail_list_default.html.erb
```diff
@@ -1,5 +1,10 @@
+<% model = document["has_model_ssim"].first.safe_constantize %>
```
cjcolvar, line 1: Same change as above.
```suggestion
<% model = document.hydra_model %>
```

cjcolvar (commented):
Looks good.  I had one suggestion for improvement.

cjcolvar (approved):
Looks good!

PR #5615, opened 2022-05-10
rework CollectionMembershipValidator to skip overwriting collections

no-reply:
the validator should limit itself to validating that the collections are
allowed.

whether or not validation is ever run, `#sync` should apply the form values to
the target resource. ensure this happens by moving the value population behavior
into a reform populator.

without this work, the `Hyrax::CollectionMembershipValidator` was erasing
existing collection membership. validators should never change the records being
validated.

@samvera/hyrax-code-reviewers


no-reply:
i think it would be a good idea to write further tests for the validator. without unit tests for its behavior i think it could become hard to understand its requirements.

elrayle:
Summary of Slack discussion.  See PR #5448 which added the validator.  There was a good bit of discussion around how to populate the collections.

no-reply:
@elrayle I tried to find the relevant discussion in #5448, but I think I'm missing it. would you mind linking directly to the comments you're hoping people will consider?

dunn (approved):

PR #5617, opened 2022-05-11
move validation tests under validators/hyrax

elrayle:
Since validator code is in `app/validators/hyrax` directory, the tests should be in `spec/validators/hyrax`.  This PR moves the `collection_membership_validator_spec.rb` to the expected directory structure.

@samvera/hyrax-code-reviewers


no-reply:
closing this as it will cause a major merge conflict with #5615, which will do the same thing.

PR #5619, opened 2022-05-11
DO NOT MERGE: this is just for finding out when `#validate` called

no-reply:
just triggering a build; please ignore.


no-reply:
confirming that this method is only raising from its own unit tests and code paths removed in https://github.com/samvera/hyrax/pull/5615

PR #5626, opened 2022-05-12
valkyrie: use storage adapters for derivatives

dunn:
Currently when derivatives are generated for uploaded files, Hyrax tries to save them directly to the filesystem, and subsequent access also assumes they will exist in a conventional local location: https://github.com/samvera/hyrax/blob/9ebda1a948fcec5e74aafc0d57da8e5dda9ad8ea/app/services/hyrax/derivative_path.rb#L7

When an application is using a different Valkyrie storage adapter, such as valkyrie-shrine for S3, these assumptions fail, and derivative storage doesn't work.  This PR teaches Hyrax to use the defined `derivatives_storage_adapter` instead of directly accessing the filesystem.

- Valkyrie-based upload tooling is abstracted into `ValkyrieUpload` in order to be used for original and derivative file upload
- a new controller pattern is used to create routes for derivatives; these are used instead of the `ThumbnailIndexer` for `FileSets` (future work should replace all uses of `ThumbnailIndexer` in Valkyrie contexts)

@samvera/hyrax-code-reviewers

PR #5628, opened 2022-05-12
valkyrie: extract upload methods into ValkyrieUpload service

dunn:
This will let us use the same tooling for uploading in other contexts (like derivatives).

Extracted from #5626


dunn:
not sure what to make of this, since we're not touching the derivativesjob here: https://app.circleci.com/pipelines/github/samvera/hyrax/7622/workflows/9afa92d5-823b-4d6d-a32a-70d4eadac85e/jobs/60714?invite=true#step-112-512

no-reply (approved):

no-reply:
looks like a nice refactor! thanks!

PR #5651, opened 2022-05-19
Fix admin dashboard filters

abelemlih:
The Collection filters in the admin dashboard are broken. Essentially, with the update to Blacklight 7, all the code we used to render those filters is deprecated, and most of the logic was moved from `facets_helper_behavior.rb` to components.

My initial approach to solve this issue was attempting to migrate some of the views we use to render facets to view components from Blacklight 7. After I managed to use components and render them successfully, I realized that the components are hard to customize for some of the use cases in Hyrax, like rendering `Admin Set` under `Collection Type` in the collections admin dashboard. Any further attempts to override components for some of the customization we need led to complex logic that is hard to read and will make it much harder for future developers to track any bugs.

The approach in this PR requires minimal edits compared to overriding Blacklight components. Instead of migrating to components to adapt to Blacklight 7, I created a helper class `Hyrax::FacetsHelper` that has the same methods as the `Blacklight::FacetsHelperBehavior` in Blacklight `6.24.0`. The helper methods in Blacklight 6 are different from those deprecated in Blacklight 7 because they do not rely on components to render the facets. The facets helper I added simply ensures the deprecated method we use to render custom facet views in Hyrax can still be used in the codebase, and no components are needed.

The PR accomplishes the following:
- Restore filtering by `Collection Type` under Dashboard > Collections
- Resolve all the specs that were failing because of the broken filtering functionality for collections

### DIFF
--- a/app/helpers/hyrax/facets_helper.rb
+++ b/app/helpers/hyrax/facets_helper.rb
@@ -1,0 +1,150 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module FacetsHelper
+    # Methods in this module are from Blacklight::FacetsHelperBehavior, blacklight v6.24.0
+    # This module is used to ensure Hyrax facet views that rely on deprecated Blacklight helper methods are still functional
+
+    include Blacklight::Facet
+
+    ##
+    # Renders the list of values
+    # removes any elements where render_facet_item returns a nil value. This enables an application
+    # to filter undesireable facet items so they don't appear in the UI
+    def render_facet_limit_list(paginator, facet_field, wrapping_element = :li)
+      safe_join(paginator.items.map { |item| render_facet_item(facet_field, item) }.compact.map { |item| content_tag(wrapping_element, item) })
+    end
+
+    ##
+    # Renders a single facet item
+    def render_facet_item(facet_field, item)
+      if facet_in_params?(facet_field, item.value)
+        render_selected_facet_value(facet_field, item)
+      else
+        render_facet_value(facet_field, item)
+      end
+    end
+
+    ##
+    # Standard display of a facet value in a list. Used in both _facets sidebar
+    # partial and catalog/facet expanded list. Will output facet value name as
+    # a link to add that to your restrictions, with count in parens.
+    #
+    # @param [Blacklight::Solr::Response::Facets::FacetField] facet_field
+    # @param [Blacklight::Solr::Response::Facets::FacetItem] item
+    # @param [Hash] options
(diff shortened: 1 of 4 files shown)


app/views/hyrax/my/_facet_limit.html.erb
```diff
@@ -3,7 +3,7 @@
       class in only necessary for the former and causes display issues in the latter. -%>
   <ul id="<%= facet_field.field.parameterize + '-dropdown-options' %>" class="<%= 'dropdown-menu ' if action_name == 'index' %>facet-values" aria-labelledby="<%= facet_field.field.parameterize %>">
     <% paginator = facet_paginator(facet_field, display_facet) %>
-    <%= render_facet_limit_list paginator, facet_field.key %>
+    <%= self.extend(Hyrax::FacetsHelper).render_facet_limit_list(paginator, facet_field.key) %>
```
dlpierce, line 6: Extending classes should not happen in views.

app/views/hyrax/my/_facet_pivot.html.erb
```diff
@@ -9,8 +9,7 @@
         <% item.fq= {} if subfacet %>
 
         <% # The unless prevents Collection from being included in the select list for the Collection Type filter. %>
-        <% facet_config= facet_configuration_for_field(item.field) %>
-        <%= (facet_config.item_presenter || Blacklight::FacetItemPresenter).new(item, facet_config, self, item.field).label unless subfacet != true && item.value == "Collection" %>
+        <%= self.extend(Hyrax::FacetsHelper).render_facet_item(item.field, item) unless subfacet != true && item.value == "Collection" %>
```
dlpierce, line 12: Extending classes should not happen in views.

app/helpers/hyrax/facets_helper.rb
```diff
@@ -0,0 +1,249 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module FacetsHelper
+    # All methods in this module are from Blacklight::FacetsHelperBehavior, blacklight v6.24.0
+    # This module is used to ensure Hyrax facet views that rely on deprecated Blacklight helper methods are still functional
```
dlpierce, line 6: Are all of these helpers required? If not, can the unneeded ones be left out?

dlpierce (changes requested):

app/helpers/hyrax/facets_helper.rb
abelemlih: I removed the helper methods that are not relevant to this ticket.

dlpierce (approved):

PR #5666, opened 2022-05-31
Allow CronJob's to use podAnnotations

mcritchlow:
This is helpful for scenarios like using Hashicorp Vault secrets using the [Agent injection solution](https://www.vaultproject.io/docs/platform/k8s/injector), which leverages annotations.

I'm planning on following up with a PR to propose supporting the Vault secret use case explicitly, but I think allowing `podAnnotations` to be used in `CronJob` (and `Job`) templates is beneficial more broadly.

* Allow existing `CronJob` templates to use `.Values.podAnnotations` if specified
* Bump chart version to `1.4.0`

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 1.3.0
+version: 1.4.0
 appVersion: 3.3.0
 dependencies:
   - name: fcrepo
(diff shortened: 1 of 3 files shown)


chart/hyrax/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 1.3.0
+version: 1.3.1
```
dunn, line 5: if we're being fussy about semver i think this counts as adding functionality so it might want to be a minor version bump

dunn (dismissed):

chart/hyrax/Chart.yaml
mcritchlow: oh that's true, thanks for that. I'll switch to `1.4.0`

dunn (approved):

orangewolf (approved):

PR #5669, opened 2022-06-08
Eliminates duplicate Collection relation.

### LINKED ISSUE #5668: Collection relationships display twice on View Work page [Rails 6 testing]
### Descriptive summary

Hyrax 3.4.1 main: if a work is part of a Collection, this relationship is displaying twice instead of once on the View Work page.

### Rationale

When a work is assigned to a collection, this relationship should only display once.

### Expected behavior

When a work is assigned to a given collection, that relationship should only display once from the Relations section of the View Work page. 

### Actual behavior

The collection relationship is displayed twice on the work detail page ([see Nurax example](https://nurax-dev.curationexperts.com/concern/generic_works/ht24wj68c?locale=en)). 
Screen Shot 2022-06-08 at 2 14 23 PM

It does not, however, display twice from Edit Work > Relationships > Collections.

1. Go to a work with no collections assigned
2. Edit the work, select the Relationships tab, and assign the work to a Collection. 
3. Save the changes to the work, and then view the Work. 
4. See the Collection listed twice in the Relations section. 

### Related work

bwatson78:
Fixes #5668 

* app/presenters/hyrax/work_show_presenter.rb: adds humanized string of "Collection" to the presenter types, since `#grouped_presenters` humanizes the keys to exclude.
* spec/features/work_show_spec.rb: adds an expectation that there will be only one link.

@samvera/hyrax-code-reviewers


app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -147,7 +147,7 @@ def tweeter
     end
 
     def presenter_types
-      Hyrax.config.registered_curation_concern_types.map(&:underscore) + ["collection"]
+      Hyrax.config.registered_curation_concern_types.map(&:underscore) + ["collection", "Collection"]
     end
 
     # @return [Array] presenters grouped by model name, used to show the parents of this object
```
cjcolvar, line 155: What about instead of adding `Collection` to `#presenter_types` you change `#grouped_presenters` to downcase the exceptions like it does for filtering?
```
grouped.reject! { |obj| except.include? obj.downcase } unless except.nil?
```
Or would that potentially break things unless the downcasing is applied to both sides?
```
grouped.select! { |obj| obj.downcase == filtered_by.downcase } unless filtered_by.nil?
grouped.reject! { |obj| except.map(&:downcase).include? obj.downcase } unless except.nil?
```

bwatson78: Looking at that now.

bwatson78:
@cjcolvar Ready for your re-review.

cjcolvar (approved):
Looks good!

PR #5682, opened 2022-06-14
Support readinessProbe in Sidekiq Deployment

mcritchlow:
Following the recommendations from the Sidekiq wiki [1], this updates
the chart to support defining a `readinessProbe` for the sidekiq worker
`Deployment`.

An example is given in the values.yaml, and would look something like:

```yaml
worker:
  readinessProbe:
    enabled: true
    command:
     - cat
     - /app/samvera/hyrax-webapp/tmp/sidekiq_process_has_started_and_will_begin_processing_jobs
    initialDelaySeconds: 10
    periodSeconds: 2
    timeoutSeconds: 1
    failureThreshold: 10
    successThreshold: 2
```

The downstream developer will also need to setup, or modify, a sidekiq
initializer to create/delete the file used for the health checks.

Example from the wiki:

```ruby
SIDEKIQ_WILL_PROCESSES_JOBS_FILE = Rails.root.join('tmp/sidekiq_process_has_started_and_will_begin_processing_jobs').freeze

Sidekiq.configure_server do |config|
  # We touch and destroy files in the Sidekiq lifecycle to provide a
  # signal to Kubernetes that we are ready to process jobs or not.
  #
  # Doing this gives us a better sense of when a process is actually
  # alive and healthy, rather than just beginning the boot process.
  config.on(:startup) do
    FileUtils.touch(SIDEKIQ_WILL_PROCESSES_JOBS_FILE)
  end

  config.on(:shutdown) do
    FileUtils.rm_f(SIDEKIQ_WILL_PROCESSES_JOBS_FILE)
  end
end
```

1. https://github.com/mperham/sidekiq/wiki/Kubernetes#health-checks

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/values.yaml
+++ b/chart/hyrax/values.yaml
@@ -201,6 +201,18 @@ worker:
   affinity: {}
   resources: {}
 
+  # see: https://github.com/mperham/sidekiq/wiki/Kubernetes#health-checks
+  readinessProbe:
+    enabled: false
+    # command:
+    #  - cat
+    #  - /app/samvera/hyrax-webapp/tmp/sidekiq_process_has_started_and_will_begin_processing_jobs
+    # initialDelaySeconds: 10
+    # periodSeconds: 2
+    # timeoutSeconds: 1
+    # failureThreshold: 10
+    # successThreshold: 2
+
 fcrepo:
   enabled: true
   externalDatabaseUsername: "hyrax"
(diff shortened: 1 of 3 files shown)


chart/hyrax/values.yaml
```diff
@@ -200,6 +200,16 @@ worker:
   tolerations: []
   affinity: {}
   resources: {}
+  readinessProbe:
+    enabled: false
+    # command:
+    #  - cat
+    #  - /app/samvera/hyrax-webapp/tmp/sidekiq_process_has_started_and_will_begin_processing_jobs
```
dunn, line 207: could add a comment linking to the wiki bit about creating the initializer too maybe?

mcritchlow: Yeah good call. I'll do that

PR #5694, opened 2022-06-14
always pull repo file if filepath param is missing

### LINKED ISSUE #5676: Rolling back file versions with the same filename can break characterization and derivative jobs
### Descriptive summary

This almost certainly affects all versions of Hyrax. For sure from 2.9.0 to 3.4.1 (and main branch).

### Rationale

Rolling back to an [older version (a.k.a. revision) of a FileSet](https://github.com/samvera/hyrax/blob/0e9fd3b98da88e678a911ceb5088b13c162f203a/app/actors/hyrax/actors/file_actor.rb#L39) is the only place to call `CharacterizeJob` or `CreateDerivativesJob` without a `filepath`, meaning it's the only place (outside of one occurrence in [rake tasks](https://github.com/samvera/hyrax/blob/0e9fd3b98da88e678a911ceb5088b13c162f203a/lib/tasks/regenerate_derivatives.rake#L8)) that causes Hyrax::WorkingDirectory to pull a copy of the file from the repository to a NOID-based-pairtree folder inside [working_path](https://github.com/samvera/hyrax/blob/0e9fd3b98da88e678a911ceb5088b13c162f203a/lib/hyrax/configuration.rb#L321).

aside: The whole `WorkingDirectory` thing was slated for removal in a TODO left on [this PR](https://github.com/samvera/hyrax/pull/1301/files). It says to use JobIOWrapper instead. I may spin off another ticket for that after this as it may be sort of forgotten at this point.

So the WorkingDirectory.find_or_retrieve() method [relies on the filename](https://github.com/samvera/hyrax/blob/main/app/services/hyrax/working_directory.rb#L15-L16) to decide whether the version that has just been "rolled back to" is the one that's already cached in said directory. Any old version that may be cached in the `working_path` will be used if the name matches the current version's `original_name`. These are never cleared out by the system itself. Admittedly we do [delete uploaded files periodically in heliotrope](https://github.com/mlibrary/heliotrope/blob/master/lib/tasks/cleanup/clean_cache.rake#L57-L66), and perhaps this is recommended in a Hyrax setup Wiki somewhere. Not sure.

It may seem unlikely that two rollbacks would occur through the UI where both versions have the same filename. But of course it's very likely that the same name needs to be used, if it's pertinent to the content (some sort of ID, or in our case a book ISBN). And, as mentioned, other calls that might be made to `CharacterizeJob` or `CreateDerivativesJob` from elsewhere, with no `filepath` parameter will cause this problem too. Like a dev working in the console or triggering a rake task. The [task linked above](https://github.com/samvera/hyrax/blob/0e9fd3b98da88e678a911ceb5088b13c162f203a/lib/tasks/regenerate_derivatives.rake#L8) would cache a `working_path` file for every FileSet in the system.

### Expected behavior

Nothing in the UI should cause `CharacterizeJob` or `CreateDerivativesJob` to run on a file that is not the FileSet's current version.

### Actual behavior

`CharacterizeJob` or `CreateDerivativesJob` will run on a file that is not the FileSet's current version if you ever roll back to different versions with the same name.

1. Upload an image FileSet to a Work. Let the jobs finish and note the thumbnail, file size and checksum in the UI
2. Upload a new version to the FileSet. Something with the same filename but a different image and size. Again, note the characterization metadata in the UI.
3. In the versions tab, revert the FileSet to the first version. Allow jobs to finish. The thumbnail and metadata will be correct. Note that this is where the working_directory copy was made.
4. Now revert to the second version. Allow jobs to finish. The thumbnail and characterization is done on the wrong file, the one cached to disk in step 3.

### Related work

TODO

conorom:
Fixes #5676

Remove Hyrax::WorkingDirectory.find_or_retrieve() as it cannot handle FileSet file versions using the same filename.
This module is marked as deprecated anyway, and its removal a suggested (future) TODO on [this PR](https://github.com/samvera/hyrax/pull/1301).

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/working_directory.rb
+++ b/app/services/hyrax/working_directory.rb
@@ -3,6 +3,7 @@ module Hyrax
   # @deprecated Use JobIoWrapper instead
   class WorkingDirectory
     class << self
+      # @deprecated No longer used anywhere in Hyrax code, naively assumes filenames of versions/revisions are distinct
       # Returns the file passed as filepath if that file exists. Otherwise it grabs the file from repository,
       # puts it on the disk and returns that path.
       # @param [String] repository_file_id identifier for Hydra::PCDM::File
@@ -10,6 +11,7 @@ module Hyrax
       # @param [String, NilClass] filepath path to existing cached copy of the file
       # @return [String] path of the working file
       def find_or_retrieve(repository_file_id, id, filepath = nil)
+        Deprecation.warn("Hyrax::WorkingDirectory.find_or_retrieve() is deprecated and no longer used in Hyrax")
         return filepath if filepath && File.exist?(filepath)
         repository_file = Hydra::PCDM::File.find(repository_file_id)
         working_path = full_filename(id, repository_file.original_name)
--- a/app/jobs/characterize_job.rb
+++ b/app/jobs/characterize_job.rb
@@ -29,7 +29,8 @@ class CharacterizeJob < Hyrax::ApplicationJob
   # @param [String, NilClass] filepath the cached file within the Hyrax.config.working_path
   def perform(file_set, file_id, filepath = nil)
     raise "#{file_set.class.characterization_proxy} was not found for FileSet #{file_set.id}" unless file_set.characterization_proxy?
-    filepath = Hyrax::WorkingDirectory.find_or_retrieve(file_id, file_set.id) unless filepath && File.exist?(filepath)
+    # Ensure a fresh copy of the repo file's latest version is being worked on, if no filepath is directly provided
+    filepath = Hyrax::WorkingDirectory.copy_repository_resource_to_working_directory(Hydra::PCDM::File.find(file_id), file_set.id) unless filepath && File.exist?(filepath)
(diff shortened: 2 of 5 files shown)


app/services/hyrax/working_directory.rb
```diff
@@ -3,23 +3,6 @@ module Hyrax
   # @deprecated Use JobIoWrapper instead
   class WorkingDirectory
     class << self
-      # Returns the file passed as filepath if that file exists. Otherwise it grabs the file from repository,
-      # puts it on the disk and returns that path.
-      # @param [String] repository_file_id identifier for Hydra::PCDM::File
-      # @param [String] id the identifier of the FileSet
-      # @param [String, NilClass] filepath path to existing cached copy of the file
-      # @return [String] path of the working file
-      def find_or_retrieve(repository_file_id, id, filepath = nil)
```
no-reply, line 12: let's deprecate this, instead of removing. it's public API now, so this would avoid the need to bump version to 4.0.0.

no-reply (approved):

PR #5695, opened 2022-06-14
wings: support non-AF FileMetadata nodes

no-reply:
when converting to/from Hyrax::FileMetadata, we have a need to handle two
separate cases.

in the first case, we want to follow ActiveFedora's pattern for ldp:describes
links. Fedora generates a node to describe a file when the file is uploaded, and
ActiveFedora reuses AF::Base but bends it to this use case. we already have some
specialized handling throughout Wings to address the unusual save patterns for
these nodes, and most of the Hyrax codebase conforms to some off-spec
requirements for handling valkyrie representations for these nodes.

this adds handling for the second case: when the file is a non-AF file, but the
app is using Wings for metadata (e.g. for derivatives stored via a disk
adapter), we need Wings to be able to save the FileMetadata resource instance
normally. i.e. it should handle it like any other resource, without the special
handling or requirements imposed by ActiveFedora.

we test this handling from three key Wings interfaces.
@dunn

### DIFF
--- a/lib/wings/active_fedora_converter.rb
+++ b/lib/wings/active_fedora_converter.rb
@@ -3,6 +3,7 @@
 require 'wings/converter_value_mapper'
 require 'wings/active_fedora_converter/default_work'
 require 'wings/active_fedora_converter/file_metadata_node'
+require 'wings/active_fedora_converter/instance_builder'
 require 'wings/active_fedora_converter/nested_resource'
 
 module Wings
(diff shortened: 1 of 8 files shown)


dunn (approved):

dunn (approved):

dunn:
hmmm
```
  1) Wings::FileConverterService#resource_to_af_file copies attributes to af_file
     Failure/Error: expect(subject.metadata_node.type).to match_array expected_attrs[:type]
     
     NoMethodError:
       undefined method `metadata_node' for #<#<Class:0x000055c8382fab88>:0x000055c83a8b7650>
       Did you mean?  metadata_streams
     # ./vendor/bundle/gems/activemodel-5.2.6/lib/active_model/attribute_methods.rb:430:in `method_missing'
     # ./spec/wings/services/file_converter_service_spec.rb:62:in `block (3 levels) in <top (required)>'
```

no-reply:
this is still the same deprecated module. the test setup isn't any good, so we just need to whack-a-mole it.

dunn (approved):

PR #5718, opened 2022-06-17
Change the way we click content block tabs in testing

abelemlih:
The way we click tabs in the spec is causing issues for some of the specs where certain tabs cannot be found. In this PR, I select the tabs by id and click them in the specs, which solves the issue.


bwatson78:
Great job, Ayoub!

PR #5720, opened 2022-06-21
add `Hyrax::FileMetadata(file)` caster

no-reply:
this new method follows the `Array(arg)` style type casting pattern to give an
appropriate FileMetadata for a given file within the storage adapter.

this uses the custom query `find_file_metadata_by` to handle casting, but has an
"ensure" style feature to give a new file identifier with appropriate references
if none already exists. this isn't intended to populate metadata, unlike the
to-be-deprecated `FileMetadata#for`.

an implementation for the custom query that works for `Wings` is given. we
populate `alternative_ids` in the `new` branch of the type cast method to make
query easier for other adapters, but don't yet provide a generic solution. some
more exploration may be needed to decide the best approach for the non-Wings
case. some adapters may benefit from a custom query of `file_identifier`, but
Hyrax is committed to a base Valkyrie implementation of the custom query.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -1,6 +1,25 @@
 # frozen_string_literal: true
 
 module Hyrax
+  ##
+  # Casts a resource to an associated FileMetadata
+  #
+  # @param [Valkyrie::StorageAdapter::File] file
+  #
+  # @return [Hyrax::FileMetadata]
+  # @raise [ArgumentError]
+  def self.FileMetadata(file)
+    raise(ArgumentError, "Expected a Valkyrie::StorageAdapter::File; got #{file.class}: #{file}") if
+      file.is_a?(Valkyrie::Resource)
+
+    Hyrax.custom_queries.find_file_metadata_by(id: file.id)
+  rescue Hyrax::ObjectNotFoundError, Ldp::BadRequest
+    Hyrax.logger.debug('Could not find an existing metadata node for file ' \
+                       "with id #{file.id}. Initializing a new one")
+
+    FileMetadata.new(file_identifier: file.id, alternative_ids: [file.id])
+  end
+
   class FileMetadata < Valkyrie::Resource
     # Include mime-types for Hydra Derivatives mime-type checking. We may want
     # to move this logic someday.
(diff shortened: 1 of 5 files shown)


app/models/hyrax/file_metadata.rb
```diff
@@ -1,6 +1,24 @@
...
 module Hyrax
+  ##
+  # Casts a resource to an associated FileMetadata
+  #
+  # @param [Valkyrie::StorageAdapter::File] file
+  #
+  # @return [Hyrax::FileMetadata]
+  # @raise [ArgumentError]
+  def self.FileMetadata(file)
+    raise ArgumentError if file.is_a?(Valkyrie::Resource) # maybe heavy handed?
```
dunn, line 12: what sort of error message could we put here? what kind of `Valkyrie::Resources` might be accidentally passed here?

lib/wings/services/custom_queries/find_file_metadata.rb
```diff
@@ -33,7 +33,13 @@ def initialize(query_service:)
       #
       # @raise [Hyrax::ObjectNotFoundError]
       def find_file_metadata_by(id:, use_valkyrie: true)
-        find_file_metadata_by_alternate_identifier(alternate_identifier: id, use_valkyrie: use_valkyrie)
+        if id.to_s.starts_with?('fedora://')
+          find_file_metadata_by_alternate_identifier(alternate_identifier: id, use_valkyrie: use_valkyrie)
+        else
```
dunn, line 38: this branch is for Valkyrie objects that need AF counterparts?

app/models/hyrax/file_metadata.rb
no-reply: the skys the limit?

if we're asking questions like this, i think i'd rather just drop the line and let the query fail in whatever way it fails.

lib/wings/services/custom_queries/find_file_metadata.rb
no-reply: that's right.

app/models/hyrax/file_metadata.rb
no-reply: ah, i take it back. the existing test covering this provides a good example:

if you pass in, e.g. a FileSet, you'd get a FileMetadata with an unfortunate reference to the FileSet. this seems like a common enough pitfall to guard against; it would result in unwanted metadata resources otherwise.

app/models/hyrax/file_metadata.rb
```diff
@@ -1,6 +1,25 @@
...
 module Hyrax
+  ##
+  # Casts a resource to an associated FileMetadata
+  #
+  # @param [Valkyrie::StorageAdapter::File] file
+  #
+  # @return [Hyrax::FileMetadata]
+  # @raise [ArgumentError]
+  def self.FileMetadata(file)
+    raise(ArgumentError, "Expected a Valkyrie File; got #{file.class}: #{file}") if
```
dunn, line 12: `Expected a Valkyrie::StorageAdapter::File` might be even clearer

no-reply:
```suggestion
    raise(ArgumentError, "Expected a Valkyrie::StorageAdapter::File; got #{file.class}: #{file}") if
```

dunn (approved):

PR #5727, opened 2022-06-22
remove deprecated code from FileMetadata

no-reply:
a lingering call from a private method is replaced to accomplish this, we probably want to backport that replacement.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/file_metadata.rb
+++ b/app/models/hyrax/file_metadata.rb
@@ -127,15 +127,6 @@ module Hyrax
     # attributes set by fits for video
     attribute :aspect_ratio, ::Valkyrie::Types::Set
 
-    # @param [ActionDispatch::Http::UploadedFile] file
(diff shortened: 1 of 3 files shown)


app/services/hyrax/valkyrie_upload.rb
```diff
@@ -28,7 +28,7 @@ def self.file(
     )
     io.close
 
-    file_metadata = find_or_create_metadata(id: streamfile.id, file: streamfile)
+    file_metadata = Hyrax::FileMetadata(file: streamfile)
```
dlpierce, line 31:
```suggestion
    file_metadata = Hyrax::FileMetadata(streamfile)
```
That should resolve the `NoMethodError:
  undefined method 'id' for #<Hash>` errors.

cjcolvar (approved):
Looks great!

cjcolvar (approved):
Looks good.

PR #5728, opened 2022-06-22
remove call to deprecated Hyrax::FileMetadata#for

no-reply:
this method was deprecated, and the private method that calls it has an encapsulated replacement. use that instead.

@samvera/hyrax-code-reviewers


dunn (approved):

no-reply:
this needed a rebase.

Issue #4016, opened 2019-09-18
Support Rails 6.0


no-reply:
### Descriptive summary

Rails 6.0 has been released. We'll want to add support as promptly as possible.

Some of the Samvera core components offer support for Rails 6 (e.g. ActiveFedora) but many do not. We'll need to work with the Core Components WG and the Blacklight community to get support in place.

### Related work

#3947

PR #5730, opened 2022-06-23
Upgrade to Blacklight 7, Bootstrap 4 and Rails 6.0

### LINKED ISSUE #3271: Add support for Blacklight 7
### Descriptive summary

Blacklight 7 has been available for some time. We need to get a handle on how hard it will be to upgrade.

### Rationale

We should keep our core dependencies up to date. Also, there may be `a11y` fixes in BL 7 that we want in Hyrax.

### Related work

There's an initial branch here: https://github.com/samvera/hyrax/compare/bl-7. `blacklight-access_controls` and ``blacklight-gallery` both need updates to support 7.x first.

dlpierce:
Fixes #3271 #5276 #4016

Migrates Hyrax to use new major versions of Blacklight, Bootstrap and Rails.


no-reply:
this looks good to me!

since it's green and `main` is already in 4.0.0 territory, i'd advocate we merge this as quickly as possible and handle any emerging issues as part of the beta/release testing cycle.

no-reply (approved):

PR #5742, opened 2022-06-27
remove deprecated class attribute from CollectionType

no-reply:
@samvera/hyrax-code-reviewers


dunn (dismissed):

dlpierce:
Is this spec still needed?

https://github.com/samvera/hyrax/blob/346cef1edb29d11a855ace467b277eb0b7459070/spec/models/hyrax/collection_type_spec.rb#L198-L208

cjcolvar (approved):
Looks good!

PR #5744, opened 2022-06-27
remove deprecated methods from CollectionBehavior

no-reply:
@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/collection_behavior.rb
+++ b/app/models/concerns/hyrax/collection_behavior.rb
@@ -52,25 +52,6 @@ module Hyrax
       self.collection_type_gid = new_collection_type.to_global_id
     end
 
-    # Add members using the members association.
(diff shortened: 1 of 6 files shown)


dunn (approved):

blancoj (approved):
This is how I tested it:
(1) As Admin, I created a collection as private
(2) Verified that only I could see it, other users could not
(3) Made it public, verified that other users could see it
(4) Added a work to the collection, and verified that the work was attached to the collection.
(5) I added a sub collection to a collection, verified that the count of sub collections for the collection was accurate.
(6) I did notice that if the sub collection is private and the parent collection is public, and the user does not have privileges to view the sub collection, the user is presented with an item count that includes the private sub collection when viewing the parent collection, which I guess could be a little confusing, but of course that does not have to do with this PR.

PR #5755, opened 2022-06-28
correctly derive paths for UploadedFileUploader

no-reply:
`UploadedFileUploader`'s path generation was extremely fragile. for instance, it
would generate an incorrect path if the configuration option didn't end with
'/'.

to generate a vaild in a system independent way, use `Pathname#/` and then cast
back to a string.

@samvera/hyrax-code-reviewers


no-reply:
this one could probably benefit from a backport to both the 2.x and 3.x release series.

dunn (dismissed):

dunn (dismissed):

dunn (approved):

PR #5757, opened 2022-06-29
add some basic documentation for CustomQueries

no-reply:
@samvera/hyrax-code-reviewers


no-reply:
@dlpierce tagging you for review since this expresses policy opinions.

Issue #5357, opened 2022-01-24
Add a dependency management bot to Hyrax


mcritchlow:
### Descriptive summary

There are several automated dependency management solutions available today.

We should adopt one for Hyrax (and perhaps Samvera more broadly?)

### Rationale

In addition to the Ruby dependencies declared in the `gemspec`, this repository also hosts a Helm chart for Hyrax which has dependencies on various sub-charts (Redis, Solr, etc.).

Keeping on top of this manually feels unsustainable, and there are great solutions in place for this.

Some solutions that I have personal experience with and are fairly easily installed as Github Apps.

- Renovate: https://github.com/marketplace/renovate
- Dependabot: https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/ (I am not sure if Dependabot supports Helm chart updates)

### Related work
Some related tickets that have come up recently in this context: #5192

mcritchlow:
Trying to surface this again, I'm not sure `dependabot` is an option. As far as I can tell from https://github.com/dependabot/dependabot-core/issues/2237 it doesn't support Helm chart dependency updates. Which is quite surprising, and of high value (from my standpoint) for this repo.

PR #5768, opened 2022-07-05
Update Hyrax chart

mcritchlow:
refs #5357

Depends on: 
- https://github.com/samvera-labs/fcrepo-charts/pull/13
- An updated `dassie` image for `3.4.1` to use as the chart default image/tag
- A `dassie-worker` image for `3.4.1` to use as the chart `worker` default image/tag (currently no image exists at all in `ghcr.io/samvera` as far as I can tell)

Upgrade Hyrax Helm chart to 2.0.0

* Create a new release of the Hyrax helm chart at version 2.0 (major version bump)
* Bump all chart dependencies to their latest versions
* Bump and set chart `image.tag` for each dependency, with the idea that #5357 will be done soon and automate keeping these up to date
* Update chart configuration as needed for each dependency upgrade

### DIFF
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 3.1.0
+version: 3.2.0
 appVersion: 3.3.0
 dependencies:
   - name: fcrepo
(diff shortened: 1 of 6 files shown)


mcritchlow:
Current status:
* New dassie 4.0 beta images pushed to ghcr and referenced as chart defaults
* New fcrepo 1.0 chart used as dependency
* Failing to start on a Valkyrie migration added to dassie https://github.com/samvera/hyrax/commit/8fea02691b6078739a43e7aa17c82c1950f6c1a1 with error: 

```sh
│ db-setup PG::InsufficientPrivilege: ERROR:  permission denied to create extension "uuid-ossp"                                              │
│ db-setup HINT:  Must be superuser to create this extension.  
```

dunn (approved):

no-reply (approved):

PR #5769, opened 2022-07-05
don't show analytics in the sidebar if it is turned off

no-reply:
prior to this, we showed the analytics link in the sidebar nav for the dashboard
even if Analytics is turned off on the repository.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/dashboard/sidebar/_activity.html.erb
+++ b/app/views/hyrax/dashboard/sidebar/_activity.html.erb
@@ -33,7 +33,7 @@
     <% end %>
   <% end %>
 
-  <% if current_ability.can_create_any_work? %>
+  <% if current_ability.can_create_any_work? && Hyrax.config.analytics? %>
     <li class="nav-item">
       <%= menu.collapsable_section t('hyrax.admin.sidebar.analytics'),
                                         icon_class: "fa fa-pie-chart",


cjcolvar (commented):
This looks good to me.  Is it worth adding a quick test to
https://github.com/samvera/hyrax/blob/main/spec/views/hyrax/dashboard/_sidebar.html.erb_spec.rb ?

no-reply:
i thought it wasn't worth a test, but I'm open to be corrected

cjcolvar (approved):

jlhardes:
This change feels like it is removing Analytics from user awareness (users won't know analytics is available) but it might not actually be doing that. It makes sense to not show Analytics in the Dashboard if it is not enabled but we can't really show it as a feature flipper (to show it is possible to have) since it requires additional setup behind the scenes. The Analytics documentation we have is currently on the Github wiki (https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#analytics-and-usage-statistics) but that doesn't seem to be a stable enough documentation location to use for pointing users to info about analytics from within the Hyrax app. I don't really want to have something like an in-app ad saying analytics is possible but I also don't want users to think Hyrax just doesn't have analytics capabilities.

All that to say I think this change is fine but we'll need to consider if better or more prominent information should be provided about Hyrax's ability to incorporate analytics (either inside the Hyrax app and/or in documentation we provide).

no-reply:
@jlhardes yes, i think the goal of this work is to make sure users don't see analytics features unless the technical work to make them available has been done.

since users can't themselves enable analytics, what is the function of making them aware of it via a nav button?

Issue #5572, opened 2022-03-23
Works cannot be saved if embargo or lease is assigned (nurax-pg)


eporter23:
### Descriptive summary

Using Hyrax 3.3/Nurax-pg, a user cannot save a new work or save changes to an existing work if an embargo has been assigned. 

### Rationale

Users should be able to assign embargoes if they are enabled for a given Admin Set.

### Expected behavior

If a user can select embargo options, they can then populate them and save their changes to the work.

### Actual behavior

When an embargo is assigned to a work, clicking Save returns the following error: 
```Failed save on # undefined method `[]' for nil:NilClass```

<img width="903" alt="Screen Shot 2022-03-23 at 1 42 11 PM" src="https://user-images.githubusercontent.com/13855987/159762553-19b128a1-61fb-46c7-b1f9-ac2d71274e1b.png">

### Steps to reproduce the behavior

1. Create a new work or edit an existing work in an Admin Set that allows embargoes/leases (e.g. Default Admin Set)
2. Select any permitted embargo or lease option (e.g. Embargo > Restricted to Private until 3/30/2023; then open to Public)
3.  Make sure all Requirements show green checks under Save Work
4. Click Save Changes
5. See error message in banner above the work form 

### Related work

Prior work: #5460

no-reply:
i think this should be closed by https://github.com/samvera/hyrax/pull/5803.

jlhardes:
This is working when testing locally via koppie but I am noticing that the Works list on the dashboard does not show the "Embargo" label for the work under Visibility of Work. Instead I am seeing the current status due to the embargo being applied ("Private" at the moment).

jlhardes:
This issue is fixed on nurax-pg. The Visibility badge problem is called out in #5845 - Manage Embargoes is not listing embargoed works so this issue can be closed.

PR #5771, opened 2022-07-12
WIP - setting embargo for Valkyrie resource

blancoj:
This is to address issue #5572

I need some help with figuring out what code would go here:
lib/hyrax/transactions/steps/set_embargo.rb

What I have there I know does not work, but not quite sure what needs to be done.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/steps/set_embargo.rb
+++ b/lib/hyrax/transactions/steps/set_embargo.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+require 'dry/monads'
+
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # Adds and removes work members
+      class SetEmbargo
+        include Dry::Monads[:result]
+
+        ##
+        # @param [Hyrax::Work] obj
+        # @param [Hash] work_members_attributes
+        #
+        # @return [Dry::Monads::Result]
+        def call(work_resource, embargo_set: nil, end_date: nil, visibility_after_embargo: nil, visibility_during_embargo: nil)
+          return Success(work_resource) unless embargo_set.eql?("embargo")
+
+          # I'm not quite sure what to do here.  I think it may be something like this.
+          embargo = Hyrax::Embargo.new(visibility_during_embargo: visibility_during_embargo,
+                                       visibility_after_embargo: visibility_after_embargo,
+                                       embargo_release_date: end_date)
+
+          resource            = Hyrax::Resource.new(embargo: embargo)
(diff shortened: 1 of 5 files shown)


lib/hyrax/transactions/steps/set_embargo.rb
```diff
@@ -0,0 +1,40 @@
...
+        def call(work_resource, embargo_set: nil, end_date: nil, visibility_after_embargo: nil, visibility_during_embargo: nil)
+          return Success(work_resource) unless embargo_set.eql?("embargo")
+
+          # I'm not quite sure what to do here.  I think it may be something like this.
+          embargo = Hyrax::Embargo.new(visibility_during_embargo: visibility_during_embargo,
+                                       visibility_after_embargo: visibility_after_embargo,
+                                       embargo_release_date: end_date)
+
+          resource            = Hyrax::Resource.new(embargo: embargo)
+          resource.visibility = 'open'
```
no-reply, line 26: shouldn't this be `visibility_during_embargo`?

no-reply: actually, i don't understand what this `resource` is for at all.

lib/hyrax/transactions/steps/set_embargo.rb
```diff
@@ -0,0 +1,40 @@
...
+module Hyrax
+  module Transactions
+    module Steps
+      ##
+      # Adds and removes work members
+      class SetEmbargo
+        include Dry::Monads[:result]
+
+        ##
+        # @param [Hyrax::Work] obj
```
no-reply, line 13: the docs look out of date. i prefer `obj` as the first argument on transaction steps, in general. but i guess i don't object to `work_resource` as long as the docs are updated to match

lib/hyrax/transactions/steps/set_embargo.rb
no-reply: i think i'm catching up to what you have in mind here.

if you're going to do it this way, you need to set `work_resource.embargo = embargo`. creating a new resource in memory isn't doing anything.

note also that these arguments are existing virtual properties on the ResourceForm changeset. maybe this transaction wants to operate against that changeset to build the embargo before `change_set.apply`?

maybe some unit tests would help guide development?

no-reply:
@blancoj what do you think about #5803 as an alternative to this?

orangewolf:
alternatives to this have been adopted. feel free to reopen or create a new issue if we missed something

PR #5778, opened 2022-07-14
comet: decouple Derivatives from Characterization

no-reply:
the tight coupling of Derivative generation to the Characterization background
job makes it hard to customize derivatives processing without changing
characterization behavior.

this takes a hack at decoupling the two by routing derivatives through an event
handler. instead of enqueing the derivatives job directly, the CharacterizeJob
now simply reports that it has characterized the given file_id within the
file_set. it also provides its path hint, in case a listener wants to log or
reuse that hint.

we then add a listener to do the work of actually characterizing. this is a bit
of indirection, but has a lot of benefits in flexibility: you no longer need to
rewrite characterization behavior to customize derivatives handling.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/publisher.rb
+++ b/lib/hyrax/publisher.rb
@@ -119,6 +119,10 @@ module Hyrax
     #     unique id), AND a `:user` (the ::User responsible for the update).
     register_event('collection.membership.updated')
 
+    # @since 3.5.0
+    # @macro a_registered_event
+    register_event('file.characterized')
+
     # @since 3.3.0
     # @macro a_registered_event
     register_event('file.downloaded')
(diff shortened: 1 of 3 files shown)


lib/hyrax/publisher.rb
```diff
@@ -119,6 +119,10 @@ class Publisher
     #     unique id), AND a `:user` (the ::User responsible for the update).
     register_event('collection.membership.updated')
 
+    # @since 3.5.0
```
no-reply, line 122: i plan to backport this quickly to the 3.x stable branch, so this is the appropriate release target

dunn (approved):

PR #5781, opened 2022-07-14
retain the default listeners for future unsubscribing

no-reply:
subscribe the hyrax default listeners in batch. provide a hook for looking up
the `Hyrax.publisher.default_listeners`. this makes it possible to unsubscribe
individual listeners at a later time.

these instances have to live in memory for the life of the Publisher anyway, so
this shouldn't impact when they are freed for garbage collection.

@samvera/hyrax-code-reviewers


lib/hyrax/publisher.rb
```diff
@@ -194,5 +194,21 @@ class Publisher
     # @since 3.2.0
     # @macro a_registered_event
     register_event('object.file.uploaded')
+
+    ##
+    # @return Array[Object] the listeners Hyrax subscribes by default.
+    def default_listeners
```
cjcolvar, line 200: How about adding a simple test since this is a public thing which shouldn't go away without red flags?
Maybe something ensuring it returns an array?

no-reply: tests are in. thanks!

cjcolvar (dismissed):
Thanks!

no-reply:
this one had fallen way behind, but i just rebased it and will follow through on any issues in the tests.

dunn (approved):

PR #5783, opened 2022-07-21
select_options is undefined for an array

ShanaLMoore:
image

When creating a new work, a 500 error is returned: undefined method select_options for array.
This was tested within the latest version of Hyku(4.0 - which upgraded to hyrax 3.4.0).

@admin_set_options => [["Default Admin Set", "admin_set/default"]]
I think we don't need to call #select_options here ^^

because [line 12](https://github.com/samvera/hyrax/blob/v3.4.0/app/helpers/hyrax/work_form_helper.rb#L12) returns similar results if they aren't @admin_set_options is not already present:

Hyrax::AdminSetOptionsPresenter.new(service).select_options => [["Default Admin Set",
"admin_set/default",
{"data-sharing"=>true}]]

### DIFF
--- a/app/helpers/hyrax/work_form_helper.rb
+++ b/app/helpers/hyrax/work_form_helper.rb
@@ -6,7 +6,7 @@ module Hyrax
     #   from views. refactor to avoid
     # @return  [Array<Array<String, String, Hash>] options for the admin set drop down.
     def admin_set_options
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
this issue may be a conflict w allinson flex and should not be contributed back afterall.

PR #5786, opened 2022-07-27
Upgrade to rails 6.1

cjcolvar:
Upgrade hyrax to Rails 6.1.  For the most part only small fixes were needed as supporting samvera core components already have support for Rails 6.1.

TODO: Upgrade dassie to Rails 6.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
         default: 2.3.13
       rails_version:
         type: string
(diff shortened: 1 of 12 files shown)


dlpierce (approved):
I'm not noticing any difficulty running Dassie with Rails 6.1, and its Gemfile doesn't preclude it either. Everything seems to still work under Rails 6.0 too.

PR #5787, opened 2022-07-27
add checked option to radio button

kirkkwang:
## Summary
Through radio buttons, give indicator what visibility the collection is currently set to. Tested change on Hyku.

* add checked option to radio button

### Screenshot
hyrax-radio-button

### DIFF
--- a/app/views/hyrax/dashboard/collections/_form_discovery.html.erb
+++ b/app/views/hyrax/dashboard/collections/_form_discovery.html.erb
@@ -5,19 +5,19 @@
 
   <div class="form-check">
     <label class="form-check-label">
-      <%= f.radio_button :visibility, Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC, checked: @collection.open_access? %>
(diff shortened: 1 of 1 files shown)


cjcolvar (dismissed):
Looks good!

no-reply:
looks like the specs need updating to match the change?

cjcolvar:
FWIW I tried this out locally and I'm not able to reproduce the original issue with hyrax's main branch, but I am able to reproduce it with nurax-dev.

no-reply:
@cjcolvar looking closer, it seems like the form should prepopulate this data?

cjcolvar:
@no-reply Do you think the form is already prepopulating in some cases but not for others?  Could this be a valkyrie/activefedora form difference?

dlpierce:
Thanks Kirk! I ended up solving this in #5812 before I saw this PR. It was also solved by hyrax-head 12 which is why it wasn't showing up on main after the big hyrax 4 merge.

kirkkwang:
Great thank you, @dlpierce !

Issue #5788, opened 2022-07-29
AdminSets can not be created in Hyrax if config.enable_noids = false


laritakr:
### Descriptive summary
AdminSets can not be created in Hyrax if `config.enable_noids = false`

Tested in v3.4.1 but appears to exist in current code as well. 

### Actual behavior
Admin Set creation fails during after save callbacks at [set_owner_attributes in ActiveFedora](https://github.com/samvera/active_fedora/blob/v13.2.7/lib/active_fedora/associations/has_many_association.rb#L30). Result is an admin set which has no permission template. 

The solr query in collection_association.rb returns `-isPartOf_ssim:[* TO *]` because [construct_query expects an id on the unsaved AdminSet.](https://github.com/samvera/active_fedora/blob/main/lib/active_fedora/associations/collection_association.rb#L295).

In Hyrax, the [model transformer assumes you either have or mint an id](https://github.com/samvera/hyrax/blob/main/lib/wings/model_transformer.rb#L112-L116).

### Related work

Link to related tickets or prior related work here.

no-reply:
i think the code you're concerned about in the model transformer was first introduced in 2019, and has been in every release since 2.5 https://github.com/samvera/hyrax/commit/f2c66bbc581e77c46e9c83103d2e7c33f04fba50.

----

my read is that this issue was introduced somewhere around here: https://github.com/samvera/hyrax/commit/f000b8b2dc179747fef066a111b112953f8a5599

orangewolf:
Back port for 3.x and patch for 4.x are both merged. This is resolved.

PR #5789, opened 2022-08-03
Backport to 3.x - AdminSets need an id to save their associations

orangewolf:
Set up a fallback for the case where NOID is false.

Fixes https://github.com/samvera/hyrax/issues/5788

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/admin_set.rb
+++ b/app/models/admin_set.rb
@@ -97,6 +97,14 @@ class AdminSet < ActiveFedora::Base
     permission_template.reset_access_controls_for(collection: self)
   end
 
+  # @api public
+  #
+  # return an id for the AdminSet.
+  # defaults to calling Hyrax::Noid, but needs a fall back if noid is off
+  def assign_id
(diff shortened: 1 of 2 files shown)


orangewolf:
not sure what is making CI mad. This one is the backport, going to try again against main.

dlpierce:
The build is broken because Circle is trying to build the test app with Rails 5.2.8.1 (with the AR store patch) but not blacklight 6.25.1. That's a bit strange since the Circle config implies it is using a specific versions of Rails for the build step.

dlpierce:
Correction, browse-everything 1.2.0 is incompatible with ruby 2.5. The error looked similar to what has been seen in the above case.

no-reply:
should we raise this as an issue with the BE folks? it seems like they should have bumped the major version when dropping ruby 2.5 support?

dlpierce:
Rebased now that 3.x-stable is green again.

dlpierce (approved):

PR #5801, opened 2022-08-17
raise a useful error when setting an unknown visibility

no-reply:
when a user tries to set a visilibity we don't know about, don't respond with
`NoMethodError` on `nil`, but give a useful error message that hints at what has
gone wrong.

@samvera/hyrax-code-reviewers


no-reply:
related to https://github.com/samvera/hyrax/issues/5572

dlpierce (approved):

PR #5804, opened 2022-08-19
Fix Hyrax 3.x build errors caused by linkeddata and json-ld gem updates

dlpierce:
Fix Hyrax 3.x build errors caused by linkeddata and json-ld gem updates

The linked data gem added dependencies that in turn added a dependency on psych. This causes bundle to want to update psych to version 4, which changes the method signature of safe_load. ActiveFedora is not yet compatible with the new format.

Additionally, the json-ld gem added support for using multi_json to load the json writer. The default output is no longer prettified, breaking a hyrax spec. This PR limits the version to prevent breakage and maintain Ruby 2.5 support.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Looks good.

PR #5807, opened 2022-08-23
provide default PermissionBadge presenter behavior when `nil`

no-reply:
if visibility wasn't indexed for any item, the `PermissionBadge` presenter will
cause any relevant page not to render.

this fixes that issue by providing a reasonable default display badge in the
case that we can't identify the visibility from the indexed data.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Looks good to me!

no-reply:
thanks! cherry picked that one here too.

dunn (approved):

PR #5808, opened 2022-08-23
soften dependendy on search_state_class for app generation

no-reply:
search_state_class isn't defined on CatalogController until after blacklight
generator has run, but `Hyrax::Engine` needs to execute before that in
generation.

cc @cjcolvar

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/generators/hyrax/install_generator.rb
+++ b/lib/generators/hyrax/install_generator.rb
@@ -71,10 +71,6 @@ module Hyrax
       end
     end
 
-    def catalog_controller
-      copy_file "catalog_controller.rb", "app/controllers/catalog_controller.rb"
-    end
-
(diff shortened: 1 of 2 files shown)


cjcolvar:
I missed that @dlpierce had already opened a PR (#5802) to cherry-pick this commit into main, but that PR didn't pass the test app generation.  Looks like this one failed in the same spot.  I think https://github.com/samvera/hyrax/pull/5786/commits/eead121318b899837fdbc9dc6d1227d05e3439f9 should resolve the issue.

no-reply:
thanks! cherry picked that one here too

cjcolvar (approved):

dlpierce:
Thanks, both of you!

PR #5812, opened 2022-08-24
[3.x] Delegate collection visibility to model

dlpierce:
Fixes the collection form object not containing any visibility data. 

This caused the collection edit form to not have the current visibility checked.

This fix is not needed with hydra-head 12 & hyrax 4 thanks to this commit:
https://github.com/samvera/hydra-head/commit/57405cce591e3a6525907825c835093ffd43e29f


cjcolvar:
Will this close https://github.com/samvera/hyrax/pull/5787?

cjcolvar (approved):
Looks good to me.

dlpierce:
Oops, I completely missed that other PR. Still, I think this is a better way to solve the problem.

kirkkwang:
Awesome thanks all!

PR #5814, opened 2022-08-24
helperize Embargo edit view dependencies on ActiveFedora behavior

no-reply:
this view depended directly on ActiveFedora and `Hydra::AccessControls` specific
behavior. helperize those dependencies so the view can be extended to support
Valkyrie.

@samvera/hyrax-code-reviewers


dlpierce (commented):
Should a modified `views/hyrax/leases/edit.html.erb` file be included too?

no-reply:
> Should a modified `views/hyrax/leases/edit.html.erb` file be included too?

i think that's in https://github.com/samvera/hyrax/pull/5810 ?

dlpierce (approved):

PR #5822, opened 2022-08-30
check json, not string equality for JSON-LD output

no-reply:
we care about the equality of the JSON structure, not the specific string output. the latter depends on the serialization code, which can change independently of our codebase

@samvera/hyrax-code-reviewers


dlpierce (approved):
Okay, I'm convinced. 😁

PR #5830, opened 2022-08-31
Create .koppie (.dassie without Fedora)

dlpierce:
The purpose of Koppie is to provide a way for developers to work on valkyrie support in hyrax, without the presence of Fedora, in the same way .dassie is currently used.

It was created by copying the current nurax-pg application into the .koppie directory and then removing the parts unneeded for running only in the docker-compose environment, such as capistrano deployment. It has its own compose file `docker-compose-koppie.yml` which can be used like so:
`docker-compose -f docker-compose-koppie.yml up`

One additional fix was made to the recently added seed data system to allow the admin role to be assigned.

As noted in the README, koppies are isolated granite outcrops in Southern Africa that are a favored habitat for hyraxes.

@samvera/hyrax-code-reviewers


app/utils/hyrax/test_data_seeders/user_seeder.rb
```diff
@@ -25,14 +25,14 @@ def generate_seeds(logger: Logger.new(STDOUT), allow_seeding_in_production: fals
         private
 
         def admin_role
-          unless defined? Role
+          unless ::User.reflect_on_association(:roles)
```
dlpierce, line 28: `Role` seemed to not be loaded while running `rake db:seed`, so this instead checks if the user model has a roles association which `add_user` also depends on existing.

jeremyf (approved):

PR #5833, opened 2022-09-03
Run `gem update --system` in hyrax-base image & set ACL on gem location

dlpierce:
Updates rubygems and bundler in the hyrax-base image, with the ability to specify a version in the event of bundler bugs.

Also adds the acl package and uses it to set default permissions for the bundle directory. This is needed because prior to this change the bundle directory is empty and will use the non-root user permissions when installing gems in a later layer. Now root creates files during the system update, causing permission errors when a regular user attempts to install additional gems.

* Update both rubygems and bundler gems in hyrax-base image
* Install a specific version of those gems (3.3.20/2.3.20 respectively) due to a bundler bug preventing dependency resolution on musl systems like Apline.
* Install the Alpine acl package and use it to set less restrictive default permissions on the location bundler installs gems too.

* `docker-compose build` completes successfully.

@samvera/hyrax-code-reviewers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,11 @@ FROM ruby:$RUBY_VERSION-alpine3.15 as hyrax-base
 
 ARG DATABASE_APK_PACKAGE="postgresql-dev"
 ARG EXTRA_APK_PACKAGES="git"
+ARG RUBYGEMS_VERSION=3.3.20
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Worked for me! :tada:

Issue #5596, opened 2022-04-06
Add a subcollection to collection fails with NoMethodError in Hyrax::Dashboard::NestCollectionsController#create_relationship_under


jlhardes:
### Descriptive summary

On nurax-pg, with 2 nestable collections of the same nestable collection type, adding a subcollection to another collection fails with "NoMethodError in Hyrax::Dashboard::NestCollectionsController#create_relationship_under"

<img width="1355" alt="Screen Shot 2022-04-06 at 3 56 16 PM" src="https://user-images.githubusercontent.com/3440166/162060057-7977fb78-70bf-42da-81c2-07ab1bd55241.png">

### Steps to reproduce the behavior

* From Dashboard -> Collections create two collections of type "nesting only"
* From Dashboard -> Collections, click the title of one of the collections
* Click Add a subcollection button
* Verify the Add a subcollection button is disabled until a selection is made
* Select the other collection from the list of potential children
* Click Add a subcollection (this is where error is encountered as shown in screenshot)

jlhardes:
Verified this is working locally on koppie and this is working on nurax-pg. Closing this issue!

PR #5835, opened 2022-09-06
Factory method for finding collection type from collection

cjcolvar:
Hyrax::PcdmCollection does not have a `collection_type` getter method so switching to a factory allows for valkyrie exclusive tests to pass.

In support of #5596.

### DIFF
--- a/app/actors/hyrax/actors/collections_membership_actor.rb
+++ b/app/actors/hyrax/actors/collections_membership_actor.rb
@@ -108,7 +108,7 @@ module Hyrax
 
         # Do not apply permissions to work if collection type is configured not to
         collection = Hyrax.config.collection_class.find(collection_id)
-        return unless collection.share_applies_to_new_works?
+        return unless Hyrax::CollectionType.for(collection: collection).share_applies_to_new_works?
 
         # Save the collection id in env for use in apply_permission_template_actor
         env.attributes[:collection_id] = collection_id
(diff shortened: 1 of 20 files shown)


app/models/hyrax/pcdm_collection.rb
```diff
@@ -70,6 +70,15 @@ def permission_manager
       @permission_manager ||= Hyrax::PermissionManager.new(resource: self)
     end
 
+    # Get the collection_type when accessed
+    def collection_type
```
no-reply, line 74: we have generally tried to avoid putting new behavior on models.

which tests does this impact? is there a more valkyrie style way of passing them?

cjcolvar: This is the test that was calling it explicitly. 
https://github.com/samvera/hyrax/blob/917b4568bdedda15bfa3c65d1c456111156764f7/spec/controllers/hyrax/dashboard/nest_collections_controller_spec.rb#L138-L144

There are a few places in the code base where a collection's collection type is accessed through it so I thought it might make sense to move this convenience method here, but I could go apply a new pattern to all places that expect it to exist.
Example:
https://github.com/samvera/hyrax/blob/917b4568bdedda15bfa3c65d1c456111156764f7/app/controllers/concerns/hyrax/collections_controller_behavior.rb#L81-L82

cjcolvar: I went with `Hyrax::CollectionType.for(collection: collection)` because I felt it was readable and the best option.

no-reply (approved):

Issue #5595, opened 2022-04-06
Add this Collection to another collection Fails with NoMethodError in Hyrax::Dashboard::NestCollectionsController#create_relationship_within


jlhardes:
### Descriptive summary

On nurax-pg, with 2 nestable collections of the same nestable collection type, adding a collection as a subcollection to another collection fails with "NoMethodError in Hyrax::Dashboard::NestCollectionsController#create_relationship_within"

<img width="1356" alt="Screen Shot 2022-04-06 at 3 48 51 PM" src="https://user-images.githubusercontent.com/3440166/162058918-6f4f001a-3b4a-47ca-a1a0-2cbee6bf5d32.png">

### Steps to reproduce the behavior

* From Dashboard -> Collections create two collections of type "nesting only"
* From Dashboard -> Collections, click the title of one of the collections
* Click Add to collection button (beside Edit and Delete buttons)
* Verify the Add to collection button is disabled until a selection is made
* Select the other collection from the list of potential parents
* Click Add to collection (this is where error is encountered as shown in screenshot)

gamontoya:
@cjcolvar I'm seeing the same reported behavior.

![5595](https://user-images.githubusercontent.com/4314914/189990250-ec645475-7b25-4833-b16d-6960f61ba9e5.png)

jlhardes:
Verified this is working locally on koppie and this is also working on nurax-pg. Closing this issue!

Issue #5492, opened 2022-03-03
Create new sub-collection leads to NoMethodError in Hyrax::Dashboard::NestCollectionsController#create_collection_under


MPLSFedResearchTZ:
### Descriptive summary

Clicking "Create new collection as subcollection" gives rise to error page of "NoMethodError in Hyrax::Dashboard::NestCollectionsController#create_collection_under".

### Steps to reproduce the behavior

1. Go to collection editing page.
2. Clicking "Create new collection as subcollection"
3. Error page appears

jlhardes:
This error is still occurring on nurax-pg:

<img width="1216" alt="Screen Shot 2022-04-06 at 5 11 07 PM" src="https://user-images.githubusercontent.com/3440166/162071649-139d1d31-3fa4-479b-9ee6-361e93f27bc7.png">

### Steps to reproduce the behavior

* From Dashboard -> Collections create a collection of type "nesting only"
* From Dashboard -> Collections, click the title of the collection
* Click Create new collection as subcollection (this is where the error occurs that is in the screenshot)

no-reply:
@jlhardes i'm confused if this is an issue with `nurax-pg` / valkyrie support, or if it's a base hyrax issue.

i don't think any of the `NestCollectionsController` supports valkyrie. AFAIK, this is the only thing not passing the full test suite when using a valkyrie collection model.

this was coming up to the top of my ToDo list, but i'm concerned about taking that work on if the controller is broken against the ActiveFedora branch.

jlhardes:
This was from testing nurax-pg so it is valkyrie-specific. I just tested again and the error is still occurring on nurax-pg. Does that help?

gamontoya:
@cjcolvar @blancoj I'm still seeing the error (testing on nurax-pg)

![5492](https://user-images.githubusercontent.com/4314914/189990993-3afd0f77-9472-4504-a0a9-3ccdc465e81e.png)

cjcolvar:
The fix hasn't been merged and deployed to nurax-pg yet.  I can tag you for retesting once it is deployed.

eporter23:
I was able to successfully create a sub collection within a User Collection type that I had previously created, using the original steps in the ticket description. I am seeing another error when trying to open the All Collections tab within the Dashboard however.  
[...]

PR #5850, opened 2022-09-09
Use solr's graph query builder for nested collection queries

cjcolvar:
This can replace the expensive nested indexing using samvera-nesting_indexer and all of the special handling for it.

Note: the graph query builder does not work with multi-sharded solr instances

All nesting indexing has been removed along with unused methods/parameters.  Deprecation notices will be added to a future backport PR to 3.x.


spec/services/hyrax/collections/nested_collection_query_service_spec.rb
```diff
@@ -122,7 +81,7 @@
           it 'returns an array of valid collections of the same collection type' do
             expect(scope).to receive(:can?).with(:deposit, coll_c).and_return(true)
             expect(described_class).to receive(:query_solr).with(collection: coll_c, access: :read, scope: scope, limit_to_id: nil, nest_direction: :as_child).and_call_original
-            expect(subject.map(&:id)).to contain_exactly(another.id, coll_e.id)
+            expect(subject.map(&:id)).to contain_exactly(another.id)
```
cjcolvar, line 84: I think this test was incorrect before.  If A -> B -> C -> D -> E then I wouldn't expect E to be valid as an "available child" for C.

spec/services/hyrax/collections/nested_collection_query_service_spec.rb
```diff
@@ -217,7 +176,7 @@
           it 'returns an array of collections of the same collection type excluding the given collection' do
             expect(scope).to receive(:can?).with(:read, coll_c).and_return(true)
             expect(described_class).to receive(:query_solr).with(collection: coll_c, access: :deposit, scope: scope, limit_to_id: nil, nest_direction: :as_parent).and_call_original
-            expect(subject.map(&:id)).to contain_exactly(coll_a.id, another.id)
+            expect(subject.map(&:id)).to contain_exactly(another.id)
```
cjcolvar, line 179: I think this test was incorrect before. If A -> B -> C -> D -> E then I wouldn't expect A to be valid as an "available parent" for C because it already is. I believe this is actually prevented by the validator if you attempt to do this in the UI.

spec/services/hyrax/collections/nested_collection_query_service_spec.rb
```diff
@@ -309,8 +268,9 @@
       let(:parent) { coll_e }
       let(:child) { another }
 
-      it 'returns false' do
-        expect(subject).to eq false
+      # Limit no longer applies so it will always return true
+      it 'returns true' do
+        expect(subject).to eq true
```
cjcolvar, line 273: We could use solr's `maxDepth` parameter to the graph query parser, but that feels like more work than it is worth.

orangewolf: that makes sense to me. 

If you exceed it there should just be a message that says "you have delved too deep. the Balrog comes"

spec/services/hyrax/collections/nested_collection_query_service_spec.rb
orangewolf: I believe in the current implementation, C can have A as an ancestor through B and as a direct parent. So

A -> B -> C -> D -> E
A -> C -> D -> E

at the same time. BUT I do think that works more by accident than by design.

cjcolvar:
This should also fix #5596, #5595, and #5492.

dlpierce:
This sounds great! Has there been any performance testing with meaningful/large hyrax data sets? Would perhaps using the `traversalFilter` graph query option provide any benefit?

no-reply:
this looks great to me.

since the performance of the old implementation is extremely bad, it seems so likely that the performance will be improved. i'd like to suggest we move forward with this and leave any testing for post merge.

i'd also like to look at a limited backport of this, keeping the old indexing behavior in place for 3.x by default, but introducing the new query behavior. i'll poke at this later today.

cjcolvar:
@dlpierce I've done some very limited manual testing and haven't seen any issues with solr.  I'm not sure if the traversalFilter is useful for our use case or not.  I'd be curious about performance as nesting depth scales but I'd suspect solr would handle most normal cases easily.

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
```diff
@@ -83,12 +82,11 @@ def available_parent_collections
         # rerouting to new_dashboard_collection_path to add the new collection as
         # a child. Since we don't yet have a child collection, the valid? option can't be used here.
         def validate_add
-          if nestable?(parent)
-            nesting_within_maximum_depth
-          else
+          unless nestable?(parent)
```
jeremyf, line 85: How are we guarding circular nesting? (e.g. A has member B has member C has member A)?

spec/services/hyrax/collections/nested_collection_query_service_spec.rb
jeremyf: Allowing E to be a direct child of C and indirect child of C was intentional but perhaps from a theoretical stand point instead of anything concrete.

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
dlpierce: According to the documentation, the graph query parser is "cyclic aware".

no-reply: on a current Hyrax, we allow circular nesting.

no-reply:
What's the right way to document that Hyrax 4.0 doesn't support sharded SolrCloud? it seems like we need to write that down somewhere, even if it's unlikely to impact any current users.

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
```diff
@@ -104,17 +102,6 @@ def remove
 
         attr_accessor :query_service, :persistence_service, :context, :collection
 
-        # ideally we would love to be able to eliminate collections which exceed the
-        # maximum nesting depth from the lists of available collections, but the queries
-        # needed to make the determination are too expensive to do for every possible
-        # collection, so we only test for this situation prior to saving the new
-        # relationship.
-        def nesting_within_maximum_depth
```
no-reply, line 112: i'm trying to write a commit message for the backport of this change---which no-ops this validation if the solr graph query is configured. but i'm a bit stuck and realizing i'm not entirely clear on why we're removing this. can you help me @cjcolvar?

orangewolf:
> What's the right way to document that Hyrax 4.0 doesn't support sharded SolrCloud? it seems like we need to write that down somewhere, even if it's unlikely to impact any current users.

I would think a note on this page https://samvera.github.io/service-stack.html (which could stand some expansion and more prominant linking in the readme, but that's a different ticket) and a comment in app/search_builders/hyrax/dashboard/nested_collections_search_builder.rb when we talk about the graph search. (because that's where things will break if you multishard).  

In a perfect world, we would throw an exception or send a notice or something on startup if multiple shards are detected, but I don't think we even mandate solr cloud usage, so I'm not sure how that API call would be reliable and again, I feel like "Check if Solr configuration is sane on start up" is a whole new ticket.

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
cjcolvar: I assumed that the max depth was only needed to limit the amount of nested reindexing to undertake.  Now that the indexing isn't necessary the technical limit seems like solr's graph query parser's efficiency.  If there is a functional reason for having a limit (and making it adopter configurable) then I could restore the depth limit enforcement with a new implementation.  The configuration of the max depth was also on the Samvera::NestingIndexer so we'd have to make a new config in Hyrax for it.

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
cjcolvar: I think it is being handled by the `parent_and_child_can_be_nested` validation on the form.

spec/services/hyrax/collections/nested_collection_query_service_spec.rb
cjcolvar: FWIW I tested again locally and this works for User Collections, but when I attempt to do this with a collection type that doesn't allow multiple membership, I get an error:
```
Hyrax::SingleMembershipError in Hyrax::Dashboard::NestCollectionsController#create_relationship_within
Error: You have specified more than one of the same single-membership collection type (type: Nesting only, collections: C and D)
```

cjcolvar: I'll look into restoring this functionality.

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
cjcolvar: @no-reply Can you give an example of currently allowed/working circular nesting?

spec/services/hyrax/collections/nested_collection_query_service_spec.rb
cjcolvar: I added support for this back in with https://github.com/samvera/hyrax/pull/5850/commits/dfc2f0963dbb45d77998954b761a15103969c384

spec/services/hyrax/collections/nested_collection_query_service_spec.rb
cjcolvar: Test was restored and support readded with https://github.com/samvera/hyrax/pull/5850/commits/dfc2f0963dbb45d77998954b761a15103969c384

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
no-reply: thanks! super helpful

app/forms/hyrax/forms/dashboard/nest_collection_form.rb
no-reply: circular-nesting

i was able to create this situation trivially in the UI.

cjcolvar: I wasn't able to reproduce this in nurax-dev.  Can you please post the exact steps that allowed this for you?  I'd like to reproduce this in nurax-dev then test the same steps with this PR to see how the behavior differs.

dlpierce (approved):
My testing of the style of solr query used here worked well and seemed to be speedy with lots of collections. :+1:

cjcolvar (commented):
-

PR #5857, opened 2022-09-12
Proxy deposit requests use valkyrie-specific query service when use_valkyrie? is true

### LINKED ISSUE #5456: Transferring ownership of work displays "NoMethodError in Hyrax::TransfersController#new" error message
### Descriptive summary

On nurax-pg, when I attempt to transfer ownership of one of my works, I see a "NoMethodError in Hyrax::TransfersController#new" message

### Expected behavior

Clicking "transfer ownership of work" should open Transfer Ownership page

### Actual behavior

Clicking "transfer ownership of work" generates "NoMethodError in Hyrax::TransfersController#new" error message

1. Make sure that transferring ownership is enabled on nurax-pg
2. Navigate to Dashboard -> Works
3. In the Actions menu next to a work, select "Transfer Ownership of Work"
4. Observe error message
hyraxtransferscontroller

cjcolvar:
I tried to replicate the behavior of WorkRelation where it should only retrieve objects of registered curation concerns (in `Hyrax::WorkResourceQueryService#work`), but I'm not sure if it is really necessary so I didn't write any tests for it yet.

I believe this should fix #5456

### DIFF
--- a/app/services/hyrax/work_resource_query_service.rb
+++ b/app/services/hyrax/work_resource_query_service.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+module Hyrax
+  # Responsible for retrieving information based on the given work.
+  #
+  # @see ProxyDepositRequest
+  # @see SolrDocument
+  # @see Hyrax::SolrService
+  # @note This was extracted from the ProxyDepositRequest, which was coordinating lots of effort. It was also an ActiveRecord object that required lots of Fedora/Solr interactions.
+  class WorkResourceQueryService
+    # @param [String] id - The id of the work
+    def initialize(id:)
+      @id = id
+    end
+    attr_reader :id
+
+    # @return [Boolean] if the work has been deleted
+    def deleted_work?
+      Hyrax.query_service.find_by(id: id)
+      false
+    rescue Valkyrie::Persistence::ObjectNotFoundError
+      true
+    end
+
+    def work
+      # Need to ensure it is a work?
+      resource = Hyrax.query_service.find_by(id: id)
+      unless Hyrax.config.curation_concerns.include?(resource.class) ||
+             Hyrax.config.curation_concerns.map(&:to_s).include?(resource.class.to_s) || # Wings-wrapped models
+             Hyrax.config.curation_concerns.map(&:to_s).include?(resource.class.name) # Wings-wrapped models
+        raise ModelMismatchError, "Expected allowed work type but got #{resource.class}"
+      end
+      resource
+    end
+
+    def to_s
+      if deleted_work?
+        'work not found'
+      else
+        solr_doc.to_s
+      end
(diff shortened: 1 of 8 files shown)


app/services/hyrax/work_resource_query_service.rb
```diff
@@ -0,0 +1,48 @@
...
+      false
+    rescue Valkyrie::Persistence::ObjectNotFoundError
+      true
+    end
+
+    def work
+      # Need to ensure it is a work?
+      resource = Hyrax.query_service.find_by(id: id)
+      unless Hyrax.config.curation_concerns.include?(resource.class) ||
+             (resource.class.respond_to?(:_canonical_valkyrie_model) && Hyrax.config.curation_concerns.include?(resource.class._canonical_valkyrie_model)) # Wings-wrapped models
```
no-reply, line 28: could `#name` be used instead? the ORMConverter base passes this through to `#_canonical_valkyrie_model` when appropriate. i think it would be something like ` unless Hyrax.config.curation_concerns.include?(resource.class.name)`

cjcolvar: I had to change this to `#to_s` due to some failing tests where the Wings wrapping class returned `Hyrax::Work` for `#name` but `#to_s` returned `GenericWork`.

cjcolvar: Ugh, more failing tests so might have to check both `#name` and `#to_s`.

app/services/hyrax/work_resource_query_service.rb
```diff
@@ -0,0 +1,49 @@
...
+    def deleted_work?
+      Hyrax.query_service.find_by(id: id)
+      false
+    rescue Valkyrie::Persistence::ObjectNotFoundError
+      true
+    end
+
+    def work
+      # Need to ensure it is a work?
+      resource = Hyrax.query_service.find_by(id: id)
```
no-reply, line 26: i'm still a little peeved (as i'm sure you must be too), to be doing all these checks. it feels like there should be one way to check if something is a registered "curation concern".

as an alternative thought, do we really need to check the registered concerns? i wonder if `resource.work?` wouldn't be good enough here.

cjcolvar: I was thinking the same.  Checking `resource.work?` feels sufficient to me.  It broadens the number of matching resources but I don't think the check to ensure it is a registered curation concern and not merely a work is really that important.

Looking ahead at other uses of `WorkRelation.new.find` I wonder if replacing them all with a `query_service.find_by` and check for `#work?` is acceptable.

cjcolvar: I went ahead and made the change to only test `#work?`.

no-reply (approved):

Issue #5797, opened 2022-08-12
Editing batch of works generates "NoMethodError in Hyrax::BatchEditsController#edit" error


jlhardes:
### Descriptive summary

On nurax-pg, editing a batch of works displays a "NoMethodError in Hyrax::BatchEditsController#edit" error

### Expected behavior

Selecting multiple works on the Works page and clicking "Edit Selected" should bring me to an edit page

### Actual behavior

Selecting multiple works on the Works page and clicking "Edit Selected" generates an error message:

![Screen Shot 2022-08-12 at 10 54 55 AM](https://user-images.githubusercontent.com/3440166/184382360-ba0b6495-3aa0-46b7-b2da-2c1b87b62278.png)

### Steps to reproduce the behavior

1. On nurax-pg, navigate to Dashboard -> Works
2. Select one or more works on the Works page
3. Click the "Edit Selected" button
4. Observe error message

### Related work

#5461 Issue with initial report of batch work edit problem but error message was different

jlhardes:
#4913 BatchEditsController needs to support valkyrie objects - code changes identified here might be part of the work that needs to happen to address this issue

jlhardes:
Still seeing error message when selecting multiple works and trying to "Edit Selected" when testing locally on koppie.

gamontoya:
@rjkati This now works on `Hyrax Version: 4.0.0.rc3`, but it fails with `Koppie v4.0.0.rc3`

**Hyrax:**

![HyraxEdit](https://github.com/samvera/hyrax/assets/4314914/0fa8268f-77b1-4f2a-af43-fbf8898c6b0d)

**Koppie:**

![batchEdit](https://github.com/samvera/hyrax/assets/4314914/0bc6612e-19a8-435e-9fab-c3cdddd70414)

alishaevn:
I've created 2 GenericWork's on the koppie app and in the Hyrax app. trying to batch edit on Koppie showed me that the form thought it was editing at least one Monograph work. I don't have any Monograph works in my local apps. I also noticed that when I accessed the rails console for koppie, it couldn't return a count for GenericWorks, but it could for Hyrax.

``` bash
# KOPPIE
[3] pry(main)> GenericWork.all.count
NoMethodError: undefined method `all' for GenericWork:Class
from (pry):3:in `__pry__'

# HYRAX
[2] pry(main)> GenericWork.all.count
=> 2
```

[...]

PR #5863, opened 2022-09-14
Resource permission persisting

cjcolvar:
Introduces handling of `permissions_attributes` params that come from the controller in the `SaveAccessControl` transaction step.  The step converts the params into `Hyrax::Permission` objects and adds them to the ACL of the object being saved.

Also in the PR is change to the works controller to pass the step arg to make use of this new functionality.

In support of #5797 

Steps for testing:
1. Go to dashboard
2. Go to works
3. Click Add new work button
4. Fill in required metadata
5. Go to Sharing tab
6. Choose group 'admin' with 'Edit' access and click Add
7. Click deposit agreement
8. Click Save button
9. Click Edit button
10. Go to Sharing tab
11. Look for group 'admin' in 'Currently Shared With' table

### DIFF
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -189,7 +189,8 @@ module Hyrax
           'work_resource.add_to_parent' => { parent_id: params[:parent_id], user: current_user },
           'work_resource.add_file_sets' => { uploaded_files: uploaded_files, file_set_params: params[hash_key_for_curation_concern][:file_set] },
           'change_set.set_user_as_depositor' => { user: current_user },
-          'work_resource.change_depositor' => { user: ::User.find_by_user_key(form.on_behalf_of) }
+          'work_resource.change_depositor' => { user: ::User.find_by_user_key(form.on_behalf_of) },
+          'work_resource.save_acl' => { permissions_params: form.input_params["permissions"] }
         )
         .call(form)
       @curation_concern = result.value_or { return after_create_error(transaction_err_msg(result)) }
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -189,7 +189,8 @@ def create_valkyrie_work
           'work_resource.add_to_parent' => { parent_id: params[:parent_id], user: current_user },
           'work_resource.add_file_sets' => { uploaded_files: uploaded_files, file_set_params: params[hash_key_for_curation_concern][:file_set] },
           'change_set.set_user_as_depositor' => { user: current_user },
-          'work_resource.change_depositor' => { user: ::User.find_by_user_key(form.on_behalf_of) }
+          'work_resource.change_depositor' => { user: ::User.find_by_user_key(form.on_behalf_of) },
+          'work_resource.save_acl' => { permissions_params: form.input_params["permissions"] }
```
cjcolvar, line 193: `form.input_params` is passed instead of the raw params because `reform` transforms the raw params into an array as part of the validation.  We could pass the raw params instead but they will be harder to work with and any other code calling the transaction outside of a controller context would have to format the params into hashes with index keys instead of the more natural arrays.

dlpierce:
This is failing features/edit_work_resource_spec.rb because the permissions_attributes parameter is not including the agent.

cjcolvar:
@dlpierce Thanks.  I made some changes to guard for missing params (probably due to unused fields on the page still being posted).  This should be ready for review as long as CI goes green.

dlpierce:
This fails when trying to edit a work and add a permission in koppie. It looks like Hyrax::PermissionController needs to change to Hyrax::Resource in its load_resource call. Strangely, when I setup the feature test to add a permission and save it, it does not fail when run in the koppie container.

dlpierce (approved):

PR #5867, opened 2022-09-15
Exposing `Hyrax.config.derivative_services`

jeremyf:
Prior to this commit, we had a somewhat opaque mechanism for overriding the candidate Hyrax::DerivativeService services.  There is precedence on doing this (see [NewspaperWorks][1]) yet the configuration was not exposed during application generation nor as part of `Hyrax::Configuration`.

This change exposes the setting of the derivative services a primary concern of configuring a Hyrax application.

Related to #5866

[1]:https://github.com/samvera-labs/newspaper_works/blob/d0c0c0595ad318178896fb7f2abd93c5139fe5b7/lib/newspaper_works/engine.rb#L15-L35

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/derivative_service.rb
+++ b/app/services/hyrax/derivative_service.rb
@@ -1,8 +1,24 @@
 # frozen_string_literal: true
 class Hyrax::DerivativeService
-  class_attribute :services
-  self.services = [Hyrax::FileSetDerivativesService]
-  def self.for(file_set)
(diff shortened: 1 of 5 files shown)


dlpierce:
@jeremyf Can you rebase this on main? There is a commit you're missing that limits our redis version and should fix the failing build.

dlpierce (approved):

PR #5868, opened 2022-09-15
Koppie fixes

dlpierce:
* Fixes koppie's storage paths so that the app and sidekiq containers can see the same files.
* Sets koppie to use cp when persisting files. If a ValkyrieIngestJob fails then the upload files will still exist when the job is retried.
* Runs `yarn` during startup to ensure UniversalViewer is installed.


fritzfreiheit (approved):
Loaded and ran branch locally, and the new koppie image works.

Issue #4854, opened 2021-04-05
Publish Hyrax container image targets to Docker Hub in CircleCI


mcritchlow:
### Descriptive summary

The current Hyrax Dockerfile provides various build targets.

Rather than rely on manual build/publish work to be done by contributors, when
they remember to, we should integrate this into the CI configuration.

### Rationale

Downstream users may start increasingly relying on the Hyrax container image,
and its various targets, for developing and deploying Hyrax applications.
Therefore we should keep the published image versions in sync with the codebase.

### Expected behavior

The following Dockerfile `target`'s should be built and published to Docker Hub
under the `samveralabs` repository. 

- `hyrax-base`
- `hyrax-worker`
- `hyrax-engine-dev`
- `hyrax-engine-dev-worker`
- `dassie`

Following a successful CI/CD pipeline in which there have been code changes,
each of the above targets should be built and published with the following tags:

- `latest` - always published
- `sha256 hash` - always published
- `hyrax version` (e.g. `hyrax-base:3.0.1` upon `Hyrax::VERSION` updates

### Related work

Some existing documentation for reference: https://github.com/samvera/hyrax/blob/master/CONTAINERS.md#docker-image-for-hyrax-based-applications

no-reply:
cc: @orangewolf

mcritchlow:
It's been quite a while since I've pushed anything to Docker Hub, but I wonder
if we might also consider using it's build tooling as a potential alternative to
within CircleCI.

https://docs.docker.com/docker-hub/builds/

I'm not sure what is preferable for Hyrax, just thought I'd put it out there for
conversation.

no-reply:
one issue with docker hub is that they've changed all their rules, and now `samveralabs` has "6 of 3" seats filled. i think we would have to pay to make it accessible to new users.

should we consider hosting images somewhere else?

no-reply:
we talked briefly in the local context about publishing at GitHub. last i checked you need GitHub credentials to download public images from GitHub registries, which is pretty high friction for developer setup, as well as in deployment contexts where secrets delivery isn't necessarily trivial.

mcritchlow:
I'm not sure what's possible here: https://www.docker.com/community/open-source/application

But it sounds like there may be an improved Docker Hub experience possible
depending on the outcome of that application?

Perhaps a related question, is "free" a requirement in an alternative image
registry?

PR #5870, opened 2022-09-16
build amd64 and arm64 images on release and push to ghcr.io

orangewolf:
Refs #4854 

Build ARM and Intel Docker images automatically based on the current recommendation from CONTAINERS.md.

Using Github Actions, we can build the docker images and push them to the Github Registry. This currently only builds hyrax-base, hyrax-base-worker and hyrax-engine-dev. There are more items described in #4854, but only those are listed in the current CONTAINERS.md. I'm not sure which list is right. It also currently does nothing about Docker Hub. I'm not super clear if we should or not at this point.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,67 @@
+name: Build and push docker image
+on:
+  workflow_dispatch:
+  release:
+  push:
+    branches:
+      - build_images_on_release
+
+env:
+  REGISTRY: ghcr.io
+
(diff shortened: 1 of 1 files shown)


dlpierce:
Running dassie in docker compose uses the hyrax-engine-dev-worker target for the sidekiq service, so that's something to consider adding here.

mcritchlow (dismissed):
This is great!

PR #5871, opened 2022-09-16
Hyrax-5844 - Cannot remove embargo from work - 2

### LINKED ISSUE #5844: Cannot remove embargo from work
### Descriptive summary

When testing locally using koppie and after a work has an embargo assigned, trying to deactivate the embargo results in an error message.

### Expected behavior

A Deactivation successful message should appear and you should be able to verify that the embargo has been removed from the Work and the Work can be accessed.

### Actual behavior

The following error message is shown when trying to deactivate an embargo on a work: Faraday::ConnectionFailed in Hyrax::PermissionsController#confirm

Screen Shot 2022-09-08 at 11 38 35 AM

1. Navigate to a Work which contains an embargo
2. Click the "Edit" button 
3. In the "Save Work" box on the right side, click the "Embargo Management Page" link in the Visibility section
4. Click the "Deactivate Embargo" button
5. Error message appears (see screenshot)

### Related work

#5572 Works cannot be saved if embargo or lease is assigned (nurax-pg) - in progress and available to test locally only but allows works to be saved with an embargo

fritzfreiheit:
Fixes #5844 ; refs #5572

Unembargoing a work under valkyrie throws an error

When deactivating an embargo on a work under koppie image/valkyrie, a TCP error is thrown by Hyrax::PermissionsController because it attempts to load ActiveFedora resource.

Screen Shot 2022-09-08 at 11 38 35 AM

Proposed change(s):
* Update Hyrax::PermissionsController to 

To test:
* In koppie/nurax-pg, navigate to a Work which contains an embargo
* Click the "Edit" button 
* In the "Save Work" box on the right side, click the "Embargo Management Page" link in the Visibility section
* Click the "Deactivate Embargo" button
* Error message appears (see screenshot above)

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/embargo_actor.rb
+++ b/app/actors/hyrax/actors/embargo_actor.rb
@@ -15,8 +15,10 @@ module Hyrax
         case work
         when Valkyrie::Resource
           embargo_manager = Hyrax::EmbargoManager.new(resource: work)
-          embargo_manager.release && Hyrax::AccessControlList(work).save
-          embargo_manager.nullify
+          return if embargo_manager.embargo.embargo_release_date.blank?
+          embargo_manager.nullify(force: true)
+          embargo_manager.release(force: true)
+          Hyrax::AccessControlList(work).save
         else
           work.embargo_visibility! # If the embargo has lapsed, update the current visibility.
           work.deactivate_embargo!
--- a/app/services/hyrax/embargo_manager.rb
+++ b/app/services/hyrax/embargo_manager.rb
@@ -164,19 +164,23 @@ module Hyrax
     ##
     # Drop the embargo by setting its release date to `nil`.
     #
+    # @param force [boolean] force the nullify even when the embargo period is current
+    #
     # @return [void]
-    def nullify
-      return unless under_embargo?
+    def nullify(force: false)
+      return false if !force && under_embargo?
       embargo.embargo_release_date = nil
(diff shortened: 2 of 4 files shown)


no-reply:
@fritzfreiheit i'm confused about the description of this one. it looks like the changes here aren't related to `PermissionsController`. could you say a bit about what these changes are addressing?

marrus-sh:
@no-reply If I am understanding correctly, the current “remove embargo” code calls out to `EmbargoManager`. However, `EmbargoManager` is currently designed for use in automated tasks, and essentially performs a no‐op when trying to release an embargo while the period is still current. This PR adds a `force` keyword to the `EmbargoManager` methods to allow for releasing embargos manually during the embargo period, instead of just in an automated fashion at the end.

I imagine a similar set of changes would be needed for leases if we decide to go this route.

alishaevn (approved):
I tested this locally and I was able to remove an embargo. it's not reflected in the "deactivated embargoes" tab, but that work can happen in a different pr so this doesn't get held up even further.

alishaevn (changes requested):
although this code passes on koppie, it errs on hyrax.

image

alishaevn (approved):
re-approving because this pr does not introduce the inability to get to the embargo edit screen for a monograph. it happens on `main` too.

PR #5880, opened 2022-09-23
Favoring `Hyrax.logger` over `Rails.logger`

### LINKED ISSUE #5879: Consolidate `Hyrax.logger` and `Rails.logger`
Throughout the code-base there are calls to `Rails.logger` and `Hyrax.logger`.  As of 2022-09-23, `Hyrax.logger` is an alias for `Valkyrie.logger`.

To help with consistent logging, Hyrax should pick a direction for logging and use it.

My proposal is that we replace all instances of `Rails.logger` with `Hyrax.logger` and then in `Hyrax.logger`'s definition, we choose `Rails.logger` if `Rails` is defined; and fallback to `Valkyrie.logger`.

jeremyf:
With this change, Hyrax will consistently use the same logger: first defaulting to `Rails.logger` and falling back to `Valkyrie.logger`.

This is a potentially breaking change as we're currently writing to two different loggers (which may be the same based on downstream configuration).

I have also exposed a means for the upstream application to configure the `Hyrax.logger`.

Closes #5879

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/create_with_files_actor.rb
+++ b/app/actors/hyrax/actors/create_with_files_actor.rb
@@ -40,7 +40,7 @@ module Hyrax
         expected_user_id = env.user.id
         files.each do |file|
           if file.user_id != expected_user_id
-            Rails.logger.error "User #{env.user.user_key} attempted to ingest uploaded_file #{file.id}, but it belongs to a different user"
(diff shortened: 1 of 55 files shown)


cjcolvar (approved):
This looks good to me.  Since `main` is pre-4.0 I think it makes sense to merge this now since it introduces a potential change in logging behavior.

dlpierce:
I would like to see a Rubocop cop setup to enforce this.

Issue #5876, opened 2022-09-21
Version Presenter Valykrization


marrus-sh:
I’ve been looking into editing FileSets through the Web UI in a Valkyrized Hyrax and from what I can tell the `VersionListPresenter` and `VersionPresenter` are current sticking points in making this work. I think fixing the following things would get us a fair chunk of the way there :—

- [x] `FileSet`s need an `original_file` method. The `ValkyrieFileSetIndexer` provides an implementation of `original_file` here :—

    <https://github.com/samvera/hyrax/blob/d5be603d9a01737145ffa2a8b9c6d8a9a1ca8641/app/indexers/hyrax/valkyrie_file_set_indexer.rb#L102-L106>

    —: but it doesn’t appear to be present on `FileSet`s themselves. Perhaps this method can just be moved there?

- [x] The resulting `FileMetadata` needs a `versions` method, and doesn’t have one. The Wings storage adapter provides a `find_versions` method so creating this should be achievable.

- [x] The `Version` object provided by the Wings storage adapter is missing a `created` method, although “in practice” it seems like this might be the `version_token`?

- [ ] Finally, `Version`s created by Wings **should** have an associated `VersionCommitter`. `VersionCommitter`s are created by the `VersioningService`, but this service does nothing for Valkyrie resources. The pieces to implement this for Wings should be present in the storage adapter; it just needs to be put into place.

marrus-sh:
Going to close this as I believe the final checkbox is probably a separate issue; the presenters themselves should work with Valkyrie now :)

PR #5886, opened 2022-10-03
Update versioning service to support Valkyrie

marrus-sh:
- Allow instances of `VersioningService` to be created and used to access the versions of a resource
- Update the Wings `Version` struct to support some additional properties needed by the versioning presenter (namely `created`)
- Initial set of tests for Valkyrie versioning behaviour

See #5876.

Things I’m curious about:

- Is this appropriation of `VersioningService` acceptable?
- What additional tests are necessary here?

### DIFF
--- a/app/presenters/hyrax/version_list_presenter.rb
+++ b/app/presenters/hyrax/version_list_presenter.rb
@@ -24,7 +24,7 @@ module Hyrax
                       else
                         Hyrax::FileSetFileService.new(file_set: file_set).original_file
                       end
-      new(original_file ? Hyrax::VersioningService.new(resource: original_file).versions : [])
+      new(Hyrax::VersioningService.new(resource: original_file).versions)
     rescue NoMethodError
       raise ArgumentError
     end
--- a/app/presenters/hyrax/version_list_presenter.rb
+++ b/app/presenters/hyrax/version_list_presenter.rb
@@ -24,7 +24,7 @@ module Hyrax
                       else
                         Hyrax::FileSetFileService.new(file_set: file_set).original_file
                       end
-      new(original_file&.versions&.all.to_a)
+      new(original_file ? Hyrax::VersioningService.new(resource: original_file).versions : [])
     rescue NoMethodError
       raise ArgumentError
     end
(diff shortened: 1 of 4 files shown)


app/services/hyrax/versioning_service.rb
```diff
@@ -1,7 +1,72 @@
...
+
+    ##
+    # Returns an array of versions for the file associated with this
+    # Hyrax::VersioningService.
+    #
+    # If the file is a Hyrax::FileMetadata and versioning is not supported in
+    # the storage adapter, an empty array will be returned.
+    def versions
+      if resource.is_a? Hyrax::FileMetadata
+        if storage_adapter.respond_to?(:"supports?") && storage_adapter.supports?(:versions)
```
no-reply, line 36: is this the same as `storage_adapter.try(:"supports?")`

no-reply:
i think the extension of `VersioningService` here is great---it doesn't seem to introduce a ton of complexity.

i'm pretty happy with the tests as-is. they seem appropriately scoped to the interface on VersioningService. the only thing that comes to mind is that a "many" case might be wanted (in addition to "one" and "two", or in place of "two"), but i wouldn't block on it, especially given how the implementation depends on `#versions_list`.

👍🏻

no-reply (dismissed):

app/services/hyrax/versioning_service.rb
marrus-sh: i believe so; i can do that instead

app/presenters/hyrax/version_list_presenter.rb
```diff
@@ -24,7 +24,7 @@ def self.for(file_set:)
                       else
                         Hyrax::FileSetFileService.new(file_set: file_set).original_file
                       end
-      new(original_file&.versions&.all.to_a)
+      new(original_file ? Hyrax::VersioningService.new(resource: original_file).versions : [])
```
no-reply, line 27: would be nice if the `VersioningService` handled the nil case. `Hyrax::VersioningService.new(resource: nil)).versions #=> []`, then you could squash out this conditional.

no-reply (approved):

app/presenters/hyrax/version_list_presenter.rb
marrus-sh: yeah i was debating which would be better; i can go that route

no-reply (approved):

no-reply (approved):

dlpierce (approved):

PR #5890, opened 2022-10-04
Add to_presenter method to Hyrax::FileSet

marrus-sh:
For parity with `Hyrax::FileSetBehaviour`, per 3a220bf88b2d4d54cf93e7a61000493a3e6aae1a.

(This is needed by the FileSet presenters, but is extending the model like this still the best approach?)

### DIFF
--- a/app/models/hyrax/file_set.rb
+++ b/app/models/hyrax/file_set.rb
@@ -70,5 +70,9 @@ module Hyrax
     def file_set?
       true
     end
+
+    def to_presenter
+      Hyrax::SearchService.new(config: CatalogController.blacklight_config).fetch(id).last
+    end
   end
 end


no-reply (approved):
`#to_presenter` creates an odd circular dependency that this could be an opportunity to move away from. is there only one caller? `/app/views/hyrax/file_sets/edit.html.erb:9`? if so i might consider changing the caller. 

for me, the ideal pattern is that the controller provides the presenter. if other views/partials require parts of model behavior, the presenter can delegate them. that might not be practical here, and i'm okay with merging this as-is (approving now for this reason). but if you have time to do the analysis and get a more principled patch in, that would be best in the longer term.

dlpierce:
I think `#to_presenter` (which is returning a SolrDocument) can be replaced with `@presenter.solr_document` and also adding `before_action :presenter` to `Hyrax::FileSetsController` to populate `@presenter`. Thoughts?

no-reply:
thanks for the alternative @dlpierce! this worked out great!

Issue #5575, opened 2022-03-24
Work view page - Items Actions menu not working : ArgumentError in Hyrax::FileSetsController#edit


MPLSFedResearchTZ:
### Descriptive summary

The actions menu on work view page is not working. Click Edit or Versions in drop-down menu gives rise to FileSetsController related errors.

### Steps to reproduce the behavior

1. Log in to user dashboard, then list works, click a work name to display work view page.
2. In the section of "Items", find the dropdown menu "Select an action" and select either Edit or Version from the drop-down menu and click.
3. You will see an error like ArgumentError in Hyrax::FileSetsController#edit

### Related work

#5136 (epic with history of previous work and testing of Works)
#5794 Issue for Download option in Items Actions menu
#5795 Issue for Delete option in Items Actions menu

fritzfreiheit:
app/presenters/hyrax/version_list_presenter.rb

Updated:
```
    ##
    # @param [Object] an object representing the File Set
    #
    # @return [Enumerable<Hyrax::VersionPresenter>] an enumerable of presenters
    #   for the relevant file versions.
    #
    # @raise [ArgumentError] if we can't build an enu
    def self.for(file_set:)
      return new(file_set.original_file&.versions&.all.to_a) if file_set.respond_to? :original_file
      original_file = Hyrax.query_service.find_by(id: Valkyrie::ID.new(file_set.original_file_id))
      new(original_file&.versions&.all.to_a)
    rescue Valkyrie::Persistence::ObjectNotFoundError
      new([])
    rescue NoMethodError
      raise ArgumentError
    end
```

app/models/hyrax/file_set.rb

Added:
```
    # Cast to a SolrDocument by querying from Solr
    def to_presenter
      Blacklight::SearchService.new(config: CatalogController.blacklight_config).fetch(id).last
    end
```

app/views/hyrax/file_sets/_permission_form.html.erb

Modified to get around issue with missing .permissions method (but I assume it shouldn't be missing, which is where I left off)

Modified:
```
<% public_perm = true if params[:controller] == 'batch' %>
<% public_perm ||= f.object.permissions.map { |perm| perm.access if perm.agent_name == "public"}.compact.first if f.object.respond_to? :permissions %>
<% public_perm ||= false %>
```

marrus-sh:
[...]

PR #5894, opened 2022-10-10
Use the ResourceForm for FileSets

marrus-sh:
This PR makes the following changes (in the interest of resolving the remaining issues with #5575) :—

- It refactors work‐specific form behaviours into a new class, `Hyrax::Forms::PcdmObjectForm`, to enable `Hyrax::Forms::ResourceForm` to be used with resource types other than PCDM objects @ 5c48b30e130bb62d00040de1141b92456d037cc4. *[Note №1]*
- It moves FileSet metadata into a new YAML‐based configuration, `file_set_metadata`, and updates the model and form to both use the same metadata (previously the model used `basic_metadata`, and the form had hardcoded fields) @ f70045238739c4e918ee02df57309a944641d1e1. *[Note №2]*
- It modifies `Hyrax::Forms::FileSetForm` to inherit from `Hyrax::Forms::ResourceForm`, with the intention of inheriting the lease and embargo handling behaviours provided there @ f63ca9cdad50d443d952012a5834f4d1a6b09ba8.
- It uses the form, instead of the model object directly, in the FileSet edit view, to allow embargo information to be populated in a Valkyrie context @ d0b785eb925bf5520a9024780d9018d1589677cb.

**Notes:**

1. `Hyrax::Forms::ResourceForm(…)` now returns a class subclassing `Hyrax::Forms::PcdmObjectForm`, not `Hyrax::Forms::ResourceForm` directly. This is a bit awkward and confusing, but it preserves backwards‐compatibility with existing Hyrax applications which defined their models using the recommended `class MyObjectForm < Hyrax::Forms::ResourceForm(MyObjectForm)` syntax.

    FileSet forms (and other future forms which want to use the ResourceForm embargo/lease implementations but do not have PCDM object properties) consequently cannot use the function‐call syntax and should just subclass `Hyrax::Forms::ResourceForm` directly.

    See commit 5c48b30e130bb62d00040de1141b92456d037cc4.

2. `Hyrax::FileSet` formerly included `Hyrax::Schema(:basic_metadata)` in the model, but had a more limited set of metadata actually exposed through the `Hyrax::FileSetForm`. The two furthermore did not agree on `license`, in that it is optional in `basic_metadata` but was required in `Hyrax::FileSetForm`.

    I’ve resolved this by creating a new metadata YAML, `file_set_metadata`, with just the attributes listed in the form or required by tests, and `license` set to `required`. But another solution might be to keep the full `basic_metadata` schema and allow `license` to be optional. See https://github.com/samvera/hyrax/blob/bd2bcffc33e183904be2c175367648815f25bc2b/lib/hyrax/specs/shared_specs/hydra_works.rb#L252-L260 which implies general `basic_metadata` support, but not all attributes are tested and compare also 792ccfdb0c4cddb542d8915a7317b9d83eeba5f0 which removed that requirement for collections considering that those attributes were never exposed in forms.

    (The attributes required by `it_behaves_like 'a model with basic metadata'` but not previously exposed on the form are: `abstract`, `label`, `relative_path`, `resource_type`, `rights_notes`, `rights_statement`, and `source`. If we’re comfortable revising this test, we could probably drop those from `file_set_metadata.yaml`.)

   > By “form” I mean in a Reform/`Hyrax::ChangeSet` sense; the actual Edit page only has a field for title. That leaves open the question of which, if any, of these attributes actually need to be defined on the FileSet model?

    In the long term, I would like for the FileSet metadata to be configurable, but it’s hardcoded to whatever is in the YAML file for now.

    See commit f70045238739c4e918ee02df57309a944641d1e1.

### DIFF
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -195,7 +195,14 @@ module Hyrax
     def initialize_edit_form
       guard_for_workflow_restriction_on!(parent: parent)
 
-      @version_list = Hyrax::VersionListPresenter.for(file_set: @file_set)
+      case file_set
+      when Hyrax::Resource
+        @form = Hyrax::Forms::ResourceForm.for(file_set)
+        @form.prepopulate!
+      else
+        @form = form_class.new(file_set)
+      end
+      @version_list = Hyrax::VersionListPresenter.for(file_set: file_set)
       @groups = current_user.groups
     end
 
--- a/app/forms/hyrax/forms/file_set_form.rb
+++ b/app/forms/hyrax/forms/file_set_form.rb
@@ -7,6 +7,11 @@ module Hyrax
     class FileSetForm < Hyrax::ChangeSet
       include Hyrax::FormFields(:core_metadata)
 
+      # The fields in +:file_set_metadata+ were hardcoded into this form in a
+      # previous version of Hyrax, but ideally in the future this metadata will
+      # be configurable.
+      include Hyrax::FormFields(:file_set_metadata)
+
       class << self
         ##
         # @return [Array<Symbol>] list of required field names as symbols
@@ -16,20 +21,6 @@ module Hyrax
             .keys.map(&:to_sym)
         end
       end
-
-      property :creator, required: true
-      property :license, required: true
-
-      property :based_near
-      property :contributor
-      property :date_created
-      property :description
-      property :identifier
-      property :keyword
-      property :language
-      property :publisher
-      property :related_url
-      property :subject
 
       property :permissions, virtual: true
       property :visibility, default: VisibilityIntention::PRIVATE
--- a/app/forms/hyrax/forms/file_set_form.rb
+++ b/app/forms/hyrax/forms/file_set_form.rb
@@ -3,8 +3,8 @@
 module Hyrax
   module Forms
     ##
-    # @api public
-    class FileSetForm < Hyrax::ChangeSet
+    # A form for +Hyrax::FileSet+s.
+    class FileSetForm < Hyrax::Forms::ResourceForm
       include Hyrax::FormFields(:core_metadata)
 
       # The fields in +:file_set_metadata+ were hardcoded into this form in a
@@ -12,26 +12,8 @@ module Hyrax
       # be configurable.
       include Hyrax::FormFields(:file_set_metadata)
 
-      class << self
-        ##
-        # @return [Array<Symbol>] list of required field names as symbols
-        def required_fields
-          definitions
-            .select { |_, definition| definition[:required] }
-            .keys.map(&:to_sym)
-        end
-      end
-
-      property :permissions, virtual: true
-      property :visibility, default: VisibilityIntention::PRIVATE
-
-      # virtual properties for embargo/lease;
-      property :embargo_release_date, virtual: true
-      property :visibility_after_embargo, virtual: true
-      property :visibility_during_embargo, virtual: true
-      property :lease_expiration_date, virtual: true
-      property :visibility_after_lease, virtual: true
-      property :visibility_during_lease, virtual: true
+      property :representative_id, type: Valkyrie::Types::String
+      property :thumbnail_id, type: Valkyrie::Types::String
     end
   end
 end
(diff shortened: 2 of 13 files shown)


app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -194,6 +194,13 @@ def add_breadcrumb_for_action
     def initialize_edit_form
       guard_for_workflow_restriction_on!(parent: parent)
 
+      case file_set
+      when Hyrax::Resource
+        @form = Hyrax::Forms::ResourceForm.for(file_set)
+        @form.prepopulate!
```
marrus-sh, line 200: Prepopulation is necessary for `:in_works_ids`

app/forms/hyrax/forms/file_set_form.rb
```diff
@@ -4,43 +4,8 @@ module Hyrax
...
-
-      # virtual properties for embargo/lease;
-      property :embargo_release_date, virtual: true
-      property :visibility_after_embargo, virtual: true
-      property :visibility_during_embargo, virtual: true
-      property :lease_expiration_date, virtual: true
-      property :visibility_after_lease, virtual: true
-      property :visibility_during_lease, virtual: true
+    class FileSetForm < Hyrax::Forms::ResourceForm
+      include Hyrax::FormFields(:basic_metadata)
```
marrus-sh, line 8: `:basic_metadata` is currently baked into `Hyrax::FileSet`, although ideally I would like this to be configurable. How are we handling it for works?

app/models/hyrax/file_set.rb
```diff
@@ -45,6 +45,13 @@ def self.model_name(name_class: Hyrax::Name)
     attribute :original_file_id, Valkyrie::Types::ID.optional # id for FileMetadata resource
     attribute :extracted_text_id, Valkyrie::Types::ID.optional # id for FileMetadata resource
 
+    attribute :admin_set_id,             Valkyrie::Types::ID
+    attribute :member_ids,               Valkyrie::Types::Array.of(Valkyrie::Types::ID).meta(ordered: true)
+    attribute :member_of_collection_ids, Valkyrie::Types::Set.of(Valkyrie::Types::ID)
+    attribute :on_behalf_of,             Valkyrie::Types::String
+    attribute :proxy_depositor,          Valkyrie::Types::String
+    attribute :rendering_ids,            Valkyrie::Types::Array.of(Valkyrie::Types::ID).meta(ordered: true)
+
```
marrus-sh, line 54: All of these attributes are currently required by `Hyrax::ResourceForm`, but they aren’t present in `Hyrax::FileSet` currently. Should they be moved off of `Hyrax::ResourceForm` somehow, or do `Hyrax::FileSet`s need more work to support these?

app/views/hyrax/file_sets/_permission.html.erb
```diff
@@ -1,5 +1,5 @@
   <div id="permissions_display" class="tab-pane">
-    <%= simple_form_for [main_app, file_set],
+    <%= simple_form_for [main_app, @form],
```
marrus-sh, line 2: Making this simple substitution “works” in the sense that it gets the page to render, but it doesn’t actually “work” in the sense that if you click save (in Koppie) it throws an error :P . I’m not sure if just passing `@form` here is a naive approach, or if it is correct and something else in Koppie is broken.

app/controllers/hyrax/file_sets_controller.rb
marrus-sh: (These can probably be written on the same line)

jeremyf:
Can you talk a bit more about the desire for configurable FileSet metadata?  I ask because this may drift the file_set towards a conceptual work.

Could these configurable attributes be better modeled in a work?

no-reply:
@jeremyf i'm not sure that i see that making FileSet metadata configurable changes FileSet's status. couldn't this configuration help us move *away* from the work-like description in the BasicMetadata we use now?

marrus-sh:
@jeremyf as far as this MR is concerned, @no-reply is right in that the changes here move the FileSet *away* from a conceptual work, in that the model used to use the same metadata as works (`:basic_metadata`) and now uses different metadata (which will hopefully be more limited).

as far as my personal desire for configurable metadata is concerned, this is of course open to discussion :) . mostly, though, i think it is a gut feeling deriving from the large number of attributes already present on the FileSet model—if FileSet only included `:core_metadata`, i wouldn’t necessarily be asking for additional configurability, but considering it currently contains a number of additional properties beyond just `:core_metadata`, i feel like those properties should be configurable.

none of that configurability is in this MR, though.

no-reply:
@jeremyf do you have any thoughts about a path to move this one through?

jeremyf (approved):

PR #5907, opened 2022-11-01
[Helm] separate the worker and web selectors 

orangewolf:
Currently the selector labels for both web deployment and worker deployment are the same. This leads to the service seeing both web and worker pods. It should just see the web pods. Also it causes some UIs to be confused about how many pods there are (looking at you Rancher UI). This isolates the two deployments.

I'd like to see a chart release pushed once this is merged. I don't see any charts in the registry since 1.0.0 and we're on 1.6.0 in the code base. I can do the push if that's ok with reviewer.

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/templates/_helpers.tpl
+++ b/chart/hyrax/templates/_helpers.tpl
@@ -52,6 +52,14 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
 
 {{/*
+Worker Selector labels
+*/}}
+{{- define "hyrax.workerSelectorLabels" -}}
+app.kubernetes.io/name: {{ include "hyrax.name" . }}-worker
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
 Create the name of the service account to use
 */}}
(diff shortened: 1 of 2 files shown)


mcritchlow:
This seems like a great idea.

Do you want to add a chart version bump to this PR? Happy to approve/merge this.

fwiw, the charts have been published: https://github.com/orgs/samvera/packages/container/package/charts%2Fhyrax

orangewolf:
I was 100% looking in the wrong place! I'll bump the version up

orangewolf:
@mcritchlow I bumped this to 2.0.0.  I'm curious what your opinion is on it.  We rolled it out to an existing cluster today (I'd only used it on a new cluster before) and found that you have to remove the worker deployment before deploying this version. It respawns no problemo, but it does require that manual step. This is because the field we're changing is not mutable.

mcritchlow:
@orangewolf - Yeah that makes sense to me. Not to make more work (we could always start doing this later), but i wonder if we should have a section in the chart readme on Upgrading, similar to the bitnami charts https://github.com/bitnami/charts/tree/main/bitnami/nginx#upgrading

Either way, approving this :+1:

PR #5914, opened 2022-11-17
Backport #5894 to 3.x

marrus-sh:
Add the changes in #5894 to the 3.x-stable branch

### DIFF
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -192,7 +192,14 @@ module Hyrax
     def initialize_edit_form
       guard_for_workflow_restriction_on!(parent: parent)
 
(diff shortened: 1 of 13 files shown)


marrus-sh:
The issue here is that in the form, `f.object.model.visibility` is correctly set to `"authenticated"`, but `f.object.visibility` has an uninitialized value of `""`. This isn’t true on `main`, where the former reflects the latter; I’m still not sure what is causing the discrepancy.

marrus-sh:
The change in behaviour appears to be a result of dd41326d5ff9747475c45f92a9cb0c5a63095311 which is fun.

marrus-sh:
It looks like https://github.com/samvera/hydra-head/pull/514 is the missing functionality. As a workaround, I’ve tried just manually initializing `@form[:visibility] = file_set.visibility` in the controller, and this seems to work.

no-reply (approved):

PR #5915, opened 2022-11-21
reinstate support for ActiveFedora to WorkflowActionsController

no-reply:
workflows still support both AF and Valkyrie, and workflows triggered from the Actor Stack initializer step still passes through AF objects, but this controller had be cut over completely to Valkryie. this impacts users with custom workflow steps. reinstating ActiveFedora support allows applications to choose when to migrate their existing custom workflows.

h/t to benpennell in slack.

@samvera/hyrax-code-reviewers


bbpennel (dismissed):
This change is working for me locally via monkey patch

cjcolvar (approved):

PR #5920, opened 2022-11-29
Fixes the missing sort call in add_sorting_to_solr (#5905).

### LINKED ISSUE #5905: Cannot sort the works under a collection
### Descriptive summary

Found in Hyrax 3.4.1 and the main branch as shown in durax-pg and durax-dev. Cannot get the works sorted by selecting any items in the "Sort by" dropdown menu:

image

### Rationale

The solr query might use a wrong sorting field or the sort is not allowed by Bliacklight (the new version of Blacklight will only pass along to solr sort parameters that have been defined in the CatalogController but this would only be a problem if this sort actually goes through Blacklight).

1. Select a sort item
2. The works are displayed in a sorted way

### Related work

bwatson78:
Fixes #5905  ; refs #5905 

Because our override method never calls `Blacklight::SearchBuilder#sort`, it will always default to `title_si asc`, making the sort options useless on this show page.

* Adds the missing sort command to the method that overrides one from Blacklight::Solr::SearchBuilderBehavior.
* Inserts a test to confirm that a sort option passed into the blacklight_params actually gets utilized.

* Login and visit Dashboard
* Choose Collections and select one with multiple works to navigate to the show page
* Choose any sort option from the pull down menu
* After page load, the order of the works should be different

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/collection_search_builder.rb
+++ b/app/search_builders/hyrax/collection_search_builder.rb
@@ -37,7 +37,7 @@ module Hyrax
     # This overrides the default 'relevance' sort.
     def add_sorting_to_solr(solr_parameters)
       return if solr_parameters[:q]
-      solr_parameters[:sort] ||= "#{sort_field} asc"
+      solr_parameters[:sort] = sort || "#{sort_field} asc"
     end
 
     # If :deposit access is requested, check to see which collections the user has
--- a/app/search_builders/hyrax/collection_search_builder.rb
+++ b/app/search_builders/hyrax/collection_search_builder.rb
@@ -37,7 +37,8 @@ module Hyrax
     # This overrides the default 'relevance' sort.
     def add_sorting_to_solr(solr_parameters)
       return if solr_parameters[:q]
-      solr_parameters[:sort] = sort || "#{sort_field} asc"
+      solr_parameters[:sort] ||= sort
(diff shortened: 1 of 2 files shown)


app/search_builders/hyrax/collection_search_builder.rb
```diff
@@ -37,7 +37,7 @@ def with_access(access)
     # This overrides the default 'relevance' sort.
     def add_sorting_to_solr(solr_parameters)
       return if solr_parameters[:q]
-      solr_parameters[:sort] ||= "#{sort_field} asc"
+      solr_parameters[:sort] = sort || "#{sort_field} asc"
```
cjcolvar, line 40: I'm wondering if this change might break code that depended on `solr_parameters[:sort]` already being set and thus not being overridden by `add_sorting_to_solr`.  That is probably an edge case but maybe it would be safer to do this instead?
```suggestion
      solr_parameters[:sort] ||= sort
      solr_parameters[:sort] ||= "#{sort_field} asc"
```

cjcolvar (commented):
Nice catch!

app/search_builders/hyrax/collection_search_builder.rb
bwatson78: @cjcolvar I committed those changes.

dlpierce (approved):

PR #5941, opened 2023-01-17
Expanding method interface

jeremyf:
Prior to this commit, we had inconsistent method interfaces for the two IIIF presenters.  Further, we also had the case where the `hostname` and `ability` were not "public", even though we had public writer methods.

With this commit, we normalize that.


cjcolvar (dismissed):
Looks good.

dlpierce (approved):

Issue #5951, opened 2023-01-30
Dashboard works page error for works without depositors


bbpennel:
### Descriptive summary

In our hyrax 3.5 repository, we experienced errors when visiting the admin dashboard > Works page if the current page contained any works that did not have a depositor set. We appear to have around 70 objects like this in our production environments (They were likely created by a deposit script). I don't know if Hyrax considers this to be invalid, if so we can probably disregard this issue.

The error is `ActionView::Template::Error (key not found: "depositor_ssim"` stemming from this line:
https://github.com/samvera/hyrax/blob/hyrax-v3.5.0/app/views/hyrax/my/_work_action_menu.html.erb#L38

Which is caused by:
https://github.com/samvera/hyrax/blob/hyrax-v3.5.0/app/models/concerns/hyrax/ability.rb#L417

### Rationale

Having the works page fail to load due to not signing depositors seems unhelpful.

### Expected behavior
The works page should not throw an error, but should render as usual, likely with the user not having the `transfer` ability since they wouldn't be the depositor.

### Actual behavior

Works page gives an uncaught error page

### Steps to reproduce the behavior

Create a work without a depositor set (either via a script or by creating the work normally, removing the depositor field in fedora, and reindexing), then go to the dashboard > works page and find that work.

PR #5952, opened 2023-01-30
Don't throw an error if depositor field isn't present

bbpennel:
Fixes https://github.com/samvera/hyrax/issues/5951

* `user_is_depositor?` will return false if there is no depositor field in solr rather than throwing a "Key not found" error.

@samvera/hyrax-code-reviewers


dlpierce (approved):
Looks great! Thank you.

PR #5959, opened 2023-02-10
Fix missing work id in file-set-in-work-download events

### LINKED ISSUE #5956: file-set-in-work-download events with (not set) label/id
### Descriptive summary

"file-set-in-work-download" events with "(not set)" as the label have been showing up in our google analytics since deploying hyrax 3.5 a little over a week ago. About 90% of our "file-set-in-work-download" events have "(not set)" as the label, while the other ~10% have work ids as the labels.

After some testing, I am finding that
1. The download link at the top of the work and file set pages ("Download the file" when no preview is available) is triggering the "(not set)" labeled "file-set-in-work-download" events.
2. The download link in the file set list ("Items" section of the work page) is recording the "file-set-in-work-download" with the work id as the label.
3. The download link in the universal viewer is not recording any events.

I was able to reproduce these results on nurax by monitoring the requests sent to GA when clicking the various download links. I used this work: https://nurax-dev.curationexperts.com/concern/generic_works/9w0323285?locale=en

It seems like the problematic download links need to have a `data-work-id` property added to them.

### Rationale

Since "file-set-in-work-download" is used to populate the download stats in the Work Analytics page, a lot of download events are effectively being lost since they have no work id.

### Expected behavior

Download links at the top of the work and fileset pages pass the work-id as the label for "file-set-in-work-download" events.

Universal Viewer download option also triggers the same download events (this isn't as important though)

### Actual behavior

See above.

Visit:
https://nurax-dev.curationexperts.com/concern/generic_works/9w0323285?locale=en
Open your browsers developer tools, open the network tab, and then click on the various download links. Two events should show up. The problematic behavior sends events with values like:
```
5(file-set*file-set-download*ns0646239)
5(file-set-in-work*file-set-in-work-download)
```

And the correct behavior sends events like:
```
5(file-set*file-set-download*ns0646239)
5(file-set-in-work*file-set-in-work-download*9w0323285)
```

bbpennel:
Add all the same data to media display download links as is present on the fileset list download link, to resolve file-set-in-work-download events not having work ids

Fixes #5956

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/file_sets/media_display/_audio.html.erb
+++ b/app/views/hyrax/file_sets/media_display/_audio.html.erb
@@ -8,7 +8,7 @@
       </audio>
       <%= link_to t('hyrax.file_set.show.downloadable_content.audio_link'),
                   hyrax.download_path(file_set),
-                  data: { label: file_set.id },
+                  data: { label: file_set.id, work_id: @presenter.id, collection_ids: @presenter.member_of_collection_ids },
                   target: :_blank,
                   id: "file_download" %>
     </div>
--- a/app/views/hyrax/file_sets/media_display/_default.html.erb
+++ b/app/views/hyrax/file_sets/media_display/_default.html.erb
@@ -4,7 +4,7 @@
     <p /><%= link_to t('hyrax.file_set.show.download'),
       hyrax.download_path(file_set),
       id: "file_download",
-      data: { label: file_set.id },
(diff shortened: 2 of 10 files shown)


dlpierce (approved):

dlpierce:
Looks good other than the specs seeming to need a presenter setup?

bbpennel:
@dlpierce the test issues should be resolved now

dlpierce (approved):

PR #5962, opened 2023-02-14
Test on Rails 7

cjcolvar:
~I had to remove samvera-nesting_indexer and power_converter as well as point to a branch of hydra-pcdm to get bundler to install rails 7.~
With ruby 3.2 support already merged to main, I rebased this PR to focus on rails 7.


cazzerson:
I wonder if we could separate the ruby and rails upgrades since the ruby 2.7 EOL is imminent.

PR #5976, opened 2023-02-28
show all works in Review Submissions page.

### LINKED ISSUE #5933: Review submissions page does not show all works
### Descriptive summary

I am noticing in both my hyrax 3 based repo and in 4.x dassie that the Review Submissions page is not returning all of my "Under Review" works. For testing, I have 10 under review works, 1 approved, which shows 9 under review and 1 published. Additional under review works after this point don't appear in the Review Submissions page at all. In our test environment, there are several hundred under review works, but only 2 are showing.

It looks like the queries being sent to solr by `Hyrax::SolrQueryService` and `Hyrax::SolrService` don't specify a `rows` parameter, so it is only returning the first 10 results total across all workflow states. Changing the number of entries to return in the dashboard UI from 10 to 25 doesn't have any affect. I am guessing we would want SolrQueryService to return a really high number of results here since it appears to be querying for every work ID in a workflow state.

The queries in question appear to start here:
https://github.com/samvera/hyrax/blob/main/app/controllers/hyrax/admin/workflows_controller.rb#L19-L20
Which call this service:
https://github.com/samvera/hyrax/blob/main/app/services/hyrax/workflow/actionable_objects.rb#L37
Which builds a query here that doesn't include a `rows` parameter:
https://github.com/samvera/hyrax/blob/main/app/services/hyrax/solr_query_service.rb#L31

### Expected behavior

"Review Submissions" page should be able to page through all under review and published works.

### Actual behavior

"Review Submissions" page only shows a total of 10 works, distributed between the two categories, with no paging available.

1. Create an admin set using the provided mediated deposit workflow
2. Deposit 11 works into the mediate deposit workflow admin set
3. Only 10 of them will show, with no extra pages
4. Navigate to the first work you deposited and approve it
5. Return to the "Review submissions" page, it will most likely now show 9 under review works and 1 published (or it might still only show 10 under review, I think it may depend on the order deposited).

### Related work

Relates to the same query causing problems here:
https://github.com/samvera/hyrax/issues/5926
But is a separate issue.

blancoj:
Fixes #5933

With this change all works that are in pending review status will be shown on the Review Submissions Page.  Before this change, if you had more than 10 works in pending review status, only 10 of the works would show up.  The reason for this is that Solr returns a max of 10 solr documents when passed an array of work ids, so I changed the code to pass it one work id at a time.

* Create an admin set using the provided mediated deposit workflow
* Deposit 11 works into the mediate deposit workflow admin set
* Go to the "Review submissions" page, and make sure you see all 11 works that are under review.

@samvera/hyrax-code-reviewers


dlpierce:
I agree with Ben here. In applications with lots of records to be displayed here, making a few thousand individual solr requests is not going to be performant. An overhaul to add paging is what's needed.

blancoj:
I agree with both of you, but have been testing by trying to load batch works, just to see. So far I have loaded 155 works in pending review, and the performance seems fine when going to the Review Submissions page.  I want to try to load 500 works and check the performance then.

PR #5981, opened 2023-03-09
Updates to CircleCI config

dlpierce:
Fixes broken build caused by mismatch between gems installed for the hyrax engine versus the generated .internal_test_app. This is solved by running `bundle install` for the internal_test_app using the ruby orb to allow the target directory to be set.

Also contains these other changes:
* Remove outdated nurax deployment workflow
* Run `yarn install` with caching for both the engine and test app
* Be more specific with rubocop caching
* Updated default ruby/bundler/rails versions
* Drop ruby patch version from workflow and job names (will need to update required checks)
* Use rspec-test command provided by ruby orb
* Reduce resource_class of bundle and build jobs
* Store both Gemfile.lock files as artifacts

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,10 +10,10 @@ jobs:
         default: 2.7.7
       bundler_version:
         type: string
-        default: 2.3.13
+        default: 2.4.8
       rails_version:
         type: string
-        default: 6.1.6.1
+        default: 6.1.7.2
     executor:
       name: 'samvera/ruby'
       ruby_version: << parameters.ruby_version >>
@@ -71,10 +71,10 @@ jobs:
         default: 2.7.7
       bundler_version:
         type: string
-        default: 2.3.13
+        default: 2.4.8
       rails_version:
         type: string
-        default: 6.1.6.1
+        default: 6.1.7.2
     executor:
       name: 'samvera/ruby_fcrepo_solr_redis_postgres'
       ruby_version: << parameters.ruby_version >>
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Could you explain more about the reason or value or switching to `ruby/rspec-test` from `samvera/parallel_rspec`?

With the use of `ruby/install-deps`, `node/install-packages`, and `ruby/rspec-test`, I'm wondering if the samvera orb supplied commands that aren't idiosyncratic are still useful.  The samvera orb met a need before, but maybe the standard orbs can meet our needs and have actually matured past the orb such that it would be a benefit to switch to them (in terms of increased functionality, robustness, and ongoing maintenance).  I guess all of this is to ask if you envision Hyrax's use of orbs to be shifting and the Samvera orb should adapt accordingly?

At face value all of these changes look fine.

dlpierce:
From what I could tell the other orbs were not available when the samvera orb was being created, so yes I do think they have evolved past the change abilities of our orb. Ours is certainly still useful for the engine cart command and the executors. 

In the case of rspec-test, yes it is to hopefully reduce needed maintenance. I expect the ruby orb would be updated in the event Circle changes how tests are split or results reported. 

For ruby/install-deps it allows using a different Gemfile than the default, a capability the Samvera orb lacks. Perhaps a version 2.x of the samvera orb could wrap the ruby orb command? That would allow us to retain the extra checksummed files used for the cache key. 

I still had to call yarn install directly for the engine in order to use caching.

PR #5983, opened 2023-03-15
chart: bump the solr version for current kubernetes support

no-reply:
Kubernetes 1.25 drops support for apis used by the solr/zk charts in the versions referenced here. use the latest chart version for compatibility.

@samvera/hyrax-code-reviewers


dunn (approved):

no-reply:
i'm running this locally to ensure solr and zk come up as expected, and will merge when that succeeds

no-reply:
success!

solrax

Issue #5794, opened 2022-08-12
Works view page - Items Actions - Download not working


jlhardes:
### Descriptive summary

The Download option on the drop-down menu for Items on work view page is not working. Selecting Download in drop-down menu gives rise to ActiveFedora::ObjectNotFoundError in Hyrax::DownloadsController#show.

### Expected behavior

File should download 

### Actual behavior

New tab opens with error message:

![Screen Shot 2022-08-12 at 9 52 16 AM](https://user-images.githubusercontent.com/3440166/184367744-86f64c65-098b-4b90-b609-f310ae1f2656.png)

### Steps to reproduce the behavior

1. Log in to user dashboard, click work name to display work view page.
2. In the section of "Items", find the dropdown menu "Select an action" and select Download from the drop-down menu and click.
3. You will see an error like ActiveFedora::ObjectNotFoundError in Hyrax::DownloadsController#show

### Related work

#5136  (epic with history of previous work and testing of Works)
#5138 (epic for loading a file and being able to download it from a Work)
#5575 Issue for Edit and Versions options in Items Actions drop-down menu
#5795 Issue for Delta option in Items Actions drop-down menu

alishaevn:
this ticket was auto closed when the attached pr was merged. reopening so that it can be qa'd.

cc: @dlpierce

gamontoya:
@rjkati This ticket can be closed. Actions --> Download is working

PR #5987, opened 2023-03-22
5794 download file redux

orangewolf:
Fixes https://github.com/samvera/hyrax/issues/5794 ; Depends on merge of https://github.com/samvera/hyrax/pull/5626

Adopts surfliner's DownloadsController and uses it when using valkyrie. Allows download of files stored in a valkyrie storage adapter.


orangewolf:
just a note. specs are working in pg now, but failing in dassie / non-valkerie. need to figure out how to have both.

orangewolf:
> @orangewolf is this something it would be helpful for me to carry over the line today/tomorrow?

@no-reply for sure - I'm sorry I missed your earlier comment. I'm out on vacation. I haven't had time to figure out why the specs are mad.

no-reply:
i think the remaining issue here is that the test setup on the valkyrie specs isn't valid for AF/Wings. an easy fix for this could be to simply not use this controller behavior when Wings is in use, and skip the tests under those conditions.

i think in that case we're missing test coverage for Valkyrie altogether, though, since `use_valkyrie` is only true in our test suite when Wings is in use.

orangewolf:
@no-reply i agree with you. There isn’t really a Valkyrie pg test set up. Do you or @dlpierce have to circleci knowledge to easily create one? If so, I’m tempted to say we do so and then I can task a junior to mark the failures pending if and only if they are use_Valkyrie and not wings. That lets us get green and move in to the new Valkyrie work (which starts Monday) with a functioning test suite we can add to.

dlpierce:
@orangewolf I'll propose getting Circle setup for this testing as a topic for this week's dev con that I'll lead.

alishaevn:
[surfliner version](https://gitlab.com/surfliner/surfliner/-/blob/trunk/comet/spec/controllers/hyrax/downloads_controller_spec.rb) of downloads_controller_spec.rb

dlpierce:
These remaining spec failures where describedby is nil seem to point to the AF::File/Hyrax::FileMetadata conversion being incomplete. The [spec for persisting a file](https://github.com/samvera/hyrax/blob/071456f40b4ead27d16e0ec8377004bc43e56a09/spec/wings/active_fedora_converter_spec.rb#L303) is a stub. I expect this is due to the special handling for some File metadata, as mention in the [commit](https://github.com/samvera/hyrax/commit/f46b29b3663e19873f2fb593116c0f3de5e6b3d3).

I suggest making sure the Hyrax::FileSet factory can correctly use its [`with_files`](https://github.com/samvera/hyrax/blob/071456f40b4ead27d16e0ec8377004bc43e56a09/spec/factories/hyrax_file_set.rb#L60-L62) trait to create valid Files.

orangewolf:
> These remaining spec failures where describedby is nil seem to point to the AF::File/Hyrax::FileMetadata conversion being incomplete. The [spec for persisting a file](https://github.com/samvera/hyrax/blob/071456f40b4ead27d16e0ec8377004bc43e56a09/spec/wings/active_fedora_converter_spec.rb#L303) is a stub. I expect this is due to the special handling for some File metadata, as mention in the [commit](https://github.com/samvera/hyrax/commit/f46b29b3663e19873f2fb593116c0f3de5e6b3d3).
> 
> I suggest making sure the Hyrax::FileSet factory can correctly use its [`with_files`](https://github.com/samvera/hyrax/blob/071456f40b4ead27d16e0ec8377004bc43e56a09/spec/factories/hyrax_file_set.rb#L60-L62) trait to create valid Files.

I spent a lot of time trying to get the specs to persist files properly. I got it almost working, but then realized that it slowed the specs way down. So instead we'll use the test adapters for this spec. That keeps it speedy and keeps it focused on testing the Valkyrie controller logic and not on testing the underlying library.

orangewolf:
@dlpierce this is down to just non-related flaky specs. are you ok with the proposed testing solution?

alishaevn (dismissed):

dlpierce (approved):

PR #5990, opened 2023-03-23
Expanding IIIF Manifest gem version

kirkkwang:
This commit will allow IIIF Manifest gem versions under 2.0.


cjcolvar (approved):
Looks good to me! :tada:

orangewolf:
2.x specs run locally but not in CI. Since 2.x is no longer in the maintenance window because 4 came out I feel like it is ok to merge this without trying to fix Circle. It was in under the wire for 2.x tickets.

PR #5999, opened 2023-03-31
Hyrax-5933 - Review submission support paging

bbpennel:
Fixes https://github.com/samvera/hyrax/issues/5933

Updates the Review Submission page to make it behave more like the my/works page, meaning:

1. Each tab now makes a separate request for works, either Under Review or Published
2. Searching, paging and updating page size now trigger new requests
3. Paging of results is performed server-side

As a result, ActionableObjects now supports query and paging related parameters, which are also now supported in `scope_entities_for_the_user`. 

**Guidance for testing**

1. Create an admin set using the provided mediated deposit workflow (using dassie)
2. Deposit 11 works into the mediate deposit workflow admin set
3. Without the changes in this PR, only the first 10 should show, without the ability to show more. With this PR, it should be possible to page to the 11th item, or to use the page size dropdown to display all items.
4. Next, navigate to one of the works and approve it
5. In the "Review submissions" page, with this PR 10 under review works should show, and in the Published tab there should be 1 item.
6. The search input should also work to filter by title.

@samvera/hyrax-code-reviewers


bbpennel:
While back porting this to hyrax 3 and testing it with my local application, I am realizing that the result count doesn't take into account the keyword filtering, since that is taking place in solr after the sipity queries have completed. I'm trying to figure out if there's any way to support that behavior.

bbpennel:
I ended up removing the option to search by title since I couldn't think of a reasonable way to make that work. I made it sort by submission date instead, since that was available in both psql and solr, and would at least make the ordering non-random.

orangewolf (dismissed):

orangewolf:
this is a great thing to see! the 'fake' js paginating on this and notifications causes lots of problems at scale =-)

alishaevn (approved):

aprilrieger (approved):

orangewolf:
I'm working on workflows approval for Valkyrie today. I'm glad to have this merged in before I start on that =-D

PR #6009, opened 2023-04-25
make the VisibilityMap configurable

no-reply:
it's a common use case to extend or otherwise change the visibilities supported in a Hyrax install. making the map instance used throughout Hyrax configurable should make this easier.

@samvera/hyrax-code-reviewers


lsitu (approved):
Looks good!

PR #6023, opened 2023-04-26
make work_show_spec.rb pass under Ruby 3

no-reply:
this feels pretty drafty. i think i may go back and write unit tests for the relevant behavior under ruby 2.7 before moving any further.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -41,12 +41,12 @@ module Hyrax::Controller
 
   ##
   # @note for Blacklight 6/7 compatibility
-  def search_service(**search_params)
+  def search_service(*search_params)
     return super if defined?(super) && search_params.empty?
 
     search_service_class.new(config: blacklight_config,
                              scope: self,
-                             user_params: search_params,
+                             user_params: search_params.first,
                              search_builder_class: blacklight_config.search_builder_class)
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/hyrax/controller.rb
```diff
@@ -41,12 +41,12 @@ def search_results(*args)
...
   ##
   # @note for Blacklight 6/7 compatibility
-  def search_service(**search_params)
+  def search_service(*search_params)
     return super if defined?(super) && search_params.empty?
 
     search_service_class.new(config: blacklight_config,
                              scope: self,
-                             user_params: search_params,
+                             user_params: search_params.first,
```
dlpierce, line 49: Are you happy with this part at least? I think it will fix a few different specs, and looks cleaner than what I came up with:
```
def search_results(*args)
    return super if defined?(super) # use the upstream if present (e.g. in BL 6)

    search_service(**args.reduce({}, :merge)).search_results
  end
```

no-reply: i think this might be better: https://github.com/samvera/hyrax/pull/6039

but i'm unsure if it will make any tests pass, or just make them fail on the PermissionQuery code.

no-reply: #6042 fixes the other half of this! it turns out there *are* unit tests for that behavior, and they were failing. i just hadn't spotted them.

PR #6028, opened 2023-04-27
`ControlledVocabularies::ResourceLabelCaching` ruby 3 compatibility

no-reply:
@samvera/hyrax-code-reviewers


dlpierce (approved):
"I don't think he knows about second splat argument, Pip."

PR #6029, opened 2023-04-27
make `Wings::Valkyrie::Storage` ruby 3 compatible

no-reply:
@samvera/hyrax-code-reviewers


lib/wings/valkyrie/storage.rb
```diff
@@ -69,8 +69,9 @@ def find_versions(id:)
 
         reader = RDF::Reader.for(content_type: response.headers['content-type'])
         version_graph = RDF::Graph.new << reader.new(response.body)
+        query = {predicate: RDF::Vocab::Fcrepo4.hasVersion}
```
no-reply, line 72:
```suggestion
        query = { predicate: RDF::Vocab::Fcrepo4.hasVersion }
```

dlpierce (approved):

PR #6033, opened 2023-04-27
do some ugly argument handling in CollectionBrandingInfo model

no-reply:
it seems like ActiveRecord won't handle these kinds of kwargs anymore

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/collection_branding_info.rb
+++ b/app/models/collection_branding_info.rb
@@ -1,16 +1,17 @@
 # frozen_string_literal: true
 class CollectionBrandingInfo < ApplicationRecord
-  def initialize(collection_id:,
-                 filename:,
-                 role:,
-                 alt_txt: "",
-                 target_url: "")
+  def initialize(*_args, **opts)
+    collection_id = opts.fetch(:collection_id)
+    filename = opts.fetch(:filename)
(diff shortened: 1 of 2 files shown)


dlpierce:
What about refactoring to use attributes and an after_initialize callback instead of overriding `initialize`?
```
  attr_accessor :collection_id, :filename, :role, :alt_txt, :target_url
  after_initialize :set_collection_attributes

  def set_collection_attributes
    self.collection_id = collection_id
    self.role = role
    self.alt_text = alt_txt
    self.target_url = target_url
    self.local_path = File.join(role, filename)
  end
```

no-reply:
> What about refactoring to use attributes and an after_initialize callback instead of overriding `initialize`?
> 
> ```
>   attr_accessor :collection_id, :filename, :role, :alt_txt, :target_url
>   after_initialize :set_collection_attributes
> 
>   def set_collection_attributes
>     self.collection_id = collection_id
>     self.role = role
>     self.alt_text = alt_txt
>     self.target_url = target_url
>     self.local_path = File.join(role, filename)
>   end
> ```

that seems fine to me. do you want to propose a separate PR?

PR #6034, opened 2023-04-27
Refactor `CollectionBrandingInfo#initialize` for Ruby 3

dlpierce:
@samvera/hyrax-code-reviewers


no-reply (approved):

no-reply:
it seems like this is genuinely failing

no-reply (approved):

PR #6036, opened 2023-04-27
fix transfers controller behavior

no-reply:
@samvera/hyrax-code-reviewers


no-reply:
this reverts https://github.com/samvera/hyrax/pull/6032, which seems to have broken all the TransfersController tests, as well as a few system tests.

i'm not sure which specs that PR originally targeted, so (at least for me) the best next step is to roll it back and continue checking off tests failing in the Ruby 3 build.

PR #6048, opened 2023-04-28
Replace kaminari_route_prefix gem with initializer patch

dlpierce:
Fixes spec/views/hyrax/users/index.html.erb_spec.rb on Ruby 3

kaminari_route_prefix is not Ruby 3 compatible, and the engine routing feature it supplies is (still!) not part of kaminari core. This replaces it with an in-tree patch. The changes to views are not included as they seem unused in hyrax.

* Remove kaminari_route_prefix gem
* Add initializer patch to support the `route_set` option for `paginate`

* Pages that use kaminari pagination still function as expected

@samvera/hyrax-code-reviewers


no-reply (approved):
love this. barely 10 lines of inline code replacing an entire dependency is always a good call as far as i'm concerned

Issue #6008, opened 2023-04-18
Metadata fields not displaying in UV


jlhardes:
### Descriptive summary

Testing hyrax 4.0.0.rc1 locally (dassie), metadata fields do not display in the "More information" menu in the universal viewer. Required fields Title, Creator, and Rights Statement are not showing. Also added Description, Abstract, Subject, Keywords, and Date Created. One field that does show is Description.

### Expected behavior

Descriptive metadata fields that show below the viewer should also show in the More Information menu inside the viewer.

### Steps to reproduce the behavior

1. Create a Generic Work with required metadata (Title, Creator, Rights Statement) and attach an image file (so the viewer will load)
2. Save work then view the work and open More Information in the viewer. See that Title, Creator, and Rights Statement do not show in More Information in the viewer.
3. Edit the work and add metadata to additional fields: Description, Abstract, Subject, Keywords, Date Created
4. Save work then view the work and open More Information in the viewer. See Description is there but no other fields.
5. View all descriptive metadata below the viewer on the Work View page.

### Related work

Reported initially as a problem on dev.nurax - https://github.com/samvera-labs/nurax/issues/5

rjkati:
metadata now displays in UV in Hyrax 4.0.0.rc2

PR #6051, opened 2023-05-01
bump Version to Hyrax 4.0.0-rc2

no-reply:
@samvera/hyrax-code-reviewers


dlpierce:
I'd like to get #6052 to green and see if there is a quick fix for #6008

dlpierce (approved):

PR #6054, opened 2023-05-09
support enumerators in FindFiles "navigator" queries

no-reply:
in addition to supporting enumerable values from
`find_many_file_metadata_by_use`, this avoids double peeking at the first element in an array. we can just give the element, and if it's `nil` fallback to error behavior.

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Nice!

PR #6055, opened 2023-05-10
reinstate breadcrumb styling

no-reply:
fixes unstyled breadcrumbs in Ruby 3.

in the meanwhile, don't hardcode the breadcrumb builder implementation in the
views. make the builder configurable so new behavior can be injected downstream.

@samvera/hyrax-code-reviewers


dlpierce (approved):
Thanks for the fix!

no-reply:
wondering if we can get this in a release, either an RC3, or are we ready to release 4.0.0 and this can be 4.0.1?

Issue #6056, opened 2023-05-10
Replace use of deprecated `Redis.current`


dlpierce:
`Redis.current` is deprecated but is used in several places in hyrax.
- https://github.com/samvera/hyrax/blob/7e091130e84400eea8032e9699f24f6f581856d3/lib/hyrax/redis_event_store.rb#L26
- https://github.com/samvera/hyrax/blob/7e091130e84400eea8032e9699f24f6f581856d3/app/services/hyrax/lock_manager.rb#L14
- https://github.com/samvera/hyrax/blob/7e091130e84400eea8032e9699f24f6f581856d3/lib/generators/hyrax/templates/config/initializers/redis_config.rb#L4
See https://github.com/redis/redis-rb/wiki/Connection-Management for replacement suggestions.

PR #6060, opened 2023-05-10
add pool based redis connections #6056 

no-reply:
moves new applications away from `Redis.current` per https://github.com/samvera/hyrax/issues/6056.

the proposed solution here is to use a global $hyrax_redis variable to capture a
`ConnectionPool::Wrapper`. the wrapper construction provides compatibility with
the existing connections, allowing us to migrate code stepwise. when the global
is not set, we continue to use `Redis.current`.

the new approach isn't setup to support Redis::Namespace in the way the older
setup is. we used to set `Redis::Namespace` everytime `RedisEventStore.instance`
was called(!). rather than repeating this pattern, we can recommend that
existing apps use `Redis::Namespace` in their pool setup:

```ruby
$redis = ConnectionPool::Wrapper.new(size: size , timeout: timeout) do
  Redis::Namespace.new(Hyrax.config.redis_namespace, config)
end
```

new apps that wish to continue to use `Redis::Namespace` can do so (e.g. Hyku
will want this). it's advisable for new apps to avoid this (we looked at
removing `redis-namespace` well over 5 years ago:
https://github.com/samvera/hyrax/issues/1671). http://www.mikeperham.com/2015/09/24/storing-data-with-redis/
is relevant.

Hyrax doesn't actually seem to use multiple namespaces, so it should be fine for
apps to just configure their Redis to use one of the numbered databases. folks
who have a problem with this will certainly know it and be able to setup
Redis::Namespace on their own.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/config/initializers/redis_config.rb
+++ b/.dassie/config/initializers/redis_config.rb
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
 require 'redis'
+require 'connection_pool'
 config = YAML.safe_load(ERB.new(IO.read(Rails.root.join('config', 'redis.yml'))).result)[Rails.env].with_indifferent_access
-Redis.current = Redis.new(config.merge(thread_safe: true))
+config[:thread_safe] = true
+
+size = ENV.fetch("HYRAX_REDIS_POOL_SIZE", 5)
+timeout = ENV.fetch("HYRAX_REDIS_TIMEOUT", 5)
+
+Hyrax.config.redis_connection =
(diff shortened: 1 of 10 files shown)


dlpierce:
Why not pin reform in the gemspec? If not there, then it would need to be pinned in spec/test_app_templates/Gemfile.extra to affect the test app generation.

no-reply:
some follow up work is proposed in #6061

dlpierce (approved):

dlpierce (approved):

PR #6074, opened 2023-05-24
[3.x] Google Analytics 4 Reporting

dlpierce:
Adds basic support for reporting to GA4 using the `ga4` provider name.

Updating the Hyrax analytics dashboard is still TODO. The existing google analytics service class is copied with only the config key changed.

### DIFF
--- a/app/services/hyrax/analytics/ga4.rb
+++ b/app/services/hyrax/analytics/ga4.rb
@@ -1,0 +1,204 @@
+# frozen_string_literal: true
+require 'oauth2'
+require 'signet/oauth_2/client'
+
+# rubocop:disable Metrics/ModuleLength
+module Hyrax
+  module Analytics
+    module Ga4
+      extend ActiveSupport::Concern
+      # rubocop:disable Metrics/BlockLength
+      class_methods do
+        # Loads configuration options from config/analytics.yml. Expected structure:
+        # `analytics:`
+        # `  ga4:`
+        # `    app_name: <%= ENV['GOOGLE_OAUTH_APP_NAME']`
+        # `    app_version: <%= ENV['GOOGLE_OAUTH_APP_VERSION']`
+        # `    privkey_path: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_PATH']`
+        # `    privkey_secret: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_SECRET']`
+        # `    client_email: <%= ENV['GOOGLE_OAUTH_CLIENT_EMAIL']`
+        # @return [Config]
+        def config
+          @config ||= Config.load_from_yaml
+        end
+
+        class Config
+          def self.load_from_yaml
+            filename = Rails.root.join('config', 'analytics.yml')
+            yaml = YAML.safe_load(ERB.new(File.read(filename)).result)
+            unless yaml
+              Hyrax.logger.error("Unable to fetch any keys from #{filename}.")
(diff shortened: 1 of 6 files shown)


app/views/layouts/_head_tag_content.html.erb
```diff
@@ -26,9 +26,12 @@ signed in %>
 <%= render 'shared/appearance_styles' %>
 
 <% if Hyrax.config.analytics? %>
-  <% if Hyrax.config.analytics_provider == 'google' %>
```
orangewolf, line 29: nice case usage =-)

app/services/hyrax/analytics/ga4.rb
```diff
@@ -0,0 +1,204 @@
+# frozen_string_literal: true
```
orangewolf, line 1: is there expectation of devergence or should we be creating a concern to house a bunch of this functionality that GA and GA4 can share?

dlpierce: I was leaving the old GA code untouched in case this is in use while both services are active.

orangewolf: that makes sense to me since its likely GA3 will be dead code before too long

bbpennel:
I've been testing this branch out with our repository today. It appears to be sending `page-view` and `scroll` events, but I'm not seeing any other events being sent, like `work-view` or `file-set-in-work-view`. 

It looks like there are a couple of issues related to https://github.com/samvera/hyrax/blob/main/app/assets/javascripts/hyrax/analytics_events.js and needing to be updated for GA4. I'm also not sure we really need to call a custom page-view event, since apparently GA tracks that automatically now, and it is an event unlike in UA. Anyways, I'll make a PR for this PR soon to try to address some of these issues.

bbpennel:
PR for this PR to get events working for all 3 providers:
https://github.com/samvera/hyrax/pull/6095

dlpierce:
Thank you @bbpennel!

dlpierce:
I agree, this should be merged to at least get data recording available.

@bbpennel or @orangewolf, have you looked at this [api wrapper](https://github.com/dailytics/google_analytics_v4_api) that was linked to in the legato GA4 PR? Do you know if it's missing any major components that hyrax would need?

bbpennel (approved):

bbpennel:
I can't claim to have a comprehensive understanding of our usage of legato, but the parts I'm familiar with are requests to the reporting API to search by event name, date, and category. My impression is that usage is supported by dailytics/google_analytics_v4_api. On the other hand, it seems like there is only a single developer (like legato) and it isn't used much elsewhere, not that those are deal breakers. It does refer to GA4 as being in beta in a few places, including some of the default URLs, so its possible it could use some updating as well.

When trying to research it, my search engine thought I wanted to look at [google-apis-analyticsreporting_v4](https://github.com/googleapis/google-api-ruby-client/blob/main/generated/google-apis-analyticsreporting_v4/OVERVIEW.md) instead, which appears to be a google developed client for talking to the GA reporting API. That got my hopes up, but honestly, its documentation is super basic and I think the code might be generated based off a definition of the API, so I'm not sure how useful it is.

So we could look into `google_analytics_v4_api`, which would require some significant refactoring of the reporting code. My worry is just that we might end up in the same situation as legato. On the other hand, the client is tiny and obviously tightly scoped to just GA4, so I guess it would come with the understanding that we'd have to replace it with a new client whenever a new major GA version was released.

Issue #5946, opened 2023-01-24
Flaky Tests related to Collection Membership


dlpierce:
These specs sometimes have failures.
https://app.circleci.com/pipelines/github/samvera/hyrax/8566/workflows/abdc14f3-49fc-4951-9966-65b6e81353f7/jobs/66661/tests

```
Failures:

  1) Hyrax::Collections::CollectionMemberService.add_member when single-membership collection error raises an error
     Failure/Error:
       expect { described_class.add_member(collection_id: collection2.id, new_member: new_member, user: user) }
         .to raise_error Hyrax::SingleMembershipError, regexp
     
       expected Hyrax::SingleMembershipError with message matching /Error: You have specified more than one of the same single-membership collection type \(type: Greedy, collections: (Col-1 and Col-2|Col-2 and Col-1)\)/ but nothing was raised
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-support-3.11.1/lib/rspec/support.rb:102:in `block in <module:Support>'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-support-3.11.1/lib/rspec/support.rb:111:in `notify_failure'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.11.1/lib/rspec/expectations/fail_with.rb:35:in `fail_with'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.11.1/lib/rspec/expectations/handler.rb:40:in `handle_failure'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.11.1/lib/rspec/expectations/handler.rb:56:in `block in handle_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.11.1/lib/rspec/expectations/handler.rb:27:in `with_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.11.1/lib/rspec/expectations/handler.rb:48:in `handle_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.11.1/lib/rspec/expectations/expectation_target.rb:65:in `to'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.11.1/lib/rspec/expectations/expectation_target.rb:139:in `to'
     # ./spec/services/hyrax/collections/collection_member_service_spec.rb:206:in `block (4 levels) in <top (required)>'
[...]

Issue #5945, opened 2023-01-24
MultipleMembershipCheckerSpec has flaky tests


dlpierce:
https://app.circleci.com/pipelines/github/samvera/hyrax/8581/workflows/fa34c591-3c80-49d4-9050-7207bda37b70/jobs/66723/tests

All specs expecting an error message sometimes fail.

```
Failures:

  1) Hyrax::MultipleMembershipChecker#check when called from actor stack and multiple single-membership collections of the same type exist and both are in the list returns an error
     Failure/Error: expect(subject).to match regexp
       expected nil to match /Error: You have specified more than one of the same single-membership collection type \(type: Greedy, collections: (Foo and Bar|Bar and Foo)\)/
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-support-3.12.0/lib/rspec/support.rb:102:in `block in <module:Support>'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-support-3.12.0/lib/rspec/support.rb:111:in `notify_failure'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.12.2/lib/rspec/expectations/fail_with.rb:35:in `fail_with'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.12.2/lib/rspec/expectations/handler.rb:38:in `handle_failure'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.12.2/lib/rspec/expectations/handler.rb:56:in `block in handle_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.12.2/lib/rspec/expectations/handler.rb:27:in `with_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.12.2/lib/rspec/expectations/handler.rb:48:in `handle_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.12.2/lib/rspec/expectations/expectation_target.rb:65:in `to'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.12.2/lib/rspec/expectations/expectation_target.rb:101:in `to'
     # ./spec/services/hyrax/multiple_membership_checker_spec.rb:102:in `block (6 levels) in <top (required)>'
     

[...]

PR #6075, opened 2023-05-24
WIP workaround for wrong query service in rspec tests

aploshay:
refs #5945 ; refs #5946

Demonstration-of-concept that works around a possible flaky test failure, without addressing the root cause -- a class variable persisting, unintentionally and unexpectedly, across different rspec tests.  The flaky test failure occurs when an earlier test runs with the `valkyrie_adapter: :test_adapter` option, and the modified class variable value for `Hyrax::SolrQueryService.query_service` persists into tests _not_ using that option.

Here's an rspec to reproduce one failure condition:
```bash
rspec './spec/features/collection_multi_membership_spec.rb[1:2:2:1:1,1:2:2:1:2,1:2:2:1:3
]' './spec/features/valkyrie_search_spec.rb[1:1:2]' --seed 57764 -f doc
```
and other ones:
```
 rspec './spec/controllers/hyrax/dashboard/collection_members_controller_spec.rb[1:1:5:1,1:1:5:2]' './spec/features/delete_resource_spec.rb[1:1]' --seed 28762

rspec './spec/features/valkyrie_search_spec.rb[1:1:2]' './spec/javascripts/jasmine_spe
c.rb[1:1]' './spec/services/hyrax/multiple_membership_checker_spec.rb[1:2:4:1:2:1,1:2:4:2:2:1,1:2:5:1:1:1:1,1:2:5:2:
2:1,1:3:5:2:1,1:3:5:3:1]' './spec/validators/hyrax/collection_membership_validator_spec.rb[1:1:3:2:3:2:1,1:1:3:2:3:3
:1]' --seed 13361

rspec './spec/features/delete_resource_spec.rb[1:1]' './spec/services/hyrax/collections
/collection_member_service_spec.rb[1:5:3:1]' --seed 53302

```

A more detailed breakdown:
* `SolrQueryService` uses [a class variable to set `query_service`](https://github.com/samvera/hyrax/blob/main/app/services/hyrax/solr_query_service.rb#L10)
* `valkyrie_search_spec` [runs with `valkyrie_adapter: test_adapter`](https://github.com/samvera/hyrax/blob/main/spec/features/valkyrie_search_spec.rb#L2), and [when the search is run](https://github.com/samvera/hyrax/blob/main/spec/features/valkyrie_search_spec.rb#L58), this locks in `Valkyrie::Persistence::Memory::QueryService` as the `query_service` class returned by both `Hyrax` and `Hyrax::SolrQueryService`
* subequent tests for `collection_multi_membership` expect [a failure trying to save the example record in more than one Collection of a type that disallows multi-collection membership](https://github.com/samvera/hyrax/blob/main/spec/features/collection_multi_membership_spec.rb#L99).  For the validation check to fail (as expected), the [query for parent collections](https://github.com/samvera/hyrax/blob/main/app/services/hyrax/solr_query_service.rb#L55-L61) has to return non-empty results.  The search class that's intended, and that's returned by `Hyrax.query_service` in this context, is `Wings::Valkyrie::QueryService` -- that returns non-empty array results.  But if the pertinent `valkyrie_search_spec` test is run first, then `Hyrax::SolrQueryService.query_service` still returns a `Valkyrie::Persistence::Memory::QueryService` -- this query service returns an empty result set for the query, so the record validation passes, which breaks all the subsequent test checks.

You can confirm this by adding a `debugger` call just before the committed change on line 12, and checking:
```ruby
Hyrax::SolrQueryService.query_service.class # => Valkyrie::Persistence::Memory::QueryService -- wrong
Hyrax.query_service.class # => Wings::Valkyrie::QueryService -- right
```

So, the committed change sort of works around the issue by forcing a reset of the class variable.   This is not an ideal solution.  But the "problem" should probably also only occur in the test context, anyway, where we are hotswapping the valkyrie_adapter during different tests.

This is more of a demonstration-of-concept workaround; a proper change might be refactoring out of using the class variable, or finding a way to get it properly re-set between tests?

(see "You can confirm this", above)

Build is also experiencing a test failure with:
```
rspec './spec/jobs/valkyrie_ingest_job_spec.rb[1:1:2:1]' --seed 19605
```
which is an unrelated issue.  Error arsises from [ActiveFedora's File#describe_by](https://github.com/samvera/active_fedora/blob/v14.0.1/lib/active_fedora/file.rb#L52-L55) and is resolved by adding a nil guard and `to_s` call:
```rb
    def described_by
      raise "#{self} isn't persisted yet" if new_record?
      # links['describedby'].first
      links['describedby']&.first
    end
 ```
 but that just kicks the can down the road to another error, so the root issue is probably `pcdm_object` missing some effectively required metadata.

@samvera/hyrax-code-reviewers


dlpierce:
#6106 is a possible way to solve this. It passed tests the first try, so that's a good sign!

orangewolf:
@aploshay should this be closed now that #6106 is merged in?

aploshay:
@orangewolf yup!

PR #6076, opened 2023-05-24
add extension behavior for FileSet show view partials

no-reply:
if a user wants to extend the FileSet show view to, e.g. show file specific
details or relationships to other objects, they should be able to inject that
behavior via the presenter

extensions were vaguely waved at around Sufia 7:
https://github.com/samvera-deprecated/sufia/issues/1481

we have some in mind in Surfliner:
https://gitlab.com/surfliner/surfliner/-/issues/1282

* adds `FileSetPresenter#show_partials` to list partial paths to render in the guts of the FileSet show page.

* no functional changes.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/file_sets/show.html.erb
+++ b/app/views/hyrax/file_sets/show.html.erb
@@ -11,7 +11,7 @@
         <%= render 'file_set_title', presenter: @presenter %>
       </header>
 
-      <%# TODO: render 'show_descriptions' See https://github.com/samvera/hyrax/issues/1481 %>
+      <%# TODO: render 'show_descriptions' See https://github.com/samvera-deprecated/sufia/issues/1481 %>
       <%= render 'show_details' %>
       <%= render 'hyrax/users/activity_log', events: @presenter.events %>
     </div><!-- /columns second -->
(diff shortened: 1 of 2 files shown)


app/views/hyrax/file_sets/show.html.erb
```diff
@@ -11,8 +11,10 @@
         <%= render 'file_set_title', presenter: @presenter %>
       </header>
 
-      <%# TODO: render 'show_descriptions' See https://github.com/samvera/hyrax/issues/1481 %>
-      <%= render 'show_details' %>
+      <% @presenter.show_partials.each do |partial_path| %>
+      <%= render partial_path %>
+      <% end %>
```
alishaevn, line 16:
```suggestion
      <% @presenter.show_partials.each do |partial_path| %>
        <%= render partial_path %>
      <% end %>
```

alishaevn (changes requested):

app/views/hyrax/file_sets/show.html.erb
no-reply: @alishaevn i don't think we want to nest indentation here? that will result in the partial HTML content being indented a level too far (since the `#each` block won't appear in the rendered HTML).

alishaevn (approved):

Issue #5594, opened 2022-04-04
NoMethodError `errors` for <GenericWork> when adding work to two collections of same single membership type (nurax-pg)


elrayle:
### Descriptive summary

Adding a work to two collections of the same single membership type raises `NoMethodError - undefined method 'errors' for <GenericWork>`

### Expected behavior

When attempting to add a work to a second collection of the same single membership type, after clicking Save, should stay in the editor and see flash message explaining why this is not allowed.

### Actual behavior

When attempting to add a work to a second collection of the same single membership type, after clicking Save, it attempts to stay in the editor but raises an exception...

<img width="838" alt="image" src="https://user-images.githubusercontent.com/6855473/161634415-dfeb34ea-492b-4c8c-88c4-055bdbfc30b7.png">

### Steps to reproduce the behavior

On nurax-pg...

Prerequisite:
1. Create a collection type that is single membership
2. Create two collections of that type

To Reproduce:
1. Dashboard -> Works -> Add new work -> select any work type
2. Fill in required metadata
3. Select tab: Relationships
4. In Add to collection selection, type name of first single membership collection and click button: Add
5. In Add to collection selection, type name of second single membership collection and click button: Add
6. Click button: Save changes

See exception shown in Actual behavior.

### Related work

Issue #5582

acklose:
@jlhardes When I reviewed this, I got a different error message.

NameError in Hyrax::GenericWorksController#update
uninitialized constant ActiveModel::DeprecationHandlingMessageArray 

<img width="944" alt="Nurax-PB_SingleMembership_twice_2022-09-15" src="https://user-images.githubusercontent.com/26173627/190511044-f8b71ad7-7895-4d78-bbee-b5f8071613f8.png">

gamontoya:
This appears to be working on Koppie. I'm not sure if seeing two errors reported to the user is expected behavior, however.

![singleMem2](https://user-images.githubusercontent.com/4314914/190527488-38ca4805-087c-4f48-8204-3210e0cb3366.png)

blancoj:
@gamontoya The issue you were seeing on Koppie, I also saw and @cjcolvar helped me resolve it.  It is happening because you need the latest version of rails running on Koppie.  This is how you get the latest rails installed on Koppie:

> docker exec -it koppie-app-1 /bin/sh
> bundle update rails

The double error, is what I also saw.  I think this is expected behavior, but it would be good to get someone else to confirm that.

gamontoya:
[...]

Issue #4011, opened 2019-09-17
Adding work to multiple single membership collections from edit work form does not forward as expected


elrayle:
### Descriptive summary

This issue lays out expected forwarding behavior when adding a work to multiple collections through the edit work form fails.

**NOTE: This likely uses the same behaviors when performing the same actions on the new work form.  This issue and issue #4010 are tightly related and the fix for one may fix both.**

### Rationale

Provide intuitive forwarding behavior that focuses on the actions a user will most likely want to take next.  Provide error message that is easy to understand.

### Expected behavior

When adding a work to multiple collections from the work edit form...

FAILURE to add the work to the collections should:  
- [ ] stay on edit form (Relationships tab)
- [ ] no changes are saved
- [ ] should maintain changes to metadata fields (Descriptions tab)
- [ ] should maintain any files added for upload (Files tab)
- [ ] show flash error (red): Collections Error: You have specified more than one collection of the same single-membership collection type (collection type: _name of collection type_, collections: _list of effected collections_).  The work was not added to those collections.

#### Testing

- [ ] Need to update specs added in PR #2687 to match the forwards described in this issue.

### Actual behavior

When adding a work to multiple collections from the work edit form...

FAILURE to add the work to the collections should:  
* stays on edit form (Descriptions tab)
* no changes are saved
* all metadata fields are reset (Descriptions tab)
* shows flash error: Collections Error: You have specified more than one of the same single-membership collection types: _name of collection type_ (_list of effected collections_)

#### Prerequisites: 

Need a collection type that is set to allow single membership.  If testing on Nurax, this collection type may already exist.  Look for Collection Type Single Membership.  If it doesn't exist, then create a single membership collection type using...

* Dashboard -> Settings -> Collection Types -> Create new collection type
* Assign type name Single Membership and Save 
* On Settings tab, uncheck MULTIPLE MEMBERSHIP and Save changes

Need two collections of type Single Membership.  You can create using...

* Dashboard -> Collections -> New Collection -> select Single Membership type -> Create Collection
* Assign suitable titles for the two collections and save.  This issue will refer to them as SM1 and SM2.

Need the work to already exist.  Create a work using...

* Dashboard -> Works -> Add new work -> select type -> fill in requirements -> Save. This issue will refer to them as Work1.

#### Test:
Verify FAILURE behaves as expected.

1. Dashboard -> Works -> click title for Work 1 -> Edit
2. change some metadata on Descriptions tab
3. add file on Files tab
[...]

PR #6089, opened 2023-06-05
5594-single-membership-error-messages

alishaevn:
### Fixes
- #5594

### Summary
move and consolidate the "work form edit" related errors to the flash message.

#### Demo
|| before | after |
| --- | --- | --- | 
| koppie | image | image |
| hyrax | image| image |

* create a single membership collection type (instructions can be found on https://github.com/samvera/hyrax/issues/4011 under the "Steps to reproduce the behavior" heading in the initial comment.)
* create two collections of the single membership type
* create/find a work and add it to one of the above collections
* edit the work and try to also add it to the second collection
* confirm you get one error message

- [X] Bug Fixes
- [X] Valkyrie Progress

### Detailed Description
before this change, the flash message would show an error above the form AND there would be another error message that showed within the form. these messages would be for the same reason. instead, this commit shows unique error messages above the form only.

@samvera/hyrax-code-reviewers


app/views/_flash_msg.html.erb
```diff
@@ -7,3 +7,14 @@
     <% flash.delete(type) %>
   <% end %>
 <% end %>
+
+<% if form_errors? %>
```
jeremyf, line 11: Having the helper include the `@form` and then assuming it on line 14 below seems like we might have knowledge spread across too many places.

What if we had `<% if @form&.errors&.any? %>` here?  That would consolidate that logic into the view.

jeremyf: And based on that, I'd begin to wonder about the helper methods that reference `@form`

alishaevn: I like it!

app/helpers/hyrax/work_form_helper.rb
```diff
@@ -99,110 +99,20 @@ def form_file_set_select_for(parent:)
...
-      when Hyrax::ChangeSet
-        form.errors[:lease_expiration_date_errors]
-      else
-        form.model.errors[:lease_expiration_date_errors]
-      end
+    # @param form [Hyrax::Forms::WorkForm]
+    # @return [String] specific error message(s)
+    def form_errors(form:)
+      error_messages = []
+      form.errors.messages.each_value { |v| error_messages << v.first }
```
alishaevn, line 106: whether the form is a `Hyrax::Forms::WorkForm` or `Hyrax::ChangeSet`, `form.errors.messages` is the way to find the error messages. see my test below.

<details>
<summary> koppie </summary>

image
</details>

<details>
<summary> hyrax </summary>

image
</details>

app/helpers/hyrax/work_form_helper.rb
```diff
@@ -97,112 +97,5 @@ def form_file_set_select_for(parent:)
         hash[presenter.title_or_label] = presenter.id
       end
     end
-
-    ##
-    # This helper retrieves errors based on form type.
```
alishaevn, line 102: since we're no longer rendering error messages from "app/views/hyrax/base/_form.html.erb", these methods are no longer called.

alishaevn:
"spec/services/hyrax/collections/collection_member_service_spec.rb:203" is failing in the `ruby3-2` pipeline only. it does not fail for me locally. I ssh'd into the pipeline and it failed twice, but did not on the third time. (unsure if the success was related to putting a byebug in the code and first confirming that `new_member` wasn't already a part of `collection2`.)

I ran the pipeline in ci again, and again it has failed. 😞

labradford (approved):

aprilrieger (approved):

PR #6091, opened 2023-06-09
thumbnails and other derivatives were not stored in the path we expected them to be

### LINKED ISSUE #5504: Generate derivatives for FileSets when Valkyrie mode is enabled
### Descriptive summary

The pub/sub and transaction architecture doesn't currently generate derivatives for Valkyrie resources, resulting in no thumbnails or manifests.

### Rationale

Valkyrie resources should be able to generate derivatives.

### Expected behavior

Derivatives generate.

### Actual behavior

Derivatives don't generate.

#### Prerequisites

Configure Hyrax:

```
```
2. restart rails app

#### To reproduce

1. Navigate to: Dashboard -> Works 
2. Click button: Add new work
3. Select type: Monograph
4. File in required metadata and other required steps
5. Add an image file
6. Click button: Save

This forwards to the new work's show page.  Look at the list of items at the bottom. There is no thumbnail.

### Related work

Found as part of https://github.com/samvera/hyrax/issues/5480

orangewolf:
### Fixes
- #5504

### Summary

Derivatives were stored in a pairtree using the first 5 non-dash items. But Hyrax expects them to be stored in a pairtree of the full id. This mismatch meant no derivatives could be found properly. 

* This is a Valkyrie only change. 
* Create a work with an attached file. After a minute or two the thumbnail should show

- [ ] Major Changes (Potentially breaking changes)
- [ ] New Features
- [ ] Deprecations
- [ ] Bug Fixes
- [x] Valkyrie Progress
- [ ] Documentation
- [ ] Containerization

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/derivative_bucketed_storage.rb
+++ b/app/services/hyrax/derivative_bucketed_storage.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+# this class overrides the Valkyrie::Storage::Disk::BucketStorage class so that file paths match
+module Hyrax
+  class DerivativeBucketedStorage
+    attr_reader :base_path
+
+    def initialize(base_path:)
+      @base_path = base_path
+    end
+
+    # rubocop:disable Lint/UnusedMethodArgument
+    def generate(resource:, file:, original_filename:)
+      raise ArgumentError, "original_filename must be provided" unless original_filename
+      Pathname.new(base_path).join(*bucketed_path(resource.id)).join(original_filename)
+    end
+    # rubocop:enable Lint/UnusedMethodArgument
+
+    def bucketed_path(id)
+      # We want to use the same code the derivative process uses so that items end up
+      # stored in the place we expect them.
+      Hyrax::DerivativePath.new(id.to_s).pair_directory
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


app/services/hyrax/derivative_bucketed_storage.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
orangewolf, line 1: I would have subclassed the override here but `Valkyrie::Storage::Disk::BucketStorage` is a inner class. That makes me think it is best to replace instead of extend.

app/services/hyrax/valkyrie_persist_derivatives.rb
```diff
@@ -25,11 +25,13 @@ def self.call(stream,
       # responds to #path -- here we only have a StringIO, so some
       # transformation is in order
       tmpfile = Tempfile.new(file_set.id, encoding: 'ascii-8bit')
-      tmpfile.write stream.read
+      stream.rewind
+      output = tmpfile.write(stream.read)
+      tmpfile.flush
```
orangewolf, line 30: the adapter does not flush the buffer first, so we need to make absolutely sure the file is in the tmp location.

dlpierce (approved):

PR #6095, opened 2023-06-19
Refactor analytics_events.js

bbpennel:
Refactor analytic_events.js to abstract away details of underlying providers implementation, since ga4 does not follow the same structure as the previous impls.

### DIFF
--- a/app/assets/javascripts/hyrax/analytics_events.js
+++ b/app/assets/javascripts/hyrax/analytics_events.js
@@ -1,75 +1,125 @@
 class TrackingTags {
   constructor(provider) {
     this.provider = provider
+    switch(this.provider) {
+    case 'matomo':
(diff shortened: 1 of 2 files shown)


bbpennel:
Does anyone know the reason why we disable the default page view event that matomo includes? 
https://github.com/samvera/hyrax/blob/main/app/views/shared/_matomo.html.erb#L4
We could easily enough reenable it and disable the pageView event in analytics_events.js. In this PR, that is the way GA4 is working since pageView is one of the many automatic events. So only UA would be sending a custom page view event (I haven't researched whether that is needed or not).

Also, I left in deprecated stubs for the previous way of sending events in analytics_events.js in case anyone was directly sending events instead of sending them via event tags, but I'm fine with removing it if it doesn't seem necessary.

PR #6098, opened 2023-06-26
5844 deactivated embargoes

alishaevn:
co-authored-by: @sephirothkod 

### refs
- #5844 

related to:
- #5871 

### Summary

with this commit we will be able to properly deactivate, and reactivate, embargoes that are related to valkyrie backed works

### Demo

https://github.com/samvera/hyrax/assets/29032869/86d9f827-5c57-43c0-b48e-4531318e0845

* create a new work with an embargo
* press the "manage embargoes" link in the sidebar
* click on the name of the work you just created
* press "deactivate embargo"
* ensure that a message with the correct info shows up in the "past embargoes" field on the same page
* go back to them "manage embargoes" page and click the "deactivated embargoes" tab
* make sure that the correct message also shows up there
* click the name of the work on the "deactivated embargoes" tab
* click "apply embargo" to put a new embargo on the work
* go back to the "manage embargoes" page and make sure the work once again shows there

### Detailed Description
previously, the embargo was getting updated and then saved to the resource (work). however, the version of the embargo that was saved to the database, did _not_ get updated. this led to issues in being able to deactivate them.

this pr proposes saving the embargo to the database only, and then saving the id of the embargo to the resource. now, when `resource.embargo` is called, it goes to find the appropriately updated embargo in the database. it also means we have one source of truth.

we are also bringing over the logic from [hydra head](https://github.com/samvera/hydra-head/blob/main/hydra-access-controls/app/models/hydra/access_controls/embargo.rb) that allows us to deactivate properly and return the correct embargo history message.

@samvera/hyrax-code-reviewers


alishaevn:
a couple of these specs are flappy. it goes between 6 then 1 fail, but works locally for @sephirothkod and I.

app/models/hyrax/resource.rb
```diff
@@ -103,6 +103,17 @@ def visibility
...
 
+    def embargo=(value)
+      raise TypeError "can't convert #{value.class} into Hyrax::Embargo" unless value.is_a? Hyrax::Embargo
+
+      @embargo = value
+      self.embargo_id = @embargo.id
+    end
+
+    def embargo
+      @embargo ||= (Hyrax.query_service.find_by(id: embargo_id) if embargo_id.present?)
```
bkiahstroud, line 114: Are the extra parenthesis necessary here? 

```suggestion
      @embargo ||= Hyrax.query_service.find_by(id: embargo_id) if embargo_id.present?
```

app/services/hyrax/embargo_manager.rb
```diff
@@ -17,6 +17,7 @@ module Hyrax
   #    - "Enforced" means the object's visibility matches the pre-release
   #      visibility of the embargo; i.e. the embargo has been applied,
   #      but not released.
+  #      "Deactivate" means that the existing embargo will be removed
```
bkiahstroud, line 20: Minor syntax consistency adjustment 

```suggestion
  #    - "Deactivate" means that the existing embargo will be removed
```

lib/hyrax/transactions/steps/save.rb
```diff
@@ -33,6 +33,7 @@ def call(change_set, user: nil)
           begin
             new_collections = changed_collection_membership(change_set)
             unsaved = change_set.sync
+            unsaved.embargo = @persister.save(resource: unsaved.embargo) if unsaved.embargo.present?
             saved = @persister.save(resource: unsaved)
```
bkiahstroud, line 37: Does saving the embargo to the resource here cascade to saving the resource itself? If it does, I think the following line might be redundant when `unsaved.embargo.present? == true`

app/models/hyrax/resource.rb
alishaevn: nope. this looks to have been left over from trying some other things.

lib/hyrax/transactions/steps/save.rb
alishaevn: I don't believe it does. the embargo would be saved on this version of the resource, but unless we then save the resource itself...it wouldn't "stay". 

also, in the event that the resource does not have an embargo...we would still need the resource to be saved.

bkiahstroud: True, but then you could just write something like this:

```suggestion
          if unsaved.embargo.present?
            unsaved.embargo = @persister.save(resource: unsaved.embargo)
          else
            @persister.save(resource: unsaved)
          end
```

But you'd only want to do that if you were positive that saving the embargo also saved the resource itself. Anyway, saving twice is better than never. I don't think it's a huge deal that should hold the PR up

bkiahstroud (approved):

lib/hyrax/transactions/steps/save.rb
alishaevn: yeah, I don't believe that saving the embargo on the resource, saves the resource itself. and also...we're adding logic about leases here in the same way in #6111. so then this would become even longer.

thanks!

orangewolf (approved):

PR #6104, opened 2023-07-03
4.x GA4 Reporting Support

dlpierce:
### Summary

Updates GA4 analytics support to match Hyrax 3.6. The analytics dashboard is disabled when using GA4. It will be restored when support for the new Data API is complete.

@samvera/hyrax-code-reviewers


orangewolf (approved):

orangewolf:
@bbpennel this looks to me like a pretty clear port of the 3.x work, but I'll won't hit merge so that you can have a chance to take a look if you want to.

bbpennel (approved):
I don't have a hyrax 4 instance to test with, but looks good based on reading

PR #6105, opened 2023-07-03
Stop testing middle ruby versions

dlpierce:
Only perform tests on our min/max ruby versions. These are not required checks already.
Also updates some versions in circle.

@samvera/hyrax-code-reviewers


orangewolf (approved):
~Looks like the current failures are flaky specs~

It now looks like there might be real failures.

alishaevn (approved):

PR #6116, opened 2023-07-24
5844-save-embargoes-again

alishaevn:
### Fixes
- #5844

### Summary
we are able to save embargoes again.

* create/edit a work
* add an embargo and save
* visit "/embargoes" and see that it's there

### Detailed Description

Idk why we were not able to. the work that was done on #6098 was to be able to deactivate embargoes, and we were able to save them as well. however, when that pr got merged to main, I was no longer able to set an embargo at all in koppie. saving the embargo in app/models/hyrax/resource.rb#embargo= is something we committed at https://github.com/samvera/hyrax/pull/6098/commits/efffe9b5b8d2e07ec73f60895649ce6f24992b39, but we then removed it and opted to save the embargo in lib/hyrax/transactions/steps/save.rb#call. however, the `value` argument that gets passed to #embargo= does not have a valid id. just an empty string. so we do not ever set or find the embargo. this change does at least work.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/resource.rb
+++ b/app/models/hyrax/resource.rb
@@ -106,7 +106,7 @@ module Hyrax
     def embargo=(value)
       raise TypeError "can't convert #{value.class} into Hyrax::Embargo" unless value.is_a? Hyrax::Embargo
 
-      @embargo = value
+      @embargo = Hyrax.persister.save(resource: value)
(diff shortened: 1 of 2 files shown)


alishaevn:
closing in favor of #6118

PR #6121, opened 2023-07-25
stop-testing-middle-ruby-versions

alishaevn:
### Summary
stop testing against ruby3-0 and ruby3-1. these were already not required.

this work was originally done by @dlpierce in #6105. that pr however also updates the circle ci versions and is failing. I expect those are legitimate failures, but do not want to hold up removing these testing versions in the meantime. removing these versions will allow our spec suite to move faster and provide a shorter pathway to a green spec suite.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -214,42 +214,6 @@ workflows:
           hyrax_valkyrie: "true"
           requires:
             - build
-  ruby3-1:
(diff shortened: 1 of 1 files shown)


alishaevn:
closing because #6105 passed when I merged in `main` today.

PR #6122, opened 2023-07-25
Update show.json.jbuilder

cjcolvar:
Test skipping converting Hyrax::Resource objects back to ActiveFedora::Base objects because both support class.fields.


rodyoukai:
Hi @cjcolvar, thanks for the test

The commented line:

# @curation_concern = Wings::ActiveFedoraConverter.convert(resource: @curation_concern)

Caused a significant delay in generating the json response, so commenting it out considerably improved response times when consuming json from a SPA, which would be my case.

cjcolvar:
There were missing keys from the generated JSON so the original reason for the conversion seems valid.  I'm not sure what keys are dropped and if they're important though.

Issue #6126, opened 2023-07-27
cannot remove lease from work


alishaevn:
### Descriptive summary
ref:
- #5844 

I'll fill more of this out later. while doing the work for #5844, @sephirothkod and I realized that leases aren't working either. since the two features are opposite sides of the same coin, we decided to fix them as well. this ticket was created to track the lease work separately from embargoes.

### Steps to reproduce the behavior in User Interface (UI)

1. Do this
2. Then do this...
3. Include note here if not testable in UI

### Actual behavior (include screenshots if available)

Include what version of Hyrax relates to this issue (3.x, 4.x, main branch, etc.) if appropriate, and any relevant error messages/tracebacks if you're reporting a bug.

### Acceptance Criteria/Expected Behavior

- [ ] Create checkbox list of what done looks like...

### Rationale (for feature request only)

Provide the rationale or user story that describes "why" this issue should be addressed. Especially if this is a new feature or significant change to the existing implementation.

### Related work

Link to related issues or prior related work here.

rjkati:
Confirmed that removing leases is working on pg.nurax

PR #6127, opened 2023-07-28
6126 deactivated leases

alishaevn:
### Fixes
- #6126

### Summary
fix deactivating leases in koppie

* create a work, or update a work, to have a lease in the future
* go to the "/leases" endpoint
* click the name of the work
* select "deactivate"
* be taken back to the work show page and notice the lease is removed

### Detailed Description
while working on #6098, braydon and I began updating some of the lease code to follow the pattern of the embargo code we were implementing. however, actually fixing the ability to deactivate leases was out of scope for #5844, so I created a ticket and the work was moved here.

#### some other things to note:
- the lease is no longer saved on the work. it us referenced using a `lease_id` attrubute
- adding a file to a work with a lease, adds a lease to the file
- removing the lease from the work does not remove the lease from the file set even if you tell it to. this is the same functionality with dassie on main though, so I did not fix it
  - on `main` however, we're able to go to "/leases" and see the file set. then click on it and deactivate it
  - in valkyrie "/leases" shows that there's a fileset with a lease on it...but the fileset doesn't have "title" metadata so we can't click into it to deactivate it. fixing fileset metadata felt out of scope so I didn't address it. (you could manually get there by going to "/leases/<fileset_id>/edit" though)

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -71,7 +71,8 @@ module Hyrax
       def save(env, use_valkyrie: false)
         return env.curation_concern.save unless use_valkyrie
 
-        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource)
+        # don't run validations again on the converted object if they've already passed
+        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource, valid: env.curation_concern.save)
 
         # we need to manually set the id and reload, because the actor stack requires
         # `env.curation_concern` to be the exact same instance throughout.
@@ -115,9 +116,9 @@ module Hyrax
         attributes.select { |_, v| v.respond_to?(:select) && !v.respond_to?(:read) }
       end
 
-      def valkyrie_save(resource:)
+      def valkyrie_save(resource:, valid:)
         permissions = resource.permission_manager.acl.permissions
-        resource    = Hyrax.persister.save(resource: resource)
+        resource    = Hyrax.persister.save(resource: resource, valid: valid)
 
         resource.permission_manager.acl.permissions = permissions
         resource.permission_manager.acl.save
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -72,7 +72,7 @@ module Hyrax
         return env.curation_concern.save unless use_valkyrie
 
         # don't run validations again on the converted object if they've already passed
-        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource, perform_af_validation: !env.curation_concern.save)
+        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource, is_valid: env.curation_concern.save)
 
         # we need to manually set the id and reload, because the actor stack requires
         # `env.curation_concern` to be the exact same instance throughout.
@@ -116,9 +116,9 @@ module Hyrax
         attributes.select { |_, v| v.respond_to?(:select) && !v.respond_to?(:read) }
       end
 
-      def valkyrie_save(resource:, valid:)
+      def valkyrie_save(resource:, is_valid:)
         permissions = resource.permission_manager.acl.permissions
-        resource    = Hyrax.persister.save(resource: resource, valid: valid)
+        resource    = Hyrax.persister.save(resource: resource, perform_af_validation: !is_valid)
 
         resource.permission_manager.acl.permissions = permissions
         resource.permission_manager.acl.save
--- a/app/actors/hyrax/actors/base_actor.rb
+++ b/app/actors/hyrax/actors/base_actor.rb
@@ -72,7 +72,7 @@ module Hyrax
         return env.curation_concern.save unless use_valkyrie
 
         # don't run validations again on the converted object if they've already passed
-        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource, valid: env.curation_concern.save)
+        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource, perform_af_validation: !env.curation_concern.save)
 
         # we need to manually set the id and reload, because the actor stack requires
         # `env.curation_concern` to be the exact same instance throughout.
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -182,13 +182,15 @@ module Hyrax
     # @return [#errors]
     def create_valkyrie_work
       form = build_form
-      return after_create_error(form_err_msg(form)) unless form.validate(params[hash_key_for_curation_concern])
+      # fallback to an empty array to avoid: # NoMethodError: undefined method `has_key?' for nil:NilClass
+      original_input_params_for_form = params[hash_key_for_curation_concern] ? params[hash_key_for_curation_concern] : {}
+      return after_create_error(form_err_msg(form), original_input_params_for_form) unless form.validate(original_input_params_for_form)
 
       result =
         transactions['change_set.create_work']
         .with_step_args(
           'work_resource.add_to_parent' => { parent_id: params[:parent_id], user: current_user },
-          'work_resource.add_file_sets' => { uploaded_files: uploaded_files, file_set_params: params[hash_key_for_curation_concern][:file_set] },
+          'work_resource.add_file_sets' => { uploaded_files: uploaded_files, file_set_params: original_input_params_for_form[:file_set] },
           'change_set.set_user_as_depositor' => { user: current_user },
           'work_resource.change_depositor' => { user: ::User.find_by_user_key(form.on_behalf_of) },
           'work_resource.save_acl' => { permissions_params: form.input_params["permissions"] }
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -180,6 +180,7 @@ module Hyrax
 
     ##
     # @return [#errors]
+    # rubocop:disable Metrics/MethodLength
     def create_valkyrie_work
       form = build_form
       # fallback to an empty array to avoid: # NoMethodError: undefined method `has_key?' for nil:NilClass
@@ -199,6 +200,7 @@ module Hyrax
       @curation_concern = result.value_or { return after_create_error(transaction_err_msg(result)) }
       after_create_response
     end
+    # rubocop:enable Metrics/MethodLength
 
     def update_valkyrie_work
       form = build_form
--- a/app/controllers/concerns/hyrax/works_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/works_controller_behavior.rb
@@ -183,7 +183,7 @@ module Hyrax
     # rubocop:disable Metrics/MethodLength
     def create_valkyrie_work
       form = build_form
-      # fallback to an empty array to avoid: # NoMethodError: undefined method `has_key?' for nil:NilClass
+      # fallback to an empty hash to avoid: # NoMethodError: undefined method `has_key?` for nil:NilClass
       original_input_params_for_form = params[hash_key_for_curation_concern] ? params[hash_key_for_curation_concern] : {}
       return after_create_error(form_err_msg(form), original_input_params_for_form) unless form.validate(original_input_params_for_form)
 
--- a/app/services/hyrax/lease_manager.rb
+++ b/app/services/hyrax/lease_manager.rb
@@ -130,7 +130,6 @@ module Hyrax
     def nullify(force: false)
       return false if !force && under_lease?
 
-      # TODO(alishaevn): why isn't this working?
       lease.lease_expiration_date = nil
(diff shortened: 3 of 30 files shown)


app/controllers/concerns/hyrax/works_controller_behavior.rb
```diff
@@ -180,15 +180,18 @@ def actor
 
     ##
     # @return [#errors]
+    # rubocop:disable Metrics/MethodLength
     def create_valkyrie_work
       form = build_form
-      return after_create_error(form_err_msg(form)) unless form.validate(params[hash_key_for_curation_concern])
+      # fallback to an empty array to avoid: # NoMethodError: undefined method `has_key?' for nil:NilClass
+      original_input_params_for_form = params[hash_key_for_curation_concern] ? params[hash_key_for_curation_concern] : {}
```
alishaevn, line 187: adjusted this error handling so that `generic_works_controller_json_spec.rb:58` passes. this is also how the #create method in this same file handles the params

lib/wings/valkyrie/persister.rb
```diff
@@ -23,15 +23,16 @@ def initialize(adapter:)
       # Persists a resource using ActiveFedora
       # @param [Valkyrie::Resource] resource
       # @return [Valkyrie::Resource] the persisted/updated resource
-      def save(resource:)
+      def save(resource:, valid: true)
```
alishaevn, line 26: by default we will not validate the converted `af_object` because validations should have been run on the Valkyrie created object. 

however, there are situations where we will have created an ActiveFedora object while `Hyrax.config.use_valkyrie?` is TRUE. (see: spec/controllers/hyrax/generic_works_controller_json_spec.rb:59)

app/actors/hyrax/actors/base_actor.rb
```diff
@@ -71,7 +71,8 @@ def apply_deposit_date(env)
       def save(env, use_valkyrie: false)
         return env.curation_concern.save unless use_valkyrie
 
-        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource)
+        # don't run validations again on the converted object if they've already passed
+        resource = valkyrie_save(resource: env.curation_concern.valkyrie_resource, valid: env.curation_concern.save)
```
alishaevn, line 75: in the event `env.curation_concern` is an AF object, BUT `Hyrax.config.use_valkyrie?` is TRUE...we need the validation to run in `Wings::Valkyrie::Persister#save` because they were NOT run on the valkyrie object. otherwise, no errors are returned although they should be.

in this case, `env.curation_concern.save` will equal false.

lib/hyrax/transactions/steps/save.rb
```diff
@@ -53,10 +56,54 @@ def call(change_set, user: nil)
...
+
+        # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+        def add_lease_to_members(change_set)
+          # TODO: account for all members and levels, ref: #6131
+          # use recursion to ensure the change_set that applies to the member is getting saved and published/indexed
+          change_set.model.member_ids&.each do |member|
+            item = Hyrax.query_service.find_by(id: member.id)
+            next unless item.is_a? Hyrax::FileSet
+
+            lease_updates = change_set.lease.instance_variable_get(:@_changes).values.select { |v| v }.any?
```
alishaevn, line 86:
``` bash
# example of what change_set.lease.instance_variable_get(:@_changes) could look like
@_changes={"visibility_after_lease"=>false, "visibility_during_lease"=>false, "lease_expiration_date"=>true}
```

lib/wings/active_fedora_converter.rb
```diff
@@ -141,11 +142,19 @@ def apply_attributes_to_model(af_object)
         members = Array.wrap(converted_attrs.delete(:members))
         files = converted_attrs.delete(:files)
         af_object.attributes = converted_attrs
+        if resource.try(:lease) && af_object.reflections.include?(:lease)
+          # af_object.lease.class has the same name as resource.lease.class; however, each class has a different object_id
+          # so a type mismatch happens. the code below coerces the one object into the other
+          # TODO(alishaevn): fix this!
```
orangewolf, line 148: is this TODO still accurate?

lib/wings/active_fedora_converter.rb
```diff
@@ -141,11 +142,19 @@ def apply_attributes_to_model(af_object)
         members = Array.wrap(converted_attrs.delete(:members))
         files = converted_attrs.delete(:files)
         af_object.attributes = converted_attrs
+        if resource.try(:lease) && af_object.reflections.include?(:lease)
+          # NOTE: af_object.lease.class has the same name as resource.lease.class; however, each class has a different object_id
```
jeremyf, line 146: Ah yes...antics.  I remember these.  There's definitely a subtle issue elsewhere that is beyond the scope of this work.

lib/wings/active_fedora_converter.rb
alishaevn: - #6134

lib/wings/valkyrie/persister.rb
jeremyf: Hmm…thinking about this parameter name `valid`?  It is a bit confusing.

Does it mean "Treat this record as valid?"  Or does it mean "validate this record".  Looking at how it is later used, it looks like it could be `perform_af_validation: false`

jeremyf (dismissed):

lib/wings/active_fedora_converter.rb
bkiahstroud: Personally, I've started using `TODO`s in my WIP PRs for pre-merge reminders. If something is going to get merged in a state that needs to be addressed later, I've started using `FIXME`. I only ever use my GitHub username when leaving `NOTE`s so it's clear who the original author was. 

This flow has been working for me and is by no means convention, just a suggestion.

alishaevn: mmmm. I do like that.
`TODO` has always been a tricky thing for me because I can't know for sure how it's interpreted. I put my name next to it for the same reason you put yours next to `NOTE`. but I suppose to some that may infer that I'm the one who's gonna fix it. which isn't always my intention. lol. nor is it always the intention that it's gonna be fixed in the present pr. (as evidenced by all the TODO's even in this repo. 😅 )

I like sticking with the `TODO` convention because it's one that I think the majority of people are familiar with. but maybe `TODO` means to handle this thing later while `TODO(name)` means **I'm** going to handle this thing before the pr is ready??

🤔

bkiahstroud: Might be worth revisiting with the wider team. I simply go with what makes the most sense to me on a case-by-case basis, but things that make the most sense to me often don't make any sense to others 😅

spec/services/hyrax/admin_set_create_service_spec.rb
```diff
@@ -292,8 +292,7 @@
         let(:admin_set) { FactoryBot.build(:invalid_hyrax_admin_set) } # Missing title
 
         it 'will not call the workflow_importer' do
-          expect { service.create! }.to raise_error(RuntimeError)
-          expect(workflow_importer).not_to have_received(:call)
+          expect { service.create! }.to raise_error(ActiveRecord::RecordNotFound)
```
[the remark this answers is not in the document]
alishaevn: I removed it because at the time I was testing it...it was no longer valid. related to the note below.

if I recall correctly...since we weren't validating in `Hyrax::Persister#save`, any longer, we did not get the runtime error anymore from `Hyrax::AdminSetCreateService#create!`. therefore the workflow importer _was_ called. later down the stacktrace in `Sipity::Workflow#activate!`, `Sipity::Workflow.find_by!` failed because we didn't have a `workflow_id`. that caused the `ActiveRecord::RecordNotFound` we're now checking for.


when I spoke to Rob about it, the different error and change in the `save` method, he said to change the error we were now expecting. I did not however explain the full background of the spec with him.

now that I'm passing a `perform_af_validation` param to `save` though...I can pass `perform_af_validation: true` inside `Hyrax::AdminSetCreateService#create!`.
hopefully that doesn't break another spec. 😩 🤞🏾 

<details>
<summary>note</summary>

Tuesday, August 1, 2023
while pairing last week, rob suggested [this change](https://github.com/samvera/hyrax/pull/6127/commits/9b64c54761216dcef982e2157539638662faa90b). I noticed last night that committing that change made the [7 failing works_controller_behavior_spec’s](https://app.circleci.com/pipelines/github/samvera/hyrax/9332/workflows/eb885c03-e719-41ba-8d71-855a3e8c35de/jobs/72774/tests), pass.
a little bit ago however, I noticed that the change caused spec/services/hyrax/admin_set_create_service_spec.rb:294 to fail.
I’m now unsure whether I should or should not use validate: false in [Wings::Valkyrie::Persister#save](https://github.com/samvera/hyrax/blob/9b64c54761216dcef982e2157539638662faa90b/lib/wings/valkyrie/persister.rb#L26). or the best practice way forward on making sure that the specs in both files pass…and that we’re doing the right thing.
</details>

spec/services/hyrax/lease_manager_spec.rb
```diff
@@ -90,7 +90,7 @@
         expect(manager.lease)
           .to have_attributes visibility_after_lease: 'authenticated',
                               visibility_during_lease: 'open',
-                              lease_expiration_date: an_instance_of(Date),
+                              lease_expiration_date: an_instance_of(DateTime),
```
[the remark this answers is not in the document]
alishaevn: I changed it because on `main`, when you add a lease to a work, it's saved as a DateTime. so I made koppie do the same. idk how the tests weren't failing before. 🤷🏾‍♀️ 

``` bash
[1] pry(main)> ActiveFedora::Base.find('dr26xx36k').lease
=> #<Hydra::AccessControls::Lease id: "0db70bc6-1bd0-4b95-a991-9b4e275b6f33", visibility_during_lease: "open", visibility_after_lease: "restricted", lease_expiration_date: "2023-08-31 00:00:00", lease_history: []>

[2] pry(main)> ActiveFedora::Base.find('dr26xx36k').lease.lease_expiration_date.class
=> DateTime
```

spec/wings/active_fedora_converter_spec.rb
```diff
@@ -362,7 +363,9 @@ class Resource < Valkyrie::Resource
       let(:work) { FactoryBot.create(:leased_work) }
 
       it 'repopulates the lease' do
-        expect(converter.convert).to have_attributes(lease_id: work.lease_id)
+        # the lease does get recreated on the converted item, but with a new id
```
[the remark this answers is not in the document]
alishaevn: I went back to test this and was able to revert the test to what it already was. 🤷🏾‍♀️ 

ref: https://github.com/samvera/hyrax/pull/6127/commits/580d817e4f976bedbbd1e9e8212abe1b6d2bff46

lib/hyrax/transactions/steps/save.rb
alishaevn: I moved this method into the LeaseManager. ref: https://github.com/samvera/hyrax/pull/6127/commits/b2a9f2303dd67e4f0a4b560da82f84d574fd84fd

lib/hyrax/transactions/steps/add_file_sets.rb
```diff
@@ -25,6 +25,14 @@ def initialize(handler: Hyrax::WorkUploadsHandler)
         # @return [Dry::Monads::Result]
         def call(obj, uploaded_files: [], file_set_params: [])
           if @handler.new(work: obj).add(files: uploaded_files, file_set_params: file_set_params).attach
+            if obj.lease
```
alishaevn, line 28: a new file set gets added to the work in this `#call` method, which happens _after_ "steps/save.rb". doing this call here correctly puts the lease on new files and existing ones.

app/services/hyrax/lease_manager.rb
```diff
@@ -47,6 +62,51 @@ def release_lease_for!(resource:, query_service: Hyrax.query_service)
         new(resource: resource, query_service: query_service)
           .release!
       end
+
+      # Creates or updates an existing lease on a member to match the lease on the parent work
+      # @param [Array<Valkyrie::Resource>] members
+      # @param [Hyrax::Work] work
+      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      def create_or_update_lease_on_members(members, work)
```
alishaevn, line 70: putting this method in the LeaseManger feels like a better place for it to be than in "transactions/save.rb". it also makes it easier to use when #6131 gets worked on.

orangewolf (approved):

spec/services/hyrax/admin_set_create_service_spec.rb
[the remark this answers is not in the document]
alishaevn: my proposed change broke several things because `Hyrax.persister.save` that gets called inside of `Hyrax::AdminSetCreateService#create!` is not the same as `Wings::Persister#save`. I've reverted my change and updated the assertion.

ref: https://github.com/samvera/hyrax/pull/6127/commits/f992e109d1903a48407a15840f45f9bd2ecef3f2

orangewolf (approved):

PR #6140, opened 2023-08-08
♻️ Adding indices to Hyrax::CounterMetrics

jeremyf:
Without these indices, what could be a large database could result in significant performance degradations at query time.  Each of these fields is very likely to be part of a "WHERE" statement.

Related to:

- https://github.com/samvera/hyrax/pull/6130
- https://github.com/scientist-softserv/palni-palci/issues/616WIP


no-reply (approved):

no-reply:
cool!

Issue #5848, opened 2022-09-09
Managers of admin sets cannot view embargoed works 


rjkati:
### Descriptive summary

When testing Hyrax 4.0.0.beta1 locally using Koppie, manager accounts cannot view embargoed works deposited in the admin set that they manage.

### Rationale

Managers of admin sets should be able to view all works deposited in the admin set

### Expected behavior

Managers of admin sets should see work show page and files for embargoed works

### Actual behavior
When a manager navigates to an embargoed work within their managed admin set, they see an "Unauthorized" message

### Steps to reproduce the behavior

Prerequisite: Make sure you have created two non-admin regular user accounts.

1. As an administrator, create an admin set and set one user account as a manager.
2. Log in as the regular user and deposit a work with an embargo.
3. Log in as the manager account from step 1 and attempt to view the embargoed work which you deposited in step 2

alishaevn:
**(testing to see if the work done for #5844 fixed this problem too)**
<hr>

alpha: admin user
beta: regular user
gamma: regular user

as alpha I created an admin set and made beta to be a manager. I then logged out and back in as gamma. I created an embargoed work, but was unable to add it to the admin set. the admin set didn't even show up.

I logged back in as alpha and saw that the collection was private. I tried to set the admin set to public, but that isn't an option. I did a search in the samvera slack and discovered that admin sets can only be seen by other admins by default. (ref: [here](https://samvera.slack.com/archives/C3QESH5FG/p1506356456000477) and [here](https://samvera.slack.com/archives/C0F9JQJDQ/p1506357835000146)) instead, on the "participants" tab of the edit screen I clicked "allow all registered users to deposit". logged back in as gamma and was able to add an embargoed work to the admin set I created.

I logged in as beta and when I viewed the managed collections page, it showed that the admin set had 0 items. clicking the name of the admin set however took me to the dashboard view and that said there was 1 item. under the "works in this set" heading however, it said "The collection is either empty or does not contain items to which you have access."

logged back in as alpha, I was able to see the work that was created by gamma.

this is still an issue.

dunn:
looks like `edit_access_person_ssim` on an adminset isn't honored when the permissions on the object are queried directly
```
[10] pry(main)> manager.can? :edit, work
[...]
[...]

Issue #6143, opened 2023-08-08
revoking manager permissions from a user does not remove object access


dunn:
### Descriptive summary

While testing https://github.com/samvera/hyrax/issues/5848 I noticed that non-Valkyrie Hyrax has a different problem: if a user's manager access of an AdminSet is revoked, they still have edit access on the objects in that AdminSet.

### Steps to reproduce the behavior in User Interface (UI)

Follow the same initial steps as in #5848:

1. As an administrator, create an AdminSet and set a user account as a manager.
2. Log in as a regular user and deposit a work with an embargo.
3. Log in as the manager account from step 1 and attempt to view the embargoed work which you deposited in step 2

In Dassie, the manger account will be able to (correctly) view the object.  However,

4. As the admin user, remove the manager account from the list of managers in the AdminSet
5. As the ex-manager user, attempt to view the embargoed work.

### Actual behavior (include screenshots if available)

You will be able to view the embargoed work, even though you can no longer access the AdminSet itself.

Tested on current main branch (c3eb490d66e9904745d447e7bf752cb51653ca9b) in Dassie.

### Acceptance Criteria/Expected Behavior

- [ ] Users' object access permissions are updated when their AdminSet roles are removed.

no-reply:
i'd like to move the discussion from https://github.com/samvera/hyrax/pull/6145#issuecomment-1672410956 over here.

PR #6145, opened 2023-08-09
add PermissionTemplate application to the `work_create` transaction 

no-reply:
users in the PermissionTemplate for the adminset/workflow should get edit access to new objects added to the admin set. test that at the controller level.

to achieve this:

when we save a resource, we should run the `PermissionTemplateApplicator`against it, to apply the template matching the resource's `AdministrativeSet`.

in the legacy Actor Stack, this step happens before the object is first saved. in the Valkyrie model, we treat the ACL as a separate resource which can (and has to) be saved separately, so it's convenient to first save the object first, and then check for permission template application.

in the old model, we fail to save the work if a permission template doesn'texist for the admin set. here, if we're missing an admin set or it is missing a permission template, we simply decline to apply and always succeed (except in cases of unhandled exceptions).

this was initially implemented as part of transactions in 05b66b90aaf, but removed when shifting to Valkyrie because it wasn't tested anywhere. we couldn't figure out what the desired behavior was, so we left it out of the new implementation.

related to #5848.

* adds a failing test at the controller layer for the behavior desired in #5848 
* reinstates an implementation of a transaction step to apply a permission template, similar to the one developed for 05b66b90aaf

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/container.rb
+++ b/lib/hyrax/transactions/container.rb
@@ -31,6 +31,7 @@ module Hyrax
       require 'hyrax/transactions/steps/add_to_collections'
       require 'hyrax/transactions/steps/add_to_parent'
       require 'hyrax/transactions/steps/apply_collection_type_permissions'
+      require 'hyrax/transactions/steps/apply_permission_template'
       require 'hyrax/transactions/steps/change_depositor'
       require 'hyrax/transactions/steps/check_for_empty_admin_set'
       require 'hyrax/transactions/steps/delete_access_control'
@@ -201,6 +202,10 @@ module Hyrax
 
         ops.register 'add_to_parent' do
           Steps::AddToParent.new
+        end
+
+        ops.register 'apply_permission_template' do
+          Steps::ApplyPermissionTemplate.new
         end
 
         ops.register 'change_depositor' do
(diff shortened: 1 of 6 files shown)


marrus-sh:
am i understanding correctly that:

- people added to administrative sets do not get access to existing works
- people removed from administrative sets do not lose access to works they previously had access to (#6143)

? this seems to match the behaviour of dassie, so in terms of having feature parity i think it’s a good thing, but i’m also a little bit 🤔 regarding what we actually want the permissions model here to look like. i tried following the steps in #5848 and this did solve the problem.

how does this interact with leases? might they revoke access unnecessarily to users in the adminset after the lease expires? (wouldn’t surprise me if this was also a problem in dassie, though…)

marrus-sh:
Actually, I’m not sure this solves the entire issue. I was able to see the deposited work in the dashboard (under “Managed Works”) of the second user, so I assumed that they had permissions, but when I actually clicked through I got an “Unauthorized—The page you have tried to access is private” error. And the work does not show up when navigating to the collection in “Managed Collections”. Curious that it appears in the other place then?

It’s possible that I need to test this again on a fresh(er) Hyrax build; I wasn’t able to build from scratch today due to internet outages.

marrus-sh (approved):
Tested it doing everything in the right order today and it does appear to work :)

I agree with treating enhancements beyond just how dassie operates as a separate issue.

PR #6146, opened 2023-08-09
Remove `bin/rails` to fix nurax deploys

dlpierce:
### Fixes

Fixes nurax deployment

This was added in an attempt to inform cli tools about the hyrax engine, but it is causing errors when nurax deploys as well as when the test app is rebuilt in circleci.

* rm bin/rails
* bump .regen

@samvera/hyrax-code-reviewers


orangewolf (approved):
I do think we should be able to have both, but the regen is clearly the more important of the two. looks like you can't run enginecart inside of docker at all (it wants to call the app Dassie and I can't figure out why), so I'm trying to configure my laptop to build it. either way, lets merge this and I'll try to find a non-breaking solution.

orangewolf:
@dlpierce AHA! I think the issue isn't that we have a bin/rails command it is that we are loading all scripts in `bin` as executables. usually in gems executables go in a separate `exe` folder so that this doesn't happen. I'm excluding bin/rails from the gemspec now and that works locally. I'm hoping it will allow the enginecart to run in ci too

dlpierce:
Seems to work for nurax here!

PR #6148, opened 2023-08-10
Valkyrize batch editing

no-reply:
The PR is to support batch editing while valkyrie is enabled. This can be tested under dassie  (w/o valkyrie) and koppie.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/batch_edits_controller.rb
+++ b/app/controllers/hyrax/batch_edits_controller.rb
@@ -61,7 +61,7 @@ module Hyrax
     end
 
     def valkyrie_update_document(obj)
-      form = Hyrax::Forms::ResourceBatchEditForm.new(obj, current_ability, nil)
+      form = form_class.new(obj, current_ability, nil)
       return unless form.validate(params[form_class.model_class.model_name.param_key])
       result = transactions['change_set.update_work']
                .with_step_args('work_resource.save_acl' => { permissions_params: form.input_params["permissions"] })
@@ -121,7 +121,7 @@ module Hyrax
     end
 
     def work_params(extra_params = {})
-      work_params = params[form_class.model_name.param_key] || ActionController::Parameters.new
+      work_params = params[form_class.model_class.model_name.param_key] || ActionController::Parameters.new
       form_class.model_attributes(work_params.merge(extra_params))
     end
 
--- a/app/controllers/hyrax/batch_edits_controller.rb
+++ b/app/controllers/hyrax/batch_edits_controller.rb
@@ -63,6 +63,9 @@ module Hyrax
     def valkyrie_update_document(obj)
       form = form_class.new(obj, current_ability, nil)
       return unless form.validate(params[form_class.model_class.model_name.param_key])
+
+      cleanu_form_fields form
+
       result = transactions['change_set.update_work']
                .with_step_args('work_resource.save_acl' => { permissions_params: form.input_params["permissions"] })
                .call(form)
@@ -149,6 +152,16 @@ module Hyrax
         redirect_to hyrax.url_for(controller: params[:return_controller], only_path: true)
       else
         redirect_to hyrax.dashboard_path
+      end
+    end
+
+    # Clean up form fields
+    # @param form Hyrax::Froms::ResourceBatchEditForm
+    def cleanu_form_fields(form)
+      form.lease = nil if form.lease && form.lease.fields['lease_expiration_date'].nil?
+      form.embargo = nil if form.embargo && form.embargo.fields['embargo_release_date'].nil?
+      form.fields.keys.each do |k|
+        form.fields[k] = nil if form.fields[k].is_a?(Array) && form.fields[k].blank?
       end
     end
   end
--- a/app/controllers/hyrax/batch_edits_controller.rb
+++ b/app/controllers/hyrax/batch_edits_controller.rb
@@ -64,7 +64,7 @@ module Hyrax
       form = form_class.new(obj, current_ability, nil)
       return unless form.validate(params[form_class.model_class.model_name.param_key])
 
-      cleanu_form_fields form
+      cleanup_form_fields form
 
       result = transactions['change_set.update_work']
                .with_step_args('work_resource.save_acl' => { permissions_params: form.input_params["permissions"] })
@@ -157,7 +157,7 @@ module Hyrax
 
     # Clean up form fields
     # @param form Hyrax::Froms::ResourceBatchEditForm
-    def cleanu_form_fields(form)
+    def cleanup_form_fields(form)
       form.lease = nil if form.lease && form.lease.fields['lease_expiration_date'].nil?
       form.embargo = nil if form.embargo && form.embargo.fields['embargo_release_date'].nil?
       form.fields.keys.each do |k|
--- a/app/controllers/hyrax/batch_edits_controller.rb
+++ b/app/controllers/hyrax/batch_edits_controller.rb
@@ -60,11 +60,27 @@ module Hyrax
       obj.save
     end
 
+    def valkyrie_update_document(obj)
+      form = Hyrax::Forms::ResourceBatchEditForm.new(obj, current_ability, nil)
+      return unless form.validate(params[form_class.model_class.model_name.param_key])
+      result = transactions['change_set.update_work']
+               .with_step_args('work_resource.save_acl' => { permissions_params: form.input_params["permissions"] })
+               .call(form)
+      obj = result.value!
+
+      InheritPermissionsJob.perform_now(obj)
+      VisibilityCopyJob.perform_now(obj)
+    end
+
     def update
       case params["update_type"]
       when "update"
         batch.each do |doc_id|
-          update_document(Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: doc_id, use_valkyrie: false))
+          if Hyrax.config.use_valkyrie?
+            valkyrie_update_document(Hyrax.query_service.find_by(id: doc_id))
+          else
+            update_document(Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: doc_id, use_valkyrie: false))
+          end
         end
         flash[:notice] = "Batch update complete"
         after_update
@@ -97,7 +113,7 @@ module Hyrax
     end
 
     def form_class
-      Forms::BatchEditForm
+      Hyrax.config.use_valkyrie? ? Forms::ResourceBatchEditForm : Forms::BatchEditForm
     end
 
     def terms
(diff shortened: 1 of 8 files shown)


spec/support/features/batch_edit_actions.rb
```diff
@@ -15,10 +15,12 @@ def fill_in_batch_edit_fields_and_verify!
       fill_in "generic_work_#{field_id}", with: "NEW #{field_id}"
 
       find("##{field_id}_save").click
+
+      ajax_wait(15)
```
tpendragon, line 19: Aw, this is a bummer. Do we know if this is still true? I won't block on this, but maybe a ticket to investigate this?

app/controllers/hyrax/batch_edits_controller.rb
```diff
@@ -60,11 +60,30 @@ def update_document(obj)
...
+      return unless form.validate(params[form_class.model_class.model_name.param_key])
+
+      cleanup_form_fields form
+
+      result = transactions['change_set.update_work']
+               .with_step_args('work_resource.save_acl' => { permissions_params: form.input_params["permissions"] })
+               .call(form)
+      obj = result.value!
+
+      InheritPermissionsJob.perform_now(obj)
```
tpendragon, line 74: Question: Should this be in the transaction?

app/controllers/hyrax/batch_edits_controller.rb
```diff
@@ -60,11 +60,30 @@ def update_document(obj)
...
+      VisibilityCopyJob.perform_now(obj)
```
tpendragon, line 75: Question: Should this be in the transaction?

app/controllers/hyrax/batch_edits_controller.rb
```diff
@@ -135,5 +154,15 @@ def redirect_to_return_controller
         redirect_to hyrax.dashboard_path
       end
     end
+
+    # Clean up form fields
+    # @param form Hyrax::Froms::ResourceBatchEditForm
+    def cleanup_form_fields(form)
```
tpendragon, line 160: Suggestion: Can we get a comment about why this is necessary? Why doesn't the form handle this somewhere? Does this mean the form's not useful outside of this controller?

app/services/hyrax/visibility_propagator.rb
```diff
@@ -10,7 +10,7 @@ class VisibilityPropagator
     # @return [#propagate]
     def self.for(source:)
       case source
-      when Hyrax::WorkBehavior # ActiveFedora
+      when ActiveFedora::Base # ActiveFedora
```
tpendragon, line 13: Question: I missed it, where did this change?

lib/wings/active_fedora_converter/default_work.rb
```diff
@@ -137,6 +137,10 @@ def enforce_future_date_for_lease?
         false
       end
 
+      def file_sets
+        members.select(&:file_set?)
```
tpendragon, line 141: Question: Is there no query scoped to file set members? Or do we just expect members to always be loaded here in entirety so it's more efficient to just iterate over them all?

app/controllers/hyrax/batch_edits_controller.rb
```diff
@@ -97,15 +116,15 @@ def destroy_batch
     end
 
     def form_class
-      Forms::BatchEditForm
+      Hyrax.config.use_valkyrie? ? Forms::ResourceBatchEditForm : Forms::BatchEditForm
```
tpendragon, line 119: nitpick: How's Hyrax feel about ternaries over explicit branches? We use an if/else above, should we use one here?

app/forms/hyrax/forms/resource_batch_edit_form.rb
```diff
@@ -0,0 +1,90 @@
...
+         :keyword, :resource_type, :license, :publisher, :date_created,
+         :subject, :language, :identifier, :based_near,
+         :related_url]
+      end
+
+      attr_reader :batch_document_ids
+
+      # Returns a list of parameters we accept from the form
+      # rubocop:disable Metrics/MethodLength
+      def self.build_permitted_params
```
tpendragon, line 37: Question: I just don't know much about hyrax forms anymore - if there's form objects with a restricted subset of fields, why do they also need to define their permitted params? Where's this get used - do all the form objects have these or is it something special about bulk edit?

app/forms/hyrax/forms/resource_batch_edit_form.rb
```diff
@@ -0,0 +1,90 @@
...
+      def column_for_attribute(name)
+        name
+      end
+
+      private
+
+      # override this method if you need to initialize more complex RDF assertions (b-nodes)
+      # @return [Hash<String, Array>] the list of unique values per field
+      def initialize_combined_fields
+        # For each of the files in the batch, set the attributes to be the concatenation of all the attributes
```
tpendragon, line 78: Maybe a little comment about why I'd want it to do this?

app/controllers/hyrax/batch_edits_controller.rb
no-reply: i'm also curious about this.
it does seem like the form should handle these transformations.

app/controllers/hyrax/batch_edits_controller.rb
no-reply: if it's not being flagged by rubocop, i think we don't block on style.

on the other hand, i'm wondering if we could just drop all the old handling? if the valkyrie code path is working, do we need the opt-in?

lib/wings/active_fedora_converter/default_work.rb
no-reply: this is just a port of how `Hydra::Works::WorkBehavior` implements this: https://github.com/samvera/hydra-works/blob/c9b9dd0cf11de671920ba0a7161db68ccf9b7f6d/lib/hydra/works/models/concerns/work_behavior.rb#L90-L92

app/forms/hyrax/forms/resource_batch_edit_form.rb
no-reply: @tpendragon you probably know more about this form setup than most of us. these are valkyrie `ChangeSet`s.

i have the same question about why this is needed, and where it's used other than the tests below.

app/forms/hyrax/forms/resource_batch_edit_form.rb
```diff
@@ -0,0 +1,90 @@
...
+        @names = []
+        @batch_document_ids = batch_document_ids
+        if @batch_document_ids.present?
+          super(model.class.new(initialize_combined_fields))
+        else
+          super(model)
+        end
+      end
+
+      def terms
```
no-reply, line 26: why is it okay to hard code these values? shouldn't they be derived from configuration?

app/controllers/hyrax/batch_edits_controller.rb
no-reply: we don't seem to enqueue these two jobs in a normal `update_work` code path. (should we?)

if we need to, i think it wouldn't be in the Transaction, but in an event listener. i.e. should we be triggering these changes whenever `object.acl.updated` gets called? or, if not, why are they needed here?

app/controllers/hyrax/batch_edits_controller.rb
lsitu: I am not sure what is the best way to handle it. The another way that we can try is to merge the submitted params to the resource attributes in [form.validate(params)](https://github.com/samvera/hyrax/pull/6148/files#diff-7eea01c84c5eba32121b213d71b5dd9f86b7968181dd4a273348d42e00463bd3R65). Please advice.

app/services/hyrax/visibility_propagator.rb
lsitu: The reason is that the monograph https://github.com/samvera/hyrax/blob/main/.dassie/app/models/monograph.rb model won't include Hyrax::WorkBehavior, which will cause tests related to visibility changes failed for fileset visibility propagating.

spec/support/features/batch_edit_actions.rb
lsitu: Yes in dassie with use_valkyrie=true. Integration tests like https://github.com/samvera/hyrax/pull/6148/files#diff-7eea01c84c5eba32121b213d71b5dd9f86b7968181dd4a273348d42e00463bd3R65 will failed with the original 5 seconds. A think this is dynamic wait time that could exit the loop when ajax call response back, which could be a better pattern than the code in https://github.com/samvera/hyrax/pull/6148/files#diff-5522864e208680e8d4491bd348df8c440287c7b89c1efd4e36c4d97b316300c6L21 that waits for constant 5 seconds for every test.

app/services/hyrax/visibility_propagator.rb
no-reply: ah, the generated model doesn't include `WorkBehavior`. it seems like the weird thing here is that the `VisibilityCopyJob` is casting the valkyrie resource back to an AF::Base, even when `use_valkyrie` is `true`(?)

i think in the ideal case, both of the following would be true:
  - `VisibilityCopyJob` should support works generated by Wings; being generous about the type check as proposed here seems good.
  - the `use_valkyrie` code path here would leverage valkyrie objects throughout.

i think it would be worth checking up about the second point as part of this work.

app/controllers/hyrax/batch_edits_controller.rb
no-reply: looking closer, i suspect the answer "because that's what the AF implementation does" should be good enough for us here.

app/forms/hyrax/forms/resource_batch_edit_form.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+module Hyrax
+  module Forms
+    class ResourceBatchEditForm < Hyrax::Forms::ResourceForm
+      include Hyrax::FormFields(:basic_metadata)
+
+      self.required_fields = []
+      self.model_class = Hyrax.primary_work_type
```
no-reply, line 8: does this form only support the primary work type?

lsitu: No. It'll work for any models, even with a batch  that mixes with different objects types. The model_class here is only served for retrieving a consistent `param_key` for different object types.

app/controllers/hyrax/batch_edits_controller.rb
tpendragon: Action item: Fix this, I can take a look at it

app/forms/hyrax/forms/resource_batch_edit_form.rb
tpendragon: Action item: See if we can use the model configurable fields. Pull @no-reply in if we get stuck with this.

tpendragon (approved):
We're going to merge this because it works and we're going to do an RC, and this feature hasn't worked really well in the past. We'll write a ticket for the adjustments we might want to do here.

PR #6152, opened 2023-08-15
Fix IIIF manifests when the presenter model is a SolrDocument

dlpierce:
### Summary

hydra_model is already a class, don't call `class` on it

* Manifests in koppie or Nurax PG include a sequence section with URLS for the IIIF endpoint.
* or
* The universal viewer displays the work's images.

### Detailed Description

Fix IIIF manifests when the presenter model is a SolrDocument. `hydra_model` already returns a class. Calling `class` on it breaks the Hyrax::Resource subclass determination causing the wrong code path to be used resulting in a manifest without any images.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/iiif_manifest_presenter.rb
+++ b/app/presenters/hyrax/iiif_manifest_presenter.rb
@@ -93,8 +93,8 @@ module Hyrax
     ##
     # @return [Array<#to_s>]
     def member_ids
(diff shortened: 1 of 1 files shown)


orangewolf (approved):

orangewolf:
looks great

PR #6160, opened 2023-08-22
Fits Servlet is a Substantial Speed Bump

orangewolf:
### Summary

Running fits as a servlet reduces fits processing time by about 4x.  This is because the majority of a given fits.sh run is loading Java classes and this gets done once at startup in the servlet. 

This code allow fits servlet as a option in Hyrax and support it in Dassie, Koppie and the helm chart.

* Spin up dassie
* See that fits service is running in docker
* Create a new work and attach a file with it
* See the fits service was called in the logs and file was successfully charaterized
* repeat for koppie

### Detailed Description

Fits is slow. It can take minutes if in properly configured (looking at you tika) and multiple seconds even with the number of processors paired down. It spikes both memory and cpu quite a bit. Running as a servlet prevents the cpu and memory spikes and lets you run more characterize jobs faster. It is my opinion that this should be the default in a dockerized world. 

* Add config value to Hyrax to support the possibility of changing the Characterize processor
* Add fits servlet as the default choice to Dassie
* Add fits servlet as the default choice to Koppie
* Add fits servlet as an option that can be enabled in the helm chart.

@samvera/hyrax-code-reviewers


orangewolf:
I'm short on time that I can spend on this, but it was a huge when for several large imports so I wanted to get the message out to the community asap.

dunn (commented):
tested successfully on koppie

chart/hyrax/README.md
```diff
@@ -89,6 +89,17 @@ To use an existing or external `fcrepo` instance, use the chart values:
...
+
+# Optional Fits Service
+
+Running fits as a servlet will result in much faster fits responses. This is because the loading cost of the
+various Java components far out weighs the processing time of those components.
+
+To enable fits as a service:
+
+  - `fits.enabled`: true
+  - make sure config/initializers/hyrax.rb has `config.characterization_options = { ch12n_tool: :fits_servlet }`
```
no-reply, line 101: i wonder if it's possible to make put this directly into the helm chart? e.g. could `characterization_options` default to this value if the environment variable is set to something remote?

i'm not willing to block on this though. seems okay to just add a note about `config.characterization_options` to whatever the relevant release notes are.

dunn: https://github.com/samvera/hyrax/pull/6160/commits/20c7ae77810b2a354fcb5ab1d6dc8b1f4ae614b2

dlpierce (dismissed):
This is great! I would like to see support for using a shared volume to avoid large file uploads, but that likely needs an enhancement to hydra-file_characterization.

> Finally, the image exposes a volume, /processing, that can be attached to local storage. Then, you can instruct FITS to analyze files in the volume by executing a query like http://localhost:8080/fits/examine?file=/processing/[PATH_TO_FILE].

dlpierce:
Also, this fitsservelet_container image has fallen behind on FITS releases. There were several security patches after 1.5.0, not to mention 1.6.0. We should investigate how updated images can be made readily available.

orangewolf:
@dlpierce I agree about the versions. I feel like increased usage may be just what the doctor ordered to get that moving again. As for the local file param I am starting with a tactic of saying "boy it sure would be nice if" and hoping someone with Java experience that's less than 15 years old does it before I have to figure it out myself.

chart/hyrax/templates/_helpers.tpl
```diff
@@ -86,6 +86,23 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- end }}
 {{- end -}}
 
+
+{{/*
+Create default fully qualified service names.
```
dunn, line 91: should these be used below in the templates?

orangewolf: hyrax.fits.fullname is uses in fits.host. fits.host is used in configmap-env.yaml line 38

dunn: ohh, i see.

dunn (approved):

PR #6172, opened 2023-08-23
use Valkryie for all test setup in edit_permissions_service_spec

no-reply:
makes this test compatible with Valkyrie by refactoring out ActiveFedora based
factories.

the methodology i used here is to:

  - stand up dassie;
  - refactor for tidiness/to remove AF based mocks;
  - rewrite with Valkyrie factories;
  - test on dassie;
  - test on koppie

@samvera/hyrax-code-reviewers


dlpierce (approved):
Great example of converting a spec

Issue #6137, opened 2023-08-07
Come up with a simple strategy for testing multiple enviornments in test suite


no-reply:
### Descriptive summary

see notes: 

https://docs.google.com/document/d/1Uc--AJ5DdAl3o41JsPsZETH2lp-JqHoT5_ueNPyPRE8/edit#heading=h.rg6kb7hpxz4w

PR #6174, opened 2023-08-23
test Ruby 3.2 with docker/dassie

no-reply:
for https://github.com/samvera/hyrax/issues/6137, we'd like to move away from using engine cart to generate an
application for test. instead we can use the same test applications we use for
the dev environment, and their respective docker containers, to run the test
suite in CI.

there are tradeoffs to this approach. on the positive side, we'll have a more
consistent setup across dev and CI, reducing friction for developers and
simplifying the pressures on the test suite (spec setup for each doesn't need to
work for multiple environments). on the negative side, we won't be regularly
testing app generation anymore, and tests will generally be run against a
mature/maintained test app, which won't necessarily represent a clean slate
"basic" hyrax application. it will also be hard to test multiple rails
versions using this strategy.

i think the benefits here outweigh the negatives and that, in general, the
added pressure to continue to maintain dassie (and later koppie) will help us
keep Hyrax usable for long time adopters.

this adds a new workflow to CircleCI with two steps:

kaniko-build uses Kaniko to build the docker container for the application and
push it to GHCR. we tag each build with `[commit-SHA]-dev`; unlike the builds we
tag on release, we don't truncate the commit SHA tags since it isn't
easy (possible?) to truncate when pulling for the next stage. we use kaniko
because is easier to run in a docker container than the docker build tool. we
install `bash` in the container because the CirceCI `ruby/rspec-test`
script (which will be injected by the Ruby Orb for the next stage) is written
for `bash`.

hyrax-test pulls the tagged image and uses it to run the rspec suite. the
environment used is adapted from the Samvera Orb, and we may want to consider
contributing it back there as a later step. rather than using the Circle
`browser-tools` orb, we use the same seleniarm browser image used by the
docker-compose dev environments. this should help normalize configuration for
the capybara setup, and hopefully make that easier to maintain too.

@samvera/hyrax-code-reviewers


no-reply:
in a follow up PR, i'll propose dropping the Ruby 2.7 build, dropping support for Ruby 2.x in Hyrax 5 in general, and i'll open a ticket to remove `engine_cart`

dlpierce:
This would drop testing of dassie with HYRAX_VALKYRIE=true. Do you think that is not worth testing anymore, or is testing koppie is a sufficient replacement?

no-reply:
i've reinstated the valkyrie and non-valkyrie versions of this build.

dlpierce:
Yeah, yarn install is needed in hyrax-engine for Karma which is needed to run the Jasmine specs. It could be added to hyrax-engine-dev in the Dockerfile instead.

no-reply:
> Yeah, yarn install is needed in hyrax-engine for Karma which is needed to run the Jasmine specs. It could be added to hyrax-engine-dev in the Dockerfile instead.

it does seem to me like that should just be part of the image build for the dev image, but i don't want to block this on getting that working. i can follow up about it after we merge this?

no-reply:
the tests we're trying to remove keep intermittently failing (mainly on capybara issues), but the good news is that the new ones don't seem to!

hackartisan (approved):

PR #6180, opened 2023-08-23
Valkyrie publish file.characterized event for derivative creation

### LINKED ISSUE #6162: ValkyrieCharacterizationService should publish file.characterized
### Descriptive summary

Currently, the ValkyrieIngestJob enqueues the derivatives job directly, and enqueues the characterization job via an `object.file.uploaded` event.

Instead, ValkyrieIngestJob should not enqueue ValkyrieCreateDerivativesJob. ValkyrieCharacterizationService should publish file.characterized and FileMetadataListener should enqueue either CreateDerivativesJob or ValkyrieCreateDerivativesJob depending on the payload's type.

See https://github.com/samvera/hyrax/blob/9d06f2b4cc23138259b2703f575a8494a7aa43d5/app/services/hyrax/listeners/file_metadata_listener.rb#L15

This is a code refactor, but may fix race condition bugs if any exist when derivatives try to run before characterization is finished.

lsitu:
### Fixes

Fixes #6162 ; refs #6162

### Summary

Valkyrie publish file.characterized event for derivative creation.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/listeners/file_metadata_listener.rb
+++ b/app/services/hyrax/listeners/file_metadata_listener.rb
@@ -19,12 +19,8 @@ module Hyrax
           CreateDerivativesJob
             .perform_later(file_set, event[:file_id], event[:path_hint])
         else
-          # With valkyrie, file.characterized event is also published for derivatives.
-          # We only need to create derivative for the original file.
-          file_id = event[:file_id]
-          file_metadata = Hyrax.custom_queries.find_file_metadata_by(id: file_id)
-
-          ValkyrieCreateDerivativesJob.perform_later(file_set.id.to_s, file_id) if file_metadata&.original_file?
+          ValkyrieCreateDerivativesJob
+            .perform_later(file_set.id.to_s, event[:file_id])
         end
       end
 
@@ -51,7 +47,9 @@ module Hyrax
       # @param [Dry::Events::Event] event
       # @return [void]
       def on_file_uploaded(event)
-        # Run characterization
+        # Run characterization for original file only
+        return unless event[:metadata]&.original_file?
+
         Hyrax.config
              .characterization_service
              .run(metadata: event[:metadata], file: event[:metadata].file)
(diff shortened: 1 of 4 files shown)


app/services/hyrax/listeners/file_metadata_listener.rb
```diff
@@ -12,8 +12,20 @@ class FileMetadataListener
...
-        CreateDerivativesJob
-          .perform_later(event[:file_set], event[:file_id], event[:path_hint])
+        file_set = event[:file_set]
+
+        case file_set
+        when ActiveFedora::Base # ActiveFedora
+          CreateDerivativesJob
+            .perform_later(file_set, event[:file_id], event[:path_hint])
+        else
+          # With valkyrie, file.characterized event is also published for derivatives.
```
no-reply, line 22: i think we don't want this to be true!

the valkyrie code path shouldn't be calling characterization on files that don't require characterization, right? i think this filter should go below around https://github.com/samvera/hyrax/pull/6180/files#diff-c1500de3cd457700a4445ea1b341164b5a822aa64a086d85e12eb9adfdfeba79R54, or else in the `ValkyrieCharacterizationService`?

no-reply: >  But in UCSD dams, currently we have file characterization for all files, including derivatives.

that seems okay. if downstream applications want to characterize under different conditions, it's easy to subscribe a new listener with the desired behavior.

app/services/hyrax/listeners/file_metadata_listener.rb
```diff
@@ -39,7 +47,9 @@ def on_file_metadata_updated(event)
       # @param [Dry::Events::Event] event
       # @return [void]
       def on_file_uploaded(event)
-        # Run characterization
+        # Run characterization for original file only
+        return unless event[:metadata].original_file?
```
no-reply, line 51:
```suggestion
        return unless event[:metadata]&.original_file?
```

`:metadata` probably SHOULD always be provided, but unhandled exceptions in listeners are very disruptive, so it's nice not to allow for `nil` errors.

no-reply (approved):
looks good, minus the minor suggestion, which i won't block on.

dunn (approved):
tested successfully locally

PR #6183, opened 2023-08-24
[semver-major] drop support for Ruby before 3.2 

no-reply:
drop support for older rubies and remove tests for them.

@samvera/hyrax-code-reviewers


no-reply:
@dlpierce this one really wants your approval, i think:

dlpierce (approved):
2.7 served us well. :wave:

Issue #6170, opened 2023-08-22
Release an RC of Hyrax 5


orangewolf:
- [ ] Review breaking changes and features
- [ ] Create release notes
- [ ] Verify that all code in the 4.x-release branch is in main
- [ ] Mint the release / tag off of main
- [ ] Create a community announcement (or get Heather or Rebecca to do so)

rjkati:
I can create and send an announcement to the community.

PR #6203, opened 2023-08-24
remove various deprecated methods with 4.0 and earlier time horizons

no-reply:
these are broken into individual commits, but it seemed nice to bundle a PR

@samvera/hyrax-code-reviewers

related to #6170


no-reply:
i think this now wants to be rebase on https://github.com/samvera/hyrax/pull/6201

dunn (approved):

PR #6209, opened 2023-08-24
refactor MulitpleMembershipChecker specs to always use valkyrie setup

### LINKED ISSUE #6196: Ensure `spec/services/hyrax/multiple_membership_checker_spec.rb` passes on `.koppie`
`spec/services/hyrax/multiple_membership_checker_spec.rb` is failing on koppie builds. there may be a number of reasons for this:

  1. the test setup uses ActiveFedora, and should be refactored to use Valkyrie;
  2. the tests themselves test behavior that only applies to ActiveFedora, and should be turned off by adding the `:active_fedora` rspec metadata;
  3. the unit under test depends on ActiveFedora and needs to be rewritten to use Valkyrie throughout;
  4. some combination of the above.

----

```sh
docker-compose -f docker-compose-koppie.yml build
docker-compose -f docker-compose-koppie.yml up
docker-compose -f docker-compose-koppie.yml exec -w /app/samvera/hyrax-engine app sh -c "bundle exec rspec spec/services/hyrax/multiple_membership_checker_spec.rb"
```

no-reply:
fixes #6196

@samvera/hyrax-code-reviewers


no-reply:
i guess one question i haven't addressed here is: does the unit under test continue to require support for ActiveFedora (probably yes?), and if so, does some subset of the prior AF tests need to remain with an added `:active_fedora` rspec tag to prevent them from running on `.koppie`?

no-reply:
i looked at this with @sdellis, and since this class doesn't have branching behavior between AF and Valkyrie models, i think it should be acceptable to drop explicit tests for AF support.

tpendragon (approved):

PR #6212, opened 2023-08-24
Reindex the work during derivative creation if it is the work's thumbnail

### LINKED ISSUE #5504: Generate derivatives for FileSets when Valkyrie mode is enabled

dlpierce:
### Fixes

Fixes #5504 

### Summary

Reindex the work during derivative creation if it is the work's thumbnail

* A work's thumbnail is shown in the catalog view without re-saving the work after initial ingest.

@samvera/hyrax-code-reviewers


app/jobs/valkyrie_create_derivatives_job.rb
```diff
@@ -1,23 +1,24 @@
...
-  end
-
-  def storage_adapter
-    Hyrax.storage_adapter
+  def reindex_parent(file_set_id)
+    file_set = Hyrax.query_service.find_by(id: file_set_id)
+    parent = Hyrax.custom_queries.find_parent_work(resource: file_set)
+    return unless parent.thumbnail_id == file_set.id
+    user = User.find_by_user_key(file_set.depositor)
+    Hyrax.publisher.publish('object.metadata.updated', object: parent, user: user)
```
no-reply, line 22: it doesn't seem right to publish `object.metadata.updated` here? we're not actually updating the object metadata.

if the goal is to reindex, shouldn't we just do that explicitly and avoid triggering other application behavior that might expect there to have been an actual change attributable to `user`?
 ```suggestion
    Hyrax.index_adapter.save(resource: parent)
```

no-reply: i think this also avoids the need to query for a user, who didn't really have anything to do with this change.

dlpierce: I was under the impression that publishing was the preferred way in general. Thanks for clearing that up.

no-reply: i think the contract with publishing events is that Hyrax will:
  - always publish them when the event occurs;
  - never publish them otherwise.

downstream applications need to be able to rely on events to trigger at the right times and not to invoke listener behavior when it's not appropriate.

tpendragon (approved):

PR #6220, opened 2023-08-25
Fix db_migrate service failing to find db-migrate-seed.sh

dlpierce:
### Fixes

db_migrate service failing to find db-migrate-seed.sh

Not sure why a login shell was specified, but it is causing problems now. PATH does not contain /app/samvera inside a login shell.


no-reply (approved):
was just looking at the same thing!

PR #6221, opened 2023-08-25
Frijya and Frigg Adapters for Migrating Collections, AdminSets and Works to Postgres or Fcrepo 6

orangewolf:
## Overview

First, apologies.  The nature of the work has been that we've been upgrading Hyku and testing Hyku against this branch.  As such, it's a larger PR than we might normally prefer.

This PR includes two primary considerations:

1. Code changes that allow for a "lazy migration" of data ([see discussion below](https://github.com/samvera/hyrax/pull/6221#issuecomment-1904756037)).
2. Refactorings that allow for downstream applications to configure behavior instead of override entire files.

### Further Discussion

The primary work introduces the Frigga and Freyja metadata adapter; these adapters are configured to:

- Read first from one storage location, and failing to find something read from another (e.g. `storage_first.find || storage_second.find` if you will)
- Write to the first storage location, ignoring the second storage location.

Through configuration and convention, the Frigga and Freyja adapters write similar looking Solr documents as their ActiveFedora based counterparts.

Frigga's adapter's first storage location is Postgres via Valkyrie, and it's second storage location is fcrepo via ActiveFedora.

Freyja's adapter's first storage location is fcrepo via Valkyrie, and it's second storage location is an fcrepo via ActiveFedora.  Note, this is likely Fedora 6 and Fedora 4 respectively.

What this means is that these adapters provide a pathway for minimal downtime during a storage migration; something demonstrated by [GBH's AMS project](https://github.com/WGBH-MLA/ams) and presented on at Samvera Connect 2023 by @orangewolf.

## Guidance for testing, such as acceptance criteria or new user interface behaviors:

There are no known new user behaviors.  Instead we're relying on the test suite as well as downstream testing of Hyku migrations to help verify.

@samvera/hyrax-code-reviewers


lib/freyja/query_service.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module Freyja
+  class QueryService
```
tpendragon, line 4: I'd :+1: adding a composite query service like this in Valkyrie if you wanted.

orangewolf: seems like a good plan to get through it in Hyrax land and then port it. Since I'm modifying a valkyrie branch to make the shared examples flexible enough for this use case, it should slot over super easy if we get to that point.

jeremyf:
The goal of `double_combo` approach is to minimize downtime by allowing for in-place migrations.

*Note: In this scenario, I’m using Fedora4 as the name for the legacy persistence layer, and Postgres as a symbolic name for the metadata adapter’s persistence layer; we could just as easily use Fedora4 but for clarifying purposes I’m separating these concepts.*

This is done by leveraging a copy on write (CoW) type strategy.  Below are two examples:

```gherkin
Given I wrote Work A to Fedora4 via ActiveFedora
And wrote Work A to Solr via ActiveFedora
And did not write Work A to Postgres
When I use Valkyrie to find Work A
Then Valkyrie will first check Postgress for Work A
And Valkyrie will next check Fedora4 for Work A
```

From the above scenario:

```gherkin
Given I have found Work A via Valkyrie
When I use Valkyrie to save Work A
Then Valkyrie will write to Postgres
And Valkyrie will update the Solr document originally written by ActiveFedora
```

At the point of completing the `write` scenario, when we go to find “Work A” via Valkyrie, we’ll first check and find the record in Postgres.

Consider that the Fedora4 repository also has works B, C, D, and E.  And all of those are part of Collection Alpha.  When rendering the contents of the collection, we leverage the Solr document representations of each of the works.  This is done by creating presenters.  To avoid presentation drift, we would like for A, B, C, D, and E to have

Put another way, the Presenter for Work A’s solr document written by ActiveFedora and the Presenter for Work A’s solr document written by Valkyrie should have the same methods and each of those methods used for rendering should return the same values.

To reduce duplication, it would be ideal for both of those presenters to in fact be the same presenter.  It then follows that we would want to minimize branching logic between documents written via ActiveFedora and those written by Valkyrie; the greatest minimization would be for ActiveFedora and Valkyrie to generate logically equivalent solr documents.  I say logically equivalent in that there are administrative benefits for tracking in Solr which documents have been valkyrized (see `valkyrie_bsi`).

This leads to the aspiration that the `has_model_ssim` should be stable between a document written via ActiveFedora and one written via Valkyrie.

The `has_model_ssim` is a foundational attribute in that presenters use this for switching logic and we use it for Solr querying logic.  See `Hyrax::SolrDocumentBehavior#hydra_model` and it’s uses.

jeremyf:
## Concerning `Freyja::CustomQueryContainer#method_missing`.

When we attempt to call a `find_access_control_for` custom query checking first service the method raises a `Valkyrie::Persistence::ObjectNotFoundError`.  Which is then rescued in `Hyrax::AccessControl.for` by instantiating a new `Hyrax::AccessControl` object.

If in the `Freyja::CustomQueryContainer` we rescue `Valkyrie::Persistence::ObjectNotFoundError`, then the second service successfully finds the resource in postgres.  However, we then encounter a problem in a subsequent call; namely that the first service will now raise `ActiveFedora::IllegalOperation`.

I am leaving this here for the end of week, sharing the code and discovery.  Some possible things for next week’s exploration is to create an explicitly defined method for `#find_access_control_for`.

Below are the two pieces of code followed by the backtrace for `ActiveFedora::IllegalOperation`; of note in the backtrace is that we attempt to call `ActiveFedora::Core#initialize`, a bit of a surprise.

```ruby
# frozen_string_literal: true
module Freyja
  class CustomQueryContainer < Valkyrie::Persistence::CustomQueryContainer
    def method_missing(method_name, *args, **opts)
      query_service.services.each do |service|
        return service.custom_queries.send(method_name, *args, **opts) if service.custom_queries.respond_to?(method_name)
      end
      super
    end

    def respond_to_missing?(method_name, _include_private = false)
      query_service.services.each do |service|
        return true if service.custom_queries.respond_to?(method_name)
      end
      false
    end
  end
end

```

```ruby
module Hyrax
  class AccessControl < Valkyrie::Resource
    ##
    # A finder/factory method for getting an appropriate ACL for a given
    # resource.
    #
    # @param resource [Valkyrie::Resource]
    # @param query_service [#find_inverse_references_by]
    #
    # @return [AccessControl]
    # @raise [ArgumentError] if the resource is not persisted
    def self.for(resource:, query_service: Hyrax.query_service)
      query_service.custom_queries.find_access_control_for(resource: resource)
    rescue Valkyrie::Persistence::ObjectNotFoundError
      new(access_to: resource.id)
    end
  end
end
```

```nil
ActiveFedora::IllegalOperation:
Attempting to recreate existing ldp_source:
`http://fcrepo:8080/rest/hykudemo/6d/90/14/a6/6d9014a6-06e7-4516-86be-87d899e62176'
# gems/active-fedora-14.0.1/lib/active_fedora/core.rb:36:in `initialize'
# gems/valkyrie-3.1.3/lib/valkyrie/persistence/postgres/orm_converter.rb:27:in `new'
# gems/valkyrie-3.1.3/lib/valkyrie/persistence/postgres/orm_converter.rb:27:in `resource'
# gems/valkyrie-3.1.3/lib/valkyrie/persistence/postgres/orm_converter.rb:19:in `convert!'
# gems/valkyrie-3.1.3/lib/valkyrie/persistence/postgres/resource_factory.rb:20:in `to_resource'
# gems/valkyrie-3.1.3/lib/valkyrie/persistence/postgres/query_service.rb:134:in `block in run_query'
# bundler/gems/hyrax-21dd6652f5f4/app/services/hyrax/custom_queries/find_access_control.rb:21:in `each'
# bundler/gems/hyrax-21dd6652f5f4/app/services/hyrax/custom_queries/find_access_control.rb:21:in `each'
# bundler/gems/hyrax-21dd6652f5f4/app/services/hyrax/custom_queries/find_access_control.rb:21:in `each'
# bundler/gems/hyrax-21dd6652f5f4/app/services/hyrax/custom_queries/find_access_control.rb:21:in `each'
# bundler/gems/hyrax-21dd6652f5f4/app/services/hyrax/custom_queries/find_access_control.rb:21:in `find'
# bundler/gems/hyrax-21dd6652f5f4/app/services/hyrax/custom_queries/find_access_control.rb:21:in `find_access_control_for'
# gems/valkyrie-3.1.3/lib/valkyrie/persistence/custom_query_container.rb:55:in `block (2 levels) in register_query_handler'
```

lib/goddess/custom_query_container.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      # None of the services responded to the method_name
+      return super unless service_responds_to
+
+      # At least one of the services had a valid query result; though it could be an empty array.
+      # Is that okay?  What if we would have had a non-empty array in a later service?
+      return returning_value if returning_value
+
+      # We didn't find anything so we need to re-raise the like error
+      raise Valkyrie::Persistence::ObjectNotFoundError, "#{self.class}#method_missing with method_name: #{method_name.inspect}, args: #{args.inspect}, opts: #{opts.inspect}"
```
orangewolf, line 33: if you create a variable (`e`) on line 14, then set it if need by on line 21, then this can simply be `raise e` and will walk the backtrace up the line. not sure if thats more info that you're getting now or less and I'd go with whatever gives us more actionable info.

lib/freyja/resource_factory.rb
```diff
@@ -0,0 +1,148 @@
...
+      #         moved to the Valkyrie storage adapter.
+      # @return [FalseClass] when the thumbnail has been moved to the Valkyrie
+      #         storage adapter.
+      # @see #move_thumbnail_to_backup
+      def thumbnail_exists?(path)
+        path.present? && File.exist?(path)
+      end
+
+      def migrate_thumbnail!(resource:)
+        return true # JEREMY and SHANA hacking away
```
dlpierce, line 87: How did this hacking go?

lib/frigg/persister.rb
```diff
@@ -0,0 +1,33 @@
...
+      #     if resource.id && resource.id.to_s != orm_object.id
+      #       raise Valkyrie::Persistence::UnsupportedDatatype,
+      #             "Postgres' primary key column can not save with the given ID #{resource.id}. " \
+      #             "To avoid this error, set the ID to be nil via `resource.id = nil` before you save it. \n" \
+      #             "Called from #{Gem.location_of_caller.join(':')}"
+      #     end
+      #   end
+      #   resource_factory.to_resource(object: orm_object)
+      # rescue ActiveRecord::StaleObjectError
+      #   raise Valkyrie::Persistence::StaleObjectError, "The object #{resource.id} has been updated by another process."
```
dlpierce, line 29: This looks like Frigg isn't persisting?

spec/lib/freyja/persister_spec.rb
```diff
@@ -0,0 +1,483 @@
...
+              .to raise_error(Valkyrie::Persistence::StaleObjectError)
+          end
+        end
+      end
+    end
+
+    context 'ordered properties' do
+      xit "orders string values and returns them in the appropriate order" do
+        validate_order ["a", "b", "a"]
+      end
```
dlpierce, line 405: All of the specs in the ordered properties section are skipped. What's the reason for that?

spec/lib/frigg/persister_spec.rb
```diff
@@ -0,0 +1,483 @@
...
+require 'valkyrie/specs/shared_specs'
+require 'wings'
+require 'frigg/metadata_adapter'
+
+RSpec.describe Frigg::Persister, :active_fedora do
+  subject(:persister) { described_class.new(adapter: adapter) }
+  let(:adapter) { Frigg::MetadataAdapter.new }
+  let(:query_service) { adapter.query_service }
+
+  xcontext "When passing a Valkyrie::Resource that was never an ActiveFedora::Base" do
```
dlpierce, line 15: This skipped spec context would go with the commented code in the frigg persister.

spec/lib/frigg/persister_spec.rb
```diff
@@ -0,0 +1,483 @@
...
+              .to raise_error(Valkyrie::Persistence::StaleObjectError)
+          end
+        end
+      end
+    end
+
+    context 'ordered properties' do
+      xit "orders string values and returns them in the appropriate order" do
+        validate_order ["a", "b", "a"]
+      end
```
dlpierce, line 405: ordered properties specs skipped here too

spec/lib/frigg/query_service_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+require 'spec_helper'
+require 'valkyrie/specs/shared_specs'
+require 'wings'
+require 'frigg/metadata_adapter'
+
+RSpec.describe Frigg::QueryService, :clean_repo do
+  let(:adapter) { Frigg::MetadataAdapter.new }
+
+  xcontext "items in postgres only" do
```
dlpierce, line 13: Everything skipped here too

dlpierce (changes requested):
I see a few spots with commented out code/specs that seem important.

lib/frigg/persister.rb
orangewolf: thats correct. it's structure is there, but there wasn't a fcrepo 6 version of Hyrax at the time I first put up the PR. I had a few requests to at least put the skeleton of an fcrepo 6 version in place even if it wasn't working yet, in order to signal to the community that it could be done. Once this is merged, we can do an investigation on getting fcrepo 6 version fully working. To be clear, I dont have any projects at the moment with fcrepo 6 on their road map.

spec/lib/frigg/persister_spec.rb
orangewolf: that's correct, and for the same reason

spec/lib/freyja/persister_spec.rb
orangewolf: no sure why these were skipped, they work fine. I brought them all back

dlpierce:
I'm continuing to review this and make sure we have a dassie config that uses Freyja.

dlpierce:
I think we're getting close. I feel like the changes in dassie to enable Freyja need an on/off env toggle.

dlpierce (dismissed):

dlpierce (approved):

Issue #6198, opened 2023-08-24
Ensure `spec/presenters/hyrax/work_show_presenter_spec.rb` passes on `.koppie`


no-reply:
`spec/presenters/hyrax/work_show_presenter_spec.rb` is failing on koppie builds. there may be a number of reasons for this:

  1. the test setup uses ActiveFedora, and should be refactored to use Valkyrie;
  2. the tests themselves test behavior that only applies to ActiveFedora, and should be turned off by adding the `:active_fedora` rspec metadata;
  3. the unit under test depends on ActiveFedora and needs to be rewritten to use Valkyrie throughout;
      * if this involves adding a switch for AF / Valkyrie behavior, ensure both branches are covered by adding another test
  5. some combination of the above.

at a glance, i think this is a combination of at leaste 1 and 3.

----

```sh
docker-compose -f docker-compose-koppie.yml build
docker-compose -f docker-compose-koppie.yml up
docker-compose -f docker-compose-koppie.yml exec -w /app/samvera/hyrax-engine app sh -c "bundle install"
docker-compose -f docker-compose-koppie.yml exec -w /app/samvera/hyrax-engine app sh -c "bundle exec rspec spec/presenters/hyrax/work_show_presenter_spec.rb"
```

hackartisan:
closed by #6225

PR #6225, opened 2023-08-25
Fix several work_show_presenter specs on koppie

hackartisan:
WIP towards #6198

### DIFF
--- a/app/indexers/hyrax/valkyrie_work_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_work_indexer.rb
@@ -9,7 +9,6 @@ module Hyrax
     include Hyrax::VisibilityIndexer
     include Hyrax::ThumbnailIndexer
     include Hyrax::Indexer(:core_metadata)
-    include Hyrax::Indexer(:basic_metadata)
 
     def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
       super.tap do |solr_doc|
--- a/app/indexers/hyrax/valkyrie_work_indexer.rb
+++ b/app/indexers/hyrax/valkyrie_work_indexer.rb
@@ -9,6 +9,7 @@ module Hyrax
     include Hyrax::VisibilityIndexer
     include Hyrax::ThumbnailIndexer
     include Hyrax::Indexer(:core_metadata)
+    include Hyrax::Indexer(:basic_metadata)
 
     def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
       super.tap do |solr_doc|
--- a/app/presenters/hyrax/pcdm_member_presenter_factory.rb
+++ b/app/presenters/hyrax/pcdm_member_presenter_factory.rb
@@ -47,14 +47,14 @@ module Hyrax
     #   we recommend making sparing use of this feature.
     #
     # @overload member_presenters
-    #   @return [Enumerator<FileSetPresenter, WorkShowPresenter>]
+    #   @return [Array<FileSetPresenter, WorkShowPresenter>]
     #   @raise [ArgumentError] if an unindexed id is passed
     # @overload member_presenters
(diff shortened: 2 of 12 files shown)


spec/support/simple_work.rb
```diff
@@ -10,7 +10,9 @@ module Test
     #
     # @example creating with FactoryBot
     #   work = FactoryBot.valkyrie_create(:hyrax_work, :public, title: ['Comet in Moominland'])
-    class SimpleWork < Hyrax::Work; end
+    class SimpleWork < Hyrax::Work
+      include Hyrax::Schema(:basic_metadata)
```
no-reply, line 14: do you want to be using `Monograph` instead?

`SimpleWork` is meant to be minimal, ensuring Hyrax works *without* including BasicMetadata (this has been a problem in the past)

no-reply: (it's also possible, i guess, that the problem here is that `WorkShowPresenter` really does only work with BasicMetadata)?

app/indexers/hyrax/valkyrie_work_indexer.rb
```diff
@@ -9,6 +9,7 @@ class ValkyrieWorkIndexer < Hyrax::ValkyrieIndexer
     include Hyrax::VisibilityIndexer
     include Hyrax::ThumbnailIndexer
     include Hyrax::Indexer(:core_metadata)
+    include Hyrax::Indexer(:basic_metadata)
```
no-reply, line 12: i think this should be avoided, since not all works include basic metadata.

spec/support/simple_work.rb
hackartisan: @no-reply It's needed for the resource_type field. Maybe that's really a core field?

app/indexers/hyrax/valkyrie_work_indexer.rb
hackartisan: @no-reply Without it resource_type doesn't get indexed. Other suggestions on how to do this?

spec/support/simple_work.rb
no-reply: i don't think it should be!

my understanding is that Hyrax promises to work with models that only have the current core properties. if the show presenter will error if it's not present, i'd think that's a bug!

app/indexers/hyrax/valkyrie_work_indexer.rb
no-reply: my best understanding is:
  - WorkShowPresenter and its callers should tolerate `resource_type` being absent;
  - tests for any (optional) behavior that depends on `resource_type` should use the `:monograph` factory

no-reply:
this looks great to me! setting aside the `:basic_metadata` issue.

app/presenters/hyrax/pcdm_member_presenter_factory.rb
```diff
@@ -28,7 +28,7 @@ def initialize(object, ability, _request = nil)
     # @return [Array<FileSetPresenter, WorkShowPresenter>]
     # @return [Enumerator<FileSetPresenter>]
     def file_set_presenters
-      return enum_for(:file_set_presenters) unless block_given?
+      return enum_for(:file_set_presenters).to_a unless block_given?
```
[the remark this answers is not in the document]
hackartisan: We could change the callers I think, we decided to do it this way because it preserved the previous api.

spec/support/simple_work.rb
hackartisan: okay, you are so very right, monograph works great, thank you for your persistence lol

app/presenters/hyrax/pcdm_member_presenter_factory.rb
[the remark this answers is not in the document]
hackartisan: We talked about this more and think `to_a` is right; if a performance enhancement is needed in future maybe that could be outside the scope of this PR?

no-reply (approved):

PR #6226, opened 2023-08-25
Fix single use links for Valkyrie.

### LINKED ISSUE #6204: Using a single use link generates argument error in valkyrie
### Descriptive summary

When I attempt to use a single use link to a fileset, I get an `ArgumentError in Hyrax::SingleUseLinksViewerController#download` page instead of the file.

This works as expected on dev.nurax, but not pg.nurax

1. Deposit a work with institution, embargoed or private visibility.
2. On the work show page for the work which you deposited in step 1, scroll down to the Items section and click on the file name
3. Click the `Create Single-Use Link` button
4. Click `Copy` under the `Single-Use Links` header
5. Go to a different browser and paste the single use link
6. Observe error message

Using a single-use link generates an error

single-use_link_error

Using a single-use link should download the file associated with the link.

tpendragon:
Closes #6204

This also adds a shortcut to allow creating a work that has a file, including a FileSet, FileMetadata, and File Identifier. I'm not sure it's the best, but it works really well for this test!

### DIFF
--- a/app/controllers/hyrax/single_use_links_viewer_controller.rb
+++ b/app/controllers/hyrax/single_use_links_viewer_controller.rb
@@ -37,6 +37,14 @@ module Hyrax
 
     private
 
+    def send_content
+      if Hyrax.config.use_valkyrie?
+        send_file_contents_valkyrie(@asset)
+      else
+        super
+      end
+    end
+
     def curation_concern
       response, _document_list = search_service.search_results
       response.documents.first
@@ -47,6 +55,12 @@ module Hyrax
     end
 
     def content_options
+      super.tap do |options|
+        options[:disposition] = 'attachment' if action_name == 'download'
+      end
+    end
+
+    def data_options(file_metadata)
       super.tap do |options|
         options[:disposition] = 'attachment' if action_name == 'download'
       end
@@ -66,7 +80,7 @@ module Hyrax
     end
 
     def asset
-      @asset ||= Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: single_use_link.item_id, use_valkyrie: false)
+      @asset ||= Hyrax.query_service.find_by(id: single_use_link.item_id)
     end
 
     def current_ability
@@ -94,7 +108,7 @@ module Hyrax
         return unless single_use_link
 
         @single_use_link = single_use_link
-        can :read, [ActiveFedora::Base, ::SolrDocument] do |obj|
+        can :read, [ActiveFedora::Base, ::SolrDocument, Hyrax::Resource] do |obj|
           single_use_link.valid? && single_use_link.item_id == obj.id && single_use_link.destroy!
         end
       end
--- a/app/controllers/hyrax/single_use_links_viewer_controller.rb
+++ b/app/controllers/hyrax/single_use_links_viewer_controller.rb
@@ -38,7 +38,7 @@ module Hyrax
     private
 
     def send_content
-      if Hyrax.config.use_valkyrie?
+      if @asset.is_a?(Hyrax::FileSet)
         send_file_contents_valkyrie(@asset)
       else
         super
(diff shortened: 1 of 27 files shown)


spec/factories/hyrax_work.rb
```diff
@@ -41,6 +42,15 @@
           .new(resource: work)
           .assign_access_for(visibility: evaluator.visibility_setting)
       end
+      if evaluator.uploaded_files.present?
+        Hyrax::WorkUploadsHandler.new(work: work).add(files: evaluator.uploaded_files).attach
+        evaluator.uploaded_files.each do |file|
+          allow(Hyrax.config.characterization_service).to receive(:run).and_return(true)
+          # I don't love this - we might want to just run background jobs so
+          # this is more real, but we'd have to stub some things.
```
no-reply, line 50: could we do

```ruby
ActiveJob::Base.queue_adapter.filter += [ValkyrieIngestJob]
```

Hyrax already makes use of the `ActiveJob` test adapter. i guess the real implementation is probably a bit fussy, since we need to reach into config and then reset it to be really clean:

```ruby
old_perform_enqueued_setting = ActiveJob::Base.queue_adapter.perform_enqueued_jobs
old_job_filter = ActiveJob::Base.queue_adapter.filter 
ActiveJob::Base.queue_adapter.perform_enqueued_jobs = true
ActiveJob::Base.queue_adapter.filter += [ValkyrieIngestJob]

allow(Hyrax.config.characterization_service).to receive(:run).and_return(true)
Hyrax::WorkUploadsHandler.new(work: work).add(files: evaluator.uploaded_files).attach

ActiveJob::Base.queue_adapter.perform_enqueued_jobs = old_perform_enqueued_setting
ActiveJob::Base.queue_adapter.filter = old_job_filter
```

maybe more than is worth messing with?

---

the stub on `Hyrax.config.characterization_service` feels like a code smell to me. currently, if i publish `file.uploaded` from any synchronous code, i'll trigger a long running `characterization_service.run` inline. since any app can publish that event from any code, that might not be the best.

this issue is probably out of scope here, but fixing it would let us rely on the `ActiveJob` adapter filter for the characterization behavior too, which would offer a lot of flexiblitiy.

app/controllers/hyrax/single_use_links_viewer_controller.rb
```diff
@@ -66,7 +80,7 @@ def not_found_exception
     end
 
     def asset
-      @asset ||= Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: single_use_link.item_id, use_valkyrie: false)
+      @asset ||= Hyrax.query_service.find_by(id: single_use_link.item_id)
```
no-reply, line 83: 👏🏻

spec/factories/hyrax_file_set.rb
```diff
@@ -24,7 +25,7 @@
           .assign_access_for(visibility: evaluator.visibility_setting)
       end
       file_set.file_ids = evaluator.files.map(&:id) if evaluator.files
-      file_set.original_file_id = evaluator.original_file.id if evaluator.original_file
+      file_set.original_file_id = evaluator&.original_file&.id || evaluator.files&.first&.id
```
no-reply, line 28: i'm not sure i'm following this. is it used somewhere i'm missing?

no-reply (approved):

spec/factories/hyrax_work.rb
tpendragon: I had a version that did `perform_enqueued jobs`

The big problem is just fits tries to run and it's not installed.

spec/factories/hyrax_file_set.rb
tpendragon: Oh this might be left over from a previous try.

spec/factories/hyrax_work.rb
no-reply: does the stub not take effect through the adapter?

no-reply (approved):

PR #6231, opened 2023-08-28
remove deprecated argument from CollectionBrandingInfo#delete

no-reply:
@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/models/collection_branding_info_spec.rb
+++ b/spec/models/collection_branding_info_spec.rb
@@ -55,9 +55,13 @@ RSpec.describe CollectionBrandingInfo, type: :model do
     before { banner_info.save(file.path) }
 
     it "removes banner file" do
-      expect { banner_info.delete }
-        .to change { storage_adapter.find_by(id: banner_info.local_path) }
-        .to raise_error Valkyrie::StorageAdapter::FileNotFound
+      expect { banner_info.delete }.to change {
+        begin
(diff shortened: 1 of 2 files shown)


dunn (dismissed):

tpendragon:
We think that this is just a test error, maybe there's something wrong with the `to_raise` - maybe you have to chain it to a thing coming from a block, and it's chained on the to_change.

spec/models/collection_branding_info_spec.rb
```diff
@@ -52,20 +52,16 @@
...
-      expect { storage_adapter.find_by(id: logo_info.local_path) }
-        .to raise_error Valkyrie::StorageAdapter::FileNotFound
+    it "removes banner file" do
+      expect { banner_info.delete }.to change {
+        begin
+          storage_adapter.find_by(id: banner_info.local_path)
+        rescue Valkyrie::StorageAdapter::FileNotFound
+          'Raised FileNotFound' # Convert the raised error so change can detect it
+        end
+      }.from(be_a(Valkyrie::StorageAdapter::File)).to(eq 'Raised FileNotFound')
```
dlpierce, line 64: I couldn't find a way to make the change matcher play nicely with raised errors, so instead I'm converting it to a string.

tpendragon (approved):

PR #6235, opened 2023-08-28
2657-auto-expiration

alishaevn:
### Summary
actually release the embargo and lease with the [embargo_lease rake tasks](https://github.com/samvera/hyrax/blob/main/lib/tasks/embargo_lease.rake)

* ensure you have works whose embargo/lease dates have passed
* run the below tasks in a container
  ``` bash
  rake hyrax:embargo:deactivate_expired
  rake hyrax:lease:deactivate_expired
  ```
* see that works with expired embargoes/leases no longer have an embargo/lease on them
* works with valid release/expiration dates, still have the embargo/lease applied

### Detailed Description
with this commit, a user is able to run `rake hyrax:embargo:deactivate_expired` or `rake hyrax:lease:deactivate_expired` manually and only release the embargoes or leases whose expiration/release dates have already passed. we are not simply calling the `deactivate!` method in each manager because that would deactivate ALL leases and embargoes, even if they are still active.

ref: #2657
ref: https://github.com/scientist-softserv/atla-hyku/issues/121

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/embargo_manager.rb
+++ b/app/services/hyrax/embargo_manager.rb
@@ -110,8 +110,18 @@ module Hyrax
       end
 
       def release_embargo_for(resource:, query_service: Hyrax.query_service)
-        new(resource: resource, query_service: query_service)
-          .release
+        em = new(resource: resource, query_service: query_service)
+        # We will only continue beyond this point if the embargo release date for the current resource has passed
+        # If the intention is to release the embargo early, then the embargo should be deactivated instead
+        return unless em.release
+
+        em.nullify
+        if resource.class <= Valkyrie::Resource
+          resource.embargo = Hyrax.persister.save(resource: em.embargo)
+        else
+          resource.embargo.save!
+          resource.save!
+        end
       end
 
       def release_embargo_for!(resource:, query_service: Hyrax.query_service)
--- a/app/services/hyrax/lease_manager.rb
+++ b/app/services/hyrax/lease_manager.rb
@@ -54,8 +54,18 @@ module Hyrax
       end
 
       def release_lease_for(resource:, query_service: Hyrax.query_service)
-        new(resource: resource, query_service: query_service)
-          .release
+        lm = new(resource: resource, query_service: query_service)
+        # We will only continue beyond this point if the lease expiration date for the current resource has passed
+        # If the intention is to release the lease early, then the lease should be deactivated instead
+        return unless lm.release
+
+        lm.nullify
+        if resource.class <= Valkyrie::Resource
+          resource.lease = Hyrax.persister.save(resource: lm.lease)
+        else
+          resource.lease.save!
+          resource.save!
+        end
       end
 
       def release_lease_for!(resource:, query_service: Hyrax.query_service)


app/services/hyrax/embargo_manager.rb
```diff
@@ -110,8 +110,18 @@ def embargo_for(resource:, query_service: Hyrax.query_service)
       end
 
       def release_embargo_for(resource:, query_service: Hyrax.query_service)
-        new(resource: resource, query_service: query_service)
-          .release
+        em = new(resource: resource, query_service: query_service)
+        # We will only continue beyond this point if the embargo release date for the current resource has passed
+        # If the intention is to release the embargo early, then the embargo should be deactivated instead
+        return unless em.release
```
alishaevn, line 116: this will return false if the embargo is still active.

app/services/hyrax/embargo_manager.rb
```diff
@@ -110,8 +110,18 @@ def embargo_for(resource:, query_service: Hyrax.query_service)
...
+
+        em.nullify
+        if resource.class <= Valkyrie::Resource
+          resource.embargo = Hyrax.persister.save(resource: em.embargo)
+        else
+          resource.embargo.save!
+          resource.save!
+        end
```
alishaevn, line 124: this follows the same pattern as `Hyrax::Actors::EmbargoActor#destroy`

app/services/hyrax/lease_manager.rb
```diff
@@ -54,8 +54,18 @@ def lease_for(resource:, query_service: Hyrax.query_service)
       end
 
       def release_lease_for(resource:, query_service: Hyrax.query_service)
-        new(resource: resource, query_service: query_service)
-          .release
+        lm = new(resource: resource, query_service: query_service)
```
alishaevn, line 57: follows the same reasoning as the embargo manager.

no-reply:
@alishaevn i think the intention was that the issue of saving the relevant objects would be one that could be handled by the caller. i.e. `release_embargo_for` is just meant to be a convenience method for `new(*args).release`, which is why it's not tested.

i'm okay with expanding its footprint, but in that case i think it needs docs and tests.

i suspect that the other side of this issue is the `hyrax.embargo/lease.deactivate_expired` tasks were written with a valkyrie native app in mind, assuming that AF/Wings apps would continue to use older infrastructure(?)

no-reply:
i'm also wondering how this is working with `EmbargoExpiryJob` and `LeaseExpiryJob` which seem to test the same behavior from the rake tasks fairly carefully.

i'd really like to see a failing test that these changes fix.

spec/services/hyrax/embargo_manager_spec.rb
```diff
@@ -220,6 +220,35 @@
...
+      let!(:vis_after_embargo) { embargo.visibility_after_embargo }
+
+      it 'removes the embargo from the resource' do
+        manager.release_embargo_for(resource: resource)
+
+        # #release will set the visibility correctly on the resource
+        # #nullify however will then set embargo.visibility_after_embargo to be nil
+        expect(resource.visibility).to eq(vis_after_embargo)
+        expect(resource.embargo.embargo_release_date).to be nil
+        expect(resource.embargo.visibility_during_embargo).to be nil
```
no-reply, line 237: is this the initial problem?

i recall thinking at initial implementation time that it was simpler and probably desirable in general to leave this data in place when releasing embargoes. since the embargo is a separate Resource, it seemed just destructive to wipe its data.

alishaevn:
> i think the intention was that the issue of saving the relevant objects would be one that could be handled by the caller. i.e. 

release_embargo_for is just meant to be a convenience method for new(*args).release, which is why it's not tested.
I made this change because while testing the rake tasks for atla, I discovered that the rake tasks aren't working in their implementation, OR hyrax `main`. so I set out to at least make the rake tasks work
<hr>

> i'm okay with expanding its footprint, but in that case i think it needs docs and tests.

I added some tests and then saw you'd asked for them 🙂 
<hr>

> i suspect that the other side of this issue is the hyrax.embargo/lease.deactivate_expired tasks were written with a valkyrie 

native app in mind, assuming that AF/Wings apps would continue to use older infrastructure(?); a valkyrie native app shouldn't need to save the object to persist the embargo change, just the ACL.
I was only running this in dassie, but added the save because the embargo _wasn't_ getting updated.
<hr>

> i'm also wondering how this is working with EmbargoExpiryJob and LeaseExpiryJob which seem to test the same behavior from the rake tasks fairly carefully.

the expiry jobs basically do the same thing as the rake tasks, both calling their respective #release_lease_for methods. so I don't expect those to work either.
<hr>

> i'd really like to see a failing test that these changes fix.

done!

spec/services/hyrax/embargo_manager_spec.rb
alishaevn: I'm unsure whether this is actually the initial problem, but I know it solves the problem. 

without calling `#nullify`, the resource's visibility would say it was updated, but the embargo wasn't actually removed. refreshing the ui would still show the work, just in a different order. (e.g. it may have initially been the first work on "/embargoes", but now it's the last). I added the nullify because it follows the pattern that `deactivate!` does.

no-reply:
@alishaevn i'm still concerned that this is changing the data interactions to solve a problem that's originating elsewhere. the new tests are great, but they test the changes to this class, not the original problem.

i think it would be helpful to get something in at the Controller layer (or a feature spec) that demonstrates the problem.

for what it's worth, i'm also concerned that `#deactivate!` is approaching this in the wrong way. this manager was intended to try to create clear definitions of what embargo states result in what behaviors, if we're finding that those don't line up to what's happening in the application, there's a deeper problem to root out.

no-reply:
looking at this a bit closer, it seems to me like the issue is in `Hyrax::EmbargoService` which *explicitly* returns all works with *any* embargo release date, and makes no attempt to check whether the embargo is enforced.

i think that's the reason this issue has been chased around for some time: the definition of an active embargo is different from what we're checking at display time. i'd suggest we fix the query.

alishaevn:
@no-reply fyi, I tested on main now that #6241 was merged. my embargoed works that should have expired, didn't. I'll try to take another look at this sometime this week.

<details>
<summary>screen shots</summary>

image
image
</details>

alishaevn:
in reading my initial comment above again, I remembered that the embargoes that are set to expire don't automatically expire. (as is the expectation. there's a ticket to change that [here](https://github.com/samvera/hyrax/issues/2657), but it's been open for 5.5 years)

to actually expire an embargo, there's a [rake task](https://github.com/samvera/hyrax/blob/main/lib/tasks/embargo_lease.rake) (`rails hyrax:embargo:deactivate_expired`), [cron job](https://github.com/samvera/hyrax/blob/main/chart/hyrax/templates/cron-embargo.yaml), and [EmbargoExpiryJob](https://github.com/samvera/hyrax/blob/main/app/jobs/embargo_expiry_job.rb) (which is nearly the same as the rake task). I used the rake task on `main` today and it properly expired the embargoed works! 🎉

Issue #6211, opened 2023-08-24
New version of file cannot be downloaded


rjkati:
### Descriptive summary

When I upload a new version of a file on pg.nurax, the upload process completes in the UI, but the new version of the file cannot be downloaded.

### Steps to reproduce the behavior in User Interface (UI)

1. Deposit a new work on pg.nurax or navigate to a work which you've previously deposited
2. On the work show page for your deposited work, scroll down to the Items list and click on the file name
3. Click the `Edit This File Set` button
4. Click the `Versions` tab
5. Click the green `Choose New Version File` button
6. Select a new file and click `Open`
7. Click the `Upload New Version` button.
8. Observe that the user activity log on the file set page has been updated to reflect your action.
9. Click the Download link and observe that the original file is downloaded, rather than the revised version.

### Actual behavior (include screenshots if available)

User activity section on the fileset page reflects that a new version has been uploaded, but download button downloads the original version.

### Acceptance Criteria/Expected Behavior

After uploading the new version of a file, the user activity section on the fileset page should reflect that a new version has been uploaded and the download button should download the new version.

tpendragon:
@marrus-sh is going to add a link to a PR that added some versioning things to the valkyrie branch of this code to help guide implementation.

tpendragon:
There might be missing API in Valkyrie for this - we just need to know what that API should be and we can get it added.

tpendragon:
We'll have to repeat QA of versioning features right after implementing this.

marrus-sh:
Here is the PR where I previously worked on versioning in a Valkyrie context: https://github.com/samvera/hyrax/pull/5886

Ultimately versioning support is provided by the file adapter, which needs `supports?(:versions)` and an implementation of a `find_versions` method. The fallback in Valkyrie if no versions are available is `resource.file_identifier`.

It’s surprising to me that №8 in the steps to reproduce works successfully in PG nurax. Does the nurax PG adapter actually support versioning, or is this effectively a no‐op despite being logged? I think getting some tests to ensure that uploading a new version actually does something on the postgres backend would be a good place to start.

tpendragon:
It's our feeling that a variety of the versioning stuff in Hyrax w/ Valkyrie just silently fails.

There's stuff for reading versioning and stuff for saving versioning - a lot of that is currently missing regardless of which storage adapter you're using. Don't know if Wings has saving versions - it has support for reading versions.

[...]

PR #6239, opened 2023-08-29
Add method to VersioningService to detect support

marrus-sh:
Right now `Hyrax::VersioningService` does some silent coersion of versioning information in the case that versioning is not supported by the storage adapter. But, it might be good to actively provide a different UI to users in some cases depending on whether multiple versions of a file can actually be stored.

This commit adds a `supports_multiple_versions?` method to `Hyrax::VersioningService` which returns `true` iff the `resource` is non‐nil and not a `FileMetadata` object, or if it is a `FileMetadata` object and the storage adapter supports versioning.

This check requires having both a resource and a storage adapter in‐hand. My guess is that this is reasonable, but if we need to detect versioning support more generally (without a specific resource), a different solution may be desirable (altho
`Hyrax.storage_adapter.try?(:"supports?", :versions)` works in this case.)

---

Related to #6211 in the sense of being a prerequisite for disabling this functionality in the UI in cases where it will definitely not work.

### DIFF
--- a/app/services/hyrax/versioning_service.rb
+++ b/app/services/hyrax/versioning_service.rb
@@ -32,14 +32,10 @@ module Hyrax
     # If the resource is nil, or if it is a Hyrax::FileMetadata and versioning
     # is not supported in the storage adapter, an empty array will be returned.
     def versions
-      if resource.nil?
+      if !supports_multiple_versions?
         []
       elsif resource.is_a?(Hyrax::FileMetadata)
-        if storage_adapter.try(:"supports?", :versions)
-          storage_adapter.find_versions(id: resource.file_identifier).to_a
-        else
-          []
-        end
+        storage_adapter.find_versions(id: resource.file_identifier).to_a
       else
         return resource.versions if resource.versions.is_a?(Array)
         resource.versions.all.to_a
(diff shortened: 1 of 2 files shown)


app/services/hyrax/versioning_service.rb
```diff
@@ -53,6 +49,16 @@ def latest_version
...
     end
 
+    ##
+    # Returns whether support for multiple versions exists on this
+    # +Hyrax::VersioningService+.
+    #
+    # Versioning is unsupported on nil resources or on Valkyrie resources when
+    # the configured storage adapter does not advertise versioning support.
+    def supports_multiple_versions?
+      !(resource.nil? || resource.is_a?(Hyrax::FileMetadata) && !storage_adapter.try(:"supports?", :versions))
```
no-reply, line 59: is the assumption that if it's not a `FileMetadata` it's an AF thing?

marrus-sh: This is the existing assumption in the codebase, yeah. This is just copying the behaviour of the previous `VersioningService` implementation (which I wrote), but my recollection is that I was following patterns I saw elsewhere.

no-reply: seems good. just making sure we weren't trying to support other valkyrie models

no-reply (approved):

PR #6241, opened 2023-08-29
refine Hyrax::EmbargoSearch behavior to find only enforced embargoes

no-reply:
since sometime in 2014, `Hyrax::EmbargoSearch` has looked for ALL resources that have ANY embargo date when determining which objects are under embargo. the actual embargo object metadata is indexed (under both AF and Valkyrie), so there's no need to be this aggressive about it.

for the UI, find only the objects that are actually under a currently enforced embargo (according to the index). as a first pass at implementing this, we find all the objects that have indexed embargo dates, then drop any that aren't currently being enforced (where the indexed visibility differs from the embargo's prescribed visibility).

we could probably improve upon this, but this should suit as a first pass.

related to ##6235

* change how the Hyrax UI determines which objects are under embargo from the index to match `Hyrax::EmbargoManager` definitions
* ensures `EmbargoManager#release` sets `#embargo_history` correctly; after running this method, it's necessary to save both the Embargo and the AccessControl to persist the changes.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/concerns/hyrax/solr_document_behavior.rb
+++ b/app/models/concerns/hyrax/solr_document_behavior.rb
@@ -91,9 +91,9 @@ module Hyrax
     end
 
     def visibility
-      @visibility ||= if embargo_release_date.present?
+      @visibility ||= if embargo_enforced?
                         Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_EMBARGO
-                      elsif lease_expiration_date.present?
+                      elsif lease_enforced?
                         Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_LEASE
                       elsif public?
                         Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
@@ -106,6 +106,23 @@ module Hyrax
 
     def collection_type_gid
       first(Hyrax.config.collection_type_index_field)
+    end
+
+    def embargo_enforced?
+      return false if embargo_release_date.blank?
+
+      indexed_embargo_visibility = first('visibility_during_embargo_ssim')
+      # if we didn't index an embargo visibility, assume the release date means
+      # it's enforced
+      return true if indexed_embargo_visibility.blank?
+
+      # if the visibility and the visibility during embargo are the same, we're
+      # enforcing the embargo
+      self['visibility_ssi'] == indexed_embargo_visibility
+    end
+
+    def lease_enforced?
+      lease_expiration_date.present?
     end
 
     private
--- a/app/actors/hyrax/actors/embargo_actor.rb
+++ b/app/actors/hyrax/actors/embargo_actor.rb
@@ -14,12 +14,9 @@ module Hyrax
       def destroy
         case work
         when Valkyrie::Resource
-          embargo_manager = Hyrax::EmbargoManager.new(resource: work)
-          return if embargo_manager.embargo.embargo_release_date.blank?
-
-          embargo_manager.deactivate!
-          work.embargo = Hyrax.persister.save(resource: embargo_manager.embargo)
-          Hyrax::AccessControlList(work).save
+          Hyrax::EmbargoManager.deactivate_embargo_for(resource: work) &&
+            Hyrax.persister.save(resource: work.embargo) &&
+            Hyrax::AccessControlList(work).save
         else
           work.embargo_visibility! # If the embargo has lapsed, update the current visibility.
           work.deactivate_embargo!
--- a/app/presenters/hyrax/embargo_presenter.rb
+++ b/app/presenters/hyrax/embargo_presenter.rb
@@ -23,5 +23,9 @@ module Hyrax
     def embargo_history
       solr_document['embargo_history_ssim']
     end
+
+    def enforced?
+      solr_document.embargo_enforced?
+    end
   end
 end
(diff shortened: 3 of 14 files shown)


spec/controllers/hyrax/embargoes_controller_spec.rb
```diff
@@ -12,6 +12,7 @@
         expect(response).to redirect_to root_path
       end
     end
+
```
no-reply, line 15:
```suggestion
```

no-reply:
if this looks good. similar changes are needed for Lease.

no-reply:
@alishaevn if this solves the UI problem for #6235, i'd suggest we move the `embargo_history` behavior from `#deactivate!` into `#release` (i think we want this history data anytime an embargo is released), and deprecate that method for removal.

app/models/concerns/hyrax/solr_document_behavior.rb
```diff
@@ -91,9 +91,9 @@ def creator
     end
 
     def visibility
```
alishaevn, line 93: > if this solves the UI problem for #6235, i'd suggest we move the `embargo_history` behavior from `#deactivate!` into `#release` (i think we want this history data anytime an embargo is released), and deprecate that method for removal.

I do think that this may solve the UI problem, but not the initial problem I was trying to solve. which is that running `hyrax:embargo:deactivate_expired` does **not** remove the embargo. in my testing yesterday, I manually ran `#release_embargo_for` on a work then checked the work show page and the embargo was still applied. as well as the work still showing on the "/embargoes" page. that's why the nullifying and saving came into play. nullifying the embargo may not be required, I added it to follow the `deactivate!` pattern. either way, I think the saving of the work/embargo is required. which may be in part to the fact that embargoes are now referenced on a work with an `embargo_id` attribute, and not saved on the work itself. this was done to be in line with how valkyrie handles embargoes. (more about that can be found in the code comment for #6098)

I'm fine with moving the history setting into `#release`. (I considered making that change on #6235 but wasn't sure if it was desired since history wasn't being set at all until #6098). however, I think that deprecating `deactivate!` will lead us to having the same problem of not being able to deactivate an embargo in valkyrie. that's why I created it. (mimicking behavior in hydra-head). I'm open to other ideas, but any changes that are made on that front would need to be tested to ensure that the intended functionality for #5844 and #6126 still work. (I won't have the time today to test this out and I'm OOO the rest of the week)

no-reply: i'm pretty confused. i had understood that the problem you were trying to solve was that releasing the embargo in `EmbargoManager` doesn't remove it from the UI's embargo management table.

it seems to me that the existing `EmbargoManager#release` tests ensure that `#release` *does* in fact end enforcement of the embargo (in that it changes the object visibility to be the correct post-embargo visibility). what i found was that the UI was not checking whether the embargo is actually enforced, just looking for any indexed embargo release date. that doesn't seem adequate, so this proposes patching it.

if there's another problem, i think it would be helpful if we could work on getting a test in that demonstrates the issue.

alishaevn: I might be able to look into adding some tests next week, depending on what my schedule looks like when I return. however, https://github.com/scientist-softserv/atla-hyku/issues/121 is original problem I was aiming to solve. 
when I tested the rake tasks against hyrax `main` yesterday, they didn't return 400 errors, but they didn't remove the embargo either. my intention was to fix that problem in hyrax, then move it to atla.

no-reply: >  the work still showed on the "/embargoes" page

this is the main issue this tries to patch. the problem, as i see it, is that *even if* the embargo is released (i.e. the correct post-embargo permissions are applied, the public can access the object, etc...), the `/embargoes` page continues to list any object that has an indexed `#embargo_release_date`. this was a shortcut implementation of the `Hyrax::EmbargoService` dating at least 2014.

i started out by trying to write a test for the `EmbargoesControllerBehavior` that demonstrated the issue, but found that it doesn't actually populate a list of embargoed objects. instead, it renders the view, which calls `EmbargoHelper`, and then the `EmbargoService`. this last had tests that explicitly ensured the behavior above, so i rewrote them to ensure the correct behavior instead: `#assets_under_embargo` should return only assets where the embargo is currently being enforced, not just ones that have an `embargo_release_date` attached.

i think the reason we've been chasing this issue around the code base is that we've not paid attention to this root cause. there may still be other issues, but i think it's necessary to ensure the UI reflects the actual object state before we try to fix the problem by manipulating object state.

----

> and still had the "embargo" badge on the show page and on "/dashboard/my/works"

i think the changes to `SolrDocument` should address this; though admittedly i didn't analyze this as closely. i just noticed that it was using the same shorthand logic that the `EmbargoService` was using, and it was helpful for both fixes to resolve it in `SolrDocument`.

----

if this doesn't get us on the same page, can we jump on a call to chat about it. i'm concerned we're talking past each other.

no-reply:
@alishaevn i added tests for the rake task in https://github.com/samvera/hyrax/pull/6241/commits/08ebfe051260f72be326c4ed9fe14e1a8e3b2d9e

no-reply:
just want to bump this to see if we can resolve this issue.

dlpierce (approved):
Let's get this merged and do some more QA on it, see where that leaves us on the other behavior brought up in this discussion.

alishaevn:
I tested this locally and it's working! (ref: [this comment](https://github.com/samvera/hyrax/pull/6235#issuecomment-1735836042))

the same will need to be done for leases.

PR #6245, opened 2023-08-29
mark JobIoWrapper specs as ActiveFedora only

no-reply:
gratefully, Valkyrie does not use this thing!

@samvera/hyrax-code-reviewers


spec/models/job_io_wrapper_spec.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-RSpec.describe JobIoWrapper, type: :model do
+RSpec.describe JobIoWrapper, type: :model, :active_fedora do
```
no-reply, line 2:
```suggestion
RSpec.describe JobIoWrapper, :active_fedora, type: :model do
```

dunn (approved):

PR #6247, opened 2023-08-29
mark Hyrax Callbacks tests as :active_fedora

no-reply:
Hyrax::Callbacks is deprecated and only used by ActiveFedora apps

@samvera/hyrax-code-reviewers


spec/config/hyrax_events_spec.rb
```diff
@@ -1,5 +1,5 @@
-# frozen_string_literal: true
-RSpec.describe "hyrax_events using Hyrax callbacks" do
+# frozen_string_literal: truep
```
dlpierce, line 1:
```suggestion
# frozen_string_literal: true
```

dlpierce (approved):

PR #6257, opened 2023-08-30
Hide versioning UI when unsupported by adapter

marrus-sh:
- Modify `Hyrax::VersionListPresenter` to wrap a `Hyrax::VersioningService`, instead directly wrapping an array of version presenters. This enables querying whether multiple versions are supported directly on the version list presenter by delegating it to the wrapped service. I also think this is a bit simpler from a code perspective.

- Suppress the UI for manually creating a new version of a file when the storage adapter does not support multiple versions. Note: Currently dassie supports multiple versions, but koppie does not.

- Suppress the UI for restoring a previous versions when the version list is empty. The fallback behaviour for the version list (i·e when versioning is unsupported) is an empty list, so this is effectively the same as the previous check, but the semantics are slightly different.

The end result of these changes is that the “Versions” tab is entirely empty in koppie (currently, with no versioning supported). We may want to either hide the tab in its entirety or else put a message there stating that versioning is not supported.

---

Related to #6211 in the sense that it prevents the confusing behaviour of users *appearing* able to upload new versions when versions are not actually supported. Probably we also want some work to prevent logging of new versions that aren’t actually being created (step №8 in that issue), but this PR handles the user‐facing side of things.

* In an application with versioning enabled (e·g dassie), no change in behaviour is observable
* In an application with versioning disabled (e·g koppie), the versioning tab is empty and users cannot upload new versions

### DIFF
--- a/app/presenters/hyrax/version_list_presenter.rb
+++ b/app/presenters/hyrax/version_list_presenter.rb
@@ -5,17 +5,19 @@ module Hyrax
   class VersionListPresenter
     include Enumerable
 
+    attr_reader :versioning_service
+
     ##
-    # @param version_list [Array<#created>]
-    def initialize(version_list)
-      @raw_list = version_list
+    # @param service [Hyrax::VersioningService]
+    def initialize(service)
+      @versioning_service = service
     end
 
     ##
     # @param [Object] an object representing the File Set
     #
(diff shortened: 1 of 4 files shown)


no-reply (approved):

no-reply:
added `notes-major` since this changes the public API for initializer on the VersionPresenter

PR #6259, opened 2023-08-30
test behavior after a published Fixity check failure on Valkyrie

no-reply:
if a Valkyrie Fixity service (to be invented) fails, it should be able to
publish an event to be picked up by `FileSetLifecycleNotificationListener` and a
notification can be published.

this updates the tests for the fixity failure notification path to avoid testing
AF behavior when Wings is disabled, and to add support for Valkyrie to the code
path.

* fixes a bug in the Fixity failure notifications that always reported the `original_file` failed, even when another file's fixity was checked
* adds support for Valkyrie to fixity failure notification code paths

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/fixity_check_failure_service.rb
+++ b/app/services/hyrax/fixity_check_failure_service.rb
@@ -14,7 +14,7 @@ module Hyrax
     end
 
     def message
-      uri = file_set.original_file.uri.to_s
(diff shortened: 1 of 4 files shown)


dlpierce (approved):
Glad to see :with_file for the FileMetadata factory!

Issue #5798, opened 2022-08-16
Show all collection descriptive metadata


jlhardes:
### Descriptive summary

Descriptive metadata fields for collections are not all showing in public view on valkyrized Hyrax. It might be that only fields that are set to be indexed appear on the show page.

### Rationale

All descriptive metadata for collections should appear if it is filled in. Descriptive metadata fields on collection are the same as on Works by default and while this may not be how collection metadata often works, it seems the best way to manage descriptive metadata for application development purposes.

### Expected behavior

All metadata fields that can be edited on Description tab of Collection will show on public view of collection (show page).

### Actual behavior

Only Title, Description, Resource type, Creator, Keyword, Date Created, Subject, and Related URL are showing. 

### Steps to reproduce the behavior

1. As a logged in user, create a new User Collection
2. On Description tab, fill in metadata on all fields 
3. Click Save changes
4. Click Cancel to view Collection in dashboard (see only fields listed above)
5. Click Public view of Collection and see only fields listed above

### Related work

PR [#5157](https://github.com/samvera/hyrax/pull/5157) - Update PcdmCollection metadata to match existing ::Collection metadata
Issue #5494 - Edit collection only shows title field for Hyrax::PcdmCollection (closed)
Issue #5799 - Enable Geonames functionality for Location field in Collections

rjkati:
Verified that collection descriptive metadata now displays in public view on nurax-pg

![collection_metadata](https://github.com/samvera/hyrax/assets/32644254/99f388d3-85e0-4fce-97a6-50471d617a27)

PR #6267, opened 2023-08-30
Display all descriptive metadata for collections on koppie.

### LINKED ISSUE #6208: Revert #6103 and fix collection display generally for Valkyrie
### Descriptive summary

#6103 fixed collection displays in Koppie, but new installs will not have all the Collection changes that were introduced there. Those changes should happen in PcdmCollection.

We should revert 6103 and fix the issue.

## Success Criteria

Collection specs pass (or https://github.com/samvera/hyrax/pull/5541 passes?) in Koppie without there being modifications to the Collection model in Koppie.

This might be blocked by getting specs passing for Koppie in CI?

lsitu:
### Fixes

Fixes #6208; refs [#5798](https://github.com/samvera/hyrax/issues/5798)

### Summary
Index all form fields for collection display on koppie.

### Screenshots
Before changes
Screen Shot - koppie collection display - before changed

After change:
Screen Shot - koppie collection display - after changed

@samvera/hyrax-code-reviewers

### DIFF
--- a/.koppie/config/metadata/collection_resource.yaml
+++ b/.koppie/config/metadata/collection_resource.yaml
@@ -40,21 +40,29 @@ attributes:
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "rights_statement_tesim"
   abstract:
     type: string
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "abstract_tesim"
   access_right:
     type: string
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "access_right_tesim"
   alternative_title:
     type: string
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "alternative_title_tesim"
   based_near:
     type: string
     multiple: true
@@ -71,6 +79,8 @@ attributes:
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "contributor_tesim"
   date_created:
     type: date_time
     multiple: true
@@ -83,6 +93,8 @@ attributes:
     multiple: true
     form:
       primary: false
+    index_keys:
+      - "identifier_tesim"
   import_url:
     type: string
   keyword:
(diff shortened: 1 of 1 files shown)


lsitu:
@no-reply I think the question I raised during DSU this morning is more about what kind of actions we want in the next step. The PR will index all missing form fields for the collection model in valkyrie, and we got a lot more fields showing up once the PR is merged. However, the fields displayed in the collection show page are defined by the static terms in [collection_presenter.rb#L45-L47](https://github.com/samvera/hyrax/blob/main/app/presenters/hyrax/collection_presenter.rb#L45-L47), and we may need to update it if we want to display other fields like `label` in [collection_resource.yaml#L101-L109](https://github.com/samvera/hyrax/blob/main/.koppie/config/metadata/collection_resource.yaml#L101-L109) for koppie. What do we want to do in this case, override the terms in hyrax, in koppie, or handle it dynamically to include all the definitions in [ollection_resource.yam](https://github.com/samvera/hyrax/blob/main/.koppie/config/metadata/collection_resource.yam)?

no-reply:
> @no-reply I think the question I raised during DSU this morning is more about what kind of actions we want in the next step. The PR will index all missing form fields for the collection model in valkyrie, and we got a lot more fields showing up once the PR is merged. However, the fields displayed in the collection show page are defined by the static terms in [collection_presenter.rb#L45-L47](https://github.com/samvera/hyrax/blob/main/app/presenters/hyrax/collection_presenter.rb#L45-L47), and we may need to update it if we want to display other fields like `label` in [collection_resource.yaml#L101-L109](https://github.com/samvera/hyrax/blob/main/.koppie/config/metadata/collection_resource.yaml#L101-L109) for koppie. What do we want to do in this case, override the terms in hyrax, in koppie, or handle it dynamically to include all the definitions in [ollection_resource.yam](https://github.com/samvera/hyrax/blob/main/.koppie/config/metadata/collection_resource.yam)?

ah! i think we should leave the presenter code as is for now, and consider improving it out of scope for this sprint. it could be worth opening a ticket though!

no-reply (approved):

PR #6268, opened 2023-08-30
fixes all but one of the test errors for collection type

sdellis:
@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/collection_type_spec.rb
+++ b/spec/features/collection_type_spec.rb
@@ -293,7 +293,7 @@ RSpec.describe 'collection_type', type: :feature do
 
     context 'when collections exist of this type' do
       let!(:collection1) do
-        FactoryBot.create(:public_collection_lw, user: build(:user), collection_type: exhibit_collection_type)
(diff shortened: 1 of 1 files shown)


no-reply:
@sdellis it seems like the build is failing due to `NameError: uninitialized constant Hydra::RoleManagement`, can you address that?

no-reply:
i think if the `Hydra::RoleManagement::UserRoles` changes are removed, and this is rebased on `main` it will pass again

sdellis:
@no-reply I reverted the Dassie User model, rebased to main, and pushed it.

.koppie/app/models/user.rb
```diff
@@ -1,4 +1,7 @@
 class User < ApplicationRecord
+  # Connects this user object to Role-management behaviors.
+  include Hydra::RoleManagement::UserRoles
```
hackartisan, line 3: does this need to be added in `lib/generators/hyrax/models_generator.rb`?

hackartisan: Oh, this was removed in https://github.com/samvera/hyrax/pull/6272/files. Should it be added back in here?

no-reply: it shouldn't, i think

hackartisan (approved):

PR #6272, opened 2023-08-30
remove Hyrax::RoleManagement from `.koppie`

no-reply:
including this makes the whole application very different from `.dassie`, in a way that complicates the test suite substantially.

@samvera/hyrax-code-reviewers


dlpierce (approved):
This is fine. I expect it will be retained for nurax pg.

Issue #5522, opened 2022-03-11
Empty collections displayed when clicking "View collections of this type" after trying deleting non-empty collection type


MPLSFedResearchTZ:
### Descriptive summary

When clicking the button "View collections of this type" it should display all visible collections of this collection type.

![image](https://user-images.githubusercontent.com/93392603/157973604-34eb8287-2f62-463b-baea-e5748ff3b684.png)

### Steps to reproduce the behavior

1. Go to Settings-> Collection types to get a list of collections types.
2. Deleting a collection type that contains some collections of this type.
3. The pop-up box shows up which contains two buttons, "Cancel" and "View collections of this type"
4. Click the "View collections of this type"
5. The user is supposed to see all the collections of this type but it shows an empty list.

eliotjordan:
I'm unable to replicate this ticket. @elrayle Is it possible some of the collection work that's already been done has fixed this?

MPLSFedResearchTZ:
The issue is still there. I reproduced it by deleting this "test collection type" collection type I created.
![image](https://user-images.githubusercontent.com/93392603/159024284-ecce74b7-69d6-434b-9b31-33accf50e927.png)

eliotjordan:
@MPLSFedResearchTZ Couple of questions:

1. Are you testing this using the Dassie or EngineCart test apps?
2. Have you pulled in the latest commits on the main branch?
3. If so, try creating a new collection type and a new collection, and then test the delete.

eliotjordan:
@MPLSFedResearchTZ nurax-pg is ~55 commits behind Hyrax main. I'll see about getting that updated and deployed.

elrayle:
I updated to latest Hyrax and this still fails on nurax-pg.  The problem is the filtering.  It filters by `Collection Type > nesting only` && `Collection Type > Collection`.  The second filter is actually the `has_model_ssim` which is `Hyrax::PcdmCollection` on nurax-pg.  I don't see the need for the second filter.  Since the results are displayed on `Dashboard -> Collections` which already filters to collections, this seems redundant.

jlhardes:
Remaining work listed in [#3820](https://github.com/samvera/hyrax/issues/3820#issuecomment-880084599) - Replace calls to ActiveFedora::Base.where (removes 7 refs to ActiveFedora) might be needed for this issue

jessicahilt:
Let's try to reproduce this problem first. If reproducible, cross-check against in 3820.

[...]

PR #6273, opened 2023-08-31
Don't force "Collection" model filter from the view 

no-reply:
### Fixes

Fixes a Valkyrie-only bug where collections are filtered en-masse from the Collection listing because they are indexed from a model other than the AF `Collection`

refs #5522 

### Summary

the search builders for the target controller already search effectively for the
model class, and may have better strategies for doing so in the future. adding a
filter here prevents applications from using Collection models not named
"Collection".

* test as in issue #5522 

* adds some controller tests (it turns out these are irrelevant to the change, but improve testing anyway)
* Fix bug: "Empty collections displayed when clicking "View collections of this type" after trying deleting non-empty collection type"

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/my/collections_controller.rb
+++ b/app/controllers/hyrax/my/collections_controller.rb
@@ -37,6 +37,8 @@ module Hyrax
         hyrax.my_collections_url(*args)
       end
 
+      # @api public
+      #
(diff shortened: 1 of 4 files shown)


mcritchlow (approved):
Thank you! This was so confusing

PR #6274, opened 2023-08-31
use Valkyrie for test setup on Collections:PremissionsService specs

### LINKED ISSUE #6254: Ensure `spec/services/hyrax/collection_types/permissions_service_spec.rb`  passes on `.koppie`
`spec/services/hyrax/collection_types/permissions_service_spec.rb is failing on koppie builds. there may be a number of reasons for this:

  1. the test setup uses ActiveFedora, and should be refactored to use Valkyrie;
  2. the tests themselves test behavior that only applies to ActiveFedora, and should be turned off by adding the `:active_fedora` rspec metadata;
  3. the unit under test depends on ActiveFedora and needs to be rewritten to use Valkyrie throughout;
      * if this involves adding a switch for AF / Valkyrie behavior, ensure both branches are covered by adding another test
  5. some combination of the above.

----

```sh
docker-compose -f docker-compose-koppie.yml build
docker-compose -f docker-compose-koppie.yml up
docker-compose -f docker-compose-koppie.yml exec -w /app/samvera/hyrax-engine app sh -c "bundle install"
docker-compose -f docker-compose-koppie.yml exec -w /app/samvera/hyrax-engine app sh -c "bundle exec rspec spec/services/hyrax/collection_types/permissions_service_spec.rb"
```

no-reply:
remove a lot of AF stubs and replace them with real setup using valkyrie.

Fixes #6254

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/collections/permissions_service_spec.rb
+++ b/spec/services/hyrax/collections/permissions_service_spec.rb
@@ -1,304 +1,382 @@
 # frozen_string_literal: true
 RSpec.describe Hyrax::Collections::PermissionsService do
+  subject(:service) { described_class }
   let(:user) { FactoryBot.create(:user) }
   let(:ability) { Ability.new(user) }
 
-  context 'collection specific methods' do
-    let(:collection) { FactoryBot.build(:collection_lw, id: 'collection_1') }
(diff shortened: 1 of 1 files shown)


hackartisan (approved):
this is a nice improvement to this test. I ran it locally and it passes now.

PR #6285, opened 2023-08-31
port most FileSetPresenter specs to Valkyrie

no-reply:
this passes most FileSetPresenter specs, but reveals a potential issue that i haven't yet been able to determine the correct behavior about.

it changes the presenter's #to_partial_path from `"file_sets/file_set"` to `"hyrax/file_sets/file_set`". the behavior in the original `::FileSet` is provided by `Hyrax::Naming`.

`PcdmCollection` uses a custom name class `Hyrax::CollectionName` to ensure ActiveModel naming behavior is compatible with `::Collection`, and it seems natural to assume `Hyrax::FileSet` should do similarly. BUT, why is this only coming up now? `.koppie` and `nurax-pg` are passing many automated and manual tests using the current `#model_name` implementation. Why?

maybe it's okay to just change the test and allow that the partial path change for Valkyrie native apps? what's going on here?

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/file_set.rb
+++ b/app/models/hyrax/file_set.rb
@@ -60,9 +60,42 @@ module Hyrax
     self.characterization_proxy = Hyrax.config.characterization_proxy
 
     attribute :file_ids, Valkyrie::Types::Array.of(Valkyrie::Types::ID) # id for FileMetadata resources
-    attribute :thumbnail_id, Valkyrie::Types::ID.optional # id for FileMetadata resource
-    attribute :original_file_id, Valkyrie::Types::ID.optional # id for FileMetadata resource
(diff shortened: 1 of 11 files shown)


dlpierce:
I updated the spec for `to_partial_path`. Looking at similar specs, they all have the "hyrax" prefix, so this was the odd one out really. My guess is it worked because the relevant partials are already in a hyrax directory.

dlpierce:
Trying to manage the use specific ids (`original_file_id`, etc) on Hyrax::FileSet proved to be very cumbersome when setting up tests, and was effectively duplicated data anyways. I've replaced them with methods that lookup the file using the use specific custom queries.

dlpierce (approved):

PR #6286, opened 2023-09-01
Koppie no wings

no-reply:
ughhh ughhh debug


dlpierce:
Wings is no longer loaded at all when disabled as of https://github.com/samvera/hyrax/pull/6312

PR #6299, opened 2023-09-06
do not fall back to generic resourceform if there is a syntax error

orangewolf:
### Summary

do not fall back to generic ResourceForm if there is a syntax error in your work type Form object. Before this commit, any method missing or other name error in GenericWorkForm would simply cause the app to use ResourceForm(GenericWork) instead. This hides the error from you while not giving you the features you'd expect. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -123,20 +123,17 @@ module Hyrax
         def for(resource)
           klass = "#{resource.class.name}Form".safe_constantize
 
-          if klass
-            klass.new(resource)
+          return  klass.new(resource) if klass
+          case resource
+          when Hyrax::AdministrativeSet
+            Hyrax::Forms::AdministrativeSetForm.new(resource)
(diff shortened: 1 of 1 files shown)


app/forms/hyrax/forms/resource_form.rb
```diff
@@ -121,18 +121,22 @@ class << self
...
-          case resource
-          when Hyrax::AdministrativeSet
-            Hyrax::Forms::AdministrativeSetForm.new(resource)
-          when Hyrax::FileSet
-            Hyrax::Forms::FileSetForm.new(resource)
-          when Hyrax::PcdmCollection
-            Hyrax::Forms::PcdmCollectionForm.new(resource)
+          klass = "#{resource.class.name}Form".safe_constantize
+
+          if klass
```
no-reply, line 126: maybe `return  klass.new(resource) if klass`, to shrink the method by a few lines and satisfy rubocop, (i think this is more readable, too)

no-reply (approved):

PR #6300, opened 2023-09-07
make valkyrie solr adapter take full url argument

orangewolf:
### Summary

Previously Valkyrie solr could only be set with host, port and core. This prevents Valkyrie from being able to work with authenticated Solr or really any Solr that doesn't follow the strict pattern. Long term, I feel like the Valkyrie config should go away and we should just support the same options as the blacklight.yml already supports. Specifying Solr twice seems unnecessary, but this at least provides a solution which is good enough for production.

* Set a `config/valkyrie_index.yml` that looks like
```yaml
development:
  adapter: solr
  url: <%= ENV['SOLR_URL'] + 'hydra-development' || "http://127.0.0.1:#{ENV.fetch('SOLR_DEVELOPMENT_PORT', 8983)}/solr/hydra-development" %>
test: &test
  adapter: solr
  url: <%= ENV['SOLR_URL'] ? ENV['SOLR_URL'] + 'hydra-test' : "http://127.0.0.1:#{ENV.fetch('SOLR_TEST_PORT', 8985)}/solr/hydra-test" %>
production:
  adapter: solr
  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/blacklight-core" %>
``` 
* Set SOLR_URL and start the app
* It should be able to see Solr

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/valkyrie/indexing/solr/indexing_adapter.rb
+++ b/lib/valkyrie/indexing/solr/indexing_adapter.rb
@@ -81,6 +81,8 @@ module Valkyrie
           # if any configuration is missing, derive it from Blacklight
           config = blacklight_based_config.with_indifferent_access.merge(config)
 
+          return config['url'] if config['url'].present?
+
           "http://#{config['host']}:#{config['port']}/solr/#{config['core']}"
         end
 


eliotjordan (approved):
Looks good to me. 👍

dlpierce (commented):
Why not use the blacklight config's url directly now? Is there a good reason to not use the same solr config for both?

orangewolf:
@dlpierce not from my perspective - but I felt like it was presumptuous since someone did work to get it to behave this way. @no-reply do you have reasons for them being seperate?

dlpierce:
I'll point out koppie has been configured to use the same core in both config files.

no-reply:
the idea was that one might want to run Blacklight against the index/core being written to by AF, and index valkyrie records elsewhere until the index was caught up, then change the blacklight config.

in principle, i don't think it's nice to make it too hard to separate the configurations. we're configuring two different things, so i should have the option to configure them separately if that's convenient for me. but i don't see any reason the default shouldn't be to use the blacklight config.

PR #6323, opened 2023-09-17
valkyrie resource affordances for find and save methods.

orangewolf:
### Summary

This PR adds `Hyrax::Resource.find`, `Hyrax::Resource.find_by`, `Hyrax::Resource#save`, `Hyrax::Resource#create`, `Hyrax::Resource#update` and their bang equivalents.

These 'nice to have' methods make it easier to do simple manipulations of resources and act as a more familiar short hand for developers who are less familiar with the `Hyrax.persister`, `Hyrax.query_service` or `Hyrax.index_adapter`.

**DRAFT** the goal of this PR is to get feedback from the community before too much work goes in to specs and comprehensive functionality. Do we want these? I sure do ;-) 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/resource.rb
+++ b/app/models/hyrax/resource.rb
@@ -59,7 +59,73 @@ module Hyrax
       def _hyrax_default_name_class
         Hyrax::ResourceName
       end
+
+      ##
+      # find a Valkyrie object by its primary identifyer.
+      #
+      # @param id [String]
+      # @return [Valkyrie::Resource]
+      def find(id, query_service: Hyrax.query_service)
+        query_service.find_by(id: id)
+      end
+
+      ##
+      # find and item by an arbitrary keyword arguements if and only if that property is supported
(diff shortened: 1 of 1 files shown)


dlpierce:
@tpendragon can chime in too, but my understanding is these convenience methods are purposefully not available in order to follow the data mapper pattern. See https://github.com/samvera/valkyrie/wiki/Saving-a-resource#persistence

dlpierce:
If we adopt tamsin's suggestion, I want them to live in a separate included file so it's clear which methods are implicated, and add rubocop rules to prevent their use within Hyrax more strongly than documentation suggestions.

orangewolf:
These seem like reasonable asks. Will let a few more folks weigh in before I do refactoring and testing. 

Tbh I use them in the console more than anything else and I got tired of the persistent save, indexer save, event send

PR #6325, opened 2023-09-19
dassie/koppie alignment and added bundle caching

dlpierce:
### Summary
dassie/koppie alignment and added bundle caching

* dassie and koppie build and run from clean compose environment. (Clean volumes with `docker compose down -v` and remove any Gemfile.*.lock files in hyrax root and .dassie/.koppie)

- Use the same images for dassie/koppie dependent services.
- Don't build an image for db_migrate, reuse the app image.
- Use `dev-entrypoint.sh` to copy gems from image build to `/app/bundle` volume to speed up runtime `bundle install`. Only set in compose to avoid affecting CircleCI build.
- `/bin` is volume mounted to allow easier dev of entrypoints, etc.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/.env
+++ b/.dassie/.env
@@ -1,5 +1,6 @@
 ANALYTICS_START_DATE=2021-08-21
 BUNDLE_GEMFILE=Gemfile.dassie
+BUNDLE_PATH=/app/bundle
 CHROME_HEADLESS_MODE=false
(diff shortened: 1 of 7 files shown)


alishaevn (approved):
I tested this locally and it's fixed the issue I was having [here](https://samvera.slack.com/archives/CA8ANGLEL/p1695138166507069).

PR #6326, opened 2023-09-19
🐛 Fix for Collection Sort_by functionality

ShanaLMoore:
Credit goes to Palni Palci for pointing out this bug + contributing the fix back.

### Summary

On the Collection show page, the sort_by dropdown button was not working. A user could select and option and a request would be made, but the sort params was not making it through to the search builder and therefore did not sort the items regardless of the selection.

## Accepted Criteria

- [ ] Works on the public view of the collection page are by default, sorted A-Z

## Testing Instructions

- [ ] Login in as an admin user
- [ ] Click on Collections. Find a collection with several works and click on it. Click to see the public view of it. 
- [ ] Change the selections of the sort_by drop down. The works' orders should change accordingly.

### Detailed Description

[DEMO](https://share.getcloudapp.com/RBuz7EQg)

* FIX TYPO? title_si => title_ssi. si 
  - title_si: The suffix _si suggests that this field should be treated as a "Short Integer" based on Solr's default dynamic field definitions.~
  - title_ssi: The suffix _ssi suggests that this field is a "String, stored, and indexed".
* Pass sort selection through the controller and onto the search builder to make it work.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/search_builders/hyrax/collection_search_builder.rb
+++ b/app/search_builders/hyrax/collection_search_builder.rb
@@ -20,7 +20,7 @@ module Hyrax
 
     # @return [String] Solr field name indicating default sort order
     def sort_field
-      "title_si"
+      "title_ssi"
     end
 
     # This overrides the models in FilterByType
--- a/app/services/hyrax/collections/collection_member_search_service.rb
+++ b/app/services/hyrax/collections/collection_member_search_service.rb
@@ -46,8 +46,10 @@ module Hyrax
       #
       # @return [Blacklight::Solr::Response]
       def available_member_works
+        sort_field = user_params[:sort] || DEFAULT_SORT_FIELD
         response, _docs = search_results do |builder|
           builder.search_includes_models = :works
+          builder.merge(sort: sort_field)
           builder
         end
         response
--- a/app/services/hyrax/collections/collection_member_search_service.rb
+++ b/app/services/hyrax/collections/collection_member_search_service.rb
@@ -46,7 +46,7 @@ module Hyrax
       #
       # @return [Blacklight::Solr::Response]
       def available_member_works
-        sort_field = user_params[:sort] || DEFAULT_SORT_FIELD
+        sort_field = user_params[:sort]
         response, _docs = search_results do |builder|
           builder.search_includes_models = :works
           builder.merge(sort: sort_field)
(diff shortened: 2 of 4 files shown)


ShanaLMoore:
I pushed this PR with assumptions. I didn't realize that more work would be needed. 

Hyrax currently doesn't have a sort by title functionality and does not appear to be indexing title anyways.

Is there a reason for the latter @dlpierce ? 

I can proceed with this fix but will remove the code related to titles.

app/search_builders/hyrax/collection_search_builder.rb
```diff
@@ -20,7 +20,7 @@ def discovery_permissions
 
     # @return [String] Solr field name indicating default sort order
     def sort_field
-      "title_si"
+      "title_ssi"
```
ShanaLMoore, line 23: Including this fix because I believe it's a typo. 


-     title_si: The suffix _si suggests that this field should be treated as a "Short Integer" based on Solr's default dynamic field definitions.~
-     title_ssi: The suffix _ssi suggests that this field is a "String, stored, and indexed".

spec/search_builders/hyrax/collection_search_builder_spec.rb
```diff
@@ -6,7 +6,7 @@
   let(:scope) { FakeSearchBuilderScope.new(current_user: user) }
 
   describe '#sort_field' do
-    its(:sort_field) { is_expected.to eq('title_si') }
+    its(:sort_field) { is_expected.to eq('title_ssi') }
```
ShanaLMoore, line 9: Including this fix because I believe it's a typo.

    title_si: The suffix _si suggests that this field should be treated as a "Short Integer" based on Solr's default dynamic field definitions.~
    title_ssi: The suffix _ssi suggests that this field is a "String, stored, and indexed".

app/services/hyrax/collections/collection_member_search_service.rb
```diff
@@ -46,8 +46,10 @@ def available_member_subcollections
       #
       # @return [Blacklight::Solr::Response]
       def available_member_works
+        sort_field = user_params[:sort]
```
ShanaLMoore, line 49: Previously the sort params were not getting through, so the sorting selections did nothing.

ShanaLMoore:
closing because it looks like this isn't a bug in hyrax after all! 🙃

PR #6330, opened 2023-09-27
refine Hyrax::LeaseService behavior to find only enforced leases

no-reply:
repeats the work in #6241, but with the lease infrastructure.

it would still be really nice to flatten these into one concept, but the data layer is still an issue (stored data has different names for functionally identical Embargo and Lease abstractions).

@samvera/hyrax-code-reviewers


dlpierce:
@no-reply Is there anything keeping this in draft status?

dlpierce (approved):

PR #6368, opened 2023-10-18
Use a hashed version identifier in IIIF manifest

### LINKED ISSUE #6352: Unable to restore previous version of a fileset on pg.nurax
### Descriptive summary

On pg.nurax, I am unable to restore a previous version of a file

1. On pg.nurax, navigate to a work for which you have edit access (or deposit a new work)
2. Scroll to the item view and click on the title of the file
3. Click "edit this file set"
4. Click "versions"
5. Click "Choose New version file", select a file, and click "Upload New Version"
6. Click "versions"
7. In the "Restore Previous Version" section, click the radio button next to the previous version and click "Save revision"
8. Open the work's show page in an incognito/private window and observe that the previous version of the file does not display.

Restoring a previous version of a file does not restore the previous version. Rather, the new version displays.

### Related work

#6234

dlpierce:
### Fixes

Fixes #6352

### Summary

Ensures the IIIF server can discern outdated image data in the event of a file revision.

* After uploading a new version of a file to a file set, the updated image is displayed in the viewer.

### Details

Despite its name, the Riiif::HttpFileResolver was capable of opening files available on a local filesytem. In doing so it would create its own copy of the file based on the id less any version data. In the event of a file revision, the cached file could not be easily cleared.

This prompted a new resolver that interfaces with valkyrie storage adapters more directly. This avoids the extraneous file copy, and attempts to prevent riiif from transforming files that have not finished copying via a Concurrent::ReadWriteLock.

The version portion of the image identifiers in the IIIF manifest is hashed to avoid potentially overly verbose version strings.

- Add RiiifFileResolver and RiiifFile to more efficiently use riiif with valkyrie storage adapters. Uses a read/write lock when loading a local file copy from storage.
- Rerun characterization/derivatives when reverting a file version.
- Include version in solr indexed original_file_id. This is now generated by Hyrax::FileSet#iiif_id instead of by the file set indexer.
- Patch Valkyrie's StreamFile #length in same manner as done for ::File.
- Set json content type for manifest response. (Browsers should render a pretty version of them now)
- The :test_disk storage adapter for specs is now a VersionedDisk adapter.

### DIFF
--- a/.dassie/config/initializers/riiif.rb
+++ b/.dassie/config/initializers/riiif.rb
@@ -1,12 +1,10 @@
 # frozen_string_literal: true
-ActiveSupport::Reloader.to_prepare do
-  Riiif::Image.file_resolver = Riiif::HttpFileResolver.new
+Rails.application.reloader.to_prepare do
   Riiif::Image.info_service = lambda do |id, _file|
     # id will look like a path to a pcdm:file
     # (e.g. rv042t299%2Ffiles%2F6d71677a-4f80-42f1-ae58-ed1063fd79c7)
     # but we just want the id for the FileSet it's attached to.
 
-    # Capture everything before the first slash
     fs_id = id.sub(/\A([^\/]*)\/.*/, '\1')
     resp = Hyrax::SolrService.get("id:#{fs_id}")
     doc = resp['response']['docs'].first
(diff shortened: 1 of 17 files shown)


.dassie/config/initializers/riiif.rb
```diff
@@ -35,5 +29,5 @@
   Riiif.not_found_image = Rails.root.join('app', 'assets', 'images', 'us_404.svg')
   Riiif.unauthorized_image = Rails.root.join('app', 'assets', 'images', 'us_404.svg')
 
-  Riiif::Engine.config.cache_duration = 365.days
+  Riiif::Engine.config.cache_duration = 1.day
```
abelemlih, line 32: What is the purpose of this change?

app/models/hyrax/file_set.rb
```diff
@@ -73,6 +73,15 @@ def original_file_id
       original_file&.id
     end
 
+    # @return [String, Nil] versioned identifier suitable for use in a IIIF manifest
+    def iiif_id
+      orig_file = original_file
+      return nil if orig_file.nil? || orig_file.file_identifier.blank?
+      latest_file = Hyrax::VersioningService.latest_version_of(orig_file)
+      version = latest_file&.version_id ? Digest::MD5.hexdigest(latest_file.version_id) : nil
+      "#{id}/files/#{orig_file.id}#{'/' + version if version}"
```
abelemlih, line 82: If version is `nil`, this would still return a string missing the version number. Is this the expected behavior?

.dassie/config/initializers/riiif.rb
dlpierce: For the dev environments the lower cache time will hopefully give any problems hidden by cached data a chance to surface. The generator for this file uses a month, which seems like a good default.

app/models/hyrax/file_set.rb
abelemlih: Would `"#{id}/files/#{orig_file.id}"` still be a valid `iiif_id` when `version` is `nil`, or would it be more accurate for the method to return `nil`?

dlpierce: I consider it valid. A IIIF server that is resolving requests for versionless ids would not know if its cached image times have become outdated due to a fileset revision, but that seems preferable to not working at all if an unversioned storage adapter is used.

abelemlih (approved):

PR #6392, opened 2023-10-26
#5098 Valkyrize FileSet Generator

### LINKED ISSUE #5098: App generated through template includes file_set model that is an AF::Base model
### Descriptive summary

The generated `app/model/file_set.rb` extends `ActiveFedora::Base`.  To move to Valkyrie, the generated file set model needs to be_a `Hyrax::Resource`.

It is unclear if the generated file set is needed.  Is there a reason not to just use `Hyrax::FileSet`?

### Rationale

All models should be_a `Hyrax::Resource` to work with Valkyrie persistence.

### Expected behavior

Generating a new Hyrax app using `template.rb` creates `FileSet < Hyrax::FileSet` in the new app's `app/models/file_set.rb`.

QUESTION:

It is unclear if the generated file set is needed.  Is there a reason not to just use `Hyrax::FileSet`?

### Actual behavior

Generating a new Hyrax app using `template.rb` creates `FileSet < ActiveFedora::Base` in the new app's `app/models/file_set.rb`.

1. Generate a new Hyrax app using...

```
rails _5.2.6_ new my_hyrax -m https://raw.githubusercontent.com/samvera/hyrax/v3.1.0/template.rb
```

_NOTE: May need to adjust the tag from v3.1.0 to the latest version tag for Hyrax._

### Related Work

Issue #5099 App generated through template includes collection model that is an AF::Base model
Issue #5100 App generated by template includes .fcrepo_wrapper for connecting to Fedora

rawOrlando:
### Fixes

Fixes #5098

### Summary

Changes FileSet generator to use to generate FileSet identical to in Koppie instead of Dassie.

`lib/generators/hyrax/templates/app/models/file_set.rb` should now match  `.koppie/app/models/file_set.rb`

@samvera/hyrax-code-reviewers


no-reply:
shouldn't this just not be called anymore? we don't want to encourage app devs to put behavior in a local FileSet, and they should just use `Hyrax::FileSet` directly instead, right?

jeremyf (approved):

PR #6399, opened 2023-10-30
Clarify purpose of FileSetFileService 

no-reply:
this service seems at first glance like it's just layering complexity on the
`#find_original_file` custom query (i almost deprecated it for this reason). but
its actually hiding an abstraction: the idea that the `pcdmuse:OriginalFile` is
the "primary" file for a FileSet.

this abstraction is useful, if Hyrax uses it consistently, because downstream
applications may want to determine the "primary" file any number of other
ways. (treating the OriginalFile as primary isn't at all in line with the PCDM
Use vocabulary's description of it, anyway).

clarify this all a bit with documentation and a method rename refactor.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/version_list_presenter.rb
+++ b/app/presenters/hyrax/version_list_presenter.rb
@@ -20,8 +20,13 @@ module Hyrax
     #   relevant file versions.
     #
     # @raise [ArgumentError] if we can't build an enumerable
-    def self.for(file_set:)
-      original_file = Hyrax::FileSetFileService.new(file_set: file_set).primary_file
+    def self.for(file_set:, file_service: Hyrax.config.file_set_file_service)
+      original_file = case file_set
+                      when ActiveFedora::Base
+                        file_set.original_file
+                      else
+                        file_service.new(file_set: file_set).primary_file
+                      end
 
       new(Hyrax::VersioningService.new(resource: original_file))
     rescue NoMethodError
--- a/app/presenters/hyrax/version_list_presenter.rb
+++ b/app/presenters/hyrax/version_list_presenter.rb
@@ -19,13 +19,10 @@ module Hyrax
     # @return [Hyrax::VersionListPresenter] an enumerable of presenters for the
     #   relevant file versions.
     #
-    # @raise [ArgumentError] if we can't build an enu
+    # @raise [ArgumentError] if we can't build an enumerable
     def self.for(file_set:)
-      original_file = if file_set.respond_to?(:original_file)
-                        file_set.original_file
-                      else
-                        Hyrax::FileSetFileService.new(file_set: file_set).original_file
-                      end
+      original_file = Hyrax::FileSetFileService.new(file_set: file_set).primary_file
+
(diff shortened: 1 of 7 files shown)


app/presenters/hyrax/version_list_presenter.rb
```diff
@@ -19,13 +19,10 @@ def initialize(service)
     # @return [Hyrax::VersionListPresenter] an enumerable of presenters for the
     #   relevant file versions.
     #
-    # @raise [ArgumentError] if we can't build an enu
-    def self.for(file_set:)
-      original_file = if file_set.respond_to?(:original_file)
```
marrus-sh, line 24: I believe this `if` was here for ActiveFedora compatibility—`FileSetFileService` probably only works in Valkyrie?

no-reply: ah, i removed it because *all* model classes for FileSet have `#original_file`, so it prevents the service from ever being reached. 

doing a class based switch statement instead.

lib/hyrax/configuration.rb
```diff
@@ -267,6 +267,14 @@ def derivative_services
       @derivative_services ||= [Hyrax::FileSetDerivativesService]
     end
 
+    ##
+    # @!attribute [rw] file_set_file_service
+    #   @return [Class] implementer of {Hyrax::FileSetFileService}
+    attr_writer :file_set_file_service
+    def file_set_file_service
+      @file_set_file_service ||= Hyrax::FileSetFileService
+    end
```
marrus-sh, line 276: We have been adding new configurations to `spec/lib/hyrax/configuration_spec.rb` (to test that they are settable and retrievable) and `lib/generators/hyrax/templates/config/initializers/hyrax.rb` (as documentation).

I don’t personally care too much either way but it might be good to follow the pattern?

spec/services/hyrax/file_set_file_service_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+    end
+
+    context "when the FileSet has an #original_file_id" do
+      it "always resolves by original_file_id"
+    end
+
+    context "when there is no OriginalFile, but files exist" do
+      it "resolves the first file"
+    end
```
marrus-sh, line 28: These specs still need to be filled out :)

marrus-sh (commented):
Looks good to me aside from the two comments!

spec/services/hyrax/file_set_file_service_spec.rb
no-reply: [these tests deliberately left blank]

this was missing unit tests and i wanted to outline coverage, but didn't intend to fully test the existing behavior. since this is just a method rename, it seems okay to leave these pending?

marrus-sh: yeah, i’m fine with that

lib/hyrax/configuration.rb
no-reply: got this in

marrus-sh (approved):

PR #6410, opened 2023-11-01
Fix embargo expiry job spec

dlpierce:
Add valkyrie specs for EmbargoExpiryJob

Marks existing spec as AF only.

Duplicates existing EmbargoExpiryJob specs but with Valkyrie resources.

Spec setup is simplified by not specifying "restricted" visibility, instead expecting the factory
default "authenticated". Additional checks to confirm the initial visibility state are also added.

The embargo traits from the hyrax_work factory are copied to the hyrax_file_set factory.

Saving of the expired embargo ACL is moved to inside the time machine to 10 days ago. Without this
change the solr doc will not contain the embargo related fields, causing
`#records_with_expired_embargoes` to not find those resources.

### DIFF
--- a/spec/factories/hyrax_file_set.rb
+++ b/spec/factories/hyrax_file_set.rb
@@ -48,6 +48,28 @@ FactoryBot.define do
       Hyrax.index_adapter.save(resource: file_set) if evaluator.with_index
     end
 
+    trait :under_embargo do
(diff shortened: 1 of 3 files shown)


bwatson78 (approved):
@dlpierce Thank you for finding the fix!

PR #6411, opened 2023-11-01
Attempts to fix the failing specs for shares_controller_spec.

bwatson78:
### Fixes

Fixes rspec failures for `spec/controllers/hyrax/my/shares_controller_spec.rb`.

### Detailed Description

Swaps out AF factory objects with Valkyrie Monographs.

@samvera/hyrax-code-reviewers


dlpierce:
After looking at this spec closer, I see let! was being used to create other documents that should be filtered out from this view. I moved them to a context so they only get created once when they are needed.

dlpierce (approved):

PR #6417, opened 2023-11-07
Skip work_usage_spec.rb tests for Valkyrie

abelemlih:
### Fixes

Per @dlpierce's recommendation, I am editing the `work_usage_spec.rb` file to skip tests until Valkyrization of analytics is complete. This will remove failing specs for instances like Koppie where Valkyrie is used.

* Skip tests in `spec/presenters/hyrax/work_usage_spec.rb` when Valkyrie is used. Once the Valkyrization of analytics is done, the test file should be edited to match the setup of a Valkyrie setup.
*
*

@samvera/hyrax-code-reviewers


dlpierce (changes requested):
Instead of an `if...else` can this be skipped by adding `:active_fedora` to the top describe? It passes already in dassie with valkyrie/wings, so it only needs to be skipped for now when wings is disabled.

dlpierce (approved):

PR #6421, opened 2023-11-07
Use class methods for Valkyrie model duck‐typing

marrus-sh:
It may be necessary to determine whether a model *class* represents a PCDM Object, File Set, etc, not just a model *instance*. This commit defines the former and aliases ~~and deprecates~~ the latter.

~~`collection?` and `pcdm_collection?` are split, in that `Hyrax::AdministrativeSet` and `Hyrax::PcdmCollection` are both the former, but not both the latter. Is this right, or should those methods be synonyms?~~ They’re effectively synonyms now.

`work?` is an alias for `pcdm_object? & !file_set?`.

`Hyrax::PcdmCollection` used to think it was a `pcdm_object?`, but it shouldn’t have.

~~There’s probably a fair amount of `model.try(:work?)` which needs to be changed to `model.class.try(:work?) || model.try(:work?)` before the old methods can be removed. We’ll need to support both for as long as we support ActiveFedora, but normalizing on a single place for querying this information, and having that place be the model class, seemed desirable for Valkyrie resources.~~ For the time being, we’ll just support both.

Notably, this enables the use of these methods for determining which form or indexer to use for a class of resource, instead of relying on class inheritance checks.

### DIFF
--- a/app/forms/hyrax/forms/pcdm_object_form.rb
+++ b/app/forms/hyrax/forms/pcdm_object_form.rb
@@ -26,6 +26,9 @@ module Hyrax
     ##
     # A form for PCDM objects: resources which have collection relationships and
     # generally resemble +Hyrax::Work+.
+    #
+    # Although File Sets are technically also PCDM objects, they use a separate
+    # form class: +Hyrax::Forms::FileSetForm+.
     class PcdmObjectForm < Hyrax::Forms::ResourceForm
       include Hyrax::FormFields(:core_metadata)
 
--- a/app/models/hyrax/administrative_set.rb
+++ b/app/models/hyrax/administrative_set.rb
@@ -33,7 +33,7 @@ module Hyrax
 
     ##
     # @return [Boolean] true
-    def collection?
+    def self.collection?
       true
     end
 
--- a/app/models/hyrax/administrative_set.rb
+++ b/app/models/hyrax/administrative_set.rb
@@ -33,7 +33,7 @@ module Hyrax
 
     ##
     # @return [Boolean] true
-    def self.collection?
+    def self.pcdm_collection?
       true
     end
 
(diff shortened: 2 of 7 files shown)


no-reply:
> This commit defines the former and aliases and deprecates the latter.

why deprecate and not just keep them?

no-reply:
> collection? and pcdm_collection? are split, in that Hyrax::AdministrativeSet and Hyrax::PcdmCollection are both the former, but not both the latter. Is this right, or should those methods be synonyms?

i think Admin Sets are ambiguous at the abstract model layer. initially, they were introduced without reference to any PCDM or Hydra::Works concepts. the Collections Extensions work effort later tried (incompletely) to wrap them into the Collection Types concept; so an Admin Set would be "just" a Collection with a certain Collection Type.

i can't see any downsides to treating AdministrativeSet instances as `pcdm:Collection`s, and would suggest we move formally in that direction going forward, but it may be appropriate for @dlpierce, @rjkati, and @samvera/hyrax-metadataists to weigh in?

marrus-sh:
> i can't see any downsides to treating AdministrativeSet instances as pcdm:Collections, and would suggest we move formally in that direction going forward

the (extremely slight) conceptual convenience to treating them separately is that they use different forms. but it’s easy enough to just check if something is an admin set first, prior to checking if it is a PCDM collection

marrus-sh:
> > This commit defines the former and aliases and deprecates the latter.
> 
> why deprecate and not just keep them?

Honestly I was going back and forth about this; it seemed kind of weird to have both (considering that it would be unexpected if they disagreed), but the convenience might outweigh those concerns. They can’t be removed in the near term regardless, so I guess it just depends whether removing them in the long term is desirable or not. (If so, it seems better to start issuing deprecation warnings as soon as possible.)

marrus-sh:
@no-reply just pushed two additional commits:

- remove the deprecation on the instance methods but add tests to the shared specs that they match the default implementation

- make AdministrativeSet return `true` for `pcdm_collection?`

no-reply (approved):

PR #6426, opened 2023-11-08
Fix trophies_controller_spec.rb tests for Valkyrie

abelemlih:
### Fixes

Fix test failures in `spec/controllers/hyrax/trophies_controller_spec.rb` when Valkyrie is used.

* Create a work in the test file that is compatible with Valkyrie
*
*

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/controllers/hyrax/trophies_controller_spec.rb
+++ b/spec/controllers/hyrax/trophies_controller_spec.rb
@@ -2,7 +2,7 @@
 RSpec.describe Hyrax::TrophiesController do
   describe "#toggle_trophy" do
     let(:user) { create(:user) }
-    let(:work) { FactoryBot.valkyrie_create(:hyrax_work, user: user) }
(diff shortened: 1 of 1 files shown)


spec/controllers/hyrax/trophies_controller_spec.rb
```diff
@@ -2,7 +2,13 @@
...
   describe "#toggle_trophy" do
     let(:user) { create(:user) }
-    let(:work) { create(:work, user: user) }
+    let(:work) do
+      if Hyrax.config.use_valkyrie?
+        FactoryBot.valkyrie_create(:monograph, depositor: user.user_key)
+      else
+        create(:work, user: user)
+      end
+    end
```
dlpierce, line 11: In many situations like this using the hyrax_work factory is good enough. Wings will create an AF GenericWork, or when wings is off a resource will be created.
```suggestion
  let(:work) { FactoryBot.valkyrie_create(:hyrax_work, user: user) }
```

dlpierce (approved):

PR #6435, opened 2023-11-09
WorkflowActionForm: run service as a background job

dunn:
### Fixes

Workflow updates can trigger slow operations; when run in the foreground this can cause browser timeouts.

@samvera/hyrax-code-reviewers


dunn:
the `WorkflowActionService` specs seem to be flaky, which makes very little sense since they were just moved over from the form spec.

PR #6436, opened 2023-11-09
Updates lease_service_spec.rb for valkyrie.

bwatson78:
### Fixes

Fixes `spec/services/hyrax/lease_service_spec.rb`

### Summary

Updates lease_service_spec.rb for valkyrie.

@samvera/hyrax-code-reviewers

Note: it may cause problems that the LeaseService relies on Solr instead of querying the actual objects.


dlpierce:
I think making use of Hyrax::TimeService to perform the indexing in the past can help resolve the problem of no expired lease data being indexed. However, there is also #6330 that has been waiting to get finished up and may impact how this spec is fixed up. Then embargo_service_spec.rb can likely share in the fixes given the similarities of leases and embargoes.

dlpierce:
As expected, merging 6330 caused some conflicts.

dlpierce (approved):

PR #6441, opened 2023-11-13
Refactor Valkyrie indexer API (to match forms)

marrus-sh:
**Note: Because there are some file renames which don’t persist across the full diff, I recommend reviewing each commit in sequence.**

This is a kind of heavy refactor of the Valkyrie indexer classes to match the API used for forms. This means :⁠—

- Using configurable classes, like `Hyrax.config.pcdm_collection_indexer`, instead of hardcoded ones.

- Renaming all of the classes to follow a pattern and be namespaced to `Hyrax::Indexers` :⁠—

  - `Hyrax::ValkyrieIndexer` 🔜 `Hyrax::Indexers::ResourceIndexer`.

    **Note:** This is a somewhat less‐basic base class, as it includes `Hyrax::ResourceIndexer`. If we still want `Hyrax::ValkyrieIndexer` as a base class which does not utilize `Hyrax::ResourceIndexer`, that’s probably an option.

  - `Hyrax::PcdmCollectionIndexer` 🔜 `Hyrax::Indexers::PcdmCollectionIndexer`.

    **Note:** This class has already been renamed once in the past (it was originally `Hyrax::ValkyrieCollectionIndexer`; see 7ca32a59062e3dbed7e4b04ebcf7bd2c261b3e28).

  - `Hyrax::ValkyrieFileSetIndexer` 🔜 `Hyrax::Indexers::FileSetIndexer`.

  - `Hyrax::ValkyrieWorkIndexer` 🔜 `Hyrax::Indexers::PcdmObjectIndexer`.

  The old aliases are supported with deprecation warnings.

- Implementing `Hyrax::Indexers::ResourceIndexer()` and `Hyrax::Indexers::PcdmObjectIndexer()` to match `Hyrax::Forms::ResourceForm()` and `Hyrax::Forms::PcdmObjectForm()`.


dlpierce (approved):
Thanks for the reorg! This is a welcome cleanup.

PR #6453, opened 2023-11-16
Updates tested collection objects to Valkyrie-compliant (spec/search_builders/hyrax/collection_search_builder_spec.rb).

bwatson78:
### Fixes

Fixes `spec/search_builders/hyrax/collection_search_builder_spec.rb`.

### Summary

Adds active_fedora tag to AF-specific test in collection_search_builder_spec.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/search_builders/hyrax/collection_search_builder_spec.rb
+++ b/spec/search_builders/hyrax/collection_search_builder_spec.rb
@@ -44,7 +44,7 @@ RSpec.describe Hyrax::CollectionSearchBuilder do
     end
   end
 
-  describe '#gated_discovery_filters' do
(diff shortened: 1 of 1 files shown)


no-reply:
Thanks for this, @bwatson78!

it's not clear to me looking at this one why this test is AF only. is this method only applicable to the ActiveFedora models for some reason?

bwatson78:
@no-reply You're right, I will circle back to this. Daniel showed me how deposit access was setup with Valkyrie objects recently. I'll apply that here.

bwatson78:
@no-reply Does that look right?

dlpierce (approved):

PR #6459, opened 2023-11-20
Makes spec/services/hyrax/file_set_fixity_check_service_spec.rb Dassie-only.

bwatson78:
### Fixes

Fixes `spec/services/hyrax/file_set_fixity_check_service_spec.rb`.

### Summary

Makes spec/services/hyrax/file_set_fixity_check_service_spec.rb Dassie-only.

Add an appropriate `notes-*` label to the PR (or indicate here) that classifies this change.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/file_set_fixity_check_service_spec.rb
+++ b/spec/services/hyrax/file_set_fixity_check_service_spec.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
-RSpec.describe Hyrax::FileSetFixityCheckService, :active_fedora do
+RSpec.describe Hyrax::FileSetFixityCheckService,
+               pending: Hyrax.config.disable_wings ? 'Valkyrie rewrite needed' : false do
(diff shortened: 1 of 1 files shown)


no-reply:
Doesn't this code need to work for koppie?

bwatson78:
@no-reply I attempted to convert the tested FileSet to a :hyrax_file_set, but was blocked by errors stating that `.files` wasn't a method of `Hyrax::FileSet`. So I went back to the Service to see if anything pointed to this being ActiveFedora only, and saw that [in the Service class](https://github.com/samvera/hyrax/blob/dac642743e8476800cb494fa8d8db539a7b27248/app/services/hyrax/file_set_fixity_check_service.rb#L29), the expected param is an ActiveFedora object.

I tried to do due diligence and track down whether there was a separate Valkyrie process to perform Fixity, but couldn't find one. Would you happen to know where the Fixity processing starts for Valkyrie FileSets?

PR #6467, opened 2023-11-21
Marks test suite as ActiveFedora-only, since importer has no Valkyrization.

bwatson78:
### Fixes

Fixes `spec/services/hyrax/user_stat_importer_spec.rb`.

### Summary

Marks test suite as ActiveFedora-only, since importer has no Valkyrization.

@samvera/hyrax-code-reviewers


dlpierce (approved):
Yes, this service is currently strongly tied to ActiveFedora and will need a rewrite for Valkyrie.

PR #6478, opened 2023-11-28
Enqueues Job that wasn't calling.

bwatson78:
### Fixes

Fixes `spec/controllers/hyrax/file_sets_controller_spec.rb`.

### Summary

Enqueues Job that wasn't calling.

@samvera/hyrax-code-reviewers


dlpierce (approved):
I'm curious what causes the number of `run` calls to decrease from this change.

PR #6499, opened 2023-12-04
Support all PCDM terms in `Hyrax::FileMetadata::Use`

marrus-sh:
Our Hyrax application makes use of `pcdmuse:PreservationFile` and `pcdmuse:IntermediateFile`, and is currently overriding Hyrax to provide these constants. But the list of possible PCDM use terms is [short and well known](https://pcdm.org/2021/04/09/use), and there’s no reason why Hyrax couldn’t just provide all of them to begin with.

I’ve also taken the liberty of changing references to `Hyrax::FileMetadata::Use::THUMBNAIL` to be `Hyrax::FileMetadata::Use::THUMBNAIL_IMAGE`, since that better matches the naming convention of the other terms. `Hyrax::FileMetadata::Use::THUMBNAIL` is still supported, but probably should be discouraged.

### DIFF
--- a/app/actors/hyrax/actors/file_actor.rb
+++ b/app/actors/hyrax/actors/file_actor.rb
@@ -80,7 +80,7 @@ module Hyrax
           :original_file
         when Hyrax::FileMetadata::Use::EXTRACTED_TEXT
           :extracted_file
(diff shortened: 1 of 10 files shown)


dlpierce (approved):
This is good! Thanks for filling in the gap.

PR #6518, opened 2023-12-07
chart: set default value for nginx.service

dunn:
deploys fail with helm 3.13.x, likely due to what they describe as 'numerous issues fixed' with regard to values handling (https://github.com/helm/helm/releases/tag/v3.13.0) and the fact that the ingress template evaluates .Values.nginx.service.port even though .Values.nginx.service itself isn't set by default https://github.com/samvera/hyrax/commit/d44e60a230314d57c2f0c462789c9819d4f6bb8b#diff-93ef0c97724a14a853549acbbd98c7524a7d85a670bbac74f5635ab8aff54033R3

@samvera/hyrax-code-reviewers


aprilrieger (approved):

orangewolf:
@dunn related to this - April and I just figured out that the nginx chart is making load balancers by default. we were legging down some extra billing in our clusters. we should likely set it to ClusterIP by default (see suggested change)

chart/hyrax/values.yaml
```diff
@@ -284,6 +284,10 @@ nginx:
   image:
     repository: bitnami/nginx
     tag: 1.23.0-debian-11-r1
+  # this needs to be set to an empty hash, since templates/ingress.yaml tries to
+  # evaluate .Values.nginx.service.port, and it needs to at least evaluate to
+  # nil, rather than throw an error (which it does if `service` itself is nil).
+  service: {}
```
orangewolf, line 290:
```suggestion
  # this needs to be set since templates/ingress.yaml tries to
  # evaluate .Values.nginx.service.port, and it needs to at least evaluate to
  # nil, rather than throw an error (which it does if `service` itself is nil).
  # you can change to NodePort or LoadBalancer if direct access is needed.
  # ClusterIP is used as we assume an ingress is in place.
  service:
    type: ClusterIP
    port: 80
```

aprilrieger (approved):
Thank you!

Issue #6517, opened 2023-12-07
Save button disabled on work page when deposit agreement is passive


bbpennel:
### Descriptive summary

If running a hyrax instance with `Active deposit agreement acceptance` disabled, if you add metadata and then add a file, the "Save" button will be disabled even though all the requirements are met. This happens with new or existing works

### Steps to reproduce the behavior in User Interface (UI)

1. In the dashboard as an admin, go to Settings > Features, then turn off "Active deposit agreement acceptance".
2. Then go to the form to create or edit a work. 
3. Input all the required metadata (if creating a new work)
4. Add a file to the work
5. The "Save" button will now be disabled. The only way to enable it is to modify your metadata.

### Actual behavior (include screenshots if available)

I've replicated this issue in hyrax 4 (dassie) and 5 (nurax), and a variation of it in our local hyrax 4 repository (our handling of deposit agreements has been modified, but it produces the same outcome). Tested in firefox and chrome, on mac and windows.

This does not produce any errors.

<img width="1405" alt="Screen Shot 2023-12-07 at 1 49 27 PM" src="https://github.com/samvera/hyrax/assets/969810/108906e3-7856-403b-8e53-f65b61120202">

### Acceptance Criteria/Expected Behavior

The "Save" button should be re-enabled after the file upload completes, if all other requirements are met.

conorom:
Noticed this bug as we are struggling to upgrade heliotrope to Hyrax 4.0.0
In the process of writing an internal Jira ticket for this.

@dlpierce @bbpennel the issue should be closed now with the merge of the linked PR in December, right?

ryanfb:
It would also be nice to have this backported for a 4.0.x point release, since this seems to be a regression introduced introduced in 4.x.

bbpennel:
It should be relatively easy to backport this to 4.x, we are [using it locally](https://github.com/UNC-Libraries/hyrax/pull/2/files) in our [4.x based fork of hyrax ](https://github.com/UNC-Libraries/hyrax/tree/unc-hyrax-4-development)(which has a number of other backported fixes). I could look into doing a PR for against samvera/hyrax in the next few days.

bbpennel:
It will probably need to wait for the build to work for 4.x again first though unfortunately, since my understanding is it needs to be updated to use gh actions.

PR #6530, opened 2023-12-11
Fix save button remaining inactive after upload completes (with passive deposit agreements)

bbpennel:
### Fixes

https://github.com/samvera/hyrax/issues/6517

### Summary

Fixes issue where the "Save" button on the work page is disabled after an upload completes even if all requirements have been met.

See the linked issue for how to replicate the issue.

notes-bugfix

### Detailed Description

The issue appears to be that the fileupload('active') value is still "1" at the time that the `fileuploadcompleted` or `fileuploaddone` events are fired. It is being retrieved here:
https://github.com/samvera/hyrax/blob/main/app/assets/javascripts/hyrax/save_work/uploaded_files.es6#L16
which is used for deciding if the "Save" button should be disabled. So when the completed event fires, it checks to see if there are any active uploads, and there still are since the plugin doesn't seem to have cleared them yet.

Since 'active' doesn't seem reliable for this purpose, I switched to externally tracking when an upload was started and ended. There is also a 'progress' attribute which records the number of bytes uploaded that might be usable for this, which we could try if that's preferred.

For what its worth, it looks like the jquery.fileupload plugin was archived in May 2023. Some parts also don't seem to be working, like the "Cancel upload" button is disabled for me and doesn't work, plus it doesn't show in progress uploads, the file just eventually appears on the list once the upload finishes. I'm guessing we need to replace it at some point, which might already be in the backlog.

* Uses an external counter for number of uploads in progress rather than the internal one, since it does not update in time for reevaluating the save button.
* Adds a test to demonstrate the bug and check for passive deposit agreement behavior

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax/save_work/uploaded_files.es6
+++ b/app/assets/javascripts/hyrax/save_work/uploaded_files.es6
@@ -3,7 +3,15 @@ export class UploadedFiles {
   constructor(form, callback) {
     this.form = form
     this.element = $('#fileupload')
-    this.element.on('fileuploadcompleted', callback)
+    this.uploadsInProgress = 0
+    this.element.on('fileuploadadded', (e, data) => {
+      this.uploadsInProgress += 1
+      callback(e, data)
+    })
+    this.element.on('fileuploadcompleted', (e, data) => {
+      this.uploadsInProgress -= 1
+      callback(e, data)
+    })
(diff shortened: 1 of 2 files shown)


dlpierce (approved):
Thanks for tracking this down! Yes, it will need to be replaced. I think some Samvera apps have been investigating alternatives already.

PR #6538, opened 2023-12-12
Remove empty “find_by_collection_type” custom query

marrus-sh:
This file wasn’t doing anything, right? It can be deleted??


orangewolf (approved):
the commit that adds this also adds `app/services/hyrax/custom_queries/find_collections_by_type.rb` which has actual content. This looks like a file that got created but then renamed and accidentally committed to me.

PR #6559, opened 2023-12-19
Cast provided `id` to string

jeremyf:
Prior to this commit, we were seeing the following issues:

```
NoMethodError:
  undefined method `split' for #<Valkyrie::ID:0x0000ffffa5f04bc8 @id="admin_set/default">
```

The string interpolation (e.g. `#{ }`) did not use the `#to_s` method of the identifier.  With this change, we treat the `Valkyrie::ID` as something we can "Noidify".


orangewolf (approved):
👍

PR #6597, opened 2024-01-09
Marks `spec/models/concerns/hyrax/solr_document/ordered_members_spec.rb` as ActiveFedora-only.

bwatson78:
### Fixes

Fixes `spec/models/concerns/hyrax/solr_document/ordered_members_spec.rb`.

### Summary

Marks `spec/models/concerns/hyrax/solr_document/ordered_members_spec.rb` as ActiveFedora-only.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/models/concerns/hyrax/solr_document/ordered_members_spec.rb
+++ b/spec/models/concerns/hyrax/solr_document/ordered_members_spec.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
-RSpec.describe Hyrax::SolrDocument::OrderedMembers do
+
+# NOTE: This class specifically queries `ordered_targets_ssim`, which is only
+#   indexed within Dassie.
+RSpec.describe Hyrax::SolrDocument::OrderedMembers, :active_fedora do
   subject(:decorated) { described_class.decorate(document) }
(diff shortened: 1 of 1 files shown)


dlpierce:
Won't we need ordered_targets_ssim to be available for MemberPresenterFactory to work? I think we still need that in valkyrie.

bwatson78:
Looking now...

bwatson78:
What I'm seeing if that we have a `PcdmMemberPresenterFactory` for Valkyrie objects. That's being switched to here: https://github.com/samvera/hyrax/blob/f2fed7aed2b6ba2168668b95a9e65ea794fb4079/app/presenters/hyrax/work_show_presenter.rb#L304. I've also checked to see where `ordered_member_ids` is being used, and everywhere I'm seeing it is in the Actor Stack(`Hyrax::Actors::ApplyOrderActor`) or a ternary is in place to not use it in Valkyrie (`Hyrax::IiifManifestPresenter` L#97 and `app/views/hyrax/base/_file_manager_members.html.erb` L#8 .

bwatson78:
@dlpierce ^ What do you think?

dlpierce (approved):
Okay, thanks for the investigation!

PR #6617, opened 2024-01-22
Marks `spec/services/hyrax/import_url_failure_service_spec.rb` as ActiveFedora-only.

bwatson78:
### Fixes

Fixes `spec/services/hyrax/import_url_failure_service_spec.rb`.

### Summary

Marks `spec/services/hyrax/import_url_failure_service_spec.rb` as ActiveFedora-only.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/services/hyrax/import_url_failure_service_spec.rb
+++ b/spec/services/hyrax/import_url_failure_service_spec.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
-RSpec.describe Hyrax::ImportUrlFailureService do
+# NOTE: This class deals with accessing a FileSet's #errors attribute,
+#   which isn't a defined attribute in Valkyrie objects.
+RSpec.describe Hyrax::ImportUrlFailureService, :active_fedora do
   let!(:depositor) { create(:user) }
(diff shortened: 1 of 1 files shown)


jeremyf:
@bwatson78 do we have an alternate for Valkyrie files?

bwatson78:
@jeremyf No, we don't (at least, as far as I can tell). I searched in the new Transactions/Steps infrastructure and didn't find an alternative to where this service is ultimately called at: https://github.com/samvera/hyrax/blob/023c9e074e9b22e454534f3ab6fc6fd09cc9fed6/app/jobs/import_url_job.rb#L97. 

In that same file, the [`use_valkyrie`](https://github.com/samvera/hyrax/blob/023c9e074e9b22e454534f3ab6fc6fd09cc9fed6/app/jobs/import_url_job.rb#L40) processing is left as no-op.

dunn (approved):

dlpierce:
According to https://github.com/samvera/hyrax/issues/4195#issuecomment-1690713708 

> Hyrax has replaced this functionality with ValkyrieIngestJob called via the WorkUploadsHandler.

PR #6647, opened 2024-01-30
♻️ Leverage more of Blacklight's Solr URL

jeremyf:
Prior to this commit, our defaults were not considering scheme, core, password, nor user from the Blacklight URL.

What this meant was that the generated URL could strip away useful information from the configured URL; in particular the user and password information.

With this change, we utilize more of the underlying Blacklight configuration; a net positive.


orangewolf (approved):
this is a good change. I'm not sure why it failed when Jeremy merged in main even through it passed on the merge before.

dlpierce (approved):
Looks good! It should pass again once valkyrie's shared specs are fixed.

no-reply:
cc: @mcritchlow @dunn

PR #6648, opened 2024-01-30
Update Dockerfile to Ruby 3.2.3, Alpine to 3.19

dlpierce:
### Summary
Update Dockerfile to Ruby 3.2.3, Alpine to 3.19
Also removes jemalloc builder and combines ENV lines.


orangewolf (approved):
seems like a reasonable step forward =-D

Issue #6521, opened 2023-12-08
Sirenia testing - VR_1.2 test: error when approving or requesting changes on a 1 step workflow deposit 


eporter23:
### Descriptive summary

When an Admin approves a deposit submitted through a basic 1-step workflow, it returns an Valkyrie::Persistence::ObjectNotFoundError in Hyrax::WorkflowActionsController#update error. Same error occurs if trying to request revisions on the work or if trying to Edit the work as an admin. The Dashboard notifications do show that the approval and review changes have occurred.

### Steps to reproduce the behavior in User Interface (UI)

1. Create an Admin Set allowing 1-step workflow, and all visibility and release options ([see test VR_1.2](https://docs.google.com/spreadsheets/d/1cBmqnbwWvpKLPOehpKBzuPZo1W1dYo7Y/edit#gid=285442197))
2. As a different user, create a new work and make sure it's assigned to this Admin Set
3. As an approving/admin user, go to review the new work
4. Fill out the approval options on the work and submit
5. See the below error message

### Actual behavior (include screenshots if available)

Include what version of Hyrax relates to this issue (3.x, 4.x, main branch, etc.) if appropriate, and any relevant error messages/tracebacks if you're reporting a bug.

Hyrax 5rc2 main, Sirenia (Fedora 6) docker image

<img width="1184" alt="Screen Shot 2023-12-07 at 2 14 20 PM" src="https://github.com/samvera/hyrax/assets/13855987/31524a22-ac9f-4d11-b68a-10366fb891b6">

Note: the approval and/or request revisions activity does show up in the users' notifications:

<img width="1085" alt="Screen Shot 2023-12-07 at 2 17 01 PM" src="https://github.com/samvera/hyrax/assets/13855987/35fff889-29f3-4395-8905-401342e9503d">

### Acceptance Criteria/Expected Behavior

- [ ] When a user with appropriate permissions is shown the option to Approve a 1-step deposit submission, they should not see an error message, and should be returned to the work with the appropriate status indicated
- [ ] The work should be transitioned to the Published/Deposited state

### Rationale (for feature request only)

Provide the rationale or user story that describes "why" this issue should be addressed. Especially if this is a new feature or significant change to the existing implementation.

1-step workflows should be supported in Fedora 6-backed repositories

### Related work

Link to related issues or prior related work here.

eporter23:
[...]

PR #6652, opened 2024-01-31
Change embargo/lease ID types to fix nil values being converted to blank strings

### LINKED ISSUE #6650: Blank values for Valkyrie::ID attributes cause URI-to-ID lookups to return Fedora base path as ID
In Hyrax using Valkyrie (i.e. koppie and sirenia) there are demonstrable scenarios where a work's attributes get changed from `nil` to type `Valkyrie::ID` containing a blank string when saved.  Specific examples are `embargo_id` and `lease_id` when a work is in a one-step workflow.  This doesn't adversely affect koppie on postgres, because the blank string seems to be functionally equivalent to `nil`, AFAIK.

But this breaks ID handling in sirenia on fedora because the URI values on those attributes get stored in fedora as just `http://fedora_url/base_path` with no trailing ID.  The result is that [this call to translate the URI to an ID](https://github.com/samvera/valkyrie/blob/main/lib/valkyrie/persistence/fedora/metadata_adapter.rb#L53) returns just the configured `base_path` part of the URI as if it were the ID.  On the flip side, [this call to translate an ID into a URI](https://github.com/samvera/valkyrie/blob/main/lib/valkyrie/persistence/fedora/metadata_adapter.rb#L60) returns `http://fedora_url/base_path/base_path`, which will produce `Ldp::NotFound` errors.

In the context of Hyrax (i.e. sirenia), the actual breaking behavior can be baffling; those IDs now contain the configured `base_path` when they should have been `nil` or blank, triggering calls in parts of the stack that it should normally slip silently through. For example, this [Hyrax issue](https://github.com/samvera/hyrax/issues/6521) describes errors while saving workflow state for a work, which is caused by `Ldp::NotFound` for the URI returned for `embargo_id`, even though it is not under embargo. 

I'm opening an issue here and in [Valkyrie](https://github.com/samvera/valkyrie/issues/944) because it's not clear if something in Hyrax or the Valkyrie persister itself is converting those ID attributes from `nil` to `Valkyrie::ID @id=""` .  Also, we may have to provide a workaround on the Hyrax side if we can't find why those get converted but the behavior doesn't change on the Valkyrie side.

This problem can be demo'd in a Hyrax (sirenia) console:

```
persister = Hyrax.persister
monograph1 = Monograph.new

# embargo_id is initially nil
monograph1.embargo_id 
# => nil

# Set embargo_id to blank string to mimic what happens for works in a one-step workflow
monograph1.embargo_id = ""  

# Value is now blank with type of Valkyrie::ID
monograph1.embargo_id 
# => #<Valkyrie::ID:0x00007f5f73fecb88 @id="">

# Saving, which currently allows blank values for the ID attributes...
monograph1 = persister.save(resource: monograph1)

# Load the saved version to a new object, which walks attributes through URI-to-ID conversion
monograph2 = Hyrax.query_service.find_by(id: monograph1.id)

# embargo_id now equal to base_path
monograph2.embargo_id 
# => #<Valkyrie::ID:0x00007f5f6eb829a8 @id="development">

# Subsequent ID-to-URI conversion returns invalid location:
Hyrax.query_service.find_by(id: monograph2.embargo_id)

Valkyrie::Persistence::ObjectNotFoundError: Valkyrie::Persistence::ObjectNotFoundError
...
Caused by Ldp::NotFound: Error: Resource /development/development not found
```

randalldfloyd:
### Fixes

Fixes #6650 ; refs #6521

### Summary

When using a Valkyrie adapter, the attribute type for `embargo_id` and `lease_id` is currently `Valkyrie::Types::ID`.  Because of the behavior of that type, if those attributes get set to a blank string, they will get persisted as `Valkyrie::ID @id=""` .    This is problematic when round-tripping them through URI/ID conversions in the fedora adapter (see Description section.)  This PR changes to `Valkyrie::Types::Params::ID`, because blank strings are always returned as `nil`, whereas `Valkyrie::Types::ID` returns exactly what it is given, and a blank string is still a valid string.

In a rails console (i.e. in sirenia), create a work and set a blank string on `embargo_id`:
```
persister = Hyrax.persister
resource = Monograph.new
resource.embargo_id = ""
monograph1 = persister.save(resource: resource)
```
Make sure resulting `embargo_id` is still `nil`, per the behavior of `Valkyrie::Types::Params::ID`
```
monograph1.embargo_id 
=> nil
```
### Detailed Description
Per the summary, this is a problem when/if IDs like `embargo_id` and `lease_id` are ever changed from `nil` to a blank string.  But why would we want to do that?  We don't, but it is consistently demonstrable when a work is related to an admin set with a one-step workflow.  It's not clear where/why this happens.

This isn't a problem for the postgres adapter, but for fedora, those IDs are stored as URIs.  When those get ran through URI-to-ID conversion by the fedora metatdata adapter, it returns what's anchored at the end, which is just the fedora base path (i.e. `development` etc.) because there is no trailing ID.   Conversely, when that value is ran through ID-to-URI conversion, it gets appended to the configured base path, so it ends up with `http://fedora_url/base_path/base_path` . 

This introduces breakage because the `embargo_id` and `lease_id`  is no longer `nil` or blank, so it triggers methods that it should have short-circuited, with invalid URIs that are guaranteed to produce  `Ldp::NotFound` errors.  This produces confusing behavior; in the case of #6521, the error came from loading the object for `embargo_id` when the work wasn't even under embargo. 

The effect of changing the attribute type is that, even though something in Hyrax workflows is inappropriately trying to change `nil` values to a blank string, the `Valkyrie::Types::Params::ID` will convert them back to `nil` when persisted; that's what the type is designed for.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/resource.rb
+++ b/app/models/hyrax/resource.rb
@@ -35,8 +35,8 @@ module Hyrax
     include Hyrax::WithEvents
 
     attribute :alternate_ids, Valkyrie::Types::Array.of(Valkyrie::Types::ID)
-    attribute :embargo_id,    Valkyrie::Types::ID
-    attribute :lease_id,      Valkyrie::Types::ID
+    attribute :embargo_id,    Valkyrie::Types::Params::ID
+    attribute :lease_id,      Valkyrie::Types::Params::ID
 
     delegate :edit_groups, :edit_groups=,
              :edit_users,  :edit_users=,


dlpierce (approved):
Thanks for tracking this down!

PR #6674, opened 2024-02-06
Repurpose `MemberCleanupListener` for parent/child

jeremyf:
This commit will repurpose the `MemberCleanupListener` to be used for cleaning up parent/child relationships.  When the child is deleted, the parent will be updated to remove the child id from its #member_ids.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/listeners/member_cleanup_listener.rb
+++ b/app/services/hyrax/listeners/member_cleanup_listener.rb
@@ -13,12 +13,14 @@ module Hyrax
         return unless event[:object]
 
         object = event[:object]
+        user = event[:user]
         return unless object.is_a?(Hyrax::Work)
 
         Hyrax.query_service.find_parents(resource: object).each do |parent|
           parent.member_ids -= [object.id]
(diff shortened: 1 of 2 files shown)


app/services/hyrax/listeners/member_cleanup_listener.rb
```diff
@@ -8,7 +8,19 @@ class MemberCleanupListener
...
+        event = event.to_h
+        return unless event[:object]
+
+        object = event[:object]
+        return unless object.is_a?(Hyrax::Work)
+
+        Hyrax.query_service.find_parents(resource: object).each do |parent|
+          parent.member_ids -= [object.id]
+          Hyrax.persister.save(resource: parent)
+          Hyrax.index_adapter.save(resource: parent)
```
no-reply, line 21: should this publish an `object.membership.changed` event instead?

no-reply (approved):

no-reply (changes requested):
oops. meant to push request changes

no-reply (approved):

Issue #6676, opened 2024-02-07
Discrepancy between `Hyrax::SolrService#query` and `ActiveFedora::SolrService#query`


kirkkwang:
### Descriptive summary

There is a discrepancy between `Hyrax::SolrService#query` and `ActiveFedora::SolrService#query` in what it returns.

### Steps to reproduce the behavior in User Interface (UI)

1. Using Bulkrax, initiate an export for all records
2. See that it fails

### Actual behavior (include screenshots if available)

Hyrax `double_combo` branch using Bulkrax `hyrax-4-valkyrie-support` branch
```rb
[32, 41] in /app/samvera/hyrax-webapp/gems/gems/bulkrax/app/factories/bulkrax/valkyrie_object_factory.rb
   32:     def self.query(q, **kwargs)
   33:       # TODO: Without the Hyrax::SolrService, what are we left with?  Someone could choose
   34:       # ActiveFedora::SolrService.
   35:       raise NotImplementedError, "#{self}.#{__method__}" unless defined?(Hyrax)
   36:       debugger
=> 37:       Hyrax::SolrService.query(q, **kwargs)
   38:     end
   39:
   40:     ##
   41:     # Retrieve properties from M3 model
(byebug) (q)
"has_model_ssim:(GenericWork OR Image OR GenericWorkResource OR ImageResource) "
(byebug) kwargs
{:fl=>"id,file_set_ids_ssim", :method=>:post, :rows=>2147483647}
(byebug) Hyrax::SolrService.query(q, **kwargs)
[]
(byebug) Hyrax::SolrService.query("*:*", **kwargs.merge(fq:q))
[{"id"=>"dc163b97-71dc-4efd-8fb6-c2b9021aa814"}]
(byebug) ActiveFedora::SolrService.query(q, **kwargs)
[{"id"=>"dc163b97-71dc-4efd-8fb6-c2b9021aa814"}]
```

Note that 
```rb
Hyrax::SolrService.query(q, **kwargs) => []
```

while
```rb
ActiveFedora::SolrService.query(q, **kwargs) => [{"id"=>"dc163b97-71dc-4efd-8fb6-c2b9021aa814"}]
```

In order for a similar return value, using this seems to work
```rb
Hyrax::SolrService.query("*:*", **kwargs.merge(fq:q)) => [{"id"=>"dc163b97-71dc-4efd-8fb6-c2b9021aa814"}]
```

### Acceptance Criteria/Expected Behavior

[...]

PR #6677, opened 2024-02-07
🐛 Restore `qt: 'standard'` to SolrService

jeremyf:
**tl;dr** Can anyone explain why we don't want `qt: 'standard'` for Valkyrie connections?

In `ActiveFedora::SolrService` we always assign the `qt: 'standard'`.[1]

However, with [this old commit][1] we dropped passing the `qt: 'standard'` when we had `use_valkyrie` of true.

We discovered this bug when upgrading Hyku and Bulkrax to Hyrax v3.x and some.

Closes https://github.com/samvera/hyrax/issues/6676

Related to:

- https://github.com/samvera/hyrax/pull/3857

[1]:https://github.com/samvera/active_fedora/blob/8e7d365a087974b4ff9b9217f792c0c049789de6/lib/active_fedora/solr_service.rb#L40-L48
[2]:https://github.com/samvera/hyrax/commit/be6104fd13117f6a6c5b0252acd5b341247f04d3

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/solr_service.rb
+++ b/app/services/hyrax/solr_service.rb
@@ -46,9 +46,8 @@ module Hyrax
     def get(query = nil, **args)
       # Make Hyrax.config.solr_select_path the default SOLR path
       solr_path = args.delete(:path) || Hyrax.config.solr_select_path
-      args = args.merge(q: query) if query.present?
+      args = args.merge(q: query, qt: 'standard') if query.present?
 
-      args = args.merge(qt: 'standard') unless query.blank? || use_valkyrie
       connection.get(solr_path, params: args)
     end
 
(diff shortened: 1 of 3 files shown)


dlpierce:
Perhaps it was to allow the application more flexibility when making solr queries?
I realize it was many moons ago, but perhaps @hackartisan has some idea?

jeremyf:
@dlpierce The main problem is that it means `ActiveFedora::SolrService` and `Hyrax::SolrService` are not interchangeable; which I assume will mean that even though the `Hyrax::SolrService` might fall back to the `ActiveFedora::SolrService` it won't be the same types of queries.  Thus introducing breaks for those that are migrating their Hyrax application.

I have an amendment in mind that should be adequate for allowing both.  It'll take a bit.

hackartisan:
Yeah, sorry I have specific memory about that param!

PR #6681, opened 2024-02-07
Adjust job enqueue timing

dlpierce:
When using the inline job queue, these jobs are run too early and encounter resources that have not yet been fully persisted. It's important to account for the inline nature of the dry-event pub/sub system and not assume that a job will actually be run later.

### Fixes

File ingest with :inline queue adapter

* With a valkyrie app using the inline job queue, can create a work with an attached file.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/work_uploads_handler.rb
+++ b/app/services/hyrax/work_uploads_handler.rb
@@ -93,7 +93,10 @@ module Hyrax
         event_payloads = files.each_with_object([]) { |file, arry| arry << make_file_set_and_ingest(file) }
         @persister.save(resource: work)
         Hyrax.publisher.publish('object.metadata.updated', object: work, user: files.first.user)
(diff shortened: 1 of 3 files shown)


app/services/hyrax/work_uploads_handler.rb
```diff
@@ -114,9 +117,7 @@ def make_file_set_and_ingest(file)
       file_set.permission_manager.acl.save if file_set.permission_manager.acl.pending_changes?
       append_to_work(file_set)
 
-      ValkyrieIngestJob.perform_later(file)
-
-      { file_set: file_set, user: file.user }
+      { file_set: file_set, user: file.user, job: ValkyrieIngestJob.new(file) }
```
jeremyf, line 120: Excellent!

jeremyf (approved):

PR #6687, opened 2024-02-08
Make generated app use Valkyrie

dlpierce:
### Summary

Updates the rails template to create a valkyrie enabled hyrax.

* Following the updated documentation results in a plain hyrax application configured for postgres valkyrie.
* A valkyrie hyrax application that does not define `::FileSet` is functional.

### Detailed Description
Following the deprecation of Hyrax's use of ActiveFedora, Hyrax should generate a minimal usable application configured for Valkyrie.

A [Lando](https://lando.dev) config providing services for Solr, Postgres, Redis, FITSservlet, and (if needed) Fedora 6 is created, along with a `.env` file with configuration pointing to these services. It is intended that, during development, the hyrax rails server not be run in a container, but on the local workstation. This `.lando.yml` is modeled on the [one that valkyrie uses](https://github.com/samvera/valkyrie/blob/main/.lando.yml).

* Modify files in `lib/generators` to preconfigure valkyrie
* Provide `.lando.yml` and `.env` files to new apps
* Avoid directly referencing `::FileSet` where its absence will prevent application from loading.
* Add an ENV var to configure which characterization tool to use.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/config/initializers/hyrax.rb
+++ b/.dassie/config/initializers/hyrax.rb
@@ -63,6 +63,11 @@ Hyrax.config do |config|
   # config.collection_model = 'Hyrax::PcdmCollection' # collection without basic metadata
   # config.collection_model = 'CollectionResource'    # collection with basic metadata
   # config.admin_set_model = 'Hyrax::AdministrativeSet'
+
+  # dassie needs legacy AF models
+  config.collection_model = '::Collection'
+  config.admin_set_model = 'AdminSet'
+  config.file_set_model = '::FileSet'
 end
 
 Date::DATE_FORMATS[:standard] = "%m/%d/%Y"
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+# Provide plain collection model if not defined by the application.
+# Needed until Hyrax internals do not assume its existence.
+class ::Collection < Hyrax.config.collection_class; end unless '::Collection'.safe_constantize
--- a/app/models/file_set.rb
+++ b/app/models/file_set.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+# Provide plain file set model if not defined by the application.
+# Needed until Hyrax internals do not assume its existence.
+class ::FileSet < Hyrax.config.file_set_class; end unless '::FileSet'.safe_constantize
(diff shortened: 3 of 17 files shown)


app/presenters/hyrax/pcdm_member_presenter_factory.rb
```diff
@@ -97,7 +97,7 @@ def work_presenters
     # @return
     def presenter_for(document:, ability:)
       case document['has_model_ssim'].first
-      when Hyrax::FileSet.name, ::FileSet.name # ActiveFedora FileSet within a Valkyrie Resource
+      when Hyrax::FileSet.name, '::FileSet'.safe_constantize&.name # ActiveFedora FileSet within a Valkyrie Resource
```
jeremyf, line 100:
```
require 'pathname'
value = nil
case value
when Pathname, nil
  puts "WHEN"
else
  puts "Else"
end
```

The above will output `WHEN` meaning that the safe constantize will begin to return a `Hyrax::FileSetPresenter.new` if there is no `document['has_model_ssim'].first`.

I'm not certain if that is an issue but it's something I just learn.

app/forms/hyrax/forms/file_set_edit_form.rb
```diff
@@ -8,7 +8,7 @@ class FileSetEditForm
 
     self.required_fields = [:title, :creator]
 
-    self.model_class = ::FileSet
+    self.model_class = '::FileSet'.safe_constantize
```
jeremyf, line 11: What will happen when this is `nil`?

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -8,7 +8,7 @@ class FileSetsController < ApplicationController
     include Hyrax::Breadcrumbs
 
     before_action :authenticate_user!, except: [:show, :citation, :stats]
-    load_and_authorize_resource class: ::FileSet, except: :show
+    load_and_authorize_resource class: Hyrax.config.disable_wings ? ::Hyrax::FileSet : ::FileSet, except: :show
```
jeremyf, line 11: Wondering if we want a `Hyrax.config.file_set_class` much like what we have for admin_set?  We could then encapsulate that logic there.

jeremyf (commented):
I'm not quite saying "no" but there's a few questions to address.

app/controllers/hyrax/file_sets_controller.rb
dlpierce: This might be the way to go, although it could make it less obvious if a piece of code is designed for use only with legacy ::FileSet objects.

dlpierce:
A missing `::FileSet` is causing some runtime issues, marking as draft for now.

config/initializers/legacy_models.rb
dlpierce, line 1: Instead of the earlier strategy of avoiding direct references to the old top level models, this defines those models if the app does not do so.

no-reply:
!! YES !!

jeremyf (approved):

PR #6712, opened 2024-02-20
🐛 Account for String and Valkyrie::ID Permissions

jeremyf:
**tl;dr** Given that we sometimes might compare `object.id` for permissions, we should treat Valkyrie::ID the same as the string.

Woe is us, for checking permissions on a String, what once was an identifier, but which hath been usurped by the more rigid Valkyrie::ID.

But loe, for the old ways still work.  Granting permission to objects based on a string identifier.  But Valkyrie::ID, one no longer a string, fails.

@samvera/hyrax-code-reviewers


dlpierce:
Yes, the `String`/`Valkyrie::ID` duality has been a source of endless fun. Thanks!

PR #6718, opened 2024-02-24
The helm chart says it evaluates extraEnvVars but it was not actually doing so

orangewolf:
### Summary
Help text was taking from Bitnami helm charts but the actual functionality was not.  This PR remedies that. It allows for things like

```yaml
extraEnvVars:
  - name: SOMETHING
     value: { .Values.some_value }
```
 
or

```yaml
extraEnvVars:
  - name: SOMETHING
     value: text

worker:
  extraEnvVars: |
      {{- toYaml .Values.extraEnvVars }}
      - name: WORKER_SOMETHING
         value: text
```

* Add the template parser for envVars as promised in the docs
* Does not change the existing functionality or require people to change their values files

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/templates/_tmplvalues.tpl
+++ b/chart/hyrax/templates/_tmplvalues.tpl
@@ -1,0 +1,38 @@
+{{/*
+Copyright VMware, Inc.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Renders a value that contains template perhaps with scope if the scope is present.
+Usage:
+{{ include "common.tplvalues.render" ( dict "value" .Values.path.to.the.Value "context" $ ) }}
+{{ include "common.tplvalues.render" ( dict "value" .Values.path.to.the.Value "context" $ "scope" $app ) }}
+*/}}
+{{- define "common.tplvalues.render" -}}
+{{- $value := typeIs "string" .value | ternary .value (.value | toYaml) }}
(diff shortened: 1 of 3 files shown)


chart/hyrax/templates/_tmplvalues.tpl
```diff
@@ -0,0 +1,38 @@
+{{/*
+Copyright VMware, Inc.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Renders a value that contains template perhaps with scope if the scope is present.
+Usage:
+{{ include "common.tplvalues.render" ( dict "value" .Values.path.to.the.Value "context" $ ) }}
```
dunn, line 10: we don't want to add a dependency on the bitnami `common` chart, do we?

dunn: oh i guess it's just re-using the name; in any case let's use `hyrax` instead

orangewolf: I don't see any references to Bitnami other than the copyright notice. The file is a copy of the upstream, but since it is Apache 2, as is Hyrax, the two are compatible license wise. I feel like leaving the attribution is a good thing here, but if you are concerned about it, I'll do some mods and then recommit it.

orangewolf: wow - I can't read apparently. I see what you mean now. I'll fix it

dunn (approved):

Issue #6725, opened 2024-02-28
/dashboard/collections error when not authenticated


bbpennel:
### Descriptive summary

Visiting the /dashboard/collections page produces an error when not logged in, "You cannot call create unless the parent is saved"

### Steps to reproduce the behavior in User Interface (UI)

Visit here while not logged in:
https://dev.nurax.samvera.org/dashboard/collections/

### Actual behavior (include screenshots if available)

The error is caused by this line:
https://github.com/samvera/hyrax/blob/main/app/controllers/hyrax/dashboard/collections_controller.rb#L30
I'm not sure of the reason for the exception carved out here, it only applies to `/dashboard/collections`. The other similar URLs (`/dashboard/works`, `/dashboard/my/collections`, `/dashboard/my/works`) all send unauthenticated users to the login page.

It seems that the rule was added in as part of this issue, but i don't see anything in the description to indicate that the page should be open to the public:
https://github.com/samvera/hyrax/issues/1493
There also don't appear to be tests for the `#index` route so it's hard to tell if it was intentional, while the other tests involve being logged in to access:
https://github.com/samvera/hyrax/blob/main/spec/controllers/hyrax/dashboard/collections_controller_spec.rb

This error applies to hyrax 2, 3, 4 and 5 (main branch). I can make a PR to remove the before action, but wasn't sure if i was missing something

The traceback was a little disconcerting since it sounds like it's trying to modify users when the user isn't logged in, but it fails so its probably harmless?:
```

ActiveRecord::RecordNotSaved in Hyrax::Dashboard::CollectionsController#index
You cannot call create unless the parent is saved

activerecord (6.1.7.3) lib/active_record/associations/has_one_association.rb:125:in `_create_record'
activerecord (6.1.7.3) lib/active_record/associations/association.rb:209:in `create!'
activerecord (6.1.7.3) lib/active_record/associations/builder/singular_association.rb:37:in `create_sipity_agent!'
hyrax (e9141b81865a) app/models/concerns/hyrax/user.rb:72:in `to_sipity_agent'
hyrax (e9141b81865a) app/search_builders/hyrax/dashboard/managed_search_filters.rb:39:in `add_managing_role_search_filter'
hyrax (e9141b81865a) app/search_builders/hyrax/dashboard/managed_search_filters.rb:14:in `apply_group_permissions'
blacklight-access_controls (6.0.1) lib/blacklight/access_controls/enforcement.rb:45:in `block in gated_discovery_filters'
[...]

PR #6726, opened 2024-02-28
Prevent unauthenticated access to /dashboard/collections/

bbpennel:
### Fixes

https://github.com/samvera/hyrax/issues/6725

### Summary

Enables authentication for the `/dashboard/collections/` endpoint, same as for `/dashboard/my/collections/`

* With Dassie, prior to checking out this branch, visit http://localhost:3000/dashboard/collections/ and experience the error
* Check out this branch and visit the url again, there should no longer be an error

notes-bugfix

### Detailed Description

See the issue for more details.

Add tests for dashboard/CatalogController#index which fail when not logged in prior to this update. Removes authentication exception for the index endpoint from this catalog controller, since it results in errors. Removing it here causes the controller to use the check from its parent class.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/hyrax/dashboard/collections_controller.rb
+++ b/app/controllers/hyrax/dashboard/collections_controller.rb
@@ -26,9 +26,6 @@ module Hyrax
       # Catch permission errors
       rescue_from Hydra::AccessDenied, CanCan::AccessDenied, with: :deny_collection_access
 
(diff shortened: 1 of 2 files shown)


dlpierce (approved):
Thanks for catching this! Let's go with this and perhaps the old `ensure_admin!` method can be cleaned up later.

PR #6741, opened 2024-03-18
chart: add support for startupProbe

dunn:
See
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

This enables apps with long startups to not be killed prematurely.

@samvera/hyrax-code-reviewers

### DIFF
--- a/chart/hyrax/Chart.yaml
+++ b/chart/hyrax/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: hyrax
 description: An open-source, Samvera-powered digital repository system
 type: application
-version: 3.6.0
+version: 3.7.0
 appVersion: 5.0.0
 dependencies:
   - name: fcrepo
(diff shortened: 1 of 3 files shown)


no-reply (approved):

no-reply (approved):
this looks great.

no idea why CI is failing, but it's clearly not related

dunn:
yeah, weird
```
Error: error resolving source context: reference not found
Usage:
  executor [flags]
  executor [command]
```

dunn:
maybe a CI cache issue?

orangewolf (approved):

dlpierce:
I wonder if this is caused by the slash in the branch name on Circle (`pull/6741`) being misinterpreted by kaniko?

dunn:
oh that could be; why would that suddenly be a problem, though?

no-reply:
@dunn can you resubmit this as a PR from `samvera/hyrax` instead of your fork?

dunn:
ohhh gotcha

dunn:
https://github.com/samvera/hyrax/pull/6750

PR #6743, opened 2024-03-20
move file related listeners to their own space

no-reply:
these listener methods don't have to do with `FileMetadata`, so they should be reorganized out into their own class. this makes it easier for downstream apps to override behaviors that are only related to files.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/publisher.rb
+++ b/lib/hyrax/publisher.rb
@@ -124,6 +124,16 @@ module Hyrax
 
     # @since 3.5.0
     # @macro a_registered_event
+    #   @note this event SHOULD be published file is characteried by the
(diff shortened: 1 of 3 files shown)


lib/hyrax/publisher.rb
```diff
@@ -208,6 +218,7 @@ def default_listeners
       @default_listeners ||=
         [Hyrax::Listeners::ACLIndexListener.new,
          Hyrax::Listeners::BatchNotificationListener.new,
+         Hyrax::Listener::FileListener.new,
```
no-reply, line 221:
```suggestion
         Hyrax::Listeners::FileListener.new,
```

dlpierce (approved):

PR #6747, opened 2024-03-26
avoid n+1 query on file_metadata delete

no-reply:
Hyrax should avoid looping over ids when it could use a custom query. e.g. we should never be doing `my_resource.member_ids.each` and looking up members instead doing `query_service.find_members(resource: my_resource)`. doing this creates two problems:

  - it prevents adapters from optimizing n+1 query problems for large sets of ids.
  - it locks hyrax into modelling details (relationship directionality) that applications may want to avoid.

there are probably other places this problem exists, but i'm aware of this one and wanted to get a patch in right away.

@samvera/hyrax-code-reviewers

### DIFF
--- a/lib/hyrax/transactions/steps/delete_all_file_metadata.rb
+++ b/lib/hyrax/transactions/steps/delete_all_file_metadata.rb
@@ -14,8 +14,7 @@ module Hyrax
 
         ##
         # @params [#save] persister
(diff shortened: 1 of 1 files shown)


dlpierce (approved):
Thanks! No reason to have n+1 situations. Thanks for offering to track down any others we've created.

PR #6752, opened 2024-03-26
use the `Hyrax::FileSetFileService` to determine "primary file"

no-reply:
`Hyrax::FileSetFileService` was introduced to un-hard code the assumption that the `:original_file` PCDM Use is the "primary" file for a given file set. use it throughout the code base to avoid requiring that assumption in downstream apps.

this should allow downstream apps to wire in a new service and resolve the "primary file" for each object according to local requirements. e.g. it may be desirable for apps to treat "preservation file" or "service file" as "primary"

* adds support for overriding the "primary" file resolution behavior

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/file_set_file_service.rb
+++ b/app/services/hyrax/file_set_file_service.rb
@@ -32,7 +32,16 @@ module Hyrax
     end
 
     ##
-    # Return the Hyrax::FileMetadata which should be considered “primary” for
+    # Return the {Hyrax::FileMetadata} which should be considered “primary” for
+    # indexing and version‐tracking.
+    #
+    # @return [Hyrax::FileMetadata]
+    def self.primary_file_for(file_set:, query_service: Hyrax.query_service)
+      new(file_set: file_set, query_service: query_service).primary_file
+    end
+
+    ##
+    # Return the {Hyrax::FileMetadata} which should be considered “primary” for
     # indexing and version‐tracking.
     #
     # If +file_set.original_file_id+ is defined, it will be used; otherwise,
--- a/app/controllers/hyrax/file_sets_controller.rb
+++ b/app/controllers/hyrax/file_sets_controller.rb
@@ -292,7 +292,7 @@ module Hyrax
     end
 
     def file_metadata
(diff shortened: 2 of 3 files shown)


app/services/hyrax/file_set_file_service.rb
```diff
@@ -32,7 +32,16 @@ def initialize(file_set:, query_service: Hyrax.query_service)
     end
 
     ##
-    # Return the Hyrax::FileMetadata which should be considered “primary” for
+    # Return the {Hyrax::FileMetadata} which should be considered “primary” for
+    # indexing and version‐tracking.
+    #
+    # @return [Hyrax::FileMetadata]
+    def self.primary_file_for(file_set:, query_service: Hyrax.query_service)
+      new(file_set: file_set, query_service: query_service).primary_file
```
marrus-sh, line 40: The fallback in `#primary_file` (at least) does

```rb
fallback_id = file_set.file_ids.first
```

This should probably use a query service too instead, yeah? (Maybe we need a “first file metadata in file set” query?)

I’d also very much like to get rid of the dependence on `file_set.original_file_id` in that method (which I think suffers from the same modelling issues that `file_ids` does), but I couldn’t figure out a way how when I wrote that method originally. Is there a query we could use for that instead?

(The difference between the service and any original file query is, and should remain, that the service has fallback behaviour if there is no original file, whereas the query would return `nil`.)

no-reply: i think you're probably right, on both points.

i remember the #primary_file setup being pretty fussy though (between Wings and PG), and i think for our purposes it's moot anyway, since we want to override this whole service in comet. getting this much through unblocks that, and next steps can be explored from there.

no-reply: re: `original_file_id`, i think `custom_queries.find_original_file(file_set: file_set)` is the preferred approach, and *is* executed in the other branch. but i presume there's a reason for `original_file_id` getting used preferentially.

dlpierce (approved):

PR #6758, opened 2024-03-27
Fix “::Collection” definition with autoloading

### LINKED ISSUE #6736: Rails autoloading fails when ::Collection is undefined
The change made in 53b0060 doesn’t actually work in environments with autoloading when there is no `::Collection`; the `.safe_constantize` creates a recursive loop which just tries to load the same file again.

We got around this by just doing `Object.const_set(:Collection, Hyrax.config.collection_class)` in an initializer, but this isn’t ideal.

marrus-sh:
In autoloaded environments which reference `::Collection` in an initializer, `safe_constantize` will create a recursive autoloading loop. We can hook into `ActiveSupport::Dependencies` to check whether this is going to happen, and pre·emptively return `false` in that case (so that `::Collection` is defined properly), without changing behaviour where `::Collection` is getting `safe_constantize`d from somewhere else.

### Fixes

Fixes #6736

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -2,4 +2,16 @@
 
 # Provide plain collection model if not defined by the application.
 # Needed until Hyrax internals do not assume its existence.
-class ::Collection < Hyrax.config.collection_class; end unless '::Collection'.safe_constantize
+class ::Collection < Hyrax.config.collection_class; end unless ActiveSupport::Dependencies.then do |deps|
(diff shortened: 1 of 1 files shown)


no-reply (commented):
fuunnnnnnnn.

i'd love if this got cross institutional review, but it checks out to me.

dlpierce (approved):
I had encountered the circular dependency issue too, thanks for the improvement!

PR #6763, opened 2024-03-29
♻️ Favor method_missing over a raft of delegates

jeremyf:
It's like we're flooding a political convention with delegates.  Instead of having all of those delegate methods, let's just implement method missing and pass things along to the `SolrDocument`.

The benefit is that as we incorporate dynamic metadata, we don't need to keep on adding `delegate :this_property, to: :solr_document`.

@samvera/hyrax-code-reviewers


dlpierce:
I like this. I was surprised to still need to customize the presenter (and other files) when we have the metadata schema yaml files in place. However, I'm going to wait until after 5.0.1 to merge this.

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -33,20 +26,14 @@ def initialize(solr_document, current_ability, request = nil)
       @request = request
     end
 
+    # We cannot rely on the method missing to catch this delegation.  Because
+    # most all objects implicitly implicitly implement #to_s
```
bkiahstroud, line 30: So nice you can say it twice (I'm guessing this was accidental) 

```suggestion
    # most all objects implicitly implement #to_s
```

Issue #6775, opened 2024-04-17
Based Near Prepopulation Creates Errors


jeremyf:
### Descriptive summary

The following spec for [./spec/forms/hyrax/forms/resource_form_spec.rb](./spec/forms/hyrax/forms/resource_form_spec.rb) fails:

``` ruby
  describe '#based_near' do
    it 'has a default of an empty based near' do
      expect(work.based_near).to be_empty
      form.prepopulate!
      expect(form.based_near).to be_empty
    end
  end
```

This is a bit surprising.  The issue with this failing is that later on we have problems with the based_near lookup service.

Namely we have the following issue:

``` ruby
ArgumentError:
  g637840 (Valkyrie::ID) is not a valid type
  
# app/services/hyrax/location_service.rb:28:in `extract_id`
```

Which is happening in the following location: [./app/indexers/concerns/hyrax/location_indexer.rb](./app/indexers/concerns/hyrax/location_indexer.rb).

Note this is showing up as a Hyku error.

PR #6790, opened 2024-04-24
:bug:Account for Valkyrie ID in location indexer

ShanaLMoore:
Without passing a string, we get an error when based_near is nil. This was discovered when testing Bulkrax in a valkyrized app. When we uploaded a CSV with no based_near defined, I was getting an error `*** ArgumentError Exception: g410220 is not a valid type` from the #extract_id method. 

- [ ] https://github.com/samvera/hyrax/issues/6775

```ruby
    def based_near_prepopulator
      self.based_near = based_near.map do |loc|
        uri = RDF::URI.parse(loc)
        if uri
          Hyrax::ControlledVocabularies::Location.new(uri)
        else
          loc
        end
      end
```
**based_near => [#<Valkyrie::ID:0x0000ffff5f734970 @id="g410220">]**

This is because at this point, obj is a neither a string or a URI. It's a Valkyrie::ID.

```ruby
    def extract_id(obj)
      uri = case obj
            when String
              URI(obj)
            when URI
              obj
            else
              raise ArgumentError, "#{obj} is not a valid type"
            end
      uri.path.split('/').last
    end
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/location_service.rb
+++ b/app/services/hyrax/location_service.rb
@@ -19,7 +19,7 @@ module Hyrax
 
     def extract_id(obj)
       uri = case obj
-            when String
+            when String, Valkyrie::ID
               URI(obj)
             when URI
               obj


no-reply (approved):
seems good, approving.

i think a unit test would improve things---it would be really easy to regress this in a refactor.

i also wonder whether a refactor like `uri = URI(obj)` wouldn't do the trick? `URI()` will already raise an informative error message if you try to cast a type it can't handle:

> `URI': bad argument (expected URI object or URI string) (ArgumentError)

this seems better than the ambiguous "not a valid type" message you get in our code.

----

looking at this closer, i definitely think we should do this refactor. it opens the whole class to extension in a way the type check strictly forecloses. if i can pass a type that ruby `URI()` will cast, i should be able to use it here.

ShanaLMoore:
> looking at this closer, i definitely think we should do this refactor. it opens the whole class to extension in a way the type check strictly forecloses. if i can pass a type that ruby `URI()` will cast, i should be able to use it here.

@no-reply Actually, I think the problem is here. It's forcing based_near to have a value even though we aren't declaring one. If we don't do this, based_near remains an empty [ ] . Is there an issue with removing[ this line? ](https://github.com/samvera/hyrax/blob/main/app/forms/concerns/hyrax/basic_metadata_form_fields_behavior.rb#L36)

image

PR #6793, opened 2024-05-09
WIP - Address valkyrie derivative issues (includes double_combo... do not merge)

laritakr:
**Keeping this in draft mode... This PR will be canceled once the other PR is merged into main, and then main is merged into double_combo.** 

Valkyrie derivatives sort of work but only by accident. We need to access them via the storage adapter, rather than assuming we will find them on the file system. This begins the process of accessing derivatives appropriately.

The idea is to always include the options to download derivatives on the actions dropdown. 

The actions download menu does not have a Hyrax::FileSet object... it has a Hyrax::IiifAv::IiifFileSetPresenter object (at least when we are using IiifPrint). The `extensions_and_mime_types` method we added to the fileset gives us an array of hashes that we can use for the actions dropdown:
```
 [{:id=>"123", :extension=>"pdf", :mime_type=>"application/pdf", :name=>nil, :use=>"OriginalFile"},
    #    {:id=>"234", :extension=>"jpeg", :mime_type=>"application/octet-stream", :name=>"thumbnail", :use=>"ThumbnailImage"}]
```

### DIFF
--- a/app/indexers/hyrax/indexers/file_set_indexer.rb
+++ b/app/indexers/hyrax/indexers/file_set_indexer.rb
@@ -101,6 +101,9 @@ module Hyrax
 
           # attributes set by fits for video
           solr_doc['aspect_ratio_tesim']      = file_metadata.aspect_ratio if file_metadata.aspect_ratio.present?
+
+          # support for derivatives download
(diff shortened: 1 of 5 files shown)


laritakr:
Pull request passes specs per https://github.com/samvera/hyrax/actions/runs/9101502970

PR #6795, opened 2024-05-15
Address valkyrie derivative issues

laritakr:
Valkyrie derivatives sort of work but only by accident. We need to access them via the storage adapter, rather than assuming we will find them on the file system. This begins the process of accessing derivatives appropriately.

The idea is to always include the options to download derivatives on the actions dropdown. 

The actions download menu does not have a Hyrax::FileSet object... it has a Hyrax::IiifAv::IiifFileSetPresenter object (at least when we are using IiifPrint). The `extensions_and_mime_types` method we added to the fileset gives us an array of hashes that we can use for the actions dropdown:
```
 [{:id=>"123", :extension=>"pdf", :mime_type=>"application/pdf", :name=>nil, :use=>"OriginalFile"},
    #    {:id=>"234", :extension=>"jpeg", :mime_type=>"application/octet-stream", :name=>"thumbnail", :use=>"ThumbnailImage"}]
``` 

<details>
<summary> Screenshots </summary>

Screenshot 2024-05-15 at 4 42 42 PM

Screenshot 2024-05-15 at 4 42 51 PM

</details>

### DIFF
--- a/app/indexers/hyrax/indexers/file_set_indexer.rb
+++ b/app/indexers/hyrax/indexers/file_set_indexer.rb
@@ -104,7 +104,7 @@ module Hyrax
 
           # support for derivatives download
           derivatives = resource.extensions_and_mime_types
-          solr_doc['extensions_and_mime_types_ssm'] = derivatives if derivatives.present?
+          solr_doc['extensions_and_mime_types_ssm'] = derivatives.to_json if derivatives.present?
         end
       end
 
--- a/app/indexers/hyrax/indexers/file_set_indexer.rb
+++ b/app/indexers/hyrax/indexers/file_set_indexer.rb
@@ -101,6 +101,9 @@ module Hyrax
 
           # attributes set by fits for video
           solr_doc['aspect_ratio_tesim']      = file_metadata.aspect_ratio if file_metadata.aspect_ratio.present?
+
+          # support for derivatives download
+          solr_doc['extensions_and_mime_types_ssim'] = resource.extensions_and_mime_types.to_json if resource.extensions_and_mime_types.present?
         end
       end
 
--- a/app/indexers/hyrax/indexers/file_set_indexer.rb
+++ b/app/indexers/hyrax/indexers/file_set_indexer.rb
@@ -103,7 +103,8 @@ module Hyrax
           solr_doc['aspect_ratio_tesim']      = file_metadata.aspect_ratio if file_metadata.aspect_ratio.present?
 
           # support for derivatives download
-          solr_doc['extensions_and_mime_types_ssim'] = resource.extensions_and_mime_types.to_json if resource.extensions_and_mime_types.present?
+          derivatives = resource.extensions_and_mime_types
+          solr_doc['extensions_and_mime_types_ssm'] = derivatives if derivatives.present?
         end
       end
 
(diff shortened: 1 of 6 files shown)


app/indexers/hyrax/indexers/file_set_indexer.rb
```diff
@@ -101,6 +101,9 @@ def to_solr # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Met
 
           # attributes set by fits for video
           solr_doc['aspect_ratio_tesim']      = file_metadata.aspect_ratio if file_metadata.aspect_ratio.present?
+
+          # support for derivatives download
+          solr_doc['extensions_and_mime_types_ssim'] = resource.extensions_and_mime_types.to_json if resource.extensions_and_mime_types.present?
```
dlpierce, line 106: - The field type probably doesn't need to be indexed.
- This is doing the work of `extensions_and_mime_types` twice.
- Could/should this be stored as multiple entries in the solr field instead of a single?

app/models/hyrax/file_set.rb
```diff
@@ -106,6 +106,32 @@ def extracted_text_id
...
+        arr << {
+          id: fm.id.to_s,
+          extension: extension[1..], # remove leading '.'
+          mime_type: fm.mime_type,
+          name: name,
+          use: use
+        }
+      end
+      # rubocop:enable Metrics/MethodLength
+    end
```
dlpierce, line 133: This needs a spec test of some sort to verify the generated json.

app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -45,8 +45,16 @@
                     class: "download",
                     data: { label: file_set.id, work_id: @presenter.id, collection_ids: @presenter.member_of_collection_ids } %>
       </li>
-    <% end %>
 
+      <% file_set.extensions_and_mime_types&.each do |hash| %>
+        <% next unless hash[:name] %>
+        <li class ="dropdown-item" role="menuitem" tabindex="-1">
+          <a href="<%= "/downloads/#{file_set.id}?locale=en&file=#{hash[:name]}&mime_type=#{hash[:mime_type]}" %>" download>
```
dlpierce, line 52: Can this use a *_url helper?

dlpierce (changes requested):

app/indexers/hyrax/indexers/file_set_indexer.rb
laritakr: Indexing it allows us to access the information on a view without having to access the data store directly. Generally solr has been the preference for views. However, you are right that it is probably optional to do this... is it your preference to not index it?

I'm not sure I understand the benefit of storing it individually. This would generally be used when you want to access all derivatives which is how we are using it (existing methods access a derivative for a particular use).

dlpierce: I meant the solr suffix should be `ssm` to avoid solr trying to analyze the contents, not that it should not be in solr at all.

dlpierce: I was thinking storing as multiple values would make the solr results easier to read, but would require more processing at load time. It's not that important.

laritakr:
@dlpierce I believe I've addressed your feedback requests. Everything passed except that flaky 'email' test.

dlpierce (approved):

PR #6806, opened 2024-05-20
Set mock response for feature in deposit agreement spec

dlpierce:
The double sign in seems to flake out when trying to find the email field. Let's mock the feature directly to avoid the need to sign in as an admin before signing in as a regular user.

### Fixes

Fixes flaky features/deposit_agreements_spec


laritakr (approved):
Much better!

Issue #5963, opened 2023-02-15
Error visiting dashboard work analytics page


bbpennel:
### Descriptive summary

I am seeing a failure when visiting the Dashboard > Works Report page, and then clicking on the title of any works under the Top Works panel. This can be demonstrated on nurax, for example:

https://nurax-dev.curationexperts.com/admin/analytics/work_reports/ht24wj68c?locale=en

### Expected behavior

Not really sure what this page is supposed to display, haven't seen it work

### Actual behavior

Error is thrown. In nurax the error is:
```
nil can't be coerced into Integer

Extracted source (around line #4):
  <div class="col-12 work-counts">
    <div class="work-count-border col-sm-2 offset-sm-3 float-left text-center">
       <% @file_sizes_total_bytes = @file_sizes.inject(0, :+) %>
       <% @file_size_total = (@file_sizes_total_bytes / 1.0.megabyte).round(2) %>
      <h4 class="work-count-title"><%= @file_size_total %>MB</p>
      <p><%= t('.total_size') %></p>

hyrax (76176286b817) app/views/hyrax/admin/analytics/work_reports/_work_counts.html.erb:4:in `+'
hyrax (76176286b817) app/views/hyrax/admin/analytics/work_reports/_work_counts.html.erb:4:in `inject'
hyrax (76176286b817) app/views/hyrax/admin/analytics/work_reports/_work_counts.html.erb:4
actionview (6.0.5.1) lib/action_view/base.rb:274:in `_run'
actionview (6.0.5.1) lib/action_view/template.rb:185:in `block in render'
activesupport (6.0.5.1) lib/active_support/notifications.rb:182:in `instrument'
actionview (6.0.5.1) lib/action_view/template.rb:385:in `instrument_render_template'
actionview (6.0.5.1) lib/action_view/template.rb:183:in `render'
actionview (6.0.5.1) lib/action_view/renderer/partial_renderer.rb:357:in `block in render_partial'
actionview (6.0.5.1) lib/action_view/renderer/abstract_renderer.rb:88:in `block in instrument'
activesupport (6.0.5.1) lib/active_support/notifications.rb:180:in `block in instrument'
activesupport (6.0.5.1) lib/active_support/notifications/instrumenter.rb:24:in `instrument'
activesupport (6.0.5.1) lib/active_support/notifications.rb:180:in `instrument'
actionview (6.0.5.1) lib/action_view/renderer/abstract_renderer.rb:87:in `instrument'
[...]

Issue #5818, opened 2022-08-26
Admin Analytic's Work Reports renders in error when attempting to call `strftime` on a String.


jeremyf:
In `app/views/hyrax/admin/analytics/work_reports/show.html.erb` we have cases where `@document.date_uploaded` is not a Date-like object.  When that view attempts to call `strftime` on the `@document.date_uploaded` we get the following error: `undefined method strftime' for "2022-08-10T23:22:45Z":String`.

Throughout the code base the `@document.date_uploaded` is rendered directly (e.g. no conversion).  In the [Hyrax::StatsUsagePresenter object](https://github.com/samvera/hyrax/blob/27d62a2338bd9e49f942e7e73829f91f2e8115eb/app/presenters/hyrax/stats_usage_presenter.rb#L38-L42) we do ensure a date like value.  Which indicates that this field cannot predicatably be called a Date.

Proposal:  Add a helper method `cast_to_formatted_date(value, format: "%Y-%m-%d")` this would attempt to parse the value from String to date then render with the given format.  Port this to the `3.4.x` and `4.0.x` branches.

PR #6815, opened 2024-05-23
create google analytics 4 compatibility

### LINKED ISSUE #5970: Google Analytics 4 Support
Per Google, on July 1, 2023, Universal Analytics properties will no longer process data. Hyrax will need to support GA4. 

`HYRAX_ANALYTICS_PROVIDER` could have an additional option (`ga4`) or GA4 properties could be detected by checking the prefix. Adding a new js snippet for GA4 properties will be relatively easy. 

It looks like GA4 supports turbolinks page navigations [out of the box](https://support.google.com/analytics/answer/9216061) (need to test) so event tracking will be simplified. No need for `setupTracking()` on every page refresh. But Hyrax file download links don't end in extensions, so there's work to do to make sure file downloads are tracked. 

The GA4 API is different enough that the easy interop between matomo and UA won't work with GA4. Instead of pushing events to a special array, you need to call the `gtag` function. For backwards compatibility, we've updated the `TrackingTags` class's `analytics()` function like so:

```
analytics() {
    if(this.provider === "matomo") {
      return _paq;
    }
    else {
      return {
        push: function(e){
          if (Array.isArray(e) && e.length == 4) {
            window.gtag('event', e[1], {'action': e[2], 'name': e[3]});
          }
        }
      };
    }
  }
```

That says 'if you're matomo, return the `_paq` array object which works as expected. If not (in our env), we're using GA4. In that case, return a new object with a function called 'push' which translates the array pushes from UA/Matomo into `gtag()` calls'. It's just a quick hack to prevent js console errors on our test IRs.

Future work:
1. Include a GA4 js snippet if the analytics provider is declared to be (or detected to be) GA4 instead of UA or Matomo. 
2. If GA4 is being used, don't fire any events GA4 can handle natively. 
3. If GA4 is being used, translate custom analytics events from the old to new API. 
4. Document setting up a GA4 property with [Enhanced event measurement](https://support.google.com/analytics/answer/9216061).
5. Investigate the impact the UA->GA4 move will have on exposing analytics in the Hyrax UI.

orangewolf:
### Fixes
Fixes #5970, #5963, and #5818 and brings GA4 analytics in to the fold. Reports are fixed to render and GA4 tracking now actually records the right information.

### Summary

* Set up a Google Analytics 4 account for testing. Set `GOOGLE_ANALYTICS_ID` `GOOGLE_ANALYTICS_PROPERTY_ID` `GOOGLE_ACCOUNT_JSON`
* View some pages
* See that analytics reports show up and render properly

* Add _ga4.html partial for displaying GA4 code
* Add ga4 support to analytics-events.js
* Ability to read GA4 analytics API and fill in the existing views

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/analytics/matomo.rb
+++ b/app/services/hyrax/analytics/matomo.rb
@@ -154,6 +154,11 @@ module Hyrax
           response["nb_visits_returning"].to_i + response["nb_visits_new"].to_i
         end
 
+        # TODO implement
+        def page_statistics(start_date, object)
+
+        end
+
         def results_array(response, metric)
           results = []
           response.each do |result|
--- a/app/services/hyrax/analytics/matomo.rb
+++ b/app/services/hyrax/analytics/matomo.rb
@@ -154,8 +154,8 @@ module Hyrax
           response["nb_visits_returning"].to_i + response["nb_visits_new"].to_i
         end
 
-        # TODO implement
-        def page_statistics(start_date, object)
+        # TODO: implement
+        def page_statistics(_start_date, _object)
           []
         end
 
--- a/app/services/hyrax/analytics/matomo.rb
+++ b/app/services/hyrax/analytics/matomo.rb
@@ -156,7 +156,7 @@ module Hyrax
 
         # TODO implement
         def page_statistics(start_date, object)
-
+          []
         end
 
         def results_array(response, metric)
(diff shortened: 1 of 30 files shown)


app/services/hyrax/analytics/matomo.rb
```diff
@@ -154,6 +154,11 @@ def total_visitors(period = 'month', date = 'today')
           response["nb_visits_returning"].to_i + response["nb_visits_new"].to_i
         end
 
+        # TODO implement
```
orangewolf, line 157: I didn't fix Matomo for this release, but I honestly think that this may be the only part that needs to be updated for Matomo 4 to work again. There is another task for adding Matomo 5

PR #6818, opened 2024-05-31
Statistics fixes

orangewolf:
### Fixes

Some Hyrax statistics models still used non-Valkyrie work types

### Summary

Move to Valkyrie work types for all Hyrax::Statistics objects

@samvera/hyrax-code-reviewers


app/services/hyrax/solr_service.rb
```diff
@@ -111,6 +111,20 @@ def query(query, **args)
       end
     end
 
+    def query_in_batches(query, **args)
```
orangewolf, line 114: written this by hand a bunch of times to do stuff in the console. finally found a reason to make it something permanent.

PR #6825, opened 2024-06-04
Change default admin set id to `admin_set_default` to avoid issues with encoded slashes

abelemlih:
### Fixes

Fixes Fedora 6.5+ compatibility, since it does not support encoded slashes in IDs.

### Summary

We currently cannot use Fedora 6.5+ with Hyrax because it does not support encoded slashes in IDs. This leads to failure every time we perform `Hyrax.query_service.find_by(id: DEFAULT_ID)`. In this PR, I change the `DEFAULT_ID` to use underscores only, and add backwards compatibility changes to helper methods finding existing admin sets.

- `notes-bugfix` Bug Fixes

### Detailed Description

After delving into the use of `DEFAULT_ID` , it’s only ever used as an actual ID in places that have a hardcoded admin set with that ID, which is not the case for Postgres or Fedora (Fedora uses UUID). The reasoning behind my PR is, to support backwards compatibility, when we need to lookup any existing unused admin set with default id, lookup for both `admin_set/default`  and `admin_set_default`, if none exist then go forward with creating a new admin set, which will use UUID with Valkyrie. This change ensures new apps using underscores if they ever have to create a default admin set with `DEFAULT_ID` , and supports any existing apps that do not use Postgres/Fedora via Valkyrie and have an admin set with the id explicitly set to `admin_set/default`.

one edge case to this approach is, if there is a repository that has a default admin set in Fedora that has an id explicitly set to `admin_set/default` , then it will still break, because Fedora 6.5+ will break every time it looks it up. My understanding is we store the default admin set id in Fedora with UUID, and we store the UUID in the `hyrax_default_administrative_set`  on the database.

* change the `DEFAULT_ID` for administrative sets to use underscores only
* Change helper method `find_unsaved_default_admin_set` to lookup both `admin_set/default` and `admin_set_default` when looking for any existing unused default admin sets.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/admin_set_create_service.rb
+++ b/app/services/hyrax/admin_set_create_service.rb
@@ -10,7 +10,7 @@ module Hyrax
   # @see Hyrax::PermissionTemplate
   # @see Sipity::Workflow
   class AdminSetCreateService # rubocop:disable Metrics/ClassLength
-    DEFAULT_ID = 'admin_set/default'
+    DEFAULT_ID = 'admin_set_default'
     DEFAULT_TITLE = ['Default Admin Set'].freeze
 
     class_attribute :permissions_create_service, :default_admin_set_persister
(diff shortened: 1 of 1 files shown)


abelemlih:
For reference:

After delving into the use of `DEFAULT_ID` , it’s only ever used as an actual ID in places that have a hardcoded admin set with that ID, which is not the case for Postgres or Fedora (Fedora uses UUID). The reasoning behind my PR is, to support backwards compatibility, when we need to lookup any existing unused admin set with default id, lookup for both `admin_set/default`  and `admin_set_default`, if none exist then go forward with creating a new admin set, which will use UUID with Valkyrie. This change ensures new apps using underscores if they ever have create a default admin set with `DEFAULT_ID` , and supports any existing apps that do not use Postgres/Fedora via Valkyrie and have an admin set with the id explicitly set to `admin_set/default`.

one edge case to this approach is, if there is a repository that has a default admin set id in Fedora that has an id explicitly set to `admin_set/default` , then it will still break, because Fedora 6.5+ will break every time it looks it up. My understanding is we store the default admin set id in Fedora with UUID, and we store the UUID in the `hyrax_default_administrative_set`  on the database.

randalldfloyd (commented):
@abelemlih This is how I would have approached it as well; did you notice that a large number of tests are marked as failing?  They all have the same root cause, which seems related:
```
RuntimeError:
  Corrupt default admin set.  Persisted admin set with saved default_admin_set_id doesn't exist.
./app/services/hyrax/admin_set_create_service.rb:119:in `rescue in find_default_admin_set'
./app/services/hyrax/admin_set_create_service.rb:103:in `find_default_admin_set'
./app/services/hyrax/admin_set_create_service.rb:27:in `find_or_create_default_admin_set'
./app/controllers/concerns/hyrax/works_controller_behavior.rb:177:in `admin_set_id_for_new'
```

randalldfloyd (approved):
This seems to cover all of the use cases for both kinds of models.  I couldn't think of any new specs to add to how this service is already being tested.  

@dlpierce , thoughts?

randalldfloyd:
@abelemlih @dlpierce 

Just another FYI... I have ran this locally as part of other work I'm doing, and the sirenia app runs using `fcrepo:6.5-tomcat9` without breaking on startup in the admin set service.  I can merge soon if no other reviewers want to ring in on it.

abelemlih:
@dlpierce I rebased this PR and it is now ready to review. Let me know if you have any feedback, thanks!

dlpierce (approved):

PR #6833, opened 2024-06-12
GH Actions workflow improvements

dlpierce:
### Summary

Only run debug step by request.
Only copy reports when fetching reports.


dlpierce:
Was merged as part of #6817

PR #6843, opened 2024-06-24
DRAFT DO NOT MERGE: :gift: Flexible double combo

kirkkwang:
# Purpose

This is the working branch for all things flexible metadata related.


lib/generators/hyrax/templates/db/migrate/20240606205215_create_hyrax_flexible_schemas.rb.erb
```diff
@@ -0,0 +1,9 @@
+class CreateHyraxFlexibleSchemas < ActiveRecord::Migration<%= migration_version %>
```
ShanaLMoore, line 1: Should this have a `unless Hyrax.config.flexible?` guard?

laritakr: I'm not sure. Does it hurt to have the models present but not use them?

PR #6846, opened 2024-06-25
Move clean_repo before hook after :index_adapter

dlpierce:
### Summary

Fixes flakiness in by_format_spec.rb

### Detailed Description
If an rspec example specified both :clean_repo and :index_adapter, the incorrect solr adapter would be cleaned leading to sporadic test failures.

Flakiness seen when 
`spec/controllers/hyrax/file_sets_controller_spec.rb[1:1:1:4:1:1]`
ran prior to 
`spec/services/hyrax/statistics/file_sets/by_format_spec.rb[1:1:1]`

* Reordering of rspec before hooks

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -282,14 +282,6 @@ RSpec.configure do |config|
   #   skip("Don't test Wings") if Hyrax.config.disable_wings
   # end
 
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
Good catch!

PR #6850, opened 2024-06-25
Dynamic attribute rows

kirkkwang:
## WIP getting attribute_rows to be dynamic

f19ef185a2e83f4d0491a40fe863579ad6894765

## 🧹 Enables dynamic display for attribute rows

2282c7ed5ccc7fd98768af39e0fdf957c7a0b382

In this commit we updated code to support dynamic rendering for attribute rows

## 🚧 WIP - handle dynamic translations

32685a7769a783d77257ca1b3b1d676b0d8df779

## Expand support for i18n labels

4740ec8d3176ee59bfe7f53120a410181ae57fed

This commit will add a few different conditions for populating the
label.  If the label is a string we'll simply set that as the label.

ex. `label: 'My Label'`

If the label is a hash we'll look for the key that matches the current
locale.

ex. `label: { en: 'My Label', es: 'Mi Etiqueta' }`

If the locale is not found it will fall back to en as the default, but
if en isn't found it will just use the first value in the hash.

ex. `label: { es: 'Mi Etiqueta', fr: 'Mon Étiquette' }`

## Remove unnecessary quotes

b21a12b718a3e318dece41f697419afae71c6aac


app/helpers/hyrax/attributes_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def conform_options(options)
+      hash_of_locales = options['view']['label'] || {}
+      current_locale = params['locale'] || I18n.locale.to_s
+      updated_options = options.deep_dup
+      return updated_options['view'].transform_keys(&:to_sym) if hash_of_locales.is_a?(String)
+
+      # If the params locale is found in the hash of locales, use that value
+      if hash_of_locales[current_locale].present?
```
ShanaLMoore, line 20: since  updated_options['view']['label'] = is in each condition block, you can move it to the outside like...

  # Determine the label based on the locale
  updated_options['view']['label'] =
    if hash_of_locales[current_locale].present?
      hash_of_locales[current_locale]
    else
      hash_of_locales['en'] || hash_of_locales.values.first
    end

ShanaLMoore (dismissed):

PR #6857, opened 2024-06-27
Install Alpine's grpc gem and use in test apps

dlpierce:
### Fixes

refs #6815 

### Summary
Make the test apps use Alpine's provided grpc gem package to avoid long compilation.

* Image build time in Actions workflow is improved

### Detailed Description

The grpc gem uses native extensions that take a long time to compile. Normally the gem can provide prebuilt binaries to avoid compilation, but Alpine Linux which we use as the base for hyrax's docker image is not compatible with those binaries. Alpine provides its own packages, but we have to match versions and specify the in-container path.

The other commits here are addressing flaky tests that made themselves known along the way.

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/Gemfile
+++ b/.dassie/Gemfile
@@ -14,9 +14,18 @@ gem 'coffee-rails', '~> 4.2'
 gem 'dalli'
 gem 'devise'
 gem 'devise-guests', '~> 0.8'
+
+# Required because grpc and google-protobuf gem's binaries are not compatible with Alpine Linux.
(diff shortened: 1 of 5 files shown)


dlpierce:
Ran into some flaky specs, but initial results look good. The build step is about 10 minutes faster!

laritakr (approved):

orangewolf (commented):
This is great @dlpierce

PR #6858, opened 2024-06-27
Version badge

kirkkwang:
This is the work to get the badges that show what version of the schema the work/collection/file_set is currently on.

### DIFF
--- a/app/indexers/hyrax/indexers/resource_indexer.rb
+++ b/app/indexers/hyrax/indexers/resource_indexer.rb
@@ -39,7 +39,7 @@ module Hyrax
       # @api public
       # @return [HashWithIndifferentAccess<Symbol, Object>]
       def to_solr
-        {
+        solr_hash = {
           "id": resource.id.to_s,
           "date_uploaded_dtsi": resource.created_at,
           "date_modified_dtsi": resource.updated_at,
@@ -49,6 +49,10 @@ module Hyrax
           "human_readable_type_tesim": resource.human_readable_type,
           "alternate_ids_sim": resource.alternate_ids.map(&:to_s)
         }.with_indifferent_access
+
+        solr_hash.merge!({ "schema_version_ssi": resource.schema_version }) if Hyrax.config.flexible?
+
+        solr_hash
       end
 
       ##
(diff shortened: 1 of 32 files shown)


app/indexers/hyrax/indexers/resource_indexer.rb
```diff
@@ -39,7 +39,7 @@ def initialize(resource:)
       # @api public
       # @return [HashWithIndifferentAccess<Symbol, Object>]
       def to_solr
-        {
+        solr_hash = {
```
laritakr, line 42: This shouldn't be necessary

kirkkwang: ah nice catch, that was from the old way we were indexing resource.id

app/views/hyrax/file_sets/_schema_version.html.erb
```diff
@@ -0,0 +1,35 @@
...
+<div class='pull-right'>
+  <% if f.object.schema_version.present? %>
+    <div class='schema-version text-center'>
+      <span>Created On Version</span>
+      <h4><%= f.object.schema_version.to_f %></h4>
+    </div>
+  <% end %>
+  <div class='schema-version text-center <%= updating %>'>
+    <span>Version After Save</span>
+    <h4><%= Hyrax::FlexibleSchema.last.id.to_f %></h4>
```
laritakr, line 11: I am adding a method to Hyrax::FlexibleSchema... method current_schema_id orders by created date and then does last.id on it. We should prob use that method.

kirkkwang: sounds good to me

laritakr (changes requested):
I'd like to hold off on merging so you can use the method I created on Hyrax::FlexibleSchema to find the most recent schema.

Other than that, if possible, I'd like to not repeat the _schema_version partial since it's the same in all three places (I think?). If the partials are required in that file location, could all three render the same partial from hyrax/base so the logic isn't in all 3 places? 

Should the styles be moved into normal css location instead of in the partial?

kirkkwang:
@laritakr good call on the shared partial, as for the css i think we should leave it in the partial because it's guarded with the `Hyrax.config.flexible?`.  This way it doesn't get loaded if we're not using flexible metadata

laritakr (approved):

PR #6860, opened 2024-07-02
Change && to if

kirkkwang:
In the case of a `GenericWorkResource` the `#hydra_model` method would still return `GenericWork` because the first condition evaluated to `true` instead.

### DIFF
--- a/app/models/concerns/hyrax/solr_document_behavior.rb
+++ b/app/models/concerns/hyrax/solr_document_behavior.rb
@@ -82,9 +82,9 @@ module Hyrax
 
     # Method to return the model
     def hydra_model(classifier: nil)
-      ((first('has_model_ssim')&.+ 'Resource')&.safe_constantize && Hyrax.config.valkyrie_transition) ||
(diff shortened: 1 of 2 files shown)


app/models/concerns/hyrax/solr_document_behavior.rb
```diff
@@ -82,7 +82,7 @@ def valkyrie?
 
     # Method to return the model
     def hydra_model(classifier: nil)
```
ShanaLMoore, line 84:
```suggestion
def hydra_model(classifier: nil)
  model = first('has_model_ssim')&.safe_constantize
  model = (first('has_model_ssim')&.+ 'Resource')&.safe_constantize if Hyrax.config.valkyrie_transition?
  model || model_classifier(classifier).classifier(self).best_model
end
```

kirkkwang: yes! this is a much cleaner way thanks!

dlpierce (approved):

PR #6862, opened 2024-07-10
Do not require custom cops in the engine

dlpierce:
This broke builds that do not install the rubocop gem. Loading of this custom cop is handled by rubocop.yml

### Fixes

Nurax failing to build due to rubocop gem not included.


kirkkwang:
Ah sorry about that, didn't think of that case!

kirkkwang (approved):

kirkkwang:
Oh right there are specs that looked for the dirty cops

dlpierce:
I forgot there was a spec for the cop.

kirkkwang (approved):

PR #6866, opened 2024-07-25
Corrects the sharing of predicates between date_created and created_at.

bwatson78:
### Summary

Corrects the sharing of predicates between `date_created` and `created_at`.

@samvera/hyrax-code-reviewers


orangewolf (approved):

orangewolf:
@bwatson78 code seems good there. My question to you is what kind of change is it? If you have existing records, does changing this predicate break things? Specifically if I have data in created at and then change the predicate, what happens the next time I open that record? Do we need a data migration of some sort here?

PR #6915, opened 2024-10-07
add contexts to the flexible metadata model.

orangewolf:
they can be assigned on admin sets and then apply the contexts to the works made with that admin set selected

### DIFF
--- a/app/forms/concerns/hyrax/flexible_form_behavior.rb
+++ b/app/forms/concerns/hyrax/flexible_form_behavior.rb
@@ -10,30 +10,18 @@ module Hyrax
 
     # OVERRIDE disposable 0.6.3 to make schema dynamic
     def schema
-      if Hyrax.config.flexible?
-        Hyrax::Forms::ResourceForm::Definition::Each.new(singleton_class.schema_definitions)
-      else
-        super
-      end
+      Hyrax::Forms::ResourceForm::Definition::Each.new(singleton_class.schema_definitions)
     end
 
     private
 
     # OVERRIDE valkyrie 3.0.1 to make schema dynamic
     def field(field_name)
-      if Hyrax.config.flexible?
-        singleton_class.schema_definitions.fetch(field_name.to_s)
-      else
-        super
-      end
+      singleton_class.schema_definitions.fetch(field_name.to_s)
     end
 
     def _form_field_definitions
-      if Hyrax.config.flexible?
-        singleton_class.schema_definitions
-      else
-        self.class.definitions
-      end
+      singleton_class.schema_definitions
     end
   end
 end
--- a/app/forms/concerns/hyrax/flexible_form_behavior.rb
+++ b/app/forms/concerns/hyrax/flexible_form_behavior.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+module Hyrax
+  module FlexibleFormBehavior
+    extend ActiveSupport::Concern
+
+    included do
+      include Hyrax::BasedNearFieldBehavior
+      property :contexts
+    end
+
+    # OVERRIDE disposable 0.6.3 to make schema dynamic
+    def schema
+      if Hyrax.config.flexible?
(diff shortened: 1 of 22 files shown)


lib/generators/hyrax/templates/db/migrate/20240606205216_add_contexts_to_hyrax_flexible_schemas.rb.erb
```diff
@@ -0,0 +1,5 @@
+class AddContextsToHyraxFlexibleSchemas < ActiveRecord::Migration<%= migration_version %>
+  def change
```
laritakr, line 2: Should this migration have a conditional check if it already exists?

app/models/hyrax/flexible_schema.rb
```diff
@@ -43,6 +50,10 @@ def schema_version
     profile['m3_version']
   end
 
+  def context_select
+    contexts.map { |k, v| [v&.[]('display_label'), k] }
```
laritakr, line 54:
```suggestion
// contexts&.map { |k, v| [v&.[]('display_label'), k] }
```

config/metadata_profiles/m3_profile.yaml
```diff
@@ -18,6 +18,8 @@ classes:
 contexts:
   flexible_context:
     display_label: Flexible Metadata Example
+  special_context:
```
laritakr, line 21: Corresponding changes to this profile should probably also be made in the .koppie (and possibly .dassie?) versions, as well as the spec version.

app/services/hyrax/m3_schema_loader.rb
```diff
@@ -21,11 +21,19 @@ def view_definitions_for(schema:, version: 1)
     ##
     # @param [#to_s] schema_name
     # @return [Enumerable<AttributeDefinition]
-    def definitions(schema_name, version)
+    def definitions(schema_name, version, contexts = nil)
       schema = Hyrax::FlexibleSchema.find_by(id: version) || Hyrax::FlexibleSchema.create_default_schema
       schema.attributes_for(schema_name).map do |name, config|
+        # We might be able to consolidate these conditions, but they have been kept seperate to make it easier to reason about
```
laritakr, line 27:
```suggestion
# We might be able to consolidate these conditions, but they have been kept separate to make it easier to reason about
```

laritakr (changes requested):
A few comments... 
- the M3 profiles in specs, .koppie, and .dassie do not match the changes in the config profile. These need to be updated
- no specs exist for the code changes here
- the migration needs to also be added to koppie & dassie
- Admin set show page should display the metadata context
- I was unable to save a work with the dimensions term

This is the error I got trying to save the work using the context term. I'm not certain of the cause, [but it originates on this line](https://github.com/samvera/valkyrie/blob/main/lib/valkyrie/persistence/postgres/persister.rb#L34C1-L34C55) when it calls `definitions` in the m3_schema_loader.

Screenshot 2024-10-08 at 5 40 09 PM

app/forms/concerns/hyrax/flexible_form_behavior.rb
```diff
@@ -0,0 +1,39 @@
...
+    extend ActiveSupport::Concern
+
+    included do
+      include Hyrax::BasedNearFieldBehavior
+      property :contexts
+    end
+
+    # OVERRIDE disposable 0.6.3 to make schema dynamic
+    def schema
+      if Hyrax.config.flexible?
```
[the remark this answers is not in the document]
orangewolf: correct, I'll take them out

lib/generators/hyrax/templates/db/migrate/20240606205216_add_contexts_to_hyrax_flexible_schemas.rb.erb
orangewolf: yep

PR #6927, opened 2024-10-17
Fix location indexing

laritakr:
### Fixes

Location indexer now includes the term needed for location faceting.

### Summary

Valkyrie resources were not indexing term `based_near_label_sim` which is needed for the location facet, so they were not appearing in the facet.

A new work using based_near location will now appear in the facet on the catalog page.

- `notes-bugfix` Bug Fixes

* Fix location facet indexing by indexing the needed term
* Add a safeguard that we don't attempt to call the location service on a blank term, as this crashes the service.

@samvera/hyrax-code-reviewers


randalldfloyd (approved):
+1 Good catch, thanks!

PR #6948, opened 2024-10-30
Parameterized lock requests

### LINKED ISSUE #6947: Global lock manager configuration lacks flexibility
### Descriptive summary

Currently `Lockable` implementations all share one set of configuration parameters. If a particular use case needs to vary these, currently it's necessary to create a separate `LockManager`.

- [ ] Clients of `Lockable` can pass in parameters to override the configured values just for a particular lock request
- [ ] Default behavior is unchanged

We discovered this in the course of adding locking to make sure Bulkrax uploads with multiple operations on the same object/collection/fileset didn't clobber each other. For this, we wanted short-lived locks, and we wanted jobs that couldn't acquire a lock to just fail quickly and be retried. Our current solution is to [create a separate lock manager for that job](https://gitlab.com/surfliner/surfliner/-/blob/5047743d923560f1e26d0122ad22355ef14563ce/comet/app/jobs/comet_create_relationships_job.rb#L224), with its own configuration, but it would be cleaner if we could just pass different parameters into this particular lock request.

### Related work

- [Surfliner MR#4458](https://gitlab.com/surfliner/surfliner/-/merge_requests/4458)

dmolesUC:
### Fixes

Fixes #6947.

### Summary

Allows callers of `LockManager#lock` and `Lockable#acquire_lock_for` to optionally pass in `ttl`, `retry_count`, and/or `retry_delay` parameters that will override the lock manager configuration for that particular lock request.

### Detailed Description

`LockManager#lock` now optionally takes `ttl`, `retry_count`, and `retry_delay` keyword arguments, which override the configured values for that lock request only, e.g.

```ruby
lock_manager.lock(some_key, ttl: 5000) do  # overrides default ttl of 60,000 ms
  some_operation
end
```

The same applies to `Lockable#acquire_lock_for`:

```ruby
class SomeTask
  include Hyrax::Lockable

  def perform
    acquire_lock_for(some_key, retry_delay: 30) do  # overrides default delay of 30 ms
      some_operation
    end
  end

  private

  def some_operation
    # ...
  end
end
```

The default behavior of `LockManager` and `Lockable` haven't changed.

### Notes

Currently, in this PR, `LockManager#lock` explicitly declares `ttl`, `retry_count`, and `retry_delay` keyword parameters. An alternative approach would be to just take `ttl` (which is passed to `Redlock::Client#lock`) and an arbitrary hash of Redlock client options (of which `retry_count` and `retry_delay` are just two possible values). I decided not to go that direction because:

1. these are the only parameters we thought were worth configuring
2. there's been some discussion of getting off Redlock (and Redis), and I didn't want to tie us too closely to that API

I did, however, redefine `Lockable#acquire_lock_for` to be a simple delegate to `lock_manager.lock` -- that's how it was behaving anyway, and it seemed cleaner than re-declaring the parameters there. I've implemented it with `Forwardable`, but if we think that's too cute, we could instead do something like:

```ruby
def acquire_lock_for(key, **kwargs, &block)
  lock_manager.lock(key, **kwargs, &block)
end
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/lock_manager.rb
+++ b/app/services/hyrax/lock_manager.rb
@@ -16,11 +16,11 @@ module Hyrax
 
     ##
     # Blocks until lock is acquired or timeout.
-    def lock(key)
+    def lock(key, ttl: @ttl, retry_count: @retry_count, retry_delay: @retry_delay)
       returned_from_block = nil
 
       pool.then do |conn|
-        client(conn).lock(key, @ttl) do |locked|
+        client(conn, retry_count: retry_count, retry_delay: retry_delay).lock(key, ttl) do |locked|
           raise UnableToAcquireLockError unless locked
           returned_from_block = yield
         end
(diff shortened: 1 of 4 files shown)


dlpierce (approved):
Looks good, thank you!

PR #6952, opened 2024-10-31
chunk upload bug fix

ShanaLMoore:
Issues have been spotted after deploying the chunked upload work. 

Sometimes the UV won't load the file and the downloaded file is corrupt. It is because it's trying to present the last chunk instead of the fully reassembled file.

### DIFF
--- a/app/controllers/hyrax/uploads_controller.rb
+++ b/app/controllers/hyrax/uploads_controller.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 module Hyrax
   class UploadsController < ApplicationController
     load_and_authorize_resource class: Hyrax::UploadedFile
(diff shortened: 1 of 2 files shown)


orangewolf (approved):
Good work Shana!

dlpierce:
I'll mention here that we ran into a flaky feature spec that stemmed from the chunked upload changes in #6931. It's fixed in #6950. 
That spec uploads a file but didn't do anything with it afterwards, only checking a couple other things on the page. My theory is the async upload JS would start, but not finish before the spec finished and started cleaning up the DB, causing the User or UploadedFile lookups to fail. Something to lookout for in other specs.

ShanaLMoore:
Ohh, thanks so much for catching and fixing it @dlpierce

PR #6976, opened 2024-11-22
Green sirenia or bust!

### LINKED ISSUE #6894: Sirenia should run in Github actions and pass
### Descriptive summary

We want to make sure that all specs pass in Github Actions for Sirenia.

dlpierce:
Fixes #6894 

### Summary
Get Sirenia specs passing and add to test matrix

### Detailed Description
Many fixes for specs broken in sirenia (fedora 6/valkyrie). See individual commmit messages.

Docker compose files have been reworked to use the same image. 

The custom queries added in support of valkyrie statistics support have been converted from SQL to Solr.


app/services/hyrax/custom_queries/find_by_date_range.rb
```diff
@@ -26,29 +26,10 @@ def initialize(query_service:)
...
-      def find_by_date_range_query
-        <<-SQL
-          SELECT * FROM orm_resources
-          WHERE created_at >= ?
-          AND created_at <= ?;
-        SQL
+        end_range = end_datetime.blank? ? '*' : end_datetime.utc.xmlschema
+        args = "system_create_dtsi:[#{start_datetime.utc.xmlschema} TO #{end_range}]"
+        args += " AND has_model_ssim: (#{models.map { |m| "\"#{m}\"" }.join(' OR ')})" unless models.empty?
+        Hyrax::SolrService.query(args)
```
orangewolf, line 32: does this change the flavor of thing returned? instead of `models [Array]` is it now `SolrHit [Array]`?

bin/worker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -e
+
+# Wait for web app to avoid racing during bundle install
+db-wait.sh web:3000
```
orangewolf, line 5: might be time to rename db-wait.sh

docker-compose-koppie.yml
```diff
@@ -33,82 +32,67 @@ services:
       - 1049:1048
     volumes:
       - ./bin:/app/samvera
-      - .koppie:/app/samvera/hyrax-webapp
+      - .koppie:/app/samvera/hyrax-koppie
```
orangewolf, line 35: can I ask what we accomplish by making this different? its extra mental overhead to remember which is which unless it serves some purpose.

lib/hyrax/transactions/steps/delete_resource.rb
```diff
@@ -27,8 +27,8 @@ def initialize(persister: Hyrax.persister, publisher: Hyrax.publisher)
         def call(resource, user: nil)
           return Failure(:resource_not_persisted) unless resource.persisted?
 
-          @persister.delete(resource: resource)
           publish_changes(resource: resource, user: user)
+          @persister.delete(resource: resource)
```
orangewolf, line 31: why? isn't it better to do the thing then publish that you did it?

spec/javascripts/jasmine_spec.rb
```diff
@@ -4,6 +4,10 @@
 # Run the jasmine tests by running the karma javascript test framework
 # The spec will fail if any jasmine tests fails.
 RSpec.describe "Jasmine" do
+  before do
+    `cd $RAILS_ROOT && bundle exec rake assets:precompile` if ENV['RAILS_ROOT']
```
orangewolf, line 8: this is really expensive to do on every spec run... is there anything we can do to mitigate that cost?

orangewolf (commented):
This looks great over all, I have a few questions but nothing that needs to stop the show

docker-compose-koppie.yml
dlpierce: This was done to better support having both .dassie and .koppie available in the hyrax-engine-dev container image, but in the end it might not be necessary. My thinking was it is easier to know which test app you're in by the directory name, but you make a good point too. I'll look into undoing the hyrax-koppie dir.

lib/hyrax/transactions/steps/delete_resource.rb
dlpierce: I agree, but fedora was complaining about tombstones. I'll double check this is still needed since there was another tombstone related fix in this PR.

spec/javascripts/jasmine_spec.rb
dlpierce: I can't find a way to make the JS tests work without compiling. This used to occur during the image build, so it's not a new time sink.

lib/hyrax/transactions/steps/delete_resource.rb
dlpierce: Still needed unless an alternative is found. Without the backwards order `find_parents` in `MemberCleanupListener` fails during `rspec spec/hyrax/transactions/work_destroy_spec.rb`:

```ruby
     Ldp::Gone:
       Discovered tombstone resource at /test/a1/38/5d/b3/a1385db3-8fbc-4330-b57d-7ffb972a5591, departed at: 2024-12-18T16:22:50.848357Z
     # /app/bundle/ruby/3.2.0/gems/ldp-1.2.1/lib/ldp/client/methods.rb:119:in `block in check_for_errors'
     # /app/bundle/ruby/3.2.0/gems/ldp-1.2.1/lib/ldp/client/methods.rb:117:in `check_for_errors'
     # /app/bundle/ruby/3.2.0/gems/ldp-1.2.1/lib/ldp/client/methods.rb:53:in `block in get'
     # /app/bundle/ruby/3.2.0/gems/activesupport-6.1.7.10/lib/active_support/notifications.rb:205:in `instrument'
     # /app/bundle/ruby/3.2.0/gems/ldp-1.2.1/lib/ldp.rb:43:in `instrument'
     # /app/bundle/ruby/3.2.0/gems/ldp-1.2.1/lib/ldp/client/methods.rb:33:in `get'
     # /app/bundle/ruby/3.2.0/gems/valkyrie-3.5.0/lib/valkyrie/persistence/fedora/query_service.rb:108:in `content_with_inbound'
     # /app/bundle/ruby/3.2.0/gems/valkyrie-3.5.0/lib/valkyrie/persistence/fedora/query_service.rb:42:in `find_parents'
     # ./app/services/hyrax/listeners/member_cleanup_listener.rb:19:in `on_object_deleted'
     # /app/bundle/ruby/3.2.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:46:in `call'
     # /app/bundle/ruby/3.2.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:46:in `block in publish'
     # /app/bundle/ruby/3.2.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:38:in `block in process'
     # /app/bundle/ruby/3.2.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:34:in `each'
     # /app/bundle/ruby/3.2.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:34:in `process'
     # /app/bundle/ruby/3.2.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:45:in `publish'
     # /app/bundle/ruby/3.2.0/gems/dry-events-1.0.1/lib/dry/events/publisher.rb:207:in `publish'
     # ./lib/hyrax/transactions/steps/delete_resource.rb:45:in `publish_changes'
```

app/services/hyrax/custom_queries/find_by_date_range.rb
dlpierce: Yeah, this should be changed to return a model.

dlpierce: I'm concerned performance of these statistics related custom queries is going to be terrible on production sized data sets.

orangewolf (approved):
This looks great. Thank you for the revisions. I agree the performance could be an issue on the date queries, but "we aren't doing performance on these" is clearly stated and should be solved with a production load in place.

PR #7003, opened 2025-02-03
Changing hydra-editor to 7.0 for Rails 7.2 compatibility

randalldfloyd:
We had been testing the Ruby/Rails upgrade work by pinning to an unmerged branch of `hydra-editor`.  That got merged so now we can properly spec a version that is compatible with Rails 7.2.


cjcolvar (approved):
:+1:

PR #7006, opened 2025-02-06
Prevent loss of attributes with reschedule

laritakr:
## Fixes

This commit changes to use a temporary variable to hold the job attributes, so that when CreateWorkJob is rescheduled due to any errors, the attributes that were deleted do not get left out of the resubmitted job's attributes.

@samvera/hyrax-code-reviewers


randalldfloyd:
We're going to roundtrip this one back to the top of `main` after a new v5.0.x release, which will contain this same fix merged in #7004.

PR #7015, opened 2025-02-11
Check if db is active?

dlpierce:
### Fixes

Failed attempts to access the database during initialization.

### Summary

This code is designed to prevent errors during initialization when the database connection is not ready. The postgres adapter does not raise an error as expected when accessed, causing errors in situations such as application generation. Calling `active?` should give us a better idea if the db is really ready.

* Can generate a hyrax application without the database being up. Other commands such as rake tasks may also benefit.
*
*

* Rely on `active?` to determine db status.
*
*

@samvera/hyrax-code-reviewers


cjcolvar (approved):
Looks good as far as I can tell.

PR #7016, opened 2025-02-11
Overriding Rails config defaults for compatibility

### LINKED ISSUE #7011: Hyrax application does not load with Rails 7.2 defaults
Setting `config.load_defaults 7.2` prevents dassie/koppie from loading with
`'require': cannot load such file -- hyrax/callbacks (LoadError)`
This is also seen when generating a new application from the template in Rails 7.2.

randalldfloyd:
Fixes #7011 

In #6913 to get running on Rails 7.2, I missed that the test apps were still set to load configuration defaults for 6.0.  When the target version gets bumped to 7.2 in any scenario, such as generating a new Rails app from the template, things start breaking.

This PR sets the test apps to load configuration defaults for 7.2, but adds several overrides to pre-7.x values for compatibility.  

Where possible, overrides were put in a Hyrax-managed initializer to keep from having to persist options in the actual application.   However, `add_autoload_paths_to_load_path` has to be in the `application.rb` of the application so that it takes effect before code starts to load.  That is accomplished by an insert in the install template.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
+
+insert_into_file 'config/application.rb', after: /config\.load_defaults [0-9.]+$/ do
+  "\n    config.add_autoload_paths_to_load_path = true"
+end
+
 gem 'hyrax', '5.0.3'
 run 'bundle install'
 generate 'hyrax:install', '-f'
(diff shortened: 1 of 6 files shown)


template.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
+
+insert_into_file 'config/application.rb', after: /config\.load_defaults [0-9.]+$/ do
+  "\n    config.add_autoload_paths_to_load_path = true"
+end
+
```
dlpierce, line 6: This autoload config needs to be setup here before attempting to load the hyrax gem and install generator.

PR #7045, opened 2025-03-26
Update location of hyrax-base image

dlpierce:
### Fixes

Outdated location of the hyrax-base image in containers doc.


randalldfloyd (commented):
@dlpierce 
I may be misunderstanding how this would work, but without a tag, would this result in trying to pull `latest`?  I wanted to verify that before approving, but I can't find a tag by that name:
```
docker manifest inspect ghcr.io/samvera/hyrax/hyrax-base
manifest unknown

docker manifest inspect ghcr.io/samvera/hyrax/hyrax-base:latest
manifest unknown
```

orangewolf (approved):

PR #7049, opened 2025-03-28
🚧 Make ControlledVocabularyInput Compatible with Valkyrie

ShanaLMoore:
# Problem
The ControlledVocabularyInput class was designed for ActiveFedora's static, RDF-based models. When used with the flexible metadata implementation, it fails because:
Fields are dynamically loaded at runtime in flexible metadata mode
The prepopulator method may not run if the field isn't available when expected
The ResourceForm reconstruction process can remove fields that the input expects

# Solution
This PR updates the ControlledVocabularyInput class to detect Valkyrie/flexible forms and use a simplified approach:
For Valkyrie/flexible forms: Generate a plain input with the necessary data attributes
For ActiveFedora forms: Preserve the original behavior
This approach makes no assumptions about how fields are defined or accessed in flexible metadata mode, allowing both implementations to work with the same view partial.

# Testing
The changes have been tested by confirming that both specs in spec/views/collections/edit_fields/_based_near.html.erb_spec.rb and spec/views/records/edit_fields/_based_near.html.erb_spec.rb now pass:
The ActiveFedora context still works as before
The Valkyrie context now correctly includes the necessary attributes
# Impact
This change allows applications to use the same view partial (_based_near.html.erb) for both ActiveFedora and Valkyrie resources without having to maintain conditional rendering logic in the view template. It supports the transition to Valkyrie while maintaining backward compatibility with existing ActiveFedora implementations.

If merged, remove the skip that was added here: https://github.com/samvera/hyrax/pull/6843/commits/de8b98a754463bc59981fe2f0d08485c7201ebcc

### DIFF
--- a/app/inputs/controlled_vocabulary_input.rb
+++ b/app/inputs/controlled_vocabulary_input.rb
@@ -7,7 +7,64 @@ class ControlledVocabularyInput < MultiValueInput
   #   'multi_value'.freeze
   # end
 
+  def input(wrapper_options = nil)
+    # For Valkyrie forms, use a special approach
+    if valkyrie_form?
+      # For Valkyrie, generate HTML that works with the JS
+      template = <<-HTML
+        <div class="multi_value controlled_vocabulary" data-autocomplete-url="#{input_html_options[:data].try(:[], :'autocomplete-url')}" data-field-name="#{attribute_name}">
+          <ul class="listing">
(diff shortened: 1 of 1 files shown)


laritakr:
lint run: https://github.com/samvera/hyrax/actions/runs/14158761193

PR #7050, opened 2025-03-28
Adds more options into attributes_helper

laritakr:
Finding the attributes depends on which class we have. This adds more fallback methods to find the right class & make sure we can display the works.


laritakr:
This does not seem to have broken any additional specs. https://github.com/samvera/hyrax/actions/runs/14138599675

PR #7053, opened 2025-04-06
Queueable index adapter

orangewolf:
### Summary

Ability to queue index / delete calls in to Redis to process them via Solr more effectively. This method accomplishes two things. 1) all ids are unique in the Redis set, so adding the same item multiple times will only index it once and 2) should give us Solr performance by committing more items to the index before a commit. 

### TODO
- [x] Add specs
- [x] Rubocop fixes 
- [x] Add a requeing job that runs the index_queue method every 5 minutes so things actually get indexed
- [x] Add a requeing job that runs the delete_queue method every 5 minutes so things actually get deleted

* Set indexing queue in an initializer: `Hyrax.config.index_adapter = :redis_queue`
* Add or edit or delete items, checking the queue with `Hyrax.config.redis_connection.smembers('toindex')` and `Hyrax.config.redis_connection.smembers('todelete')` to see that items are added to the queue
* Verify that items do get indexed properly when the indexing job runs
* Verify that items do get deleted properly when the delete job runs

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/concerns/hyrax/queued_job_behavior.rb
+++ b/app/jobs/concerns/hyrax/queued_job_behavior.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+module Hyrax
+  # Grants the user's edit access on the provided FileSet
+  module QueuedJobBehavior
+    extend ActiveSupport::Concern
+
+    included do
+      queue_as Hyrax.config.ingest_queue_name
+    end
+
+    private
+
+    def redis_queue
+      Valkyrie::IndexingAdapter.find(:redis_queue)
+    end
+
+    def requeue(*args)
+      self.class.set(wait_until: 5.minutes.from_now).perform_later(*args)
+    end
+  end
+end
--- a/app/jobs/concerns/hyrax/queued_job_behavior.rb
+++ b/app/jobs/concerns/hyrax/queued_job_behavior.rb
@@ -6,6 +6,7 @@ module Hyrax
 
     included do
       queue_as Hyrax.config.ingest_queue_name
+      cattr_accessor :requeue_frequency
     end
 
     private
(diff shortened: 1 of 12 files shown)


dlpierce:
This is a great idea!

I see this uses sets for the redis queue datatype so the queued ids are not repeated, but `SPOP` will return a random subset. If we instead used a sorted set using a timestamp as the score that would allow the index job to work on the oldest queued records first, and give the queued records a bigger window to receive updates before being indexed. Another possibility is to exclude timestamps that are less than X seconds old, although that might not be possible with only one redis call.

app/jobs/concerns/hyrax/queued_job_behavior.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    private
+
+    def redis_queue
+      Valkyrie::IndexingAdapter.find(:redis_queue)
+    end
+
+    def requeue(*args)
+      self.class.set(wait_until: 5.minutes.from_now).perform_later(*args)
+    end
```
dlpierce, line 19: The number of minutes (or seconds) to wait ought to be configurable. Or perhaps dynamic depending on the queue size?

orangewolf:
> This is a great idea!
> 
> I see this uses sets for the redis queue datatype so the queued ids are not repeated, but `SPOP` will return a random subset. If we instead used a sorted set using a timestamp as the score that would allow the index job to work on the oldest queued records first, and give the queued records a bigger window to receive updates before being indexed. Another possibility is to exclude timestamps that are less than X seconds old, although that might not be possible with only one redis call.

Good call. I didn't realize `spop` was random. Switched it to `zadd` and `zpopmin` to fix this issue. Seems to work well.

app/jobs/concerns/hyrax/queued_job_behavior.rb
orangewolf: configurable seems reasonable, dynamic probably isn't needed.

orangewolf: good idea. I've added it

PR #7054, opened 2025-04-07
Catch errors that cause unnecessary retries

laritakr:
### Fixes

In order to speed up migrations, we want to skip out of jobs which error that don't need 5 retries.

### DIFF
--- a/app/jobs/migrate_files_to_valkyrie_job.rb
+++ b/app/jobs/migrate_files_to_valkyrie_job.rb
@@ -4,7 +4,7 @@
 class MigrateFilesToValkyrieJob < Hyrax::ApplicationJob
   # Define a logger for this job
   def logger
-    @logger ||= Logger.new(Rails.root.join('log', 'migrate_files_to_valkyrie_job.log'))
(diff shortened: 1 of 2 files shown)


app/jobs/migrate_files_to_valkyrie_job.rb
```diff
@@ -2,6 +2,10 @@
 # Responsible for conditionally enqueuing the file and thumbnail migration
 # logic of an ActiveFedora object.
 class MigrateFilesToValkyrieJob < Hyrax::ApplicationJob
+  # Define a logger for this job
+  def logger
+    @logger ||= Logger.new(Rails.root.join('log', 'migrate_files_to_valkyrie_job.log'))
```
orangewolf, line 7:
```suggestion
    @logger ||= Logger.new(Rails.root.join('tmp', 'imports', 'migrate_files_to_valkyrie_job.log'))
```

this prevents the log from being deleted when the pod ends

orangewolf (changes requested):
concept is right, just need to change the file location

PR #7058, opened 2025-04-10
Change base image to debian

dlpierce:
Moving from Alpine to Debian will avoid lengthy native compilation when installing the grpc gem (which lacks appropriate prebuilt file for Alpine). This work is based on similar changes done for hyku by Rob.

Debian does not yet package ImageMagick 7, so we install it here by extracting files from the official AppImage archive.

Also included are provisions for running the nurax test servers from the hyrax-engine-dev target.

@samvera/hyrax-code-reviewers


randalldfloyd (approved):
Verified that I'm able to build/up the test apps locally with this setup

PR #7060, opened 2025-04-11
🐛 Creates a fallback for m3 flexible schema

sjproctor:
### Fixes

Ref:
- https://github.com/notch8/wvu_knapsack/issues/22

### Summary

Allows for introduction of m3 flexible metadata profiles by creating a default resource. 

- `notes-bugfix` Bug Fixes

### Detailed Description

Prior to this PR, adding a m3 flexible metadata profile resource types that not include all default resource types produced a no method error. This change removes the error message in favor of a default resource that allows loading of a custom schema.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/m3_schema_loader.rb
+++ b/app/services/hyrax/m3_schema_loader.rb
@@ -23,7 +23,9 @@ module Hyrax
     # @return [Enumerable<AttributeDefinition]
     def definitions(schema_name, version, contexts = nil)
       schema = Hyrax::FlexibleSchema.find_by(id: version) || Hyrax::FlexibleSchema.create_default_schema
-      schema.attributes_for(schema_name).map do |name, config|
+      attributes = schema.attributes_for(schema_name)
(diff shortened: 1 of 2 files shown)


app/services/hyrax/m3_schema_loader.rb
```diff
@@ -37,8 +39,37 @@ def definitions(schema_name, version, contexts = nil)
     rescue ActiveRecord::StatementInvalid
       Rails.logger.error "Skipping definition load for migrations to run"
       []
-    rescue NoMethodError
-      raise UndefinedSchemaError, "Flexible schema not found in version #{version} for #{schema_name}"
     end
+
+    # rubocop:disable Metrics/MethodLength
+    def fall_back_schema_for(_schema_name)
```
kirkkwang, line 45: very non-blocking, but consider the name to be `fallback_schema_for` instead?

kirkkwang (approved):

PR #7074, opened 2025-04-22
Use HTTP POST for Matomo 5 API

### LINKED ISSUE #6769: Matomo 5 API calls should use POST instead of GET
### Descriptive summary

Matomo 5 (released Dec 2023) changes how auth tokens work when using the reporting API ([see notes](https://developer.matomo.org/changelog#usage-of-authentication-tokens)). Hyrax currently ([in Hyrax v5.0.1](https://github.com/samvera/hyrax/blob/v5.0.1/app/services/hyrax/analytics/matomo.rb#L172)) issues `GET` requests to the reporting API that include a confidential auth token, and this should be changed to `POST`.

See the following example API call issued via a Hyrax application to a Matomo instance (w/token & IP obscured):
```
xxx.x.xxx.xx - - [01/Apr/2024:14:43:56 +0000] "GET /?date=2023-03-31%2C2024-04-01&flat=1&format=JSON&idSite=6&label=7m01bk723+-+file-set-in-work-download&method=Events.getName&module=API&period=day&token_auth=xxx HTTP/1.1" 200 5889
```

- [ ] Calls to the Matomo reporting API are made via `POST` and not `GET`

This change is recommended by Matomo for improved security. Anyone setting up a new auth token in Matomo 5 (e.g., to support Matomo analytics in a Hyrax app) will by default have it restricted to `POST`, which would make it not work with Hyrax.

ryanfb:
### Fixes

Fixes #6769

### Summary

> Matomo 5 (released Dec 2023) changes how auth tokens work when using the reporting API: https://developer.matomo.org/changelog#usage-of-authentication-tokens
>
> Anyone setting up a new auth token in Matomo 5 (e.g., to support Matomo analytics in a Hyrax app) will by default have it restricted to POST, which would make it not work with Hyrax.

* Fresh Matomo 5 API tokens with POST restriction should now work

- `notes-bugfix` HTTP POST now used for Matomo API calls

* switch from GET to POST for all Matomo API calls

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/analytics/matomo.rb
+++ b/app/services/hyrax/analytics/matomo.rb
@@ -179,14 +179,6 @@ module Hyrax
           response.is_a?(Hash) && response["result"] == "error"
         end
 
-        def get(params)
-          response = Faraday.get(config.base_url, params)
-          return [] if response.status != 200
-          api_response = JSON.parse(response.body)
-          return [] if contains_matomo_error?(api_response)
-          api_response
-        end
-
         def post(params)
           response = Faraday.post(config.base_url, params)
           return [] if response.status != 200


ryanfb:
NB: this is likely to start affecting all Hyrax users who use Matomo for statistics. We ran across this when an expired Matomo key for our Hyrax app stopped view/download statistics from displaying correctly, and the re-issued "secure" auth token being restricted to POST meant it needed this corresponding fix.

PR #7075, opened 2025-04-23
Compile imagemagick on arm64

dlpierce:
### Fixes

Fixes unable to build Dockerfile on arm64

### Detailed Description
The upstream AppImage version of ImageMagick does not provide arm64 binaries, so we must compile and install our own from source. Hopefully this is a short lived situation that will be resolved when the next version of Debian releases with ImageMagick 7 packages.

* During docker build, compile ImageMagick when a precompiled binary is not available.
*
*

@samvera/hyrax-code-reviewers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,6 @@ RUN apt-get update && \
     ffmpeg \
     ghostscript \
     git \
-    imagemagick \
     less \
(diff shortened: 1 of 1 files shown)


bkiahstroud (approved):
It works on my machine

PR #7080, opened 2025-05-02
[5.0 forward port] Redis Index Queue 

orangewolf:
### Fixes

This brings the redis queue work in the 5.0 branch forward to main.

@samvera/hyrax-code-reviewers


randalldfloyd (approved):

dlpierce:
How are the jobs initially enqueued? Could we do it automatically at startup (if enabled) and avoid having duplicates using https://github.com/veeqo/activejob-uniqueness ?

dlpierce (approved):
Let's get this in and worry about the kickstarter in another PR.

PR #7126, opened 2025-06-13
Flexible lay ground work

orangewolf:
### Summary

Lays the groundwork for Flexible metadata by making including core and basic metadata by default completely optional and toggled by config variables. This version allows configuring the Collection, AdministrativeSet, FileSet and each individual work type independently. 

It depreciates two existing functions in a way that if you are using them you won't have to make changes right away. 

The first is that Hyrax includes core_metadata in all work types. This can still be done in the future but should be explicit in each work types class and the generators have been updated to support that.

The second is the removal of the auto inclusion of `"#{SchemaName}FormFieldsBehavior"` when adding `schema_name` schemas to the form objects. This is difficult to trace as there is no explicit in include so its hard to know where this code comes from when one is debugging. It also doesn't save much in terms of LOC to have it explicit. In current Hyrax this is only used by based near. based_near form attributes have been moved to be generally included but to noop if the object doesn't have a based near field instead.

This is step 1 or arguably step 0 of the flexible metadata work. My plan is to chain additional PRs off this one so that they can be reviewed and merged in order only.

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/collection_type_spec.rb
+++ b/spec/features/collection_type_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe 'collection_type', type: :feature do
     end
   end
 
-  describe 'create collection type' do
+  describe 'create collection type', :clean_repo do
     let(:title) { 'Test Type' }
     let(:description) { 'Description for collection type we are testing.' }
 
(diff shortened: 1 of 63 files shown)


spec/features/collection_type_spec.rb
```diff
@@ -42,7 +42,7 @@
     end
   end
 
-  describe 'create collection type' do
+  describe 'create collection type', :clean_repo do
```
orangewolf, line 45: fixed a flapping spec that was making me crazy ;-)

christhepianist:
[celebrate]     Christine Peterson reacted to your message:
________________________________

randalldfloyd:
Also, I will probably do some strategic commit editing before merging.  I don't want to squash the whole thing.

PR #7128, opened 2025-06-13
Fix notification bell aria issues.

### LINKED ISSUE #7122: Visible label and accessible name do not match
## Descriptive summary
On dev.nurax.samvera.org (Hyrax 5.1.0), the visible label and accessible name do not match and do not meet WCAG 2.2 accessibility requirements

## Steps to reproduce the behavior in User Interface (UI)

Pre-req: You must have at least one unread notification. The easiest way to get an unread notification is to add another user as a proxy.

1. In Chrome, download the Site Improve browser extension
2. Log into dev.nurax.samvera.org as a user with at least one unread notification
3. Run the SiteImprove browser extension. 

Image

dlpierce:
### Fixes

Fixes #7122 

### Summary

aria-label did not match the displayed values. Moving that verbose string to aria-description and adding a label to the bell icon should resolve the issue and be more succinct.

The view spec is refactored to allow test messages to render, and some additional auth related specs are added. The commented lines depend on PR #7127 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/hyrax_helper_behavior.rb
+++ b/app/helpers/hyrax/hyrax_helper_behavior.rb
@@ -68,10 +68,10 @@ module Hyrax
       mailbox = UserMailbox.new(user)
       unread_notifications = mailbox.unread_count
       link_to(hyrax.notifications_path,
-              'aria-label' => mailbox.label(params[:locale]),
+              'aria-description' => mailbox.label(params[:locale]),
               class: 'notify-number nav-link') do
         capture do
-          concat tag.span('', class: 'fa fa-bell')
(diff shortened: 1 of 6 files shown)


dlpierce:
Apologies for the off-topic commits here. This PR hit several inconsistent spec failures along the way that I tried to find solutions for.

In particular, it seems like DatabaseCleaner does not consistently clear records, leading to duplicate Sipity::Role objects. A theory for this behavior is bad interactions between multiple/threaded database connections and the transaction rollback used by DatabaseCleaner. I've added explicit clearing of connections after cleaning, but it remains to be seen if this is effective as reproducing the spec failures locally has not been possible.

PR #7135, opened 2025-07-08
🔄 Revert changes to select type of work label

sjproctor:
### Fixes

Ref:
- https://github.com/notch8/hykuup_knapsack/issues/383

### Summary

This commit reverts the changes made to the label for selecting the type a work type.

### Detailed Description

This bug was creating by incorrectly fixing an issue that arose from adding flexible metadata profiles that had `multi_value: false` on the title property.

Flexible metadata profiles must have `multi_value: true` on the title.

@samvera/hyrax-code-reviewers


kirkkwang (approved):
Looks good to me!

Just noting that if you want to ensure `#first` does what you expect you can always do something like this as an option.  But I do recognize that's being overly cautious.

```rb
Array.wrap(source.title).first
```

PR #7176, opened 2025-08-29
Add to collection UI improvement

kirkkwang:
This commit will favor using the select2 with the drop down with a pre- populated list of collections over the current implementation where it makes a query to find the collections.  Prior, the user would need to know what the collection's name is before using this interface.  Now the user has the option of scrolling through and also using the type function to find the collection for deposit.

image

image

image

image


ShanaLMoore:
Issue: 
- https://github.com/notch8/hykuup_knapsack/issues/466

PR #7179, opened 2025-09-03
Workflow factory loading fix

### LINKED ISSUE #7166: A copy of Hyrax::Workflow::WorkflowFactory has been removed from the module tree but is still active
While working on updating our Hyrax-based repository MorphoSource to Hyrax 5.0.5, in our development environment we are intermittently getting the error "A copy of Hyrax::Workflow::WorkflowFactory has been removed from the module tree but is still active." In at least one case, we've tracked this down to be caused by this line in `Hyrax::Workflow::WorkflowFactory`:

https://github.com/samvera/hyrax/blob/ca7efb6f85eb3475ea5880f361ac2aeb84f6e137/app/services/hyrax/workflow/workflow_factory.rb#L66

Changing `WorkflowActionInfo` to `Hyrax::WorkflowActionInfo` in this line fixes the issue for us, though that requires creating a downstream copy of this module in our app just to update the module reference, which isn't ideal. 

Thanks!

randalldfloyd:
Issue loading workflow factory seems to be fixed if prefixed with namespace

### Fixes

Fixes #7166

### DIFF
--- a/app/services/hyrax/workflow/workflow_factory.rb
+++ b/app/services/hyrax/workflow/workflow_factory.rb
@@ -63,7 +63,7 @@ module Hyrax
       end
 
       def run_workflow_action!(action:)
-        subject = WorkflowActionInfo.new(work, user)
+        subject = Hyrax::WorkflowActionInfo.new(work, user)
         Workflow::WorkflowActionService.run(subject: subject,
                                             action: action)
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
👍 Makes sense.

Issue #7185, opened 2025-09-15
:bug: Batch add and valkyrized works are incompatible


ShanaLMoore:
# Summary

Refactor **Batch Add (Create)** to be fully **Valkyrie-based** and generate Step 2 (common metadata) dynamically from the **selected work type**.
For **Bulk Change (Edit)**, **remove metadata editing** and support **permissions and visibility only**.

# Acceptance Criteria
GIVEN an authenticated user has permission to create works
AND the Batch Upload feature is enabled for the tenant in their features settings
WHEN the user 

# Implementation notes

* [ ] **Batch Add (Create)**

  * [ ] Type selection is **required** before Step 2.
  * [ ] **Step 1:** upload files + per-file **Title** and **Resource Type** (options reflect selected work type).
  * [ ] **Step 2:** fields come from the selected work type’s **Valkyrie form/changeset** (or profile):

    * [ ] Only valid terms for that type are shown.
    * [ ] **Required fields** are indicated and enforced; server-side validation blocks submission with inline errors.
  * [ ] **Batch create** makes one work per file, attaches files, applies Step-1 (Title/Resource Type) + Step-2 common metadata.
  * [ ] No reliance on the **fake ActiveFedora** form; compatible with Hyrax **6.1/6.2** (covered by tests).
  * [ ] Error states: missing required fields block submission; no phantom validations for fields not in schema.

* [ ] **Bulk Change (Edit) — Permissions & Visibility only**

  * [ ] UI shows **Permissions** and **Visibility** controls only; **all metadata fields are removed/hidden**.
  * [ ] Server accepts and applies only:

    * [ ] **Permissions** (sharing/ACLs as currently supported).
    * [ ] **Visibility** (`open`, `authenticated`, `restricted`).
      *(Embargo/lease management is out of scope for this ticket.)*
  * [ ] Mixed work-type selection is allowed (no type gating needed).
  * [ ] Submitting any **metadata params** via the request is ignored and **not** applied (test coverage included).
  * [ ] Success and error messaging reflect the limited scope (“Bulk Edit updates permissions and visibility only”).

# Screenshots or Video

<details>
<summary>What to capture</summary>

* **Batch Add**:

  * Type picker gating “Next” until a type is selected.
  * Step 2 for two distinct work types showing different fields.
  * Inline validation on a required field.
  * Success summary: N works created with files attached.

* **Bulk Change (Edit)**:

  * Updated panel showing **Permissions & Visibility** only.
[...]

PR #7186, opened 2025-09-15
Warn batch upload is not compatible with flex

ShanaLMoore:
## Warn batch upload is not compatible with flex

0b372eea21e5243347a2b3938215ddb7062fee6d

Batch upload and valkyrized works are incompatible. We will add help text to warn the user of this case: https://github.com/samvera/hyrax/issues/7185

image

Related issues: 
- https://github.com/samvera/hyrax/issues/7185
- https://github.com/notch8/palni_palci_knapsack/issues/473

### DIFF
--- a/config/features.rb
+++ b/config/features.rb
@@ -30,7 +30,7 @@ Flipflop.configure do
 
   feature :batch_upload,
           default: false,
-          description: "Enable uploading batches of works. <br /><strong>WARNING:</strong> This feature is not compatible with flexible metadata.".html_safe
+          description: "Enable uploading batches of works. <br /><strong>WARNING:</strong> Do not enable. This feature is broken and will be addressed in https://github.com/samvera/hyrax/issues/7185.".html_safe
 
   feature :hide_private_items,
           default: false,


ShanaLMoore:
@orangewolf also batch upload doesn't work with 6.2 at all, regardless of flexible on/off

orangewolf (approved):

config/features.rb
```diff
@@ -30,7 +30,7 @@
 
   feature :batch_upload,
           default: false,
-          description: "Enable uploading batches of works"
+          description: "Enable uploading batches of works. <br /><strong>WARNING:</strong> Do not enable. This feature is broken and will be addressed in https://github.com/samvera/hyrax/issues/7185.".html_safe
```
ShanaLMoore, line 33: I already merged this but should we wrap the url in <a>? so it's linked? It looks like this deployed as is: 

Image

ShanaLMoore: I updated it to 

```
  feature :batch_upload,
          default: false,
          description: "Enable uploading batches of works. <br /><strong>WARNING:</strong> Do not enable. This feature is broken and will be addressed in <a href='https://github.com/samvera/hyrax/issues/7185' target='_blank'>https://github.com/samvera/hyrax/issues/7185</a>.".html_safe
```

PR #7195, opened 2025-09-18
Upgrade base image to debian trixie

dlpierce:
trixie has ImageMagick 7 (finally!) so we should be able to use the official package now.

* Upgrade Docker base image to debian trixie
* Use debian's imagemagick package


orangewolf:
🥳

PR #7196, opened 2025-09-19
Fix fileset migration - to 5.0-stable

laritakr:
### Summary

PDF and some other types of files had a text derivative file attached to the same fileset as the original file. When migrating these situations, the fileset never was submitted for migration due to the incorrect assumption that filesets would only have 1 file in Fedora. 

### Changes included

- Modifies the logic for triggering file migration on a fileset.
- Brings in some changes from 5.0-flexible branch that were missed here, to facilitate Hyku's rake task for making fedora sample works
- removes based_near from the terms list used for batch changes, as it fails trying to load existing location information.


orangewolf (approved):
are any of these changes needed in main or are they all there already?

laritakr:
I am making separate PRs for main and for 5.0-flexible branches... that's still underway once I get this to pass lint.

PR #7207, opened 2025-09-24
Contribute back locale bug fix (5.0-flexible)

laritakr:
### Fixes

Ref https://github.com/notch8/hykuup_knapsack/issues/526


ShanaLMoore:
I think we need one for 5.0-stable as well?

PR #7213, opened 2025-09-29
Allow additional metadata to be set on FileSets via the WorkUploadsHandler

maxkadel:
### Fixes
Refs https://github.com/notch8/hykuup_knapsack/issues/454

### Summary
Non-UI uploaders, such as Bulkrax, cannot set metadata on FileSets that then allow them to re-find the uploaded object, such as :source_identifier.

* Tested via the UI against 
  * [Hyku](https://github.com/samvera/hyku/tree/file_set_flexible_metadata)
  * [Bulkrax](https://github.com/samvera/bulkrax/pull/1067/commits/33f883b0e78af01539fc23b3b83897c21e42f3b3)
* Example testing CSV
```
file,model,source_identifier,parents,title,creator,rights_statement,visibility
,GenericWork,gw-123,,"Custom Generic Work Title",KKW,http://rightsstatements.org/vocab/NKC/1.0/,open
dummy.jpg,FileSet,fs-123,gw-123,Custom File Set Title,KKW,,restricted
```
  * My example directory looks like this - Archive.zip is the files directory and fs_test.csv zipped together, and that's what I uploaded to my Bulkrax importer:
```
├── file_sets
│   ├── Archive.zip
│   ├── files
│   │   └── dummy.jpg
│   └── fs_test.csv
```

I'm not sure I got the label right - feel free to change or suggest a change.

### Detailed Description
Previously, we were passing `file_set_params` to the `#add` method and setting them to an instance variable, but then we weren't doing anything with them. Instead, the only attributes that were set on the FileSet were in a pre-defined hash in `file_set_args`, not anything from the passed in parameters. 

This update passes the `file_set_params` through to the `file_set_args` method and merges them. The pre-defined values take precedence, and only attributes available on the schema can be set.

In order to pair up the files with the file sets we are using the index of the arrays to match them up.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/forms/hyrax/forms/batch_edit_form.rb
+++ b/app/forms/hyrax/forms/batch_edit_form.rb
@@ -5,8 +5,7 @@ module Hyrax
       # Used for drawing the fields that appear on the page
       self.terms = [:creator, :contributor, :description,
                     :keyword, :resource_type, :license, :publisher, :date_created,
-                    :subject, :language, :identifier, :based_near,
-                    :related_url]
+                    :subject, :language, :identifier, :related_url]
       self.required_fields = []
       self.model_class = Hyrax.primary_work_type
 
@@ -39,7 +38,6 @@ module Hyrax
          { subject: [] },
          { language: [] },
          { identifier: [] },
-         { based_near: [] },
          { related_url: [] },
          { permissions_attributes: [:type, :name, :access, :id, :_destroy] },
          :on_behalf_of,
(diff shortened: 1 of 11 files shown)


app/services/hyrax/work_uploads_handler.rb
```diff
@@ -133,14 +138,19 @@ def append_to_work(file_set)
...
+    #
     # @return [Hash{Symbol => Object}]
-    def file_set_args(file)
-      { depositor: file.user.user_key,
-        creator: file.user.user_key,
-        date_uploaded: file.created_at,
-        date_modified: Hyrax::TimeService.time_in_utc,
-        label: file.uploader.filename,
-        title: file.uploader.filename }
+    def file_set_args(file, file_set_params = {})
```
maxkadel, line 145: Interested in feedback on whether the file_set_params should override values in the built hash or not. Right now the built hash overrides values passed in, for any values that have been passed in.

maxkadel: For now I've made the decision to allow the pre-filled hash to be overridden.

PR #7240, opened 2025-10-13
Reload work when attaching FileSets

maxkadel:
### Fixes
Connected to https://github.com/notch8/hykuup_knapsack/issues/558

### Summary
When the WorkUploadsHandler is instantiated multiple times with different FileSets, it should attach all the FileSets to the Work.

In order to make sure we are updating the most recent version of the work, and not lose any associated FileSets, we should first lock the work, then re-load it from the persistence layer, and then perform any transformations on it.

Since the `:work` was set as an instance variable with an `attr_reader`, we needed to avoid re-using `work` as a variable name, in case we used the cached `@work`.

* On Hyku ([this commit](https://github.com/samvera/hyku/commit/9877ba16bd5b1796ee75554649256bfb6ad343b6) or later), create an importer using the attached zip file and the CSV parser
  * [Example zip to upload](https://github.com/user-attachments/files/22888830/fs_lots.zip)
  * Do not manually kick off any background jobs - allow them to be set off by the background job manager
* On 5.0-flexible right now, only a couple of the FileSets will be attached to the work. On this branch, all FileSets should be attached

- `notes-bugfix` Bug Fixes

@samvera/hyrax-code-reviewers

### DIFF
--- a/.dassie/config/metadata_profiles/m3_profile.yaml
+++ b/.dassie/config/metadata_profiles/m3_profile.yaml
@@ -64,7 +64,7 @@ properties:
       default: Enter a standardized title for display. If only one title is needed,
         transcribe the title from the source itself.
     display_label:
-      default: Title
+      default: blacklight.search.fields.show.title_tesim
     index_documentation: displayable, searchable
     indexing:
     - title_sim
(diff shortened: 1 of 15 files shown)


app/services/hyrax/work_uploads_handler.rb
```diff
@@ -92,47 +82,63 @@ def add(files:, file_set_params: [])
...
+        reloaded_work.representative_id = file_ids.first if reloaded_work.respond_to?(:representative_id) && reloaded_work.representative_id.blank?
+        reloaded_work.thumbnail_id = file_ids.first if reloaded_work.respond_to?(:thumbnail_id) && reloaded_work.thumbnail_id.blank?
+        @persister.save(resource: reloaded_work)
+        Hyrax.publisher.publish('object.metadata.updated', object: reloaded_work, user: user)
+      end
+    end
 
+    ##
+    # @api private
+    def update_file_set(file_set, file)
```
maxkadel: Better name? set_permissions_and_visibility

app/services/hyrax/work_uploads_handler.rb
```diff
@@ -92,47 +82,63 @@ def add(files:, file_set_params: [])
...
+      files.each_with_object([]).with_index do |(file, arry), index|
+        file_set = find_or_create_file_set(file, @file_set_params[index] || {})
+        update_file_set(file_set, file)
+        arry << { file_set: file_set, user: file.user, job: ValkyrieIngestJob.new(file) }
+      end
+    end
+
+    ##
+    # @api private
+    def append_file_sets_to_work(file_ids:, user:)
```
maxkadel: Can this live in the Work?

maxkadel: This has to reload the @work instance variable, so I think we can't just move it part and parcel to the Work.

laritakr (approved):
Good work

PR #7256, opened 2025-10-22
[5.0-flexible] Convert view:label to display_label 

orangewolf:
### Fixes

Changing the display_label should change both the index page and the show page

Added a rake task `hyrax:metadata:update_labels` to update profile labels easily.

@samvera/hyrax-code-reviewers


laritakr (commented):
I believe this should include changes to all of the various sample m3 profiles as well? Don't want to approve without those present as well.

Out of time for a detailed look but this seemed reasonable on a quick look.

PR #7257, opened 2025-10-22
Flexible display labels fixes

orangewolf:
### Fixes

Flexible metadata has display_labels but they are not respected by the UI. after this commit they will be. 

We added a rake task to change the labels in your profile if needed. Use `./bin/rails hyrax:metadata:update_labels[config/metadata_profiles/m3_profile.yaml]` to update a profile on disk.

@samvera/hyrax-code-reviewers


randalldfloyd (commented):
@orangewolf 
Is there something more I need to do for keys to get interpreted in the show view?  Running koppie, using this branch and its m3 profile, I just get the key names literally as text:
image

PR #7276, opened 2025-11-10
Fix releasing embargo and leases on the day they are expired

orangewolf:
this solves a 1 second bug around how midnights are calculated, which causes day of leases to incorrectly set the permissions to private. This fix has been used in Hyku for a couple of years now. It becomes exacerbated because Hyku does automatic embargo/lease releases as a timed job.

See https://github.com/samvera/hydra-head/pull/566 for a corresponding hydra-head fix.

@samvera/hyrax-code-reviewers


mark-dce (approved):
Nice work 👍

PR #7291, opened 2025-12-01
Override hydra-editor screen reader text to 'this field' for accessib…

### LINKED ISSUE #7238: As a screen reader user editing work descriptions, I want the Remove label for multivalued entries to not reference "previous" values
### Descriptive summary

During screen reader testing, tester noted label for "Remove" for multivalued works to reference "previous" rather than value with the "Remove" option

1. Use a screen reader to view your Hyrax application (Windows: Narrator, NVDA, or JAWS; Mac: VoiceOver)
2. In the Dashboard, edit a work
3. Add multiple titles
4. Note the screen reader label refers to the Remove action referring to the "previous" entry

Image

- [ ] Screen reader label for remove says "this" instead of "previous"

Screen reader testing task was to add and remove another title, tester was uncertain that the "Remove" action would remove the correct value.

trmccormick:
Accessibility fix for screen reader text in hydra-editor (#7238)

### Fixes

Fixes #7238 ; refs #7238

### Summary

Override hydra-editor screen reader text to "this" for accessibility

* The screen reader text for the remove button should now say "this" instead of "previous".
* Test with a screen reader to confirm the updated text is announced.
* Ensure the change appears on all relevant forms where hydra-editor is used.

`notes-bugfix` Bug Fixes

### Detailed Description

This PR adds a JavaScript override to update the screen reader text for the remove button in hydra-editor fields. Previously, the text read "previous", which was confusing for users relying on assistive technology. The override changes the text to "this" for clarity and improved accessibility.

The change is implemented in `app/assets/javascripts/hyrax/hydra_editor_override.js` and loaded via the asset pipeline. No changes were made to the hydra-editor gem itself.

* Add `hyrax/hydra_editor_override.js` to override screen reader text
* Update asset manifest to load the override after hydra-editor

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/assets/javascripts/hyrax.js
+++ b/app/assets/javascripts/hyrax.js
@@ -94,3 +94,4 @@
 
 // this needs to be after batch_select so that the form ids get setup correctly
 //= require hyrax/batch_edit
+//= require hyrax/hydra_editor_override
--- a/app/assets/javascripts/hyrax/hydra_editor_override.js
+++ b/app/assets/javascripts/hyrax/hydra_editor_override.js
@@ -1,0 +1,6 @@
+// app/assets/javascripts/hyrax/hydra_editor_override.js
+$(document).on('ready page:load turbolinks:load', function() {
+  $('.remove .sr-only').each(function() {
+    $(this).html('this <span class="controls-field-name-text">field</span>');
+  });
+});


dlpierce:
I'm concerned the css selector is not specific enough and might cause unintended side effects. I'm also not sure if using "this" is a significant improvement over "previous".

I did a little hacking in a browser inspector and found it should be possible to have the current value of the input field be read out by the screen reader (e.g. "Remove Title [alternate title of work]"). Doing so requires the generated inputs to have their own unique id which is non-trivial in the existing hydra-editor code, but doable.

orangewolf (approved):

orangewolf (changes requested):
@trmccormick I approved this too soon - @dlpierce makes a good point in the comments. I'm hoping its not a big lift to get this to a better state.

PR #7298, opened 2025-12-03
Speed up the IIIF Manifest feature test

mark-dce:
### Fixes

The IIIF manifest feature tests are extremely slow.

### Summary
The IIIF test setup created 3 works, and 12 additional file sets, all persisted to disk for the test. Only a single resource of each type is needed for the primary test. We need to persist an even smaller set of items for the access restriction test.

### Changes proposed in this pull request
Refactor the test to avoid creating unnecessary test objects. Also use factories where possible rather than manually constructing resources. We also do not need to clear the repository for this test.

#### TEST SPEED BEFORE *(over 5 minutes!!!)*
```
Slowest examples
  building a IIIF Manifest with a user missing read permissions on children generates a bare manifest
    ./spec/features/iiif_manifest_spec.rb:39
    162.382843552 seconds
  building a IIIF Manifest gets a full manifest for the work
    ./spec/features/iiif_manifest_spec.rb:21
    153.629054135 seconds

Finished in 5 minutes 18 seconds (files took 13.1 seconds to load)
2 examples, 0 failures
```

#### TEST SPEED AFTER *(more than 5 times faster)*
```
Slowest examples
  building a IIIF Manifest gets a full manifest for the work
    ./spec/features/iiif_manifest_spec.rb:23
    32.677574705 seconds
  building a IIIF Manifest with a user missing read permissions on children generates a bare manifest
    ./spec/features/iiif_manifest_spec.rb:42
    14.026897021 seconds

Finished in 48.9 seconds (files took 13.47 seconds to load)
2 examples, 0 failures
```

@samvera/hyrax-code-reviewers

### DIFF
--- a/spec/features/iiif_manifest_spec.rb
+++ b/spec/features/iiif_manifest_spec.rb
@@ -1,76 +1,44 @@
 # frozen_string_literal: true
 require 'rails_helper'
 
-RSpec.describe 'building a IIIF Manifest', :clean_repo do
-  let(:work) { valkyrie_create(:comet_in_moominland, :public, description: ['a novel about moomins']) }
+RSpec.describe 'building a IIIF Manifest', :aggregate_failures do
+  let(:work) { valkyrie_create(:comet_in_moominland, :public, description: ['a novel about moomins'], members: children) }
+  let(:children) { valkyrie_create_list(:monograph, 1, members: file_sets) + file_sets }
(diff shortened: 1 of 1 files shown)


dlpierce (approved):
Looks good, thank you! I expect the large number of objects was a holdover from performance testing on the manifest side.

PR #7299, opened 2025-12-04
Pull relevant work from 5.0-flexible for OAI mappings

laritakr:
### Fixes

Refs https://github.com/samvera/hyrax/pull/7297

### Summary

Adds new mappings_data_for method to extract mappings from profile to be used for OAI-PMH. 

For OAI-PMH we need the mappings and indexing info for properties with the specified mapping. The result of this will be used to configure the OAI-PMH provider with the flexible metadata.

Includes a bugfix to correctly load hydra_model in the solr document when using valkyrie_transition.

### DIFF
--- a/app/models/concerns/hyrax/solr_document_behavior.rb
+++ b/app/models/concerns/hyrax/solr_document_behavior.rb
@@ -82,7 +82,7 @@ module Hyrax
 
     # Method to return the model
     def hydra_model(classifier: nil)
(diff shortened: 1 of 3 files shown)


mark-dce (approved):
👍

laritakr:
Moving this back to draft, as we may need additional content returned for the main branch, since flexibility is defined by model.

Issue #7303, opened 2025-12-04
Deprecation warnings should display full warning text


mark-dce:
### Descriptive summary

Hyrax is not displaying the text of deprecation messages.

### ISSUE
The problem affects multiple Samvera gems and is fully described in this maintenance ticket: https://github.com/samvera/maintenance/issues/175

#### Warning using wrong call signature
```
DEPRECATION WARNING: You are using deprecated behavior which will be removed from the next major or minor release. 
          (called from <class:Work> at /app/samvera/hyrax-engine/app/models/hyrax/work.rb:98)
```

#### Warning using the correct call signature
```
DEPRECATION WARNING:  globally including `Hyrax::Schema(:core_metadata)` at the work level is deprecated.
          Add `include Hyrax::Schema(:core_metadata)` to your individual work classes or set
          up flexible metadata, then set `work_default_metadata = false` in config/hyrax.rb to remove this warning.
          (called from <class:Work> at /app/samvera/hyrax-engine/app/models/hyrax/work.rb:98)
```

### Acceptance Criteria/Expected Behavior

- [x] Update Hyrax to consistently use the Rails call signature
- [x] Add a patch to display message text until all the underlying dependencies are updated (per the maintenance ticket https://github.com/samvera/maintenance/issues/175 )

mark-dce:
Initial updates to Hyrax complete. We'll open a new issue when the other dependencies are updated and we're ready to do the final Hyrax updates.

PR #7306, opened 2025-12-04
Handle old and new style Deprecation#warn calls

mark-dce:
### Summary
The code (and dependencies) use a mix of signatures to call the Deprecation#warn method. In some cases, this swallows the text of the warning.

### Changes proposed in this pull reques
Patch the method to inspect arguments and ensure they get passed in the correct positions to the older method provided by the gem.

#### BEFORE
```
DEPRECATION WARNING: You are using deprecated behavior which will be removed from the next major or minor release.
          (called from <class:Work> at /app/samvera/hyrax-engine/app/models/hyrax/work.rb:98)
```

#### AFTER
```
DEPRECATION WARNING:  globally including `Hyrax::Schema(:core_metadata)` at the work level is deprecated.
          Add `include Hyrax::Schema(:core_metadata)` to your individual work classes or set
          up flexible metadata, then set `work_default_metadata = false` in config/hyrax.rb to remove this warning.
          (called from <class:Work> at /app/samvera/hyrax-engine/app/models/hyrax/work.rb:98)
```
### References
* https://github.com/samvera/hyrax/issues/7303
* https://github.com/samvera/maintenance/issues/175

@samvera/hyrax-code-reviewers

### DIFF
--- a/config/initializers/deprecation_patch.rb
+++ b/config/initializers/deprecation_patch.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+#
+# OVERRIDE deprecation v1.1.0  – handle calls using either the deprecation gem method signature
+#                                or newer Rails method signatues.
+#                                see: https://github.com/samvera/hyrax/issues/7303
+# Remove when: dependencies no longer pull in the deprecation gem
+
+module Deprecation
+  module DeprecationWarningPatch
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
This is such a helpful fix!  Good catch!

Issue #7302, opened 2025-12-04
Enhance Google Scholar indexing


maxkadel:
### Descriptive summary
There is some metadata we could add to the header to enhance works' indexing in Google Scholar.

### Steps to reproduce the behavior in User Interface (UI)

1. Go to a scholarly work (e.g. a dissertation)
2. Inspect the html using the dev tools, look in the header for the `<!-- Google Scholar metadata -->` comment and look at the meta data available
3. Note that on main there is a comment on a number of scholarly fields that are not currently available, in addition to `citation_dissertation_institution`, which is required for theses and dissertations

### Acceptance Criteria/Expected Behavior

- [x] All fields currently in the comment can be displayed in the header meta data
- [x] `citation_dissertation_institution` is in the header meta data

### Rationale (for feature request only)

See [Google Scholar's indexing documentation](https://scholar.google.com/intl/en/scholar/inclusion.html#indexing) for more context. Since so many institutions use Hyrax for institutional repositories, or for theses and dissertations, it makes sense to prioritize Google Scholar SEO.

### Related work

See https://github.com/notch8/hykuup_knapsack/issues/578 for the ticket that led to this

PR #7307, opened 2025-12-05
Add more properties to google scholar presenter

maxkadel:
### Fixes
refs #7302

Pulled from work done in https://github.com/UNC-Libraries/hy-c/pull/897 by @bbpennel

### Summary
In order to enhance the indexing of works by Google Scholar, we should follow their recommendations for what metadata to make available.

* In order to test this in the UI, you need an application that has work types that include these types of fields - Hyku includes ETDs, which has `degree_grantor` as a property. UNC's Hy-C has an [Article model](https://github.com/UNC-Libraries/hy-c/blob/main/app/models/article.rb) which has the journal-related properties.

I do not know whether it would be appropriate to add to the dassie and koppie test apps to add an ETD model, a Journal Article model, or both.

* Add possible metadata for Google Scholar indexing, when available on a given model
* Refactors view specs to pull spec for a partial into their own spec

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/controllers/concerns/hyrax/collections_controller_behavior.rb
+++ b/app/controllers/concerns/hyrax/collections_controller_behavior.rb
@@ -3,7 +3,8 @@ module Hyrax
   module CollectionsControllerBehavior
     extend ActiveSupport::Concern
     include Blacklight::AccessControls::Catalog
-    include Blacklight::Base
+    include Blacklight::Configurable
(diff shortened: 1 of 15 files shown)


dlpierce (approved):
Thanks for this!
I'm fine with adding these fields to koppie's Monograph if that would be sufficient for testing purposes.

PR #7312, opened 2025-12-09
Admin Set Create Form For Valkyrie - main version

### LINKED ISSUE #7263: Admin Set titles do not save when admin set is created
### Descriptive summary

On pg.nurax, the title of an admin set is not saved when the admin set is created. I can edit the admin set and add a title and then the value saves as expected. 

1. Log in to pg.nurax as a repository administrator
2. Go to `Dashboard` -> `Collections` -> `Add New Collection`
3. Select `Admin Set` and click `Create Collection`
4. Add a value in the `Title` field and click `Save`
5. Note that the `Title` field is now blank, but you can still click `Cancel` and navigate away from the page without an error

orangewolf:
Same as #7311 but this version is for main

### Fixes

Fixes #7263

### Summary

For Valkyrie works, if a field is missing in the admin set create, the admin set gets created anyway. This code shows a user level error flash instead and reloads the form. If flexible is true, the creator is mandatory but never set. Also additional fields are never shown to the user, only title and description. This code expands the fields display to handle whatever properties the AdministrativeSetForm object knows about.

* Set flexible on (see docs/flexible.md)
* Create an admin set via the UI
* See that creator is present and editable
* Save the admin set without creator and see that an error is returned
* Save the admin set with creator and see that the admin set saves correctly.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/presenters/hyrax/google_scholar_presenter.rb
+++ b/app/presenters/hyrax/google_scholar_presenter.rb
@@ -82,5 +82,44 @@ module Hyrax
     def title
       Array(object.try(:title)).sort.first || ""
     end
+
+    ##
+    # @return [String] a string representing the degree granting institution(s)
+    #   as a semicolon delimited list
+    def degree_grantor
+      Array(object.try(:degree_grantor)).join('; ')
+    end
+
+    ##
+    # @return [String] a string representing the journal title
+    def journal_title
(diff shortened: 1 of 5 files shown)


maxkadel (approved):
Looks good!

Issue #7314, opened 2025-12-14
Add blacklight_dynamic_sitemap, deprecate resourcelist


maxkadel:
### Descriptive summary

In order to sustainably create sitemaps, we should use the pre-existing blacklight_dynamic_sitemap gem, rather than trying to fix the resourcelist.

### Actual behavior

See https://github.com/samvera/hyrax/issues/59 - this issue has been open almost 10 years and not resolved. Instead of trying to fix the existing class, which probably would have been done by now if it were easy, instead add the blacklight_dynamic_sitemap gem, and deprecate the resourcelist.

### Acceptance Criteria/Expected Behavior

- [x] Blacklight dynamic sitemap is added as part of generated app
- [ ] Come up with a plan / follow-up ticket for adding a robots.txt that points to the generated sitemap
- [ ] The ResourceList is marked as deprecated for next major release

### Rationale (for feature request only)

Many institutions using Hyrax want their sites indexed by "good bots", e.g. Googlebot. Without an index page for sitemaps for large repositories, their sitemaps cannot be added. 

It's possible that by adding the sitemap to the robots.txt, it's less likely we will have the issue of bots crawling through all the combinations of facets, which can be very expensive for Solr.

### Related work

- For previous ticket, see https://github.com/samvera/hyrax/issues/59

PR #7315, opened 2025-12-15
Add Blacklight Dynamic Sitemap

maxkadel:
### Fixes
refs #7314

### Summary
Adds blacklight_dynamic_sitemap to engine, test apps, and generators. Allows for paginated sitemap that should work for any size repository.

* Go to `/sitemap`
* Select one of the sub-pages - if there are very few works in your index, you may have to find an existing work and use that to find the right index (for non-noid using apps). Otherwise you may just have to check through all of them
* Administrative-type objects should not be returned, such as `Hydra::AccessControl` etc.
* Only public works should be listed

### Detailed Description
- Add `lib/generators/hyrax/templates/app/views/blacklight_dynamic_sitemap/sitemap/show.xml.builder` to the install generator and the test apps. It has to be in the local application's code, not the engine's, because otherwise the gem won't find it and will render its own version.
- Adds route for blacklight_dynamic_sitemap in generator
- Add initializer to configure for Hyrax apps
  - We have to use different configuration for apps that use noids vs. those that don't - if an app doesn't use noids, we can use their ids directly. Otherwise we have to add to the solrconfig.xml (see below, and the change to the Dassie app's solrconfig.xml).
- Add a note on how to test the generator locally. It may be obvious to other folks, but it wasn't to me, so I figured I'd make it easier for future-me.

### Upgrade notes
Newly generated apps should come with blacklight_dynamic_sitemap fully configured out of the box after this PR.

If you want your pre-existing app to be able to take advantage of the `blacklight_dynamic_sitemap` gem:
0. Update your Hyrax gem and bundle install, it should add blacklight_dynamic_sitemap as a dependency of Hyrax
1. If your app uses noids (check `Hyrax.config.enable_noids?`), you must add the following to your `solrconfig.xml`:
```xml
  <updateRequestProcessorChain name="add_hashed_id">
    <processor class="solr.processor.SignatureUpdateProcessorFactory">
      <bool name="enabled">true</bool>
      <str name="signatureField">hashed_id_ssi</str>
      <bool name="overwriteDupes">false</bool>
      <str name="fields">id</str>
      <str name="signatureClass">solr.processor.Lookup3Signature</str>
    </processor>

    <processor class="solr.LogUpdateProcessorFactory" />
    <processor class="solr.RunUpdateProcessorFactory" />
  </updateRequestProcessorChain>

  <requestHandler name="/update" class="solr.UpdateRequestHandler">
    <lst name="defaults">
      <str name="update.chain">add_hashed_id</str>
    </lst>
  </requestHandler>
```
2. You must add the view from `lib/generators/hyrax/templates/app/views/blacklight_dynamic_sitemap/sitemap/show.xml.builder` to your instance's `app/views/blacklight_dynamic_sitemap/sitemap/show.xml.builder` path
3. Add `mount BlacklightDynamicSitemap::Engine => '/'` to your `config/routes.rb` file

@samvera/hyrax-code-reviewers

### DIFF
--- a/hyrax.gemspec
+++ b/hyrax.gemspec
@@ -38,7 +38,7 @@ SUMMARY
   spec.add_dependency 'almond-rails', '~> 0.1'
   spec.add_dependency 'awesome_nested_set', '~> 3.1'
   spec.add_dependency 'blacklight', '~> 7.29'
-  spec.add_dependency 'blacklight_dynamic_sitemap'
+  spec.add_dependency 'blacklight_dynamic_sitemap', '~> 1.0'
   spec.add_dependency 'blacklight-gallery', '~> 4.6.4'
   spec.add_dependency 'breadcrumbs_on_rails', '~> 3.0'
   spec.add_dependency 'browse-everything', '>= 0.16', '< 2.0'


dlpierce (approved):
Thank you!

PR #7328, opened 2026-01-19
straighten out the Riiif file nameing 

orangewolf:
so that classes are not initialized in the initalizer. solves superclass mismatch in development issue. This is a longstanding annoyance, not really any new functionality. If the old style initializer is present in an app it should still work just fine, the new models just won't be used until said target app updates their initializers. 

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/models/hyrax/riiif/file.rb
+++ b/app/models/hyrax/riiif/file.rb
@@ -34,7 +34,7 @@ else
   module Hyrax
     module Riiif
       class File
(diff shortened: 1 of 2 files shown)


dlpierce:
This looks like it will make the riiif gem a dependency, which is something I got feedback about back when the current setup was committed.

dlpierce (approved):
I'm good with this. Thank you!

Issue #7255, opened 2025-10-22
Create new test app with Flexible turned on


maxkadel:
### Descriptive summary
Create a new test app with Flexible metadata turned on

Discussed during the [October 22, 2025 Samvera tech call](https://samvera.atlassian.net/wiki/spaces/samvera/pages/3454107649/Samvera+Tech+Call+2025-10-22).

### Acceptance Criteria/Expected Behavior
- [ ] There is a new test app named "allinson" in honor of Julie Allinson, who originally did a ton of work on flexible metadata
- [ ] It is based on the existing .koppie config, following the pattern of sirenia, with its own docker compose file (and possibly env file)
- [ ] It sets `HYRAX_FLEXIBLE=true` in development and test
- [ ] It runs as one of the apps in the CI matrix
- [ ] Any tests that are not passing are either made to pass or made pending only for when flexible metadata is on

orangewolf:
@maxkadel I believe this work is done and am closing this ticket. If you notice anything missing feel free to reopen.

PR #7333, opened 2026-01-28
Adding new compose file for a test app with flexible metadata enabled

randalldfloyd:
refs #7255

Currently, flexible is enabled in the Allinson app added via this PR, and functionally works when desk-checking the UI. It doesn't seem to introduce any breaking changes (though full QA on a Nurax is still needed.).  It's [enabled here in its docker compose file](https://github.com/samvera/hyrax/blob/allinson_test-app/docker-compose-allinson.yml#L21-L23).

TODO:
Figure out how to handle the test suite.   

Flexible seems to work in the test app, but it blows up the test suite spectacularly with roughly 2/3 tests failing.  This is potentially an effect of setting `HYRAX_DISABLE_INCLUDE_METADATA=true` , which is kind of necessary so that models only use the profile definitions and avoid including core/basic metadata modules.  But that might be at odds with how tests use base factories, setup expectations, mocks etc.

So, flexible is currently [disabled here in the spec helper](https://github.com/samvera/hyrax/blob/allinson_test-app/spec/spec_helper.rb#L13-L15), which overrides Allinson enabling it for itself.  But this begs the question on the strategy for testing.  Should we:

1. Get the whole suite passing with flexible fully enabled, and add yet another full pass through the suite in the matrix.
2. Develop a subset set of feature tests/work type model tests, and conditionally only run them when flexible is dynamically flipped on during the existing koppie/sirenia passes.


spec/spec_helper.rb
```diff
@@ -6,8 +6,13 @@
 
 # Analytics is turned off by default
 ENV['HYRAX_ANALYTICS'] = 'false'
+
+# Controlling Flexible Metadata is done per test application, but the test suite needs
+# to disable it for now because disabling the inclusion of core/basic metadata has a huge
+# impact on the suite.  It can be explicitly enabled or mocked where appropriate.
 ENV['HYRAX_FLEXIBLE'] = 'false'
 ENV['HYRAX_DISABLE_INCLUDE_METADATA'] = 'false'
+ENV['HYRAX_FLEXIBLE_CLASSES'] = 'none'
```
orangewolf: is `"none"` the right choice here of should it be `""`?

orangewolf (approved):

PR #7343, opened 2026-02-04
[i27] - includes workflow_state_name_ssim for indexing

ShanaLMoore:
Issue:
- https://github.com/samvera/bulkrax/issues/1080

A Bulkrax bug revealed a bug in Hyrax. When exporters ran with a set workflow status, they would never complete or even error. This is because Bulkrax queries for a solr index term that appears to have never been valkyrized.

ref: 
- https://github.com/samvera/bulkrax/blob/main/app/parsers/bulkrax/parser_export_record_set.rb#L147

<details> <summary> EXPORTER </summary> 

image

</details> 

before: 

image

```ruby
doc = SolrDocument.find 'd31f1527-1774-46
10-a1d8-0b1e854f89f7'
doc.keys.include?("workflow_state_name_ssim")
=> false
```

after: 

image

```ruby
doc = SolrDocument.find 'd31f1527-1774-46
10-a1d8-0b1e854f89f7'
doc.keys.include?("workflow_state_name_ssim")
=> true
```

### DIFF
--- a/app/indexers/concerns/hyrax/workflow_indexer.rb
+++ b/app/indexers/concerns/hyrax/workflow_indexer.rb
@@ -2,20 +2,12 @@
 
 module Hyrax
   module WorkflowIndexer
-    # Write the suppressed status and workflow roles and state into the solr_document
+    # Write the workflow roles and state into the solr_document
     # @param [Hash] solr_document the solr document to add the fields to
     def to_solr
       super.tap do |solr_doc|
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
Thanks for that catch! I removed it from app/indexers/concerns/hyrax/workflow_indexer.rb @laritakr

laritakr (approved):

PR #7346, opened 2026-02-09
Robots - fix typo

maxkadel:
Robots - fix typo in sitemap url


app/views/hyrax/homepage/robots.text.erb
```diff
@@ -7,11 +7,12 @@ User-agent: *
 # Disable crawling filters - it'll just slow down discovery of useful resources - better that they page.
 Disallow: /catalog*f[
 Disallow: /catalog*f%5B
+Disallow: /*?locale=
```
maxkadel: Ensure that this doesn't effectively disallow all paths because of automatically adding `locale=en` when browsing.

PR #7350, opened 2026-02-17
Add install generator for Ramp

kirkkwang:
#### Video Example
image

#### Audio Example
image

### Summary

Add an install generator for Ramp IIIF player for A/V resources.

* ideally someone will run the generator in a deployed environment like Nurax
  - `rails g hyrax:ramp:install`
* they would also need to restart the server
* add a audio or video resource and check the show page
* the viewer should be the Ramp player instead of Universal Viewer

#### Add install generator for Ramp

8a55f2dd422e0f8ea3bf4f7f3cea991704e00ecf

This commit will add an install generator for Ramp and also set the A/V
IIIF Viewer to use Ramp.

Usage:
```
rails generate hyrax:ramp:install
```

#### Switch from `shakapacker` to `jsbundling-rails`

3dd6861a7613a4f66bd4e14f5402af0623f090c9

This commit switches from using `shakapacker` to `jsbundling-rails` with
aligns better with Avalon's approach.

* add an install generator so Hyrax has another option for viewing A/V IIIF resources

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/views/hyrax/base/iiif_viewers/_ramp.html.erb
+++ b/app/views/hyrax/base/iiif_viewers/_ramp.html.erb
@@ -1,29 +1,5 @@
 <%# run `bundle exec rails g hyrax:ramp:install` to install Ramp IIIF A/V player %>
-<% content_for(:head) do %>
-  <%= javascript_pack_tag 'application' %>
-  <%= stylesheet_pack_tag 'application' %>
-<% end %>
-<div id="ramp-player" class="ramp-player">
-  <%= react_component("RampPlayer", { manifestUrl: main_app.polymorphic_url([main_app, :manifest, presenter], { locale: I18n.locale }) }) %>
-</div>
-<script>
-  function loadRampPlayer() {
-    // RAMP player caches the previous manifest URL, so we must unmount/remount
-    // the React component to ensure it loads the correct manifest for this work.
-    // We also need to wait for the Webpacker pack to load asynchronously before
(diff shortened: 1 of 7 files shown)


masaball (approved):
This looks good to me. Exciting to see this option being added. 

While I echo @cjcolvar that it would be good for our different applications to be using the same/similar pipleines, I want to caution that `react_on_rails` is maintained by Shakacode and intended to be used in tandem with Shakapacker. It has been working with `jsbundling-rails` in Avalon with our fairly basic configuration and this configuration seems even simpler than ours. But that is a potential concern with following Avalon's footsteps. We decided the risk of `react_on_rails` compatibility was worth it to get onto a more standard/modern asset pipeline and we could reassess if issues arose. Just another aspect for the hyrax team to consider before moving forward with this specific implementation.

kirkkwang:
@masaball Thanks for the insight, we can definitely reassess as well as time goes on and it's nice to have the Avalon team to model after

@cjcolvar that sounds good to me, also since this is just a generator, how would we even test it out on Nurax?  Someone would need to run it on the server and give it a restart I would think.  But then I would imagine it goes away after the next merge/deploy.  I suppose one way is that we check in the generated files to the internal apps like Universal Viewer

cjcolvar:
Draft PR created with new ramp standalone UMD package: https://github.com/samvera/hyrax/pull/7417

PR #7352, opened 2026-02-18
Expose Hyrax.schema_for and ResourceForm.for for context-aware schemas

ShanaLMoore:
# Summary

Adds a single place in Hyrax for context-aware schema and form resolution so that callers (e.g. Bulkrax) can support flexible metadata without depending on `HYRAX_FLEXIBLE`, `contexts`, or the Flexibility concern.

- **`Hyrax.schema_for(klass, admin_set_id: nil)`** — Returns the Valkyrie schema for a model class. When the admin set has contexts and the model is flexible, loads context-gated properties; otherwise returns the base schema. Reads `admin_set.contexts` only when the admin set responds to `:contexts`; rescues `ObjectNotFoundError`. One public API for “what properties are valid for this class + admin set?”
- **`ResourceForm.for(resource:, admin_set_id: nil)`** — When `admin_set_id` is provided and the resource is flexible, applies the admin set’s contexts to the resource before building the form so context-specific fields appear. Callers no longer set `resource.contexts` themselves.

Backward compatible: when `HYRAX_FLEXIBLE=false` or the admin set has no contexts, behavior is unchanged.

# Ticket Number

Resolves https://github.com/samvera/hyku/issues/2935 (with Bulkrax PR).

# Screenshots / Video

N/A — API and form behavior only. End-to-end behavior is visible in the Bulkrax PR (validation/import with context-specific columns).

# Expected Behavior

1. **`Hyrax.schema_for(klass, admin_set_id: 'id')`** — When the admin set has contexts (e.g. `special_context`) and the model is flexible, the returned schema includes context-gated properties (e.g. `dimensions`). When the admin set is missing or has no contexts, or the model is not flexible, returns the base schema and does not raise.
2. **`ResourceForm.for(resource: work, admin_set_id: 'id')`** — When the admin set has contexts and the resource is flexible, the built form shows context-specific fields. When `admin_set_id` is nil or the resource is not flexible, form building is unchanged.
3. **Backward compatibility** — When `HYRAX_FLEXIBLE=false`, `schema_for` and `ResourceForm.for` do not pass or read contexts; no `ArgumentError` or `NoMethodError`.

# Sample Files

N/A — no CSV or profile changes in Hyrax. Sample M3 profile and CSV are described in the [Bulkrax PR.](https://github.com/samvera/bulkrax/pull/1114)

# Notes

- **Files:** `lib/hyrax.rb` (`schema_for`), `app/forms/hyrax/forms/resource_form.rb` (`for(admin_set_id:)`, `apply_admin_set_contexts`).
- **Consumers:** Bulkrax (and any other importer/validator) should call `Hyrax.schema_for` and `ResourceForm.for(resource:, admin_set_id:)` instead of resolving contexts or calling `flexible?` themselves.
- **Testing:** New specs in `spec/lib/hyrax_spec.rb` (`.schema_for`) and `spec/forms/hyrax/forms/resource_form_spec.rb` (`.for` with `admin_set_id`).

### DIFF
--- a/lib/hyrax.rb
+++ b/lib/hyrax.rb
@@ -150,7 +150,17 @@ module Hyrax
     query_service.custom_queries
   end
 
-  def self.schema_for(klass, admin_set_id: nil)
+  ##
+  # Returns the schema for a resource class, optionally scoped by admin set contexts.
+  #
+  # When +admin_set_id+ is present and flexible schema is enabled, the admin set's
+  # contexts are used so flexible resources get a context-aware schema. If the
+  # admin set is not found, falls back to the base schema without raising.
+  #
+  # @param klass [Class] a Valkyrie resource class (e.g. Hyrax::Work)
+  # @param admin_set_id [String, nil] optional admin set id to resolve contexts from
+  # @return [Dry::Types::Schema, Array] the schema (keys/types or enumerable of keys)
+  def self.schema_for(klass:, admin_set_id: nil)
     contexts = if admin_set_id.blank? || !config.flexible?
                  []
                else
@@ -161,6 +171,11 @@ module Hyrax
     fallback_schema(klass)
   end
 
+  ##
+  # @api private
+  #
+  # @param admin_set_id [String]
+  # @return [Array<String>] context identifiers from the admin set, or []
   def self.schema_contexts_for(admin_set_id)
     return [] if admin_set_id.blank?
     admin_set = query_service.find_by(id: admin_set_id)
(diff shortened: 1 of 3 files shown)


lib/hyrax.rb
```diff
@@ -149,4 +149,36 @@ def self.query_service
...
+               else
+                 schema_contexts_for(admin_set_id)
+               end
+    resolve_schema(klass, contexts)
+  rescue Valkyrie::Persistence::ObjectNotFoundError
+    fallback_schema(klass)
+  end
+
+  def self.schema_contexts_for(admin_set_id)
+    return [] if admin_set_id.blank?
```
ShanaLMoore: redudant

PR #7415, opened 2026-04-15
Apply permission template when work moves to a different admin set

laritakr:
### Summary

When a work's admin set changes during an update, the new admin set's permission template is now applied and the old template's grants are removed (preserving shared, manual, and depositor permissions)
File set permissions are propagated via InheritPermissionsJob after the work's permissions are updated
Follows the existing changed_collection_membership pattern in the Save step to detect admin set changes

- `notes-bugfix` Bug Fixes

@samvera/hyrax-code-reviewers


orangewolf (approved):
this looks great

PR #7416, opened 2026-04-16
Retrieve all collections for forming relationships

laritakr:
### Fixes

When searching for collections to create relationships, we should return all collections that match the search criteria, not just the first 100. 

- `notes-bugfix` Bug Fixes

### Detailed Description

This commit removes the limit on collections when preparing the list for selection. A paginated search is introduced, limited by config variables.


app/services/hyrax/uncapped_solr_query.rb
```diff
@@ -0,0 +1,24 @@
...
+  #
+  # The caller provides a block that accepts a row count and returns
+  # a Solr response object (one that responds to .response['numFound']
+  # and .documents).
+  module UncappedSolrQuery
+    # @yield [Integer] rows — the number of rows to request
+    # @yieldreturn [Blacklight::Solr::Response] Solr response
+    # @return [Blacklight::Solr::Response] the full response, or the empty
+    #   count response when there are no results
+    def self.call
```
orangewolf, line 16: I feel like this needs to cap the single result set at some value (say 1k) that should probably be exposed as a settings. if the results are greater than that amount you really need to be doing paginated loops to gather all the results. And even if you do that, you need to be careful not to overload memory by loading too many solr objects in to memory at once. 
100k ids? no problem. 100k full objects? 💣

orangewolf (changes requested):

orangewolf (approved):
this looks great

PR #7418, opened 2026-04-17
Display VTT transcripts in audio/video players

### LINKED ISSUE #305: Annotations tab fails in Vanilla JS
When loading a manifest with annotations in vanilla js Clover, this message appears:

```
Something went wrong
Error message: l is not a function 
```

Here is an example:
https://markpbaggett.github.io/static_iiif/vtts/seipp.html

eltiffster:
## Summary For Release Notes

Allow users to select and display [WebVTT](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API) files as transcripts/captions for audio/video files.

If using Flexible Metadata please add the following to your profile.
```yaml
properties:
  transcript_ids:
    available_on:
      class:
        - Hyrax::FileSet
    indexing:
      - transcript_ids_ssim
    data_type: array
    display_label: Transcripts
    form:
      primary: false
    property_uri: http://vocabulary.samvera.org/ns#transcriptIds
```

### Guidance Setup:
* Create a work with an audio/video representative file set and a VTT file set.
* Once both files have been uploaded, go to the edit page for the audio/video file set. There should be a form field titled "Transcripts" that lists your VTT file as an option. Select your VTT file and click Save.
* If IIIF/AV Support is not enabled (via the Features Dashboard):
  * Go to a file set or work page
  * Click the 3 dots in the audio/video player and click "Captions" to enable captions.
* If IIIF/AV Support is enabled, you'll need to first install the Clover IIIF viewer since UV doesn't support captions at time of writing:
   * In your webapp (`.dassie` or `/app/samvera/hyrax-webapp` in Docker), run `rails generate hyrax:iiif_viewer clover` to install
   * **NOTE: If you previously installed Clover, you'll need to run `rails destroy hyrax:iiif_viewer clover`, then `rails generate hyrax:iiif_viewer clover` to reinstall fresh copies of the viewer files.** The `clover.js` file was altered to fix [this issue](https://github.com/samvera-labs/clover-iiif/issues/305) and an extra class was added to a `div` in `clover.html.erb`.
  * Once Clover is installed and the transcript has been saved to the audio/video file, go the work page. You should be able to display the captions by clicking the 3 dots in the audio/video player toolbar and enabling them. If you click on the "Annotations" tab, you should also see the interactive transcript with timestamps (example pictured below).
screenshot of the interactive transcript in Clover's Annotations tab, side-by-side with the video player

---

- `notes-major` (I think) due to needing to install a new gem for converting `language` field values into a language code readable by the IIIF viewer

### Detailed Description

This is a continuation of work started before/during the March 2026 Community Sprint. More context/discussion on implementation was recorded on the [the Sprint Board](https://github.com/orgs/samvera/projects/56/views/1?pane=issue&itemId=160119656&issue=samvera%7Chyrax%7C7258).

After uploading VTT file(s) to a work with an audio/video (AV) file, users can choose to use the VTT file as the subtitles/captions file for the corresponding audio/video. This is done by editing the AV file set, selecting the VTT file by title, and saving the AV file set (see screenshot below). Under the hood, the VTT file set ids are saved to the `transcript_ids` attribute of the AV file and indexed as `transcript_ids_ssim` in the AV file's Solr document.

The transcript(s) form is populated by a Solr query that searches for "sibling" file sets (i.e. file sets of the same parent work as the AV file set) with a `text/vtt` mime type. Currently, `text/vtt` is the only accepted mime type, but other mime types could be added in future. Users can select multiple VTT files per audio/video file. This approach also supports a nested work structure where a child work has a different transcript than other child works or its parent work.

captions_edit_interface

Selected transcripts are displayed via a `<track>` element in default audio/video partials. When using a IIIF AV viewer, the transcript is displayed in a IIIF manifest via an annotation, following the pattern of this [IIIF cookbook recipe](https://iiif.io/api/cookbook/recipe/0219-using-caption-file/):

```json
"annotations": [
  {
   "type": "AnnotationPage",
   "items": [
      {
         "type": "Annotation",
         "motivation": "supplementing",
         "body": {
              "id": "http://localhost:3000/transcripts/file_id.vtt",
              "type": "Text",
              "format": "text/vtt",
              "label": {
                 "none": ["title_or_label.vtt"]
               }
           },
        "id": "http://localhost:3000/concern/generic_works/id/manifest/canvas/id",
        "target": "http://localhost:3000/concern/generic_works/id/manifest/canvas/id"
      }
    ]
  }
 ]
```
Thanks to @kirkkwang and @trmccormick for their work on this!

* Add `transcript_ids` and `language` properties/attributes to ActiveFedora and Hyrax file set classes, their respective form classes, and their respective indexers
* Add UI hints/guidance for `transcript_ids` and `language` form fields in the file set edit form
* Pass a `Hyrax::FileSetPresenter` to `render_media_display_partial` instead of a Solr document (in `app/views/hyrax/file_sets/edit.html.erb`)
* Add VTT transcripts as annotations to the IIIF manifest and sort them by language/locale by @kirkkwang
* Add [homepage property](https://iiif.io/api/cookbook/recipe/0047-homepage/) to IIIF manifest by @kirkkwang 
* Add route and controller for serving transcripts to IIIF AV viewers by @kirkkwang
* Add [LanguageList gem](https://github.com/scsmith/language_list) for parsing 2-letter language code from a Solr document's language field
* Add `<track>` elements to `app/views/hyrax/file_sets/media_display/_video.html.erb` and `app/views/hyrax/file_sets/media_display/_audio.html.erb`: initial idea and technical plan by @trmccormick

### Possible Future Work
* Localization/translation for UI hints for the language and transcript fields
* The file set edit form simply renders the `views/hyrax/file_sets/_form.html.erb` partial for ActiveFedora file sets, while the Valkyrie version uses the `hydra-editor` gem, which renders partials in `views/records/edit_fields`. Is there a reason for this difference? Should the form be refactored?
* Ideally, Clover should be easier to configure, maybe with a JSON object/file like with Universal Viewer. This would require using a JS framework to modify it and rebuild the `clover.js` file.

@samvera/hyrax-code-reviewers


kirkkwang (dismissed):
This looks good to me, i think the only thing is to maybe take out the `ActiveTriples::Resource` case, it doesn't seem likely as a case at the moment anyways.

app/controllers/hyrax/transcripts_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Hyrax
+  class TranscriptsController < ApplicationController
+    def show
+      file_metadata = Hyrax.custom_queries.find_file_metadata_by(id: params[:id])
+      file_object = Hyrax.storage_adapter.find_by(id: file_metadata.file_identifier)
+
+      # Using .extracted_text is blocked by https://github.com/samvera/hyrax/issues/7410
+      transcription = file_object.read
```
orangewolf: does this create a potential memory issue? I believe we can stream the file directly instead of loading it in to memory and then sending it

app/models/concerns/hyrax/file_set/transcripts.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module Hyrax
+  class FileSet
+    module Transcripts
+      extend ActiveSupport::Concern
+
+      included do
+        property :transcript_ids, predicate: ::RDF::URI.new('http://vocabulary.samvera.org/ns#transcriptIds'), multiple: true do |index|
```
orangewolf: this should be conditional on `Hyrax.config.file_set_include_metadata` and this field should be added to the default m3 profile.

app/presenters/hyrax/iiif_manifest_presenter.rb
```diff
@@ -129,6 +129,17 @@ def sequence_rendering
       end
     end
 
+    ##
+    # @return [Array<Hash{String => String}>]
```
orangewolf, line 137: where is this one used?

app/services/hyrax/file_set_type_service.rb
```diff
@@ -42,11 +39,22 @@ def audio?
...
+    def video?
+      video_types.include?(mime_type)
+    end
+
     private
 
     def audio_types
       return ::FileSet.audio_mime_types if defined?(::FileSet) && ::FileSet.respond_to?(:audio_mime_types)
-      DEFAULT_AUDIO_TYPES
+      Hyrax.config.mime_types_map.fetch(:audio_mime_types)
```
orangewolf, line 52: I love this optimization

app/views/hyrax/file_sets/media_display/_video.html.erb
```diff
@@ -5,6 +5,10 @@
         <source src="<%= hyrax.download_path(file_set, file: 'webm', mime_type: 'video/webm') %>" type="video/webm" />
         <source src="<%= hyrax.download_path(file_set, file: 'mp4', mime_type: 'video/mp4') %>" type="video/mp4" />
         <%= t('hyrax.file_set.show.downloadable_content.video_tag_not_supported') %>
+        <% file_set.transcripts.each do |transcript| %>
```
orangewolf, line 8: is file_set.transcripts memoized?

orangewolf (changes requested):
This is really great work. I have a few questions and concerns, but the feature as a whole looks great. Let me know if you want to pair up on the m3 part or the file streaming.

app/views/hyrax/file_sets/media_display/_video.html.erb
eltiffster: It's memoized in the file set presenter (via Hyrax::DisplaysTranscripts):

```ruby
def transcripts
  return [] if transcript_ids.blank?
  @transcripts ||= begin
                      results = Hyrax::SolrQueryService.new
                                                       .accessible_by(
                                                         ability: (try(:current_ability) || ability),
                                                         action: :read
                                                       )
                                                       .with_ids(ids: transcript_ids)
                                                       .solr_documents
                      sort_transcripts_by_language(results)
                    end
end
```

app/presenters/hyrax/iiif_manifest_presenter.rb
eltiffster: In Clover, it's used for the [Homepage](https://iiif.io/api/cookbook/recipe/0047-homepage/) section in the About sidebar. Not strictly necessary for the manifest/viewer, but nice to have if someone reuses the manifest on an external site.

screenshot of the homepage attribute in IIIF manifest

eltiffster:
> This is really great work. I have a few questions and concerns, but the feature as a whole looks great. Let me know if you want to pair up on the m3 part or the file streaming.

Thanks, @orangewolf! I'm not very familiar with the flexible metadata features, so I'd appreciate any help I can get.

I tried adding `transcript_ids` to `config/metadata_profiles/m3_profile.yml` like so:
```yaml
properties:
  transcript_ids:
    available_on:
      class:
        - Hyrax::FileSet
    indexing:
      - transcript_ids_ssim
    data_type: array
    display_label: Transcripts
    form:
      primary: false
    property_uri: http://vocabulary.samvera.org/ns#transcriptIds
```
And also to both `.dassie` and `.koppie` in the same place. But I don't see `transcript_ids` field if I download the default profile through the interface. What am I doing wrong?

eltiffster:
> > This is really great work. I have a few questions and concerns, but the feature as a whole looks great. Let me know if you want to pair up on the m3 part or the file streaming.
> 
> Thanks, @orangewolf! I'm not very familiar with the flexible metadata features, so I'd appreciate any help I can get.
> 
> I tried adding `transcript_ids` to `config/metadata_profiles/m3_profile.yml` like so:
> 
> ```yaml
> properties:
>   transcript_ids:
>     available_on:
>       class:
>         - Hyrax::FileSet
>     indexing:
>       - transcript_ids_ssim
>     data_type: array
>     display_label: Transcripts
>     form:
>       primary: false
>     property_uri: http://vocabulary.samvera.org/ns#transcriptIds
> ```
> 
> And also to both `.dassie` and `.koppie` in the same place. But I don't see `transcript_ids` field if I download the default profile through the interface. What am I doing wrong?

So I had to delete the first default profile from the database, which allowed `Hyrax::FlexibleSchema.create_default_schema` to pick up the new field.

How should this change be handled for those who have already created a flexible schema? 

Should there be documentation advising people to add `transcript_ids` to their flexible metadata profile? Or a button that compares the default properties to the current profile's properties and creates a new schema with the added properties?

app/controllers/hyrax/transcripts_controller.rb
eltiffster: Thanks for catching this! I've changed the `TranscriptsController` to inherit from `DownloadsController` so it can use `send_file file.disk_path, data_options(file_metadata)` to stream the file.

orangewolf:
> 
> So I had to delete the first default profile from the database, which allowed `Hyrax::FlexibleSchema.create_default_schema` to pick up the new field.
> 
> How should this change be handled for those who have already created a flexible schema?
> 
> Should there be documentation advising people to add `transcript_ids` to their flexible metadata profile? Or a button that compares the default properties to the current profile's properties and creates a new schema with the added properties?

I updated the description so we can catch it in the release notes.

orangewolf (approved):

PR #7420, opened 2026-04-28
Add :redirects Flipflop feature flag

ShanaLMoore:
## Summary

- Adds a `:redirects` Flipflop feature flag (default off) to the `site_configuration` group in `config/features.rb`
- First slice (Slice 0) of the URL Redirects feature (notch8/palni_palci_knapsack#620)
- Dark-launch posture: toggling the flag on/off has no observable effect yet since no other code references it

## Test plan

- [ ] `:redirects` feature appears in the Hyrax admin Flipflop UI
- [ ] Toggling it on/off has no observable effect
- [ ] Existing Hyrax test suite passes unchanged

### DIFF
--- a/config/features.rb
+++ b/config/features.rb
@@ -36,6 +36,10 @@ Flipflop.configure do
     feature :read_only,
             default: false,
             description: "Put the system into read-only mode. Deposits, edits, approvals and anything that makes a change to the data will be disabled."
+
+    feature :redirects,
+            default: false,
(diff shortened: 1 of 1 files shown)


laritakr (dismissed):
Looks good to me.

Non-blocking to consider - do we want to put this into an experimental group (where we put the Hyku guided import) instead of the group you put it in at this point, since it's under construction?

laritakr (approved):
Looks great

PR #7421, opened 2026-04-28
Add structured `redirects` metadata field to works and collections

laritakr:
# Summary

Implements the data layer for URL Redirects (Phase 1, slice 1): 
Adds the data layer for a new URL Redirects feature: works and collections can carry a list of redirect entries (path + canonical flag + sequence) that downstream slices use to resolve legacy URLs to canonical Hyku URLs

Refs notch8/palni_palci_knapsack#621.

## Files involved: 
- New Hyrax::Redirect Valkyrie::Resource with path, canonical, sequence
- New :redirects schema in config/metadata/redirects.yaml (flexible: false mode)
- New `redirects` property in m3_profile.yaml (flexible: true mode)
- Hyrax::Work and Hyrax::PcdmCollection opt into :redirects (gated on work_default_metadata / collection_include_metadata? to avoid double-defining the attribute when flexible mode is active)
- New Hyrax::Indexers::RedirectsIndexer extracts paths from the nested entries into redirects_path_ssim; included into PcdmObjectIndexer and PcdmCollectionIndexer
- Schema loader's type_for falls back to Hyrax::* namespace when a type isn't found under Valkyrie::Types::*, letting schemas reference nested-resource types declared in Hyrax without monkey-patching the Valkyrie::Types module
- Redirect resource spec and FactoryBot factory

## What's new

- Hyrax::Redirect — a small Valkyrie::Resource with path, canonical, and sequence attributes. Designed to live as a nested entry inside a parent work or collection.
- redirects attribute on Hyrax::Work and Hyrax::PcdmCollection — a multi-valued list of Hyrax::Redirect entries. Available in both Hyrax flexible-metadata modes:
- flexible: false: new config/metadata/redirects.yaml schema, opted into via include Hyrax::Schema(:redirects) (gated on the same work_default_metadata / collection_include_metadata? configs the rest of the schema includes use).
- flexible: true: new redirects property in config/metadata_profiles/m3_profile.yaml with available_on.class listing Hyrax's work and collection classes.
- Solr indexing — Hyrax::Indexers::RedirectsIndexer mixin extracts each entry's path into a redirects_path_ssim field. Included in PcdmObjectIndexer (works) and PcdmCollectionIndexer (collections).

## Generic capability added (not specific to redirects)

- SchemaLoader#type_for now resolves nested-resource types from YAML. A schema entry with type: <name> resolves first as a Valkyrie::Types::* primitive (existing behavior, unchanged), then as a Valkyrie::Resource subclass (new).
- Short forms resolve under Hyrax:: first (type: redirect → Hyrax::Redirect); fully-qualified names work for downstream apps (type: MyApp::Citation).
- Uses String#classify instead of capitalize (so type: access_control would resolve to Hyrax::AccessControl, not Hyrax::Access_control).
- Validates resolved candidates with < Valkyrie::Resource so accidental matches against unrelated classes raise the standard "Unrecognized type" error.

This generic capability is what makes the redirects schema work without monkey-patching Valkyrie::Types, and is reusable for any future nested-resource attribute Hyrax or its consumers want to declare via YAML.

### DIFF
--- a/app/indexers/hyrax/indexers/pcdm_collection_indexer.rb
+++ b/app/indexers/hyrax/indexers/pcdm_collection_indexer.rb
@@ -10,7 +10,7 @@ module Hyrax
       include Hyrax::LocationIndexer
       include Hyrax::ThumbnailIndexer
       include Hyrax::Indexer(:core_metadata) if Hyrax.config.collection_include_metadata?
-      include Hyrax::Indexers::RedirectsIndexer
+      include Hyrax::Indexers::RedirectsIndexer if Hyrax.config.redirects_enabled?
       check_if_flexible(Hyrax::PcdmCollection)
 
       self.thumbnail_path_service = CollectionThumbnailPathService
--- a/app/indexers/hyrax/indexers/pcdm_collection_indexer.rb
+++ b/app/indexers/hyrax/indexers/pcdm_collection_indexer.rb
@@ -10,6 +10,7 @@ module Hyrax
       include Hyrax::LocationIndexer
       include Hyrax::ThumbnailIndexer
       include Hyrax::Indexer(:core_metadata) if Hyrax.config.collection_include_metadata?
+      include Hyrax::Indexers::RedirectsIndexer
       check_if_flexible(Hyrax::PcdmCollection)
 
       self.thumbnail_path_service = CollectionThumbnailPathService
--- a/app/indexers/hyrax/indexers/redirects_indexer.rb
+++ b/app/indexers/hyrax/indexers/redirects_indexer.rb
@@ -7,9 +7,8 @@ module Hyrax
     # resources that carry a `redirects` attribute. The redirect resolver
     # (`Hyrax::RedirectsController`) queries this field by path.
     #
-    # Safe to include unconditionally on resource indexers — when the resource
-    # has no `redirects` attribute (or an empty value), the Solr field is
-    # set to an empty array.
+    # Safe to include unconditionally on resource indexers — the Solr field
+    # is only written when the `:redirects` Flipflop feature is enabled.
     #
     # @example
     #   class WorkIndexer < Hyrax::Indexers::PcdmObjectIndexer
@@ -18,6 +17,7 @@ module Hyrax
     module RedirectsIndexer
       def to_solr(*args)
         super.tap do |document|
+          next document unless Flipflop.redirects?
           document['redirects_path_ssim'] = Array(resource.try(:redirects))
                                             .map { |entry| entry.try(:path) }
                                             .compact
--- a/app/indexers/hyrax/indexers/redirects_indexer.rb
+++ b/app/indexers/hyrax/indexers/redirects_indexer.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Indexers
+    ##
+    # Indexer mixin that emits the `redirects_path_ssim` Solr field for
+    # resources that carry a `redirects` attribute. The redirect resolver
+    # (`Hyrax::RedirectsController`) queries this field by path.
+    #
+    # Safe to include unconditionally on resource indexers — when the resource
+    # has no `redirects` attribute (or an empty value), the Solr field is
+    # set to an empty array.
+    #
+    # @example
+    #   class WorkIndexer < Hyrax::Indexers::PcdmObjectIndexer
+    #     include Hyrax::Indexers::RedirectsIndexer
+    #   end
+    module RedirectsIndexer
+      def to_solr(*args)
+        super.tap do |document|
+          document['redirects_path_ssim'] = Array(resource.try(:redirects))
+                                            .map { |entry| entry.try(:path) }
+                                            .compact
+        end
+      end
+    end
+  end
+end
--- a/app/indexers/hyrax/indexers/redirects_indexer.rb
+++ b/app/indexers/hyrax/indexers/redirects_indexer.rb
@@ -21,9 +21,8 @@ module Hyrax
       def to_solr(*args)
         super.tap do |document|
           next document unless Flipflop.redirects?
-          document['redirects_path_ssim'] = Array(resource.try(:redirects))
-                                            .map { |entry| entry.try(:path) }
-                                            .compact
+          next document unless resource.respond_to?(:redirects)
+          document['redirects_path_ssim'] = Array(resource.redirects).map(&:path).compact
         end
       end
     end
--- a/app/indexers/hyrax/indexers/redirects_indexer.rb
+++ b/app/indexers/hyrax/indexers/redirects_indexer.rb
@@ -7,12 +7,15 @@ module Hyrax
     # resources that carry a `redirects` attribute. The redirect resolver
     # (`Hyrax::RedirectsController`) queries this field by path.
     #
-    # Safe to include unconditionally on resource indexers — the Solr field
-    # is only written when the `:redirects` Flipflop feature is enabled.
+    # Include this mixin only in apps where `Hyrax.config.redirects_enabled?`
+    # is true (the inclusion site is the config gate). The mixin's body
+    # then needs only the Flipflop check — when the config is on, the
+    # `:redirects` feature is registered and `Flipflop.redirects?` is safe
+    # to call.
     #
     # @example
     #   class WorkIndexer < Hyrax::Indexers::PcdmObjectIndexer
-    #     include Hyrax::Indexers::RedirectsIndexer
+    #     include Hyrax::Indexers::RedirectsIndexer if Hyrax.config.redirects_enabled?
     #   end
     module RedirectsIndexer
       def to_solr(*args)
--- a/app/models/hyrax/pcdm_collection.rb
+++ b/app/models/hyrax/pcdm_collection.rb
@@ -41,22 +41,8 @@ module Hyrax
   #
   class PcdmCollection < Hyrax::Resource
     include Hyrax::Schema(:core_metadata) if Hyrax.config.collection_include_metadata?
-    # The `redirects` attribute is registered directly on the resource class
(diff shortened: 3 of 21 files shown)


orangewolf (changes requested):
looks great so far, but this should have a config variable and be off by default during development. It might make sense to do the config var as a seperate pr and then merge that in here

ShanaLMoore:
> looks great so far, but this should have a config variable and be off by default during development. It might make sense to do the config var as a seperate pr and then merge that in here

There is a separate PR that has already been merged  into main (#7420); it adds the :redirects Flipflop flag, defaulting to off.   https://github.com/samvera/hyrax/pull/7420.

orangewolf:
> > looks great so far, but this should have a config variable and be off by default during development. It might make sense to do the config var as a seperate pr and then merge that in here
> 
> There is a separate PR that has already been merged into main (#7420); it adds the :redirects Flipflop flag, defaulting to off. #7420.

That's great, so we just need to use it here.

laritakr:
We could use a config variable to allow it to happen at all. The flip flop
only controls the behavior but the attributes are present all the time.

ShanaLMoore:
[committed](https://github.com/samvera/hyrax/pull/7421/changes/b01879ae92e078e91cf459aeef9a3bdbf1caf056) on the branch and ready for re-review. Gated the RedirectsIndexer behind Flipflop.redirects? so the redirects_path_ssim Solr field is only written when the feature is enabled. cc @orangewolf

app/indexers/hyrax/indexers/redirects_indexer.rb
```diff
@@ -18,6 +17,7 @@ module Indexers
     module RedirectsIndexer
       def to_solr(*args)
         super.tap do |document|
+          next document unless Flipflop.redirects?
```
ShanaLMoore, line 23: @orangewolf Will this suffice or are you thinking of something different? like a hyrax config

ShanaLMoore:
TODO: Per the samvera tech call we need to use a config var. We need to honor flexible metadata and not always include them cc @laritakr we need to rework this

documentation/redirects.md
```diff
@@ -0,0 +1,164 @@
...
+
+Hyrax can register arbitrary URL paths as redirects to a work or collection's canonical URL. This is the migration safety net for institutions moving from DSpace, CONTENTdm, Islandora, bepress, or any other repository system whose URLs are cited in published scholarship, embedded in LibGuides, indexed by search engines, and bookmarked by researchers — those URLs need to keep resolving after the move.
+
+This feature is **disabled by default** and must be explicitly enabled in two layers.
+
+## Two-layer feature gating
+
+The redirects feature is gated by **two** independent switches:
+
+1. **`Hyrax.config.redirects_enabled?`** — application-level config. Controls whether the schema and properties exist in this Hyrax application at all. Set in `config/initializers/hyrax.rb` or via the `HYRAX_REDIRECTS_ENABLED` environment variable. Default: `false`.
```
ShanaLMoore, line 11: 👍🏿

app/indexers/hyrax/indexers/redirects_indexer.rb
```diff
@@ -0,0 +1,31 @@
...
+    #
+    # @example
+    #   class WorkIndexer < Hyrax::Indexers::PcdmObjectIndexer
+    #     include Hyrax::Indexers::RedirectsIndexer if Hyrax.config.redirects_enabled?
+    #   end
+    module RedirectsIndexer
+      def to_solr(*args)
+        super.tap do |document|
+          next document unless Flipflop.redirects?
+          document['redirects_path_ssim'] = Array(resource.try(:redirects))
```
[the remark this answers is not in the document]
laritakr: # Suggested code
```
redirects = resource.try(:redirects)
redirects ? redirects.to_a : []
# nil          → []
# []           → []
# [a, b]       → [a, b]
```

# Current code
```
Array(resource.try(:redirects))
# nil          → []
# []           → []
# [a, b]       → [a, b]
```

Both cases would be correct, so the change isn't necessary.

app/services/hyrax/schema_loader.rb
```diff
@@ -105,13 +105,31 @@ def admin_only?
...
+                              Valkyrie::Types::Set.constructor(&Coerce)
+                            else
+                              Valkyrie::Types::Array.constructor(&Coerce)
+                            end
                           else
                             Identity
                           end
 
-        collection_type.of(type_for(config['type']))
+        collection_type.of(member_type)
```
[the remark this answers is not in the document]
laritakr: Verified: Hyrax::Redirect does respond to .meta (it's defined on Dry::Struct as a class method), so the NoMethodError case doesn't arise. `attributes_for` doesn't raise on a nested resource today.

The `multiple: false` case isn't a concern in practice because the nested-resource schema shipped in the metadata yaml files use `multiple: true`. To make that constraint explicit at the loader level, I added a validation and spec.

spec/services/hyrax/simple_schema_loader_spec.rb
```diff
@@ -145,5 +145,42 @@
       expect(permissive_schema.values.all? { |v| v.value.present? }).to be_truthy
       expect(permissive_schema[:sample_attribute].value).to eq("http://hyrax-example.com/sample_attribute")
     end
+
+    context 'when Hyrax.config.redirects_enabled? is true' do
+      before { allow(Hyrax.config).to receive(:redirects_enabled?).and_return(true) }
+
+      it 'includes the redirects predicate in the permissive schema' do
+        expect(permissive_schema.size).to eq(68)
+        expect(permissive_schema[:redirects].value).to eq('http://samvera.org/ns/hyku/redirects')
```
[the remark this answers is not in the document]
laritakr: Use of size was pre-existing, but is a brittle assertion. Removing it entirely in this location (but leaving it in locations outside the changes here). The presence-and-correct-predicate check (permissive_schema[:redirects].value) is the meaningful behavior.

app/indexers/hyrax/indexers/redirects_indexer.rb
orangewolf: I would think this should be the config var and not the flipper based on our discussion.

app/models/hyrax/pcdm_collection.rb
```diff
@@ -41,6 +41,8 @@ module Hyrax
   #
   class PcdmCollection < Hyrax::Resource
     include Hyrax::Schema(:core_metadata) if Hyrax.config.collection_include_metadata?
+    # See documentation/redirects.md for the redirects feature.
+    include Hyrax::Schema(:redirects) if Hyrax.config.collection_include_metadata? && Hyrax.config.redirects_enabled?
```
orangewolf, line 45: this looks right to me

app/indexers/hyrax/indexers/redirects_indexer.rb
orangewolf: I would also say that this line can return document if resource doesn't have a redirects property. If you do that you don't need the try on the line below

app/indexers/hyrax/indexers/pcdm_collection_indexer.rb
```diff
@@ -10,6 +10,7 @@ class PcdmCollectionIndexer < Hyrax::Indexers::ResourceIndexer
       include Hyrax::LocationIndexer
       include Hyrax::ThumbnailIndexer
       include Hyrax::Indexer(:core_metadata) if Hyrax.config.collection_include_metadata?
+      include Hyrax::Indexers::RedirectsIndexer if Hyrax.config.redirects_enabled?
```
orangewolf, line 13: 👍 its ok that this hits even if flexbile is on and no field is defined, since it will no op.

app/services/hyrax/simple_schema_loader.rb
```diff
@@ -49,7 +51,21 @@ def config_paths(schema_name)
...
       config_search_paths.each { |root_path| file_name_arr += Dir.entries(root_path.to_s + "/config/metadata/") }
-      file_name_arr.reject { |fn| !fn.include?('.yaml') }.uniq.map { |y| y.gsub('.yaml', '') }
+      file_name_arr
+        .reject { |fn| !fn.include?('.yaml') }
+        .uniq
+        .map { |y| y.gsub('.yaml', '') }
+        .reject { |schema_name| disabled_schemas.include?(schema_name) }
+    end
+
+    # Names of schemas in `config/metadata/` that the host app has not
```
orangewolf, line 61: I'm not sure how this is related to this pr?

orangewolf (changes requested):

app/services/hyrax/simple_schema_loader.rb
laritakr: This filter is what enforces the two-layer gating (config + feature flipper) for the redirects.yaml schema added in this pull request. 

The schema file ships in the gem unconditionally — it's present on disk whether or not the host app has set `Hyrax.config.redirects_enabled = true`. `SimpleSchemaLoader#metadata_files` walks `config/metadata/*.yaml` to build the schema list used by `permissive_schema_for_valkrie_adapter`. Without the filter, that map would include the redirects predicate for every Hyrax application — even ones that never opted in — which contradicts the "feature is fully absent when the config is off" behavior documented in `documentation/redirects.md`. Filtering at `metadata_files` is a single point that catches all paths that read from it.

orangewolf: got it, that makes senes to me then

app/indexers/hyrax/indexers/redirects_indexer.rb
laritakr: Config var gates the entire indexer inclusion. The flipper's primary use is to prevent indexing if the flipper is off. 

Added the return document & removed the try.

orangewolf (approved):

PR #7422, opened 2026-04-29
Add Hyrax::RedirectsController and catch-all redirect route

laritakr:
# Summary

Implements read-only redirect resolution.

When both gates are open (`Hyrax.config.redirects_enabled?` and `Flipflop.redirects?`), a catch-all route at the end of the engine routes dispatches any unmatched path to `Hyrax::RedirectsController#show`. The controller queries Solr against `redirects_path_ssim` for a record with that path, caches the result with a 60-second TTL, and 301-redirects to the permanent URL produced by Rails' `polymorphic_path` for the resolved work or collection.

When either gate is closed the route's constraint returns false, the route does not match, and Rails falls through to its standard 404. The config check short-circuits `Flipflop.redirects?` since the feature is unregistered when the config is off.

Refs notch8/palni_palci_knapsack#622.

### DIFF
--- a/app/controllers/hyrax/redirects_controller.rb
+++ b/app/controllers/hyrax/redirects_controller.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Hyrax
+  # See documentation/redirects.md for the redirects feature.
+  class RedirectsController < ApplicationController
+    CACHE_TTL = 60.seconds
+
+    def show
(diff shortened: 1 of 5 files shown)


laritakr:
Lint for this branch: https://github.com/samvera/hyrax/actions/runs/25183337606

orangewolf (dismissed):

orangewolf (approved):

PR #7425, opened 2026-05-01
624 aliases tab work form

laritakr:
## Summary
Refs #624, #625

Implements the user-facing UI for the URL Redirects feature on the work and collection edit forms. Builds on the validation, normalization, and ledger-sync infrastructure.

- **Aliases tab on the work edit form** (#624). Visible whenever the redirects feature is fully enabled (config + Flipflop). Renders existing redirect entries with a per-row remove control and a trailing blank row for adding new aliases on the next save.
- **Aliases tab on the collection edit form** (#625). Mirrors the work-form tab; reuses the `_form_redirects.html.erb` partial — no duplicate UI markup. Form-side wiring (populator, validator, normalizer, sync step) is shared via `Hyrax::Forms::ResourceForm`, so this slice adds only the UI surface.
- **Server-only flow** — no JavaScript dependency for adds (trailing blank row becomes the new entry on submit) or removes (inline `onclick` removes the `<tr>`; populator drops the missing row server-side).
- **UX**: input-group prefix shows the site root so users can see the full URL the path resolves from. Gray-X close button matches the Sharing tab convention.
- **Accessibility**: sr-only labels on every input, error region with `role="alert"` and `aria-live="polite"`, per-row remove buttons with the path included in the `aria-label`.

## Screenshots

Sample of tab:

Screenshot 2026-05-01 at 1 50 04 PM

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -44,6 +44,7 @@ module Hyrax
       property :version, virtual: true, prepopulator: LockKeyPrepopulator
 
       validates_with Hyrax::RedirectValidator, attributes: [:redirects] if Hyrax.config.redirects_enabled?
+      property :redirects_attributes, virtual: true, populator: :redirects_populator if Hyrax.config.redirects_enabled?
 
       ##
       # @api public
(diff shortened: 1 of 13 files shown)


laritakr:
Closing this PR. The redirects feature is being restructured into a five-PR stack to give each PR a tightly scoped purpose and decouple the data-model fix from the UI work. Spike testing surfaced a persistence bug: Set.of(Hyrax::Redirect) round-trips badly through Valkyrie's Postgres adapter — sub-fields get stripped on save and reads fail coercion.

The new stack (each PR builds on the previous):

1. **#7430 — basednear-form-hardening** (open): form-layer foundation. Makes BasedNearFieldBehavior compose via super, fixes the validates_with heritage-replay crash. Independent of redirects.
2. **redirects-data-model** (next): Hyrax::Redirect becomes a presenter; the parent attribute becomes Array.of(Hash); schema_loader gains a `hash` shortcut. Ships the data plumbing (RedirectPath ledger, normalizer, lookup, config gates, migration) with no UI.
3. **redirects-resolver**: RedirectsController, catch-all route, sync/remove transaction steps, indexer.
4. **redirects-form-and-validator**: aliases tab UI, RedirectsFieldBehavior (with hash-shape populator + presenter-hydrating prepopulator), RedirectValidator, helpers, locales.
5. **nested-field-behavior-pattern**: documentation/forms/field_behaviors.md.

The work in this PR isn't lost — most files come back largely unchanged in PRs 2-4, just regrouped and with the data-model fix folded in.

PR #7426, opened 2026-05-01
Extract Hyrax.config.redirects_active? helper

laritakr:
### Fixes

Refs #643

### Summary

Adds a config helper to consolidate duplication of the two redirects gating mechanisms, replacing use of the dual gating throughout the code.


laritakr:
lint run: https://github.com/samvera/hyrax/actions/runs/25233218787

PR #7427, opened 2026-05-01
Redirects integration fixes

laritakr:
### Fixes

Fixes various issues encountered while UI testing the redirects feature.
- fixes routing
- hide alias tab when object has no redirects property
- validate the redirects property correctly so m3 profile works
- strip trailing query from root path on redirects tab

PR #7428, opened 2026-05-02
Introduce composable Field Behavior pattern

laritakr:
# Summary

Introduces a new documented pattern for Field Behaviors that compose via super, and uses it to add a new Hyrax::RedirectsFieldBehavior providing support for the Aliases tab in the UI. This is gated behind the HYRAX_REDIRECTS_ENABLED flip, so the new behavior is a no-op when the flip is off and doesn't affect existing ResourceForms.

Rewires Hyrax::BasedNearFieldBehavior along with Hyrax::RedirectsFieldBehavior so the two compose via super.

Also fixes a prior bug where any subclass of ResourceForm crashed at load time once HYRAX_REDIRECTS_ENABLED=true. Reform's validates_with helper accidentally shares one options hash ({attributes: [:redirects]}) across every form that inherits from ResourceForm. The first time it runs, ActiveModel's validator empties that hash by deleting :attributes. The second time it runs — on a subclass — the hash is already empty, so the validator crashes with :attributes cannot be blank. The fix sidesteps the helper so a fresh hash is built each time, and the sharing problem goes away.

### DIFF
--- a/app/forms/concerns/hyrax/based_near_field_behavior.rb
+++ b/app/forms/concerns/hyrax/based_near_field_behavior.rb
@@ -1,18 +1,27 @@
 # frozen_string_literal: true
 module Hyrax
+  # Field Behavior for the `based_near` (location) controlled vocabulary form
+  # field. The form expects a ControlledVocabularies::Location object as input
+  # and produces a hash like those used with accepts_nested_attributes_for.
+  #
+  # See documentation/forms/field_behaviors.md for the pattern this module
+  # follows (why `deserialize` strips and calls `super`, why it's prepended
+  # in `inherited`, etc).
   module BasedNearFieldBehavior
-    # Provides compatibility with the behavior of the based_near (location) controlled vocabulary form field.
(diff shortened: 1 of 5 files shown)


laritakr:
Closing this PR. See #7425 and #7427 for the full restructuring rationale. The work in this PR is being split across the new stack:

- The composable Field Behavior pattern (BasedNearFieldBehavior super-chaining + validates_with → validation block fix) is in **#7430** — already open against main.
- `Hyrax::RedirectsFieldBehavior` (with the new hash-shape data model and a presenter-hydrating prepopulator) will be in the upcoming `redirects-form-and-validator` PR.
- `documentation/forms/field_behaviors.md` will be in the upcoming `nested-field-behavior-pattern` PR (last in the stack).

The data-model rewrite that surfaced during integration testing (Hyrax::Redirect as a presenter, Array.of(Hash) on the parent, schema_loader 'hash' shortcut) lands in the upcoming `redirects-data-model` PR.

No code is being lost — just regrouped under coherent themes.

PR #7436, opened 2026-05-07
Fix redirects form crash on non-flexible installs

laritakr:
## Summary
- Fix a crash on the work edit page when the redirects feature is enabled in a non-flexible install. The Aliases tab now renders cleanly instead of raising `NoMethodError`.

## Details
The redirects feature adds a per-work alias list, exposed on the work edit form under an Aliases tab. Without this fix, opening that tab in a non-flexible install (`HYRAX_FLEXIBLE=false`) crashes the page: the form partial reads `f.object.redirects`, but the form class never had the `redirects` property registered.

In flexible mode the m3 schema loader populates form properties from the metadata profile, so `redirects` is on the form and the partial works. In non-flexible mode that path doesn't run, and the existing redirects field behavior only registered the virtual `redirects_attributes` (used for nested form params) — never the persisted `redirects` property the partial actually reads.

The fix loads `redirects` onto the form from `config/metadata/redirects.yaml` via `Hyrax::FormFields(:redirects)`, the same mechanism every other YAML-backed form uses. The `redirects.yaml` form options were also corrected from `primary: false` (no effect on form display gating) to `display: false`, matching how the loader actually decides what to render.

## Testing
- [ ] On an install with `HYRAX_REDIRECTS_ENABLED=true` and `HYRAX_FLEXIBLE=false`, edit any work, click the **Aliases** tab — the tab renders without error and shows the redirects table.
- [ ] On the same install, add an alias path, save the work — the alias persists and reappears when the form is re-opened.
- [ ] Repeat the above with `HYRAX_FLEXIBLE=true` to confirm flexible mode still behaves correctly (no regression).
- [ ] With `HYRAX_REDIRECTS_ENABLED=false`, edit a work — the Aliases tab is not present and the edit page renders normally.

### DIFF
--- a/app/forms/concerns/hyrax/redirects_field_behavior.rb
+++ b/app/forms/concerns/hyrax/redirects_field_behavior.rb
@@ -30,6 +30,13 @@ module Hyrax
   module RedirectsFieldBehavior
     def self.included(descendant)
       return unless Hyrax.config.redirects_enabled?
+      # Load `redirects` (the persisted property the form partial reads
+      # via `f.object.redirects`) from `config/metadata/redirects.yaml`.
+      # In non-flexible mode this is the only thing that puts the
+      # property on the form. In flexible mode the m3 loader adds an
+      # equivalent definition on initialize; redefining via `property`
+      # is idempotent so this is a safe no-op there.
+      descendant.include Hyrax::FormFields(:redirects)
       descendant.property :redirects_attributes,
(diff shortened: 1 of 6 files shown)


ShanaLMoore:
Retested from Hyku with Flex true and false; this solution works now.

ShanaLMoore (approved):

PR #7443, opened 2026-05-12
Add editor_only visibility flag and use it for redirects display

laritakr:
## Summary
- Adds an `editor_only` flexible-metadata visibility flag so schemas can hide selected properties on show pages from users without edit ability for the record, and switches the redirects show-page display to use it (broadening visibility from admins-only to anyone with edit ability).
- Refs samvera/hyrax#656, notch8/palni_palci_knapsack#657

## Details

A property can now declare `editor_only: true` (alongside the existing `admin_only`) to restrict its show-page visibility. Editor visibility uses CanCan's `:edit` ability against the specific record being viewed, so admins (who have edit ability on everything) see editor-only fields, and per-record editors see them on records they can edit.

Each flag is an independent restrictor. When both are set on the same property, both must pass — the field renders only for a user who is admin *and* an editor of the record. The same semantics apply uniformly across the work, collection, and file_set show pages.

Restricted properties (with either flag) are excluded from Blacklight catalog registration entirely — no column header, no facet, not added to free-text search. This is a stronger guarantee than per-row gating: the data is not present in catalog responses for any user, including anonymous visitors.

A new `field_visible?` view helper centralizes the show-page gate, replacing duplicated inline checks. A previously hardcoded `admin_note` check in the FileSet metadata partial is removed; the helper now does the gating, and the field's schema can declare `editor_only: true` to opt in. (FileSet schemas don't currently declare `admin_note`, so no Hyrax-side schema change is needed; downstream apps that add `admin_note` to their own profiles can adopt the flag at their own pace.)

The redirects display (added in #7442) is switched from `admin_only` to `editor_only`, so per-record editors — not just site admins — can audit registered aliases on the show pages of works and collections they can edit.

While tightening the catalog-exclusion path, a pre-existing bug in `remove_old_properties!` was uncovered and fixed in its own commit: `_tesim` was being deleted from `facet_fields` instead of `index_fields`, so the intended cleanup was silently dropped. The fix and the new restricted-field eviction now share a single `remove_from_blacklight_config!` helper that covers all three Blacklight surfaces (facet, index, and the all_fields qf), with spec coverage demonstrating the qf cleanup across middle, first, and only-term positions.

Documentation for the visibility flags was previously missing entirely; this PR adds a "Property visibility flags" section to `documentation/flexible_metadata.md` covering both `admin_only` and `editor_only`, including the YAML and m3 declaration patterns and the catalog-exclusion behavior. `documentation/redirects.md` is updated to reference the new flag, with a cross-link to the general visibility-flags reference.

## Testing
- Confirm a logged-out visitor on a work, collection, or file_set show page does not see any property declared `editor_only: true`.
- Confirm an editor (non-admin user with edit ability for the record) sees the property on records they can edit, and does not see it on records they cannot edit.
- Confirm an admin sees properties declared `editor_only: true` everywhere, and also sees properties declared `admin_only: true`.
- Confirm a property declared with both `admin_only: true` and `editor_only: true` renders only when the user is both an admin and an editor of the record.
- With the redirects feature enabled, register an alias on a work, then visit the work's show page as a non-admin user with edit ability — confirm the registered aliases appear. Repeat as a logged-out visitor and confirm nothing appears. Repeat for a collection.
- Visit `/catalog` as an anonymous user and confirm that no property declared `admin_only` or `editor_only` appears as a column, facet, or search suggestion. Repeat as an admin and confirm the same — restricted fields stay out of catalog regardless of who is signed in.
- Verify the same behavior under both `HYRAX_FLEXIBLE=false` (top-level `editor_only:` key in the YAML schema) and `HYRAX_FLEXIBLE=true` (`editor_only` entry in the m3 profile property's `indexing:` array).

### DIFF
--- a/app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
+++ b/app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
@@ -23,7 +23,7 @@ module Hyrax
           # to prevent them from being exposed in catalog search results.
           # They remain available on show pages, based on visibility.
           if restricted_field?(indexing)
-            blacklight_config.facet_fields.delete("#{itemprop}_sim")
+            remove_from_blacklight_config!(itemprop)
             next
           end
 
@@ -139,16 +139,22 @@ module Hyrax
 
       def remove_old_properties!(previous_properties, current_properties)
         props = previous_properties - current_properties
-        return if props.empty?
+        props.each { |prop| remove_from_blacklight_config!(prop) }
+      end
 
-        props.each do |prop|
-          # remove from facet field
-          blacklight_config.facet_fields.delete("#{prop}_sim")
-          # remove from index field
-          blacklight_config.index_fields.delete("#{prop}_tesim")
-          # remove from qf
-          blacklight_config.search_fields['all_fields'].solr_parameters[:qf].slice!("#{prop}_tesim")
-        end
+      # Evict every Blacklight registration for `itemprop`.
+      def remove_from_blacklight_config!(itemprop)
+        blacklight_config.facet_fields.delete("#{itemprop}_sim")
+
+        name = blacklight_config.index_fields.keys.detect { |key| key.start_with?(itemprop) }
+        name ||= "#{itemprop}_tesim"
+        blacklight_config.index_fields.delete(name)
+
+        qf = blacklight_config.search_fields['all_fields']&.solr_parameters&.dig(:qf)
+        return if qf.nil?
+
+        qf.slice!(" #{name}")
+        qf.slice!(name)
       end
     end
 
--- a/app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
+++ b/app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
@@ -19,12 +19,16 @@ module Hyrax
           indexing = prop['indexing']
           next if indexing.nil?
 
+          # prevents all restricted fields from being added to blacklight config
+          # to prevent them from being exposed in catalog search results.
+          # They remain available on show pages, based on visibility.
+          if restricted_field?(indexing)
+            blacklight_config.facet_fields.delete("#{itemprop}_sim")
+            next
+          end
+
           if stored_searchable?(indexing, itemprop)
             index_args = { itemprop:, label: }
-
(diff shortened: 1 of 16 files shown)


app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
```diff
@@ -19,13 +19,17 @@ def load_flexible_schema
           indexing = prop['indexing']
           next if indexing.nil?
 
+          # prevents all restricted fields from being added to blacklight config
+          # to prevent them from being exposed in catalog search results.
+          # They remain available on show pages, based on visibility.
+          if restricted_field?(indexing)
+            blacklight_config.facet_fields.delete("#{itemprop}_sim")
```
orangewolf: needs to check for more cases (ssim for example)

orangewolf (changes requested):

app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
```diff
@@ -134,16 +139,22 @@ def facetable?(indexing, itemprop)
...
+
+        name = blacklight_config.index_fields.keys.detect { |key| key.start_with?(itemprop) }
+        name ||= "#{itemprop}_tesim"
+        blacklight_config.index_fields.delete(name)
+
+        qf = blacklight_config.search_fields['all_fields']&.solr_parameters&.dig(:qf)
+        return if qf.nil?
+
+        qf.slice!(" #{name}")
+        qf.slice!(name)
```
orangewolf:
```suggestion
        blacklight_config.facet_fields.each do |field|
          blacklight_config.facet_fields.delete(field) if field.start_with?(item prop)  
        end

        names = blacklight_config.index_fields.keys.select do |field|
          blacklight_config.index_fields.delete(field) if field.start_with?(item_prop)
        end
        
        qf = blacklight_config.search_fields['all_fields']&.solr_parameters&.dig(:qf)
        return if qf.nil?
        names.each do |name|
          qf.slice!(" #{name}")
          qf.slice!(name)
        end
```

orangewolf (dismissed):

orangewolf (approved):

PR #7444, opened 2026-05-12
Add copy permalink button to show pages

laritakr:
## Summary
- Adds a "Copy permalink" button to work and collection show pages (behind a feature flipper) that copies the record's canonical UUID-based URL to the clipboard, so users have a stable link to share or cite when the browser address bar shows an alias path.
- Refs notch8/palni_palci_knapsack#651

### DIFF
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -1,4 +1,7 @@
 <% provide :page_title, @presenter.page_title %>
+<% content_for :canonical_url do %>
+  <link rel="canonical" href="<%= permalink_for(@presenter) %>" />
+<% end %>
 
 <%= render 'shared/citations' %>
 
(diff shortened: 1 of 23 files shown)


app/views/hyrax/base/show.html.erb
```diff
@@ -1,4 +1,7 @@
 <% provide :page_title, @presenter.page_title %>
+<% content_for :canonical_url do %>
+  <link rel="canonical" href="<%= permalink_for(@presenter) %>" />
```
orangewolf, line 3: https://claude.ai/share/46d0d190-6974-4fc6-a4ef-5b475bb75018 I agree with Claude - this should be display_link

orangewolf (commented):
this PR looks good except for the permalink vs display link for the link rel. I'm putting that as just a comment so someone can override me and a approve if you want.

app/views/hyrax/base/show.html.erb
laritakr: Possibly, but we don't have display_url yet. That work isn't yet done, so the change to use it would have to come along with that work.

ShanaLMoore (approved):

PR #7446, opened 2026-05-14
Hide attribute rows when all values are blank

laritakr:
## Summary
- Attribute rows on show pages no longer render an empty row when every value is blank.
- Single-value flexible-schema attributes now coerce blank strings to `nil`, matching multi-value behavior.

## Details
- `Hyrax::Renderers::AttributeRenderer#render` / `#render_dl_row` previously short-circuited only when `values` was Ruby-blank. An array of all-blank strings rendered empty `<dt>`/`<dd>` (or `<tr>`/`<td>`) rows. A new `blank_values?` predicate treats an all-blank array the same as an empty one.
- The schema loader's single-value branch (`Identity.of`) was a no-op, so blank strings survived round-trips. It now applies a `CoerceSingular` constructor that strips the dry-types `Undefined` placeholder and coerces blank strings to `nil`. Booleans, numbers, and other non-string types are unchanged.
- Single-value and multi-value attributes now apply equivalent blank-cleanup at the type boundary.

### DIFF
--- a/app/renderers/hyrax/renderers/attribute_renderer.rb
+++ b/app/renderers/hyrax/renderers/attribute_renderer.rb
@@ -109,7 +109,6 @@ module Hyrax
       end
 
       def blank_values?
-        return true if values.blank?
         Array(values).all?(&:blank?)
       end
     end
--- a/app/renderers/hyrax/renderers/attribute_renderer.rb
+++ b/app/renderers/hyrax/renderers/attribute_renderer.rb
@@ -27,7 +27,7 @@ module Hyrax
 
       # Draw the table row for the attribute
       def render
-        return '' if values.blank? && !options[:include_empty]
+        return '' if blank_values? && !options[:include_empty]
 
         markup = %(<tr><th>#{label}</th>\n<td><ul class='tabular'>)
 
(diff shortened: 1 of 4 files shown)


app/services/hyrax/schema_loader.rb
```diff
@@ -135,19 +135,20 @@ def multiple?
...
+      # Booleans, numbers, and other types pass through unchanged.
       #
       # @example
-      #   Identity.of(Valkyrie::Types::String) # => Valkyrie::Types::String
+      #   Identity.of(Valkyrie::Types::String) # => Valkyrie::Types::String with blank-string → nil coercion
       class Identity
-        ##
-        # @param [Dry::Types::Type]
-        # @return [Dry::Types::Type] the type passed in
+        CoerceSingular = lambda do |value|
```
orangewolf: I’m pretty sure you don’t need a lambda for this. It can just be implemented directly in the of method

app/renderers/hyrax/renderers/attribute_renderer.rb
```diff
@@ -107,6 +107,11 @@ def li_value(value)
       def work_type_label_key
         options[:work_type] ? options[:work_type].underscore : nil
       end
+
+      def blank_values?
+        return true if values.blank?
```
orangewolf: The first line doesn’t cover any cases that are not already covered by the second line.

orangewolf (changes requested):

orangewolf (approved):

PR #7451, opened 2026-05-17
Display URL flag for redirect aliases

laritakr:
## Summary
Lets institutions designate one alias as the **display URL** for a work or collection — visitors who follow any alias for that record now land at the display alias path, not the UUID URL. Refs notch8/palni_palci_knapsack#626 and notch8/palni_palci_knapsack#653.

## Details
- Adds a "Display URL" radio button column on the Aliases form. Selecting a radio designates that alias as the record's display URL; only one per record.
- Schema rename: `canonical` boolean → `display` boolean on each redirect entry; unused `sequence` field dropped.
- `hyrax_redirect_paths` table restructured: `path` column renamed to `source_path`; new `target_path` column stores where each alias resolves to (display alias's path, or UUID path when no display is set); resolver now reads from DB directly instead of Solr.
- Reversible Rails migration handles the column rename and a backfill of `target_path` for existing rows; runs inline so post-`db:migrate` state is consistent.
- Locale strings updated in all 7 languages.
- Documentation refreshed: `redirects.md` migration playbook now uses the new column names; `field_behaviors.md` exemplar matches the current `RedirectsFieldBehavior` source.

## Testing
- Edit a work, click the **Aliases** tab. Add two or more aliases. Select one as the display via the radio button. Save.
- Visit each alias path in the browser. Confirm a 301 redirect to the display alias's path.
- Unselect the display flag (remove the display alias or use the form to clear it). Save. Re-visit any remaining alias. Confirm a 301 redirect to the UUID path (`/concern/<type>/<uuid>` or `/collections/<uuid>`).
- Repeat on a Collection.
- For installations using Flexible Metadata, verify the m3 profile's `redirects` property still validates.


laritakr:
Closing in favor of alternate plan, retaining a single path and a display_url boolean.  The simpler shape avoids the self-referential row this design produced for the display alias.

PR #7454, opened 2026-05-18
Only run one workflow at a time per PR

dlpierce:
Adds concurrency group and cancel-in-progress to the CI workflow. The intent is if a PR or branch is updated with new commits then any specs running will be cancelled.


laritakr (approved):
I believe the fix for the prior spec failure is already in main, so I pulled in main. Assuming it is green when it finishes the build, this should be good to go. If not, we need to re-evaluate and try another spec fix.

PR #7455, opened 2026-05-18
Fix redirects form property registration in non-flexible mode

laritakr:
## Summary

In non-flexible mode the `redirects` property is now included on `Hyrax::Forms::ResourceForm` at the class level (gated on `redirects_enabled?` and `!Hyrax.config.flexible?`), replacing a per-instance install on the form's singleton class that silently failed to register the property in Reform's `definitions` registry. Without the property registered there, saved entries never reach the resource.

Flexible mode is unaffected by the code change — it uses the m3 schema loader's singleton-class `schema_definitions` path inside `initialize`, which the change does not touch. Documentation is updated to flag that the m3 profile snippet requires a `form:` block; without it the m3 form-definition loader filters the property out and the Aliases tab errors when the partial reads `f.object.redirects`.

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -28,6 +28,7 @@ module Hyrax
 
       include BasedNearFieldBehavior
       include RedirectsFieldBehavior
+      include Hyrax::FormFields(:redirects) if Hyrax.config.redirects_enabled? && !Hyrax.config.flexible?
       class_attribute :model_class
 
       property :human_readable_type, writable: false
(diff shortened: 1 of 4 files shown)


app/forms/hyrax/forms/resource_form.rb
```diff
@@ -28,6 +28,7 @@ class ResourceForm < Hyrax::ChangeSet # rubocop:disable Metrics/ClassLength
 
       include BasedNearFieldBehavior
       include RedirectsFieldBehavior
+      include Hyrax::FormFields(:redirects) if Hyrax.config.redirects_enabled? && !Hyrax.config.flexible?
```
orangewolf: This should probably be work_include_metadata or however one spells thst

orangewolf (approved):

Issue #653, opened 2017-03-24
Dashboard partial doesn't work in non-hyrax controller


jcoyne:
In hyku we use the dashboard layout on a non-Hyrax controller and get this exception:
```
     ActionView::Template::Error:
       No route matches {:action=>"index", :controller=>"hyrax/my/collections", :locale=>nil}
```

PR #7456, opened 2026-05-18
Add display_url column for redirect aliases

laritakr:
## Summary
Adds a `display_url` boolean column to `hyrax_redirect_paths` and renames the per-entry 'canonical`/`sequence` fields to a single `display_url`. Subsequent PRs will add the form UI (radio column) and the resolver wiring that uses the flag.
- Refs Pitt #653.

## Details

Note: we are changing some behavior in ways that are not backward compatible. Since the feature is new and in active development, I did not worry about backward compatibility in the changes. This was intentional. 

Changes in this PR:
- Schema-only foundation. The column is added and the persistence layer carries it, but nothing in this PR can set the flag — the form UI and any import path land in follow-up PRs. Every row this code writes today persists `display_url: false`.
- One alias per record will eventually be flagged as the institution's preferred display URL; the rest are plain redirects with no ordering semantics. This PR replaces the prior `canonical`/`sequence` shape with that simpler model.

## Note

I have been weighing whether to add a suggested source_path / target_path pair, or drop target_path and have the controller compute where to send a visitor at request time.

That schema mixes a per-row fact ("what alias points to what resource") with a per-resource fact ("which alias is the institution's preferred URL"). It forces a rewrite of every alias row on the record whenever the display flag moves, and produces self-referential rows where source_path == target_path on the display alias itself — which causes an infinite redirect loop unless the controller special-cases it.

It is my opinion that having to rewrite every row whenever the display flag moves, plus the special-case the controller needs for self-referential rows outweighs the savings of a single-query lookup.

### DIFF
--- a/.dassie/db/schema.rb
+++ b/.dassie/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2026_04_30_000001) do
+ActiveRecord::Schema.define(version: 2026_05_18_000001) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -313,6 +313,7 @@ ActiveRecord::Schema.define(version: 2026_04_30_000001) do
     t.string "resource_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "display_url", default: false, null: false
     t.index ["path"], name: "index_hyrax_redirect_paths_on_path", unique: true
     t.index ["resource_id"], name: "index_hyrax_redirect_paths_on_resource_id"
   end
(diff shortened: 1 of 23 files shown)


.dassie/db/schema.rb
```diff
@@ -313,6 +313,7 @@
     t.string "resource_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "display_url", default: false, null: false
```
orangewolf, line 316: I do not understand why this is a boolean. why not store where you are trying to go instead of forcing an extra lookup on every request?

orangewolf (changes requested):
I'm concerned about both maintainability and performance with this approach. Let's discuss

.dassie/db/schema.rb
laritakr: We may be able to get there more cleanly now. I know you originally asked for that structure, but I ran into some cyclical redirect issues that were getting very messy to resolve. Now that Rack is in place and a few other issues have gotten resolved, perhaps the work I've done since will make it more straightforward. I'm looking into it to see what I can come up with.

laritakr:
Went a different direction.

PR #7457, opened 2026-05-19
Add Display URL radio column to redirects form

laritakr:
## Summary
- Adds a Display URL radio column to the Aliases form with a "None" option, plus an "Add another alias" button so users can enter multiple aliases before saving.
- Refs samvera/hyrax#7451 (replaces the prior approach), Pitt #653.

## Details
- Stacks on top of the schema-rename PR. The `display_url` column added there now has a UI.
- Enforces "at most one per record" rule server-side by the existing redirect validator, so Bulkrax imports and direct console writes follow the same constraint.
- The redirects resolver is unchanged in this PR — visiting any alias still 301s to the record's UUID URL. 
- Existing inline and new javascript moved to `app/assets/javascripts/hyrax/redirects.js`.

### DIFF
--- a/app/assets/javascripts/hyrax.js
+++ b/app/assets/javascripts/hyrax.js
@@ -89,6 +89,7 @@
 //= require hyrax/turbolinks_events
 //= require hyrax/i18n_helper
 //= require hyrax/collapse
+//= require hyrax/redirects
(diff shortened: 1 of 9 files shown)


laritakr:
Reimplemented over a different schema pattern elsewhere.

PR #7458, opened 2026-05-19
Resolve aliases via Rack middleware

laritakr:
## Summary
- Replaces the redirect resolver with a Rack middleware backed by the Postgres `hyrax_redirect_paths` ledger. When the visited path is a record's display URL, the show page renders in place at that path; every other alias and the record's UUID URL 301-redirect to the display URL. Show pages emit a `<link rel="canonical">` pointing at the display URL.
- Refs Pitt #644, #626, and #653.

## Details
- `Hyrax::RedirectsController`, the install generator's catch-all `get '*alias_path'` route, and the `redirects_path_ssim` Solr field are removed. The `redirects_path_tesim` field that powers the show-page display is unchanged.
- 301 responses include `Cache-Control: no-cache` so a stale browser-cached redirect from a prior display-URL configuration is revalidated on every visit.
- Turbolinks compatibility: the middleware sets `Turbolinks-Location` on 301s, and the show controllers stamp it on responses that follow a Turbolinks XHR redirect chain, so the browser address bar reflects the resolved URL rather than the originally-requested alias.
- The `<link rel="canonical">` tag in the work and collection show layouts now return the record's display alias URL when one is set and falls back to the UUID-based permalink otherwise. The Copy Permalink button continues to expose the UUID URL — it's the only address guaranteed to never change.

## Notes

Out of the box Hyrax uses a path-only cache key (correct for single-tenant installs), and Hyku will need a one-time initializer file like the one below to fold tenant identity into the cache key. 

```
# In Hyku: config/initializers/hyku_redirects_middleware_decorator.rb
module Hyku
  module RedirectsMiddlewareDecorator
    def cache_key_for(path)
      [Apartment::Tenant.current, super].join('/')
    end
  end
end

Hyrax::Redirects::Middleware.singleton_class.prepend(Hyku::RedirectsMiddlewareDecorator)
```

- Treated as notes-minor because this is a new feature, so the removal of the routes has no released version to break.


lib/hyrax/engine.rb
```diff
@@ -104,6 +104,11 @@ class Engine < ::Rails::Engine # rubocop:disable Metrics/ClassLength
       require 'hyrax/rails/routes'
     end
 
+    initializer 'hyrax.redirects.middleware' do |app|
+      require 'hyrax/redirects/middleware'
+      app.middleware.use Hyrax::Redirects::Middleware
+    end
+
```
dlpierce, line 111: Can this middleware setup be conditional on `Hyrax.config.redirects_enabled?`

PR #7465, opened 2026-05-21
Redirects schema changes

laritakr:
## Summary
Reshape the `hyrax_redirect_paths` table to add `to_path`, `permalink_path`, and `is_display_url`, and rename `path` to `from_path`. Existing alias behavior is unchanged for users.
Refs notch8/palni_palci_knapsack#653

## Details
- New migration drops and recreates the table with the new column shape. Drop & recreate:
  - allows skipping migrating data into the new columns that don't allow NULL
  - is acceptable because the feature is unreleased and behind both a config flag and a Flipflop.
- Remaining functionality is modified to work with the new table structure.

### DIFF
--- a/lib/hyrax/transactions/steps/sync_redirect_paths.rb
+++ b/lib/hyrax/transactions/steps/sync_redirect_paths.rb
@@ -6,8 +6,8 @@ module Hyrax
     module Steps
       # A `dry-transaction` step that mirrors a saved resource's `redirects`
       # entries into the `hyrax_redirect_paths` redirects table. The unique
-      # index on `path` enforces global uniqueness at the DB level — if a
-      # concurrent save already claimed a path, the insert raises
+      # index on `from_path` enforces global uniqueness at the DB level — if
+      # a concurrent save already claimed a path, the insert raises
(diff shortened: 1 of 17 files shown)


lib/hyrax/transactions/steps/sync_redirect_paths.rb
```diff
@@ -40,27 +40,36 @@ def syncable?(object)
...
-                  .reject(&:blank?)
-                  .uniq
+          permalink = Hyrax::PermalinkPath.call(object)
+          resource_id = object.id.to_s
           now = Time.current
-          paths.map { |path| { path: path, resource_id: object.id.to_s, created_at: now, updated_at: now } }
+          alias_paths_for(object).map do |alias_path|
+            { from_path: alias_path,
+              to_path: permalink,
+              permalink_path: permalink,
```
orangewolf, line 49: in my version I had display_path as the to_path very time and permalink needed to be the permanent_path. here they match. is that intentional?

orangewolf: I see, its in PR #7466

orangewolf (approved):

PR #7470, opened 2026-05-21
Strip default locale from default_url_options

ShanaLMoore:
## Summary

`Hyrax::Controller#default_url_options` currently appends `locale: I18n.locale` to every URL the app generates, even when the active locale is the default. The result is that sitemap entries, canonical tags, and outbound links all carry `?locale=en` (or whichever locale is the default), and search engines index those locale-parameterized URLs as canonical.

This was originally reported downstream as a Hyku Up issue where Google was returning Italian/French/Spanish/Portuguese URLs as top results for English queries on a Hyku tenant whose default locale was English. The polluted sitemap was the discovery signal that drove Googlebot to index the locale-parameterized variants in the first place.

This PR makes the locale inclusion conditional: it is only added when the current locale differs from `I18n.default_locale`. That preserves the original intent of the method (persist a user's non-default locale across requests via route helpers) while keeping default-locale URLs clean.

Refs: notch8/hykuup_knapsack#677

## Behavior

| Active locale | Before | After |
| --- | --- | --- |
| `I18n.default_locale` | `{ locale: :en }` | `{}` |
| non-default (e.g. `:es`) | `{ locale: :es }` | `{ locale: :es }` |

## Test plan

- [ ] CI runs the new `spec/controllers/concerns/hyrax/controller_spec.rb` covering both branches
- [ ] Existing controller/feature specs still pass
- [ ] Manual check on a downstream Hyku app: hit `/sitemap` and confirm work URLs no longer contain `?locale=en`

### DIFF
--- a/app/assets/javascripts/hyrax/select_work_type.es6
+++ b/app/assets/javascripts/hyrax/select_work_type.es6
@@ -42,9 +42,12 @@ export default class SelectWorkType {
       let url = this.form.find('input[type="radio"]:checked').data(this.type)
       const select = this.form.find('select')
 
-      // Only append admin_set_id if the select exists and has a value
+      // Only append admin_set_id if the select exists and has a value.
+      // Pick the right separator: URLs lose their locale query string when
+      // I18n.locale == I18n.default_locale, so we cannot assume `?` is present.
       if (select.length && select.val()) {
-          url = url + "&admin_set_id=" + select.val()
+          const separator = url.indexOf('?') === -1 ? '?' : '&'
+          url = url + separator + "admin_set_id=" + select.val()
       }
       return url
   }
--- a/app/controllers/concerns/hyrax/controller.rb
+++ b/app/controllers/concerns/hyrax/controller.rb
@@ -27,9 +27,15 @@ module Hyrax::Controller
     hyrax.dashboard_path
   end
 
-  # Ensure that the locale choice is persistent across requests
+  # Ensure that the locale choice is persistent across requests, but only
+  # when it differs from the default. Appending +?locale=+ for the default
+  # locale on every URL pollutes sitemaps, canonical tags, and outbound
+  # links, which causes search engines to index and rank duplicate
+  # locale-parameterized URLs as canonical.
   def default_url_options
-    super.merge(locale: I18n.locale)
+    opts = super
+    opts[:locale] = I18n.locale unless I18n.locale == I18n.default_locale
+    opts
   end
 
   # @note for Blacklight 6/7 compatibility
--- a/app/presenters/hyrax/admin_set_presenter.rb
+++ b/app/presenters/hyrax/admin_set_presenter.rb
@@ -51,7 +51,7 @@ module Hyrax
     end
 
     def show_path
-      Hyrax::Engine.routes.url_helpers.admin_admin_set_path(id, locale: I18n.locale)
+      Hyrax::Engine.routes.url_helpers.admin_admin_set_path(id, locale: (I18n.locale unless I18n.locale == I18n.default_locale))
     end
 
     def available_parent_collections(*)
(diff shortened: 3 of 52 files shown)


ShanaLMoore:
@laritakr heads up on this one - your recent redirects/permalink work (#7449 "Strip query string and fragment from permalink URL", #7458, etc.) is in the same neighborhood, since `default_url_options` is what was *adding* the `?locale=` that `permalink_for` then has to strip.

This change makes `default_url_options` not append `:locale` at all when it equals `I18n.default_locale`, so default-locale URLs come out clean at the source. `permalink_for`'s `strip_query_string` still does the right thing as a belt-and-suspenders for non-default locales and any other registered `default_url_options`, so nothing here should disturb your redirects work.

Driver: notch8/hykuup_knapsack#677 (Google was indexing Hyku tenant pages under wrong languages because the sitemap emitted `?locale=en` on every URL).

Wanted to flag in case anything you're touching expects `default_url_options` to keep returning `{ locale: ... }` unconditionally.

ShanaLMoore:
CI is red, but the failures are all expected consequences of this change. Triaging now and pushing a spec sweep.

The pattern: many existing specs assert that URLs contain `?locale=en` (the default locale), because the old `default_url_options` always appended it. With this change, default-locale URLs come out without the param, so assertions like

```ruby
expect(response).to redirect_to(hyrax.edit_admin_admin_set_path(..., locale: 'en', anchor: 'participants'))
# got: .../edit#participants  (no ?locale=en)
```

now fail. These specs were correctly testing the old behavior; they need updating to drop `locale: 'en'` from expectations, since the new (correct) behavior is "don't pollute default-locale URLs."

I am spinning up Dassie locally to sweep these, will push an update once green. Will leave the PR in draft-equivalent state mentally until then.

ShanaLMoore:
**EDIT (2026-05-22):** the "pre-existing flakes" claim in the original comment below was wrong. Those failures *were* caused by this PR, through a hidden coupling I missed on the first read:

`app/assets/javascripts/hyrax/select_work_type.es6` concatenated `"&admin_set_id=" + select.val()` onto the new-work URL. That worked only because the old `default_url_options` always appended `?locale=en`, so every URL already had a query string. Once the default-locale strip removed that, the concatenation produced `/concern/generic_works/new&admin_set_id=...` and the browser routed to a 404. The new-work form never rendered, so the `"Files"` tab Capybara was looking for did not exist.

Fixed in 0ec9f80f1 with a proper `?`/`&` separator picker, mirroring the same pattern LaRita used in `Hyrax::RedirectsController#with_locale` from the redirects merge.

The branch has also been rebased onto current `main` (was 21 commits behind after the redirects stack landed); the only spec conflict was in `spec/helpers/hyrax/permalink_helper_spec.rb`, resolved by taking `main`'s version (it rewrote `permalink_for` to use `Hyrax::PermalinkPath.call` + `strip_query_string`, and its tests already cover locale stripping correctly).

All 16 shards green on run 26297622041.

---

_[Original comment preserved below for context; the "pre-existing flakes" portion is the part that turned out to be wrong.]_

Sweep done. The PR now contains:

1. **Production fix:** `Hyrax::Controller#default_url_options` only appends `:locale` when the current locale differs from `I18n.default_locale`. Four presenters (`admin_set_presenter`, `collection_presenter`, `work_show_presenter`, `file_set_presenter`) and two renderers (`linked_attribute_renderer`, `faceted_attribute_renderer`) were also passing `locale: I18n.locale` explicitly to URL helpers; they now use the same conditional so they don't reintroduce `?locale=en`.

2. **Spec sweep:** 42 spec files updated to drop `locale: 'en'` / `locale: I18n.locale` / `?locale=en` from URL expectations. Validated locally in Dassie (454 examples, 0 failures across the previously red areas).

orangewolf (approved):

ShanaLMoore:
cc @orangewolf I think we need to revert this change. LaRita is pointing out that it may be causing other issues: [here](https://notch8.slack.com/archives/C087FQ5PG22/p1779829482672769)

image

PR #7476, opened 2026-05-26
Revert #7470 (default_url_options strip) -- exposes locale-leak regression

ShanaLMoore:
## Summary

Reverts samvera/hyrax#7470. That PR (merged earlier today) changed `Hyrax::Controller#default_url_options` to only emit `locale:` when the active locale differs from `I18n.default_locale`. After deploy to koppie (`https://pg.nurax.samvera.org`), Hyrax began rendering pages in random non-default locales for users who had never requested those locales.

The regression is real and user-visible. The fix in #7470 is correct *in concept* but uncovered a latent cross-request `I18n.locale` leak in Hyrax that the previous always-append behavior was inadvertently masking. The leak deserves its own diagnosis and fix; this PR restores the pre-#7470 behavior so staging/production aren't degraded while that happens.

## Reproduction (live on pg.nurax)

Six consecutive cookie-less, Accept-Language-less curls all return a sign-in form whose action carries a non-default locale:

```bash
for i in 1 2 3 4 5 6; do
  curl -s https://pg.nurax.samvera.org/users/sign_in \
    | grep -o 'action=\"/users/sign_in[^\"]*\"' | head -1
done
```

Output (during my session):

```
action=\"/users/sign_in?locale=it\"
action=\"/users/sign_in?locale=it\"
action=\"/users/sign_in?locale=it\"
action=\"/users/sign_in?locale=it\"
action=\"/users/sign_in?locale=it\"
action=\"/users/sign_in?locale=it\"
```

LaRita and I also saw it interactively: navigating from catalog search to the dashboard randomly flips the UI to Portuguese, French, Italian, German etc., even though the user only browses in English and never picked a non-English locale.

## Why #7470's symptom is the leak being unmasked

Pre-#7470, `default_url_options` unconditionally merged `locale: I18n.locale`. Every link, redirect, and form action carried `?locale=X`. Every subsequent request hit `Hyrax::Controller#set_locale`:

```ruby
def set_locale
  I18n.locale = params[:locale] || I18n.default_locale
end
```

with a non-nil `params[:locale]`, so `I18n.locale` got *forcibly pinned* to a known value at the start of every request. Any other code path that mutated `I18n.locale` or `I18n.default_locale` was effectively overwritten on the next request.

Post-#7470, default-locale users navigate URLs with no `?locale=` param. `set_locale` falls through to `I18n.default_locale`. If anything in the request lifecycle is mutating `I18n.locale` (or `I18n.default_locale`) and the value persists across requests on the same Puma worker, the next visitor inherits that stale value. The deterministic per-worker stickiness in the curl repro above is consistent with this.

I have not yet identified the exact mutator. Candidates worth checking:
- `app/models/user_mailbox.rb` -- uses `I18n.t(..., locale: preferred_locale)` (kwarg form should be safe but worth ruling out)
- `app/channels/hyrax/notifications_channel.rb` -- writes `current_user.update(preferred_locale: data['locale'])`
- `app/controllers/concerns/hyrax/works_controller_behavior.rb` -- has its own `locale = params[:locale] || current_user&.preferred_locale || I18n.default_locale` branch
- presenters and renderers in `app/presenters/` / `app/renderers/` that read `I18n.locale` at URL-build time
- anything setting `I18n.default_locale =` (vs `I18n.locale =`) anywhere in the boot/initializer path

## Refs

- Original fix: samvera/hyrax#7470
- Downstream Hyku PR being held: samvera/hyku#3071 (will be marked draft)
- Downstream Hyku issue: notch8/hykuup_knapsack#677 (original Google-indexing issue)
- Follow-up locale-leak issue: will be filed in this repo with the repro above

## Plan to re-land

1. This revert merges and pg.nurax goes back to normal (random language flips stop).
2. New issue filed for the I18n.locale cross-request leak with the curl repro.
3. Once the leak is fixed (independent PR), re-open the equivalent of #7470 layered on top.
4. Resume samvera/hyku#3071.

## Test plan

- [ ] CI green on this branch.
- [ ] After merge + deploy to pg.nurax, the curl loop above returns 6 forms with `action=\"/users/sign_in?locale=en\"` (or other stable locale) instead of 6 differing locales.


ShanaLMoore:
Filed samvera/hyrax#7477 with the cross-request locale-leak repro and candidates list. Once that lands, this revert can be reversed.

orangewolf (approved):

PR #7479, opened 2026-06-04
Add compound (hierarchical) metadata foundation

laritakr:
## Summary
Introduces compound (hierarchical) metadata — a repeatable list of entries, each a set of named sub-properties — defined entirely in schema YAML with no per-field code, on works and collections.

## Details
- A compound is a `type: hash, multiple: true` parent property; its members are flat top-level properties pointing back with `subproperty_of: <parent>`. The generic form, populator, indexer, and show renderer all read the declaration from the schema.
- Sub-property types: open text, `controlled` (inline `values:` or a QA `authority:`), auto-linked `url`, and `work_or_url` (search readable internal works or collections, or enter an external URL).
- Per-sub-property Solr indexing (`index_keys:`/`indexing:`) and show display (`display:`); optional `group:` clustering and `form: { cols:, as: }` layout.
- Compounds render inline or as a card (`view: { display: card }`), and sit in the form's primary or "Additional fields" section per `form: { primary: }`.
- Sub-properties (and whole compounds) can be marked `required:`; on save the form blocks with one error per compound, and required fields show a `*`.
- Works identically under `HYRAX_FLEXIBLE=false` and `true`; an m3 profile validator catches a misconfigured compound on save.
- Ships sample compounds (participants, identifiers, compound_rights, relationships) on works and collections; an app removes or overrides them in its own schema.
- `Hyrax::CompoundFieldBehavior` is the generic, schema-driven member of the same Field Behavior family as the hand-written `based_near`/`redirects` behaviors; the docs explain when to use each.

## Screenshots

<details>
<summary>Work & forms with compound metadata</summary>

Screenshot 2026-06-04 at 12 10 29 PM

Screenshot 2026-06-04 at 12 10 44 PM

Screenshot 2026-06-04 at 4 37 39 PM

Screenshot 2026-06-04 at 4 37 19 PM

</details>

<details>
<summary>Collection show page with compound metadata</summary>

Screenshot 2026-06-03 at 5 44 35 PM

Screenshot 2026-06-04 at 4 37 03 PM

</details>

### DIFF
--- a/.koppie/config/metadata_profiles/m3_profile.yaml
+++ b/.koppie/config/metadata_profiles/m3_profile.yaml
@@ -943,7 +943,7 @@ properties:
     form:
       primary: false
     property_uri: http://id.loc.gov/vocabulary/relators/asn
-    subfields:
+    subproperties:
       title: { type: string }
       agent_name:
         type: string
@@ -972,7 +972,7 @@ properties:
     form:
       primary: false
     property_uri: http://purl.org/dc/terms/identifier
-    subfields:
+    subproperties:
       identifier:
         type: string
         required: true
@@ -998,7 +998,7 @@ properties:
     form:
       primary: false
     property_uri: http://purl.org/dc/terms/rights
-    subfields:
+    subproperties:
       rights_statement:
         type: controlled
         authority: rights_statements
@@ -1026,7 +1026,7 @@ properties:
     form:
       primary: false
     property_uri: http://purl.org/dc/terms/relation
-    subfields:
+    subproperties:
       related_item:
         type: work_or_url
         required: true
--- a/.koppie/config/metadata_profiles/m3_profile.yaml
+++ b/.koppie/config/metadata_profiles/m3_profile.yaml
@@ -933,7 +933,7 @@ properties:
   # parent; its members are declared as separate properties pointing back with
   # `subproperty_of: <parent>`, each carrying its own `available_on`, indexing,
   # and form placement. `subproperty_of` entries are not standalone resource
-  # attributes. See documentation/forms/compound_fields.md.
+  # attributes. See documentation/compound_fields.md.
   participants:
     available_on:
       class:
--- a/.koppie/config/metadata_profiles/m3_profile.yaml
+++ b/.koppie/config/metadata_profiles/m3_profile.yaml
@@ -928,3 +928,109 @@ properties:
       primary: false
     range: http://www.w3.org/2001/XMLSchema#string
     property_uri: http://vocabulary.samvera.org/ns#transcriptIds
+  # Sample compound (hierarchical) metadata, demonstrating the Hyrax compound
+  # foundation in flexible mode. See documentation/forms/compound_fields.md.
+  agent:
+    available_on:
+      class:
+        - GenericWork
+        - CollectionResource
+    data_type: array
+    type: hash
+    range: http://www.w3.org/2001/XMLSchema#string
+    display_label:
+      default: Agent
+    form:
+      primary: false
+    property_uri: http://id.loc.gov/vocabulary/relators/asn
+    subfields:
+      title: { type: string }
+      agent_name:
+        type: string
+        indexing: [agent_name_sim, agent_name_tesim]
+      agent_role:
+        type: controlled
+        values: [Author, Editor, Contributor, Creator, Data Collector, Funder, Publisher, Other]
+        indexing: [agent_role_sim, agent_role_tesim]
+    groups:
+      - { label: ~, cols: 4, fields: [title, agent_name, agent_role] }
+    view:
+      render_as: compound
+      html_dl: true
+  identifiers:
+    available_on:
+      class:
+        - GenericWork
+        - CollectionResource
+    data_type: array
+    type: hash
+    range: http://www.w3.org/2001/XMLSchema#string
+    display_label:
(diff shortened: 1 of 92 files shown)


jrochkind:
Amazing!

> Compounds render inline or as a card 

Is there (already?) an easy way to provide custom rendering? (with a custom view_component I guess?).  Seems likely to be a need for this.

laritakr:
> Amazing!
> 
> > Compounds render inline or as a card
> 
> Is there (already?) an easy way to provide custom rendering? (with a custom view_component I guess?). Seems likely to be a need for this.

I agree that we need to expand the view abilities... this is likely the next step (and a good stepping off point for anyone else who wants to contribute to this work. Getting the foundation in place was a big enough piece that I just wanted to get the basics going before trying to present them!)

orangewolf (approved):

dmolesUC:
I'm a little concerned that we have a 5,000-line PR approved with no comments (`houndci-bot`'s CSS opinions don't really count). Is there more conversation about this work somewhere else?

laritakr:
> I'm a little concerned that we have a 5,000-line PR approved with no comments (`houndci-bot`'s CSS opinions don't really count). Is there more conversation about this work somewhere else?

@dmolesUC I understand the concern, and I agree it would ideally be recorded on the PR itself.

First, the reassuring part: this change is a **no-op unless you opt in**. Nothing activates the compound metadata unless a downstream includes it in their YAML profile. The YAML changes in the PR are just demonstrations of the features — they don't turn anything on for existing installs. And it was **deliberately excluded from the release that just went out**, so it's not in anyone's hands yet; the intent is to leave room for fine-tuning before it ships.

On the process: Notch8 has a large client driving fairly rapid work on some foundational pieces, and we're contributing back as much as seems feasible for the wider community. Under that time pressure we lean on two patterns — anything with repercussions goes behind a config var or feature flag; genuinely no-op changes we implement directly (as here). That's meant a lot of the design conversation happened 1:1 rather than on the PR, which I recognize isn't ideal for a change this size.

For what it's worth, the plan was discussed with Rob during his twice-weekly Notch8 architecture office hours, and in the Hyku channel during Team Violet office hours; it also came up on the Samvera tech call, though I think that was after the merge.

We did consider just building this into Hyku and not contributing it upstream. The consensus in those conversations was that it's an important capability for Hyrax, and that doing it downstream by overriding Hyrax would be both more work and more fragile — so upstreaming was the deliberate choice.

It's intentionally staged: this is the foundation, with more enhancements to come. If you have concerns, I'd genuinely welcome you testing it out and telling us what more you think it needs — participation is very welcome.

SixtyGrit:
Furthermore, the need for hierarchical metadata was discussed during Hyku interest group meetings as well as the metadata interest group. The community need here was clear and LaRita had demonstrated it to one or both of those groups while we were developing it for the specific client to align the needs with that of the community.

dmolesUC:
I'm not arguing with the value of / validity of / need for the feature, just a little concerned at such a large PR going in without evidence of code review -- thanks for clarifying the circumstances.

I don't like having large volumes of code gated by feature flags rather than modularized into gems -- it's already very difficult to tell which code paths are / aren't going to be executed in a large Hyrax app -- but I freely admit that's the way Hyrax has done things for at least the past several years and possibly since before it was called "Hyrax", so that's not really on this PR.

marrus-sh:
Hyrax has a pretty good test suite and a very good QA process, so I am not overly concerned about code breaking existing functionality. However, I am very concerned about the maintenance burden of these code changes in the future. By my calculation this MR added over 1000 lines of actual Ruby code—excluding JS, YAML, comments, and blank lines—to the Hyrax codebase, an inflation of the total codebase size of 2.67%. Some of these lines are things like

```rb
    def truthy?(value)
      ActiveModel::Type::Boolean.new.cast(value) == true
    end
```

which implies that this behaviour is, for some reason, doing `ActiveModel` casts from values like `"off"` to `false`. It seems that this is being used to enable people to type

```yaml
   required: "off"
```

in their YAML files. Can you justify why this is a feature which is necessary, when Hyrax does not exhibit this behaviour for any of its other YAML configuration files? And for that matter, why does the implementation of `#truthy?` end in `== true`, which does nothing assuming the cast is already casting to `true` or `false`?

There seems to be an impression that the purpose of code review is to validate that a feature is well‐designed and meets an existing need. This is not the purpose of code review—it is the purpose of stakeholder review. In that respect, I agree: It is good to get features merged early, tried out by stakeholders, and refined with additional functionality as needed prior to release.

But the purpose of code review is to actually review the code. And it looks to me like this code has not been reviewed. None of the conversations or feedback which I would expect to have been made about this code are present, and there is no justification documented for why things were implemented in the manner they were implemented. I have a feeling that the reason for this is because nobody actually _knows_ why features were implemented in this manner, because there was never any conscious decision about it. I would love to be wrong about that.

It is the policy of Hyrax that every merge request needs at least 1 approval before it can be merged. If that approval is not actually dependent on somebody actually reviewing the code, it makes me wonder what the purpose of the policy is.

orangewolf:
since this PR is already closed and has been for some time, I've moved the existential part of this discussion to the Samvera slack https://samvera.slack.com/archives/C0F4P3FFF/p1784327116904079. This will provide greater visibility to the community because I want to make sure such concerns get addressed openly.

PR #7514, opened 2026-06-26
Revert #7508 to diagnose main CI failures (OOM)

ShanaLMoore:
**Diagnostic revert** of #7508 (`ca85349`) to check whether it's behind the current CI failures on `main`.

`main` has been red since #7508 and #7512 landed — last green run was `5b5a470` (Jun 24); current HEAD `1f130133b` is failing. The failures are an **OOM/crash, not assertions**: every shard exits `137`, the Postgres logs show `unexpected EOF on client connection`, and the `spec-reports-*` artifacts are ~152 bytes (empty), i.e. rspec is killed before writing any results.

This PR isolates #7508 (a 2-line change to `Hyrax::PermissionsController`) by reverting it on top of current `main`:
- if CI here is **green**, #7508 is implicated and we can dig into why;
- if it's **still red**, #7508 is exonerated and the cause is elsewhere (#7512, or CI/runner memory pressure).

Not meant to merge as-is — purely to get a clean signal. If it goes green, worth a re-run to rule out flaky OOM. cc @bwatson78 @dlpierce

_Surfaced while #7513's CI went red purely from the Update-branch merge pulling in current `main`._

### DIFF
--- a/app/controllers/hyrax/permissions_controller.rb
+++ b/app/controllers/hyrax/permissions_controller.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 module Hyrax
   class PermissionsController < ApplicationController
-    resource_klass = Hyrax.config.use_valkyrie? ? Hyrax::Resource : ActiveFedora::Base
-    load_resource class: resource_klass, instance_name: :curation_concern
+    load_resource class: Hyrax::Resource, instance_name: :curation_concern
 
     attr_reader :curation_concern
     helper_method :curation_concern


ShanaLMoore:
Closing this out. Summary of what the revert told us:

### Failure signature
Every test shard (allinson/dassie/koppie/sirenia, 16/16) dies ~80s into the rspec step with `##[error]Process completed with exit code 137` (SIGKILL). That is the test stack being OOM-killed, not a spec assertion failing. The trailing Postgres log lines (`database "hyrax_test" does not exist`, `relation "sipity_roles" does not exist`) show up in the end-of-job container-log dump and are consistent with Postgres being taken down when the stack is killed, rather than a real schema/migration problem.

### Key finding: #7508 is not the cause
- This branch reverted only #7508, which touches a single file (`app/controllers/hyrax/permissions_controller.rb`).
- **`main` currently passes with #7508 in it.** HEAD `1f130133b` (merge of #7512, which landed after #7508's merge commit `b4b5a0681`) has a long run of green *Test Results of main* runs today (14:59-17:56 UTC).
- `main` hit the same `exit 137` signature once (run at 14:58 on `1f130133b`) and passed on the very next run of the same commit at 14:59.

So removing #7508 doesn't fix anything, because `main` is already green with it. The revert's premise (that #7508 caused the main CI failures) isn't supported by the data.

### Caveat
Unlike main's single transient hit, this branch failed all 16 shards on the original run **and** again on a re-run of the failed jobs (same OOM signature both times). So the OOM is reproducible, not a one-off flake here, but it is not attributable to #7508.

### Recommendation
Track this as a CI infra / memory problem (the docker-compose stack getting OOM-killed under runner memory pressure) rather than via app-level reverts. Likely levers: reduce per-shard container memory / parallelism, add an OOM-aware retry, or move to a larger runner.

PR #7515, opened 2026-06-26
Attempt to get CI working

laritakr:
## What this fixes

CI's `Lint Build Test` matrix was failing all test jobs. Three distinct problems were involved; this branch addresses all three.

## Problem 1 — OOM (exit 137)

Every test job was being SIGKILL'd by the out-of-memory killer. Each job ran the entire docker-compose stack (web + worker, Solr, Postgres, fcrepo, Chromium, FITS, redis, memcached) **plus** RSpec on a single ~7GB `ubuntu-latest` runner — sitting at the memory ceiling and tipping over under variable load (which is why it was intermittent on `main`).

Addressed by reducing the running footprint rather than enlarging the runner:

- Capped the Solr heap so it doesn't auto-size off host RAM.
- Health-gated the backing services (web waits for `service_healthy`).
- Switched the test step from `docker compose up` to `docker compose run web`, and dropped the sidekiq `worker` from web's `depends_on` — the suite uses the ActiveJob `:test` adapter, so a live worker is never exercised.
- Skipped the dev entrypoint (`--entrypoint sh`); specs create/migrate their own DB in `spec_helper`, so `db:prepare`/`db:seed` aren't needed.

Result: jobs now run the full suite to completion (~1300–1800 examples) instead of dying at ~46 seconds.

## Problem 2 — `wrong number of arguments (given 3, expected 1)`

Once the suite ran, hundreds of specs failed with this `ArgumentError`, and seeding crashed with the same error.

Root cause: `sprockets 3.7.2` calls the legacy positional `ERB.new(data, trim_mode, eoutvar)` API, which **`erb 6.x` removed** — so every asset/template render raised the arity error. Because `Gemfile*.lock` is gitignored, CI re-resolves gems to latest on every run, so this broke all branches at once the moment `erb 6.x` was published upstream (unrelated to any merge).

Fixed by pinning `gem 'erb', '~> 4.0'`.

## Problem 3 — jasmine spec (`net::ERR_NAME_NOT_RESOLVED`)

After the above, the only remaining failure was `spec/javascripts/jasmine_spec.rb`, where the remote browser couldn't reach the karma server at `http://web:9876`. The switch to `docker compose run` had dropped the `web` service's network alias, so the hostname no longer resolved.

Fixed by passing `--use-aliases` so the run container keeps the `web` network alias.

## Also added

- Marked the checkout as a git `safe.directory` (via `GIT_CONFIG_*`) so git doesn't abort with a dubious-ownership error under `docker compose run`.

### Prevent gem release breakage:

`Gemfile*.lock` being gitignored makes CI non-deterministic — any bad upstream gem release breaks every branch simultaneously (as `erb 6.x` did here). Committing the lockfiles for the test apps would make CI reproducible and prevent recurrence.

### DIFF
--- a/.github/workflows/lint-build-test.yml
+++ b/.github/workflows/lint-build-test.yml
@@ -133,9 +133,16 @@ jobs:
         # rather than every service in the compose file, reducing CI memory use.
         # Let the container's default entrypoint run in the webapp dir (it does
         # db:prepare/migrate), then cd into the engine for the specs.
+        #
+        # GIT_CONFIG_* marks every repo path as a git safe.directory before the
+        # entrypoint runs. The checkout is owned by a different uid than the
+        # container user, so without this git aborts (dubious ownership) while
+        # the entrypoint and bundler touch the engine and git-sourced gems.
         run: >-
-          docker compose -f docker-compose-${{ matrix.ci_test_app }}.yml run -T --quiet-pull --pull missing web sh -c
-          "git config --global --add safe.directory /app/samvera/hyrax-engine && cd ../hyrax-engine && bundle install && yarn install && rspec_booster --job ${{ matrix.ci_node_index }}/${{ matrix.ci_node_total }}"
+          docker compose -f docker-compose-${{ matrix.ci_test_app }}.yml run -T --quiet-pull --pull missing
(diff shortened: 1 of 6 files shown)


ShanaLMoore (approved):

orangewolf (commented):
Thanks for fixing these LaRita!

PR #7517, opened 2026-06-26
Don't `up` the whole stack, `run` instead.

dlpierce:
Experimenting with not running unneeded services in github actions tests.


laritakr:
Issue is resolved by https://github.com/samvera/hyrax/pull/7515 (which incorporated this attempt)

PR #7518, opened 2026-06-29
Tune fcrepo heap to speed up dassie CI

laritakr:
## Summary

Try to speed up Dassie run.

## Details

Dassie's fcrepo4 ran with the JVM default heap, which on the CI runner made the Fcrepo HTTP round-trips its specs depend on slow (dassie took ~2-3x longer than the other apps). Pin the Jetty JVM to a 1500m max heap via JAVA_OPTIONS.


laritakr:
Did not speed up Dassie, so closing

PR #7525, opened 2026-07-07
Move erb pin to gemspec

dlpierce:
### Fixes

Fixes nurax servers hitting erb ArgumentError during asset:precompile

### Summary

Pinning a gem should be done in the gemspec. Doing so in the Gemfile only affects spec runs, but not the actual application.


laritakr (approved):
Thanks for the followup and catching this. :)

PR #7534, opened 2026-07-10
Install clover into dev apps

dlpierce:
### Fixes

Fixes #7407 

### Summary

Ran `rails g hyrax:iiif_viewer clover` for .koppie and .dassie.
Note: The docker compose stacks have a volume for /public which will obscure the clover files installed there. Running the generator from inside the container will recreate them.

Clover will be used when AV support is flipped on and the file is AV media.


cjcolvar (approved):
👍

PR #7540, opened 2026-07-15
Add typeahead + multi-select opt-ins for controlled compound members

laritakr:
Issue: 
- research-technologies/enact_knapsack#92
- https://github.com/research-technologies/enact_knapsack/issues/93

## Summary

Adds two opt-in, per-member `form:` flags for compound `controlled` metadata members in the M3 profile. Both default **off**, so existing controlled members render exactly as before.

- **`autocomplete: true`** — binds a select2 typeahead over the member's `<select>`. The full option list is already in the DOM, so filtering is client-side (no extra request). Works for single- and multi-select.
- **`multiple: true`** — renders a `<select multiple>`. At submit time each selected value **fans out into its own compound entry**, sharing the row's other members (cartesian product when a row has more than one `multiple` member). A contributor picked with three roles is stored as three single-role entries, not one entry holding an array.

## Screenshots

**Deposit form — Name and Role controls aligned (empty state):**
pr-shot-1-empty-matched

**Multi-select typeahead with several roles chosen (pills wrap in one box, no overflow):**

image

**Edit round-trip — one saved row's roles reopen as separate preselected entries:**

pr-shot-3-edit-roundtrip-fixed

**Show pg**

image

** MOBILE VIEW **

pr-shot-3-mobile

## Details

The key design decision is **fan-out**: the stored shape stays uniform — every member value is a scalar, whether or not the member is multi-select. Consumers never branch on array-vs-scalar, and each term indexes and facets independently. The array only exists transiently between form submit and fan-out; the indexer and show renderer were hardened to tolerate a transient array value because an invalid submission is echoed back before fan-out runs.

Changed (all form / index / render path — nothing touches persistence):

- `app/services/hyrax/compound_schema.rb` — `normalize_subproperty` reads `form.autocomplete` / `form.multiple` into the spec (default false).
- `app/forms/concerns/hyrax/compound_field_behavior.rb` — the load-bearing change: fans a multi-select row out into one entry per value.
- `app/views/hyrax/compounds/_compound_row.html.erb` — `multiple` → `<select multiple>` with `[]` name; `autocomplete` → `data-hyrax-compound-controlled` + placeholder.
- `app/assets/javascripts/hyrax/compound_metadata.js` — `bindControlledSelects` binds select2 v3 to native selects, on load and on cloned rows.
- `app/indexers/.../compound_indexer.rb`, `app/renderers/.../compound_attribute_renderer.rb`, `app/helpers/hyrax/compound_fields_helper.rb`, `app/services/hyrax/compound_subproperty_labeler.rb` — tolerate/represent an array member value.
- `config/locales/hyrax.en.yml` — new `compound_fields.search` placeholder key.
- `documentation/compound_fields.md` — new section + type-table row.

Behavior holds in both flexible and non-flexible modes (the compound list comes from the effective schema either way). The Enact profile opt-in (adding these flags to `contributor_role`) is a separate follow-up in `enact_knapsack`, not part of this PR.

## Testing

Unit/spec coverage added for: schema flag parsing, fan-out (several values → several entries; single value → one scalar; blank handling), array-tolerant indexer/renderer/helper/labeler, and a new view spec for the multi/autocomplete markup. Green on all CI test apps (koppie, dassie, allinson, sirenia).

Manual end-to-end verification in a flexible-mode deposit form (allinson), with `autocomplete: true` + `multiple: true` on a controlled compound member:

- role control renders as a select2 typeahead over a native `<select multiple>`; typing filters client-side;
- picking two roles for one participant, saving → the row **fans out** into two separate scalar entries on the show page (`Name / Role` per entry), each faceting independently;
- reopening edit → the two entries re-collapse into their rows, each preselected;
- "Add" clones a row whose select is also a searchable multi-select.

### select2 styling

select2 v3 copies the control's `.form-control` classes onto its own container (Bootstrap border + fixed `.form-control-sm` height) while rendering its own bordered inner box, producing a "box within a box" and clipping a multi-select's chosen pills. `_compound_metadata.scss` flattens the inner box so only the form-control frame shows and lets the frame grow to fit the selected values. Verified in a flexible-mode deposit form (single clean box; multiple pills wrap without overflowing into the next field).

### DIFF
--- a/app/assets/stylesheets/hyrax/_compound_metadata.scss
+++ b/app/assets/stylesheets/hyrax/_compound_metadata.scss
@@ -93,36 +93,32 @@
 // (Bootstrap border + fixed `.form-control-sm` height) while drawing its own
 // bordered inner box, so a bound control shows a "box within a box" and clips
 // its chosen pills. Flatten the inner box and let the frame grow to fit.
-.hyrax-compound-row {
-  .select2-container.form-control {
-    height: auto;
-    min-height: calc(1.5em + 0.5rem + 2px); // matches .form-control-sm
-    padding: 0;
+// (bindControlledSelects tags the container so this stays off other select2
+// controls in the row.)
+.hyrax-compound-controlled-select2.form-control {
+  height: auto;
+  min-height: calc(1.5em + 0.5rem + 2px); // matches .form-control-sm
+  padding: 0;
 
-    .select2-choices, // multi-select
-    .select2-choice {  // single-select
-      border: 0;
-      background: transparent;
-      box-shadow: none;
-      min-height: 0;
-    }
+  .select2-choice,  // single-select
+  .select2-choices { // multi-select
+    background: transparent;
+    border: 0;
+    box-shadow: none;
+    min-height: 0;
+  }
+}
+
+// Multi-select: match an empty control to a sibling .form-control-sm input so
+// fields line up, and keep pills tidy as they wrap.
+.select2-container-multi.hyrax-compound-controlled-select2 {
+  .select2-choices {
+    padding: 0 0.5rem;
   }
 
-  // Multi-select: match an empty control to a sibling .form-control-sm input so
-  // fields line up, and keep pills tidy as they wrap.
-  .select2-container-multi.form-control {
-    .select2-choices {
-      padding: 0 0.5rem;
-    }
-
-    .select2-search-field input.select2-input {
-      height: calc(1.5em + 0.5rem);
-      margin: 0;
-      line-height: 1.5;
-    }
-
-    .select2-search-choice {
-      margin: 2px 0 2px 5px;
-    }
+  .select2-search-field .select2-input {
+    height: calc(1.5em + 0.5rem);
+    line-height: 1.5;
+    margin: 0;
   }
 }
--- a/app/assets/stylesheets/hyrax/_compound_metadata.scss
+++ b/app/assets/stylesheets/hyrax/_compound_metadata.scss
@@ -88,3 +88,45 @@
     margin-top: 0.25rem;
   }
 }
+
+// A `controlled` sub-property with `form: { autocomplete: true }` binds select2
+// v3 over the native <select>. select2 copies the control's `.form-control`
+// classes onto its own container (a Bootstrap border + the fixed
+// `.form-control-sm` height) while rendering its own bordered inner box (a
+// "box within a box"), and the fixed height clips a multi-select's chosen pills.
+// Flatten the inner box so only the form-control frame shows, and let that frame
+// grow to fit the selected values instead of overflowing into the next field.
+.hyrax-compound-row {
+  .select2-container.form-control {
+    height: auto;
+    min-height: calc(1.5em + 0.5rem + 2px); // matches .form-control-sm
+    padding: 0;
+
+    .select2-choices, // multi-select
+    .select2-choice {  // single-select
+      border: 0;
+      background: transparent;
+      box-shadow: none;
+      min-height: 0;
+    }
+  }
+
+  // Multi-select: size the empty control to match a sibling .form-control-sm
+  // input (so Name and Role line up), and keep the chosen pills tidy as they
+  // wrap and grow the box.
+  .select2-container-multi.form-control {
+    .select2-choices {
+      padding: 0 0.5rem;
+    }
+
+    .select2-search-field input.select2-input {
+      height: calc(1.5em + 0.5rem);
+      margin: 0;
+      line-height: 1.5;
+    }
+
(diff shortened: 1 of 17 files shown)


app/forms/concerns/hyrax/compound_field_behavior.rb
```diff
@@ -56,34 +56,67 @@ def compound_field_names
...
+      present = multiple_keys.select { |key| entry[key].is_a?(Array) && entry[key].any? }
+      return [entry.transform_values { |v| v.is_a?(Array) ? nil : v }] if present.empty?
+
+      value_lists = present.map { |key| entry[key] }
+      value_lists.first.product(*value_lists[1..]).map do |combo|
+        entry.merge(present.zip(combo).to_h)
+      end
     end
   end
 end
```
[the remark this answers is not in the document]
ShanaLMoore: Fixed in bb2b0fa7e. `expand_multiple_members` now only collapses `multiple` members (via `multiple_keys - present`): non-`multiple` fields are never nil'd (so a malformed array on another member is left intact rather than wiped), and an unselected `multiple` member becomes `nil` instead of leaking a `[]` into the stored row.

app/assets/stylesheets/hyrax/_compound_metadata.scss
```diff
@@ -88,3 +88,45 @@
...
+      box-shadow: none;
+      min-height: 0;
+    }
+  }
+
+  // Multi-select: size the empty control to match a sibling .form-control-sm
+  // input (so Name and Role line up), and keep the chosen pills tidy as they
+  // wrap and grow the box.
+  .select2-container-multi.form-control {
+    .select2-choices {
```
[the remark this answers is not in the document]
ShanaLMoore: Fixed in bb2b0fa7e. `bindControlledSelects` now tags each generated container with `.hyrax-compound-controlled-select2` and the styles are scoped to that class, so other select2 controls (work_or_url / linked_record) in the same row are untouched. Verified by enabling the flags on a second member (`relationship_type`): its control picked up the styling while the sibling work_or_url picker in the same row rendered normally.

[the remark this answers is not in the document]
ShanaLMoore: Follow-up (7f5925bb2): per maintainer direction we intentionally broadened this to flatten the box-in-box on *all* select2 controls within a compound row (Item/work_or_url, linked_record, and the controlled typeahead), so the compound form reads consistently. The selector is bounded to `.hyrax-compound-row`, so it does not affect select2 controls elsewhere in the app.

ShanaLMoore:
I ran this locally and can confirm this is working. cc @laritakr  screenshots attached in the description.

app/views/hyrax/compounds/_compound_row.html.erb
```diff
@@ -45,12 +45,23 @@
             <% end %>
             <% case spec[:type] %>
             <% when 'controlled' %>
+              <%# `multiple` submits an array (`[]` suffix) that fans out server-side;
```
orangewolf: there is a lot of variable setting in this view. it makes me think the whole thing needs to be a helper (and therefor tested)

config/metadata/compound_metadata.yaml
```diff
@@ -77,7 +77,7 @@ attributes:
   identifiers:
     type: hash
     multiple: true
-    predicate: http://purl.org/dc/terms/identifier
+    predicate: http://id.loc.gov/ontologies/bibframe/identifiedBy
```
orangewolf, line 80: why the change here?

orangewolf (changes requested):

config/metadata/compound_metadata.yaml
ShanaLMoore: this can be removed from the PR. the sirenia test was failing because there was another property using this predicate. We've since opened a new PR for this here. see [main's pipeline](https://github.com/samvera/hyrax/runs/87698781731) @orangewolf

ShanaLMoore: https://github.com/samvera/hyrax/pull/7546

app/views/hyrax/compounds/_compound_row.html.erb
ShanaLMoore: Good call; extracted the whole `controlled` branch into a `compound_controlled_select` helper (plus a small `compound_controlled_data`), so the view is now just `<%= compound_controlled_select(spec, value, input_name, input_id) %>`. The multiple / autocomplete / force-select logic is now unit-tested in spec/helpers/hyrax/compound_fields_helper_spec.rb. Done in 90e7b01cb.

orangewolf (approved):

PR #7541, opened 2026-07-15
Make derivatives configurable

kirkkwang:
This commit introduces a new configuration option so it will make it easier for consuming applications to override default derivative settings.


laritakr:
I know I can't approve this because Kirk and I are from the same organization, but I think it's a fantastic idea. Anything making Hyrax configurable so we can have less overrides is great.

cjcolvar (approved):
This looks great!  Love how this is backwards-compatible and opens up for configuration values that were hard-coded!

PR #7543, opened 2026-07-16
Fix blank file detected error

kirkkwang:
Previously, if a PDF with no text layer was uploaded, it would error out in the create derivatives background job and keep retrying until it died.  Each retry re-ran derivative generation and would append another thumbnail, so a single file set accumulated multiple duplicate thumbnails.

The error came from `ValkyriePersistDerivatives` raising "blank file detected" when an output wrote zero bytes, which is exactly what a text-less PDF produces when full-text extraction returns nothing.  In this case an empty extraction is fine and shouldn't be a failure.

This commit adds an early return on an extracted text derivative when the output is empty, so it skips the raise and the job completes instead of retrying.  A blank thumbnail still raises, since that indicates something went wrong with the derivative creation process.

### DIFF
--- a/app/services/hyrax/valkyrie_persist_derivatives.rb
+++ b/app/services/hyrax/valkyrie_persist_derivatives.rb
@@ -28,7 +28,11 @@ module Hyrax
       stream.rewind
       output = tmpfile.write(stream.read)
       tmpfile.flush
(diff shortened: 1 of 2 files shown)


cjcolvar (approved):
Good find!

PR #7548, opened 2026-07-17
refactor: add storage-neutral file API to Hyrax::UploadedFile

orangewolf:
## Summary

Part of the chain moving Hyrax's **non-ActiveFedora** upload/storage paths onto Active Storage (see #7547). This PR is a **pure refactor** — no behavior change.

`Hyrax::UploadedFile` gains a small storage-neutral API for reading staged content and metadata:

- `#filename`, `#content_type`, `#byte_size`, `#stored?`
- `#with_io` — yields a rewound binary IO with a real filesystem `#path` (what Valkyrie storage adapters need), closed deterministically after the block
- `#with_local_path` — yields a local path for copy-based consumers

and the storage-agnostic consumers adopt it instead of reaching through the CarrierWave uploader:

| Consumer | Before |
|---|---|
| `ValkyrieIngestJob#upload_file` | `file.uploader.file.to_file` (leaked the IO), `.original_filename` |
| `Hyrax::WorkUploadsHandler#file_set_args` | `file.uploader.filename` |
| `uploads/create.json.jbuilder` | `uploaded_file.file&.file&.filename` / `.size` |
| `SaveCollectionBanner` / `SaveCollectionLogo` steps, `Dashboard::CollectionsController` branding | `f.file_url` used as a **local filesystem path** (only works because CarrierWave's `store_dir` is absolute) |

ActiveFedora-specific consumers (`JobIoWrapper`, `FileActor`, `FileSetActor`) intentionally keep the CarrierWave interface — the AF ingest path is out of scope for this whole chain.

## Why

These call sites were coupled to CarrierWave storing files at absolute local paths. Routing them through one narrow interface lets a follow-up PR add an opt-in Active Storage staging backend without touching every consumer (and keeps that follow-up reviewable).

## Dependencies

None (independent of #7547; the two merge in either order).

## Tests

Run locally against the koppie (Valkyrie) test app — dockerized Postgres/Solr/Redis:

- `spec/models/hyrax/uploaded_file_spec.rb` — new coverage for the API (16 examples, 0 failures)
- `spec/controllers/hyrax/uploads_controller_spec.rb`, `spec/services/hyrax/work_uploads_handler_spec.rb`, `spec/jobs/valkyrie_ingest_job_spec.rb`, `spec/hyrax/transactions/steps/save_collection_banner_spec.rb`, `spec/hyrax/transactions/steps/save_collection_logo_spec.rb` (32 examples, 0 failures, 1 pending)
[...]

PR #7547, opened 2026-07-17
[ActiveStorage For All] enable Active Storage in test apps and install generator

orangewolf:
## Summary

First step of a small-PR chain moving Hyrax's **non-ActiveFedora** upload/storage paths onto Rails **Active Storage** (staging uploads now; an Active Storage-backed Valkyrie storage adapter and a modern upload UI follow in stacked PRs).

This PR only enables the plumbing; **no Hyrax behavior changes**:

- Adds the canonical Rails `create_active_storage_tables` migration to the `.dassie` and `.koppie` test apps (both already ship `config/storage.yml` + `config.active_storage.service` for every environment — the tables were the missing piece) and updates their `schema.rb`.
- Teaches `hyrax:models` generator to run `active_storage:install` so new host applications get the tables alongside Hyrax's own migrations.
- Documents Active Storage setup (including the upgrade path for existing apps) in `documentation/developing-your-hyrax-based-app.md`, and that disk↔S3 is a standard `config/storage.yml` / `config.active_storage.service` change.

## Why

Upcoming PRs in this chain let `Hyrax::UploadedFile` staging and (separately) Valkyrie file storage run on Active Storage, so that switching non-AF storage between local disk and S3 is one Rails config change rather than bespoke Hyrax configuration. ActiveFedora file-set storage is explicitly **out of scope** and unchanged throughout the chain.

## Dependencies

None — safe to merge first. PRs that follow are stacked on this one.

## Tests

- `.koppie` migration run for real against dockerized Postgres (`rails db:create db:migrate`, test env) — `schema.rb` here is that regenerated output.
- `.dassie/db/schema.rb` updated with the identical dumper output (same migration, same Postgres); dassie's full stack (Fedora etc.) was not run locally — CI exercises it.
- `git diff --check` clean; RuboCop clean on touched Ruby (test-app `db/` is outside RuboCop's scan).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7549, opened 2026-07-17
feat: optional Active Storage backend for staged uploads

orangewolf:
## Summary

Adds an **opt-in Active Storage backend for staged uploads** (`Hyrax::UploadedFile`), preserving the existing chunked upload protocol. Part of the non-AF → Active Storage chain; **stacked on #7547 and #7548** (their commits appear in this diff until they merge).

New config:

```ruby
config.uploaded_file_storage_backend = :active_storage # default :carrierwave
```

- **`:carrierwave`** (default): byte-for-byte the existing behavior. ActiveFedora apps should stay here — the AF ingest path reads staged files through CarrierWave (documented in the generated initializer).
- **`:active_storage`**: staged bytes are written through Rails Active Storage, so `config/storage.yml` / `config.active_storage.service` decide where they live — **switching staging from local disk to S3 becomes that one Rails config change.**

Design points:

- `Hyrax::UploadedFile` gains `has_one_attached :file_attachment`. **Reads dispatch per record** (attachment present → Active Storage; CarrierWave file → CarrierWave), so records created under either backend remain readable whatever the setting — no migration required to flip.
- The **client chunk protocol is unchanged** (two-step create, serial `CONTENT-RANGE` chunks, Cloudflare-size-safe requests; the blueimp widget needs no changes). Under Active Storage, `Hyrax::UploadsController` assembles chunks in a local staging file under `config.cache_path` and attaches the completed file when the final chunk arrives (`last byte + 1 == total`); a chunk that doesn't continue the pending assembly restarts it, mirroring the CarrierWave replacement behavior. Assembly assumes sequential chunks on one node or a shared staging path — the same constraint CarrierWave append has always had.
- A bare String assignment (`file = "name.png"`, the protocol's pre-create step) records the **intended filename** ahead of content.
- **Virus scanning** runs once per newly attached content from its local source, instead of CarrierWave's rescan-of-the-growing-file on every chunk append. Unchanged-content saves don't rescan.
- Destroying an upload **purges** its attachment (`ActiveStorage::PurgeJob`).
- `FileSetsController#uploaded_file_from_path` passes the raw multipart upload straight through (both backends accept it; the `CarrierWave::SanitizedFile` wrap was redundant).

## Dependencies

- #7547 (Active Storage tables in test apps / installer)
- #7548 (storage-neutral read API this backend implements)

## Tests

Run locally against the koppie (Valkyrie) test app — dockerized Postgres/Solr/Redis:

- `spec/models/hyrax/uploaded_file_spec.rb`: the storage-neutral API examples now run as **shared examples against both backends**, plus AS-specific coverage (intended filename, `store_file`, virus scan on attach, no rescan on unchanged saves, purge on destroy) — 39 examples, 0 failures
- `spec/controllers/hyrax/uploads_controller_spec.rb`: legacy CarrierWave examples untouched and passing; new AS context covers pre-create, single-shot attach, **sequential chunk assembly + finalize with content verification**, staging-file cleanup, mismatched-chunk restart, purge on destroy — with the existing chunk examples, 140 examples total, 0 failures
- `spec/lib/hyrax/configuration_spec.rb` (backend validation), `spec/controllers/hyrax/file_sets_controller_spec.rb`, `spec/jobs/valkyrie_ingest_job_spec.rb`, `spec/services/hyrax/work_uploads_handler_spec.rb` — green

AF-side (dassie) suite portions not run locally; CI covers them. RuboCop clean on touched files.

## Notes for reviewers

- Attaching staged files enqueues Active Storage's `AnalyzeJob` (image metadata); harmless for staging records, called out for awareness.
- Uploads-controller chunk-append authorization (any signed-in user can PATCH any upload id) is a **pre-existing** gap left as-is here for parity; a follow-up PR adds ownership enforcement at the attach seam.

### DIFF
--- a/.dassie/db/migrate/20260716000001_create_active_storage_tables.active_storage.rb
+++ b/.dassie/db/migrate/20260716000001_create_active_storage_tables.active_storage.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+# This migration comes from active_storage (originally 20170806125915)
+class CreateActiveStorageTables < ActiveRecord::Migration[7.0]
+  def change
+    # Use Active Record's configured type for primary and foreign keys
+    primary_key_type, foreign_key_type = primary_and_foreign_key_types
+
+    create_table :active_storage_blobs, id: primary_key_type do |t|
+      t.string   :key,          null: false
+      t.string   :filename,     null: false
+      t.string   :content_type
+      t.text     :metadata
+      t.string   :service_name, null: false
+      t.bigint   :byte_size,    null: false
+      t.string   :checksum
+
+      if connection.supports_datetime_with_precision?
+        t.datetime :created_at, precision: 6, null: false
+      else
+        t.datetime :created_at, null: false
+      end
+
+      t.index [:key], unique: true
+    end
+
+    create_table :active_storage_attachments, id: primary_key_type do |t|
+      t.string     :name,     null: false
+      t.references :record,   null: false, polymorphic: true, index: false, type: foreign_key_type
+      t.references :blob,     null: false, type: foreign_key_type
+
+      if connection.supports_datetime_with_precision?
+        t.datetime :created_at, precision: 6, null: false
+      else
+        t.datetime :created_at, null: false
+      end
+
+      t.index [:record_type, :record_id, :name, :blob_id], name: :index_active_storage_attachments_uniqueness, unique: true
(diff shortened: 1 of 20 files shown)


orangewolf:
Superseded by #7555 — same change moved to an in-repo branch (active-storage-for-all/uploaded-file-staging) and retitled.

PR #7550, opened 2026-07-17
feat: add Active Storage-backed Valkyrie storage adapter

orangewolf:
## Summary

Adds **`Hyrax::Storage::ActiveStorage`** — a Valkyrie `StorageAdapter` that stores repository files (PCDM files belonging to file sets, including versions) as `ActiveStorage::Blob` records. Part of the non-AF → Active Storage chain; **stacked on #7547** (its commit appears in this diff until it merges). Independent of #7548/#7549 — this adapter serves the *repository* side, they serve the *staging* side.

With this adapter registered, which backend holds repository file bytes is decided by standard Rails configuration — `config/storage.yml` + `config.active_storage.service`, or a per-instance `service_name:`. **Switching a Valkyrie app's repository files from local disk to S3 is that one config change**; no bespoke Hyrax S3 configuration.

Design:

- **Versioning mirrors `Valkyrie::Storage::VersionedDisk`** exactly, encoded in blob keys instead of paths: versions live at `<key_prefix><resource_id>/v-<millis>-<filename>`, and file identifiers circulate in the stable `v-current-` reference form — `Hyrax::FileMetadata#file_identifier` stays stable as versions arrive, `Hyrax::VersioningService` and the file-set **Versions tab work unchanged** (`supports?(:versions)` and `:version_deletion` are true).
- **Lazy IO**: `find_by` returns a `StreamFile` whose blob IO downloads nothing until content is actually read; `#disk_path` consumers (FITS characterization, derivatives, downloads, riiif) keep working via local materialization — the same pattern as the Fedora adapter.
- **Blobs only, no `ActiveStorage::Attachment` rows** — these are repository files, not app-level attachments (also avoids analyze jobs).
- Multiple instances can target different key namespaces/services (e.g. a derivatives instance).
- Registered in koppie as `:active_storage_storage`, selectable via `VALKYRIE_STORAGE_ADAPTER=active_storage_storage`; **koppie's default adapter is unchanged** (a later PR proposes the flip). ActiveFedora file-set storage is untouched, as throughout this chain.
- Docs: registration + S3 example in `documentation/developing-your-hyrax-based-app.md`.

## Dependencies

- #7547 (Active Storage tables in the test apps)

## Tests

Run locally against the koppie (Valkyrie) test app — dockerized Postgres/Solr/Redis:

- `spec/services/hyrax/storage/active_storage_spec.rb` — valkyrie's full **`'a Valkyrie::StorageAdapter'` shared compliance suite** (upload/find_by/delete/checksums plus the complete `upload_version` / `find_versions` / version-deletion semantics), plus adapter-specific examples: stable current-reference ids, byte-identical round-trips, `disk_path` materialization, **no download on lookup**, prefix-scoped `handles?`, slash-containing filenames (collection-branding style), per-instance `service_name`, and an **integration example through `Hyrax::ValkyrieUpload` + `Hyrax::VersioningService`** (ingest → version on re-upload → `find_versions` == 2). Result: 23 examples, 0 failures, 1 skip.
- The single skip is valkyrie's lsof file-descriptor-count shared example, which asserts strict fd equality and is environment-sensitive under the full Hyrax test stack (connection pools); the property it guards — no eager download on `find_by` — is covered by a dedicated example. The skip carries this explanation in the spec.

RuboCop clean on new files; `git diff --check` clean.

### DIFF
--- a/.dassie/db/migrate/20260716000001_create_active_storage_tables.active_storage.rb
+++ b/.dassie/db/migrate/20260716000001_create_active_storage_tables.active_storage.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+# This migration comes from active_storage (originally 20170806125915)
+class CreateActiveStorageTables < ActiveRecord::Migration[7.0]
+  def change
+    # Use Active Record's configured type for primary and foreign keys
+    primary_key_type, foreign_key_type = primary_and_foreign_key_types
+
+    create_table :active_storage_blobs, id: primary_key_type do |t|
+      t.string   :key,          null: false
+      t.string   :filename,     null: false
+      t.string   :content_type
+      t.text     :metadata
+      t.string   :service_name, null: false
+      t.bigint   :byte_size,    null: false
+      t.string   :checksum
+
+      if connection.supports_datetime_with_precision?
+        t.datetime :created_at, precision: 6, null: false
+      else
+        t.datetime :created_at, null: false
+      end
+
+      t.index [:key], unique: true
+    end
+
+    create_table :active_storage_attachments, id: primary_key_type do |t|
+      t.string     :name,     null: false
+      t.references :record,   null: false, polymorphic: true, index: false, type: foreign_key_type
+      t.references :blob,     null: false, type: foreign_key_type
+
+      if connection.supports_datetime_with_precision?
+        t.datetime :created_at, precision: 6, null: false
+      else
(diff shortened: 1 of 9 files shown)


orangewolf:
Superseded by #7556 — same change moved to an in-repo branch (active-storage-for-all/valkyrie-storage-adapter) and retitled.

Issue #7544, opened 2026-07-16
[Sirenia] Test fails when attempting to valkyrie_create with string value for identifier with compount metadata enabled


cjcolvar:
### Descriptive summary
I'm seeing the following test failure (`spec/features/work_show_spec.rb:190`) that is blocking a clean build of my PR (#7542).  It appears to only happen with Sirenia and not Koppie or Dassie so the mix of valkyrie and fedora adapater. I'm guessing that this is because both `identifier` and `identifiers` share the same predicate so the string value being passed to `valkyrie_create` gets checked in `identifiers` and fails the dry-types validation.

I think this is breaking builds now is because the docker image in the github cache got replaced with a new build when clover was installed in the test apps (#7534).

```
1) work show view in valkyrie shows a work
     Failure/Error: super

     Dry::Struct::Error:
       [Monograph.new] ["867-5309"] (Array) has invalid type for :identifiers violates constraints (type?(Hash, "867-5309") failed)
     # /app/bundle/ruby/3.3.0/gems/dry-types-1.9.1/lib/dry/types/schema.rb:330:in `rescue in block in resolve_unsafe'
     # /app/bundle/ruby/3.3.0/gems/dry-types-1.9.1/lib/dry/types/schema.rb:327:in `block in resolve_unsafe'
     # /app/bundle/ruby/3.3.0/gems/dry-types-1.9.1/lib/dry/types/schema.rb:322:in `each'
     # /app/bundle/ruby/3.3.0/gems/dry-types-1.9.1/lib/dry/types/schema.rb:322:in `resolve_unsafe'
     # /app/bundle/ruby/3.3.0/gems/dry-types-1.9.1/lib/dry/types/schema.rb:60:in `call_unsafe'
     # /app/bundle/ruby/3.3.0/gems/dry-types-1.9.1/lib/dry/types/constructor.rb:84:in `call_unsafe'
     # /app/bundle/ruby/3.3.0/gems/dry-struct-1.8.1/lib/dry/struct/class_interface.rb:254:in `new'
     # ./app/models/concerns/hyrax/compound_normalization.rb:24:in `new'
     # /app/bundle/ruby/3.3.0/gems/valkyrie-3.5.1/lib/valkyrie/types.rb:65:in `block in <module:Types>'
     # /app/bundle/ruby/3.3.0/gems/dry-types-1.9.1/lib/dry/types/constructor/function.rb:17:in `call'
[...]

PR #7559, opened 2026-07-20
Match identifiers predicate change in compound metadata config

cjcolvar:
The predicate used for identifiers in the compound metadata changed (#7546) due to conflicts with the identifier field in basic metadata leading to errors saving identifier when both fields were present (sirenia).

Related to #7544


cjcolvar:
Duplicate of #7560

PR #7560, opened 2026-07-20
Update `identifiers` predicate

kirkkwang:
This is a follow up to a previous change in the `identifiers` predicate to fix CI issues.  The m3 profiles that have the same property should also be updated.

Ref:
- https://github.com/samvera/hyrax/pull/7546/changes/387817e4


cjcolvar (approved):
Thanks!

PR #7566, opened 2026-07-28
Linked record picker enhancements

laritakr:
## Summary
Strengthens the `linked_record` compound picker so a cataloger can recognize and reuse an existing record instead of creating a duplicate, and reaches the create form without hunting for it.
https://github.com/research-technologies/enact_knapsack/issues/138

## Details
Supports two behaviors a downstream picker needs to avoid duplicate records, expressed entirely in the generic `linked_record` layer:

- **Tighten matching before create.** Adds `similar`, a new optional source capability on `Hyrax::CompoundLinkedRecordResolver` alongside `search`/`create`/`match` — a fuzzy, multi-row "did you mean" lookup exposed through a generic `linked_record_similar` QA authority (one authority per every source; no per-source class). When a source registers it, clicking Create first surfaces near-matches with a "Use this" action; the primary Create button is the override (press again to create anyway, editing the name re-checks). This is what lets a picker catch a near-duplicate the exact/substring search would miss.
- **Distinguish results.** Picker result rows may carry an optional `detail:` string, rendered as a muted second line, so records that share a label are told apart at a glance. Sources that omit it render the plain label unchanged.
- **Reach create without an empty search.** The "Add new" affordance is now always visible for a creatable source, so the create form is reachable alongside results rather than only after a search returns nothing.

A source that registers no `similar:` proc keeps today's behavior end to end. The always-visible "Add new" is a behavior change, but no Hyrax-bundled profile uses a creatable `linked_record`, so nothing shipped changes — flagged for maintainer review of the default.

<details>
<summary>Screenshot from consuming application</summary>
Screenshot 2026-07-28 at 6 28 07 PM
</details>

## Notes

I locked down valkyrie's version to get the specs to pass until Valkyrie is fixed. See https://samvera.slack.com/archives/C5F0P8K0X/p1782316745553639

### DIFF
--- a/app/assets/javascripts/hyrax/compound_metadata.js
+++ b/app/assets/javascripts/hyrax/compound_metadata.js
@@ -46,12 +46,16 @@
                     quietMillis: 250,
                     data: function(term, page) { lastTerm = term; return { q: term }; },
                     results: function(data, page) {
-                        // For a creatable linked_record, surface the "Add new"
-                        // trigger when the search came back empty.
+                        // Keep the creatable linked_record's "Add new" term stashed
+                        // for the create-form prefill (the affordance is always
+                        // visible; see toggleAddNew).
                         if (isLinkedRecord) toggleAddNew(wrapEl, data.length === 0, lastTerm);
                         return {
                             results: data.map(function(obj) {
-                                return { id: obj.id, text: [].concat(obj.label)[0] };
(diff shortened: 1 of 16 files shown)


dlpierce (dismissed):
Looks good. Valkyrie 3.6.1 should be out shortly.

kirkkwang (approved):

PR #7568, opened 2026-07-31
Fix trailing characters in citations

kirkkwang:
This commit will change the way we derive the MLA, APA, and Chicago citation format so in certain conditions, we don't get a trailing : or , character.  Previously, if a work had a publisher but no location set, then the citation would look something like:

"Revelle, R. (1960). Ocean Circulation. : Scripps Institution of Oceanography."

Also, for MLA, if the date is set with no publisher or location, it would produce a citation like:

"Revelle, Roger. Ocean Circulation. , 1960."


cjcolvar (approved):
Looks great!  Thanks for finding this and then fixing it!

PR #7570, opened 2026-08-04
Improve performance of dashboard/works page

bbpennel:
### Summary

Pull list of depositable admin sets from solr rather than from repository, otherwise for wings or valkyrie/fedora it retrieves the record for every admin set when all we need is info stored in solr. Our dashboard/works page was taking 25 seconds to load in hyrax 5, which was new.

notes-bugfix

### Detailed Description

We observed that the dashboard/works page was taking 25 seconds to load in our QA environment after upgrading from hyrax 4 to 5, but it was only taking 1-2 seconds in our Hyrax 4 instance with identical data. We are using Wings at this point, before we start looking at migrating to Fedora 7. 

It turns out that it was populating the list of admin sets by querying activefedora for each admin set, and our repository contains 30 of them. This was a new behavior in Hyrax 5, since it was now using the queryService.find_many_by_ids method, but that retrieves far more information than is needed for populating this UI component.

This PR switches to pulling the list directly from solr, which gets performance back to the Hyrax 4 level. I have the behavior change for both Wings and Valkyrie, since it looks like the Valkyrie/fedora behavior will have the same problem, but I have not tested there yet.

* Using solr directly for retrieving admin sets info for dashboard/works facet filter.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/helpers/hyrax/hyrax_helper_behavior.rb
+++ b/app/helpers/hyrax/hyrax_helper_behavior.rb
@@ -52,11 +52,13 @@ module Hyrax
       indexed_ids = admin_sets_list.map(&:last)
       missing_ids = source_ids.map(&:to_s) - indexed_ids
 
-      admin_sets_list.concat(
-        Hyrax.query_service.find_many_by_ids(ids: missing_ids).map do |source|
-          [source.title.first, source.id.to_s]
-        end
-      )
+      if missing_ids.present?
+        admin_sets_list.concat(
+          Hyrax.query_service.find_many_by_ids(ids: missing_ids).map do |source|
+            [source.title.first, source.id.to_s]
+          end
+        )
+      end
 
       # Sorts the default admin set to be first, then the rest by title.
       admin_sets_list.sort do |a, b|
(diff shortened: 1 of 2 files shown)


orangewolf (approved):
thanks @bbpennel this looks great =-)

dmolesUC:
I would expect Valkyrie/Postgres to behave better, or at least no worse, when pulling directly from the database… honestly, at this point, we're doing our best to eliminate Solr calls wherever we can. 🙁

bbpennel:
I considered scoping this change to just Wings, but the valkyrie/fedora implementation likely would have a similar performance problem. So scoping it to Wings and valkyrie/fedora but not valkyrie/postgres felt awkward, particularly without modifying the actual valkyrie implementation.

dmolesUC:
Two thoughts on that:

1. We have various places in the code where we switch behavior depending on whether we're using `ActiveFedora`, for instance [`AdminSetsController#admin_set_params`](https://github.com/samvera/hyrax/blob/main/app/controllers/hyrax/admin/admin_sets_controller.rb#L209) or [`DeepIndexingService#fetch_external`](https://github.com/samvera/hyrax/blob/main/app/indexers/hyrax/deep_indexing_service.rb#L35-L36). We could do something like that here.
2. Extracting the construction of `admin_sets_list` into its own method would make it easier to swap a `Hyrax.query_service`-based implementation back in, or to modularize it later.

Tangentially (and this is a problem with the existing code, not the PR), it feels sketchy to call the method `available_admin_sets…` and the variable `admin_sets_list` when it's really just an array of title-ID pairs (contra the documentation, which says it's an `Array<String>`). Plus, we use this data for creating collections as well, so the `…for_creating_works` part isn't really accurate either.

PR #7573, opened 2026-08-05
Let engines add action buttons to a work's show page

laritakr:
## Summary

Adds show_actions_for, a helper returning the names of extra action partials to render in the button row, defaulting to none. Mirrors `form_progress_sections_for`.

## Rationale

Engines wanting a button there previously had to override `_show_actions.html.erb` and keep a copy of Hyrax's markup in sync with it.


cjcolvar (approved):
Looks good.  Nice to have more clear seams for modifying base Hyrax!

PR #7575, opened 2026-08-10
Render license labels under flexible metadata

ShanaLMoore:
### What

On flexible-metadata apps, a work's license renders as a bare auto-linked URI ("https://creativecommons.org/licenses/by/4.0/") instead of its authority label ("Creative Commons BY Attribution 4.0 International").

### Why

Attribute rendering under flexible metadata follows the m3 profile's view options, and the profiles Hyrax ships (engine default, koppie, dassie) mark license as `render_as: external_link`. Every app seeded from those profiles also carries the setting in its stored flexible schema, so fixing the profiles alone would not heal existing apps.

### Fix

* `AttributesHelper#conform_options` applies the semantic renderer for the two authority-backed fields (`license` -> `:license`, `rights_statement` -> `:rights_statement`), overriding a profile-supplied `render_as`. The semantic renderer's label still links out to the URI, so an `external_link` rendering is strictly worse than the label - which is why overriding the stored value is safe, and why this heals existing stored schemas without a migration.
* The three shipped m3 profiles change license's view to `render_as: license` so newly seeded schemas agree with the code. `related_url` keeps its `external_link`, and the helper leaves fields without a semantic mapping untouched (covered by spec).

### Testing

Spec additions in the existing `#conform_options` block cover both semantic fields, the override of a stored `external_link`, and non-semantic fields passing through unchanged. Written and verified downstream first: this fix originated as a Hyku decorator (samvera/hyku#3212), verified live against seeded flexible-metadata tenants where the license row now shows the authority label; upstreaming it here per review feedback since the helper and the profiles are Hyrax's. I could not run the full engine suite locally, so leaning on CI for the suite run.

Downstream note: once this lands, the Hyku decorator in samvera/hyku#3212 becomes redundant and that PR slims to nothing (its profile fix is Hyku's own copy, also being handled).

### DIFF
--- a/app/helpers/hyrax/attributes_helper.rb
+++ b/app/helpers/hyrax/attributes_helper.rb
@@ -63,6 +63,16 @@ module Hyrax
       options_hash&.with_indifferent_access&.fetch('render_term', nil) || field_name
     end
 
+    # Authority-backed fields whose values are URIs with human-readable
+    # labels. Their semantic renderers always win over a profile-supplied
+    # render_as: the label still links to the URI, so rendering the bare
+    # URI (e.g. as an external_link) is strictly worse, and profiles
+    # already stored in existing flexible schemas carry the old setting.
+    SEMANTIC_RENDERERS = {
+      license: :license,
+      rights_statement: :rights_statement
+    }.freeze
+
     # @param [String] field name
     # @param [Hash<Hash>] a nested hash of view options...
     #        {:label=>{"en"=>"Title", "es"=>"Título"}, :html_dl=>true}
@@ -88,6 +98,8 @@ module Hyrax
         )
       end
       view_options[:base_url] = request.base_url if respond_to?(:request) && request.respond_to?(:base_url)
+      semantic_renderer = SEMANTIC_RENDERERS[field_name.to_sym]
+      view_options[:render_as] = semantic_renderer if semantic_renderer
       view_options
     end
 
--- a/app/helpers/hyrax/attributes_helper.rb
+++ b/app/helpers/hyrax/attributes_helper.rb
@@ -63,15 +63,15 @@ module Hyrax
       options_hash&.with_indifferent_access&.fetch('render_term', nil) || field_name
     end
 
-    # Authority-backed fields whose values are URIs with human-readable
-    # labels. Their semantic renderers always win over a profile-supplied
-    # render_as: the label still links to the URI, so rendering the bare
-    # URI (e.g. as an external_link) is strictly worse, and profiles
-    # already stored in existing flexible schemas carry the old setting.
-    SEMANTIC_RENDERERS = {
-      license: :license,
-      rights_statement: :rights_statement
-    }.freeze
+    # These override any stored render_as: the semantic label still links
+    # to the URI, so a bare external_link rendering is strictly worse.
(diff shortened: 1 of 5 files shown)


spec/helpers/hyrax/attributes_helper_spec.rb
```diff
@@ -42,6 +42,29 @@
...
+      # render_as: external_link for license; the authority label must win
+      # anyway - it still links to the URI, and a bare URI is strictly worse.
+      result = helper.conform_options(:license, { render_as: 'external_link' }.with_indifferent_access)
+      expect(result[:render_as]).to eq(:license)
+    end
+
+    it 'leaves fields without a semantic renderer alone' do
+      result = helper.conform_options(:related_url, { render_as: 'external_link' }.with_indifferent_access)
+      expect(result[:render_as]).to eq('external_link')
+    end
```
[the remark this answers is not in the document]
ShanaLMoore: Good catch, added the parallel example for rights_statement in 49accbe10.

app/helpers/hyrax/attributes_helper.rb
```diff
@@ -88,6 +98,8 @@ def conform_options(field_name, view_options)
         )
       end
       view_options[:base_url] = request.base_url if respond_to?(:request) && request.respond_to?(:base_url)
+      semantic_renderer = SEMANTIC_RENDERERS[field_name.to_sym]
```
orangewolf: how were right statements handled before this change?

app/helpers/hyrax/attributes_helper.rb
```diff
@@ -63,6 +63,16 @@ def conform_field(field_name, options_hash)
...
 
+    # Authority-backed fields whose values are URIs with human-readable
+    # labels. Their semantic renderers always win over a profile-supplied
+    # render_as: the label still links to the URI, so rendering the bare
+    # URI (e.g. as an external_link) is strictly worse, and profiles
+    # already stored in existing flexible schemas carry the old setting.
+    SEMANTIC_RENDERERS = {
+      license: :license,
+      rights_statement: :rights_statement
+    }.freeze
```
orangewolf: lets avoid constants whenever possible. they are a common but inflexible pattern and antithetical to Ruby's design principles. Leaving this is a hash is a tiny bit of over-engineering, but I think that's probably fine.

orangewolf (changes requested):

app/helpers/hyrax/attributes_helper.rb
ShanaLMoore: Before this change rights_statement was already rendering correctly: all three shipped profiles (engine, koppie, dassie) already set `render_as: rights_statement`, so it showed the authority label. Only license was misconfigured as `external_link`. I included rights_statement in the helper mapping so custom or already-stored schemas that carry a different render_as still get the label; for the shipped profiles it is a no-op.

app/helpers/hyrax/attributes_helper.rb
ShanaLMoore: Done in 1375f8942: replaced the constant with a `semantic_renderers` method returning the hash, so including classes can override it.

laritakr:
I've got a rework plan to show labels for Hyku's new UI-based vocabulary terms that may revise the need for this, so I recommend holding off on it and letting that rework make these changes as needed.

PR #7580, opened 2026-08-11
Allow a compound to declare an ordered validation

laritakr:
## Summary
A compound metadata property can declare a rule relating two of its sub-properties, so an entry whose end date falls before its start date is refused.

## Details
Nothing previously stopped a reversed range from saving: validation was per-sub-property, with no way to express a rule across two. This validation is declared on the compound rather than a sub-property:

```yaml
  dates:
    data_type: array
    type: hash
    validations:
      - { type: ordered, before: start_date, after: end_date }
```
 
- `ordered` compares dates. A blank `after` (the usual case for an optional end date) is allowed, as are equal values; a value that does not parse as a date is skipped rather than flagged.
- One message per broken rule, not per offending row, matching how `missing_required_subproperties` dedupes. A row still missing a `required:` sub-property reports only that, so a half-filled entry doesn't raise two complaints.
- An unrecognized `type:` is carried through rather than rejected, so a profile written against a newer Hyrax still loads. A new profile-validation warning surfaces it, which stops a misspelling (`ordred`) from silently disabling the rule.
- Opt-in: compounds without `validations:` behave exactly as before.


app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module FlexibleSchemaValidators
+    # Warns about a compound `validations:` rule that will never fire.
+    #
+    # {Hyrax::CompoundSchema} normalizes the rule's shape without checking its
+    # `type`, so an unrecognized one is stored and then quietly ignored at
+    # validation time. That keeps a profile written against a newer Hyrax loadable
+    # here, but it also means a misspelling (`ordred`) silently disables the rule.
```
dmolesUC, line 10: Is this a situation we expect to get into, where we're trying to make sure new profiles are backwards-compatible with earlier versions of Hyrax? That feels like it could be a lot of work.

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def warn_about(compound, rule)
+        type = rule['type'].to_s
+
+        if Hyrax::CompoundEntryValidation::RULE_TYPES.exclude?(type)
+          add(:unknown_type, compound: compound, type: type.presence || '(blank)')
+        elsif rule['before'].blank? || rule['after'].blank?
+          add(:incomplete_rule, compound: compound, type: type)
+        end
```
dmolesUC, line 40: I'm not excited about the prospect of this `if` statement just getting longer and longer over time — I think it'd be cleaner if it was just easy to add small validators focused on individual rule types.

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+      def initialize(profile, warnings)
+        @profile = profile
+        @warnings = warnings
+      end
+
+      def validate!
+        (@profile['properties'] || {}).each do |name, config|
+          next unless config.is_a?(Hash)
+
+          rules(config).each { |rule| warn_about(name, rule) }
```
dmolesUC, line 22: Reading this line in isolation, it feels like every rule is going to generate a warning. I'd expect to see something more like:

```ruby
rules(config).each do |rule|
  validator = validator_for(rule) # presumably switches on type
  validator.validate!
  @warnings << validator.warnings
end
```

(There are a lot of different ways you could record the validation state, this is just off the top of my head -- but you get the idea.)

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+module Hyrax
+  module FlexibleSchemaValidators
+    # Warns about a compound `validations:` rule that will never fire.
+    #
+    # {Hyrax::CompoundSchema} normalizes the rule's shape without checking its
+    # `type`, so an unrecognized one is stored and then quietly ignored at
+    # validation time. That keeps a profile written against a newer Hyrax loadable
+    # here, but it also means a misspelling (`ordred`) silently disables the rule.
+    # Warning rather than erroring preserves the first while surfacing the second.
+    class CompoundValidationsValidator
```
dmolesUC, line 12: I find myself questioning a bit whether this class should exist at all, given it currently only validates the well-formedness of `ordered` rules (plus noting the presence of unknown rule types). Maybe those should be separate validators?

dmolesUC: Also, it took me a while to figure out what this class actually does, which makes me think `CompoundValidationsValidator` isn't the best name for it. I think `ValidationsValidator` is just inherently confusing.

It makes some sense if you're coming from `CompoundEntryValidation`, but because till quite recently we've had `Validator` classes but not `Validation` classes, it's not immediately obvious what the difference would be.

I wonder if `Validations` shouldn't maybe be `ValidationRules`, in which case (assuming we keep it) this class would be something like `CompoundValidationRuleValidator` or `ValidationRuleSetValidator` -- making it clear that it validates the rules but doesn't apply them.

That, though, makes me wonder if `CompoundEntryValidation` should exist at all -- why isn't it just part of `CompoundValidator`?

spec/services/hyrax/flexible_schema_validators/compound_validations_validator_spec.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+RSpec.describe Hyrax::FlexibleSchemaValidators::CompoundValidationsValidator do
+  subject(:validator) { described_class.new(profile, warnings) }
+
+  let(:warnings) { [] }
+
+  def compound(validations)
```
dmolesUC, line 8: If this is just `dates`, let's call it `dates` (when it does something else, we can rename it)

app/services/hyrax/compound_entry_validation.rb
```diff
@@ -12,11 +12,20 @@ module Hyrax
...
   #   * every populated row must fill all of the compound's `required`
-  #     sub-properties.
+  #     sub-properties;
+  #   * a declared `ordered` rule (`validations: [{ type: ordered, before:, after: }]`)
+  #     requires `after` to be greater than or equal to `before` in every row that
+  #     supplies both comparable values.
   #
   # Rows are the post-populator persisted hashes (all-blank rows already
   # dropped), so a no-required compound with no rows is valid.
   class CompoundEntryValidation
```
dmolesUC, line 22: I know this is ten weeks old now, but why isn't this class just part of `CompoundEntryValidator`?

app/services/hyrax/compound_entry_validation.rb
```diff
@@ -26,11 +35,16 @@ def initialize(definition, entries)
...
+    #   `:out_of_order`.
     def violations
       return [{ type: :required_but_empty, missing: required_keys }] if required_but_empty?
 
-      rows_missing_required.map { |missing| { type: :missing_required_subproperties, missing: missing } }
+      missing = rows_missing_required.map { |keys| { type: :missing_required_subproperties, missing: keys } }
+      # An unfinished row cannot be compared, so the required-field error stands alone.
+      return missing if missing.any?
+
+      ordering_violations
```
dmolesUC, line 47: I'm not excited about this method just getting longer and longer as more rules are added (and the class getting longer and longer as methods are added to implement those rules). I think smaller, more focused validation rule classes would be easier to reason about.

spec/services/hyrax/compound_entry_validation_spec.rb
```diff
@@ -69,6 +69,62 @@ def sub(required: false)
...
+    # One message however many rows are wrong, as with missing_required_subproperties.
+    it 'reports the rule once when several rows are reversed' do
+      rows = [{ 'start' => '2026-06-30', 'finish' => '2026-01-01' },
+              { 'start' => '2025-06-30', 'finish' => '2025-01-01' }]
+
+      expect(violations_for(rows).count { |v| v[:type] == :out_of_order }).to eq(1)
+    end
+
+    it 'ignores values it cannot compare' do
+      expect(violations_for([{ 'start' => 'circa 1920', 'finish' => 'later' }])).to be_empty
```
dmolesUC, line 108: Should this be a warning?

(And if we want to be able to handle and also validate this kind of data in a GLAM-standard way, should we just be using EDTF?)

documentation/compound_fields.md
```diff
@@ -700,6 +700,38 @@ Sub-property format and controlled-vocabulary correctness belong in an
...
+
+`ordered` compares dates: it requires `after` to fall on or after `before` in every
+row supplying both, so an entry ending before it starts is refused. A blank `after`
+(the usual case for an optional end date) is allowed, as is a pair of equal values.
+Values that do not parse as dates — a partial or free-text date the profile permits
+— are skipped rather than treated as an error.
+
+One message is added per broken rule, not per offending row, and a row still
+missing a `required:` sub-property reports only that, so a half-filled entry does
+not raise two complaints at once.
```
dmolesUC, line 732: It's not obvious to me how this works in practice -- or whether it's true in practice that if a required field is missing it's never worth doing any other validation -- but I suppose as long as people do try to clear all the warnings it's probably fine, even if it requires more iteration on the schema.

documentation/compound_fields.md
```diff
@@ -700,6 +700,38 @@ Sub-property format and controlled-vocabulary correctness belong in an
...
+
+`Hyrax::CompoundEntryValidation` owns the rules and
+`Hyrax::CompoundEntryValidator` renders them, keyed on
+`hyrax.compound_fields.errors.out_of_order`.
+
+An unrecognized `type:` is carried through rather than rejected, so a profile
+written against a newer Hyrax still loads — but profile validation warns about it,
+which is what stops a misspelled rule from silently doing nothing. A rule missing
+`before:` or `after:` warns the same way.
```
dmolesUC, line 741: As I noted on `CompoundValidationsValidator`, it's not clear why we should go out of our way to support loading newer profiles in older Hyraxes. But I'll take your word for it if you say it's a real use case you've run into.

app/services/hyrax/compound_entry_validation.rb
```diff
@@ -66,6 +80,43 @@ def populated_rows
...
+    # (blank, or a partial/free-text date the profile permits) is skipped rather than
+    # flagged. Comparing other types would mean declaring the type in the rule.
+    def comparable(row, key)
+      value = row[key] || row[key.to_sym]
+      return if value.blank?
+
+      Date.parse(value.to_s)
+    rescue ArgumentError, TypeError
+      nil
+    end
```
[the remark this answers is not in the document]
dmolesUC: `1920` and `2026-01` strike me as completely legit dates that should still be compared. It's not clear to me, actually, what kind of "partial" date we'd want to skip. (Again, let me put a plug in for EDTF.)

app/services/hyrax/flexible_schema_validator_service.rb
```diff
@@ -40,6 +40,7 @@ def validate!
       validate_sort_properties
       validate_redirects
       validate_compound
+      validate_compound_validations
```
dmolesUC, line 43: Yeah, "validate validations" just sounds weird. Maybe `CompoundValidationValidator` should just be part of `CompoundValidator`?

dmolesUC (commented):
Apologies if some of the design questions I'm raising here are ones I really should have raised ten weeks ago on #7479

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
laritakr: As much as possible, it's always good to keep things backward compatible in the profiles (and in general), but it's always a balance of course. An incompatibility would  mean it's a major version (which we just did and don't want another one right away) and perhaps even a migration, and it's my opinion that we want to keep the effort low with regard to keeping your apps up to day.

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
laritakr: Very good point. I'm not sure how much time I have right now to rethink this... I just matched this validator with other existing validation structures when I contributed it back from one of our clients' work (because it seemed to be something that would be helpful to others). 

But I like the idea of a refactor to make it more of a plugin feature. And I think ALL of the schema validations could use a more flexible structure. Whether this change is approved now or later, that is a very useful refactor to work on. :)

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
laritakr: Inherently, I think there's a difference between validating the entire compound entry and validating the properties within a compound entry. So it does make sense in my mind to separate them. But your opposite perspective could just as easily be argued, and you're right that the name could definitely be more intuitive.

spec/services/hyrax/flexible_schema_validators/compound_validations_validator_spec.rb
laritakr: I preferred to build a more generic structure, knowing there WILL be more validations than just dates. This just happened to be the first one that a client requested.

spec/services/hyrax/compound_entry_validation_spec.rb
laritakr: I had some questions on what we DID want to validate. Our client is using a datepicker, but contributing it back led to needing a more generic date comparison. I'm open to feedback on what the desired behavior might be.

documentation/compound_fields.md
laritakr: The most important is to catch the errors in some way, but I'm not sure what's the best way either? This worked as a one-off validation requested by a client, but perhaps isn't the best way to start building a validator pattern.

documentation/compound_fields.md
laritakr: I just make it a practice to not try to break things that I don't _have_ to break, partly because Hyku generally uses Hyrax main branch instead of pinning to a released version.

laritakr:
Thanks for the detailed feedback @dmolesUC 

In my spare time (ha!) I'll do some thinking of how we could better structure this (and the other) m3 profile validations. If we're beginning a pattern, it's a good idea to build the right pattern, though I find it easier to find the right pattern when I have more than one thing to find the pattern between. 

Hoping for more feedback on how to better structure validations. What other validation use cases might be good to consider?

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
dmolesUC: That says to me that we want old profiles to work with new Hyrax, rather than new profiles to work with old Hyrax. Or am I misunderstanding?

app/services/hyrax/flexible_schema_validators/compound_validations_validator.rb
dmolesUC: Yeah, not being familiar with the underlying code it took me a fair bit to understand the distinction between the code for validating the schema and the code for validating objects against the schema. But given this is the first use of `Validation` as a noun I think it might be good to nip it in the bud. 🙃

spec/services/hyrax/flexible_schema_validators/compound_validations_validator_spec.rb
dmolesUC: Still, this spec could be scoped to just dates, it's not going to significantly change the amount of code you need to test the next validation, it's just going to change how it's nested.

documentation/compound_fields.md
dmolesUC: But as I noted above, it's not (at least if I understand the comments correctly) about trying to make sure newer code doesn't break old profiles, it's about trying to make sure newer profiles can run on older code. That seems like an unlikely scenario to me, and not worth adding code complexity for.

dmolesUC:
> Hoping for more feedback on how to better structure validations. What other validation use cases might be good to consider?

IDK, but my first instinct would be to look at other kinds of validations (Rails, Dry, whatever) and see if there are any models there -- not necessarily in the implementations, which are often pretty baroque, but in the interfaces they present to developers.

orangewolf:
@dmolesUC Is it fair to say there are two layers of feedback on this PR? some specific to the date validation themselves and others that are more broadly about the m3 profile validation? If that's the case, then I wonder how you'd feel about tackling the date validation level ones on this PR and then opening up a ticket and new work stream to do a broader refactor on the validations? I do think the needs for m3 profile validation are a bit emergent and that its hard to see what shape they need to take w/o some concrete examples, but it feels like we may be getting to a place where the design will be more apparent than when these validations were first conceived and I do agree with your comment about borrowing from the broader community. I'm just thinking that we don't want to refactor the whole validation engine on the PR about dates

PR #7581, opened 2026-08-12
Save a typed URL in a work_or_url compound field

### LINKED ISSUE #7572: compound metadata URLs don't update on dev.nurax
### Descriptive summary

When I edit a URL for a compound metadata field which allows URLs, the new values do not display on save. This is occurring on dev.nurax/dassie

Prereq: Compound metadata needs to be enabled on dev.nurax

1. Deposit a new work with a value in the `Relationship` compound metadata group
2. Edit the work that you deposited in step 1 and change the value in the `Item` field of the `Relationship` compound metadata group to a URL value and click `Save changes`.
3. Observe that the value has not update on the work show page in the `Relationships` section

Work show page displays values from first deposit and does not update on edit.

laritakr:
## Summary
Typing an external URL into a Relationship's `Item` field and saving now stores it. Previously the value was silently discarded and the old one kept. The form now includes two boxes to be able to appropriately search, enter, show, and display the work or url.
- Fixes #7572

## Details
A new work now shows both a search row AND a row to enter text. Edit shows the
work in the search box with the text box hidden when the stored value resolves to
an indexed work; otherwise the value appears in the text box, editable.
- On the show page: a value beginning http:// or https:// becomes a link; a value
  that resolves to a work links to that work by its title; anything else (including
  a scheme-less URL, or an id whose work is gone) renders as plain text.
- The search is a typeahead that omits the work itself, so it can't create a
  relationship to itself. It searches title, description, creator and keyword, and
  matches partial words in titles (so "jour stud" finds "Journal of Studies").
  Pasting a full work id also finds that work.

## Testing
- Edit a work with a Relationship row. Type a URL into the field below the search box — don't touch the dropdown — and save. The show page shows the new URL.
- Clear the field, save, and confirm the required-field error still fires.
- Search the picker by title or creator, click a result: the work's title shows in the picker and the show page links to the work.
- Paste a work id into the picker: it resolves and offers that work by title.
- Confirm the work you're editing does not appear in its own picker results.
- With a work selected, click the picker's × — the URL field reappears.
- Add a second Relationship row, type a URL, save, and confirm both rows persist.

## Screenshots

**A form with a work as a relationship**

Screenshot 2026-08-12 at 2 48 37 PM

**And empty form**

Screenshot 2026-08-12 at 2 48 45 PM

**A form with a URL as a relationship**

Screenshot 2026-08-12 at 2 49 49 PM


app/views/hyrax/compounds/_work_or_url_field.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    Locals: input_id, input_name, value, and optionally exclude_id (the work being
+    edited, kept out of the picker's own results). %>
+<% exclude_id = local_assigns[:exclude_id].presence %>
+<% current_option = compound_work_or_url_option(value) %>
+<%# Hide the URL field only for a value that resolves to an indexed work, which
+    compound_work_or_url_option signals with a title differing from the value.
+    Anything unresolved (a URL, a scheme-less "www.example.org/x", an id whose work
+    is gone) labels as itself and stays editable — testing "not a URL" instead would
+    hide a scheme-less URL, which is neither. %>
+<% work_selected = current_option.present? && current_option.first != value.to_s %>
```
orangewolf: these should be helpers please

orangewolf (changes requested):
the rest of this PR seems great and appears to do what it says it will do. Good work!

orangewolf (approved):

PR #7583, opened 2026-08-13
Fix compound metadata removal without restart

laritakr:
## Summary

Fix compound metadata removal without restart. If an app is booted with flexible metadata and includes compound metadata, removing the compound metadata from the m3 profile didn't take effect until restarting the app.

## Details

The first commit adds a failing spec that reproduces a compound metadata field continuing to render on work forms after it is removed from the m3 profile, until the app is restarted.

The second commit fixes the bug: flexible resources now read compound declarations from the loader's attribute map at the current profile version, rather than from the resource's Dry schemas. A compound removed from the m3 profile disappears from work forms immediately, with no restart.

The Dry schemas cannot represent a removal. `Hyrax::Flexibility.attributes` only merges attributes in, and builds the singleton schema from the class schema, so a compound present when the class loaded survives in both the class and singleton schemas — folded `subproperties:` and all — for the life of the process. Ordinary single-value properties were unaffected because `ResourceForm#initialize` already prunes definitions absent from the current profile; compounds bypass that prune, rendering via `#compound_terms` instead.

## Testing

- Boot the app with an m3 profile that declares a compound (a `type: hash` parent property with at least one member declaring `available_on: { properties: [<parent>] }`). Confirm the compound renders on a work's edit form.
- Without restarting, upload a profile with the compound parent and its subproperties removed.
- Reload the work's edit form. The compound field should be gone.
- Add the compound back via another profile upload, still without restarting, and confirm it reappears on the form.

### DIFF
--- a/app/services/hyrax/compound_schema.rb
+++ b/app/services/hyrax/compound_schema.rb
@@ -64,28 +64,55 @@ module Hyrax
     end
     private_class_method :flexible_attributes_for
 
-    # For a class, its own schema. For an instance, both the singleton schema
-    # (flexible mode) and the class schema (non-flexible mode), so {.for} works
-    # in both.
+    # A flexible instance resolves compounds from the loader's attribute map at
+    # the current profile version, NOT from its Dry schemas. Neither Dry schema
+    # can reflect a compound being removed from the profile:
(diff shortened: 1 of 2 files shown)


orangewolf (approved):
this works for me locally

PR #7586, opened 2026-08-15
Make the Helm chart's DB connection pool size configurable

maxkadel:
This is the non-breaking version. We could also make a breaking change and remove the `?pool=` in the url entirely, and it would fall back to the value in the `database.yaml`.

---

`DATABASE_URL`'s hardcoded `?pool=5` silently overrides a consuming app's own `database.yml` pool setting — apps have no way to control their real connection pool size. Hit this directly: `RAILS_MAX_THREADS=12` but only ever 5 real DB connections per pod, causing `ActiveRecord::ConnectionTimeoutError` under load.

Adds `postgresql.pool` (default `5`, unchanged behavior) and `postgresql.poolInUrl` (default `true`, unchanged behavior; set `false` to omit `?pool=` entirely so the app's own `database.yml` is authoritative). Both apply to `externalPostgresql` deployments too.

Verified via `helm template` across internal/external DB and all pool/poolInUrl combinations.

### DIFF
--- a/chart/hyrax/values.yaml
+++ b/chart/hyrax/values.yaml
@@ -122,13 +122,24 @@ externalPostgresql: {}
 #  port:
 
 # ActiveRecord connection pool size, written into DATABASE_URL's ?pool=
-# param. Size this to the app's own thread concurrency (e.g. Puma's
-# RAILS_MAX_THREADS, or a job runner's own thread count) — a pool smaller
-# than the thread count causes threads to queue for a connection even
-# when the database itself has headroom, and a pool much larger than
-# needed can exhaust a shared database's max_connections faster than
-# expected across multiple replicas/consumers.
+# param when dbPoolInUrl is true. Size this to the app's own thread
+# concurrency (e.g. Puma's RAILS_MAX_THREADS, or a job runner's own
+# thread count) — a pool smaller than the thread count causes threads to
+# queue for a connection even when the database itself has headroom, and
+# a pool much larger than needed can exhaust a shared database's
+# max_connections faster than expected across multiple replicas/consumers.
 dbPool: 5
+
+# Whether DATABASE_URL includes a ?pool= param at all. Defaults to true
+# to preserve existing behavior on upgrade. DATABASE_URL's ?pool= (when
+# present) is treated by ActiveRecord as authoritative, silently
+# overriding whatever pool: a consuming app's own database.yml computes
+# (e.g. from its own RAILS_MAX_THREADS) — which is surprising if you
+# don't know to look for it. If your app already sizes its own pool via
+# database.yml and you want that to be the single source of truth, set
+# this false so DATABASE_URL omits the param entirely and your app's own
+# config takes over unmodified. dbPool is ignored when this is false.
+dbPoolInUrl: true
 
 embargoRelease:
   enabled: true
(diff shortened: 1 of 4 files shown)


chart/hyrax/values.yaml
```diff
@@ -269,6 +269,10 @@ minio:
 
 postgresql:
   enabled: true
+  # ActiveRecord pool size, written into DATABASE_URL's ?pool= (ignored if poolInUrl is false). Applies to externalPostgresql too.
+  pool: 5
+  # DATABASE_URL's ?pool= overrides a consuming app's own database.yml. Set false to let that be authoritative instead.
```
orangewolf: I don't think this is true. I think its either in the URL or not set if DATABASE_URL exists

chart/hyrax/values.yaml
```diff
@@ -269,6 +269,10 @@ minio:
 
 postgresql:
   enabled: true
+  # ActiveRecord pool size, written into DATABASE_URL's ?pool= (ignored if poolInUrl is false). Applies to externalPostgresql too.
+  pool: 5
```
orangewolf, line 274: feels like this can be in 3 states:
- its a digit: `pool: 5` sets it to 5 (or whatever number)
- its blank: `pool: ` or really any non number (so `pool: -`) which should remove the `?pool=` from the url though as I said below, that probably doesn't do any good
- its missing (aka no `pool:` at all). This would default to 5 for backward compatibility.

orangewolf: if this is true poolInURL doesn't need to be here

maxkadel: Removed poolInURL & added test to verify

chart/hyrax/values.yaml
maxkadel: Updated comment and added RSpec test to verify behavior (this is validating normal Rails behavior, but it's a fast test on a somewhat confusing config so I think it's worth keeping, but happy to remove it if you think it's redundant). 

| # | Scenario | `database.yml` `pool:` | `ENV['DATABASE_URL']` | `url:` key in `database.yml` | Resolved pool | Winner |
|---|----------|:-----------------------:|------------------------|-------------------------------|:--------------:|--------|
| 1 | `DATABASE_URL` not set | `13` | *(unset)* | — | **13** | `database.yml`'s `pool:` |
| 2 | `DATABASE_URL` set, includes `?pool=` | `13` | `...db?pool=99` | — | **99** | `DATABASE_URL`'s `?pool=` |
| 3 | `DATABASE_URL` set, no `?pool=` | `13` | `...db` | — | **13** | `database.yml`'s `pool:` |
| 4 | `url:` key in `database.yml`, includes `?pool=` | `13` | *(unset)* | `...pg?pool=99` | **99** | `url:`'s `?pool=` |
| 5 | `url:` key in `database.yml`, no `?pool=` | `13` | *(unset)* | `...pg` | **13** | sibling `pool:` |
| 6 | `url:` key in `database.yml` **and** `DATABASE_URL` both set | `13` | `...envdb?pool=77` | `...yamldb?pool=55` | **55** | `url:` — `DATABASE_URL` ignored *entirely* (host/db too, not just pool) |

orangewolf (approved):

PR #7587, opened 2026-08-15
Allow setting of thumbnail on empty parent work

kirkkwang:
## Allow setting of thumbnail on empty parent work

image

image

---

958c45e6e71e3502ac9b78852c47844ab2463d62

Prior, if you have an empty parent work (no file set of its own) and add
child work(s) (with file sets on them), the Representative Media /
Thumbnail / Rendering input fields render on the edit form of the parent
work.  However, there's nothing to set them to, just empty dropdowns.
This commit allows the parent work's edit form to get all of the
descendant works' file sets as well as its own so they can be used in
the parent's fields.  This also changes #form_file_set_select_for to
return an array of arrays instead of a hash in case we have a situation
where multiple child works have the same file set label.  A hash would
have collapsed that down to one ambiguous option.  Now instead of just
0001.tif as the option it'll be "0001.tif (Volume 1)", the "Volume 1"
being the work's title.

The change in the `PcdmMemberPresenterFactory` is for efficiency's sake.
Previously, each call returning the work and file set presenters would
make a query to Solr.  The memoization fixes that and since in the
`WorkFormHelper` we're asking for both work and file set presenters, we
adjust the call to query for all of them and use Ruby to select/reject
what we need so only one query is needed.  A separate memoization was
added to the edit form because all three fields (Representative Media /
Thumbnail / Rendering) have the same options so we don't need to do
three separate queries to populate them.

The change to the views is so we get an empty option so it doesn't look
like the thumbnail is set when it really isn't.  Without the empty
option when a depositor visits the empty parent work, the thumbnail
looks like it's already selected which isn't actually true.

### DIFF
--- a/app/helpers/hyrax/work_form_helper.rb
+++ b/app/helpers/hyrax/work_form_helper.rb
@@ -84,21 +84,31 @@ module Hyrax
     end
 
     ##
-    # Constructs a hash for a form `select`.
+    # Constructs options for a form `select`.
     #
     # @param form [Object]
     #
-    # @return [Array<Hash{String => String}>] a map from file set labels to ids for
-    #   the parent object
+    # @return [Array<Array(String, String)>] label/id pairs for the work's own
+    #   file sets and all descendant works' file sets (legacy forms return their
+    #   own label => id Hash via #select_files)
     def form_file_set_select_for(parent:)
       return parent.select_files if parent.respond_to?(:select_files)
       return {} unless parent.respond_to?(:member_ids)
 
-      file_sets =
-        Hyrax::PcdmMemberPresenterFactory.new(parent, nil).file_set_presenters
+      @form_file_set_select_for ||= begin
+        factory = Hyrax::PcdmMemberPresenterFactory.new(parent, nil)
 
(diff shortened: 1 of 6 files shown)


app/helpers/hyrax/work_form_helper.rb
```diff
@@ -84,21 +84,31 @@ def form_progress_sections_for(*)
...
+        factory = Hyrax::PcdmMemberPresenterFactory.new(parent, nil)
 
-      file_sets.each_with_object({}) do |presenter, hash|
-        hash[presenter.title_or_label] = presenter.id
+        factory.file_set_presenters.map { |fsp| [fsp.title_or_label, fsp.id] } +
+          descendant_file_set_options(factory.work_presenters)
+      end
+    end
+
+    def descendant_file_set_options(work_presenters, seen: Set.new)
```
orangewolf, line 106: the select box already suffers from an issue where lots of attached file sets can make it slow to load. this continues that problem, but doesn't really make it worse

orangewolf (approved):
Discussed during 2026-08-19 tech call

PR #7589, opened 2026-08-17
Home content touch ups

kirkkwang:
### Before
image

### After
image

38b2bdfeaf6f10d4c2d6eeec62bceea56c583844

This commit will align the handle that displays to admins on the home
page featured works.  Previously the handle was over extended compared
to the card itself.

## Add confirmation when unfeaturing

image

1a6b838bf05c0604f99208be85544ab052070f59

Prior, when you unfeature a work from the home page it would immediately
remove it.  This commit will add a confirmation before the work gets
unfeatured.

## Make removing featured works look correct

a5c8569bb3ae1c5cf41f5ef281d6961fb7fe73db

### Before
featured-works

### After
featured-works-after

Prior, when a featured work was removed, the only thing that goes away
is the x button, the work card still remains until the user refreshes
the page.  This commit will make it so that the card goes away as well.
If the last featured work gets removed then the list returns back to the
original state where nothing has been added.

## Clean up admin featured works card styling

d10370786f673b4e2367d5f83a16f7c08567d90d

### Before
image

### After
image

This commit just cleans up the admin featured works card, notably there
was a PR to make the label NOT muted (WCAG reasons) but the admin side
of things was left out.

Ref:
- https://github.com/samvera/hyrax/commit/7014aea9f

## Match recent uploads tab with featured works

76cf2b1f4203547f939d73744a39dbfa97a0688b

### Before
image

### After
image

Previously, featured works tab was updated to fix some UI issues, this
commit will match the look with that.

### Before
image

### After
image

d93a9c6c0ac0b8a91f44bbec8d47d851a5fa1956

This commit will make the admin view of the featured works look like the
public view, so it uses the same size fonts.  Also the dragged ghost
state looks proper now as well.

### DIFF
--- a/app/assets/javascripts/hyrax/featured_works.js
+++ b/app/assets/javascripts/hyrax/featured_works.js
@@ -48,6 +48,9 @@ Blacklight.onLoad(function() {
   $('a[data-behavior="unfeature"]').on('click', function(evt) {
     evt.preventDefault();
     anchor = $(this);
+    var confirmText = anchor.data('confirmText');
+    if (confirmText && !confirm(confirmText)) { return; }
+
     $.ajax({
        url: anchor.attr('href'),
        type: "post",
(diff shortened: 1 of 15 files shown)


orangewolf (approved):
thank you for this

PR #7591, opened 2026-08-20
Handle missing object key in TrophyCleanupListener in Wings

bbpennel:
### Summary
When using Wings in hyrax 5.2.0 we are getting errors like the following every time an object is deleted:
```
Failed to delete trophies for z890s9938. These trophies might be orphaned.
        key not found: :object
```
This is probably preventing trophies from getting deleted in Wings, although for us it is just a nuisance since we don't allow trophies to be created.

Trophies should now be deleted in wings and there shouldn't be a warning, whether or not there are trophies associated with a work that is being delete.

notes-bugfix

### Detailed Description

This PR prevents the error by using the same approach used in ObjectLifecycleListener to ensure the event is a hash before trying to access optional keys, and then does an early return if there is no ID to save an unnecessary query.

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/services/hyrax/listeners/trophy_cleanup_listener.rb
+++ b/app/services/hyrax/listeners/trophy_cleanup_listener.rb
@@ -9,10 +9,12 @@ module Hyrax
       # @param [Dry::Events::Event] event
       # @return [void]
       def on_object_deleted(event)
+        event = event.to_h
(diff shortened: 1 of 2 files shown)


orangewolf (approved):
I wonder some times if the trophies are worth the bother. Thank you for the fix!

PR #7594, opened 2026-08-21
Add a reusable hyrax-nginx image and build it in CI

maxkadel:
bitnamilegacy/nginx has no free updated releases; rebase on nginxinc/nginx-unprivileged instead, keeping the Bitnami nginx chart's expected log/config paths so the Helm chart can keep using that chart unchanged.

Part of a stack — nothing in the Helm chart references this image yet.

Connected to #7599

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,6 +53,16 @@ jobs:
           tags: |
             ${{ env.REGISTRY }}/${{ env.REPO_LOWER }}/hyrax-worker-base:${{ env.TAG }}
             ${{ env.REGISTRY }}/${{ env.REPO_LOWER }}/hyrax-worker-base:${{ github.ref_name }}
+      - name: Build and push nginx
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          target: hyrax-nginx
+          push: true
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.REPO_LOWER }}/hyrax-nginx:${{ env.TAG }}
+            ${{ env.REGISTRY }}/${{ env.REPO_LOWER }}/hyrax-nginx:${{ github.ref_name }}
       - name: Build and push dassie
         uses: docker/build-push-action@v7
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -53,6 +53,16 @@ jobs:
           tags: |
             ${{ env.REGISTRY }}/${{ env.REPO_LOWER }}/hyrax-worker-base:${{ env.TAG }}
             ${{ env.REGISTRY }}/${{ env.REPO_LOWER }}/hyrax-worker-base:${{ github.ref_name }}
+      - name: Build and push nginx
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          target: hyrax-nginx
+          push: true
```
maxkadel, line 62: Would it be worth it to use a `cache-from` step here, or is the build so small it wouldn't make a real difference?

docker/nginx/nginx.conf
```diff
@@ -0,0 +1,29 @@
...
+
+    include       /etc/nginx/mime.types;
+    default_type  application/octet-stream;
+
+    access_log  /opt/bitnami/nginx/logs/access.log;
+
+    sendfile        on;
+    keepalive_timeout  65;
+
+    include /opt/bitnami/nginx/conf/server_blocks/*.conf;
```
maxkadel, line 28: Would this be breaking prior to updating our helm chart to use it, or would it be fine? Could we make it safer by making making an empty dummy file in the Dockerfile?

.github/workflows/build.yml
maxkadel: I think the layers we're adding to the base image are so tiny it doesn't make sense - it takes a few seconds to build locally.

docker/nginx/nginx.conf
maxkadel: Added an `nginx -t` check to make sure the nginx configuration is valid (only runs when there are changes to files related to the nginx build)

orangewolf (approved):
its great to check one more bitnami legacy images off the pile

maxkadel:

Issue #7339, opened 2026-02-02
Update image for nginx


maxkadel:
- [ ] Update to a non-bitnami image for nginx in chart/hyrax/values.yaml
- [ ] Ensure current functionality is maintained

PR #7595, opened 2026-08-21
Point the chart's nginx.image default at the new hyrax-nginx image

maxkadel:
Stacked on #7594. bitnamilegacy/nginx has no free updated releases. The nginx subchart has no Chart.AppVersion fallback for a blank tag (unlike the main image.tag), so it needs a real pinned value — adds a drift-check script + test so it can't silently fall out of sync with appVersion.

Not running in CI yet because of an unrelated issue with the fcrepo chart

Connected to #7339
Connected to #7599 


maxkadel:

PR #7596, opened 2026-08-21
Add Helm chart CI (ct lint + helm-unittest)

maxkadel:
Adds a chart/**-scoped workflow running `helm/chart-testing-action` (`ct lint`) and `helm-unittest`. Fixes two pre-existing yamllint nits so `ct lint` passes clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7597, opened 2026-08-21
Add global.hyraxFullnameOverride / hyrax.effectiveFullname

maxkadel:
Preparation step for getting nginx mounting defaults working: adds a global hyraxFullnameOverride so that the nginx chart will be able to find the correctly-named PVCs. 

`hyrax.fullname`'s `.Chart.Name` fallback resolves to the wrong chart name when re-rendered inside a subchart's own tpl context (confirmed while working on nginx.extraVolumes — it silently computed `RELEASE-NAME-nginx-uploads` instead of the real PVC name). Adds `global.hyraxFullnameOverride` as an explicit escape hatch and routes the PVC's existing consumers through it; default behavior for anyone not using nameOverride/fullnameOverride is unchanged (verified via test).

### DIFF
--- a/chart/hyrax/templates/_helpers.tpl
+++ b/chart/hyrax/templates/_helpers.tpl
@@ -25,6 +25,14 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 
 {{/*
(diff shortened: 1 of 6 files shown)


maxkadel:

PR #7598, opened 2026-08-21
Sane defaults for nginx extraVolumes/extraVolumeMounts

maxkadel:
Stacked on #7597. Mounts the same uploads PVC the main app already uses (system/uploads/uv/assets), using the previous PR's effectiveFullname override so the claimName can't drift from the real PVC. Does not touch nginx.image — that's stack 1's job (#7595).


maxkadel:

PR #7603, opened 2026-08-22
[Option B] Own nginx chart, no Bitnami dependency, no republished image (major bump)

maxkadel:
~**Either/or with #7606 — don't merge both.**~

Consensus seems to be for this option.

Drops the Bitnami nginx subchart entirely in favor of three chart-owned templates (Deployment/Service/ConfigMap) using our own image directly, with a sane default `serverBlock`/`nginx.conf`, `extraServerBlockConfig`/`extraHttpConfig` override hooks, and a couple of nginx+Rails performance defaults (upstream keepalive, gzip, open_file_cache).

Major version bump: `nginx.*`'s shape changes, so any Bitnami-specific value not reimplemented here (metrics, HPA/PDB, etc.) would silently stop doing anything for anyone who already has `nginx.enabled: true`.

Based on #7607.

**Validated against a real app's `staging` environment**: deployed a prerelease build end-to-end, including an in-place upgrade from an existing `nginx.enabled: true` release still on the Bitnami subchart.


chart/hyrax/values.yaml
```diff
@@ -295,143 +295,43 @@ postgresql:
...
-  #       client_body_buffer_size 32K;
-  #       client_max_body_size 0;
-  #       access_log /opt/bitnami/nginx/logs/access.log loki;
-  #       # if=$loggable;
-
-  #       sendfile on;
-  #       send_timeout 300s;
-
-  #       include /opt/bitnami/nginx/conf/bots.d/ddos.conf;
-  #       include /opt/bitnami/nginx/conf/bots.d/blockbots.conf;
```
orangewolf, line 389: it would be cool if we could keep the ability to add things like these 2 lines w/o having to override the whole file.

maxkadel: Agreed - you should be able to add to the `server` block with `extraServerBlockConfig`, which will add to the existing file, without overriding, or, if it needs to go at the top level instead of inside the server block (like with logs), you can put it in `extraHttpConfig`.

There are comments inline for that functionality, but do you think there's a better way to document it to make that clearer?

maxkadel: see `chart/hyrax/files/nginx/server-block.conf:39` for where it inserts the `extraServerBlockConfig`, and `chart/hyrax/files/nginx/nginx.conf:38` for where it inserts the `extraHttpConfig`.

[PR doing a test deploy to check functionality](https://github.com/notch8/palni_palci_knapsack/pull/726) (currently against a temporary test chart, until the Hyrax one is released).

orangewolf (dismissed):
I see now how this is supposed to work

chart/hyrax/values.yaml
orangewolf: I thought there was a 3rd case, but I can see now that inside the server block and outside the server block are the only two real cases and it doesn't matter which config file the "outside the server block" code ends up in.

orangewolf (approved):

PR #7609, opened 2026-08-25
Show IIIF viewer when child works are viewable

kirkkwang:
## Screenshot
image

### V2

image

### V3

image

## Summary

Previously, if we had an empty parent work that holds child works with viewable content, the IIIF manifest that gets created is a Collection manifest.  The manifest is viewable but Hyrax view logic did not render the IIIF viewer.  This commit will make it so that a viewer does render and child works are viewable, albeit through a Collection manifest.

---
_**🤖 Assisted-by: Opus 5**_

### DIFF
--- a/app/presenters/hyrax/work_show_presenter.rb
+++ b/app/presenters/hyrax/work_show_presenter.rb
@@ -11,16 +11,11 @@ module Hyrax
     attr_writer :member_presenter_factory
     attr_accessor :solr_document, :current_ability, :request
 
-    class_attribute :collection_presenter_class, :presenter_factory_class, :iiif_viewer_max_child_depth
+    class_attribute :collection_presenter_class, :presenter_factory_class
 
     # modify this attribute to use an alternate presenter class for the collections
     self.collection_presenter_class = CollectionPresenter
     self.presenter_factory_class = MemberPresenterFactory
-
-    # how many member levels #iiif_viewer? descends into when a work has no
-    # representative of its own; bounds the number of Solr queries an empty, deeply
-    # nested collection can trigger to one per direct child
-    self.iiif_viewer_max_child_depth = 1
 
     # @param [SolrDocument] solr_document
     # @param [Ability] current_ability
@@ -375,15 +370,8 @@ module Hyrax
         (av_viewable? || image_viewable? || pdf_viewable?)
     end
 
-    # depth lives in Thread.current, not an ivar, since each recursive call
-    # builds a new presenter instance
     def child_works_viewable?
-      depth = Thread.current[:hyrax_iiif_viewer_child_depth] ||= 0
-      return false if depth >= iiif_viewer_max_child_depth
-      Thread.current[:hyrax_iiif_viewer_child_depth] = depth + 1
-      work_presenters.any?(&:iiif_viewer?)
-    ensure
-      Thread.current[:hyrax_iiif_viewer_child_depth] = depth
+      Hyrax::ViewableChildWorksService.viewable?(solr_document: solr_document, ability: current_ability)
     end
   end
 end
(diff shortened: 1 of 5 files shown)


app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -85,10 +85,12 @@ def download_url
     end
 
     # @return [Boolean] render a IIIF viewer
```
maxkadel, line 87: The name of the memo (`@iiif_viewer`) is a little confusing - it seems like it should be an instance of a some kind of IiifViewer object, rather than a boolean representing whether or not a IIIF viewer should be displayed. Would rubocop get mad if you changed it to `@display_iiif_viewer` or something like that?

Also, since as far as I can tell this is part of the public api, what are the implications of changing the signature? I know any caller of the method would not be broken because of the default value, ~but if someone was overriding and calling `super`, would that now be broken? (I'm going to look into this more so that I understand it better, but I wanted to "show my work")~

app/presenters/hyrax/work_show_presenter.rb
```diff
@@ -364,5 +366,12 @@ def pdf_viewable?
       return false unless Flipflop.iiif_pdf?
       representative_presenter.pdf?
     end
+
+    def child_works_viewable?(seen)
+      work_presenters.any? do |presenter|
+        next unless seen.add?(presenter.id)
+        presenter.iiif_viewer?(seen: seen)
```
maxkadel: This works if the child work uses a WorkShowPresenter, but if it uses a different presenter that doesn't have the new signature, it will break.

app/presenters/hyrax/work_show_presenter.rb
maxkadel: Ok, so I looked into it more, and the problem is not a method calling `super` on this method, but if there's another *presenter* implementing `iiif_viewer?`, that then the `child_works_viewable?` would call with the `seen` argument, that would raise an `ArgumentError`. Working on a test that fails for the right reason.

maxkadel (changes requested):
I really like the idea of this.

I think there is a breaking change in here that we can fix - I put a test for it and a proposed fix on a PR against this PR - https://github.com/samvera/hyrax/pull/7610.

maxkadel (dismissed):
Thanks @kirkkwang! Looks good

app/presenters/hyrax/work_show_presenter.rb
kirkkwang: Nice catch, that's a great point!

kirkkwang:
Thanks @maxkadel for the help, that's much better!  I'd like to get one more pair of eyes on it because even though the manifest was always being generated, the viewer wasn't rendering and that is a change in behavior.

I'd love to get @cjcolvar's thoughts if he has time.  The ultimate goal is to add IIIF Ranges into Hyrax and make that a feature flipper.  When flipped on, instead of the default behavior of generating a Collection manifest that references child work manifests, it would generate a Manifest manifest with a table of contents for each child work.

app/services/hyrax/viewable_child_works_service.rb
```diff
@@ -0,0 +1,70 @@
...
+      @member_ids ||= Array(solr_document.member_ids)
+    end
+
+    def viewable_member_count
+      solr_service.query_result(join_query, fq: filter_queries, rows: 0)
+                  .dig('response', 'numFound')
+    end
+
+    def join_query
+      "{!join from=hasRelatedMediaFragment_ssim to=id cache=false}{!terms f=id}#{member_ids.join(',')}"
```
maxkadel, line 43: I would love some of these methods to be less Solr-specific / more abstracted, and also not going to block on that.

maxkadel (approved):
Looks good - my preference would be for more abstracted queries for the Solr clauses, but I'm also ok with them as-is, especially since they're in private methods we can update later.

app/services/hyrax/viewable_child_works_service.rb
kirkkwang: good idea, as a follow up PR i can abstract these into the `Hyrax::SolrQueryService`, at least that way it can be reusable elsewhere

PR #7611, opened 2026-08-25
Remove dead view partials

kirkkwang:
## Remove dead collection work action menu partials

98a92c36aa82440490cd6075326e15c0eccaa582

These partials are no longer being rendered from anywhere.  The
2a4646784 commit stopped rendering _form_default_group (renamed to
_form_default_group_delt) from _document_list, which rendered
_list_works, which rendered _work_action_menu.  The commit also dropped
`render 'edit_actions'` from edit.html.erb.

Ref
- https://github.com/samvera/hyrax/commit/2a4646784

## Remove _work_button_row

61238522fd0b082314f126e83b98e1a75a4a585e

This partial has never been rendered from anywhere.  The 613c4bce0
commit added it but show.html.erb renders _show_actions instead, so no
render call ever existed.

Ref
- https://github.com/samvera/hyrax/commit/613c4bce0

## Remove unused file set flash partials

acf468e74300b02b08d479966f3815da1d0c1e9a

These partials are no longer being rendered from anywhere.  The
e3b80fa4b commit removed the commented out render_to_string calls for
both, back when they lived in generic_files, and left the partials
behind.

Ref
- https://github.com/samvera/hyrax/commit/e3b80fa4b

## Remove leftovers from Blacklight catalog

e753c383f8e0d8cf7fb3525958c26977f6933d7d

These partials are no longer being rendered from anywhere.  The
180a22953 commit copied over Blacklight's catalog partials.

Ref
- https://github.com/samvera/hyrax/commit/180a22953

---
_**🤖 Assisted-by: Opus 5**_


orangewolf (changes requested):
thanks for the clean up. please think through and select an appropriate notes label for this PR. I can't decide if it's just a patch or if it's a major breaking change just from the info here. Also, I wonder if the _delt partial should also be removed since it says its not used?

kirkkwang:
@orangewolf yeah i think let's be safe and call it a major, though no call within Hyrax calls these I guess we can't be sure there are apps out there that call these partials.  Oh and the _delt partial was already deleted in https://github.com/samvera/hyrax/commit/06b05ed93

PR #7613, opened 2026-08-26
Fixes a bug where fits characterization silently fails

maxkadel:
### Fixes
Fixes bug discovered by @kirkkwang, described in https://samvera.slack.com/archives/C0F9JQJDQ/p1785199309995839. 

### Summary
When fits fails to characterize something, it still returns a 200 and an xml document, it's just that that xml document explains the error. We should raise and log the error, instead of swallowing it.

* Upload a file over 2 gigs to an application using the `ghcr.io/samvera/fitsservlet` docker image, watch background jobs.
  * *Before the fix*: The background job for characterization will succeed, but the file will not have been characterized or gotten derivatives created
  * *After the fix*: The background job will fail with a clear error message

* Pass errors in fits on to the background job - when fits fails, the characterization job should fail

@samvera/hyrax-code-reviewers

### DIFF
--- a/app/jobs/valkyrie_characterization_job.rb
+++ b/app/jobs/valkyrie_characterization_job.rb
@@ -1,6 +1,13 @@
 # frozen_string_literal: true
 class ValkyrieCharacterizationJob < Hyrax::ApplicationJob
   queue_as Hyrax.config.ingest_queue_name
+
+  # The characterization tool reported a definitive failure (e.g. a file
+  # exceeding its configured size limit); retrying won't change the outcome.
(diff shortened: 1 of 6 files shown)


orangewolf (approved):

orangewolf:
this passed in my evaluation of it. my only concern is if there are fits cases where one tool fails and others pass and the doc is still an error. I couldn't prove that it works that way, but I could figure out how to prove that it doesn't work that way either. I think being noisy about it is better than silent, so I'm moving this forward

PR #7616, opened 2026-08-28
Fix facet for a name: surrogate property

laritakr:
## Summary
Fixes a metadata property that used `name:` to stand in for another. If the property was facetable or searchable, it would show an always-empty facet or search-results column.

## Details
- A profile property can declare `name:` to represent a different attribute for a particular class. The attribute named is what gets indexed.
- Registration derived every Solr field name from the property's own profile key, so the facet and column pointed at fields nothing is ever written to and listed no values for any search.
- Registration now derives from the attribute actually indexed, and removes only the surrogate's own keys — a surrogate that declares the target attribute's field names no longer deletes the target's registration.
- Only affects profiles that use `name:` surrogates; installations without one see no change.

## Testing
- In an installation whose m3 profile has a property declaring `name:` for another attribute on a different class, confirm the sidebar no longer offers a facet under the surrogate's own name.
- Confirm the facet for the attribute it stands in for is present and lists values.
- Reload a catalog page several times without restarting the server, and confirm no error.

#### Sample M3 addition for testing

The two entries must be on **disjoint** classes — the profile validator rejects two properties sharing a resolved name on overlapping classes.

```yaml
  edition:
    available_on:
      class:
      - GenericWork
    cardinality:
      minimum: 0
    data_type: array
    display_label:
      default: Edition
    indexing:
    - edition_sim
    - edition_tesim
    - facetable
    property_uri: http://purl.org/ontology/bibo/edition
    range: http://www.w3.org/2001/XMLSchema#string
  monograph_edition:
    name: edition
    available_on:
      class:
      - Monograph
    cardinality:
      minimum: 0
    data_type: array
    display_label:
      default: Edition
    indexing:
    - edition_sim
    - edition_tesim
    - facetable
    property_uri: http://purl.org/ontology/bibo/edition
    range: http://www.w3.org/2001/XMLSchema#string
```

### DIFF
--- a/app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
+++ b/app/controllers/concerns/hyrax/flexible_catalog_behavior.rb
@@ -19,6 +19,20 @@ module Hyrax
           indexing = prop['indexing']
           next if indexing.nil?
 
+          # A property may declare `name:` to stand in for another on a
+          # particular class. The named attribute is what gets indexed, so it is
+          # the name every solr field derives from — registering under the
+          # surrogate's own key produces a facet nothing is indexed into.
+          indexed_name = indexed_name_for(itemprop, prop)
+          if indexed_name != itemprop
+            # Only the surrogate's own keys. Passing its `indexing:` array would
+            # fold in any field names it declares — which for a surrogate are
+            # the target attribute's — and delete the target's registration.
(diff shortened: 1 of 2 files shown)


orangewolf (approved):
This is a great fix LaRita. thank you

PR #7618, opened 2026-08-31
Index controlled vocabulary labels

laritakr:
## Summary
Adds an opt-in seam for resolving controlled-vocabulary term ids to their labels. **No behavior changes unless an application registers a label service** — Hyrax ships no resolver, and the default returns values unchanged.
- Refs samvera/hyku#3254

## Details

**What is unchanged.** With no label service registered — which is every Hyrax installation today — indexing, show pages, search results, and facets are byte-identical to before. Shipped m3 profiles also declare `sources: ['null']` everywhere, and the properties Hyrax does treat as controlled (`license`, `rights_statement`, `resource_type`) get that from dedicated edit-field partials and service classes rather than the profile. This PR does not touch those.

**What changes, and only then.** Behavior changes for a property that meets all three conditions:

1. Its `controlled_values.sources` names an authority (not the `'null'` sentinel).
2. A registered label service reports that authority resolvable.
3. Its `indexing:` array names the Solr fields themselves — `stored_searchable` / `facetable` alone carry no field name, so such a property is left alone.

For those properties: labels are indexed as `<property>_label_*` beside the untouched ids, the show page and search row display the label, the facet reads labels, and the label field joins free-text search.

**Design notes for review.**
- Ids are never replaced — they are the link target for a URI-valued authority and what OAI harvests.
- Row and facet move together, because Blacklight queries a facet with whatever the row displayed.
- The id facet stays configured but hidden, so saved searches and bookmarks carrying `f[<property>_sim][]` keep resolving; it is restored if the property stops being controlled.

**Unrelated change**
Gitignores uploads the koppie and dassie dev stacks write into the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,10 @@ db/schema.rb
 .byebug_history
 .tool-versions
 
+# Files uploaded through the app while developing against the test stacks
+/.koppie/uploads/
+/.dassie/uploads/
+
 ## Carried over from hyrax-models (RIP)
 *.gem
 *.rbc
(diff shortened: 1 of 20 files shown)


orangewolf (dismissed):
LaRita and I have done several rounds of trying to make this simpler. I'd love to see a more straightforward implementation of this, but do not see a way to do so that keeps the flexibility we need.

orangewolf (approved):

PR #7626, opened 2026-09-03
Prune schema fields the flexible resource cannot answer

ShanaLMoore:
A tenant on flexible metadata whose M3 profile does not declare `redirects` gets a 500 on work and collection `new` and `edit`. Hyrax's own `config/metadata_profiles/m3_profile.yaml` does not declare it, so `HYRAX_FLEXIBLE=true` with `HYRAX_REDIRECTS_ENABLED=true` reproduces this with no adopter customisation. `redirects` is installed onto the forms from `config/metadata/redirects.yaml` on the env flag alone, while the matching resource attribute comes from the profile, and the two can disagree.

`ResourceForm#initialize` already pruned definitions the current schema does not declare, but only ones marked `display`. `redirects.yaml` sets `form: display: false`, so the field could be neither found nor pruned, and Disposable then read it off a resource with no such attribute.

Worth arguing with:

- Candidacy is keyed on the definition carrying a `:display` option at all, which is what separates a metadata-schema field from the form's structural properties (`permissions`, `embargo`, the `<name>_attributes` writers). Those carry no `:display` key and must survive despite the resource not answering them. Keying on `respond_to?` alone prunes them and breaks the form.
- The deprecated positional-argument path now shares the same helper. Its comment already claimed it mirrored the keyword path; it did not.

Testing:

`bundle exec rspec ./spec/forms` gives 288 examples, 0 failures, 18 pending. Rubocop clean on both changed files.

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -93,12 +93,7 @@ module Hyrax
         if resource.nil?
           if !deprecated_resource.nil?
             Deprecation.warn "Initializing Valkyrie forms without an explicit resource parameter is deprecated. Pass the resource with `resource:` instead."
-            # Remove form definitions for attributes the model doesn't support,
-            # mirroring the cleanup in the resource: keyword path below.
-            if deprecated_resource.respond_to?(:flexible?) && deprecated_resource.flexible?
-              to_remove = singleton_class.definitions.select { |k, v| !deprecated_resource.respond_to?(k) && v.instance_variable_get("@options")[:display] }
-              to_remove.keys.each { |removed_field| singleton_class.definitions.delete(removed_field) }
-            end
+            prune_unbacked_definitions!(deprecated_resource, current_schema_fields.keys.map(&:to_s)) if deprecated_resource.respond_to?(:flexible?) && deprecated_resource.flexible?
             super(deprecated_resource)
           else
             super()
@@ -109,12 +104,7 @@ module Hyrax
             hash = resource.attributes.dup
             hash[:schema_version] = Hyrax::FlexibleSchema.current_schema_id
             resource = resource.class.new(hash)
-            # find any fields removed by the current schema
-            current_field_keys = current_schema_fields.keys.map(&:to_s)
-            to_remove = singleton_class.definitions.select { |k, v| !current_field_keys.include?(k.to_s) && v.instance_variable_get("@options")[:display] }
-            to_remove.keys.each do |removed_field|
-              singleton_class.definitions.delete(removed_field)
-            end
+            prune_unbacked_definitions!(resource, current_schema_fields.keys.map(&:to_s))
           end
 
           super(resource)
@@ -301,6 +291,20 @@ module Hyrax
       delegate :flexible?, to: :model
 
       private
+
+      # Only metadata-schema definitions carry a `:display` option. The form's
+      # own structural properties (`permissions`, `embargo`, the
+      # `<name>_attributes` writers) do not, and must survive even though the
+      # resource has no matching attribute.
+      def prune_unbacked_definitions!(resource, current_field_keys)
+        removable = singleton_class.definitions.select do |name, definition|
+          next false if current_field_keys.include?(name.to_s)
+          options = definition.instance_variable_get("@options")
+          next false unless options.key?(:display)
+          options[:display] || !resource.respond_to?(name)
+        end
+        removable.each_key { |name| singleton_class.definitions.delete(name) }
+      end
 
       # A copy of the resource loaded at the current schema version, so its
       # singleton schema reflects the live profile (including compounds added
--- a/spec/forms/hyrax/forms/resource_form_spec.rb
+++ b/spec/forms/hyrax/forms/resource_form_spec.rb
@@ -285,6 +285,69 @@ RSpec.describe Hyrax::Forms::ResourceForm do
     end
   end
 
+  describe 'flexible form with a field the profile never declared' do
+    # `redirects` is the shipped case: installed from config/metadata whenever
+    # `Hyrax.config.redirects_enabled?`, never from the profile. Stated without
+    # it so the example runs regardless of HYRAX_REDIRECTS_ENABLED.
+
+    let(:current_form_defs) do
+      { 'title' => { required: true, primary: true, display: true } }
+    end
+
+    let(:schema_loader) do
+      loader = instance_double(Hyrax::M3SchemaLoader)
+      allow(loader).to receive(:current_version).and_return(1)
+      allow(loader).to receive(:attributes_for).and_return({})
+      allow(loader).to receive(:form_definitions_for) { current_form_defs }
+      allow(loader).to receive(:index_rules_for).and_return({})
+      loader
+    end
+
+    let(:work_class) do
+      klass = Class.new(Hyrax::Work) do
+        def self.name
+          'TestUndeclaredFieldWork'
+        end
+      end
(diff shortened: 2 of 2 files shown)


ShanaLMoore:
Sharper on the root cause, having traced the two gates.

The model and the form gate the same feature on two different config methods, and only one of them is flexible-aware:

- `app/models/hyrax/work.rb:99` — `include Hyrax::Schema(:redirects) if Hyrax.config.work_default_metadata && Hyrax.config.redirects_enabled?`
- `lib/hyrax/configuration.rb:1079` — `work_default_metadata` is `flexible? ? false : true`
- `app/forms/hyrax/forms/pcdm_object_form.rb:20` — `include Hyrax::FormFields(:redirects) if Hyrax.config.redirects_enabled? && Hyrax.config.work_include_metadata?`
- `lib/hyrax/configuration.rb:1068` — `work_include_metadata` is `disable_include_metadata ? false : true`, independent of `flexible?`

So in flexible mode with `HYRAX_DISABLE_INCLUDE_METADATA` unset, the model skips the redirects schema while the form installs it. The resource can then only answer `redirects` if the adopter's M3 profile declares it, and Hyrax's own default profile does not. `PcdmCollectionForm:20` has the same shape via `collection_include_metadata?`.

That suggests a second, narrower fix: gate the form include the way the model is gated. I have left this PR on the prune because the prune also covers the schema-shrank case and any other statically wired field, but if maintainers prefer aligning the gates I am happy to switch or to add it.

Note on evidence: the regression examples here reproduce the prune gap with a synthetic `display: false` property, not with `redirects` itself. Exercising `redirects` needs a genuinely flexible boot, and koppie does not currently come up with `HYRAX_FLEXIBLE=true` (`Dry::Struct::RepeatedAttributeError: Attribute :date_modified has already been defined` from `.koppie/app/models/monograph.rb:8`). The production trace behind this report is `NoMethodError (undefined method \`redirects' for #<GenericWorkResource>)` on a Hyku 7 tenant.

ShanaLMoore:
Reproduced end to end on a multitenant Hyku 7 app, and my previous comment's root-cause theory was wrong. Retracting it: the knapsack in question sets `HYRAX_DISABLE_INCLUDE_METADATA=true` whenever flexible is on, so `work_include_metadata?` is false there and the static `FormFields(:redirects)` include never happens. The two-gate asymmetry is real in the code but is not what fires here.

The actual trigger is **cross-tenant**, and it is an asymmetry in when the model and the form resolve the flexible schema:

- the resource class re-bakes its attributes from the **current tenant's** schema, so `respond_to?(:redirects)` tracks the tenant correctly
- the form class's flexible fields come from `check_if_flexible` doing a class-level `include Hyrax::FormFields(model.to_s, definition_loader: m3_schema_loader)`, baked **once per process** from whichever tenant was served first

Serve a tenant whose profile declares `redirects`, then a tenant whose profile does not, and the form carries `redirects` while the model has no such attribute. Because `redirects.yaml` is `display: false`, the prune could not remove it.

Verified on Hyku 7, `HYRAX_FLEXIBLE=true`, `HYRAX_REDIRECTS_ENABLED=true`, two tenants (A declares `redirects`, B does not), alternating form builds in one process.

Before, on this branch's merge-base (`resource_form.rb` byte-identical to `main`):

```
A form_ok model=true  defs_redirects=true
B RAISED NoMethodError: undefined method `redirects' for #<GenericWorkResource:0x0000ffff46883a88>
A form_ok model=true  defs_redirects=true
B RAISED NoMethodError: undefined method `redirects' for #<GenericWorkResource:0x0000ffff3c2d77b0>
```

After, same app, only `resource_form.rb` swapped for this branch's version:

```
A form_ok model=true  defs_redirects=true
B form_ok model=false defs_redirects=false
```

Single-tenant profile edits never reproduce it: I added and removed `redirects` from one tenant's profile in both directions and the model tracked the change each time, so model and form stayed consistent. It takes two tenants that disagree.

laritakr:
## The prune isn't reaching the real problem

I dug into this while investigating the reported Hyku bug, and I think this PR is treating a symptom. Sharing what I found.

### The form class is being used as scratch space

A form's field list lives in one place per form class (`definitions`). Only about 30 of those are the form's own static properties (`permissions`, `embargo`, `member_ids`, and the `*_attributes` writers). **Every metadata field — `title`, `creator`, `redirects` — is absent from the class and added fresh on each build** from the M3 profile.

The catch is that those per-build fields don't land on the instance. They land on the class. `singleton_class.schema_definitions = self.class.definitions` makes the two the same object, because Declarative resolves `definitions` as `@definitions ||= ...` on whichever object receives it. So the "per-instance" schema has always been the class's, mutated per request.

That's invisible in a single-profile install. Where profiles differ it isn't: whichever profile is built first leaves its fields on the class, and the next form inherits them.

### Why `respond_to?` doesn't fire

`ResourceForm#initialize` rebuilds the resource at the current schema version before `super`, and `Hyrax::Flexibility.load` re-applies the schema per instance. I checked this directly — a work whose stored `schema_version` is stale answers `false` to `respond_to?(:redirects)`, but the *rebuilt* resource answers `true`.

So by the time the prune runs, the resource can answer anything the current profile declares. I removed the `respond_to?` clause entirely and could not get a crash out of it in either flex mode.

### What is actually broken

The existing prune on main already drops fields the current profile no longer declares — but only ones marked `display`. That is the real gap, and it is exactly `redirects`, which `redirects.yaml` sets to `display: false`.

Demonstrated with one hidden field, two builds, second profile doesn't declare it:

    display: true   -> build2: class has field? false   (existing prune handles it)
    display: false  -> build2: class has field? true    (leaks)

So this PR is right that `display: false` is the hole. It's the framing I'd question: the fix isn't "prune fields the resource can't answer" (the resource can answer them), it's "don't leave one profile's fields on the shared class."

### Reproduced against the Hyku bug

Two tenants whose profiles genuinely disagree — `cat` declares `redirects`, `dog` does not — serving both in one process against the real `GenericWorkResourceForm`:

    start:                     defs=32  has_redirects=false
    CAT build (declares yes):  defs=69  has_redirects=true   singleton==class? TRUE
    DOG build (declares NO):   defs=68  has_redirects=true   singleton_redirects=TRUE

`dog` inherits `redirects` from `cat`. That's the 500 in the reported bug.

### What I'd suggest instead

Reset the class's field list back to its static baseline at the start of each flexible build, so the previous profile's fields can't survive into the next one. It's ~10 lines of production code, needs no `respond_to?` probing, and fixes the hidden-field case that motivated this PR.

I've put that up as https://github.com/samvera/hyrax/pull/7627 — happy to fold it into this branch instead if you'd rather keep the discussion in one place.

With that change, `dog` reports `singleton_redirects=false` and the class settles at 69/67 instead of drifting, stable across repeated alternation in both directions. `spec/forms` is green in both flex modes (allinson 269, koppie 288).

The specs there deliberately use a `display: false` field: with a visible one they pass without the fix, since the pre-existing prune already handles that case. That distinction is the whole bug, so it's worth keeping explicit in any version of this.

Worth noting the deeper issue remains: profile fields being written to the class at all. Fixing that properly means making the per-instance schema genuinely per-instance and updating `required?`/`form[]`/`expose_class` to match — I tried it and it breaks 9 specs, so it's separate work.

laritakr:
Going to re-close this in favor of https://github.com/samvera/hyrax/pull/7627

The refactor of the pruning is a nice touch but I don't think the `respond_to?` check fires — the resource gets rebuilt at the current schema version before the prune runs, so it can always answer. So I took the `display: false` lead from here and cleared the leftover fields off the form class instead.

PR #7627, opened 2026-09-04
Keep one profile's fields off the form class

laritakr:
## Summary
Fixes work and collection forms picking up metadata fields from a different metadata profile than the one they were built for, which could make a form fail to load.
- Refs notch8/hyku-community-issues#154

## Details
- Flexible form fields are declared on the instance's singleton, but `Declarative` resolves `definitions` on whichever object receives it, so the singleton and the form class share one container.
- Each build therefore writes its profile's fields onto the shared class, where the next form built inherits them.
- The pre-existing cleanup only drops fields marked `display`, so a hidden field — `redirects` is the shipped one — survived onto the class even for a profile that never declared it.
- A flexible build now resets the class to its own static properties first. ~10 lines of production code; no `respond_to?` probing.
- Verified against a multi-profile reproduction: the form class stops accumulating fields, and each build sees only what its own profile declares.
- Known follow-up, out of scope here: `required_fields=` mutates `Definition` objects in place, which reaches the class the same way.

### DIFF
--- a/app/forms/hyrax/forms/resource_form.rb
+++ b/app/forms/hyrax/forms/resource_form.rb
@@ -76,6 +76,7 @@ module Hyrax
         r = resource || deprecated_resource
         if r.flexible?
           self.class.deserializer_class = nil # need to reload this on first use after schema is loaded
+          reset_flexible_definitions!
           singleton_class.schema_definitions = self.class.definitions
           contexts = r.respond_to?(:contexts) ? r.contexts : nil
           current_schema_fields = Hyrax::Schema.m3_schema_loader.form_definitions_for(schema: r.class.name, version: Hyrax::FlexibleSchema.current_schema_id, contexts: contexts)
@@ -199,6 +200,14 @@ module Hyrax
           required_fields
         end
 
+        ##
+        # The form's own properties. Memoized on the first call, which
+        # {ResourceForm#reset_flexible_definitions!} makes before any profile
+        # field has been added, so the snapshot excludes them.
+        def static_schema_definitions
+          @static_schema_definitions ||= definitions.dup
+        end
+
         def schema_definitions
           @definitions
         end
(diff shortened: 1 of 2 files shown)


ShanaLMoore:
Verified against the Hyku reproduction from notch8/hyku-community-issues#154. Two tenants whose profiles disagree, real `GenericWorkResourceForm`, one process, fresh process per run, only `resource_form.rb` swapped between runs.

Order A, B, A, B (A declares `redirects`, B does not):

```
main   A ok / B RAISED NoMethodError: undefined method `redirects' / A ok / B RAISED
7627   A ok / B ok                                                / A ok / B ok
```

Order B, A, B, A:

```
main   B ok / A ok / B RAISED NoMethodError: undefined method `redirects' / A ok
7627   B ok / A ok / B ok                                                / A ok
```

Field sets are right, not just non-crashing: A keeps `redirects` (62 definitions), B does not have it (61). On `main` the class stays at 62 once A has been served, which is the leak this PR removes.

The second block shows why it presents as intermittent: B is fine until A has been served on that worker.

I also checked the memoization risk in `static_schema_definitions`, that a polluted baseline would make the reset a no-op. It does not happen: the class has 31 definitions (statics only) at first call in every run, including when the first build is tenant B.

Your diagnosis is the better one and this is the right fix. One correction for the record: 7626 does not error on this reproduction. I ran it through the identical harness and both tenants passed, because the resource answers `respond_to?(:redirects)` false in tenant B. Your point likely holds for the stale `schema_version` case, where the rebuild does make the resource answer, which is a different scenario. Does not change which PR should land.

ShanaLMoore:
Verified against the Hyku repro in notch8/hyku-community-issues#154. Two tenants whose profiles disagree on `redirects`, served in one process, real `GET /concern/generic_works/:id/edit`:

```
main   A 200 / B 500 NoMethodError in Hyrax::GenericWorksController#edit
7627   A 200 / B 200
```

Matches the production report. Field sets are right too: A keeps `redirects`, B does not, and the class stops accumulating. Holds in both tenant orderings. Your diagnosis is the correct one.

One note so it is not carried forward: 7626 did not error on this repro, the resource answers `respond_to?(:redirects)` false in tenant B. Likely true for the stale `schema_version` case instead. Does not change which PR lands.

orangewolf (approved):
This makes sense to me and though I was only able to reproduce it in production mode, it did seem to fail and then work consistently in the before/after of this change.

PR #7632, opened 2026-09-07
Pin json below 3.0 to fix broken CI

maxkadel:
`json 3.0.0`'s `JSON.generate` dropped the `quirks_mode` keyword that `ActiveSupport::JSON::Encoding` still passes, raising `ArgumentError: unknown keyword: quirks_mode` on essentially any ActiveRecord JSON serialization. This broke CI across every test app (allinson/dassie/koppie/sirenia) today, unrelated to any in-flight PR. Pins `json` below 3.0 in the gemspec.


cjcolvar (commented):
Thanks for fixing this!  Agree that the `Gemfile`s feel like the best place to handle this.  Do you think it should also be added to the top-level `Gemfile` for anyone doing development outside of docker or any tasks that load rails outside of docker?

cjcolvar (approved):

PR #7635, opened 2026-09-09
Add a QA-backed vocabulary label resolver

laritakr:
## Summary
Ships a working controlled-vocabulary label resolver backed by the local Questioning Authority vocabularies in `config/authorities/`. **Not installed by default** — nothing changes until an application registers it.

Refs #7618 to add a default resolver

## Details
- Register it in `config/initializers/hyrax.rb`:
  `Hyrax.config.controlled_vocabulary_label_service = Hyrax::QaControlledVocabularyLabelService.new`
- A property's `controlled_values.sources` entry is matched to a local subauthority of the same name. Remote authorities (Geonames, LOC) are never resolved — that would mean a network request per value during indexing.
- Labels are read one map per authority rather than one lookup per value: a file-based authority re-parses its YAML on every call, which measured ~640× slower for indexing.
- Reindex after registering; the show page and catalog read the label fields the indexer writes.

## Testing
- Register the service, add a real authority name to a property's `controlled_values.sources`, reindex, and confirm the term's label appears on the show page, in search rows, and in the facet.
- Confirm an installation that registers nothing sees no change.

### DIFF
--- a/app/services/hyrax/qa_controlled_vocabulary_label_service.rb
+++ b/app/services/hyrax/qa_controlled_vocabulary_label_service.rb
@@ -1,0 +1,111 @@
+# frozen_string_literal: true
+
+module Hyrax
+  ##
+  # Resolves controlled vocabulary term ids to labels using the local
+  # Questioning Authority vocabularies an application has configured in
+  # `config/authorities/`.
+  #
+  # Not installed by default — Hyrax ships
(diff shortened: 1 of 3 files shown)


orangewolf (dismissed):
discussed on several calls including the 9/18/2026 office hour

orangewolf (approved):

PR #7642, opened 2026-09-16
Memoize FlexibleSchema.current_version/current_schema_id per-request

maxkadel:
`current_version`/`current_schema_id` re-query and re-parse the YAML-serialized `profile` column on every call, uncached. Profiling a single page render in a real Hyku app (Notch8's `palni_palci_knapsack`, [notch8/palni_palci_knapsack#742](https://github.com/notch8/palni_palci_knapsack/pull/742)) found this accounting for ~20% of wall time — called a dozen+ times per render across different code paths.

Memoizes the latest record per-request via `Hyrax::Current` (`ActiveSupport::CurrentAttributes`, Rails-native since 5.2 — no new dependency, resets automatically between requests/jobs via Rails' executor). Same shape as the existing fix for Hyku's own `Site.instance` (unmerged: [samvera/hyku#3166](https://github.com/samvera/hyku/pull/3166)), which predates `CurrentAttributes` being the obvious choice and used `RequestStore` instead — worth revisiting there too.

**Validated live** under concurrent load against a real (production-cloned) dataset on staging: work-show p50 1394ms → 729ms, p90 4233ms → 3048ms. Re-validated after the `CurrentAttributes` switch specifically — confirmed it resets correctly between requests under this app's real Apartment tenant-switching.


dlpierce (commented):
This is a good optimization and implemented better than something similar IU did for allinson flex. Is there anywhere that `Hyrax::Current` will need to be explicitly reset? For instance, when a new flexible schema version is created?

maxkadel:
> This is a good optimization and implemented better than something similar IU did for allinson flex. Is there anywhere that `Hyrax::Current` will need to be explicitly reset? For instance, when a new flexible schema version is created?

The only time it should need to be explicitly reset is if someone is using it in a rails console or a rake task, where someone would need to call `Hyrax::Current.reset` if they created a new schema and then wanted to call `current_version`.

When a new schema is created, it redirects, and with a redirect `ActiveSupport::CurrentAttributes` will automatically be reset.

There is currently no job/rake task use of this code path, only views / forms / presenters / services, which are all in the request cycle.

dlpierce (approved):
Great, thanks for checking. 👍

PR #7645, opened 2026-09-16
Add config flag for indexing FileMetadata as its own Solr document

maxkadel:
`Hyrax::FileMetadata` gets its own Solr document via `MetadataIndexListener`, but `FileSetIndexer` already projects the same characterization fields onto the parent `FileSet`'s document, and nothing queries FileMetadata directly - on one production tenant this was ~55,000 redundant documents (samvera/hyku#3290).

Per review on that PR, adds `Hyrax.config.index_file_metadata_as_document?` here instead of a Hyku account setting - app-level, defaults to `true`, overridable via `HYRAX_INDEX_FILE_METADATA_AS_DOCUMENT`, same pattern as `use_valkyrie?`/`redirects_enabled?`. Open to also deprecating the indexing itself for the next major if that's the preferred direction.


orangewolf (approved):
As discussed in 9/18/2026 office hour

PR #7646, opened 2026-09-17
Fix chart HPA: autoscaling/v2beta1 was removed in Kubernetes 1.26

maxkadel:
`templates/hpa.yaml` renders `apiVersion: autoscaling/v2beta1`, which was removed from Kubernetes in 1.26 (per the [deprecation guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v1-26)). Any `helm upgrade` with `autoscaling.enabled: true` fails outright on any currently-supported Kubernetes version:

```
Error: UPGRADE FAILED: resource mapping not found for name: "..." namespace: "" from "": no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2beta1"
ensure CRDs are installed first
```

- Moves the web HPA to `autoscaling/v2`'s nested `metrics[].resource.target` schema (`v2beta1`'s flat `targetAverageUtilization` doesn't exist in `v2`).
- Adds an equivalent `worker.autoscaling` block with its own `hpa-worker.yaml`, so web and worker scale independently — useful if you want to autoscale request-serving pods without also autoscaling background job workers.
- Adds `helm-unittest` coverage for both HPAs (rendering, target API version, CPU-only and CPU+memory metrics, and that each `enabled` flag only affects its own Deployment).

Found this while debugging a production crash-loop caused by CPU/thread-pool exhaustion, where the chart's own autoscaling support turned out to be the way we'd have addressed it.


maxkadel:
Closing in favor of #7647 - reopened from a branch pushed directly to this repo instead of my fork, so CI's build/push step doesn't fail on the fork-token restriction.

PR #7647, opened 2026-09-17
Fix chart HPA: autoscaling/v2beta1 was removed in Kubernetes 1.26

maxkadel:
`templates/hpa.yaml` renders `apiVersion: autoscaling/v2beta1`, which was removed from Kubernetes in 1.26 (per the [deprecation guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v1-26)). Any `helm upgrade` with `autoscaling.enabled: true` fails outright on any currently-supported Kubernetes version:

```
Error: UPGRADE FAILED: resource mapping not found for name: "..." namespace: "" from "": no matches for kind "HorizontalPodAutoscaler" in version "autoscaling/v2beta1"
ensure CRDs are installed first
```

- Moves the web HPA to `autoscaling/v2`'s nested `metrics[].resource.target` schema (`v2beta1`'s flat `targetAverageUtilization` doesn't exist in `v2`).
- Adds an equivalent `worker.autoscaling` block with its own `hpa-worker.yaml`, so web and worker scale independently — useful if you want to autoscale request-serving pods without also autoscaling background job workers.
- Adds `helm-unittest` coverage for both HPAs (rendering, target API version, CPU-only and CPU+memory metrics, and that each `enabled` flag only affects its own Deployment).

Found this while debugging a production crash-loop caused by CPU/thread-pool exhaustion, where the chart's own autoscaling support turned out to be the way we'd have addressed it.

Also corrected the `minReplicas`/`maxReplicas`/`targetCPUUtilizationPercentage` defaults for both blocks — they were unexamined `helm create` scaffold boilerplate from 2020 (1/100/80), not considered defaults:
- `minReplicas: 1` → `2`: turning on autoscaling at all implies the workload needs elastic capacity; running a single replica defeats basic HA.
- `maxReplicas: 100` → `10`: 100 is an unrealistic ceiling for any real deployment and mainly risks runaway scale-out cost from a misbehaving metric. Still fully overridable per-deployment.
- `targetCPUUtilizationPercentage: 80` → `70`: more headroom to absorb pod-startup + readiness-probe lag during a scale-up event.

None of these are schema changes, and `autoscaling.enabled: false` by default means this is invisible to most consumers — but anyone who already runs with `autoscaling.enabled: true` and didn't override these three values will see a behavior change on upgrade (notably, a lower `maxReplicas` ceiling than before). Worth a release-note callout rather than silent.


orangewolf (approved):
thanks, this is a great fix. discussed on 9/18/2026 office hour

PR #7649, opened 2026-09-18
Stop caching zero-count analytics rows per day; prune existing bloat

maxkadel:
`file_view_stats`/`file_download_stats`/`work_view_stats` cache one row per object per day forever, even on zero-view days — on a long-running instance that's 99%+ dead weight (one tenant example: 8.85M rows for 16k rows of real activity, 75% of the whole database).

Zero-count days now collapse into a single per-object marker row (`Hyrax::Statistic#advance_zero_marker`) instead of accumulating daily; `Hyrax::FileUsage` reconstructs the zero-filled chart at render time instead of relying on the DB to hold it. 

Adds `hyrax:stats:prune_zero_stats` (`DRY_RUN=true`, `BATCH_SIZE=n`) to clean up the existing backlog without one long blocking transaction.


orangewolf (approved):
Discussed on 9/18/2026

